PR #9036, opened 2020-05-19
Add some notes on what tidy does and how to see if your cluster can handle it

ncabatoff:


vishalnayak:
This is a great addition. The only feeling that floated up during the read was that it uses technical terms like "token store", "secondary index", "accessor index". I understand that the primary job that this doc PR is doing is to explain the underlying technical workings and that these could be inevitable.

But I do wonder if we can think deliberately to alleviate some of it. For example, "reload the entirety of token store" to "reload all the existing tokens"; "secondary indexes" to "secondary entries" probably?

This is an optional comment.

ncabatoff:
> This is a great addition. The only feeling that floated up during the read was that it uses technical terms like "token store", "secondary index", "accessor index". I understand that the primary job that this doc PR is doing is to explain the underlying technical workings and that these could be inevitable.
> 
> But I do wonder if we can think deliberately to alleviate some of it. For example, "reload the entirety of token store" to "reload all the existing tokens"; "secondary indexes" to "secondary entries" probably?

My hope was to instead go in the other direction: rather than continue down the path of vagueness we've been on so far, let's expand the Internals docs to define these concepts.  That said, I did use your suggestions to slightly simplify the discussion of accessors.

vishalnayak:
Expanding the internals is a good idea IMO. We could then optionally link to those definitions wherever required. Thanks for simplifying a bit.

PR #8562, opened 2020-03-13
Enable encrypted config parameters

jefferai:
This provides a new command, `vault operator config encrypt/decrypt` that can use a defined KMS within the configuration marked for `config` purpose to encrypt or decrypt values. These values can be decrypted with a corresponding seal/kms block at runtime, so in practice, an operator that has access to the secrets can run the encrypt command, but the resulting file is then safe to pass through e.g. CI/CD or store in Git for deployment.

An example (you'd not want to actually use an AEAD KMS in real life, but this works well for demo/test):

```
listener "tcp" {
        address = "127.0.0.1:8205"
        tls_disable = true
}
telemetry {
        prometheus_retention_time = "{{encrypt(24h)}}"
        disable_hostname = true
}
storage "inmem" {
}

kms "aead" {
        purpose = "autoseal"
        key = "{{encrypt(kcI2Pyo6jZBRE9Nr7sdniVAYCRwEmOE93tw7qS0Hjq4=)}}"
        aead_type = "aes-gcm"
}

kms "aead" {
        purpose = "config"
        key = "kcI2Pyo6jZBRE9Nr7sdniVAYCRwEmOE93tw7qS0Hjq4="
        aead_type = "aes-gcm"
}
```
This becomes encrypted via the command to:
```
listener "tcp" {
        address = "127.0.0.1:8205"
        tls_disable = true
}
telemetry {
        prometheus_retention_time = "{{decrypt(Ch-feItbSZwpND15ovFKly_nFxfEnjlpz3g-f-4RVIon)}}"
        disable_hostname = true
}
storage "inmem" {
}

kms "aead" {
        purpose = "autoseal"
        key = "{{decrypt(Ckg1MovuQwkMyK4m2VKNG0WLnOjsqFxU0WDPu2KIL8_dbQWf4xAOxdG9U8suhu9zargZOo4X50IvxHeyfbT6mHgvQfaZEarnSWk)}}"
        aead_type = "aes-gcm"
}

kms "aead" {
        purpose = "config"
        key = "kcI2Pyo6jZBRE9Nr7sdniVAYCRwEmOE93tw7qS0Hjq4="
        aead_type = "aes-gcm"
}
```
At runtime Vault will pass config files, before decoding, to a function that decrypts with the given KMS.

As is apparent from the example, a nice feature of this is the ability to encrypt the other seal/KMS block parameters.

PR #9037, opened 2020-05-19
Port encrypted config shared bits to a separate PR

jefferai:
See #8562 for more information about this. This simply ports shared parts over to a separate PR.

### DIFF
--- a/internalshared/configutil/encrypt_decrypt.go
+++ b/internalshared/configutil/encrypt_decrypt.go
@@ -1,0 +1,92 @@
+package configutil
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"regexp"
+
+	"github.com/gogo/protobuf/proto"
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+)
+
+var (
+	encryptRegex = regexp.MustCompile(`{{encrypt\(.*\)}}`)
+	decryptRegex = regexp.MustCompile(`{{decrypt\(.*\)}}`)
+)
+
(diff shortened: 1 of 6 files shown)


internalshared/configutil/encrypt_decrypt.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"regexp"
+
+	"github.com/gogo/protobuf/proto"
```
briankassouf, line 11: Did you mean to use gogo here?

internalshared/configutil/encrypt_decrypt_test.go
```diff
@@ -0,0 +1,82 @@
...
+	circonus_api_key = "barfoo"
+}
+`
+
+	reverser := new(reversingWrapper)
+	out, err := EncryptDecrypt(rawStr, false, false, reverser)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
briankassouf, line 41: Should you check the encrypted output contains `{{decrypt(oofrab)}}`?

internalshared/configutil/encrypt_decrypt.go
jefferai: Sure didn't. That stuff is insidious.  Will fix.

internalshared/configutil/encrypt_decrypt_test.go
jefferai: I actually was thinking about that -- technically it contains a marshaled proto but I can just unmarshal in the test and verify. I sort of figured so long as what came in == what goes out it doesn't truly matter what it looks like in the middle 😂  But I'll add it.

PR #9038, opened 2020-05-19
Add note regarding LDAP regression.

ncabatoff:


kalafut:
Is this still needed following https://github.com/hashicorp/vault/pull/9041 ?

PR #9039, opened 2020-05-19
update deps for security

noelledaley:
This PR updates a few dependencies to resolve security alerts. These are primarily used in test and dev environments so should not need to be backported.


mladlow:
Since there won't be a backport, I'm going to change the milestone to 1.5.

PR #9052, opened 2020-05-20
Token creation counters

mgritter:
vault.token.creation counter added, both for direct creation via tokenstore and creation via an auth method.
Unit tests covering a single case for each.


vault/token_store.go
```diff
@@ -2716,6 +2717,20 @@ func (ts *TokenStore) handleCreateCommon(ctx context.Context, req *logical.Reque
 		return logical.ErrorResponse(err.Error()), logical.ErrInvalidRequest
 	}
 
+	// Count the successful token creation.
```
ncabatoff, line 2720: Do we want to include batch tokens in these metrics?

It looks like req.MountPoint is the only piece of data we don't have in `TokenStore.create`, which would be the ideal place to put this so that future changes in token creation don't accidentally skip metric updates.  Can we change the `create` signature to propagate that information there?

If not, should we also update wrapping.go, which calls `create`?

mgritter: The intent was to include batch tokens, yes.

Thanks for the pointer to response wrapping; I had not considered this as a use case but it should be included.

mgritter:
Example Telegraf output:

```
{
  "_value": 1,
  "auth_method": "token",
  "cluster": "vault-cluster-f5455742",
  "config:hecRouting": false,
  "config:multiMetric": false,
  "creation_ttl": "10m",
  "metric_name": "vault.token.creation.value",
  "metric_type": "counter",
  "mount_point": "auth/token/",
  "namespace": "root",
  "time": 1591042760,
  "token_type": "batch"
}
{
  "_value": 1,
  "auth_method": "token",
  "cluster": "vault-cluster-f5455742",
  "config:hecRouting": false,
  "config:multiMetric": false,
  "creation_ttl": "10m",
  "metric_name": "vault.token.creation.value",
  "metric_type": "counter",
  "mount_point": "ns1/auth/token/",
  "namespace": "ns1",
  "time": 1591042810,
  "token_type": "batch"
}
{
  "_value": 1,
  "auth_method": "userpass",
  "cluster": "vault-cluster-f5455742",
  "config:hecRouting": false,
  "config:multiMetric": false,
  "creation_ttl": "2h",
  "metric_name": "vault.token.creation.value",
  "metric_type": "counter",
  "mount_point": "ns1/auth/userpass/",
  "namespace": "ns1",
  "time": 1591042890,
  "token_type": "service"
}
```

vault/token_store.go
mgritter: Updating wrapping.go.

Long-term I would like to find agreement on how to pass context downstream to lower-layer functions: is it through context.Context? Individual arguments? Some new container? I didn't want to get into it right now, nor do something half-right.

With three call sites, it's right at the edge of "must solve", but I think these three can easily be refactored away if we come up with an architectural solution to the general problem.

PR #9053, opened 2020-05-20
Ui/replication mgmt action block

chelshaw:
This does some low-impact work to prepare for the refactor of replication-actions. Included:

* Move modal to addon in lib/core
* Update modal to take a "type" param, plus tests
* Add action-block style only component
* Add styles-only replication-action grid that the action-blocks will live inside of

Screen Shot 2020-05-20 at 6 18 34 PM

Note: The modal changes really only affect the header -- that being said, they don't exactly match the mocks so we will need to reconcile those changes or possibly have alternative modal stylings


ui/app/styles/components/action-block.scss
```diff
@@ -0,0 +1,58 @@
+@mixin stacked-grid {
```
Monkeychip, line 1: nice use of mixins!

ui/app/styles/core.scss
```diff
@@ -88,6 +88,8 @@
 @import './components/search-select';
 @import './components/selectable-card';
 @import './components/selectable-card-container.scss';
+// action-block extends selectable-card
```
Monkeychip, line 91: helpful comment, thank you!

ui/lib/core/addon/components/modal.js
```diff
@@ -10,12 +10,32 @@
  * @param {function} onClose - onClose is the action taken when someone clicks the modal background or close button (if shown).
  * @param {string} [title] - This text shows up in the header section of the modal.
  * @param {boolean} [showCloseButton=false] - controls whether the close button in the top right corner shows.
+ * @param {String} type=null  - The header type. This comes from the message-types helper.
```
Monkeychip, line 13: I'm not sure it matters, but should this be a lower case "string" instead of "String".  Also, it looks like maybe there's an extra space between "null  -"

ui/lib/core/addon/templates/components/modal.hbs
```diff
@@ -1,9 +1,20 @@
 {{#ember-wormhole to="modal-wormhole"}}
-  <div class="modal {{if isActive 'is-active'}}" aria-modal="true">
+  <div class="{{modalClass}} {{if isActive 'is-active'}}" aria-modal="true">
```
Monkeychip, line 2: nice!

PR #8135, opened 2020-01-10
WIP: Vault Agent: Add configuration for maximum connection attempts

catsby:
WIP

Vault Agent can be configured to attempt to authenticate with a Vault cluster that is not responsive. When this occurs, the goroutine running the Auth Method can enter an infinite loop of attempting to connect and receiving tcp dial errors like so:

```
2020-01-10T16:33:13.848-0600 [INFO]  auth.handler: authenticating
2020-01-10T16:33:13.849-0600 [ERROR] auth.handler: error authenticating: error="Put http://127.0.0.1:8200/v1/auth/approle/login: dial tcp 127.0.0.1:8200: connect: connection refused" backoff=2.496237872
```

This repeats until either Vault becomes responsive or the Agent process is terminated. Note that this is different than a failed auth attempt or responses from a sealed Vault; the dial error occurs when no connection can be made at the tcp layer. Because of that fact the existing `VAULT_CLIENT_TIMEOUT` and `VAULT_MAX_RETRIES` configuration options are never even considered, and Agent attempts to authenticate indefinitely. 

This pull request adds an additional configuration, `max-connection-attempts` which is specific to Vault Agent. This option can limit the number of failed connection attempts before exiting the Agent process. The default is `0`, which indicates "unlimited", so that the current behavior is unchanged and this change is backwards compatible. 

The maximum connection attempts can be set with the `-max-connection-attempts` command line flag, the `VAULT_MAX_CONNECTION_ATTEMPTS` environment variable, or the `max_connection_attempts` configuration variable, in that order of precedence. 

------

Remaining items:

- [ ] Clean up / add code comments
- [ ] Agent web documentation

PR #9059, opened 2020-05-21
vault agent - respect consul-template retry configuration

### LINKED ISSUE #6001: vault agent retry maximum
It would be great if vault agent would have a retry maximum like consul-template.

Currently using it in an init container, and when vault is unavailable, this blocks the entire container from starting up.

Configurable maximum retry or once-attempt like consul-template, configurable in config file or cli.

Currently sleeping + killing via pidfile, which is not very graceful.

clatour:
The consul template retry stanza from configuration was not being passed
through appropriately, this should fix that.

Since a duration is not a valid HCL type, the swap from a simple hcl
decode had to be changed to `mapstructure`.

### DIFF
--- a/command/agent/config/config.go
+++ b/command/agent/config/config.go
@@ -32,14 +32,15 @@ type Config struct {
 
 // Vault contains configuration for connnecting to Vault servers
 type Vault struct {
-	Address          string      `hcl:"address"`
-	CACert           string      `hcl:"ca_cert"`
-	CAPath           string      `hcl:"ca_path"`
-	TLSSkipVerify    bool        `hcl:"-"`
-	TLSSkipVerifyRaw interface{} `hcl:"tls_skip_verify"`
-	ClientCert       string      `hcl:"client_cert"`
-	ClientKey        string      `hcl:"client_key"`
-	TLSServerName    string      `hcl:"tls_server_name"`
+	Address          string                `hcl:"address" mapstructure:"address"`
+	CACert           string                `hcl:"ca_cert" mapstructure:"ca_cert"`
+	CAPath           string                `hcl:"ca_path" mapstructure:"ca_path"`
+	TLSSkipVerify    bool                  `hcl:"-" mapstructure:"-"`
+	TLSSkipVerifyRaw interface{}           `hcl:"tls_skip_verify" mapstructure:"tls_skip_verify"`
+	ClientCert       string                `hcl:"client_cert" mapstructure:"client_cert"`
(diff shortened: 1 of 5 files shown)


ncabatoff:
Hi @clatour,

Thanks for the PR!

I'm not too familiar with Agent's use of templating, but I suspect this may not fully address the original issue.  A brief look at the code suggests that the consul-template lib's connection to Vault is managed independently of Agent's auto-auth code.  So even if you make consul-template give up trying to talk to Vault via this config, Agent itself would still be spinning its wheels trying to talk to Vault.  There's an incomplete PR (#8135) that tried to address that other bit.

Have you tested that this fix solves #6001 for you?

clatour:
Hi @ncabatoff, on second glance, I think I may have linked the wrong issue. This patch _does not_ alleviate the retry issue when the host is invalid or unreachable, only respects the retry configuration block for the vault agent configuration.

avoidik:
hi @clatour there are two issues at the moment with retry-logic in vault agent:
1. an issue when the consul rendering server inside the vault agent service have failed to render templates and was terminated, but vault-agent service itself left in a half-working state (e.g. auto-auth works, rendering templates don't), and the only solution is to restart it
2. vault not available due to sealed, not initialized, so vault agent pushes the maximum exceeded attempts also switching in to a half-working state where the only solution is to restart it

clatour:
Hi @avoidik,

I have edited my initial comment (and title) to remove the link to the outstanding vault agent retry issue. This is solely a patch to respect some configuration that is supported by consul template, and was documented as supported in vault-agent, but did not look to be wired through.

Currently the only instance in which I am using vault-agent is rendering templates as part of a deployment, where I’d like vault-agent to quickly set up, render template(s), and tear down. Similar to the idea of `consul-template`s `once` flag. So I’m not super familiar with any of the other retry mechanisms that exist when running vault-agent as a daemon.

What this solves for my use case is largely when a secret doesn’t exist or the template can’t otherwise be rendered, I’d like the `consul-template` component to allow itself to be configured to fail faster. Which I think is more related to your first point, but as-implemented would just cause the template rendering to be able to be shut down after a configurable retry, and not actually helping the retry/restart of that component overall.

VioletHynes:
As this PR is older and cannot currently be merged due to merge conflicts, I’m going to go ahead and close it. Please feel free to open a new PR if your desired changes are still applicable!

However, additionally, I think there's a good chance the problem you're trying to solve was solved here: https://github.com/hashicorp/vault/pull/16970

Issue #9001, opened 2020-05-14
cli: operator raft join should accept file paths for TLS parameters


calvn:
**Is your feature request related to a problem? Please describe.**
The flags related to TLS parameters currently only accepts data values as literal single-line strings which can be cumbersome to provide.

**Describe the solution you'd like**
The `vault operator raft join` command should be able to accept file paths for TLS parameters.

**Describe alternatives you've considered**
Pass in the data as a single-line string value for each of the TLS flags.

**Additional context**
Related to #8753
Related to #8894

PR #9060, opened 2020-05-21
Support reading Raft TLS flags from file

alexanderbez:
Support reading leader TLS flags from file in `vault operator raft join`. Not sure how/if we can test these changes directly. Let me know if there is a way to do this.

closes: #9001


command/operator_raft_join.go
```diff
@@ -139,10 +167,25 @@ func (c *OperatorRaftJoinCommand) Run(args []string) int {
...
+	out := []string{
+		"Key | Value",
+		fmt.Sprintf("Joined | %t", resp.Joined),
+	}
 	c.UI.Output(tableOutput(out, nil))
 
 	return 0
 }
+
+func parseArg(raw string) (string, error) {
```
calvn, line 179: Since they are technically flag values and not arg values, maybe rename it to `parseFlagVal` or something along those lines. We could also move this to `base_helper.go` (and maybe include a unit test for it).

vishalnayak:
Looking great! I can get back to looking at this again once Calvin's comment is addressed and the testing is successful.

command/operator_raft_join.go
alexanderbez: Good idea -- updated

command/base_helpers.go
```diff
@@ -273,3 +275,20 @@ func humanDurationInt(i interface{}) interface{} {
 	// If we don't know what type it is, just return the original value
 	return i
 }
+
+// parseFlagFile accepts a flag value returns the contets of that value. If the
+// value starts with '@', that indicates the value is a file and it's contets
```
calvn, line 280:
```suggestion
// value starts with '@', that indicates the value is a file and its content
```

command/base_helpers_test.go
```diff
@@ -160,3 +160,52 @@ func TestTruncateToSeconds(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.value, func(t *testing.T) {
+			content, err := parseFlagFile(tc.value)
+			if err != nil {
+				t.Errorf("unexpected error parsing flag value: %v", err)
```
calvn, line 203: Prefer `t.Fatalf` over here to fail fast since an error here would result in an error on the content check anyways.

PR #9061, opened 2020-05-21
Replication Management Sidebranch: Replication Action Disable

chelshaw:
This PR does a few things:

- Add `confirmation-modal` component that automatically disables the confirmation button until some text is entered
- Create manage page where the `replication-dr-secondary` splash page once was
- Refactor the first `replication-action-${action}` block
- Add tabbed links to DR replication details page

The interface for how the action blocks will be updated. Screenshot after changes of Primary and Secondary, repectively:
Screen Shot 2020-05-21 at 1 38 15 PM
Screen Shot 2020-05-21 at 1 38 24 PM

And here are the confirmation modals for each of the replication configurations
1) Performance / Primary
repl-perf-prim

2) Performance / Secondary
repl-perf-sec

3) DR / Primary
repl-dr-prim

4) DR / Secondary
repl-dr-sec 

TODO: 
[x] Add/fix tests for action-block-disable
[/] Add/fix tests for confirmation-modal


ui/app/templates/vault/cluster/replication-dr-promote/details.hbs
```diff
@@ -1,44 +1,44 @@
 <section class="section">
   <div class="container is-widescreen">
     <ReplicationPage @model={{model}} as |Page|>
-        <Page.header
-          @showTabs={{false}}
+      <Page.header
+        @showTabs={{true}}
```
chelshaw, line 5: This line was the only change, the rest is whitespace

ui/lib/core/addon/templates/components/replication-action-disable.hbs
```diff
@@ -1,67 +1,45 @@
-<h4 class="title is-5 is-marginless">
-  Disable Replication
-</h4>
-<div class="content">
-  <p>
-    Disable {{replicationDisplayMode}} Replication entirely on the cluster.
-    {{#if model.replicationAttrs.isPrimary}}
```
chelshaw, line 7: Matching the designs for this component means that much of this additional context was lost. We may want to add some of this back into the confirmationModal

ui/lib/core/addon/templates/components/replication-action-disable.hbs
```diff
@@ -1,67 +1,45 @@
...
+    "disable"
+    (if
+      (eq model.replicationAttrs.modeForUrl 'bootstrapping')
+      mode
+      model.replicationAttrs.modeForUrl
+    )
+  }}
+>
+  <p>
+    Disabling Performance replication entirely on this secondary cluster means that:
```
noelledaley, line 35: looks like the content here needs to be swapped out depending on whether it's a primary or secondary -- which you should have access to on the `model`, or perhaps from this `clusterMode` property? https://github.com/hashicorp/vault/blob/5a75ef2f7aae44b7a94bdd96738dc074efbef955/ui/app/templates/vault/cluster/replication-dr-promote/index.hbs#L34

looks like that property needs to be added to the list of properties inside the `ReplicationActions` component too: https://github.com/hashicorp/vault/blob/f60d159674d64aa8b58ca3bcf51f6d54d4eacef3/ui/lib/core/addon/components/replication-actions.js#L20

ui/lib/core/addon/templates/components/confirmation-modal.hbs
```diff
@@ -0,0 +1,44 @@
...
+        <h4 class="has-text-weight-semibold is-size-6">
+          Confirm
+        </h4>
+        <p class="is-help">Type <strong>{{confirmText}}</strong> to confirm {{toConfirm}}</p>
+        {{input
+            type="text"
+            value=confirmationInput
+            name="confirmationInput"
+            class="input has-margin-top"
+            autocomplete="off"
```
noelledaley, line 23: very nice ⚡

ui/lib/core/addon/templates/components/replication-action-disable.hbs
```diff
@@ -1,67 +1,45 @@
...
+    </button>
   </div>
 </div>
+
+<ConfirmationModal
+  @title="Disable Replication?"
+  @onClose={{action (mut isModalActive) false}}
+  @isActive={{isModalActive}}
+  @confirmText={{if (eq replicationDisplayMode "DR") "Disaster Recovery" replicationDisplayMode}}
+  @toConfirm="disabling {{replicationDisplayMode}} replication on this cluster"
```
noelledaley, line 23: nit: mind renaming this `toConfirmMsg` or something similar? at first pass i expected this property to be a route to transition to or action to perform after confirming :o

ui/lib/core/addon/templates/components/replication-actions.hbs
```diff
@@ -2,14 +2,29 @@
...
+      </div>
+    {{/each}}
+  {{else}}
+  <div class="replication-actions-grid-layout">
+    {{#each (if clusterMode (replication-action-for-mode replicationMode clusterMode) (replication-action-for-mode replicationMode model.replicationAttrs.modeForUrl)) as |replicationAction index|}}
+        <div class="replication-actions-grid-item">
+          {{component (concat 'replication-action-' replicationAction)
+            onSubmit=(action "onSubmit")
+            replicationMode=replicationMode
+            model=model
```
noelledaley, line 23: oh yeah! looks like you can pass in the `clusterMode` here!

ui/lib/core/addon/templates/components/replication-actions.hbs
```diff
@@ -2,14 +2,29 @@
...
+          onSubmit=(action "onSubmit")
+          replicationMode=replicationMode
+          model=model
+          replicationDisplayMode=replicationDisplayMode
+        }}
+      </div>
+    {{/each}}
+  {{else}}
+  <div class="replication-actions-grid-layout">
+    {{#each (if clusterMode (replication-action-for-mode replicationMode clusterMode) (replication-action-for-mode replicationMode model.replicationAttrs.modeForUrl)) as |replicationAction index|}}
```
noelledaley, line 18: hmm, this conditional hurts my 🧠a little bit -- is there any way to make this a computed property, that way you can just do ``{{#each replicationAction as |replicationAction index|}}` or something similar?

ui/lib/core/addon/components/confirmation-modal.js
```diff
@@ -0,0 +1,34 @@
...
+import Component from '@ember/component';
+import layout from '../templates/components/confirmation-modal';
+
+export default Component.extend({
+  layout,
+  buttonClass: 'is-danger',
+  buttonText: 'Confirm',
+  confirmText: 'Yes',
+  type: 'warning',
+  actionDescription: '',
```
noelledaley, line 33: what does this property do? should this be added to the JSDoc comments above?

noelledaley:
left a couple small comments/suggestions, otherwise this looks awesome! the one failing UI test is due to a linting error and should be an easy fix. nice work!

ui/lib/core/addon/templates/components/replication-action-disable.hbs
chelshaw: Great catch, I've updated the content per Ivana's mockups!

chelshaw: I was actually able to use the values that were already here, but I'd love a sanity check on the logic: 

- Only DR Secondaries would need to pass `clusterMode` explicitly
- Everywhere else `replicationAttrs` is available on the model, which has an `isPrimary` value

SO my assumption is that if `isPrimary` is not true, the node is a secondary.  Or, we can do that logic from the `replication-actions` file and pass it in explicitly 🤔

ui/lib/core/addon/templates/components/replication-actions.hbs
chelshaw: Updating this one now, or maybe even a reformat would help a bit!

ui/lib/core/addon/templates/components/confirmation-modal.hbs
```diff
@@ -0,0 +1,46 @@
...
+      {{buttonText}}
+    </button>
+    <button
+      class="button is-secondary"
+      onclick={{action (mut isActive) false}}
+      data-test-cancel-button>
+      Cancel
+    </button>
+  </footer>
+</Modal>
```
Monkeychip, line 46: missing extra space at the bottom of file.

ui/lib/core/addon/templates/components/replication-actions.hbs
```diff
@@ -2,14 +2,33 @@
...
+            onSubmit=(action "onSubmit")
+            replicationMode=replicationMode
+            model=model
+            replicationDisplayMode=replicationDisplayMode
+          }}
+        </div>
+    {{/each}}
+  </div>
+  {{/if}}
+{{/if}}
```
Monkeychip, line 34: missing extra space here too

ui/lib/core/addon/templates/components/replication-action-disable.hbs
noelledaley: regarding #1, yeah, it's strange that DR secondaries seem to be the only place where you have to explicitly pass it in. could we refactor it so you don't have to? that way you could get rid of the conditional here: https://github.com/hashicorp/vault/pull/9061/files#diff-6cae373607effdd513024f7145a722f3R18

for #2, it looks like `replicationAttrs` should also have a property called `isSecondary`, and so my hunch would be to use that value explicitly, however upon testing this in a DR secondary i noticed that the actual value of `isSecondary` might be incorrect? i'm not sure why this would be -- i got this value by inspecting the cluster model, then console logging the `replicationAttrs`. mind taking a look into that?

image

noelledaley: (also lol at those links above linking to some of the first vault PRs!!!)

ui/tests/integration/components/confirmation-modal-test.js
```diff
@@ -0,0 +1,34 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import sinon from 'sinon';
+import { fillIn, find, render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
+module('Integration | Component | confirmation-modal', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it works as expected', async function(assert) {
```
noelledaley, line 10: this test description makes me chuckle, but we should probably update it to describe what it's actually testing 😉

something a la `it renders with a disabled confirm button` or similar

ui/tests/integration/components/replication-actions-test.js
```diff
@@ -111,16 +121,32 @@ module('Integration | Component | replication actions', function(hooks) {
...
       );
-
-      assert.equal(find('h4').textContent.trim(), headerText, `${testKey}: renders the correct component`);
 
       if (typeof fillInFn === 'function') {
         await fillInFn.call(this);
       }
-      await click('[data-test-confirm-action-trigger]');
-      await click('[data-test-confirm-button]');
+      if (oldVersion) {
```
noelledaley, line 138: nice thinking on these tests! hopefully we can delete the old versions very soon ಠ‿ಠ

ui/lib/core/addon/templates/components/replication-action-disable.hbs
chelshaw: So, it turns out I didn't have to pass it `clusterMode`, we have the same cluster properties passed down from DR secondary as the others! 🎉 Thanks for the 🦅👁️ on that!

As far as using `isSecondary` explicitly, if there are only two modes it doesn't really matter which one we check for and which is the `else`... but, if there is another mode I'm not aware of (or might be in the future!) that would be helpful information. I'm going to go ahead and close out this comment since the code itself is outdated, but thanks again for the thorough feedback!

noelledaley:
that sounds like a solid plan -- i don't foresee any additional modes coming along. thanks for making all the other fixes, too!

PR #9079, opened 2020-05-26
Ui/summary dashboard

Monkeychip:
This PR adds the replication summary dashboard.  This dashboard is a summary view of when you have both a DR primary and Performance Primary enabled on the same cluster.  This is the only summary dashboard.  

The dashboard lives on the replication.index route, which is used to display the following when both DR and Performance primaries were enabled.

image

This is what it looks like with the changes:
image

Screen Recording 2020-05-27 at 04 47 PM

Notes:
- We use the computed property `replicationDetails` to fill in much of what goes into the dashboard.  However, replicationDetails is usually just the this.model.[dr or performance] information.  In this case, were we need both.  To solve for this, I pass only to the new `replication-card-summary` component a property called `replicationDetailsSummary` which is an object with both the dr and performance replication details.  I then parse through this object in the replication-summary-card js file. 

- we moved `key-value-header` component into the addons folder so I could use it in the replication engine.

- Added acceptance and component tests.


ui/lib/core/addon/components/replication-page.js
```diff
@@ -29,22 +30,45 @@ export default Component.extend({
...
   }),
+  isSummaryDashboard: computed('model.dr.{mode}', 'model.performance.{mode}', function() {
+    const router = this.router;
+    const currentRoute = router.get('currentRouteName');
+
+    // we only show the summary dashboard in the replication index route
+    if (currentRoute === 'vault.cluster.replication.index') {
+      const drMode = this.model.dr.mode;
+      const performanceMode = this.model.performance.mode;
+      return drMode && performanceMode === 'primary';
```
chelshaw, line 41: Do you want to check if the dr mode is also primary? In that case this line should be `return drMode === 'primary' && performanceMode === 'primary'`

ui/lib/core/addon/components/replication-page.js
```diff
@@ -29,22 +30,45 @@ export default Component.extend({
...
+    }
+  }),
   formattedReplicationMode: computed('model.{replicationMode}', function() {
```
chelshaw, line 44: Do you need to also compute based off `isSummaryDashboard`? Can you compute from a computed property? 🤔

ui/lib/core/addon/components/replication-page.js
```diff
@@ -29,22 +30,45 @@ export default Component.extend({
...
     // dr or performance 🤯
+    const { isSummaryDashboard } = this;
+    if (isSummaryDashboard) {
+      return 'Disaster Recovery & Performance';
+    }
     const mode = this.model.replicationMode;
     return MODE[mode];
   }),
-  clusterMode: computed('model.{replicationAttrs}', function() {
+  clusterMode: computed('model.{replicationAttrs}', 'isSummaryDashboard', function() {
```
chelshaw, line 53: Ah, i see here you can 🤓

ui/lib/core/addon/components/replication-summary-card.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @example
+ * ```js
+ * <ReplicationSummaryCard
+    @title='States'
+    @replicationDetails=replicationDetails
+    />
+ * ```
+ * @param {string} [title=null] - The title to be displayed on the top left corner of the card.
+ * @param replicationDetails=null{DS.Model.replicationDetails} - An Ember data object off the Ember data model.  It is computed at the parent component and passed through to this component.
```
chelshaw, line 17: The formatting of this one looks a little funky 🤔 is that right?

ui/lib/core/addon/templates/components/replication-dashboard.hbs
```diff
@@ -9,16 +9,41 @@
...
-    {{yield (hash
-      card=(component componentToRender replicationDetails=replicationDetails)
-    )}}
-    {{#if (not isSecondary)}}
-      {{yield (hash secondaryCard=(component 'known-secondaries-card'))}}
-    {{/if}}
-  </div>
+  {{#if isSummaryDashboard}}
+    <div class="summary-state">
+      <h6 class="title is-5 {{if (get (cluster-states summaryState) "isOk") "" "has-text-danger"}}" data-test-error>state</h6>
```
chelshaw, line 14: Very small nit, but this ternary could be rewritten as
`{{unless (get (cluster-states summaryState) "isOk") "has-text-danger"}}`

ui/lib/core/addon/templates/components/replication-dashboard.hbs
```diff
@@ -9,16 +9,41 @@
...
+        <p class="has-text-grey is-size-8">The current operating state of both Disaster Recovery and Performance. This value should be ‘running.’ If the value is idle, it indicates an issue and needs to be investigated.</p>
+      {{/unless}}
+      <h2 class="title is-3" data-test-summary-state>{{summaryState}}
+        {{#if (get (cluster-states summaryState) "isOk")}}
+          <Icon @glyph={{get (cluster-states summaryState) "glyph"}} class="{{if (get (cluster-states summaryState) "isOk") "has-text-success" "has-text-danger"}}" data-test-icon />
+        {{/if}}
+      </h2>
+    </div>
+    <div
+      class="selectable-card-container summary secondary"
```
noelledaley, line 29: does this need the `secondary` class? 'cause it might not necessarily be a secondary, right?

ui/lib/core/addon/templates/components/replication-dashboard.hbs
```diff
@@ -9,16 +9,41 @@
...
+      data-test-selectable-card-container-summary>
+        {{yield (hash
+          card=(component componentToRender replicationDetails=replicationDetailsSummary)
```
noelledaley, line 32: i wonder if in the case of the summary dashboard we should just hardcode the `componentToRender` here -- it's a little confusing to have to go back and find out what component we're passing in here. the downsides of contextual components (ᵟຶ︵ ᵟຶ)

ui/lib/core/addon/components/replication-header.js
```diff
@@ -3,7 +3,9 @@ import layout from '../templates/components/replication-header';
 
 export default Component.extend({
```
noelledaley, line 4: would you be willing to add JSDoc comments for this component? there are a lot of very similar properties in here and i think it'll help us later to remember what each thing is.

ui/lib/core/addon/components/replication-page.js
```diff
@@ -12,6 +12,7 @@ const MODE = {
 export default Component.extend({
```
noelledaley, line 12: would you be willing to add JSDoc comments for this component? there are a lot of very similar properties in here and i think it'll help us later to remember what each thing is.

ui/lib/core/addon/components/replication-summary-card.js
noelledaley: nice catch! chelsea is right, the format should match the example in the README here:
https://github.com/hashicorp/vault/blob/dd81d580622050357f3e2ec90e51d30c77cb99f4/ui/README.md#L157

ui/lib/core/addon/components/replication-summary-card.js
```diff
@@ -0,0 +1,52 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import layout from '../templates/components/replication-summary-card';
+
+/**
+ * @module ReplicationSummaryCard
+ * ReplicationSummaryCard components
```
noelledaley, line 7: looks like this needs the full description 😉

ui/lib/core/addon/components/replication-summary-card.js
noelledaley: while you're at it, i'd recommend updating this description to include a little more about what the contents of `replicationDetails` are. what information does it provide?

ui/lib/core/addon/components/replication-summary-card.js
```diff
@@ -0,0 +1,52 @@
...
+ * ```
+ * @param {string} [title=null] - The title to be displayed on the top left corner of the card.
+ * @param replicationDetails=null{DS.Model.replicationDetails} - An Ember data object off the Ember data model.  It is computed at the parent component and passed through to this component.
+ */
+
+export default Component.extend({
+  layout,
+  title: null,
+  replicationDetails: null,
+  lastDrWAL: computed('replicationDetails.dr.{lastWAL}', function() {
```
noelledaley, line 24: i haven't tested this syntax, but do you know if this might need to be recomputed like `replicationDetails.{dr}.{lastWAL}` to ensure that the property is updated when dr changes too?

ui/lib/core/addon/components/replication-summary-card.js
```diff
@@ -0,0 +1,52 @@
...
+    return this.replicationDetails.dr && this.replicationDetails.dr.lastWAL
```
noelledaley, line 25: i think all of the conditionals in this component can be refactored to `return this.replicationDetails.performance.lastWAL|| 0`

ui/lib/replication/addon/templates/components/replication-summary.hbs
```diff
@@ -262,69 +262,90 @@
...
       {{/if}}
     </div>
-  {{/unless}}
+      {{#unless (and submit.isRunning (eq cluster.dr.mode "bootstrapping"))}}
+      <div class="box is-bottomless is-fullwidth is-marginless">
+        <h3 class="title is-flex-center is-5 is-marginless">
+          <Icon @size="xl" aria-hidden="true" @glyph="perf-replication" />
+          Performance
+        </h3>
+        {{#if cluster.dr.replicationEnabled}}
```
noelledaley, line 327: is this the right conditional here? it looks like it's checking for DR but then rendering links to performance 🤔

noelledaley:
ooh, another fix i just noticed -- looks like the dashboard is making a bad request to the `replication/${mode}/status`. we probably don't want to make this request until we have the modes on the summary page.

ui/lib/core/addon/components/replication-page.js
Monkeychip: Nice catch!  This one was a slippery one to catch because we only compute this property when the summary dashboard shows, which only happens when both are a primary so by default it was correctly returning true, but not because drMode was a primary, but because peroformanceMode always was.  I'll tag @noelledaley  as I believe she wrote this.

ui/lib/core/addon/templates/components/replication-dashboard.hbs
Monkeychip: Hmm.. I agree, but I also don't like that this would be a one-off, where we use componentToRender elsewhere.  I think I would rather change all of them, instead of just the one, or keep it as is.  Thoughts?

ui/lib/core/addon/components/replication-summary-card.js
Monkeychip: Good catch.  This was missing on both the ReplicationSecondaryCard and ReplicationPrimaryCard components as well, so I went ahead and added them.

ui/lib/core/addon/components/replication-summary-card.js
Monkeychip: I looked at the documentation, and suspect, though can't confirm that the double {} is not valid.  You can watch multiple properties, like this `replication.dr.{lastWAL, clusterId}`, but both the dr object, and the lastWAL property.  

However, lastWAL changes frequently, and that's the property we care about for the computed `lastDrWAL`.  Additionally, this is the same as watching as we do in the other dashboards, `replicationDetails.{lastWAL}`, it's just our object has another nested object inside of it.  The lastWAL will change whenever we enable, disabled, or make any cluster changes (adding a policy, token, etc.).

I'll do some testing today to make sure these computed properties change as we expect them to.

ui/lib/replication/addon/templates/components/replication-summary.hbs
Monkeychip: Wow, OK this was logic I moved and pulled over (e.g. I didn't write this, I just moved it).  _However_, it does not make any sense.  This is the functionality of what shows if I remove the conditional.  Notice the first one is a clickable box, and the second one is not. I think I can safely remove this logic and just let the box be clickable, _but_ it's a little outside of the scope.  Thoughts?

conditional

Monkeychip:
> ooh, another fix i just noticed -- looks like the dashboard is making a bad request to the `replication/${mode}/status`. we probably don't want to make this request until we have the modes on the summary page.

Good catch. I fixed this with a conditional in the replication-page which is where we are calling the adapter (replication-mode) that is making this request.

ui/lib/core/addon/components/replication-page.js
Monkeychip: Sure thing.  However, the replication-page only takes in the cluster model, so the write up is pretty small.  I'll do the same for the replication-dashboard, which takes in a couple more properties.

ui/lib/core/addon/templates/components/replication-dashboard.hbs
noelledaley: oh yeah, i think changing the rule to apply to `&.secondary, &.summary { //grid stuff }` would be a great fix. that way the classes of the element accurately describe what it is.

ui/lib/core/addon/templates/components/replication-dashboard.hbs
noelledaley: yeah, that totally makes sense! i agree that it'd be better to change all of 'em vs just this one. in a future refactor it might be nice to update the dashboard template to something like:

``
if summary dashboard
  show summary stuff
else if primary dashboard
  show primary stuff
else if secondary stuff
  show secondary stuff
```

this would eliminate some of the abstraction (and also flexibility) within the dashboard template which could be helpful for debugging. just something to consider for later if you think it'd help & we have time.

ui/lib/replication/addon/templates/components/replication-summary.hbs
noelledaley: woah, that is bizarre 😳 yeah, i'd try removing the conditional and seeing if the tests still pass. additionally, i'd check to make sure the box correctly show/hides when DR is disabled & performance is enabled, when both are enabled, when both are disabled etc.

PR #9081, opened 2020-05-26
updated usage example

devops-rob:
I changed the code in the docs for the transit secrets engine example because it assumes that the user is using a bash shell where the <<< redirect will work.  I went through the example using the docker image that we publish and that doesn't use a bash shell so it didn't work.

I think using echo and piping into base64 command is much more universal and will work across all of the shells


aphorise:
IMO - request is reasonable - as opposed to using [here strings (`<<<`)](https://www.landoflinux.com/linux_bash_scripting_here_strings.html) the `echo` is more readable especially for OS without it.

mladlow (approved):
On the one hand, I feel like this is very subjective. On the other hand I personally find here strings very confusing and would prefer the new syntax. @devops-rob I think you might need to update this PR from main before it can be merged.

taoism4504 (approved):

PR #9085, opened 2020-05-27
Test pre-1.4 seal migration 

mjarmy:
This PR refactors the seal migration code into `vault.Core` so it's easier to test, and rewrites the seal migration tests for pre-1.4 shamir->transit and transit->shamir so that the tests simulate bringing the cluster down to do the migration.

In addition, there are a couple of minor changes:
* Add a few minor improvements to the reusable storage test suite.
* Remove the test that existed solely to exercise reusable storage.


vault/core.go
```diff
@@ -2224,7 +2230,113 @@ func (c *Core) PhysicalSealConfigs(ctx context.Context) (*SealConfig, *SealConfi
...
+		newRecoveryConfig.StoredShares = 0
+		newSeal.SetCachedRecoveryConfig(newRecoveryConfig)
+	}
+
+	c.setSealsForMigration(migrationSeal, newSeal, unwrapSeal)
+
+	return nil
+}
+
+func (c *Core) setSealsForMigration(migrationSeal, newSeal, unwrapSeal Seal) {
```
vishalnayak, line 2339: It feels like this function can be pulled into the `adjustForSealMigration` function. Since we are now doing this in NewCore, I wonder if we need to grab the `stateLock` anymore.

mjarmy: `setSealsForMigration()` gets called from two different places in `adjustForSealMigration`, so I left it as its own method.  Its true though that there is no need for the state lock, so I pushed up a commit that removes it.

vishalnayak: Sounds good.

vault/external_tests/seal_migration/seal_migration_pre14_test.go
```diff
@@ -0,0 +1,134 @@
+// +build !enterprise
+
+package seal_migration
```
calvn, line 3: May I suggest using `sealmigration` as the package name instead? [According to this blog post](https://blog.golang.org/package-names), package names shouldn't have `_` (or be `camelCased`).

vault/external_tests/seal_migration/seal_migration_test.go
```diff
@@ -0,0 +1,517 @@
...
+	// Unseal
+	if storage.IsRaft {
+		testhelpers.RaftClusterJoinNodes(t, cluster)
+		if err := testhelpers.VerifyRaftConfiguration(leader, numTestCores); err != nil {
+			t.Fatal(err)
+		}
+	} else {
+		cluster.UnsealCores(t)
+	}
+	testhelpers.WaitForNCoresUnsealed(t, cluster, numTestCores)
```
calvn, line 275: It seems like this would time out if the cluster's storage is Raft since we're not unsealing for that case.

calvn: Wrote too quick, looks like `raftClusterJoinNodes` also performs unsealing.

vault/external_tests/seal_migration/seal_migration_pre14_test.go
mjarmy: good idea, done

vault/external_tests/seal_migration/seal_migration_test.go
```diff
@@ -0,0 +1,517 @@
...
+		cluster.Cleanup()
+	}()
+
+	leader := cluster.Cores[0]
+	client := leader.Client
+
+	// Unseal
+	if storage.IsRaft {
+		testhelpers.RaftClusterJoinNodes(t, cluster)
+		if err := testhelpers.VerifyRaftConfiguration(leader, numTestCores); err != nil {
```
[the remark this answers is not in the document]
mjarmy: What I'd like to do then is keep `numTestCores`, with its setting of 5, and use it to set `opts.NumCores` as we are already doing.  But everywhere else in the tests I will use `len(cluster.Cores)`.  I believe this is in consonance with what you are proposing, while at the same time sticking with 5 for the number of test cores to prove that we haven't hardwired the default value of 3 anywhere in the tests.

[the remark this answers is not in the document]
calvn: Is there a specific reason to use 5 vs 3 (the `vault.DefaultNumCores` value) for a multi-cluster setup?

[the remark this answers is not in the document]
mjarmy: The specific reason is that I want to make sure I don't accidentally hard-wire the `ReusableStorage` stuff to use the default 3 cores, and since this is the only place right now in the code base that is using reusable storage, its the only place I can do that.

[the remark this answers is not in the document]
vishalnayak: Keeping in mind the `simplify use of numTestCores` commit (https://github.com/hashicorp/vault/pull/9085/commits/7271985fc6d797ba40e365c9a60475259039c399), which addresses some parts of Calvin's suggestion, I think the changes looks good.

The `numTestCores` variable as it is used now is only applicable to the seal migration package and is only used in places where the full cluster is not present. So, we can't do `len(cluster.Cores)`. Also, the reason quoted by Mike to have 5 cores seems valid as it is walking a case that ensures creation of necessary number of reusable storage artifacts.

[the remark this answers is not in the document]
calvn: The only remaining usage of `numTestCores` seems to be on here:
https://github.com/hashicorp/vault/blob/7271985fc6d797ba40e365c9a60475259039c399/vault/external_tests/sealmigration/seal_migration_pre14_test.go#L80

My suggestion was not to use `len(cluster.Cores)` in that instance since `opts.NumCores` is the field that determines the number of cores the cluster should have, but rather re-use the const value, `vault.DefaultNumCores`, that we already have for multi-core cluster setups.

PR #9087, opened 2020-05-27
CI - go mod test

sarahethompson:
This makes use of Jeff's changes in - https://github.com/hashicorp/vault/pull/9000
to allow us to lock build-time dependencies to specific versions in CI as well (https://hashicorp.atlassian.net/browse/VLTES-102).
We will hopefully be able to move to Artifactory for our GOPROXY soon in CI once the Artifactory server is upgraded to fix - https://www.jfrog.com/jira/browse/RTFACT-19395.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,7 +86,7 @@ jobs:
           mkdir ./pkg
 
           # Build dev binary
-          make ci-bootstrap dev
+          make bootstrap dev
         name: Build dev binary
     - persist_to_workspace:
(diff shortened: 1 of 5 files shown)


samsalisbury:
@sarahethompson there are failing tests on this branch, and on master, however they are not the same tests. I stil don't think your changes would have caused them, but I just pulled the latest from master into this branch to check we now get the same failing tests, assuming we do then it's ok to merge, thanks for doing this!

PR #9089, opened 2020-05-27
Initial 1.5.0 Upgrade Doc

alexanderbez:
The aim of this PR is to be a (temporary) living PR that contains the necessary upgrade instructions for the upcoming Vault 1.5.0 release.


website/pages/docs/upgrading/upgrade-to-1.5.0.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+# Overview
+
+This page contains the list of deprecations and important or breaking changes
+for Vault 1.4.x compared to 1.5.0. Please read it carefully.
+
+## Raft Configuration
+
+A new Raft configuration value, `max_entry_size`, has been introduced. This value
+limits the size in bytes for a Raft entry. It applies to both Put operations and
```
mladlow, line 18: If this suggestion is correct I think it would be helpful to clarify.

As far as Put vs put, I think we should be consistent with the capitalization, but I don't much care which we use.

```suggestion
limits the size in bytes for a Raft K/V entry. It applies to both put operations and
```

website/pages/docs/upgrading/upgrade-to-1.5.0.mdx
```diff
@@ -0,0 +1,25 @@
...
+transactions. Any put or transaction operation exceeding this configuration value
+will cause the respective operation to fail. The default value for this
+configuration is 1048576 -- two times the chunking size.
+
+In addition, a new metric has been introduced, `entrySize`, that allows for
+operators to view the average entry size and adjust the configuration value as
```
mladlow, line 24: I'm not entirely sure this is correct to the implementation but I think something along these lines better reflects what we talked about this morning.
```suggestion
operators to sample the entry size, view the average, and adjust the configuration value as
```

alexanderbez:
> Could you link to the docs for the config value and the metric? Otherwise I think this looks good, but I plan to wait to approve the PR generally until closer to 1.5, as I expect to continue looking at this PR.

Yes, good idea.

website/pages/docs/upgrading/upgrade-to-1.5.0.mdx
```diff
@@ -0,0 +1,25 @@
...
+This page contains the list of deprecations and important or breaking changes
+for Vault 1.4.x compared to 1.5.0. Please read it carefully.
+
+## Raft Configuration
+
+A new Raft configuration value, `max_entry_size`, has been introduced. This value
+limits the size in bytes for a Raft K/V entry. It applies to both put operations and
+transactions. Any put or transaction operation exceeding this configuration value
+will cause the respective operation to fail. The default value for this
+configuration is 1048576 -- two times the chunking size.
```
briankassouf, line 21: I’m not sure we need to mention chunking size here as it’s more of an internal detail

website/pages/docs/upgrading/upgrade-to-1.5.0.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+In addition, a new metric has been introduced, `entrySize`, that allows for
```
briankassouf, line 23: Can you give the full metric name? vault.raft_storage...

website/pages/docs/upgrading/upgrade-to-1.5.0.mdx
briankassouf: Also can you convert this to mb?

website/pages/docs/upgrading/upgrade-to-1.5.0.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+## Raft Configuration
+
+A new Raft configuration value, `max_entry_size`, has been introduced. This value
+limits the size in bytes for a Raft K/V entry. It applies to both put operations and
+transactions. Any put or transaction operation exceeding this configuration value
+will cause the respective operation to fail. The default value for this
+configuration is 1MiB.
+
+In addition, a new metric has been introduced, `vault.raft-storage.entrySize`,
```
calvn, line 23: The upgrade page usually only contains changes around breaking or deprecated behavior. Although this bit might be useful, I wonder if it's okay to leave it out since that's also in the metrics doc page and in the changelog.

alexanderbez: Thoughts @mladlow?

mladlow: If I were reading this as an operator, my first question would be "will this change cause problems if I upgrade?" and I'd want to know how to answer that question. So, I think it's a nice convenience for someone like that to know where they can go without having to find the right metric. It can be a little difficult to find the metric you're looking for on the metrics page, I have found.

briankassouf: Maybe we should put the metric into 1.4.3 then? That way someone upgrading from that version or greater can check the metric before upgrading

mladlow: Oh, that would be good. @alexanderbez would you mind making a backport PR for the metric to release/1.4.x?

alexanderbez: Yeah, good idea. Will create (my first) backport PR 👍

alexanderbez: https://github.com/hashicorp/vault/pull/9105

PR #9091, opened 2020-05-27
Provide token ttl and issue time in the audit log.

sgmiller:
Populate a token_ttl and token_issue_time field on the Auth struct of audit log entries, and in the Auth portion of a response for login
methods.


ncabatoff:
Curious why you left out the other LeaseOptions fields?

audit/format.go
```diff
@@ -212,6 +218,11 @@ func (f *AuditFormatter) FormatResponse(ctx context.Context, w io.Writer, config
 			NumUses:                   resp.Auth.NumUses,
 			EntityID:                  resp.Auth.EntityID,
 			TokenType:                 resp.Auth.TokenType.String(),
+			TokenTTL:                  int64(resp.Auth.TTL.Seconds()),
+		}
+		var zt time.Time
+		if resp.Auth.IssueTime != zt {
+			respAuth.TokenIssueTime = resp.Auth.IssueTime.Format(time.RFC3339)
```
mgritter, line 225: I'm not sure why this is useful to include? Does it differ significantly from the time on the response itself?

mgritter: Hm, could be useful for renew calls, I guess?

sgmiller: On a login I guess it would be a similar time.  I was just trying to add it consistently.

sgmiller:
> Curious why you left out the other LeaseOptions fields?

The remaining fields are MaxTTL (which is sort of irrelevant to an active token which has an actual ttl), Renewable, which proabably could be useful, and Increment, which doesn't apply to a token but only a renew request.

sgmiller:
Actually, I take that back, Renewable isn't available from the TokenEntry data we get during checkToken, so we can't populate it.

audit/format.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"github.com/hashicorp/vault/sdk/helper/salt"
```
ncabatoff, line 7: Can you revert this rearrangement of imports please?

audit/format.go
```diff
@@ -127,6 +128,11 @@ func (f *AuditFormatter) FormatRequest(ctx context.Context, w io.Writer, config
 		},
 	}
 
+	var zt time.Time
+	if auth.IssueTime != zt {
```
ncabatoff, line 132: I used to do this all the time, but I think you can now use the `.IsZero` method instead.

audit/format.go
```diff
@@ -127,6 +128,11 @@ func (f *AuditFormatter) FormatRequest(ctx context.Context, w io.Writer, config
...
+		reqEntry.Auth.TokenIssueTime = auth.IssueTime.Format(time.RFC3339)
```
ncabatoff, line 133: Is there precedent for stringifying times before auditing?  Does it work if you don't?  I'd prefer to defer formatting as late as possible.  This also means if we decide that times don't need to be redacted, it's easy to not to do so.

ncabatoff: Sorry, I should clarify: I know we do it for some "envelope" fields like AuditResponseEntry.Time, I mean for the actual Auth/Request/Response fields that go through HashAuth/HashRequest/HashResponse.  But it looks like you're doing this after redaction, so it's moot.  Which is a bit questionable, mind you.  It's probably ok, and seems to achieve the desired effect. but it's a risky practice to get into.

ncabatoff:
Can we update the formatter tests please, e.g. TestFormatJSONx_formatRequest?

audit/format.go
sgmiller: ah, much better.  Was not liking that construct.

audit/format.go
sgmiller: My concern was with parseability by Splunk and other systems.  I assume unix time would also work but it makes the audit log not human readable, a problem I didn't want to make worse.  Re doing after redaction, that's probably just an oversight on my part, but these fields if redacted are pretty much useless.  In particular, due to the small domain size of ttl, its pretty easy to reverse the hmac via a dictionary if you have the ability to create tokens anyway.

sgmiller:
Will update the tests...

audit/format.go
ncabatoff: Oh, I agree that we probably don't want these fields redacted.  I just want redaction (and the decision as to whether to redact) to be in the same place as much as possible.  I also want times rendered with RFC3339, as you were doing.

sgmiller:
Is there generalized time formatting for values in the audit log that I 
should have been using?  I'll definitely see if I can move those value 
assignments to pre-redact.

PR #9093, opened 2020-05-27
docs: document raft and mlock interaction

calvn:


website/pages/docs/configuration/index.mdx
```diff
@@ -81,14 +81,22 @@ to specify where the configuration is.
   process's entire memory address space from spilling to disk and is therefore
   automatically disabled on unsupported platforms.
 
+  Disabling `mlock` is strongly recommended if using [integrated
+  storage](https://www.vaultproject.io/docs/internals/integrated-storage) due to
+  the fact that `mlock` does not interact well with memory mapped files such as
+  those created by BoltDB, which is used by Raft to track state. Additionally,
+  swap should be disabled to prevent these files from being dumped into disk.
```
ncabatoff, line 88: I thought the danger lay with Vault's unencrypted memory being swapped to disk (i.e. the original motivation for our using mlock), and not with the BoltDB mmaps which are already encrypted?

calvn: You're right, the reason is less so from a security perspective since the data is already in an  encrypted state. I'm thinking it might be due to the fact that BoltDB simply doesn't play nicely with its files being dumped to disk, but I'd defer to @briankassouf for the exact reasoning.

website/pages/docs/configuration/index.mdx
```diff
@@ -81,14 +81,22 @@ to specify where the configuration is.
   process's entire memory address space from spilling to disk and is therefore
   automatically disabled on unsupported platforms.
 
+  Disabling `mlock` is strongly recommended if using [integrated
+  storage](https://www.vaultproject.io/docs/internals/integrated-storage) due to
+  the fact that `mlock` does not interact well with memory mapped files such as
```
briankassouf, line 86: Let's go into a little more detail here: when using `mlock` memory-mapped files are loaded into resident memory. This causes Vault to store the full dataset in memory and is particularly problematic when using larger-than-RAM datasets.

website/pages/docs/configuration/index.mdx
briankassouf: Yeah it's Vault's memory that we are concerned about. The BoltDB file will be on disk already. Maybe something along the lines of "If you disable `mlock` it's highly recommended that you also disable swap on the server to prevent Vault's memory from being swapped to disk."

calvn: I guess the question still remains on the why for “prevent Vault’s memory from being swapped to disk”? Does it come down to a disk utilization concern, or is it something else?

website/pages/docs/configuration/index.mdx
calvn: Should be addressed by 8fa6fabb240ccf0ce4d740564ad63b4f5ad7972f

website/pages/docs/configuration/index.mdx
```diff
@@ -81,14 +81,27 @@ to specify where the configuration is.
   process's entire memory address space from spilling to disk and is therefore
   automatically disabled on unsupported platforms.
 
+  Disabling `mlock` is strongly recommended if using [integrated
+  storage](https://www.vaultproject.io/docs/internals/integrated-storage) due to
```
briankassouf, line 85:
```suggestion
  storage](/docs/internals/integrated-storage) due to
```

PR #9095, opened 2020-05-28
Fix FreeBSD build by pinning gopsutil to 48177ef5f880

samsalisbury:
- This is a workaround for https://github.com/shirou/gopsutil/issues/853


samsalisbury:
That specific revision was chosen only because Consul are using it, and I know it does not cause the build problem I was trying to solve. Closing this now, as @ncabatoff has a better solution, opening an upstream PR to fix it, and temporarily depending on our own fork until that's merged.

PR #9097, opened 2020-05-28
Extend agent template tests to also validate that updated templates get re-rendered

ncabatoff:


command/agent_test.go
```diff
@@ -770,19 +771,22 @@ func TestAgent_Template_Basic(t *testing.T) {
 
 	for tcname, tc := range testCases {
 		t.Run(tcname, func(t *testing.T) {
-			// make some template files
-			var templatePaths []string
-			for i := 0; i < tc.templateCount; i++ {
-				path := makeTempFile(t, fmt.Sprintf("render_%d", i), templateContents(i))
```
catsby, line 776: Any particular reason to not use `makeTempFile` here?

ncabatoff: When I was debugging it was easier to have the inputs and outputs in the same directory and using the same basename.

PR #9113, opened 2020-06-01
UI/link to changelog version in ui footer

noelledaley:
# Link to Changelog version in UI Footer

image

This PR adds a link to the changelog at the version of Vault currently being used. This allows users to quickly view relevant changes, particularly when we have added or changed behavior in the UI.

## Notes
The helper used to generate the URL assumes that changelog items will match the following format:
```
## v1.5.0
### Month, DD, YYYY

## v1.4.5
### Month, DD, YYY
```

Since this new header format will only be used for Vault versions 1.4.3 and beyond, all previous versions just link back to the base changelog link.

### DIFF
--- a/ui/app/templates/vault.hbs
+++ b/ui/app/templates/vault.hbs
@@ -10,7 +10,7 @@
       &copy; {{date-format (now) "YYYY"}} HashiCorp
     </span>
     <span>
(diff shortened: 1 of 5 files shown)


noelledaley:
Closed and reopened here so these changes are separated from the replication project. https://github.com/hashicorp/vault/pull/9216

PR #9119, opened 2020-06-02
Simple typos

JeffHemmen:
Fixes two very simple typos.

Mostly I want to ease myself into the collaboration workflow.


JeffHemmen:
Clarification: typos in API Docs.
See current live: https://www.vaultproject.io/api-docs/system/rekey

PR #9125, opened 2020-06-02
Add JSDocs to components

Monkeychip:
This PR adds JSDocs to the components that still need it.    

I also went ahead and removed the TODO that was no longer relevant. 

As a side note, I find it helpful to see a list of all the components we've added in this replication project.  Here they are for reference.  After this PR all of these will have JSDocumentation.

1. replication-dashboard
2. replication-documentation-link
3. replication-header
4. replication-page
5. replication-primary-card
6. replication-secondary-card
7. replication-summary-card
8. replication-table-rows
9. known-secondaries-card
10. known-secondaries-table

### DIFF
--- a/ui/lib/core/addon/components/replication-table-rows.js
+++ b/ui/lib/core/addon/components/replication-table-rows.js
@@ -2,6 +2,21 @@ import Component from '@ember/component';
 import { computed } from '@ember/object';
 import layout from '../templates/components/replication-table-rows';
 
+/**
+ * @module ReplicationTableRows
+ * The `ReplicationTableRows` component is table component.  It displays cluster mode details specific to the cluster of the Dashboard it is used on.
+ *
+ * @example
+ * ```js
+ * <ReplicationTableRows
+    @replicationDetails={{replicationDetails}}
+    @clusterMode="primary"
+    />
+ * ```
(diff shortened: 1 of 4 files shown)


ui/lib/core/addon/components/replication-table-rows.js
```diff
@@ -2,6 +2,21 @@ import Component from '@ember/component';
...
+ *
+ * @example
+ * ```js
+ * <ReplicationTableRows
+    @replicationDetails={{replicationDetails}}
+    @clusterMode="primary"
+    />
+ * ```
+ * @param {Object} replicationDetails=null - An Ember data object pulled from the Ember Model. It contains details specific to the whether the replication is dr or performance.
+ * @param {String} clusterMode=null - The cluster mode passed through to a table component.
```
noelledaley, line 17: it would be helpful to specify here whether this is 'dr' or 'performance' or 'primary' or 'secondary', since i know i always get them confused 😅

Monkeychip: Good call! Added.

Monkeychip:
> one tiny change, then good to go! thank you for doing this!

Fixed!  I'll merge once the checks go through.

PR #9128, opened 2020-06-02
set path properly in NewRaftBackend()

mjarmy:
Set the path properly in `NewRaftBackend()`, so that `EnvVaultRaftPath` is taken into account.


ncabatoff:
Fixes #9484.

PR #9120, opened 2020-06-02
Replication Sidebranch: Management Actions

chelshaw:
This is the management sidebranch to be merged into the replication sidebranch when the work is finished. The management pages exist for all types of replication (DR primary, DR secondary, Perf primary, Perf secondary). 

Tasks to complete before merging:

- [x]  Disable action
- [ ] Promote action
- [ ] Update action
- [ ] Recover action
- [ ] Reindex action
- [ ] Fix tests
- [ ] (Stretch) Generate token action (DR sec only)

PR #9136, opened 2020-06-03
UI: Fix replication management tests

noelledaley:
# Update Replication Management Tests
This PR fixes several test selectors and updates our replications tests so they're compatible with the new modal workflow. I have commented out several test case scenarios that will be re-enabled as we add new actions to the https://github.com/hashicorp/vault/pull/9120 branch.

### DIFF
--- a/ui/lib/core/addon/components/replication-header.js
+++ b/ui/lib/core/addon/components/replication-header.js
@@ -22,6 +22,7 @@ import layout from '../templates/components/replication-header';
 
 export default Component.extend({
   layout,
+  data: null,
   classNames: ['replication-header'],
   isSecondary: null,
   secondaryId: null,
--- a/ui/lib/core/addon/templates/components/replication-action-reindex.hbs
+++ b/ui/lib/core/addon/templates/components/replication-action-reindex.hbs
@@ -7,7 +7,6 @@
       Reindex the local data storage
     </p>
   </div>
-
   <div class="action-block-action">
     <button
       class="button is-secondary"
@@ -18,7 +17,6 @@
     </button>
   </div>
 </div>
-
 <Modal
   @title="Begin reindex?"
   @onClose={{action (mut isModalActive) false}}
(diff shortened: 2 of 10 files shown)


ui/lib/core/addon/components/replication-header.js
```diff
@@ -22,6 +22,7 @@ import layout from '../templates/components/replication-header';
 
 export default Component.extend({
   layout,
+  data: null,
```
chelshaw, line 25: Is this change part of the test fixes, or why is it necessary?

ui/lib/core/addon/templates/components/replication-action-reindex.hbs
```diff
@@ -28,25 +26,28 @@
...
   <section class="modal-card-body">
     <p class="has-bottom-margin-m">
-      Reindexing can cause a very long delay on the number and size of objects in the data store. {{if model.replicationAttrs.isPrimary 'You should always re-index your secondary first.'}}
+      Reindexing can cause a very long delay depending on the number and size of objects in the data store.
+      {{if model.replicationAttrs.isPrimary 'You should always re-index your secondary first.'}}
     </p>
     <p>
-      Progress will be shown, and you will {{ if model.replicationAttrs.isPrimary 'not' }} be able to use Vault during this time.
+      Progress will be shown, and you will
+      {{if model.replicationAttrs.isPrimary 'not'}}
```
chelshaw, line 34: Let's make sure the formatting here didn't remove all the spaces around `not` so it doesn't look like `you willnotbe able...`

chelshaw: We can just add the spaces around "not" if it did break the spaces 👍

ui/tests/integration/components/replication-actions-test.js
```diff
@@ -39,22 +39,54 @@ module('Integration | Component | replication actions', function(hooks) {
...
+      ['disable', 'primary'],
+      false,
+    ],
+    [
+      'performance',
+      'secondary',
+      'disable',
+      'Disable Replication',
+      async function() {
+        fillIn('[data-test-confirmation-modal-input]', 'Disaster Recovery');
```
chelshaw, line 70: This one should be 'Performance'

ui/lib/core/addon/components/replication-header.js
noelledaley: ah yes, this was to ensure the property was documented & would show up in the Ember debugger. previously `data` was being used in the template but wasn't explicitly referenced in the component JS file, which meant that the component in the Ember debugger didn't have a `data` attribute. i needed access to it to debug the tests.

you can see that the attribute is now listed here:
image

so fun fact: defining a default value (even if it's `null`) in a component ensures it's listed in the Ember debugger too!

ui/lib/core/addon/templates/components/replication-action-reindex.hbs
noelledaley: you know, that's interesting -- i totally would've thought that this would smoosh the words together too, but it didn't 🤔 

image

ui/tests/integration/components/replication-actions-test.js
noelledaley: nice catch! fixed 🦅

PR #9137, opened 2020-06-03
updated 1.3.2 and 1.4.0 upgrade guides to note AWS STS region issue

dedamico:
1.4.1 resolves an issue with AWS STS region selection in the AWS Client that Vault creates.  This issue appears to have been introduced in 1.3.2 & 1.4.0.

(This came from an issue from a customer who upgraded to 1.4.0, and we confirmed 1.4.1 resolved their issue - I can provide details internally).


kalafut:
Thanks!

PR #9143, opened 2020-06-03
Integrate password policies into RabbitMQ secret engine

pcman312:
- Integrates password policies into the RabbitMQ secret engine
  - `password_policy` field is optional. If not specified, it defaults to using the `base62` package instead of a uuid as a password.
- Updates docs for RabbitMQ
  - Tabified the API docs for RabbitMQ to show both cURL & Vault CLI commands


builtin/logical/rabbitmq/path_config_connection.go
```diff
@@ -91,6 +100,33 @@ func (b *backend) pathConnectionUpdate(ctx context.Context, req *logical.Request
...
+	}
+
+	var connConfig connectionConfig
+	if err := entry.DecodeJSON(&connConfig); err != nil {
+		return connectionConfig{}, err
+	}
+	return connConfig, nil
+}
+
+func writeConfig(ctx context.Context, storage logical.Storage, config connectionConfig) (err error) {
```
tvoran, line 119: nit: It doesn't seem like the named return `err` is being used in this function since a new `err` is declared every time it's used.

```suggestion
func writeConfig(ctx context.Context, storage logical.Storage, config connectionConfig) error {
```

builtin/logical/rabbitmq/path_config_connection.go
```diff
@@ -91,6 +100,33 @@ func (b *backend) pathConnectionUpdate(ctx context.Context, req *logical.Request
 	return nil, nil
 }
 
+func readConfig(ctx context.Context, storage logical.Storage) (config connectionConfig, err error) {
```
tvoran, line 103: nit: It doesn't seem like the named parameters are being used in this function, maybe they're not necessary?

```suggestion
func readConfig(ctx context.Context, storage logical.Storage) (connectionConfig, error) {
```

PR #9149, opened 2020-06-04
Ui/replication mgmt/update primary action

chelshaw:
The update-primary action is now updated to use the modal flow! Before/After below

Before, DR:
Screen Shot 2020-06-04 at 10 51 02 AM

After, DR:
dr-update-primary

Before, Perf:
Screen Shot 2020-06-04 at 11 06 46 AM

After, Perf: 
Screen Shot 2020-06-04 at 11 27 24 AM

This PR also includes a change to the modal styling, which keeps the modal from being tucked under the header when the window is shorter than the modal content:
Screen Shot 2020-06-05 at 12 26 57 PM

### DIFF
--- a/ui/app/styles/components/modal.scss
+++ b/ui/app/styles/components/modal.scss
@@ -5,6 +5,8 @@
 .modal-card {
   box-shadow: $box-shadow-highest;
   border: 1px solid $grey-light;
+  max-height: calc(100vh - 70px);
+  margin-top: 60px;
 
   &-head {
     border-radius: 0;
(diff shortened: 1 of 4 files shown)


noelledaley:
oh, one other question because i can't remember 😅 -- since this PR removes the Operation Token workflow from the UI, is that action coming next? for some reason i thought i remembered it being out of scope with this release. but if it _is_ out of scope for this release, how will users generate an operation token?

chelshaw:
> oh, one other question because i can't remember 😅 -- since this PR removes the Operation Token workflow from the UI, is that action coming next? for some reason i thought i remembered it being out of scope with this release. but if it _is_ out of scope for this release, how will users generate an operation token?

Very good question @noelledaley! That was removed with the first code merged into the replication-mgmt-sidebranch, but in any case the set of modals for generating an operation token is complex and I believe we did list it as a stretch goal. That being said, I can prioritize that for next to make sure it gets in before launch, but worst case there is a CLI command we could list (in the copyable format) for them to run in terminal

PR #9150, opened 2020-06-04
Ui/secondary token flow dr

Monkeychip:
This PR implements the designs for generating a secondary token.  We are changing the flow so that it opens a modal, which when closed, will automatically copy the token to your clipboard.  The point of this is to avoid the previous "back" button flow and to make the whole transition a little smoother.  

I was originally going to make this it's own branch into master, but because we're on schedule with that release and it will be nice to have when doing a final QA of the replication project, I am going to include it in the sidebranch.

We also now show the TTL and the expiration date which is generated from decoding the jwt token.

**Previous Flow**

**New Flow**
_In the gif I paste the copied token in the URL so you can see that it's been copied to my clipboard. 
 It is copied to the clipboard when I either click the Copy & Close button or click off the modal._


ui/lib/core/addon/components/modal.js
```diff
@@ -26,7 +26,7 @@ export default Component.extend({
   glyph: computed('type', function() {
     const modalType = this.get('type');
     if (!modalType) {
-      return {};
+      return;
```
Monkeychip, line 29: this was to fix an empty icon element for the models (like this one and the transit-key-actions modal) that did not have an icon.

ui/lib/core/package.json
```diff
@@ -5,6 +5,7 @@
   ],
   "dependencies": {
     "autosize": "*",
+    "date-fns": "*",
```
Monkeychip, line 8: for testing, I needed to add the date-fns package as a core dependency

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
```diff
@@ -6,72 +6,74 @@
     <p>Generate a token to enable {{replicationMode}} Replication or change primaries on secondary cluster.</p>
   </div>
   {{message-error errors=errors}}
-  {{#if token}}
```
Monkeychip, line 9: removing the previous token flow section

ui/tests/helpers/date-format.js
```diff
@@ -0,0 +1,8 @@
+import { helper } from '@ember/component/helper';
```
Monkeychip, line 1: Add date formate to test helpers to use in enterprise-replication-test

ui/tests/integration/helpers/date-format-test.js
```diff
@@ -1,27 +1,37 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { dateFormat } from '../../../helpers/date-format';
```
Monkeychip, line 3: because this helper was moved to the core, it can no longer be accessed and rendered like it previously was.  Therefore, I pretty much rewrote the tests for this helper.  I attempted to keep in spirit with the previous test, while working with the new render flow I had to implement.

ui/tests/acceptance/enterprise-replication-test.js
```diff
@@ -88,6 +89,22 @@ module('Acceptance | Enterprise | replication', function(hooks) {
     await click('[data-test-secondary-add]');
 
     await pollCluster(this.owner);
+    // checks on secondary token modal
+    assert.dom('#modal-wormhole').exists();
+
+    const date = new Date();
```
Monkeychip, line 95: The point of this test is to make sure:
1: the modal is rendering the date in the correct format
2: that the time of adding the default 30 min matches to the generated token.  Adding 30 min to the the current timestamp vs decoding the token could result in a 1-2 second difference because "computer things".  Thus, I check the strings are the same up until the seconds.  

I think these checks are helpful and cover a lot of functionality, but if you can think of ways to make them easier to read or clean up, I'm all ears 👂.

ui/lib/replication/addon/controllers/application.js
```diff
@@ -105,7 +116,26 @@ export default Controller.extend(copy(DEFAULTS, true), {
...
+      }
+      this.toggleProperty('isModalActive');
+      this.transitionToRoute('mode.secondaries');
+    },
+    toggleModal(successMessage) {
+      if (!!successMessage && typeof successMessage === 'string') {
+        this.get('flashMessages').success(successMessage);
+      }
+      // use copy browser extension to copy token if you close the modal by clicking outside of it.
+      const htmlSelectedToken = document.querySelector('.textarea');
```
chelshaw, line 132: You may consider using the `<CopyButton>` component on the modal and then hooking into closing the modal close on the success action, the querySelector seems a little tenuous if there's ever more than one element with `.textArea` (see app/templates/components/secret-edit.hbs for an example)

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
```diff
@@ -6,72 +6,74 @@
...
+      {{/link-to}}
     </div>
-  {{/if}}
+  </div>
 </form>
+
+{{#if isModalActive}}
+  <Modal @title="Copy your token" @onClose={{action "toggleModal" "Token copied!"}} @isActive={{isModalActive}}>
+    <section class="modal-card-body">
+      <p>This token can be used to enable {{replicationMode}} replication or change primaries on the secondary cluster.</p>
```
noelledaley, line 54: looks like this needs to be updated to `Disaster Recovery Replication` instead of`dr replication`:

image

ui/lib/replication/addon/controllers/application.js
noelledaley: yes -- to echo chelsea's advice i'd recommend using the `HoverCopyButton` component & using the same codeblock styles that we use when generating an operation token and elsewhere in the UI. the codeblock (vs an input) makes it clearer that this value is meant to be copied and not edited. here's an example of how to do that:
https://github.com/hashicorp/vault/blob/061bc8ed2bbed49eb4f54c3a04647fbbefb13e48/ui/app/templates/components/shamir-flow.hbs#L23-L34

image

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
```diff
@@ -6,72 +6,74 @@
...
+        <h2 class="title is-6">Activation token</h2>
+        <div class="copy-text level">
+          <div class="is-fullwidth">
+            <textarea readonly value={{token}} class="textarea level-left"/>
+          </div>
+        </div>
+        <div class="has-top-margin-xl has-bottom-margin-s">
+          {{info-table-row
+            label="TTL"
+            value=ttl}}
```
noelledaley, line 65: is this value correct? i noticed when i entered a TTL of 60 seconds, the result displayed 60 minutes. not sure if this is a problem with the TTL picker or what but it's worth investigating:

image

_image_

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
```diff
@@ -6,72 +6,74 @@
...
-        This will be used to identify secondary cluster once a connection has been established with the primary.
-      </p>
-    </div>
-    <div class="field">
-      {{ttl-picker onChange=(action (mut ttl)) class="is-marginless"}}
-      <p class="help has-text-grey">
-        This is the Time To Live for the generated secondary token. After this period, the generated token will no longer be valid.
-      </p>
+    <p class="help has-text-grey">
+      This will be used to identify secondary cluster once a connection has been established with the primary.
```
noelledaley, line 17: nit: this is missing an "a" -- "will be used to identify a secondary cluster"

ui/tests/acceptance/enterprise-replication-test.js
noelledaley: one thing that'd help with making the tests easier to read would be moving this to its own test. in an ideal scenario testing that the modal shows the right stuff would be inside a component test, but in lieu of that i think this should be kept separate from the rest of this test since this test is already super long (and really ought to be broken up 😅). 

whaddya think about moving this to a separate test for just adding a secondary using this new flow? 

and there's a few small naming nits that might help too. up to you!

ui/tests/acceptance/enterprise-replication-test.js
```diff
@@ -88,6 +89,22 @@ module('Acceptance | Enterprise | replication', function(hooks) {
...
+    date.setMinutes(date.getMinutes() + 30); // add default 30 min TTL to current date to return expires
+    const dateFromHelper = dateFormat([date, 'MMM DD, YYYY hh:mm:ss A']);
```
noelledaley, line 97: this could be renamed `formattedDate`. `dateFromHelper` makes me think that it's returning the actual helper, vs. a string

ui/tests/acceptance/enterprise-replication-test.js
```diff
@@ -88,6 +89,22 @@ module('Acceptance | Enterprise | replication', function(hooks) {
...
+    // because timestamp might be off by a 1 sec due to different in exp from token vs. adding 30 min, checking the string before the seconds.
+    const shortendedDateFromHelper = dateFromHelper.slice(0, 18);
```
noelledaley, line 99: this could be `shortenedDate`

ui/tests/integration/helpers/date-format-test.js
```diff
@@ -1,27 +1,37 @@
...
+      .dom('[data-test-date-format]')
+      .includesText(today.getFullYear(), 'it renders the date in the year format');
   });
 
-  test('it supports date timestamps', function(assert) {
-    let today = new Date().getTime();
-    let result = dateFormat([today, 'YYYY']);
-    assert.ok(Number(result) >= 2017);
+  test('it formats the date as specified', async function(assert) {
+    let today = new Date();
```
noelledaley, line 20: i think you'll want to keep the same value we were using before in this test -- `Date().getTime()`, since that is testing that the helper can handle various types of inputs. it's not testing the various types of outputs, if that makes sense.

ui/tests/acceptance/enterprise-replication-test.js
Monkeychip: I think that's a great idea.  I'll move it and work on some of the naming.

ui/lib/replication/addon/controllers/application.js
Monkeychip: I chatted this through with Chelsea.  I originally tried this solution, however, this modal is a bit unique from the others.  After chatting with design, we decided that we wanted the modal to copy the token if you click *off* the modal (not just on a button click).  The reason is that the moment you click "Generate Token" (so before the modal even opens) a token has been generated.  If you click off the modal and we don't copy the token or redirect them to the secondaries page, then if they try and click "Generate Token" again they'll get an error that the token already exists.  Thus, we really want to force them, no matter what they do, to copy the token and redirect to the secondaries page.  Additionally, because the `@onClose` action of the modal needs to both copy and send the success message, I moved this to an action inside the controller.  And the `@clipboardText` is only an hbs helper, and is not something I can use in the controller.  Hence the whole action -> select and copy the token + send a success message flow.  

Design originally had the copy button icon as you both mentioned, but again, because we needed them to copy the token when they closed the modal, we removed it as it was redundant.  Right now, they have two ways to copy the token, 1. click the copy & close button, or 2. click off the modal (e.g. close it).  We didn't need another copy button that ultimately no matter what their next step was, would copy the token again.  

Let me know if that makes sense.

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
Monkeychip: We have this issue elsewhere in the app.  Therefore I went ahead and added a cluster property called `replicationModeForDisplay` and replaced elsewhere in the app where we hardcode or conditionally render the display version of the mode.  I made this one commit as Chelsea will likely need to implement these changes into her manage actions sidebranch.

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
Monkeychip: fixed with setting initalValue of ttl-picker.  This is the quick fix without replacing the old ttl-picker with the new one. That can be something we tackle in another, post project, ticket.

ui/tests/integration/helpers/date-format-test.js
Monkeychip: Makes sense.  Amended.

ui/tests/acceptance/enterprise-replication-test.js
Monkeychip: I added the chain the assert.equal as I was having issues chaining on dateFormat and document.query...

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
```diff
@@ -18,7 +18,7 @@
     </p>
   </div>
   <div class="field">
-    {{ttl-picker onChange=(action (mut ttl)) class="is-marginless"}}
+    {{ttl-picker onChange=(action (mut ttl)) initialValue="30m" class="is-marginless"}}
```
noelledaley, line 21: this is a good interim fix, but it is a bit jarring to see `1800 seconds` as the default on the page (even though it does technically amount to 30 minutes), whereas everywhere else in the UI the TTL defaults to 30s. @chelshaw do you have any ideas about why this bug might be happening?

in any case, would be good to drop a comment reminder here + in JIRA to fix this bug & add it to our backlog 🐸

Monkeychip: Agreed - I'll go ahead and make a JIRA ticket for it and add it to the backlog. I'll also add a comment in the code.

PR #9151, opened 2020-06-04
Document and give an example of the input size limits when using the FF3-1 transform.

sgmiller:
A calculator might have been nice. :)


website/pages/docs/secrets/transform/index.mdx
```diff
@@ -221,6 +250,8 @@ The following builtin alphabets are available for use in the secret engine:
 - builtin/alphanumericupper
 - builtin/alphanumeric
 
+Custom alphabets may contain between 2 and 65536 characters.
```
calvn, line 253:
```suggestion
Custom alphabets must contain between 2 and 65536 unique characters.
```

calvn:
Added the `docs-cherrypick` label to this so that it can be automatically cherry-picked.

PR #9152, opened 2020-06-05
Add known primaries info table

noelledaley:
# Show known_primary_cluster_addrs
This PR replaces the last wal and primary cluster address on the replication secondary dashboards with the known_primary_cluster_addrs. 

It also adds a new component, the `InfoTable`, which uses our existing `InfoTableRow` component and `vlt-table` style-only component to render a list-like table of `InfoTableRow`s. We can reuse this elsewhere in the Vault UI, hopefully refactoring it to accept multiple columns so it can replace the `KnownSecondariesTable` component, for instance.

*Note:* I haven't yet figured out what size the max-height should be of the Primary Clusters card, or where to set it. Ideas welcome!

#### Primary Cluster card displays the last_remote_wal and primary_cluster_add
image

#### The Primary Cluster Card on a secondary now has a scrollable list of known_primary_addrs
image

#### The Known Secondaries Card on a primary is now scrollable
image

#### An info table row component with a label, no change
image

### DIFF
--- a/ui/app/styles/components/selectable-card.scss
+++ b/ui/app/styles/components/selectable-card.scss
@@ -67,6 +67,11 @@
     font-weight: 500;
     line-height: 1.33;
   }
+
+  .vlt-table {
+    max-height: 240px;
+    overflow: scroll;
+  }
 }
 
 .selectable-card.is-rounded {
--- a/ui/app/styles/components/selectable-card.scss
+++ b/ui/app/styles/components/selectable-card.scss
@@ -70,7 +70,7 @@
 
   .vlt-table {
     max-height: 240px;
-    overflow: scroll;
+    overflow-y: scroll;
   }
 }
 
--- a/ui/app/styles/components/vlt-table.scss
+++ b/ui/app/styles/components/vlt-table.scss
@@ -4,6 +4,15 @@
     height: 0;
   }
 
+  &.sticky-header {
+    thead th {
+      position: sticky;
+      background: #fff;
+      box-shadow: 0 1px 0px 0px rgba($grey-dark, 0.3);
+      top: 0;
+    }
+  }
+
   th,
   td {
     padding: $spacing-s;
--- a/ui/lib/core/addon/templates/components/info-table-row.hbs
+++ b/ui/lib/core/addon/templates/components/info-table-row.hbs
@@ -1,12 +1,14 @@
 {{#if (or alwaysRender value)}}
-  <div class="column is-one-quarter">
-    <span class="is-label has-text-grey-dark" data-test-row-label="{{label}}">{{label}}</span>
-    {{#if helperText}}
-      <div>
-        <span class="is-label helper-text has-text-grey">{{helperText}}</span>
-      </div>
-    {{/if}}
-  </div>
+  {{#if label}}
+    <div class="column is-one-quarter">
+      <span class="is-label has-text-grey-dark" data-test-row-label="{{label}}">{{label}}</span>
+      {{#if helperText}}
+        <div>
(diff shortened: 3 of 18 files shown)


ui/lib/core/addon/templates/components/info-table-row.hbs
```diff
@@ -1,12 +1,14 @@
 {{#if (or alwaysRender value)}}
-  <div class="column is-one-quarter">
-    <span class="is-label has-text-grey-dark" data-test-row-label="{{label}}">{{label}}</span>
-    {{#if helperText}}
-      <div>
-        <span class="is-label helper-text has-text-grey">{{helperText}}</span>
-      </div>
-    {{/if}}
-  </div>
+  {{#if label}}
```
noelledaley, line 2: this change allows us to display info-table-rows with no lefthand column. this was **necessary**

image

ui/lib/core/addon/templates/components/replication-secondary-card.hbs
```diff
@@ -1,39 +1,59 @@
-<div class="selectable-card is-rounded card-container {{if hasErrorClass "has-border-danger"}}" data-test-replication-secondary-card>
-  {{!-- Check if State or WAL Card --}}
-  {{#if (eq title "States")}}
-    <h3 class="title is-5 grid-item-top-left card-title">{{title}}</h3>
+<div
+  class="selectable-card is-rounded card-container {{if hasErrorClass 'has-border-danger'}}"
```
noelledaley, line 2: all of these are whitespace changes, sorrrrry :(

ui/lib/core/addon/templates/components/replication-secondary-card.hbs
```diff
@@ -43,34 +63,38 @@
...
-            rel="noreferrer noopener" 
-            data-test-primaryClusterAddr
-          >
-            {{primaryClusterAddr}}
-          </a>
-        </span>
-      {{else}}
-        <p class="has-text-grey is-size-8">unknown</p>
-      {{/if}}
+  {{else}}
```
noelledaley, line 80: down here and below 👇  are where the actual changes occur

ui/app/styles/components/selectable-card.scss
```diff
@@ -67,6 +67,11 @@
     font-weight: 500;
     line-height: 1.33;
   }
+
+  .vlt-table {
+    max-height: 240px;
+    overflow: scroll;
```
Monkeychip, line 73: A couple thoughts.  One to prevent the scroll bar from showing on the x axis, you might want to add either overflow-y: scroll instead of overflow scroll, or add overflow-x: hidden.

I worked on this one a little and got it so the header was sticky and scrolled.  It makes it a little harder, but I'd be happy to pair quickly through the solution if you think it's worth it.  The header being : 
image

ui/app/styles/components/selectable-card.scss
```diff
@@ -67,6 +67,11 @@
     font-weight: 500;
     line-height: 1.33;
   }
+
+  .vlt-table {
+    max-height: 240px;
```
Monkeychip, line 72: I think we need a min-height here, as if there are no know addrs then it looks a bit funny.  Also, there doesn't appear to be the empty state showing anymore.  Here's a screenshot where I expanded the min-height so you could see it.
image

ui/app/styles/components/vlt-table.scss
```diff
@@ -4,6 +4,15 @@
     height: 0;
   }
 
+  &.sticky-header {
```
Monkeychip, line 7: nice!

PR #9155, opened 2020-06-05
docs: adding a storage concepts page

mladlow:
This new concepts page very briefly covers some considerations for taking backups.

I have a google doc containing the same content. It's had some review. I'm happy to have a split of conversation here and in the doc, if people would prefer to look at the doc.


website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,40 @@
...
+Due to the highly flexible nature of Vault's potential storage configurations, providing exacting guidance on backing up and restoring Vault is challenging.
+
+When backing up Vault, there are two pieces to consider:
+
+1. Vault's encrypted data in the storage backend
+2. Configuration files and management scripts for running the Vault server
+
+### Vault's Persisted Data
+
+In an ideal world, backups and restores would be performed while Vault is offline. If offline backups are impossible, we have recommended using a storage backend that supports atomic snapshots (such as [Consul](https://www.consul.io/docs/commands/snapshot.html) or [Integrated Storage](https://www.vaultproject.io/docs/commands/operator/raft#snapshot)).
```
ncabatoff, line 30: "have recommended" sounds weird here, why not just "we recommend"?

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+**If your storage backend does not support atomic snapshots, we recommend only taking offline backups.**
+
+We recommend taking backups **before**, but not during, write operations to the `/sys` API (excluding the `/sys/leases`, `/sys/namespaces`, `/sys/tools`, `/sys/wrapping`, `/sys/policies`, and `/sys/pprof` endpoints). Some examples of workflows that write to the `/sys` API are upgrades, server moves, and rekeys. In the future, this guidance may change for the Integrated Storage backend.
+
+To perform a backup or restore of Vault's encrypted data, refer to the snapshotting documentation for your storage backend.
```
ncabatoff, line 36: Can we add links for Raft and Consul snapshot APIs or commands?  This sentence implies that all storage backends support snapshotting, I'd rather we point to how to do it for consul/raft and leave out the word "snapshot" for other backends.

mladlow: Yup - I'll poke at this paragraph a little more.

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+### Configuration
```
ncabatoff, line 38: Another suggested thing to backup: any user-installed plugins.

Should we put in a warning here about backing up sensitive configuration?  You might have a vault token for transit autounseal in your config, or you might have a TLS private key in your config dir, etc.

mladlow: Someone would have to help me write that part, but I think it would be a great addition.

website/pages/docs/concepts/storage.mdx
mladlow: I'm attempting to further communicate that we recommend it with slightly less preference than taking a backup while Vault is offline. It is a bit awkward, I would prefer to keep it however.

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+## Supported Storage Backends
+
+For enterprise customers, HashiCorp provides official support for Consul and Vault's Integrated Storage as storage backends.
+
+Many other options for storage are available with community support - see our [Storage Configuration](/docs/configuration/storage) section for more information.
+
+## Backups & Restores
+
+Due to the highly flexible nature of Vault's potential storage configurations, providing exacting guidance on backing up and restoring Vault is challenging.
```
calvn, line 21:
```suggestion
Due to the highly flexible nature of Vault's potential storage configurations, providing exact guidance on backing up and restoring Vault is challenging.
```

website/pages/docs/concepts/storage.mdx
calvn: I think Nick's suggestion here is not to remove this, but to slightly re-word it.

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,40 @@
...
+1. Vault's encrypted data in the storage backend
+2. Configuration files and management scripts for running the Vault server
+
+### Vault's Persisted Data
+
+In an ideal world, backups and restores would be performed while Vault is offline. If offline backups are impossible, we have recommended using a storage backend that supports atomic snapshots (such as [Consul](https://www.consul.io/docs/commands/snapshot.html) or [Integrated Storage](https://www.vaultproject.io/docs/commands/operator/raft#snapshot)).
+
+**If your storage backend does not support atomic snapshots, we recommend only taking offline backups.**
+
+We recommend taking backups **before**, but not during, write operations to the `/sys` API (excluding the `/sys/leases`, `/sys/namespaces`, `/sys/tools`, `/sys/wrapping`, `/sys/policies`, and `/sys/pprof` endpoints). Some examples of workflows that write to the `/sys` API are upgrades, server moves, and rekeys. In the future, this guidance may change for the Integrated Storage backend.
```
calvn, line 34: This bit is somewhat hard to enforce on production environments. Spitballing here, but I wonder if we could/should introduce a "maintenance mode" that essentially puts a cluster to be in read-only mode.

website/pages/docs/concepts/storage.mdx
kalafut, line 30:
```suggestion
Backups and restores are ideally performed while Vault is offline. If offline backups are not feasible, we recommend using a storage backend that supports atomic snapshots (such as [Consul](https://www.consul.io/docs/commands/snapshot.html) or [Integrated Storage](https://www.vaultproject.io/docs/commands/operator/raft#snapshot)).
```

website/pages/docs/concepts/storage.mdx
tomasbackman: > Should we put in a warning here about backing up sensitive configuration? 

Yes I would say that would be best  =)

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,40 @@
...
+Many other options for storage are available with community support - see our [Storage Configuration](/docs/configuration/storage) section for more information.
+
+## Backups & Restores
+
+Due to the highly flexible nature of Vault's potential storage configurations, providing exacting guidance on backing up and restoring Vault is challenging.
+
+When backing up Vault, there are two pieces to consider:
+
+1. Vault's encrypted data in the storage backend
+2. Configuration files and management scripts for running the Vault server
```
vishalnayak, line 26: What scripts are we referring here? The customer written scripts?
If yes, they will likely not be in Vault's store; and do we really want to warn about that?. It might make readers wonder about a storage related artifact to back up which doesn't exist.

website/pages/docs/concepts/storage.mdx
mikegreen: This is tough for 24/7 production/enterprise usage, maybe the guidance is always run a backend that is atomic snapshottable (ie, Consul + IS). And, if not, we rec they're performed offline.

website/pages/docs/concepts/storage.mdx
mladlow: Yes - these are things that the customer might use to interact with Vault that are not stored within Vault itself. I do think it's valuable to mention both the desire to save things internal to Vault and things external to Vault as a reminder to operators for when they're coming up plans - like a DR plan, or what needs to be on a new node when it's created and added to a cluster.

So, I want to include both items, but if you have suggestions for how to rephrase the second one to better represent "external-to-Vault stuff that you as an operator might want to save", I'm happy to rephrase?

website/pages/docs/concepts/storage.mdx
mladlow: I know it's hard to enforce, but I think the guidance is still worthwhile and that any maintenance mode conversation would be separate. Let me know if you disagree on this though.

calvn: Yep, this is a good heads-up to keep anyways.

website/pages/docs/concepts/storage.mdx
mladlow: @ncabatoff would you take another look at this paragraph sometime? If you want to do significant workshopping, let's chat on Slack or Zoom though.

ncabatoff: LGTM

website/pages/docs/concepts/storage.mdx
devops-rob: I can see both points of view here.  Maybe some like this for 2

"Disk backup for the underlying server running Vault.  _Note: This would include Vault's configuration file and any custom scripts written to help with the management of Vault and/or the server itself_"

devops-rob (commented):
This looks good to me.  I think adding some clarity around the disk back ups for the underlying server is the only thing we need to address but all in all it's useful for practitioners.  Good work @mladlow

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,74 @@
...
+### Vault's Persisted Data
+
+Backups and restores are ideally performed while Vault is offline. If offline backups are not
+feasible, we recommend using a storage backend that supports atomic snapshots (such as
+[Consul](https://www.consul.io/docs/commands/snapshot.html) or [Integrated
+Storage](/docs/commands/operator/raft#snapshot)).
+
+**If your storage backend does not support atomic snapshots, we recommend only taking offline backups.**
+
+We recommend taking backups **before**, but not during, write operations to the `/sys` API
```
ncabatoff, line 52: Do we want to give any explanation for why we recommend this?  e.g. "Even with atomic snapshots, certain Vault operations may consist of multiple writes, and a snapshot taken in the middle of them may result in an inconsistent state and thus a less useful backup."

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,74 @@
...
+* Consul [snapshots](https://www.consul.io/docs/commands/snapshot.html)
+
+### Configuration
+
+In addition to backing up Vault's encrypted data via the storage backend, you may also wish to
+save the server configuration files, any scripts for managing the Vault service, and ensure you
+can reinstall any user-installed plugins. The location of these files will be specific to your
+installation of Vault.
+
+~> **NOTE**: Some of your configuration may be sensitive (a Vault token for Transit Autounseal
```
ncabatoff, line 72: Suggest we add something after this to the effect that snapshots are encrypted and thus do not share this concern.

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,74 @@
...
+feasible, we recommend using a storage backend that supports atomic snapshots (such as
+[Consul](https://www.consul.io/docs/commands/snapshot.html) or [Integrated
+Storage](/docs/commands/operator/raft#snapshot)).
+
+**If your storage backend does not support atomic snapshots, we recommend only taking offline backups.**
+
+We recommend taking backups **before**, but not during, write operations to the `/sys` API
+(excluding the `/sys/leases`, `/sys/namespaces`, `/sys/tools`, `/sys/wrapping`, `/sys/policies`, and
+`/sys/pprof` endpoints). Some examples of workflows that write to the `/sys` API are upgrades,
+server moves, and rekeys. In the future, this guidance may change for the Integrated Storage
```
calvn, line 55: Is this referring to replication state changes, or does it entail more than that?

ncabatoff: I read it as changes to raft peers.

website/pages/docs/concepts/storage.mdx
mladlow: I followed up with Nick and will follow up with Mark, but the short answer is no. Feel free to reach out to me for further discussion.

website/pages/docs/concepts/storage.mdx
mladlow: I believe I copied it from someone's suggestion (which may have been on the original google doc), and so am not certain what it means and will remove it.

website/pages/docs/concepts/storage.mdx
briankassouf: Do we also want to mention here, or somewhere below, that the backups will only be able to be decrypted by the seal that is in use at the time of backup?

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+1. Vault's encrypted data in the storage backend
+2. Configuration files and management scripts for running the Vault server
+
+There's also a big question - what is the error case you're trying to guard against by saving a backup?
+
+### The Big Question - Why Take Backups?
+
+It's important to consider the question of "why take a backup" while developing your ongoing
+backup strategy.
```
briankassouf, line 33: This section mostly talks about when not to use backups, should we add a bit more around when backups are useful? i.e. before an upgrade or major change to the cluster, to combat accidental data deletions, etc?

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+We do not recommend backups as protection against the failure of an individual machine or data
+center. Vault servers can run in clusters. Vault Enterprise supports replicated clusters for
```
briankassouf, line 36: For OSS I do think backups cover the loss of a data center

website/pages/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,74 @@
...
+2. Configuration files and management scripts for running the Vault server
+
+There's also a big question - what is the error case you're trying to guard against by saving a backup?
+
+### The Big Question - Why Take Backups?
+
+It's important to consider the question of "why take a backup" while developing your ongoing
+backup strategy.
+
+We do not recommend backups as protection against the failure of an individual machine or data
```
briankassouf, line 35: I think this is what you're getting at but could be be explicit and say that backups are not a replacement for running in HA?

website/pages/docs/concepts/storage.mdx
briankassouf: This section may fit better in the "Why take backups" section.

mladlow: I think understanding when to take a backup is more of a "how do I" and less of a "should I" kind of thing - could you elaborate on the connection to the "should I" section?

briankassouf: I more read it as you "should" take a backup **before** doing system changing operations. Which i think is more useful than saying don't take a snapshot **during** these operations.

website/pages/docs/concepts/storage.mdx
mladlow: @briankassouf I moved some things around - could you re-read the "Why take Backups" section and see what you think?

website/content/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,110 @@
...
+---
+
+# Storage
+
+As described on our [Architecture](/docs/internals/architecture) page, Vault's
+storage backend is untrusted storage used purely to keep encrypted information.
+
+## Supported Storage Backends
+
+For enterprise customers, HashiCorp provides official support for Consul and
```
vishalnayak, line 16: Do we want to remove "For enterprise customers" here?

website/content/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,110 @@
+---
+layout: docs
+page_title: Storage
+sidebar_title: Storage
+description: >-
+  Vault relies on external storage to save its information.
```
vishalnayak, line 6:
```suggestion
  Vault relies on external storage to save its durable information.
```

website/content/docs/concepts/storage.mdx
```diff
@@ -0,0 +1,110 @@
...
+There's also a big question - what is the error case you're trying to guard
+against by saving a backup?
+
+### The Big Question - Why Take Backups?
+
+It's important to consider the question of "why take a backup" while developing
+your ongoing backup and disaster recovery strategy.
+
+Taking a backup is recommended prior to upgrades, as downgrading Vault storage
+is not always possible. Generally, a backup is recommended any time a major
```
vishalnayak, line 42: We may want to reword this to keep people from thinking that it is sometimes possible to downgrade storage. Some ideas: "downgrading Vault storage is not recommended", or "not supported"?

vishalnayak (commented):
This is good information to go into the docs!

website/content/docs/concepts/storage.mdx
mladlow: Yeah, that's fair.

website/content/docs/concepts/storage.mdx
mladlow: I don't think so? I think you have to be paying us to have support?

vishalnayak: My thought was that if support encompasses fixing bugs reported in the open and reviewing community PRs for Consul and Raft storage, it is free for non-enterprise folks.

PR #8075, opened 2019-12-27
Add documentation about flag -target=recovery for auto-unseal mode

tomasbackman:
The exact wording or style does not matter for me. The important thing is that information about this extra flag is given. (As mentioned in https://github.com/hashicorp/vault/issues/6658)

Issue #6658, opened 2019-04-30
Rekey operation new shares and threshold always show 1, and new key not output


terrych0u:
**Describe the bug**
When I try to use rekey operator and give the flag `-key-shares=5 -key-threshold=3`, the screen show `New Shares` and `New Threshold` always show `1`, and after I give it unseal key, the screen didn't show up the new key

**To Reproduce**
1. vault operator rekey -init -key-shares=5 -key-threshold=3
my screenshot: https://d.pr/free/i/UvSdDe

2. vault operator rekey -nonce=<nonce number> and give it 3 unseal key
nothing shows up on the screen.
my screenshot: https://d.pr/free/i/Y1PDbC

**Expected behavior**
The `key-shares` and `key-threshold` will be the same as I input, and the screen will show new key, just like documents: https://www.vaultproject.io/guides/operations/rekeying-and-rotating

**Environment:**
* Vault Server Version (retrieve with `vault status`):
1.1.1, https://d.pr/free/i/KdmVef
* Vault CLI Version (retrieve with `vault version`):
1.1.1
* Server Operating System/Architecture:
single container on k8s, image is 'vault:v1.1.1'

Vault server configuration file(s):

```hcl
{
  "listener": {
    "tcp": {
      "address": "0.0.0.0:8200",
      "tls_disable": 1
    }
  },
  "seal": {
    "awskms": {
      "kms_key_id": "XXXXXXXXXXXXX",
      "region": "ap-southeast-1"
    }
  },
  "storage": {
    "s3": {
      "bucket": "my-bucket",
      "region": "ap-southeast-1"
    }
  },
  "ui": true
}
```
 Thanks for the help.

michelvocks:
Hi @terrych0u,

I can see from your configuration that you use auto-unseal which, when enabled, internally transforms your key-shares to recovery keys. When you add the "-target=recovery" (https://www.vaultproject.io/docs/commands/operator/rekey.html#target) parameter to your rekey operation then it should work.

However, I agree that our documentation does not reflect that very well and is also partially outdated. I will try to fix that. Thanks for reporting this issue!  

Cheers,
Michel

terrych0u:
Hi @michelvocks 

The flag `-target=recovery` did solve `key-shares` and `key-threshold` number issue, however, the new unseal key still doesn't show up on the screen: https://d.pr/free/i/DXLUMS

I there something I'm missing?

Thanks for the help.

michelvocks:
[...]

PR #9163, opened 2020-06-08
Add note about flag -target=recovery for auto-unseal mode

tomasbackman:
Same as earlier PR #8075 (But based on latest master)

(Comes from issue: https://github.com/hashicorp/vault/issues/6658)


ncabatoff:
Thank you!

PR #9166, opened 2020-06-08
Add core of plugin test framework Stepwise 

catsby:
This is the start of implementing a new plugin testing framework codenamed `stepwise`. The goal is to implement a testing framework comparable to (and to eventually replace) the existing ["logicaltest"][logicaltest] framework found in `helper/testhelpers/logical/testing.go`, however using an actual, external, running instance of Vault. The canonical and first coming example would be using Docker to run Vault and run plugin tests against it. 

This PR introduces the basic types and interface to establish this framework, and the `Run` method that executes the user defined steps, and some simple tests. ~This PR is currently in DRAFT form for early review and feedback.~ 

**Update 2020-06-12:** added some tests, removing "draft"

[logicaltest]: https://github.com/hashicorp/vault/blob/4e6c3f23691a060dc7a6132cc8d956b7cdd0fc57/helper/testhelpers/logical/testing.go#L30


sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// StepCheckFunc is the callback used for Check in Steps.
+type StepCheckFunc func(*api.Secret, error) error
+
+// Case is a collection of tests to run for a plugin
+type Case struct {
+	// Driver is used to setup the Vault instance and provide the client used to
+	// drive the tests
+	Driver StepDriver
```
m1kola, line 154: Few questions:

1. Is driver going to be only responsible for setting up a plugin against a pre-existing vault instance? Or the intention to make it responsible for deploing/tearing down a vault instance too?

2. In a case when we need to not only register a plugin with a vault instance, but also set up/tear down some other external resources. Where do we do that? In driver setup? I think it is going to be relevant for auth plugins.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,345 @@
...
+	// TODO use test logger if available?
+	logger := logging.NewVaultLogger(log.Trace)
+	if c.Driver == nil {
+		tt.Fatal("nil driver in acceptance test")
+	}
+
+	err := c.Driver.Setup()
+	if err != nil {
+		driverErr := fmt.Errorf("error setting up driver: %w", err)
+		if err := c.Driver.Teardown(); err != nil {
```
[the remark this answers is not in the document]
catsby: Great catch, thank 😄  I added a check in b50c3f3f2 to only invoke `Teardown()` here if `SkipTeardown` is `true`

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,345 @@
...
+				tt.Error(fmt.Errorf("error revoking lease: %w", err))
+				failedRevokes = append(failedRevokes, secret)
+				continue
+			}
+		}
+
+		// If we have any failed revokes, log it.
+		if len(failedRevokes) > 0 {
+			for _, s := range failedRevokes {
+				tt.Fatal(fmt.Sprintf(
```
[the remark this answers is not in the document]
catsby: Indeed! Thanks again for the catch. I've updated that in b50c3f3f2

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,345 @@
...
+	case PluginTypeSecrets:
+		return "secret"
+	default:
+		return "unsupported"
+	}
+}
+
+// DriverOptions are a collection of options each step driver should
+// support
+type DriverOptions struct {
```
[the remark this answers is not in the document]
catsby: This is good feedback, and in my mind I've imagined that we'd likely need to eventually support passing in some kind of custom configuration that's specific to the `driver` (now called `environment`). I suspect it would be some kind of `map[string]interface{}` that the environment will need to know how to parse. 

For now I'm going to leave this as is and think more on it. The use of the options is more apparent in the example tests which will be shared in a future PR when the first environment is added.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,345 @@
...
+
+	// Unauthenticated will make the request unauthenticated.
+	Unauthenticated bool
+}
+
+// StepCheckFunc is the callback used for Check in Steps.
+type StepCheckFunc func(*api.Secret, error) error
+
+// Case is a collection of tests to run for a plugin
+type Case struct {
```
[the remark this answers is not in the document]
catsby: Yeah that's poorly worded. I updated the wording in b50c3f3f2 to (hopefully) better reflect the purpose. If you have some spare time, please let me know what you think!

sdk/testing/stepwise/stepwise.go
catsby: Thanks for the feedback/questions! Replies below:

1) Right now the environments are responsible for setup and teardown of the Vault instance itself, as well as compiling the plugin under test and registering it with the Vault instance. An early, very rough first version of a Docker environment (driver) can be see on the (now poorly named) [stepwise-driver-legacy][branch], specifically [here][environment]. An example test can be found in the [transit secret engine][transit].

2) Indeed auth plugins will need this in the future. For now setting up external resources such as that are beyond the initial scope of work we've set out for this framework, so it will need to be deferred. 

I hope that answers your questions, albeit maybe not sufficiently for the second one. The primary goal of this work is to effectively overhaul the existing but not often used plugin test framework to enable better externally driven testing, using an actual instance of Vault. 

[branch]: https://github.com/hashicorp/vault/tree/stepwise-driver-legacy
[environment]: https://github.com/hashicorp/vault/blob/5b8f2f71dbd1fbb4291cb90ba4048b3f817a4f0b/sdk/testing/stepwise/environments/docker/environment.go
[transit]: https://github.com/hashicorp/vault/blob/5b8f2f71dbd1fbb4291cb90ba4048b3f817a4f0b/builtin/logical/transit/stepwise_test.go

catsby:
Hey @m1kola thanks for the review! I'll reply to your items in-line below.

> Why do we want to have own testing framework with own types (`Case`, `Steps`, etc)?

Vault already has an existing testing framework that's the predecessor to this work in `helper/testhelpers/testing.go`:

- https://github.com/hashicorp/vault/blob/d11f15704df252bb724d1252559f152ce4735f51/helper/testhelpers/logical/testing.go

The primary goal(s) of this work here is to effectively overhaul or reintroduce it but utilize an actual instance of Vault instead of some of our Vault core test functions, and to avoid calling plugin methods directly through the plugin's `HandleRequest` methods. As for using our own types, we're loosely following the pattern set by Terraform Provider acceptance testing, which has served us (HashiCorp at large) very well over the years.

>     * It is going to be very limiting. For example, some plugins might need to perform actions on external systems between steps, etc.

It may be limiting at times or in certain areas, yes. The framework is not envisioned as an end-all for plugin testing, but instead another tool available, and one that comes with the Vault SDK. We felt that a plugin testing framework that manages an external Vault instances, other external infrastructure, and the required coordination between them may be better suited to be a separate external tool/framework. While this framework does some of that and has limitations, we hope it to be a compromise. Hopefully you can agree!

>     * Because of the above, it will be tempting to add more features into the framework and it will increase maintanance costs. The fact that the package is going to be used externally adds more flavour into this.

The intent / design is meant for a specific use-case or style of tests, and as a result will likely not be the most robust framework possible. I'm hopeful we can keep the framework focused on that use-case and resist the temptation to add too much and increase the maintenance costs.

>     * It is going to be a quite big piece for a very specific purpose which plugin authors will need to learn.

Plugin authors don’t have to learn this if they don’t choose to, but I agree that if they do there will be a learning curve. We feel this specific purpose is much needed in Vault and the Vault ecosystem and we certainly hope people will find it worthwhile to learn. 

> I'm wondering if an alternative approach where we are not trying to wrap Go's testing package with abstractions would be a better fit? I (vaguely at this stage) see it as collection of functions you can use inside of the Go's standard test. 

As I mentioned above, the framework has a specific use-case and so we're explicitly trying to be prescriptive in how to write or construct these tests. 

> Something similar to [gomock's](https://github.com/golang/mock) approach where you call controller functions inside of a test:

Unfortunately I haven't yet invested the time to learn how to use `golang/mock`, so I'm not sure what your example provides. Could you elaborate and educate me here? If it fits within the paradigm I've described above I'll certainly investigate more. 

> Apologies if I am bothering you with stupid suggestions. I'm not super familiar with the project and probably missing some background for this PR.

No apologies needed! The project was discussed and planned internally and unfortunately not shared with the community until this first PR. Hopefully it will make more sense as follow-up PRs land. 

Thanks again for the feedback and review!

Cheers,
Clint

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	Path string
+
+	// Arguments to pass in the request. These arguments represent payloads sent
+	// to the API.
+	Data map[string]interface{}
+
+	// Check is a function that is called after this step is executed in order to
+	// test that the step executed successfully. If this is not set, then the next
+	// step will be called
+	Check StepCheckFunc
```
pcman312, line 133: What do you think about making this into a slice of checks? Barring that, I can see the value in creating a helper function that merges multiple check functions into one:

```go
func MergeStepCheckFuncs(funcs ...StepCheckFunc) StepCheckFunc {
  return func(secret *api.Secret, actualErr error) error {
    for _, f := range funcs {
      err := f(secret, actualErr)
      if err != nil {
        return err
      }
    }
    return nil
  }
}
```

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	// the response. It is the responsibility of the StepCheckFunc to validate the
+	// error is appropriate or not, if expected.
+	ExpectError bool
+
+	// Unauthenticated will make the request unauthenticated.
+	Unauthenticated bool
+}
+
+// StepCheckFunc is the callback used for Check in Steps.
+type StepCheckFunc func(*api.Secret, error) error
```
pcman312, line 148: `Assertion`
`StepAssertion`
`AssertStep`
?

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+		// sent to the Check function to validate.
+		if step.Check != nil {
+			if err := step.Check(resp, respErr); err != nil {
+				tt.Error(fmt.Errorf("failed step %d: %w", index, err))
+			}
+		}
+	}
+}
+
+func makeRequest(tt TestT, driver StepwiseEnvironment, step Step) (*api.Secret, error) {
```
pcman312, line 320: Should this be a `tt.Helper()`?

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+		// "index" here is only used for logging / output, and not to reference the
+		// step in the slice of steps.
+		index := i + 1
+		if logger.IsWarn() {
+			progress := fmt.Sprintf("%d/%d", index, stepCount)
+			logger.Warn("Executing test step", "step_number", progress)
+		}
+
+		// reset token in case it was cleared
+		client, err := rootClient.Clone()
```
pcman312, line 291: This client isn't used beyond `SetToken` below - should it be passed into `makeRequest` instead of the environment?

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+		// If a step returned an unexpected error, fail the entire test immediately
+		if respErr != nil && !step.ExpectError {
+			tt.Fatal(fmt.Errorf("unexpected error in step %d: %w", index, respErr))
+		}
+
+		// run the associated StepCheckFunc, if any. If an error was expected it is
+		// sent to the Check function to validate.
+		if step.Check != nil {
+			if err := step.Check(resp, respErr); err != nil {
+				tt.Error(fmt.Errorf("failed step %d: %w", index, err))
```
pcman312, line 314: Would some sort of name on the step be valuable here? I'm concerned about the index being misinterpreted as off-by-one when debugging.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	DeleteOperation               = "delete"
+	ListOperation                 = "list"
+	HelpOperation                 = "help"
+)
+
+// StepwiseEnvironment is the interface Environments need to implement to be used in
+// Case to execute each Step
+type StepwiseEnvironment interface {
+	Setup() error
+	Client() (*api.Client, error)
```
pcman312, line 35: Can `Client()` get a comment on it - In particular I'd like it to be clear whether or not this should be creating a new client every time it's called vs returning a cached client.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	path := driver.ExpandPath(step.Path)
+	switch step.Operation {
+	case WriteOperation, UpdateOperation:
+		return client.Logical().Write(path, step.Data)
+	case ReadOperation:
+		// Some operations support reading with data given.
+		// TODO: see how the CLI parses args and turns them into
+		// map[string][]string, or change how step.Data is defined (currently
+		// map[string]interface{})
+		// resp, respErr = client.Logical().ReadWithData(path, step.Data)
```
pcman312, line 342: Commented out code?

sdk/testing/stepwise/stepwise_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+	return nil
+}
+
+// Client creates a Vault API client configured to the mock environment's test
+// server
+func (m *mockEnvironment) Client() (*api.Client, error) {
+	m.l.Lock()
+	defer m.l.Unlock()
+	// this shouldn't be needed but being defensive
```
pcman312, line 236: Should this be an assertion? Would it be valuable to know that this is being called?

sdk/testing/stepwise/stepwise_test.go
```diff
@@ -0,0 +1,282 @@
...
+	if m.ts == nil {
+		if err := m.Setup(); err != nil {
+			return nil, err
+		}
+	}
+
+	if m.client != nil {
```
pcman312, line 243: I'm not seeing where `m.client` is set - am I missing it or does it need to be set prior to the return?

sdk/testing/stepwise/stepwise_test.go
```diff
@@ -0,0 +1,282 @@
+package stepwise
```
pcman312, line 1: I'm seeing a 63% code coverage of the `stepwise.go` file - mostly on the happy path. Are you planning on adding more tests for this in subsequent PRs?

sdk/testing/stepwise/stepwise.go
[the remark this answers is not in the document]
m1kola: > in my mind I've imagined that we'd likely need to eventually support passing in some kind of custom configuration that's specific to the `driver` (now called `environment`).

Yep, I see the following cases:

* Environment supports exactly what specified in `EnvironmentOptions`
* Environment supports what in `EnvironmentOptions` + some extra environment-specific options
* Environment supports less than what specified in `EnvironmentOptions`.

Given the above and the fact that environment is going to be instantiated in plugin's test itself (like in [`TestBackend_basic_docker`, for example](https://github.com/hashicorp/vault/blob/5b8f2f71dbd1fbb4291cb90ba4048b3f817a4f0b/builtin/logical/transit/stepwise_test.go#L18-L40)), I think it makes sense to let environments to define their own options and the way how they accept options (constructor arguments for each option, a struct for everything or functional options as @pcman312 suggested or whathever). 

This way we will keep the framework a bit simpler: it will rely on the interface only and will not care about how an envirionment was instantiated.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	UpdateOperation               = "update"
+	ReadOperation                 = "read"
+	DeleteOperation               = "delete"
+	ListOperation                 = "list"
+	HelpOperation                 = "help"
+)
+
+// StepwiseEnvironment is the interface Environments need to implement to be used in
+// Case to execute each Step
+type StepwiseEnvironment interface {
```
m1kola, line 33: Nit: I think there is not need to add `Stepwise` as a prefix here: the package is called `stepwise` already.

m1kola:
@catsby thanks for the clarifications. This is very helpful in learning more background for this PR and Vault in general.

I understand that the idea is to provide a framework for a very specific use-case of tests, but I'm not convinced that there is a need to enforce such a strict structure (defining a `Case`, `Steps` for each test, etc).

> As for using our own types, we're loosely following the pattern set by Terraform Provider acceptance testing, which has served us (HashiCorp at large) very well over the years.

Thanks for this reference. I had a quick look at the testing framework from `terraform-plugin-sdk`: I think the number of configuration options the terraform plugin testing framework provides and amount of commits it receives supports my poin about adding complexity: it is going to be hard to resist the temptation to add too much :) I think we should give as little options as possible and make the framework as minimal as possible. Give more power and responsibilities to the plugin authors.

I'm looking at [`TestBackend_basic_docker`](https://github.com/hashicorp/vault/blob/5b8f2f71dbd1fbb4291cb90ba4048b3f817a4f0b/builtin/logical/transit/stepwise_test.go#L18-L40) and keeping in mind other cases... I think I'll be able to come up with a design which will have the following characteristics (among others):

* Gives a way to setup/teardown an environment
* Exports fewer types and options
  * Ideally the core of framework only exports something like the `StepwiseEnvironment` interface with `Setup`/`Teardown`, `Client` and (maybe) few other  functions.
    * But it should be ok to add and export implementations of the interfaces (eg. docker environment) to the SDK as a subpackage.
    * Shoukd be easy to add own implementation of the interfaces
  * Less types & options should make it easier to maintain it as part of the SDK
  * Should also make it easier to learn (looking at 10s of options of `terraform-plugin-sdk` again and it's not easy to digest)
* Does not impose a test structure
  * Plugin authors will be using the testing framework from Go standard library more direclty. It is a tool familiar to many Go developers (easier adoption)
  * Our testing SDK will work in a similar way to `gomock`'s `Controller`: plugin authors will be calling functions from the SDK as part of standard Go test (I know you are not familiar with `gomock`, will try to provide a better example ASAP: see suggestion below)
  * Gives plugin authors more control & power (can change state of the external system, for example)
  * But makes plugin authors to be responsible for a bit more. For example, they will be explicitly calling the client and will have to do assertions such as [`ExpectError`](https://github.com/hashicorp/vault/pull/9166/files#diff-b4517a2127f04dc1845c02097e79af67R141).
* (Maybe) Provides utilities (such as `CompilePlugin`).
* Will not be compatible with `helper/testhelpers/testing.go` (not sure if it's a goal as part of current Stepwise efforts)
* Will not be feel familiar to people who already used `terraform-plugin-sdk` or `helper/testhelpers/testing.go` (again, not sure if it's important)

@catsby @pcman312 what do you think? Is it worth exploring? If yes, as a starting point I can rewrite [`TestBackend_basic_docker`](https://github.com/hashicorp/vault/blob/5b8f2f71dbd1fbb4291cb90ba4048b3f817a4f0b/builtin/logical/transit/stepwise_test.go#L18-L40) to use the design I have in mind (roughly, without implementation of the framework). It will demonstrate us how it is going to look from plugin author's perspective. After this I can write more detailed design document or jump straight to the implementation (going to re-use a lot of things from this PR).

sdk/testing/stepwise/helpers.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// calculate sha256
+	f, err := os.Open(binPath)
+	if err != nil {
+		return "", "", "", err
+	}
+
+	h := sha256.New()
+	if _, ioErr := io.Copy(h, f); ioErr != nil {
+		panic(ioErr)
```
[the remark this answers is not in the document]
catsby: Indeed! I think early on I was thinking this kind of testing code should panic but in hindsight I don't think that's true anymore. Updated in 5d9ce7023

sdk/testing/stepwise/helpers.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	_ = f.Close()
```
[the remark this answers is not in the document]
catsby: Sure, updated in 3b3ad87fb

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+
+	log "github.com/hashicorp/go-hclog"
+
+	"github.com/hashicorp/vault/api"
+	"github.com/hashicorp/vault/sdk/helper/consts"
+	"github.com/hashicorp/vault/sdk/helper/logging"
+)
+
+// StepOperation defines operations each step could preform
+type StepOperation string
```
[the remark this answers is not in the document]
catsby: This was intentional; I wanted developers to consistently use the `stepwise` package when building these tests, instead of using mostly `stepwise` and then 1 part (the operations) from the `logical` package. 

I've updated the code comment in c88cff0fb to reflect this.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+
+	// ExpandPath adds any Namespace or mount path to the user defined path
+	ExpandPath(string) string
+
+	// MountPath returns the path the plugin is mounted at
+	MountPath() string
+
+	// BarrierKeys returns the keys used to seal/unseal the cluster. Used for
+	// debugging. TODO verify we should provide this
+	//BarrierKeys()        [][]byte
```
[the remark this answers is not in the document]
catsby: Yup, removed in 54b575711

sdk/testing/stepwise/stepwise_test.go
catsby: Whoops! Good catch, it was recreating the client each time and never setting it. Fixed in fb96110b7

sdk/testing/stepwise/stepwise.go
catsby: Removed in fb96110b7

sdk/testing/stepwise/stepwise_test.go
catsby: No, I don't think it should or that would be valuable. I changed it in fb96110b7 to simply return `nil` with an error if `ts` is `nil`.

sdk/testing/stepwise/stepwise.go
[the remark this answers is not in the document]
catsby: Resolving this for now

sdk/testing/stepwise/stepwise.go
catsby: Updated in 359a4ff15, along with other methods in the interface

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	// Unauthenticated will make the request unauthenticated.
+	Unauthenticated bool
+}
+
+// StepCheckFunc is the callback used for Check in Steps.
+type StepCheckFunc func(*api.Secret, error) error
+
+// Case represents a scenario we want to test which involves a series of
+// steps to be followed sequentially, evaluating the results after each step.
+type Case struct {
```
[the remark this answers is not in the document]
catsby: I originally named it `Case` to differentiate from the legacy test `TestCase`. In hindsight I think I should move the legacy files to their own package at the same level as `stepwise` and split the two, because right now the legacy testing is a part of the `stepwise` package. I'm going to defer that change and make a note internally to do so after this PR is merged, otherwise it will introduce a new set of changes here.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	case PluginTypeSecrets:
+		return "secret"
+	default:
+		return "unsupported"
+	}
+}
+
+// EnvironmentOptions are a collection of options each step driver should
+// support
+type EnvironmentOptions struct {
```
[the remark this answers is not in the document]
catsby: The intent of these options is for users to customize things specific to Vault or the usage of this plugin, e.g. mount path (prefix), or the plugin type (auth, secret). The functional options pattern looks promising for configuring things specific to the environment, which isn't the intent of this type. 

That said of I see the type is poorly named. I renamed it to `MountOptions` in f22b7b69a to better reflect it's purpose. 

As for the pattern, I'm going to defer that for now as we don't currently plan to offer much configuration of the environment. When that comes though, that seems like a likely pattern to apply here.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	// omitted the mount path will default to the PluginName with a random suffix.
+	MountPathPrefix string
+
+	// Name is used to register the plugin. This can be arbitray but should be a
+	// reasonable value. For an example, if the plugin in test is a secret backend
+	// that generates UUIDs with the name "vault-plugin-secrets-uuid", then "uuid"
+	// or "test-uuid" would be reasonable. The name is used for lookups in the
+	// catalog. See "name" in the "Register Plugin" endpoint docs:
+	// - https://www.vaultproject.io/api-docs/system/plugins-catalog#register-plugin
+	Name string
```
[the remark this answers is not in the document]
catsby: Agreed, renamed in 0cbfe670a

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	// step will be called
+	Check StepCheckFunc
+
+	// ExpectError indicates if this step is expected to return an error. If the
+	// step operation returns an error and ExpectError is not set, the test will
+	// fail immediately and the StepCheckFunc will not be ran. If ExpectError is
+	// true, the StepCheckFunc will be called (if any) and include the error from
+	// the response. It is the responsibility of the StepCheckFunc to validate the
+	// error is appropriate or not, if expected.
+	ExpectError bool
```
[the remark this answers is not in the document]
catsby: Yeah that makes sense. `ExpectError` was a holdover from the legacy frameworks [`ErrorOk`][errorok] that may have had use in early development of `stepwise` but you're correct I don't think it's useful here. I removed it in a8c2db64a, thanks 😄 


[errorok]: https://github.com/hashicorp/vault/blob/2806998bdeb8b2e88d98223fc607bfba211f8100/sdk/testing/stepwise/legacy.go#L97

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+	PreCheck func()
+
+	// Steps are the set of operations that are run for this test case.
+	Steps []Step
+
+	// SkipTeardown allows the TestTeardownFunc to be skipped, leaving any
+	// infrastructure created during Environment setup to remain. Depending on the
+	// Environment used this could incur costs the user is responsible for.
+	// TODO maybe better wording here
+	SkipTeardown bool
```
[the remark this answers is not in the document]
catsby: This is an interesting suggestion that I'm going to defer on for now. In general I don't imagine `Teardown()` is something that should offer or require customization other than what is specific to the environment itself. The environment should completely tear down any infrastructure created, be it a docker container or a cluster in a cloud provider.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+func Run(tt TestT, c Case) {
+	tt.Helper()
+	// We only run acceptance tests if an env var is set because they're
+	// slow and generally require some outside configuration.
+	if os.Getenv(TestEnvVar) == "" {
+		tt.Skip(fmt.Sprintf(
+			"Acceptance tests skipped unless env '%s' set",
+			TestEnvVar))
+		return
+	}
```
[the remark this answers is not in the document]
catsby: 👍 added in 8479973c8 moved the `testing.Verbose()` check in there too

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+
+	// We require verbose mode so that the user knows what is going on.
+	if !testing.Verbose() {
+		tt.Fatal("Acceptance tests must be run with the -v flag on tests")
+		return
```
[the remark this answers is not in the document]
catsby: Agreed. Removed in ebccf4af2

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+// steps to be followed sequentially, evaluating the results after each step.
+type Case struct {
+	// Environment is used to setup the Vault instance and provide the client that
+	// will be used to drive the tests
+	Environment StepwiseEnvironment
+
+	// Precheck, if non-nil, will be called once before the test case
+	// runs at all. This can be used for some validation prior to the
+	// test running.
+	PreCheck func()
```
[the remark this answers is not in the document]
catsby: This is a vestigial item from the legacy framework that I'm going to simply remove for now. I agree that if this kind of check is needed, it should at least accept `TestT` and return an error. Removed in 7bf3e1a19 for now.

sdk/testing/stepwise/stepwise.go
catsby: Renamed to `AssertionFunc`, which I think makes more sense

sdk/testing/stepwise/stepwise.go
catsby: Yes! Added in b0945c403

sdk/testing/stepwise/stepwise_test.go
catsby: Yes. As the first environment gets added and the first few actual tests to use the framework arrive, I plan to expand on the test coverage

sdk/testing/stepwise/stepwise.go
catsby: Indeed, the name stutters. I've renamed it to simply `Environment`

sdk/testing/stepwise/legacy.go
```diff
@@ -492,4 +492,5 @@ type TestT interface {
 	Error(args ...interface{})
 	Fatal(args ...interface{})
 	Skip(args ...interface{})
+	Helper()
```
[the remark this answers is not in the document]
catsby: I'd like to defer this to a follow-up PR unless you strongly object

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,351 @@
...
+// Tests will fail unless the verbose flag (`go test -v`, or explicitly
+// the "-test.v" flag) is set. Because some acceptance tests take quite
+// long, we require the verbose flag so users are able to see progress
+// output.
+func Run(tt TestT, c Case) {
+	tt.Helper()
+	// We only run acceptance tests if an env var is set because they're
+	// slow and generally require some outside configuration.
+	if os.Getenv(TestEnvVar) == "" {
+		tt.Skip(fmt.Sprintf(
```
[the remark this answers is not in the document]
catsby: Deferring, it will come with https://github.com/hashicorp/vault/pull/9166#discussion_r440365296

sdk/testing/stepwise/stepwise.go
catsby: I had meant to add this, the legacy framework has one. I'm deferring this for now, while it's a simple addtion as you've shown, I don't have any immedate tests that use it and I feel we can add it later (and add tests for it)

sdk/testing/stepwise/stepwise.go
catsby: I'm going to defer addressing this for now. In the future I would like to support each test case or test step being ran with a token created with user defined policies attached instead of using the root token. At this time I don't know what that looks like, so for now I would like to leave this as is to continue to support the `Unauthenticated` feature and revisit this when I have a better idea on that feature.

sdk/testing/stepwise/stepwise.go
catsby: An optional `name` could be useful here, yes. I'm going to defer the addition of `name` for now. I've updated the documentation on `case.Steps` to reflect the 1-based indexing used in logging in 72e24e7a8

sdk/testing/stepwise/legacy.go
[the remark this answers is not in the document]
pcman312: :+1:

catsby:
Hey @m1kola - 

Thank you again for your continued feedback! 

Regarding adding complexity, we certainly don’t consider the framework “done” as of this PR, with more features, changes, and improvements planned in the near future. The focus will remain the same though: providing a prescriptive way to test a plugin through invoking the Vault API itself. I would agree with your sentiments on empowering plugin authors here if we intended or even envisioned `stepwise` to be the end-all solution for testing plugins, however that is not the case. The `stepwise` framework is meant to be a structured method of testing plugins in a specific way, effectively interacting with the plugin through Vault as a user would.  We realize that a testing framework specific to the project it is included in will be unfamiliar for many Go developers, which is why `stepwise` should not be the only way to test plugins. The intent is to provide a structured way of opaque testing of plugins, with as little mocking as possible (preferably zero mocking). 

I don’t mean to discourage your efforts to add or improve Vault or Vault plugin testing, I hope you don’t receive this reply in that way. There currently exists several internal testing functions and methods for wiring a plugin up with a Vault core (often a simulated one) such as [TestCore][testcore] or [NewTestCluster][newtestcluster] in `vault/testing.go`. Alternatively you can often bypass having a core at all and directly call backend methods with minimal “Vault” included, by invoking the `HandleRequest` method on the `framework.Backend` type (some examples in the [Transit backend here][transit]). These methods and testing styles meant to give developers flexibility in how they structure and test plugins, and are absolutely a valuable tool that I’m sure would benefit from many of the additions you mention. 

I want to reiterate that I don’t mean to discourage you from pursuing the ideas you have for improving plugin testing for Vault, quite the opposite! I would love to see and hear more on your ideas, but I want to be clear that they run contrary to the goals of the `stepwise` framework as it stands today. With time and reflection, perhaps some of those ideas will come to be included, if appropriate. 

Thank you again for your input here! I hope this reply answers your questions, at least for now.

Cheers,
Clint



[testcore]: https://github.com/hashicorp/vault/blob/18f73cf35c4751f0bdd6811062e1c17ef771e824/vault/testing.go#L97
[newtestcluster]: https://github.com/hashicorp/vault/blob/18f73cf35c4751f0bdd6811062e1c17ef771e824/vault/testing.go#L1094
[transit]: https://github.com/hashicorp/vault/blob/18f73cf35c4751f0bdd6811062e1c17ef771e824/builtin/logical/transit/backend_test.go#L116

sdk/testing/stepwise/stepwise.go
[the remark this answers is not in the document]
catsby: I renamed `DriverOptions` to `EnvironmentOptions`, and then later simply `MountOptions`. The intent for this type is to configure how the plugin is mounted, specifically the path. `Driver` was outdated and `Environment` was a misnomer, so I changed to `MountOptions`. 

Going forward, if it's appropriate to enable developers to specify environment specific options, I think the functional options method suggested makes the most sense. For now though, that feature is deferred until we know more about how it would be used.

PR #9167, opened 2020-06-08
Add backend type to audit logs

sgmiller:
Especially interested in feedback on whether this is the right name for the
new field.


audit/format.go
```diff
@@ -350,6 +351,7 @@ type AuditRequest struct {
 
 type AuditResponse struct {
 	Auth     *AuditAuth             `json:"auth,omitempty"`
+	Backend  string                 `json:"backend,omitempty"`
```
mgritter, line 354: I'd feel a little bit better if it was backend_type, but I'm also worried that "backend" is new terminology for an end user?  We generally refer to "secret engine" or "auth method" but the documentation does mention "backend" as a concept too.

I looked for other examples in our API or audit entries.  Mount status has just "type".

The new metrics use auth_method or secret_engine, because I split usage so they didn't overlap.

backend_type seems preferable to be clearer that it's not the mount point or path?

sgmiller: Trying MountType instead, thoughts?

sgmiller:
Added MountType to requests as well per bk's comment.  @mgritter: the unit tests for audit log test by having NoopAudit populate in memory structs of the output using the same datastructures and then deepEqual comparing them.  So the new fields are covered by that.

ncabatoff:
Looks like you're getting a test panic now due to entry.Type being nil in handleCancelableRequest.

vault/request_handling.go
```diff
@@ -548,6 +548,7 @@ func (c *Core) handleCancelableRequest(ctx context.Context, ns *namespace.Namesp
 				Auth:                auth,
 				Request:             req,
 				Response:            auditResp,
+				MountType:           entry.Type,
```
mgritter, line 551: OK, just to note what we said in our offline conversation, this now seems unnecessary because we get the MountType from Request instead.

PR #8340, opened 2020-02-12
Add support for telemetry option `metrics_prefix`.  

ncabatoff:
When set, instead of starting with `vault_`, metrics will start with `${prefix}_`.

PR #9171, opened 2020-06-09
add metrics_prefix doc note

mikegreen:
Adding metrics_prefix documentation 
added via https://github.com/hashicorp/vault/pull/8340


website/pages/docs/configuration/telemetry.mdx
```diff
@@ -33,6 +33,7 @@ The following options are available on all telemetry configurations.
 - `enable_hostname_label` `(bool: false)` - Specifies if all metric values should
   contain the `host` label with the local hostname. It is recommended to enable
   `disable_hostname` if this option is used.
+- `metrics_prefix` `(string: "vault")` - The prefix used while writing all telemetry data. By default, this is set to "vault". This was added in Vault 1.4.
```
ncabatoff, line 36: We don't usually note when a new feature/option was introduced.  It's not a bad idea, but I think we should be consistent and agree collectively on whether we want to do that.

PR #9177, opened 2020-06-09
Add a simple sealed gauge, updated when seal status changes

sgmiller:


vault/core.go
```diff
@@ -783,6 +783,8 @@ func NewCore(conf *CoreConfig) (*Core, error) {
 	c.rawConfig.Store(conf.RawConfig)
 
 	atomic.StoreUint32(c.sealed, 1)
+	metrics.SetGauge([]string{"core", "sealed"}, 1)
```
mgritter, line 786: Can we use core.metricSink instead, so that we get a cluster label?

I think that setting the guage only on transitions is not good enough.  Say Vault is up for 60 days but we have a 30 day retention period on metrics.  Then the dashboard stops working.  I think that we need to put this gauge in the emitMetrics goroutine instead.

mgritter: FWIW, Telegraf does remember the value of a gauge for some period of time if it hasn't received an update, but not indefinitely.  That could give us a little extra window, but probably not enough.

sgmiller: I see.  I was expecting prometheus-like behavior where the value is polled on each scrape.

vault/core.go
```diff
@@ -1710,6 +1713,7 @@ func (c *Core) sealInternalWithOptions(grabStateLock, keepHALock, shutdownRaft b
 	if swapped := atomic.CompareAndSwapUint32(c.sealed, 0, 1); !swapped {
 		return nil
 	}
+	metrics.SetGauge([]string{"core", "sealed"}, 1)
```
mgritter, line 1716: Missed this one.

vault/core.go
```diff
@@ -2090,6 +2094,13 @@ func (c *Core) emitMetrics(stopCh chan struct{}) {
 			if c.expiration != nil {
 				c.expiration.emitMetrics()
 			}
+			//Refresh the sealed gauge
+			if c.Sealed() {
+				metrics.SetGauge([]string{"core", "unsealed"}, 1)
+			} else {
+				metrics.SetGauge([]string{"core", "unsealed"}, 0)
+			}
```
briankassouf, line 2102: I think the values here are switched, elsewhere unsealed==0 when c.Sealed() is true.

sgmiller: Thanks guys, fixed above.

PR #9182, opened 2020-06-10
Changes to expiration manager to walk tokens 

mgritter:
(including non-expiring ones.)
Count by namespace in token manager.


mgritter:
Posting for early review of changes to expiration manager.
Still to do: uniquify the policy lists so they share a single array when identical.
Also TODO: start the actual collector and test the gauge output (and implement the other two gauges along the same lines.)

vault/expiration.go
```diff
@@ -1341,13 +1355,37 @@ func (m *ExpirationManager) inMemoryLeaseInfo(le *leaseEntry) *leaseEntry {
 	//   policies -- stored in Auth object
 	//   auth method -- derived from lease.Path
 	if le.Auth != nil {
-		// TODO: make unique so all copies share a single array
-		ret.Auth.Policies = le.Auth.Policies
+		// Ensure that list of policies is not copied more than
+		// once. This menthod is called with pendingLock held.
```
ncabatoff, line 1359:
```suggestion
		// once. This method is called with pendingLock held.
```

vault/expiration.go
```diff
@@ -569,6 +589,11 @@ func (m *ExpirationManager) Stop() error {
 		m.leaseCount--
 		return true
 	})
+	m.nonexpiring.Range(func(key, value interface{}) bool {
+		m.nonexpiring.Delete(key)
+		return true
+	})
```
briankassouf, line 595: Do we need to itterate over the list or can we just re-assign the nonexpiring variable to a newly created sync.Map?

mgritter: Hm, I assumed the behavior was similar enough to pending that we wanted to treat it in the same way.  But nothing gets looked up in nonexpiring in the data path.

vault/expiration.go
```diff
@@ -147,15 +157,19 @@ func expireLeaseStrategyRevoke(ctx context.Context, m *ExpirationManager, le *le
...
+		tokenView:   view.SubView(tokenViewPrefix),
+		tokenStore:  c.tokenStore,
+		logger:      logger,
+		pending:     sync.Map{},
+		nonexpiring: sync.Map{},
+		leaseCount:  0,
+		tidyLock:    new(int32),
+
+		uniquePolicies:      make(map[string][]string),
+		emptyUniquePolicies: time.NewTicker(7 * 24 * time.Hour),
```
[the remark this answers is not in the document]
mgritter: Yeah... I thought about it.  It seems like it would be substantial additional overhead, because go-cache does its own locking, and tracks expiration per entry. We really only need to expire because we don't want indefinite growth.

For the number of distinct policies we have, maybe the per-entry costs are fine, but it seemed to introduce a lot of uncertainty.

vault/expiration.go
mgritter: OK, I remembered (or re-reasoned) about why it is done this way.

Replacing m.pending and m.nonexpiring with a new value is a race condition because WalkTokens doesn't grab pendingLock before accessing them. That's the race condition Jeff warned me about in the initial API design. I added a comment explaining this.

However, I do think it's possible we could grab a pointer to the sync.Map with the lock held and then use it without the lock. I didn't introduce that in this PR.

PR #9191, opened 2020-06-10
Validate physical CockroachDB table config value before using it

pcman312:
Adding validation to the CockroachDB physical backend's table config value.

From the [CockroachDB docs](https://www.cockroachlabs.com/docs/stable/keywords-and-identifiers.html#identifiers)
- All values that accept an identifier must:
  - Begin with a Unicode letter or an underscore (_). Subsequent characters can be letters,
  - underscores, digits (0-9), or dollar signs ($).
- Not equal any SQL keyword unless the keyword is accepted by the element's syntax.

The rules implemented here are a bit more restricted than what is listed on the docs for ease of implementation. Also explicitly rejecting backticks, single quotes, and double quotes to prevent problems when using it in SQL queries.


physical/cockroachdb/keywords.go
```diff
@@ -0,0 +1,440 @@
...
+		"VALUE":                     true,
+		"VARYING":                   true,
+		"VIEW":                      true,
+		"WITHIN":                    true,
+		"WITHOUT":                   true,
+		"WRITE":                     true,
+		"YEAR":                      true,
+		"ZONE":                      true,
+	}
+)
```
Valarissa, line 440: Long list is long :)

PR #9192, opened 2020-06-10
Implement lease expiration as a priority queue rather than Go timers.

sgmiller:
For 100k active leases, seems to use 90% less RAM and depending on concurrency, competitive in lease registration speed (30% faster single threaded, starts to lose to go timers on contention ~16 threads)

Modified our Priority Queue implementation to add Peek and Clear, and disabled it's deep copy of the items stored in the queue.  This probably breaks other code, so it'd need to be configurable on New.

Credit to @ncabatoff for the idea.

### DIFF
--- a/sdk/queue/priority_queue.go
+++ b/sdk/queue/priority_queue.go
@@ -8,9 +8,9 @@ package queue
 import (
 	"container/heap"
 	"errors"
-	"sync"
-
+	sync2 "github.com/hashicorp/vault/sdk/helper/sync"
 	"github.com/mitchellh/copystructure"
+	"sync"
 )
 
 // ErrEmpty is returned for queues with no items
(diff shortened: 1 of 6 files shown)


sdk/queue/priority_queue.go
```diff
@@ -150,6 +150,46 @@ func (pq *PriorityQueue) PopByKey(key string) (*Item, error) {
...
+	if i != nil {
+		return i.(*Item)
+	}
+	return nil
+}
+
+// WhileEmpty blocks the caller until the queue contains an item
+func (pq *PriorityQueue) WaitWhileEmpty(doneChan chan struct{}) {
+	for pq.Len() == 0 {
+		pq.lock.Lock()
```
kalafut, line 177: How are items added with this lock held?

sdk/queue/priority_queue.go
```diff
@@ -150,6 +150,46 @@ func (pq *PriorityQueue) PopByKey(key string) (*Item, error) {
...
+	i := pq.data.peek()
+	if i != nil {
+		return i.(*Item)
+	}
+	return nil
+}
+
+// WhileEmpty blocks the caller until the queue contains an item
+func (pq *PriorityQueue) WaitWhileEmpty(doneChan chan struct{}) {
+	for pq.Len() == 0 {
```
kalafut, line 176: This looks racy given there is no lock held. Once we enter the block length could be > 0

PR #9194, opened 2020-06-10
replacing "a key usage mode" on transit doc page as it is confusing

assareh:
Since the context of this page is transit and encryption keys, the use of the word "key" to mean effectively common seems ill advised. Proposing an alternative wording.


ncabatoff:
Thanks!

PR #9195, opened 2020-06-11
Update kubernetes auth plugin

kalafut:


riuvshyn:
here is also related doc upd PR https://github.com/hashicorp/vault/pull/9142

PR #9198, opened 2020-06-11
CONTRIBUTING.md: Link to Discussion forum

evanj:
The Vault Google Group has been replaced with Discuss. Link to the
correct place.


evanj:
Re: CLA: I work for Datadog, and this edit was done at work, on my work computer. My understanding is Datadog has a corporate CLA on file with Hashicorp.

kalafut:
Thanks for finding this! I can't merge w/o CLA but have applied these fixes directly since it's a fixed link and not IP. I know some other Datadog devs have contributed code under the CLA, so you might want to check into that on your end. If you were to submit code patches, for example, they would be blocked until the CLA check cleared.

evanj:
Awesome thanks!

PR #9199, opened 2020-06-11
kv secrets mount case-sensitivity note

zalary:


website/pages/docs/secrets/kv/index.mdx
```diff
@@ -13,6 +13,8 @@ modes. It can be a generic Key-Value store that stores one value for a key.
 Versioning can be enabled and a configurable number of versions for each key
 will be stored.
 
+Note that `kv` secrets mounts are case-sensitive.
```
ncabatoff, line 16: What does this mean?  That we preserve case for keys?

yhyakuna: > What does this mean? That we preserve case for keys?

@ncabatoff  apparently so?  https://app.asana.com/0/112957393038545/1109620483178716/f

mikegreen, line 16: To comply with the use of "note" in other places, could we use this syntax to call it out in a box? 
```suggestion
~> **NOTE:** 
```

aphorise:
I believe this request may still be relevant particularly since there are confusions regarding the default case sensitivity that's expected from most systems.

@yhyakuna @brianshumate please advice.

yhyakuna:
Since this PR was originally opened, the folder structure have changed. I opened a [PR 16473](https://github.com/hashicorp/vault/pull/16743), so closing this.

PR #9216, opened 2020-06-12
Ui/add changelog link

noelledaley:
# Link to Changelog version in UI Footer

image

This PR adds a link to the changelog at the version of Vault currently being used. This allows users to quickly view relevant changes, particularly when we have added or changed behavior in the UI.

## Notes
The helper used to generate the URL assumes that changelog headers will match the following format:
```
## v1.5.0
### Month, DD, YYYY

## v1.4.5
### Month, DD, YYY
```

Since this new header format will only be used for Vault versions 1.4.3 and beyond, all previous versions just link back to the base changelog link. As a result this PR does not need to be backported.


ui/lib/core/addon/helpers/changelog-url-for.js
```diff
@@ -0,0 +1,33 @@
...
+## v1.5.0
+### Month, DD, YYYY
+
+## v1.4.5
+### Month, DD, YYY
+
+etc.
+*/
+
+export function changelogUrlFor([version]) {
```
chelshaw, line 16: Why do we expect version to be in an array?

noelledaley: this is ember weirdness that i don't fully understand, to be honest. ember helpers receive arguments as an array ¯\_(ツ)_/¯

chelshaw: That is bizarre. But ok 😂

Monkeychip: Re: helpers taking arrays: I read a little about this a while ago, and though I can't say I've done this, it's because helpers can take named arguments.  https://guides.emberjs.com/release/components/helper-functions/

ui/lib/core/addon/helpers/changelog-url-for.js
```diff
@@ -0,0 +1,37 @@
...
+  let url = 'http://www.github.com/hashicorp/vault/blob/master/CHANGELOG.md#';
+
+  try {
+    // strip the '+prem' from enterprise versions and remove periods
+    let versionNumber = version
```
mladlow, line 21: I have no idea what all the various versions look like from the perspective of the UI (like the HSM ones). So, I can't comment if this will work in all cases, which is my only concern.

ui/lib/core/addon/helpers/changelog-url-for.js
```diff
@@ -0,0 +1,37 @@
...
+### Month, DD, YYYY
+
+## v1.4.5
+### Month, DD, YYY
+
+etc.
+*/
+
+export function changelogUrlFor([version]) {
+  let url = 'http://www.github.com/hashicorp/vault/blob/master/CHANGELOG.md#';
```
mladlow, line 17: Should this be https?

noelledaley: ah, yes! nice catch.

ui/lib/core/addon/helpers/changelog-url-for.js
noelledaley: yeah, we're not totally sure what all of the versions look like either which is why i wrapped this in a try block. this way, if we encounter any versions that don't meet the expected format we'll catch the error log the version. if anything fails we default to the regular 'ol changelog which should prevent users from having to deal with broken links.

PR #9219, opened 2020-06-13
Add real connected state and API address

noelledaley:
This PR leverages the [newly available](https://github.com/hashicorp/vault/pull/9029) `api_address` and `connected_state` from the status endpoint to add a URL from a Replication Primary dashboard to a secondary UI.

image

## Notes
- when a secondary is disconnected, there will be no `api_address`, which is why we leave that table data box empty
- these links are expected to be broken during development. unfortunately testing the links is very difficult because our ember server proxies requests from port `:4200` to `:8200` [see here](https://github.com/hashicorp/vault/blob/aaf7245c34622a36ccd9751ec93d9c9c18ece84d/ui/package.json#L19-L20). this means that in the example screenshot above the vault `api_address` _is_ indeed `http://127.0.0.1:8202/`, but ember is actually serving the ui at `http://127.0.0.1:4202/ui/`. as a result, _direct_ requests to `:8202` cannot be found. i chatted through this with @briankassouf and @calvn and they believe this should only be an issue during development. in a production environment the UI should always be served at the `api_address`.
- after discussion, it seems like the best way for us to test these links would be to build an enterprise binary that includes these changes after this PR is merged. from there we can run `make-static-dist` to see what the UI looks like in a production environment. if anyone has any other suggestions, please share! 🎉 

See https://github.com/hashicorp/vault/pull/9029 for more details on the backend implementation.

### DIFF
--- a/ui/app/templates/vault/cluster/replication-dr-promote/details.hbs
+++ b/ui/app/templates/vault/cluster/replication-dr-promote/details.hbs
@@ -32,7 +32,7 @@
       {{!-- passing in component to render so that the yielded components are flexible based on the dashboard --}}
         @componentToRender='replication-secondary-card' as |Dashboard|>
         <Dashboard.card
-          @title="States"
+          @title="Status"
         />
         <Dashboard.card
           @title="Primary cluster"
--- a/ui/app/models/replication-attributes.js
+++ b/ui/app/models/replication-attributes.js
@@ -18,6 +18,7 @@ export default Fragment.extend({
   isPrimary: match('mode', /primary/),
 
   knownSecondaries: attr('array'),
+  secondaries: attr('array'),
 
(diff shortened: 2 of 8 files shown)


ui/app/templates/vault/cluster/replication-dr-promote/details.hbs
```diff
@@ -32,7 +32,7 @@
       {{!-- passing in component to render so that the yielded components are flexible based on the dashboard --}}
         @componentToRender='replication-secondary-card' as |Dashboard|>
         <Dashboard.card
-          @title="States"
+          @title="Status"
```
noelledaley, line 35: this fixed a bug in which the secondary dashboard rendered 2 cards with the `known_primary_cluster_addr` content inside it

PR #9226, opened 2020-06-15
Move sdk/helper/random -> helper/random

pcman312:
This package is being moved because this code was originally intended to be used within plugins, however the design of password policies has changed such that this is no longer needed. Thus, this code doesn't need to be in the public SDK.

**Note:** This package is new for 1.5 so this is not a breaking change.


pcman312:
Following up from my offline discussion with @Valarissa:
There are tests in place on the password generation code and the change to `StaticSystemView` is only used in tests. We lose the connection between the actual generation code in tests that are in the `vault` package however both sides are still being tested independently. The `dynamicSystemView` also has tests that are using the real password generation code.

PR #9233, opened 2020-06-16
Calculate percentage when displaying progress in tidy operation

jramosf:
The current implementation displays "progress" as a counter sent to log every 500 records, but we don't really know what's the actual **progress**. This replaces that number with a percentage number.


vault/token_store.go
```diff
@@ -1818,10 +1818,11 @@ func (ts *TokenStore) handleTidy(ctx context.Context, req *logical.Request, data
 				}
 
 				var deletedChildrenCount int64
-				for _, child := range children {
+				for index, child := range children {
 					countParentList++
 					if countParentList%500 == 0 {
-						ts.logger.Info("checking validity of tokens in secondary index list", "progress", countParentList)
+						progress := float64(index)/float64(len(children))*100
+						ts.logger.Info("checking validity of tokens in secondary index list", "progress", progress)
```
calvn, line 1825: Instead of replacing the count value, can we include that in addition to the percentage value? 

```
ts.logger.Info("checking validity of tokens in secondary index list", "progress", countParentList, "percent_complete", percentComplete)
```

jramosf: Sure thing! I've pushed a new commit with the change

vault/token_store.go
```diff
@@ -1878,10 +1879,11 @@ func (ts *TokenStore) handleTidy(ctx context.Context, req *logical.Request, data
 			// For each of the accessor, see if the token ID associated with it is
 			// a valid one. If not, delete the leases associated with that token
 			// and delete the accessor as well.
-			for _, saltedAccessor := range saltedAccessorList {
+			for index, saltedAccessor := range saltedAccessorList {
 				countAccessorList++
 				if countAccessorList%500 == 0 {
-					ts.logger.Info("checking if accessors contain valid tokens", "progress", countAccessorList)
+					percentComplete := float64(index)/float64(len(saltedAccessorList))*100
+					ts.logger.Info("checking if accessors contain valid tokens", "progress", countParentList, "percent_complete", percentComplete)
```
calvn, line 1886:
```suggestion
					ts.logger.Info("checking if accessors contain valid tokens", "progress", countAccessorList, "percent_complete", percentComplete)
```

vault/token_store.go
```diff
@@ -1974,10 +1976,11 @@ func (ts *TokenStore) handleTidy(ctx context.Context, req *logical.Request, data
 			}
 
 			// Revoke invalid cubbyhole storage keys
-			for _, key := range cubbyholeKeys {
+			for index, key := range cubbyholeKeys {
 				countCubbyholeKeys++
 				if countCubbyholeKeys%500 == 0 {
-					ts.logger.Info("checking if there are invalid cubbyholes", "progress", countCubbyholeKeys)
+					percentComplete := float64(index)/float64(len(cubbyholeKeys))*100
+					ts.logger.Info("checking if there are invalid cubbyholes", "progress", countParentList, "percent_complete", percentComplete)
```
calvn, line 1983:
```suggestion
					ts.logger.Info("checking if there are invalid cubbyholes", "progress", countCubbyholeKeys, "percent_complete", percentComplete)
```

vault/token_store.go
```diff
@@ -1818,10 +1818,11 @@ func (ts *TokenStore) handleTidy(ctx context.Context, req *logical.Request, data
 				}
 
 				var deletedChildrenCount int64
-				for _, child := range children {
+				for index, child := range children {
 					countParentList++
 					if countParentList%500 == 0 {
-						ts.logger.Info("checking validity of tokens in secondary index list", "progress", countParentList)
+						percentComplete := float64(index)/float64(len(children))*100
```
calvn, line 1824: This should either be `index+1` or use `countParentList` (same for the other ones).

jramosf: Thanks! I made a typo when changing it.

jramosf:
From what I see the failing pipeline is due to other things rather than this patch (it's also broken on `master` branch). Can you confirm @calvn ?

Thanks

PR #9234, opened 2020-06-16
docs: add role_type to jwt role example

bonifaido:
`role_type` is `"oidc"` by default, so this has to specified in case of JWT.


kalafut:

PR #9238, opened 2020-06-16
Add new Telemetry config options

sgmiller:
Add  maximum_gauge_cardinality, and usage_gauge_period 
   configuration options to the config stanza.

   Update unit tests.

   Document.


internalshared/configutil/telemetry.go
```diff
@@ -35,6 +37,11 @@ type Telemetry struct {
 	DisableHostname     bool   `hcl:"disable_hostname"`
 	EnableHostnameLabel bool   `hcl:"enable_hostname_label"`
 	MetricsPrefix       string `hcl:"metrics_prefix"`
+	ClusterName         string `hcl:"cluster_name"`
```
mgritter, line 40: Scott,

I am really sorry to have not made this clearer (I only left a comment in VLT-099 rather than editing it)  I think this attribute is unnecessary since the config already includes a cluster name elsewhere; it doesn't seem necessary to allow  telemetry to have a different user-provided name?  I just wanted to avoid the machine-generated one.

What do you think?

website/pages/docs/configuration/telemetry.mdx
```diff
@@ -28,6 +28,9 @@ parameters on this page are grouped by the telemetry provider.
 
 The following options are available on all telemetry configurations.
 
+- `cluster_name` `(string: "")` - Specifies the name of the cluster for use in metric labeling.
+- `usage_gauge_period` `(string: "10m")` - Specifies the duration between gauge collections.
```
mgritter, line 32:
```suggestion
- `usage_gauge_period` `(string: "10m")` - Specifies the interval at which high-cardinality usage data is collected, such as token counts, entity counts, and secret counts.
```

command/server.go
```diff
@@ -941,12 +941,14 @@ func (c *ServerCommand) Run(args []string) int {
...
-		UserAgent:   useragent.String(),
-		ClusterName: config.ClusterName,
+		Config:              config.Telemetry,
+		Ui:                  c.UI,
+		ServiceName:         "vault",
+		DisplayName:         "Vault",
+		UserAgent:           useragent.String(),
+		ClusterName:         config.ClusterName,
+		MaxGaugeCardinality: config.Telemetry.MaximumGaugeCardinality,
+		GaugeInterval:       config.Telemetry.UsageGaugePeriod,
```
mgritter, line 951: It feels redundant to have the latter two in SetupTelemetryOpts when they are already in config.Telemetry? I think this is fine, but you could get rid of a few lines if you were inclined?

internalshared/configutil/telemetry.go
```diff
@@ -168,16 +174,32 @@ func parseTelemetry(result *SharedConfig, list *ast.ObjectList) error {
 		result.Telemetry.PrometheusRetentionTime = PrometheusDefaultRetentionTime
 	}
 
+	if result.Telemetry.UsageGaugePeriodRaw != nil {
+		var err error
+		if result.Telemetry.UsageGaugePeriod, err = parseutil.ParseDurationSecond(result.Telemetry.UsageGaugePeriodRaw); err != nil {
```
mgritter, line 179: I took another look at VLT-099; one thing we need is a way to disable this feature entirely.

My suggested solution was to accept "none" as a string, and set the gauge period to zero.  But if that's infeasible we can just document "0" as the special value.

sgmiller: It seems easy enough to check for none right here.

command/server.go
sgmiller: Yeah, good point.

PR #9239, opened 2020-06-16
Token gauge metrics implementation.

mgritter:
Implementation of all three new token count gauges.
Still need to test on enterprise branch to make sure non-root namespaces are working correctly. And add a unit test.  Verified locally that the new code is providing the right output.


mgritter:
OK, broke a bunch of unit tests because emitMetrics is called with an uninitialized set of options.

vault/core.go
```diff
@@ -2093,6 +2093,49 @@ func (c *Core) emitMetrics(stopCh chan struct{}) {
 	writeTimer := time.Tick(c.counters.syncInterval)
 	identityCountTimer := time.Tick(time.Minute * 10)
 
+	// The gauge collection processes are started and stopped here
+	// because there's more than one TokenManager created during startup,
+	// but we only want one set of gauges.
+	if c.MetricSink().GaugeInterval != time.Duration(0) {
+		tokenCount1, err := c.MetricSink().NewGaugeCollectionProcess(
```
mgritter, line 2100: This is the pattern that needs to be copied for each of the newly introduced gauges.

I tried refactoring but it was something of a mess because of the need to use a different function, on whatever object (tokenStore) is hooked up to the core, at the time we collect the metrics.

vault/core.go
```diff
@@ -2093,6 +2093,49 @@ func (c *Core) emitMetrics(stopCh chan struct{}) {
...
+			[]string{"token", "count"},
+			[]metrics.Label{{"gauge", "token_by_namespace"}},
+			func(ctx context.Context) ([]metricsutil.GaugeLabelValues, error) {
+				if c.tokenStore == nil {
```
briankassouf, line 2104: I believe we should have a read on the stateLock mutex to safely access the the token store variable. Same goes for the below function

vault/core.go
```diff
@@ -2093,6 +2093,49 @@ func (c *Core) emitMetrics(stopCh chan struct{}) {
...
+					return c.tokenStore.gaugeCollector(ctx)
+				}
+			},
+			c.logger,
+		)
+		if err != nil {
+			c.logger.Error("failed to start token count", "error", err)
+		} else {
+			go tokenCount1.Run()
+			defer tokenCount1.Stop()
```
briankassouf, line 2116: Could we use a more descriptive variable name?

vault/core.go
```diff
@@ -2093,6 +2093,49 @@ func (c *Core) emitMetrics(stopCh chan struct{}) {
...
+		}
+
+		tokenCount2, err := c.MetricSink().NewGaugeCollectionProcess(
+			[]string{"token", "count", "by_policy"},
+			[]metrics.Label{{"gauge", "token_by_policy"}},
```
briankassouf, line 2121: Is this adding a label of `gauge=token_by_policy`? Is that redundant given the name of the metric above?

vault/token_store.go
```diff
@@ -3452,6 +3459,152 @@ func (ts *TokenStore) gaugeCollector(ctx context.Context) ([]metricsutil.GaugeLa
...
+	prefixTree := radix.New()
+
+	pathToPrefix := func(path string) string {
+		_, method, ok := prefixTree.LongestPrefix(path)
+		if ok {
+			return method.(string)
+		}
+
+		// This method needs a context that has a namespace
+		// FIXME: does it need to match the namespace in the lease?
```
briankassouf, line 3541: I’m not sure off the top of my head but I believe you may need to use a context with the correct ns

vault/token_store.go
```diff
@@ -3452,6 +3459,152 @@ func (ts *TokenStore) gaugeCollector(ctx context.Context) ([]metricsutil.GaugeLa
...
+		mountEntry := ts.core.router.MatchingMountEntry(rootContext, path)
+		if mountEntry == nil {
+			return "unknown"
+		}
+
+		// FIXME: is there a better way to do this, rather than accessing
+		// the router twice? mountEntry.Path is incomplete.
```
briankassouf, line 3548: You may need to join the ns path with the mount path. Is that what you meant by incomplete?

vault/core.go
mgritter: The label is used for reporting errors and run times; it's attached to a set of "metadata" metrics:
vault.metrics.collection.interval -- current gauge collection interval
vault.metrics.collection -- sample of how long it takes
vault.metrics.collection.error -- error count on collection

vault/token_store.go
mgritter: The specific issue here was that mountEntry.Path doesn't include auth/, but the lease name does.

I run into the same problem in various places; it makes a certain amount of sense because the auth mount table is separate and the mounts can only be under auth/. I should probably make a function that converts the mountEntry back to its user-visible version, but I'm not sure if there is some existing functionality that does the job.

I was a little wary of just appending auth/ to the name, because of namespaces, but I do need to test that this works correctly in that case-- maybe it does not.

mgritter:
> We may want to add in toggles to turn these new metrics on/off in the event it causes too much load or issues with the metrics sinks...

The telemetry gauge_collection_interval setting (introduced in https://github.com/hashicorp/vault/pull/9238) can be set to zero or "none" to disable them all. We went back and forth on the RFC; one version had a more fine-grained filter but all-or-nothing seems OK for the first release, I think.

mgritter:
Oops, I forgot to enable one of the gauges!

Tested on enterprise with two namespaces (and additional debugging output) and the methods here appear to work:
```
2020-06-18T22:07:22.433-0700 [INFO]  core: gauge: key=[token, count] value=2 labels=["{namespace root}"]
2020-06-18T22:07:22.433-0700 [INFO]  core: gauge: key=[token, count] value=6 labels=["{namespace ns1}"]
2020-06-18T22:07:22.497-0700 [WARN]  core.licensing: core: licensing warning: expiration_time="2020-06-18 22:35:22 -0700 PDT" time_left=28m0s
2020-06-18T22:07:30.274-0700 [INFO]  core: gauge: key=[token, count, by_method] value=1 labels=["{namespace root}", "{auth_method userpass}"]
2020-06-18T22:07:30.275-0700 [INFO]  core: gauge: key=[token, count, by_method] value=1 labels=["{namespace root}", "{auth_method token}"]
2020-06-18T22:07:30.275-0700 [INFO]  core: gauge: key=[token, count, by_method] value=5 labels=["{namespace ns1}", "{auth_method userpass}"]
2020-06-18T22:07:30.275-0700 [INFO]  core: gauge: key=[token, count, by_method] value=1 labels=["{namespace ns1}", "{auth_method token}"]
2020-06-18T22:07:50.942-0700 [INFO]  core: gauge: key=[token, count, by_policy] value=2 labels=["{namespace root}", "{policy admins}"]
2020-06-18T22:07:50.942-0700 [INFO]  core: gauge: key=[token, count, by_policy] value=1 labels=["{namespace root}", "{policy default}"]
2020-06-18T22:07:50.942-0700 [INFO]  core: gauge: key=[token, count, by_policy] value=1 labels=["{namespace ns1}", "{policy default}"]
2020-06-18T22:07:50.943-0700 [INFO]  core: gauge: key=[token, count, by_policy] value=6 labels=["{namespace ns1}", "{policy admins}"]
```

That suggests I can remove one of my FIXMEs, we can look up a lease's mount point even if the context is the root namespace.

vault/token_store.go
mgritter: Verified I don't need to join the namespace (see example output in a comment.)

vault/token_store.go
mgritter: It looks like this works OK with multiple namespaces, will remove the FIXME.

mgritter: Nope, I was wrong. If I create two different auth methods with the same path, but in different namespaces, then they are aliased.  So need to incorporate the namespace into the lookup.

2020-06-18T22:26:57.579-0700 [INFO]  core: gauge: key=[token, count, by_method] value=1 labels=["{namespace root}", "{auth_method userpass}"]
2020-06-18T22:26:57.579-0700 [INFO]  core: gauge: key=[token, count, by_method] value=1 labels=["{namespace ns1}", "{auth_method userpass}"]

vault/token_store.go
```diff
@@ -3439,10 +3440,16 @@ func (ts *TokenStore) gaugeCollector(ctx context.Context) ([]metricsutil.GaugeLa
 		}
 	})
 
+	if err != nil {
+		return []metricsutil.GaugeLabelValues{}, nil
```
ncabatoff, line 3444: If we're deliberately suppressing this error, can we have a comment explaining why?

vault/core.go
briankassouf: Ah okay, thanks for that. I was missing that context.

vault/token_store.go
mgritter: Good catch, thanks. I replaced this to look for the specific error I wanted to suppress (the expiration manager is still in recovery mode) and pass others through.

vault/core.go
mgritter: Fixed.

vault/core.go
mgritter: Fixed in the new code which loops over an array instead.

vault/token_store.go
mgritter: Now uses a context with the namespace we got from the lease ID.

mgritter:
Testing on enterprise, verified that now two different mount types mounted at /auth/myauth/ and /ns1/auth/myauth/ do resolve to different auth types, as expected:

```
2020-06-19T15:40:06.210-0700 [INFO]  core: gauge debug: key=[token, count, by_auth] labels=["{namespace root}", "{auth_method userpass}"] value=1
2020-06-19T15:40:06.210-0700 [INFO]  core: gauge debug: key=[token, count, by_auth] labels=["{namespace root}", "{auth_method token}"] value=2
2020-06-19T15:40:06.211-0700 [INFO]  core: gauge debug: key=[token, count, by_auth] labels=["{namespace ns1}", "{auth_method github}"] value=1
2020-06-19T15:40:06.211-0700 [INFO]  core: gauge debug: key=[token, count, by_auth] labels=["{namespace ns1}", "{auth_method ns_token}"] value=1
```

Maybe we want to report "ns_token" as just token, though?

vault/core_metrics.go
```diff
@@ -2,15 +2,196 @@ package vault
...
+// emitMetrics is used to start all the periodc metrics; all of them should
+// be shut down when stopCh is closed.
+func (c *Core) emitMetrics(stopCh chan struct{}) {
+	// The gauge collection processes are started and stopped here
+	// because there's more than one TokenManager created during startup,
+	// but we only want one set of gauges.
+
+	metricsInit := []struct {
+		MetricName    []string
+		MetadataLabel []metrics.Label
```
briankassouf, line 129: Would it make sense to rename this `MetricErrorLabel`? Or are we using this for more things now?

vault/token_store.go
```diff
@@ -3452,6 +3469,230 @@ func (ts *TokenStore) gaugeCollector(ctx context.Context) ([]metricsutil.GaugeLa
...
+		}
+
+		// Look up the path from the lease within the correct namespace
+		mountEntry := ts.core.router.MatchingMountEntry(ctx, path)
+		if mountEntry == nil {
+			return "unknown"
+		}
+
+		// mountEntry.Path lacks the "auth/" prefix; perhaps we should
+		// refactor router to provide a method taht returns both the matching
```
briankassouf, line 3632:
```suggestion
		// refactor router to provide a method that returns both the matching
```

vault/core_metrics.go
```diff
@@ -2,15 +2,196 @@ package vault
...
+		}
+	}
+}
+
+// These wrappers are responsible for redirecting to the current instance of
+// TokenStore; there is one per method because an additional level of abstraction
+// seems confusing.
+func (c *Core) tokenGaugeCollector(ctx context.Context) ([]metricsutil.GaugeLabelValues, error) {
+	// stateLock or authLock protects the tokenStore pointer
+	c.stateLock.Lock()
```
briankassouf, line 81: I think a read lock will suffice here and in the below functions

vault/token_store.go
```diff
@@ -3452,6 +3469,230 @@ func (ts *TokenStore) gaugeCollector(ctx context.Context) ([]metricsutil.GaugeLa
...
+		ctx := namespace.ContextWithNamespace(rootContext, ns)
+
+		key := ns.Path + path
+		_, method, ok := prefixTree.LongestPrefix(key)
+		if ok {
+			return method.(string)
+		}
+
+		// Look up the path from the lease within the correct namespace
+		mountEntry := ts.core.router.MatchingMountEntry(ctx, path)
```
briankassouf, line 3626: We may want to read lock core's stateLock before accessing the router. Otherwise i think we could cause some races here (and when we access the router below)

vault/core_metrics.go
mgritter: Yes, thank you.

vault/core_metrics.go
mgritter: It is used for the three metrics: error cont, time to do the collection, and the cardinality before filtering is applied.

vault/token_store.go
mgritter: OK; is there a different entry point I could be using that does that already?

mgritter: Or I'm wondering if metricsMutex will do, but I'm not sure exactly what it protects?

mgritter: Fixed, used stateLock for now.

vault/core_metrics.go
mgritter: fixed.

vault/core_metrics.go
briankassouf: Ok, disregard then!

helper/metricsutil/bucket.go
```diff
@@ -20,7 +20,7 @@ var bucketBoundaries = []struct {
 	{30 * 24 * time.Hour, "30d"},
 }
 
-const overflowBucket = "+Inf"
+const OverflowBucket = "+Inf"
```
briankassouf, line 23: Instead of `+Inf` do we want do `+30d`?

mgritter: While writing the RFC, I thought about ">30d" or "30d+". I think the main problem I have is that they are visually very similar to the bare "30d", so in a visualization it would be unclear which bucket was which.

Issue #9213, opened 2020-06-12
Make TLS ConnectionState available to external plugins


tep:
_n.b. I came close to doing the PR to make this happen then quickly concluded I should write this feature request first, to ensure the PR wouldn't be summarily rejected for one reason or another. Otherwise, I'd be perfectly happy to do the leg work._
 
**Problem:**
The `*tls.ConnectionState` from the origin request is not available in the `*logical.Request` passed to external plugin callbacks. This greatly hinders development of plugins intent on examining incoming client certificates. 

**Solution:**
The original request's `*tls.ConnectionState` should be marshalled through to external plugins as part of the `pb.Connection` protobuf message that's included when making gRPC calls to the plugin. 

As an optimization, inclusion could be optional and the ConnState only added for plugins (or perhaps just Paths or Operations) that request it.

**Alternatives:**
In my current use-case I'm developing a plugin that needs to examine several x.509v3 custom extensions attached to a client's certificate and, as a work around, I've extracted those values on the client side and sent them to the plugin as field values -- but, this is an ugly hack and introduces several limitation.

ncabatoff:
Hi @tep, 

We discussed this internally and would be fine with this feature.  If you want to submit a PR that would be great!

tep:
Thanks @ncabatoff  -- as luck would have it, I submitted #9240 a few days after I wrote this issue (and have been running w/ that patch in my dev cluster for the past few weeks).

tep:
Hey @ncabatoff,  The original PR for this was old and crufty so it's been closed and replaced with #12581.  Let me know soonest if there's anything else you need from me.

PR #9240, opened 2020-06-16
Update plugin proto to send tls.ConnectionState

tep:
Fixes: #9213

In order for external plugins to examine the client certificate(s)
presented during the original request into Vault, we need to marshal
that request's tls.ConnectionState across the gRPC boundary from the
Vault server to the external plugin process.

To that end, this change updates the protocol buffer (and associated
translation logic) used for communicating with external plugins to
include the logical.Connection's "ConnState" value -- thus making it
available to the logical.Request passed to callbacks defined by
external plugins.


tep:
Closing in lieu of #12581

PR #9242, opened 2020-06-16
UI: Disallow kv2 with too large 'max versions' value

chelshaw:
This PR addresses a bug in the UI that causes an error when saving the kv2 secret with too large of a  `max versions` value. Now, when the max versions value is over a threshold, we throw an error indicating that the max versions value is too large. 

Before:
kv2-versions-before

After:
kv2-versions-after


ui/app/components/secret-edit.js
```diff
@@ -294,16 +294,22 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
...
       let model = this.modelForData;
+      let secret = this.model;
       // prevent from submitting if there's no key
-      // maybe do something fancier later
       if (type === 'create' && isBlank(model.path || model.id)) {
+        this.flashMessages.danger('Please provide a path for the secret');
+        return;
+      }
+      const maxVersions = secret.get('maxVersions');
+      if (maxVersions > 9999999999999999) {
```
noelledaley, line 311: you might want to drop a comment in here to explain the magic number, but otherwise this seems like a reasonable solution to me 👍

PR #9250, opened 2020-06-17
Added gauges to count KV secrets.

mgritter:
This still needs to be hooked up to core.emitMetrics in the same way as https://github.com/hashicorp/vault/pull/9239.  But I want to move emitFunctions into this new file first.


vault/core_metrics.go
```diff
@@ -0,0 +1,112 @@
...
+				Operation: logical.ListOperation,
+				Path:      currentDirectory,
+			}
+			resp, err := c.router.Route(ctx, listRequest)
+			if err != nil {
+				c.logger.Error("failed to perform internal KV list", "mount_point", m.MountPoint, "error", err)
+				// TODO: mark just one gauge as failed?
+				return []metricsutil.GaugeLabelValues{}, err
+			}
+			rawKeys, ok := resp.Data["keys"]
```
ncabatoff, line 90: Probably best not to assume we have a non-nil response.

vault/core_metrics.go
```diff
@@ -0,0 +1,112 @@
...
+			currentDirectory := subdirectories[0]
+			subdirectories = subdirectories[1:]
+
+			listRequest := &logical.Request{
+				Operation: logical.ListOperation,
+				Path:      currentDirectory,
+			}
+			resp, err := c.router.Route(ctx, listRequest)
+			if err != nil {
+				c.logger.Error("failed to perform internal KV list", "mount_point", m.MountPoint, "error", err)
```
ncabatoff, line 86: I've said this before in another patch and I don't think you were persuaded, but: when it comes to errors encountered while trying to generate metrics, I think it's best to report those errors using other metrics rather than logs.  That way it's easy for metrics consumers to be aware that there may be issues with the metrics they're seeing.

vault/core_metrics.go
```diff
@@ -0,0 +1,112 @@
...
+				// TODO: mark just one gauge as failed?
```
ncabatoff, line 87: This could happen in a healthy system if something gets unmounted between when we list the mounts and run this code, right?  In which case it's probably nicer for us to continue with the other mounts.

vault/core_metrics_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/hashicorp/vault/helper/namespace"
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+func TestCoreMetrics_KvSecretGauge(t *testing.T) {
+	core, _, root := TestCoreUnsealed(t)
+
+	// I think I can't test with the real kv-v2 because it's
+	// a plugin. But we can fake it by using "metadata" as
```
ncabatoff, line 14: It's a builtin plugin so you should be able to test with kvv2 - plenty of existing tests do.

mgritter: What do I need to do to make that happen? Can you give me a pointer?

One example I found was in external_tests which explicitly linked the plugin package, so I guess that would work. The others I found were in the command package which starts a test server, which seems inappropriate here.

vault/core_metrics.go
mgritter: I did add a metric that covers reported errors:
https://github.com/hashicorp/vault/blob/4bb08b7a7ef41b61d41a09a242bad8e0df018e56/helper/metricsutil/gauge_process.go#L182
```
		p.sink.IncrCounterWithLabels([]string{"metrics", "collection", "error"},
			1,
			p.labels)
```

vault/core_metrics.go
mgritter: That's a good point, I wasn't clear on whether it would return an error.  (One of my frustrations with the list operation is how often it just returns nothing when it *should* return an error.)

We probably need to report the metrics.collection.error count within this code to get coverage that the counts might be bad, if we don't report a top-level error. I prefer having the text message for troubleshooting, though, and that seems inappropriate to put in a metric label.

It would be nice to understand expected errors like "that mount doesn't exist any more", I will see if I can make the error handling a little smarter.

vault/core_metrics.go
mgritter: Added check.

vault/core_metrics.go
mgritter: I change the error handling so that it increments this counter for each KV mount that fails, rather than reporting a single error at the top level.

vault/core_metrics.go
mgritter: Improved with the new handling of errors (by increasing the error counter per mount that fails.)

vault/core_metrics_test.go
ncabatoff: Shouldn't need more than

```
	if err := testClient.Sys().Mount("kv", &api.MountInput{
		Type: "kv",
		Options: map[string]string{
			"version": "2",
		},
	}); err != nil {
		t.Fatal(err)
	}
```

ncabatoff: Or look at `testCoreAddSecretMount`

mgritter: OK; I admit I don't understand why going directly to core.mount would behave any differently than going through HandleRequest, but I can give it a try.

I thought the problem is that the test configuration uses
        logicalBackends["kv"] = LeasedPassthroughBackendFactory
so we always get v1.

mgritter: Yeah, changing how I mounted the secret engine isn't the problem; it still goes to PassthroughBackend. I think I need to replace core's configuration with something like:

```
	coreConfig := &vault.CoreConfig{
		LogicalBackends: map[string]logical.Factory{
			"kv": logicalKv.Factory,
		},
	}
```

mgritter: I managed to make this work, but using a mechanism it looks like nobody else ever has (for swapping in a different backend factory into TestCore.)

vault/core_metrics.go
```diff
@@ -0,0 +1,143 @@
...
+			if path[len(path)-1] == '/' {
+				subdirectories = append(subdirectories, currentDirectory+path)
+			} else {
+				m.NumSecrets += 1
+			}
+		}
+	}
+}
+
+func (c *Core) kvSecretGaugeCollector(ctx context.Context) ([]metricsutil.GaugeLabelValues, error) {
```
briankassouf, line 113: Can we add an env var here that short circuits this function? I'd like for us to have an escape hatch (that is more granular than turning all the new metrics off) incase this creates too much load.

vault/core_metrics.go
```diff
@@ -0,0 +1,143 @@
...
+		listRequest := &logical.Request{
+			Operation: logical.ListOperation,
+			Path:      currentDirectory,
+		}
+		resp, err := c.router.Route(ctx, listRequest)
+		if err != nil {
+			c.kvCollectionErrorCount()
+			// ErrUnsupportedPath probably means that the mount is not there any more,
+			// don't log those cases.
+			if !strings.Contains(err.Error(), logical.ErrUnsupportedPath.Error()) {
```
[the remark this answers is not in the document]
mgritter: The error count is incremented on line 79, before this check; it only suppresses a log message.

vault/core_metrics.go
mgritter: I will do that where the process is initialized.

vault/core_metrics.go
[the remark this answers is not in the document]
mgritter: Oh, sorry, I got what you were saying backwards.  I think the counter should be incremented any time the measurement is suspect, even if it's an error we "expect" will happen occasionally.  It's a signal that the data is potentially bad.

vault/core_metrics.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (c *Core) findKvMounts() []*kvMount {
+	mounts := make([]*kvMount, 0)
+
+	c.mountsLock.RLock()
+	defer c.mountsLock.RUnlock()
+
+	for _, entry := range c.mounts.Entries {
+		if entry.Type == "kv" {
```
alexanderbez, line 29: nit: we should consider having a `const` for these kinds of checks.

PR #9252, opened 2020-06-17
Replaced ClusterMetricSink's cluster name with an atomic.Value.

mgritter:
Replaced ClusterMetricSink's cluster name with an atomic.Value.  This should permit go-race tests to pass which seal and unseal the core.

Alternatives considered: 
 * refer back to Core.cluster every time. This would create a circular dependency if we did it directly. 
 * call a function every time.  This would break the circular dependency, but seems cumbersome.
 * put a mutex around the cluster name

The approach here seems like the least heavyweight solution to the potential race.


mgritter:
Another failure, in an odd place, this is @sgmiller's new code:

```
				c.metricSink.SetGaugeWithLabels([]string{"core", "unsealed"}, 1, nil)
```

Does metricSink get modified by seal or unseal?

```
==================
 WARNING: DATA RACE
 Read at 0x00c002e38278 by goroutine 613: 
  github.com/hashicorp/vault/vault.(*Core).emitMetrics()
      /go/src/github.com/hashicorp/vault/vault/core.go:2107 +0x794
 
Previous write at 0x00c002e38278 by goroutine 738:
  [failed to restore the stack]

Goroutine 613 (running) created at:
  github.com/hashicorp/vault/vault.(*Core).postUnseal()
      /go/src/github.com/hashicorp/vault/vault/core.go:1993 +0x574
  github.com/hashicorp/vault/vault.(*Core).unsealInternal()
      /go/src/github.com/hashicorp/vault/vault/core.go:1485 +0xd75
  github.com/hashicorp/vault/vault.(*Core).unseal()
      /go/src/github.com/hashicorp/vault/vault/core.go:1109 +0x9cf
  github.com/hashicorp/vault/vault.TestCoreUnseal()
      /go/src/github.com/hashicorp/vault/vault/core.go:1022 +0x62
  github.com/hashicorp/vault/vault.testCoreUnsealed()
      /go/src/github.com/hashicorp/vault/vault/testing.go:330 +0x1a1
  github.com/hashicorp/vault/vault.TestCoreUnsealed()
      /go/src/github.com/hashicorp/vault/vault/testing.go:307 +0x9f
  github.com/hashicorp/vault/vault.TestRequestHandling_LoginMetric()
       /go/src/github.com/hashicorp/vault/vault/request_handling_test.go:209 +0x5f
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:909 +0x199
 

Goroutine 738 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:960 +0x651
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1202 +0xa6
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:909 +0x199
  testing.runTests()
      /usr/local/go/src/testing/testing.go:1200 +0x521
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:1117 +0x2ff
  main.main()
      _testmain.go:924 +0x223
 
==================
 ```

**Comment by mgritter** [comment]
Ah, that test swaps in a different metricSink after the core is unsealed. Need to think about how to do this in a race-free manner.

File: helper/metricsutil/wrapped_metrics.go [hunk 1]
```diff
@@ -59,17 +63,30 @@ func (m *ClusterMetricSink) MeasureSinceWithLabels(key []string, start time.Time
 func BlackholeSink() *ClusterMetricSink {
 	sink, _ := metrics.New(metrics.DefaultConfig(""),
```
calvn, line 64: Could we do `return NewClusterMetricSink("", &metrics.BlackholeSink{})` over here?

mgritter: Yeah, good point, I added the second factory later and then forgot to refactor this one.   I'll fix that and commit.

mgritter: On second thought, I'll commit now and fix later.

PR #9254, opened 2020-06-18
raft: allow retry join block within ha_storage stanza

calvn:


vishalnayak:
This is looking right. LMK how the testing goes.

calvn:
Bumped across a few issues during testing, going to hold off on merging this one.

PR #9258, opened 2020-06-18
builtin/database: fix creds rotation panic for nil resp

jasonodonnell:
A panic occurs if `resp` is a nil pointer in the event of a password rotation failure. Added a small check if `resp` is nil before trying to extract WAL ID from it.


jasonodonnell:
Similar bug in OpenLDAP secret engine: https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/14

PR #9263, opened 2020-06-19
Add counter for creation of new entity.

mgritter:


vault/identity_store.go
```diff
@@ -568,6 +569,23 @@ func (i *IdentityStore) CreateOrFetchEntity(ctx context.Context, alias *logical.
 		entity.Aliases = []*identity.Alias{
 			newAlias,
 		}
+
+		// Emit a metric for the new entity
+		ns, err := namespaceByID(ctx, entity.NamespaceID, i.core)
```
mgritter: This isn't working as I expected it to on enterprise, I'm getting "unknown"

mgritter: Oh, should be capital-N namespace.

PR #9270, opened 2020-06-19
VLT091 plugin testing framework stepwise

catsby:
This PR adds a new plugin testing framework `stepwise` into the Vault SDK. This represents the core of the framework with environments and tests to follow shortly. This work was reviewed and approved to merge into this feature branch in https://github.com/hashicorp/vault/pull/9166 , and is now being opened to be merged into the main branch. 

The main changes/additions are summarized as:

- Move [`helper/testhelpers/logical/testing.go`][logicaltest] to the Vault SDK package
- Update all plugin tests that use the above referenced `logicaltest` package 
- Adds a new package `stepwise`, a spiritual successor to the `logicaltest` package

Future PRs will introduce:

- An implementation of an `stepwise.Environment`; the first will be a Docker environment
- Multiple plugin tests that utilize the stepwise framework and docker environment

Remaining:

- [x] resolve any merge conflicts
- [ ] README for the `stepwise` package

[logicaltest]: https://github.com/hashicorp/vault/blob/2f1e5b66d45266a090b7f8fb45c20804f310f5da/helper/testhelpers/logical/testing.go


sdk/testing/stepwise/legacy.go
```diff
@@ -220,7 +229,10 @@ func Test(tt TestT, c TestCase) {
 
 	// Create an HTTP API server and client
 	ln, addr := http.TestServer(nil, core)
-	defer ln.Close()
+	defer func() {
+		// intentionally ingore this error
+		_ = ln.Close()
```
calvn, line 234: Calling `defer ln.Close()` should also ignore the error.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,332 @@
...
+// user defined steps against.
+package stepwise
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	log "github.com/hashicorp/go-hclog"
+
```
calvn, line 13:
```suggestion
```

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,332 @@
...
+	"github.com/hashicorp/vault/api"
+	"github.com/hashicorp/vault/sdk/helper/consts"
+	"github.com/hashicorp/vault/sdk/helper/logging"
+)
+
+// Operation defines operations each step could preform. These are
```
calvn, line 19:
```suggestion
// Operation defines operations each step could perform. These are
```

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,332 @@
...
+		// Fatal will exit
+		return
+	}
+
+	logger := logging.NewVaultLogger(log.Trace)
+
+	if err := c.Environment.Setup(); err != nil {
+		driverErr := fmt.Errorf("error setting up driver: %w", err)
+		if !c.SkipTeardown {
+			if err := c.Environment.Teardown(); err != nil {
```
calvn, line 195: Is calling `Teardown` necessary here if the `Setup` failed in the first place?

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,332 @@
...
+			if err := rootClient.Sys().Revoke(secret.LeaseID); err != nil {
+				tt.Error(fmt.Errorf("error revoking lease: %w", err))
+				failedRevokes = append(failedRevokes, secret)
+				continue
+			}
+		}
+
+		// If we have any failed revokes, log it.
+		if len(failedRevokes) > 0 {
+			for _, s := range failedRevokes {
```
calvn, line 247: If the steps below consist of testing creation and then revocation, this defer call will likely produce a false positive when trying to revoke an already-revoked secret (which was done as part of one of the steps).

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+		}
+	}()
+
+	stepCount := len(c.Steps)
+	for i, step := range c.Steps {
+		// range is zero based, so add 1 for a human friendly output of steps.
+		// "index" here is only used for logging / output, and not to reference the
+		// step in the slice of steps.
+		index := i + 1
```
calvn, line 261: Nit: you could you `i + 1` directly in here and below.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,332 @@
...
+
+func makeRequest(tt TestT, env Environment, step Step) (*api.Secret, error) {
+	tt.Helper()
+	client, err := env.Client()
+	if err != nil {
+		return nil, err
+	}
+
+	if step.Unauthenticated {
+		client.ClearToken()
```
calvn, line 300: Since `ClearToken` is updates the client's token value and the client is cached, we might want to defer adding back the client's original token after `makeRequest`.

sdk/testing/stepwise/stepwise_test.go
```diff
@@ -0,0 +1,285 @@
...
+		ExpectedRequestID string
+		ExpectErr         bool
+	}
+	testRequests := map[string]testRequest{
+		"list": {
+			Operation:         ListOperation,
+			Path:              "keys",
+			ExpectedRequestID: "list-request",
+		},
+		"read": {
```
calvn, line 142: Looks like `mockEnvironment` has a revoke endpoint but we're not testing it here. Might be also good to test create and update operations, as well as auth'ed/unauth'ed requests.

sdk/go.mod
```diff
@@ -2,34 +2,44 @@ module github.com/hashicorp/vault/sdk
 
 go 1.13
 
+replace github.com/hashicorp/vault/api => ../api
+
+replace github.com/hashicorp/vault => ../
```
calvn, line 7: Won't this create a dep cycle, since vault imports the sdk, now the sdk imports vault?

catsby: Oddly enough this was the only way I could get the tests to run in `stepwise_test.go`. Otherwise I get errors about a `helper/random` package. I'll continue to poke around and try to sort things out.

sdk/testing/stepwise/helpers.go
```diff
@@ -0,0 +1,52 @@
...
+		return "", "", "", err
+	}
+
+	// calculate sha256
+	f, err := os.Open(binPath)
+	if err != nil {
+		return "", "", "", err
+	}
+	defer func() {
+		_ = f.Close()
```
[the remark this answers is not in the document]
catsby: I was using a linter locally that was complaining when implicitly ignoring those errors, but I reverted to just `defer f.Close()` in 14557d1bf as I prefer that style anyway

sdk/testing/stepwise/legacy.go
catsby: Changed in 14557d1bf

sdk/testing/stepwise/stepwise.go
catsby: Likely not; I changed that in 483bc3080 to simply call `Fatal` with the error.

sdk/testing/stepwise/stepwise.go
catsby: Using a variable like `index` gave me a better opportunity to document in code why the `i + 1` what happening at all, so I'd like to keep this unless you feel strongly that it should change

sdk/testing/stepwise/stepwise.go
catsby: Great idea, I added that in 09b2307bd

sdk/testing/stepwise/stepwise_test.go
catsby: Unless you object I'd like to defer those additions until after the merge, and the first driver is added which will bring some plugin tests that use those endpoints as well.

sdk/testing/stepwise/stepwise.go
catsby: Great idea, I added that in 09b2307bd

sdk/testing/stepwise/stepwise.go
catsby: I'm not sure, I believe (and my local testing with a dev cluster confirmed) that revoking an already revoked lease simply returns a `204 No Content` response, which is the same response for successfully revoking a lease. Does that sound right?

sdk/testing/stepwise/helpers.go
```diff
@@ -0,0 +1,52 @@
...
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+)
+
+const pluginPrefix = "vault-plugin-"
+
+// CompilePlugin is a helper method to compile a source plugin
+// TODO refactor compile plugin input and output to be types
```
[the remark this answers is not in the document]
catsby: I'm tracking this internally as I'm not 100% sure on the inputs/outputs as is, so I'm deferring this for a future PR.

sdk/go.mod
catsby: Resolved this by moving the `logicaltest` package back to `/helpers` , so the SDK no longer depends on Vault

sdk/testing/stepwise/stepwise.go
calvn: Ah right, this should be fine then.

sdk/testing/stepwise/stepwise.go
calvn: Yep, I was suggesting keeping the `range is zero based, so add 1 for a human friendly output of steps` comment and removing the need for the index since that should be clear enough, but not a big deal if it stays.

sdk/testing/stepwise/stepwise_test.go
calvn: Since `mockEnvironment` is not dependent on external plugins or drivers, having it here now can ensure that `makeRequest` will work properly when plugins tests do end up calling it. I'd lean towards adding those cases for completeness.

sdk/go.mod
```diff
@@ -3,33 +3,47 @@ module github.com/hashicorp/vault/sdk
 go 1.13
 
 require (
-	github.com/armon/go-metrics v0.3.0
-	github.com/armon/go-radix v0.0.0-20180808171621-7fddfc383310
-	github.com/aws/aws-sdk-go v1.25.37
+	github.com/armon/go-metrics v0.3.3
```
calvn, line 6: Are the dep updates in here required by your changes? If not, it might be good to revert them, and address vendor updates separately.

sdk/testing/stepwise/helpers.go
```diff
@@ -0,0 +1,51 @@
...
```
calvn, line 17: Since all non-error return values are strings, might be good to add a comment around them so consumers don't need to look into the source to figure this out.

sdk/testing/stepwise/stepwise_test.go
```diff
@@ -0,0 +1,336 @@
...
+	t.SkipCalled = true
+	t.SkipArgs = args
+	t.f = true
+}
+
+func (t *mockT) failed() bool {
+	return t.f
+}
+
+func (t *mockT) failMessage() string {
```
calvn, line 324: This (and `failed()`) no longer seem to be referenced elsewhere.

sdk/testing/stepwise/stepwise.go
catsby: I changed to `i + 1` in 073737336

sdk/go.mod
alexanderbez: This could be due to a simple `go mod tidy`. If so, we should encourage that, right?

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,340 @@
...
+
+	// Steps are the set of operations that are run for this test case. During
+	// execution each step will be logged to output with a 1-based index as it is
+	// ran, with the first step logged as step '1' and not step '0'.
+	Steps []Step
+
+	// SkipTeardown allows the Environment TeardownFunc to be skipped, leaving any
+	// infrastructure created after the test exists. Depending on the Environment
+	// used this could incur costs the user is responsible for.
+	SkipTeardown bool
```
alexanderbez, line 166: Are there situations where `SkipTeardown:true` and `Environment.Teardown ` is not a no-op. If not, then a client can just have `Teardown` perform a no-op instead of requiring an additional bool field.

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,340 @@
...
+		client, err := rootClient.Clone()
+		if err != nil {
+			tt.Fatal(err)
+		}
+
+		// TODO: support creating tokens with policies listed in each Step
+		client.SetToken(rootToken)
+
+		resp, respErr := makeRequest(tt, c.Environment, step)
+
```
alexanderbez, line 270:
```suggestion
```

sdk/testing/stepwise/stepwise.go
```diff
@@ -0,0 +1,340 @@
...
+		if resp != nil {
+			responses = append(responses, resp)
+		}
+
+		// run the associated AssertionFunc, if any. If an error was expected it is
```
alexanderbez, line 275:
```suggestion
		// Run the associated AssertionFunc, if any. If an error was expected it is
```

sdk/testing/stepwise/stepwise_test.go
catsby: I've updated the tests to do basic request type counting (revokes, reads, et. al.) in 8d5e1e65f , please take a look when you can!

sdk/go.mod
catsby: Indeed I believe this is simply from `go mod tidy`. I'm inclined to leave as-is unless you object @calvn

sdk/testing/stepwise/stepwise_test.go
catsby: Those were copied over from the legacy framework, thanks for catching that. I removed them in 1bb36b0ab

sdk/testing/stepwise/stepwise.go
catsby: Environments are responsible for knowing how to destroy and/or clean up any infrastructure they created during testing. The purpose of `SkipTeardown` is for developers to preserve the environment for inspection after a test runs, for whatever debugging reasons they may have. I've update the code comment above in 4951eabb5 to mention this, does that clarify things here? Or perhaps I'm not following your suggestion/question correctly?

alexanderbez: Cool, makes complete sense. Thanks!

sdk/go.mod
calvn: Makes sense, let's leave as-is.

sdk/testing/stepwise/stepwise_test.go
calvn: Looks great, thanks!

catsby:
Test failures are unrelated so I'm going to merge now. Thanks @calvn, @alexanderbez , and @pcman312 for the reviews 😄

PR #9271, opened 2020-06-19
Update TTL picker on add replication secondary

chelshaw:
This change updates the TTL picker to the new version to match most updated designs. The component also allows the default value to be more obvious.

**BEFORE**
ttl-secondary-old

**AFTER (default)**
replication-secondary-ttl-default

**AFTER (enabled/custom)**
replication-secondary-ttl-enabled


ui/lib/replication/addon/templates/mode/secondaries/add.hbs
```diff
@@ -18,11 +18,14 @@
...
-    <p class="help has-text-grey">
-      This is the Time To Live for the generated secondary token. After this period, the generated token will no longer be valid.
-    </p>
+    <TtlPicker2
+      @initialValue="30m"
+      @label="Time to Live (TTL) for generated secondary token"
+      @helperTextDisabled="If not set, the default value (30 minutes) will be used"
+      @helperTextEnabled="After this period, the generated token will no longer be valid."
+      @onChange={{action "updateTtl"}}
+      @changeOnInit={{true}}
```
chelshaw, line 27: Added this param because I found that if you make a secondary with a custom TTL, and then immediately make another secondary without touching the TTL picker, it uses the previous value instead of the actual default (which is passed here as `initialValue`

ui/lib/core/addon/components/ttl-picker2.js
```diff
@@ -100,6 +107,7 @@ export default Component.extend({
       seconds,
       timeString: time + unit,
     };
+    console.log({ ttl });
```
Monkeychip, line 110: did you want to keep this console in there?

ui/tests/integration/components/ttl-picker2-test.js
```diff
@@ -164,4 +165,21 @@ module('Integration | Component | ttl-picker2', function(hooks) {
...
+        @onChange={{onChange}}
+        @initialValue="100m"
+        @initialEnabled="true"
+        @changeOnInit={{true}}
+      />
+    `);
+    assert.ok(changeSpy.calledOnce, 'it calls the passed onChange after render');
+    // assert.dom('[data-test-ttl-value]').hasValue('6000', 'time value is initialValue as seconds');
+    // assert.dom('[data-test-ttl-unit]').exists('Unit is shown on mount');
+  });
```
Monkeychip, line 184: did you want to keep some of those tests commented out?

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
```diff
@@ -18,11 +18,14 @@
     </p>
   </div>
   <div class="field">
-    {{!-- TODO fix so it defaults to 30s like in other places or replace with new TTL picker --}}
-    {{ttl-picker onChange=(action (mut ttl)) initialValue="30m" class="is-marginless"}}
-    <p class="help has-text-grey">
-      This is the Time To Live for the generated secondary token. After this period, the generated token will no longer be valid.
-    </p>
+    <TtlPicker2
+      @initialValue="30m"
```
noelledaley, line 22: image

image

Judging from the code here I'd expect to see the initial value as 30 minutes, but for some reason I'm seeing 1800 seconds in the UI. I know these values are equal, but everywhere else in the UI we default to "minutes" as the unit. If possible it'd be great to fix this so it's consistent.

ui/lib/core/addon/components/ttl-picker2.js
chelshaw: Nope! 😄

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
chelshaw: This is expected behavior for the current design, but not ideal so I went ahead and updated it (with test coverage!) so that the unit matches what is passed in on the value

chelshaw: I updated the TTL picker UI to match the unit passed in, but the unit being passed to the API is being standardized to seconds for a couple reasons:

1. Some of the APIs expect units, and others don't (defaulting to seconds), so the new TTL picker was designed to account for either of those
2. None of the APIs accept 'days' as a unit, but we provide that as a dropdown option, so always converting to seconds when we send to the API ensures that we are consistent at that level.

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
BOYTOYFORYOUXXX: 💨🦁🐑

ui/lib/replication/addon/templates/mode/secondaries/add.hbs
noelledaley: awesome, thank you for tracking that down! shoot, i found another bug:

1. toggle the TTL picker
2. enter a value (3 minutes)
3. toggle the TTL picker off
4. the value from step 2 still gets sent to the API 🙈

Issue #9257, opened 2020-06-18
Plugin docs unclear that the plugin also needs CAP_IPC_LOCK


onetwopunch:
**Describe the bug**
When writing a custom plugin, due to the way Vault communicates over RPC, if you want to enable mlock and don't want to run Vault as root, you have to also give the custom plugins `CAP_IPC_LOCK` or you'll face an incredibly unhelpful error where Vault can't communicate with the plugin over RPC.

More context at: https://github.com/hashicorp/go-plugin/issues/149

**To Reproduce**
1. Create a custom Vault plugin and place the executable in the directory specified by `plugin_directory`
2. Give the vault executable the IPC LOCK capability per the docs: `/sbin/setcap cap_ipc_lock=+ep /usr/local/bin/vault`
3. Run `vault server` as a non-privileged user
4. Assuming the plugin compiles and can run locally, you should be able to add it to the catalog just fine with: 
```
vault write sys/plugins/catalog/auth/$CATALOG_NAME} sha_256=$SHA256SUM command=$PLUGIN_NAME
```
5. The error should appear when you first try to communicate with the plugin. In my case it was an auth plugin so I tried to configure it with:
```
vault auth enable -path=my_path -listing-visibility=unauth $CATALOG_NAME
```
6. The error was:
```
Unrecognized remote plugin message:

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.
```

Since the plugin worked just fine when I ran it in `-dev` mode as well as when running as root, I knew it was a permissions issue so I tried disabling mlock in the Vault config and running again and it worked fine. Then to fix the issue I simply ran:

```
/sbin/setcap cap_ipc_lock=+ep /etc/vault.d/plugins/my-vault-plugin
```

Now everything works perfectly. This should be documented as part of the Vault plugin docs.

**Expected behavior**
I expected that the Vault executable was the only thing that needed the IPC_LOCK capability, since the docs didn't mention anything about that for plugin creation and there were no samples of how to run Vault in production with a custom plugin.

**Environment:**
* Vault Server Version (retrieve with `vault status`): `1.4.2`
* Vault CLI Version (retrieve with `vault version`): `1.4.2`
* Server Operating System/Architecture: `Debian`

PR #9275, opened 2020-06-20
docs: clarify when plugin executables must also be given mlock syscall ability

austingebauer:
Closes https://github.com/hashicorp/vault/issues/9257

This PR adds a note to the [internals/plugins](https://www.vaultproject.io/docs/internals/plugins) documentation to clarify when plugin executables must also be given the `mlock` syscall ability. It's not obvious that this is needed from the plugin documentation.

There currently is documentation stating that this is needed, but it's somewhat difficult to find in [configuration#disable_mlock](https://www.vaultproject.io/docs/configuration#disable_mlock). See:
```
Note: Since each plugin runs as a separate process, you need to do the same for each plugin in your plugins directory.
```

I plan to update the [ops-deployment-guide](https://learn.hashicorp.com/vault/operations/ops-deployment-guide#step-2-install-vault) to mention this as well.


austingebauer:
Good idea! Done in 54a98d9. Thanks 👍

PR #9282, opened 2020-06-22
Set focus on the #modal-card-{{modalId}} when a modal is active.

Monkeychip:
This PR sets the browser's focus on a new ID called #modal-card-{{modalId}} that is on `modal-card` inside the `<Modal /> component. Setting focus allows a user to use their keyboard to navigate through the modal, and do things like hit enter when it's on the Copy & Close button, or close the modal by hitting enter when they're on the Close button. 

Before this PR, the focus was set on whatever it was before the modal launched, which was generally the button they clicked on to launch the modal (i.e. the "Reindex button, etc).  This _could_ potentially lead to confusing behavior when the user hits tab + enter thinking they're on the modal and instead they do some action in the background.

**Note**: The escape button still doesn't do anything.  Ideally, it would close the modal, but we're just setting the focus, not handling keyboard events. 

See functionality in this GIF.  I tested all instances of `<Modal />` and it works as expected.
focus

Because we have several cases (like the manage pages) where we have multiple models loaded at once, I had to wrap each instance of the modal in a conditional, checking to see whether or not it was active.  This is not ideal.  However, I use the modal hook `[didInsertElement](https://guides.emberjs.com/v3.4.0/components/the-component-lifecycle/#toc_integrating-with-third-party-libraries-with-didinsertelement)` to set the focus.  `didInsertElement`, waits for the dom to completely load, which is crucial because no other hook does this, and we need to wait for the dom to load so we can set focus on the modal-card.  If you don't use this hook,  you hit an error saying it can't set focus to null; it's null because the dom isn't done loading and `#modal-card-{{modalId}}` doesn't exist yet.

The downside to `didInsertElement`, is that it only happens once, on load.  It doesn't happen if a component receives new properties, etc.  There are some discussions about this problem on [StackOverflow](https://stackoverflow.com/questions/9510634/is-there-an-after-view-change-hook-much-like-didinsertelement) (the solutions are not great). 

I checked every ember hook and none of them do what we want, which is to wait for all the elements and check on rerender.  Thus, to solve this, I wrap each instance of the modal in a conditional.  _Now_ the DOM only shows one instance of the modal, which solves some other problems we have encountered in the past with testing.  However, the strange part, as Chelsea pointed out, is that you have `isActive` or `isModalActive` being used around the modal and passed through the modal. 

In addition to this solution, I tried using the `ember-focus-trap` library, but it too has issues.  Most notably it breaks the `CopyButton` component.  We investigated why this is, and the fix has been merged only recently - [see PR here](https://github.com/jkusa/ember-cli-clipboard/pull/206) - and is not available to us on our version.  We updated the version and it still broke everything.  We didn't go too much farther down this rabbit hole, but we may consider trying to go back this route after we update Ember.  

We ended up ditching the Ember-focus-trap route because one we couldn't get it to work without breaking the `CopyButton` component.  Second, we ultimately are just trying to set focus on an element, and shouldn't really need a library to do that.  Well at least in theory...

I'm open to any other suggestions, or if folks are OK with this solution, which works as expected, I'll move forward and start amending tests accordingly. 

Todo (if get OK on solution)
- [ ] tests all browsers
- [ ] amend and add tests


ui/app/templates/components/shamir-modal-flow.hbs
```diff
@@ -1,194 +1,197 @@
...
+          {{#if otp}}
+          There are three values below. The encoded operation token and the OTP are used in the <span class="has-text-semibold">DR operation token command</span>, which you will need to copy and paste into your CLI in order to generate the operation token. Save your encoded operation token and OTP.
+          {{else}}
+          The encoded operation token and the OTP are used in the <span class="has-text-semibold">DR operation token command</span>, which you will need to copy and paste into your CLI in order to generate the operation token. Save your encoded operation token and OTP.
+          {{/if}}
+        </p>
+        <div class="has-bottom-margin-m">
           <h4 class="title is-7 has-bottom-padding-m is-fullwidth">
-            DR operation token command
+            Encoded operation token
```
Monkeychip, line 21: I honestly don't know what that's showing as different.  I double-checked and I only added a conditional.  I'll investigate further if concept is approved.

ui/app/templates/components/shamir-modal-flow.hbs
```diff
@@ -1,194 +1,197 @@
...
+      <button
+        class="button is-secondary"
+        onclick={{action 'onCancelClose'}}
+        data-test-shamir-modal-cancel-button
+      >
+        {{if encoded_token 'Close' 'Cancel'}}
+      </button>
+    </footer>
+  </Modal>
+{{/if}}
```
Monkeychip, line 197: I'll fix all of these missing extra bottom space if concept is approved.

PR #9288, opened 2020-06-23
Update kubernetes.mdx

Eadinator:
The documentation has an example with the application using the vault-auth service account (the app would send vault-auth's token to vault), but also vault-auth is specified in the ClusterRoleBinding which is the account vault uses against the token review API.

This could be confusing as they are separate accounts with separate purposes but they have been named the same here. The proposed change distinguishes between the two.


aphorise:
Request does seem reasonable and a different appellation / naming does seem to be in order. @tomhjp @jasonodonnell your inputs and opinions welcome.

tomhjp (approved):
Thanks, this is a good point to differentiate 👍

PR #9326, opened 2020-06-26
request.connection.remote_addr only has IP

rberlind:
The request.connection.remote_addr property exposed to Sentinel only has an IP.
It does not include a port.
I tested this in a policy with `print("remote address:", request.connection.remote_addr)` and got back 150.10.0.26.

### DIFF
--- a/website/pages/docs/enterprise/sentinel/properties.mdx
+++ b/website/pages/docs/enterprise/sentinel/properties.mdx
@@ -30,7 +30,7 @@ The following properties are available in the `request` namespace.
 
 | Name                     | Type                  | Description                                                                                 |
 | :----------------------- | :-------------------- | :------------------------------------------------------------------------------------------ |
-| `connection.remote_addr` | `string`              | TCP/IP source IP of the client                                                    |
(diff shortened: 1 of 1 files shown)


rberlind:
Thanks @jescalan .  I did not really think you were the right person, but GitHub suggested you and I did not want to argue with it. :)
Roger

website/pages/docs/enterprise/sentinel/properties.mdx
```diff
@@ -30,7 +30,7 @@ The following properties are available in the `request` namespace.
 
 | Name                     | Type                  | Description                                                                                 |
 | :----------------------- | :-------------------- | :------------------------------------------------------------------------------------------ |
-| `connection.remote_addr` | `string`              | TCP/IP source address/port of the client                                                    |
+| `connection.remote_addr` | `string`              | TCP/IP source IP of the client                                                    |
```
calvn, line 33: "TCP/IP source IP" sounds a bit redundant, we should keep the address terminology here.

```suggestion
| `connection.remote_addr` | `string`              | TCP/IP source address of the client                                                    |
```

PR #9332, opened 2020-06-26
Add Algolia indexing script to CI

pruett:
Adds CI job to push search content to Algolia

**Note**: I did add an `ALGOLIA_API_KEY` environment variable to the `vault-docs` context in CCI


website/.env
```diff
@@ -0,0 +1,3 @@
+NEXT_PUBLIC_ALGOLIA_APP_ID=YY0FFNI7MF
+NEXT_PUBLIC_ALGOLIA_INDEX=poc_VAULT
+NEXT_PUBLIC_ALGOLIA_SEARCH_ONLY_API_KEY=5037da4824714676226913c65e961ca0
```
samsalisbury, line 3: For any other reviewers, this looks safe, according to https://www.algolia.com/doc/guides/security/api-keys/#search-only-api-key

pruett:
Thanks for the review @samsalisbury . Yeah, with regard to the Node.js script/we've got that implemented on other properties, so no review necessary there, considering the fact that we may tweak going forward as well.

As far as the failing tests...any insight here?

samsalisbury:
> As far as the failing tests...any insight here?

@pruett Only that the failing tests are not related to your change, they are also broken on the head of master (currently https://github.com/hashicorp/vault/commit/b3f0e9badfaac44ef4300acf13144abb2cec25d3), so you can merge this now. I'm trying to get us to a point where master does not get broken, but we have a little work to do around GH permissions and removing bad or flaky tests.

pruett:
@samsalisbury I don't have the permissions necessary to do that I don't believe: I get the following message underneath the checks:

```
Required statuses must pass before merging
All required statuses and check runs on this pull request must run successfully to enable automatic merging.
```

Feel free to merge if your permissions allow (I'm assuming they do 👍 )

pruett:
Oh right, I guess I just don't have the admin override checkbox

PR #9334, opened 2020-06-29
sdk: ability to parse days, weeks, and years

greut:
I've got myself in trouble by using `1w` in the _period_ configuration when using the Terraform provider. Only to realize that the `time.Duration` goes up to the hour. I guess there is another internal library that could do that (instead of Prometheus' which is already used but not by the sdk)


greut:
that code got moved to https://github.com/hashicorp/go-secure-stdlib

PR #9341, opened 2020-06-29
monitor: watch for seal state during monitor request

calvn:
This PR partially addresses monitor blocking state read lock release by looking at the seal state. Other blocking scenarios that hold the state lock regardless of seal status will be addressed in the separate PR, likely  https://github.com/hashicorp/vault/pull/9312

### DIFF
--- a/vault/logical_system.go
+++ b/vault/logical_system.go
@@ -2703,7 +2703,6 @@ func (b *SystemBackend) handleMetrics(ctx context.Context, req *logical.Request,
 func (b *SystemBackend) handleMonitor(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
 	ll := data.Get("log_level").(string)
 	w := req.ResponseWriter
-	resp := &logical.Response{}
 
 	if ll == "" {
 		ll = "info"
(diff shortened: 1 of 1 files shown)


vault/logical_system.go
```diff
@@ -2744,22 +2742,32 @@ func (b *SystemBackend) handleMonitor(ctx context.Context, req *logical.Request,
...
+			return nil, nil
 		case l := <-logCh:
-			_, err = fmt.Fprint(w, string(l))
-
-			if err != nil {
-				return resp, err
-			}
+			// If we error here, it means we couldn't write to the response
+			// writer so there's nothing we can use to send over the error.
+			_, _ = fmt.Fprint(w, string(l))
```
briankassouf, line 2770: I think we should probably still return if we see an error though, right?

calvn: Yeah, I guess we can still return the error. I'll re-word the comment above since that error does indeed get swallowed by this:

https://github.com/hashicorp/vault/blob/c68e270863cd5518a3380b73c721fe822592b6b8/http/handler.go#L823-L833

calvn: Updated now.

PR #9348, opened 2020-06-30
Fix wrong err return value in plugin reload status command

sgmiller:


alexanderbez:
Looks like some `TestSystemBackend_Plugin_reload` tests are failing?

Issue #9335, opened 2020-06-29
panic when visiting /ui


greut:
**Describe the bug**

vault panic when visiting `/ui`

**To Reproduce**
1. Run `vault server -dev`
2. Run `curl http://localhost:8200/ui`
3. See error

```
2020-06-29T14:08:24.261+0200 [INFO]  http: panic serving 127.0.0.1:53046: runtime error: slice bounds out of range [4:3]
goroutine 56 [running]:
net/http.(*conn).serve.func1(0xc000ad17c0)
        /usr/lib/go/src/net/http/server.go:1772 +0x139
panic(0x34edb20, 0xc000cd5920)
        /usr/lib/go/src/runtime/panic.go:975 +0x3e3
github.com/hashicorp/vault/http.buildLogicalRequestNoAuth(0xc0000c1800, 0x3f7dd80, 0xc000958380, 0xc0005caf00, 0x5e2a2d8, 0x5dcb800, 0x3f99bc0, 0xc00067a2a0, 0x42fcb9, 0x7f0e75925f98)
        /home/yoan/go/src/github.com/hashicorp/vault/http/logical.go:47 +0x14eb
github.com/hashicorp/vault/http.wrapGenericHandler.func1(0x3f7dd80, 0xc000958380, 0xc0005cad00)
        /home/yoan/go/src/github.com/hashicorp/vault/http/handler.go:307 +0x5d7
net/http.HandlerFunc.ServeHTTP(0xc000b58630, 0x3f7dd80, 0xc000958380, 0xc0005cad00)
        /usr/lib/go/src/net/http/server.go:2012 +0x44
github.com/hashicorp/go-cleanhttp.PrintablePathCheckHandler.func1(0x3f7dd80, 0xc000958380, 0xc0005cad00)
        /home/yoan/go/pkg/mod/github.com/hashicorp/go-cleanhttp@v0.5.1/handlers.go:42 +0xba
net/http.HandlerFunc.ServeHTTP(0xc000b781c0, 0x3f7dd80, 0xc000958380, 0xc0005cad00)
        /usr/lib/go/src/net/http/server.go:2012 +0x44
net/http.serverHandler.ServeHTTP(0xc0009582a0, 0x3f7dd80, 0xc000958380, 0xc0005cad00)
        /usr/lib/go/src/net/http/server.go:2807 +0xa3
net/http.(*conn).serve(0xc000ad17c0, 0x3f99b00, 0xc0006ad400)
[...]

PR #9361, opened 2020-07-01
http: check path before parsing

alexanderbez:
The introduction of [resource quotas](https://github.com/hashicorp/vault/pull/9330) introduced changes to the `wrapGenericHandler` middleware, where subsequent handlers expect a `LogicalResponseWriter`. Before, in `wrapGenericHandler`, we did not call `buildLogicalRequestNoAuth`. Now we do, however, `buildLogicalRequestNoAuth` assumes a `/v1/` path prefix, which is not the case for some routes (e.g. UI).

There are still issues with the changes introduced in [resource quotas](https://github.com/hashicorp/vault/pull/9330), specifically some Raft failures, so I'm not sure this fix is a stopgap or a partial solution.

closes: #9335


http/handler.go
```diff
@@ -276,10 +276,14 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
 		} else {
 			ctx, cancelFunc = context.WithTimeout(ctx, maxRequestDuration)
 		}
+
+		defer cancelFunc()
```
calvn, line 280: I believe not using defer in here was on purpose since this is a hot-path and `defer` adds a bit of overhead (albeit this has been greatly reduced in recent Go releases).

http/handler.go
```diff
@@ -309,6 +311,7 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
 			respondError(w, status, err)
```
calvn, line 311: If we stick with no defer, this is missing a call to `cancelFunc()`

http/handler.go
alexanderbez: I'll rever the `defer`.

vishalnayak:
Closing in favor of https://github.com/hashicorp/vault/pull/9390

PR #9374, opened 2020-07-01
Don't return quota error on revoke.

mgritter:
Changed log messages to be clearer about quota operations.
This should fix enterprise unit test failures.


mgritter:
Replaces https://github.com/hashicorp/vault-enterprise/pull/1306

PR #9379, opened 2020-07-01
Add nil check for quota manager

alexanderbez:


calvn:
We should also nil check in these places:
https://github.com/hashicorp/vault/blob/c68e270863cd5518a3380b73c721fe822592b6b8/vault/auth.go#L342
https://github.com/hashicorp/vault/blob/c68e270863cd5518a3380b73c721fe822592b6b8/vault/mount.go#L667

PR #9387, opened 2020-07-02
Ui/small replication action fixes

Monkeychip:
This PR fixes the following issues:

1. **Fix breadcrumb link.**  If a DR cluster was not enabled, the breadcrumb link on the landing page did not work.  I changed it so now if replication is not enabled the link takes you back to the index page `vault.cluster.replication.index`.  I also changed the breadcrumb label to "Replication" whereas before it was Disaster Recovery.  This label change only happens if the DR is not enabled.

breadcrumb1

2. **Another breadcrumb change**.  Before when you were on the manage page and you clicked the breadcrumb it was trying to redirect you to the manage page.  I confirmed with Design that in the case of a DR Secondary, there should be no breadcrumb because you only have two routes in this state: manage and details, and toggle back and forth between those with a breadcrumb didn't make much sense.  Here's an image of a DR secondary, with no breadcrumb.

image

3. We had two situations where the transition needed some help (no errors just look/feel fixes). The first fix was to add a black nav during the **initialization of a secondary DR cluster.**  To reproduce, add a secondary DR cluster.  Before it showed only a loader.  Now you can see a black nav bar above the loader. (in the gif, I wait for the network request so that the state of the loader is on the screen the longest).

transition1

4. The **second transition issue occurred after you enabled a DR secondary and then demoted that same DR**.  This was fixed by adding a new property to the cluster to capture the `drMode`.  I then saved the initial drMode when the route's model was generated and set it on a property called `drModeInit`.  Because the mode changes before any transition, to capture the change (computed properties did not work here), I compared the saved `drModeInit` property to the `drMode` property in the template.  This situation only occurs when the `drMode` goes from primary to secondary (e.g. during demotion).  Once the transition is complete the model is reset and the loader will go away. 

_*note, not very often the `drMode` will be 'disabled' and not 'primary' during the transition. In this case, the loader will not show.  However, because this is rare, and I can't really seem to understand under which circumstance this happens. I feel better about not trying to capture this particular situation and be ok with the occasional flash of template screen and no loader.  My feelings are that this is a feel/look fix, and if the fix is wrong the user could end up seeing a loader and not getting out of it, and that's a much worse situation than fixing a flash of the template page._

Here is a gif with a debugger so you can see the two properties.
transition2

Here is a gif without the debugger so you can see how it looks.
transition3

5. The last fix was to prevent the super slow modal issues we were seeing after demoting a cluster.  This issue was happening because the `shamir-modal` was not found as it was not in the addon component library.  The fix was to move the `shamir-modal` and `shamir-modal-flow` into the addon components so it could be accessed from the replication engine.


ui/app/templates/vault/cluster/replication-dr-promote/index.hbs
```diff
@@ -3,6 +3,7 @@
     <ReplicationPage @model={{model}} as |Page|>
       <Page.header
         @showTabs={{true}}
+        @pageType="manage"
```
Monkeychip, line 6: used to determine on manage page or not, and if you are change the link the breadcrumb uses.

ui/lib/replication/addon/templates/mode.hbs
```diff
@@ -1,57 +1,62 @@
 <section class="section">
   <div class="container is-widescreen">
-    {{#if model.replicationAttrs.replicationEnabled}}
-      <PageHeader as |p|>
-        <p.top>
-          <nav class="key-value-header breadcrumb">
+    {{#if (and (eq model.drMode "secondary") (eq model.drModeInit "primary"))}}
+      <nav class="navbar"></nav>
```
chelshaw, line 4: This one doesn't have the `page-container` wrapping div and the one above does. Should we make them consistent?

ui/lib/replication/addon/templates/mode.hbs
```diff
@@ -1,57 +1,62 @@
 <section class="section">
   <div class="container is-widescreen">
-    {{#if model.replicationAttrs.replicationEnabled}}
-      <PageHeader as |p|>
-        <p.top>
-          <nav class="key-value-header breadcrumb">
+    {{#if (and (eq model.drMode "secondary") (eq model.drModeInit "primary"))}}
```
Monkeychip, line 3: This is the only change to this file, the rest of just indenting in to accommodate the new conditional

ui/lib/core/addon/templates/components/replication-header.hbs
```diff
@@ -25,9 +25,21 @@
...
-          {{#link-to "vault.cluster.replication-dr-promote"}}
-            Disaster Recovery
-          {{/link-to}}
+          {{#if isDisabled}}
+            {{#link-to "vault.cluster.replication.index"}}
+              Replication
+            {{/link-to}}
+          {{else}}
+            {{#if (eq pageType "manage")}}
+              {{#link-to "vault.cluster.replication-dr-promote.details"}}
```
chelshaw, line 34: A little curious why this link wouldn't go back to general Replication dashboard (`vault.cluster.replication.index`) instead. Or at the very least, maybe naming the breadcrumbs `Disaster Recovery Details` and `Disaster Recovery Manage` to be more indicative of what it's doing.

Monkeychip: Good thoughts.  @noelledaley do you have any opinions here?

ui/lib/replication/addon/templates/mode.hbs
Monkeychip: Good call.  I think the solution is to remove the page-container from above.  I'll play around and amend.  Thanks!

ui/lib/core/addon/templates/components/replication-header.hbs
Monkeychip: Confirmed with design that we'll remove breadcrumbs if it's a DR Secondary.

ui/lib/core/addon/templates/components/replication-header.hbs
```diff
@@ -15,7 +15,7 @@
 {{/if}}
 <PageHeader as |p|>
   <p.top>
-    {{#if (not isSummaryDashboard) }}
+    {{#if (not (or isSummaryDashboard isSecondary)) }}
```
Monkeychip, line 18: If is drSecondary then don't show breadcrumb, there is a conditional higher up that is making sure this is DR.

ui/lib/core/addon/templates/components/replication-header.hbs
```diff
@@ -25,8 +25,8 @@
           <span class="sep">
             /
           </span>
-          {{#link-to "vault.cluster.replication-dr-promote"}}
-            Disaster Recovery
+          {{#link-to "vault.cluster.replication.index"}}
```
Monkeychip, line 28: The only time a breadcrumb shows on the route `ui/vault/replication-dr-promote/details` or the `ui/vault/replication-dr-promote/` is when there is no DR Secondary and you land on the empty state component.  In this case. you can use the breadcrumb to navigate back to the replication index page which lets you setup replication.  See image here: 
image

ui/lib/replication/addon/templates/mode.hbs
Monkeychip, line 3: The rest of this is indentation changes due to wrapping everything in conditional

ui/lib/replication/addon/templates/index.hbs
```diff
@@ -1,6 +1,7 @@
 <section class="section">
   <div class="container is-widescreen">
     {{#if model.replicationIsInitializing }}
+      <nav class="navbar"></nav>
```
noelledaley, line 4: nice fix! ✨

PR #9389, opened 2020-07-02
Add mount type to request earlier

sgmiller:
In the case of requests with no vault token, routing never happens, where
MountType is typically populated on request.  We have it from the
MatchingMountEntry check earlier in request handling, so this populates it
earlier, ensuring "mount_type" is on all audit entries.

### DIFF
--- a/vault/request_handling.go
+++ b/vault/request_handling.go
@@ -582,6 +582,8 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
 	var nonHMACReqDataKeys []string
 	entry := c.router.MatchingMountEntry(ctx, req.Path)
 	if entry != nil {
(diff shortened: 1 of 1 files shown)


vault/request_handling.go
```diff
@@ -582,6 +582,8 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
 	var nonHMACReqDataKeys []string
 	entry := c.router.MatchingMountEntry(ctx, req.Path)
 	if entry != nil {
```
alexanderbez, line 584: nit: I think formatting is off here.

sgmiller:
> LGTM (just format). Should this be backported to 1.4.x?

No, the field was introduced in 1.5 in the first place.

PR #9390, opened 2020-07-02
http: revert resource quota changes

alexanderbez:
- Revert HTTP changes made in https://github.com/hashicorp/vault/pull/9330
- Replaces https://github.com/hashicorp/vault/pull/9361

/cc @vishalnayak @calvn @briankassouf


http/logical.go
```diff
@@ -39,12 +39,53 @@ func (b *bufferedReader) Close() error {
 	return b.rOrig.Close()
 }
 
-func buildLogicalRequestNoAuth(perfStandby bool, w http.ResponseWriter, r *http.Request) (*logical.Request, io.ReadCloser, int, error) {
+func buildLogicalPathNoAuth(r *http.Request) (string, int, error) {
```
calvn, line 42: nit: This is purely fore building a logical path using the http request so rename it to be simply `buildLogicalPath`?

http/logical.go
```diff
@@ -39,12 +39,53 @@ func (b *bufferedReader) Close() error {
...
 	ns, err := namespace.FromContext(r.Context())
 	if err != nil {
-		return nil, nil, http.StatusBadRequest, nil
+		return "", http.StatusBadRequest, nil
+	}
+
+	path := r.URL.Path
+	if strings.HasPrefix(path, "/v1/") {
+		path = ns.TrimmedPath(path[len("/v1/"):])
```
calvn, line 50: We want to always trim the namespace path regardless of the `v1/` presence so this should be outside of the prefix check.

http/logical.go
```diff
@@ -58,10 +99,14 @@ func buildLogicalRequestNoAuth(perfStandby bool, w http.ResponseWriter, r *http.
 		op = logical.DeleteOperation
 		data = parseQuery(r.URL.Query())
 	case "GET":
+		var (
```
calvn, line 102: Nit: Within functions, you usually want var declarations as close to the usage as possible to keep its context close so might be good to move it back down.

http/util.go
```diff
@@ -36,29 +36,49 @@ func rateLimitQuotaWrapping(handler http.Handler, core *vault.Core) http.Handler
...
 			if core.RateLimitAuditLoggingEnabled() {
+				requestId, err := uuid.GenerateUUID()
+				if err != nil {
+					respondError(w, http.StatusBadRequest, errwrap.Wrapf("failed to generate identifier for the request: {{err}}", err))
+					return
+				}
+
+				req := &logical.Request{
+					ID: requestId,
+					// TODO: Do we need op? If so, we should consider refactoring buildLogicalPathNoAuth and/or buildLogicalRequestNoAuth.
```
calvn, line 75: Yeah, we probably need the Operation as well.

calvn: We can move the logic within `buildLogicalRequestNoAuth` into `buildLogicalPathNoAuth` (again, this seems to be a bit confusing so consider renaming) and have it return an operation.

http/util.go
```diff
@@ -36,26 +37,46 @@ func rateLimitQuotaWrapping(handler http.Handler, core *vault.Core) http.Handler
...
+
+			req := &logical.Request{
+				ID:         requestId,
+				Operation:  op,
+				Path:       path,
+				Connection: getConnection(r),
+				Headers:    r.Header,
 			}
 
 			if core.RateLimitAuditLoggingEnabled() {
```
calvn, line 82: Since we're always returning within here, we could call `buildLogicalRequestNoAuth` to have it build us the logical request for auditing instead of manually instantiating one and possibly leaving out fields (noticed that `Data` was left out).

http/util.go
```diff
@@ -36,26 +37,41 @@ func rateLimitQuotaWrapping(handler http.Handler, core *vault.Core) http.Handler
...
-			quotaErr := errwrap.Wrapf(fmt.Sprintf("request path %q: {{err}}", req.Path), quotas.ErrRateLimitQuotaExceeded)
+			quotaErr := errwrap.Wrapf(fmt.Sprintf("request path %q: {{err}}", path), quotas.ErrRateLimitQuotaExceeded)
 			respondError(w, http.StatusTooManyRequests, quotaErr)
 
 			if core.Logger().IsTrace() {
-				core.Logger().Trace("request rejected due to lease count quota violation", "request_path", req.Path)
+				core.Logger().Trace("request rejected due to lease count quota violation", "request_path", path)
+			}
+
+			req, _, status, err := buildLogicalRequestNoAuth(core.PerfStandby(), w, r)
```
calvn, line 71: Since `req` is only used by the audit logic, let's move it into core.RateLimitAuditLoggingEnabled() so we can conditionally call this (vs calling this all the time)

vishalnayak: Done.

http/util.go
vishalnayak: Done.

http/util.go
```diff
@@ -36,29 +37,44 @@ func rateLimitQuotaWrapping(handler http.Handler, core *vault.Core) http.Handler
...
 			}
 
 			if core.RateLimitAuditLoggingEnabled() {
+				req, _, status, err := buildLogicalRequestNoAuth(core.PerfStandby(), w, r)
+				if err != nil || status != 0 {
+					respondError(w, status, err)
+					return
+				}
+
 				_ = core.AuditLogger().AuditRequest(r.Context(), &logical.LogInput{
```
calvn, line 78: I noticed that we ignore error in other place that we call this, but I wonder if we should at least log it on the server side (as a WARN maybe?) that this audit was dropped.

vishalnayak: Done.

alexanderbez:
Tested UI and auditing ✅

alexanderbez:
Let me wrap up the race faults WRT to rate-limit quotas before merging.

http/logical.go
```diff
@@ -141,7 +141,6 @@ func buildLogicalRequestNoAuth(perfStandby bool, w http.ResponseWriter, r *http.
 		}
 
 	case "OPTIONS":
-	case "HEAD":
```
briankassouf, line 144: This change seems unrelated to the resource quotas changes. Do you remember why it was added?

vishalnayak: I remember adding it as part of quotas changes. It was needed to be able to call buildLogicalRequestNoAuth for all the requests. I don't remember the exact failure that had triggered this (it was from tests though). Now that we reverted these changes, I figured there is no reason to keep this unjustified change, and reverted this as well.

briankassouf:
@alexanderbez I think you accidentally committed a test output file.

alexanderbez:
> @alexanderbez I think you accidentally committed a test output file.

Indeed, apologies!

PR #9391, opened 2020-07-02
Minor wording update in subnav

jescalan:
This wording change comes at the request of the education team!

### DIFF
--- a/website/data/subnav.js
+++ b/website/data/subnav.js
@@ -16,7 +16,7 @@ export default [
     url: 'https://www.hashicorp.com/products/vault/enterprise',
   },
   'divider',
-  { text: 'Learn', url: 'https://learn.hashicorp.com/vault' },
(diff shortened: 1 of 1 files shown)


website/data/subnav.js
```diff
@@ -16,7 +16,7 @@ export default [
     url: 'https://www.hashicorp.com/products/vault/enterprise',
   },
   'divider',
-  { text: 'Learn', url: 'https://learn.hashicorp.com/vault' },
+  { text: 'Tutorials', url: 'https://learn.hashicorp.com/vault' },
```
alexanderbez, line 19: I don't have a strong preference here, but if we change the nomenclature here, we should probably change it everywhere, correct? Including the subdomain.

jescalan: We are changing it everywhere that we point to learn in primary navigation, but not planning to change the subdomain. I was not the one to make this call, the education team might be able to provide more context if you want to reach out!

PR #9393, opened 2020-07-02
Changing changelog headers and update version

mladlow:


website/data/version.js
```diff
@@ -1,3 +1,3 @@
-export const VERSION = '1.4.2'
+export const VERSION = '1.4.3'
 export const CHANGELOG_URL =
-  'https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#142-may-21st-2020'
+  'https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#143'
```
pbernal, line 3: this link doesn't work, i think it has to be: https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#143-july-2nd-2020

mladlow: The changelog edits will change that.

PR #9397, opened 2020-07-03
Updating token documentation

devops-rob:
This pull request adds some additional detail to the docs to make it clearer that for a token to have the ability to create a child token, it needs to be created as an unlimited use token.  

A community member had an issue with this and I went through it with him after he posted on discuss - https://discuss.hashicorp.com/t/restricted-use-token-cannot-generate-child-tokens/10748/3 . 

@calvn pointed me in the right direction as to why the token_num_uses parameter needs to be set to 0 for a token to be able to create a child token. I thought this could be made a bit clearer in the docs to help other practitioners who may run into the same issue.

### DIFF
--- a/website/pages/docs/auth/approle.mdx
+++ b/website/pages/docs/auth/approle.mdx
@@ -95,6 +95,8 @@ management tool.
        secret_id_num_uses=40
    ```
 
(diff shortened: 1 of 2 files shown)


kalafut:
LGTM!

PR #9416, opened 2020-07-07
Don't require AWS_ROLE_SESSION_NAME to sts:AssumeRoleWithWebIdentity

### LINKED ISSUE #9415: Assuming an IAM Role with sts:AssumeRoleWithWebIdentity requires optional parameter AWS_ROLE_SESSION_NAME to be specified
**Describe the bug**
I recently upgraded to 1.4.3 so that I could use [IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) for assigning an IAM Role to my vault instances using its identity in Kubernetes.

The role session name is technically a [required parameter on the STS API](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html#API_AssumeRoleWithWebIdentity_RequestParameters). Though for other applications running in my cluster, I'd normally omit the role session name and allow it to be [chosen by the AWS SDK](https://github.com/aws/aws-sdk-go/blob/98a279965b6174326bd3fe3283cc534f2027b79d/aws/credentials/stscreds/web_identity_provider.go#L105-L109).

It looks like vault forces the AWS_ROLE_SESSION_NAME to be specified to even attempt to assume a role with a web identity though, unable to let the AWS SDK fill in a value.

**To Reproduce**
1. Follow steps to setup an IAM Role that can be assumed from a Kubernetes service account.
2. Deploy vault to kubernetes without AWS_ROLE_SESSION_NAME specified in the environment variables.
3. vault will fail to unseal if it requires AWS API access to unseal (for example via using KMS)

**Expected behavior**
I expect vault to assume an IAM Role via IRSA, generating a role session name if one is not provided.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.4.3
* Vault CLI Version (retrieve with `vault version`): 1.4.3
* Server Operating System/Architecture: linux/amd64

```json
{
  "api_addr": "https://example.com",
  "listener": {
    "tcp": {
      "address": "0.0.0.0:8200",
      "cluster_address": "[::]:8201",
      "tls_cert_file": "/vault/tls/vault.pem",
      "tls_disable": false,
      "tls_key_file": "/vault/tls/vault.pem"
    }
  },
  "seal": {
    "awskms": {
      "kms_key_id": "alias/vault",
      "region": "us-west-2"
    }
  },
  "storage": {
    "dynamodb": {
      "ha_enabled": "true",
      "region": "us-west-2",
      "table": "vault_us-west-2_lab"
    }
  },
  "ui": true
}
```

**Additional context**

orirawlings:
Fixes #9415


tvoran:
The test failures seem to be pre-existing on master, and pass locally.

TestSealMigration_TransitToShamir_Post14 - sealmigration
TestSealMigration_TransitToShamir_Post14/raft

PR #9429, opened 2020-07-08
📌 Hard Pin Website Dependencies

jonathantneal:
- 🎟️ [Asana](https://app.asana.com/0/0/1180156776385991/f)
- 🔍 [Preview](https://deploy-preview-9429--vault-www.netlify.app/)

This PR makes a change to the dependency management strategy for the website's Node.js dependencies. There should be 
no impact at all for anyone working with the website. The change makes the website more stable and less likely to have bugs at the hands of a bug in one of our dependencies.

More details on the reasoning behind this change:

- The way dependencies are currently handled, is that they will automatically update minor or patch versions behind the scenes on npm i, but not major versions. This is the node default and is normally referred to as "semver ranges".
- Unfortunately sometimes accidents happen with package authors and a bug is pushed through a minor/patch version. This causes our sites to break when it does happen, even without an explicit update to that dependency.
- We do regular dependency maintenance across all our sites to keep everything up to date anyway.
- Therefore, we have concluded that it's safer to "hard pin" to a set version of each dependency, not allowing for automatic updates across semver ranges. All dependency updates will happen during our manual update rounds and be subject to specific review by the webdev team.


website/pages/docs/secrets/databases/mongodbatlas.mdx
```diff
@@ -62,18 +63,17 @@ more information about setting up the database secrets engine.
...
+    of the role:
+
+        $ vault read database/creds/my-role
+        Key                Value
+        ---                -----
+        lease_id           database/creds/my-role/2f6a614c-4aa2-7b19-24b9-ad944a8d4de6
+        lease_duration     1h
+        lease_renewable    true
+        password           A1a-QwxApKgnfCp1AJYN
+        username           v-5WFTBKdwOTLOqWLgsjvH
```
jescalan, line 76: Let's hit this with a fenced code block while we're here

website/pages/docs/secrets/openldap/index.mdx
```diff
@@ -17,49 +17,49 @@ This plugin currently supports LDAP v3.
 
 1. Enable the OpenLDAP secret engine:
 
-    ```sh
-    $ vault secrets enable openldap
-    ```
+   ```sh
+   $ vault secrets enable openldap
+   ```
```
jescalan, line 22: this can be `shell-session`

website/pages/api-docs/secret/consul/index.mdx
```diff
@@ -149,12 +149,12 @@ $ curl \
 
 ### Sample request
 
-```sh
-curl \
-→     --request POST \
-→     --header "X-Vault-Token: ..."\
-→     --data @payload.json \
-→     http://127.0.0.1:8200/v1/consul/roles/example-role
```
jonathantneal, line 157: This is the only example where a character was removed (the `→` arrow). No other examples of `curl` included these arrows. I hope this bug fix can come along for the ride.

kalafut:
@BRKalow Given so many changes to the docs platforms since this PR, is it now obsolete/closable?

PR #9432, opened 2020-07-09
Update the seal migration docs

mjarmy:
This PR clarifies the seal migration docs a bit, and adds a section about using `step-down` on the old active node.

### DIFF
--- a/website/pages/docs/concepts/seal.mdx
+++ b/website/pages/docs/concepts/seal.mdx
@@ -141,37 +141,48 @@ These steps are common for seal migrations between any supported kinds and for
 any storage backend.
 
 1. Take a standby node down and update the [seal
-   configuration](/docs/configuration/seal). If the migration is from Shamir seal
-   to KMS seal, add the desired new KMS seal block to the config. If the migration
-   is from KMS seal to Shamir seal, add `disabled = "true"` to the old seal block.
-   If the migration is from KMS seal to another KMS seal, add `disabled = "true"`
-   to the old seal block and add the desired new KMS seal block. Now, bring the
-   standby node back up and run the unseal command on each by supplying the
-   `-migrate` flag. Supply Shamir unseal keys if old seal was Shamir, which will be
-   migrated as the recovery keys for the KMS seal. Supply recovery keys if the old
-   seal is one of KMS seals, which will be migrated as the recovery keys of the new
-   KMS seal or as Shamir unseal keys if the new seal is Shamir.
+   configuration](/docs/configuration/seal). 
 
-2. Perform step 1 for all the standby nodes, one at a time. It is necessary to
+    * If the migration is from Shamir seal to KMS seal, add the desired new KMS
+      seal block to the configuration. 
+    * If the migration is from KMS seal to Shamir seal, add `disabled = "true"`
+      to the old seal block.
+    * If the migration is from KMS seal to another KMS seal, add `disabled =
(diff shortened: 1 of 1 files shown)


website/pages/docs/concepts/seal.mdx
```diff
@@ -141,37 +141,48 @@ These steps are common for seal migrations between any supported kinds and for
...
+    Now, bring the standby node back up and run the unseal command on each key, by
+    supplying the `-migrate` flag. 
+
+    * Supply Shamir unseal keys if old seal was Shamir, which will be migrated
+      as the recovery keys for the KMS seal. 
+    * Supply recovery keys if the old seal is one of KMS seals, which will be
+      migrated as the recovery keys of the new KMS seal, or as Shamir unseal
+      keys if the new seal is Shamir.
+
+1. Perform step 1 for all the standby nodes, one at a time. It is necessary to
```
vishalnayak, line 162: Why is it that all the points are named `1.`?

alexanderbez: This is just how Markdown works I think. You can view the rendered Markdown [here](https://github.com/hashicorp/vault/blob/04c327803112c53f6a128517afd61ebf4464c6a0/website/pages/docs/concepts/seal.mdx)

website/pages/docs/concepts/seal.mdx
```diff
@@ -141,37 +141,48 @@ These steps are common for seal migrations between any supported kinds and for
...
+      seal block to the configuration. 
+    * If the migration is from KMS seal to Shamir seal, add `disabled = "true"`
+      to the old seal block.
+    * If the migration is from KMS seal to another KMS seal, add `disabled =
+      "true"` to the old seal block and add the desired new KMS seal block. 
+
+    Now, bring the standby node back up and run the unseal command on each key, by
+    supplying the `-migrate` flag. 
+
+    * Supply Shamir unseal keys if old seal was Shamir, which will be migrated
```
alexanderbez, line 156:
```suggestion
    * Supply Shamir unseal keys if the old seal was Shamir, which will be migrated
```

website/pages/docs/concepts/seal.mdx
alexanderbez, line 162: One thing @rberlind pointed out to me earlier while testing this, is that it might be helpful to add "expected" output for operators. Maybe a specific logline or subset of a logline.

rberlind:
@alexanderbez : Would you like me to test with a Vault Enterprise cluster using the `vault operator stepdown` command instead of `sudo systemctl vault stop` on the active?

website/pages/docs/concepts/seal.mdx
mjarmy: Yep, you can just use 1. for every point, and it will label them for you.

rberlind:
I did test using `vault operator step-down` instead of `sudo systemctl restart vault`, but the migration still failed with same errors.  So, using `step-down` might be better technique, but does not address the current issue with the post 1.4.0 migration process failing for Vault Enterprise 1.4.2.

website/pages/docs/concepts/seal.mdx
mjarmy: Can you tell me what you were thinking we could add here?  Maybe something about how to monitor the integrated storage status?

alexanderbez: Yeah that could be a good idea. Is there something that an operator should see regardless of the HA/storage engine used?

PR #9436, opened 2020-07-09
Remove namespace from mount_point label.

mgritter:
This makes mount_point consistent across metrics; previously some had the namespace label and some did not:

image


mgritter:
Should go in 1.5 GA, but not necessary for RC.

ncabatoff:
We don't need to apply this fix to the other metrics that specify a mount_point, e.g. wrapInCubbyhole?

mgritter:
> We don't need to apply this fix to the other metrics that specify a mount_point, e.g. wrapInCubbyhole?

No, you're right.  Should have moved everything to one place like you originally suggested.  :)

mgritter:
Test on enterprise: 
image

auth/github/, prefix/mysecret/, and auth/myauth/ are in ns1
auth/myauth/ (same name) and secret/ are in root

vault/token_store.go
```diff
@@ -1822,7 +1822,7 @@ func (ts *TokenStore) handleTidy(ctx context.Context, req *logical.Request, data
 				for index, child := range children {
 					countParentList++
 					if countParentList%500 == 0 {
-						percentComplete := float64(index)/float64(len(children))*100
```
mgritter, line 1825: Oops, somebody's not using gofmt I guess?

PR #9463, opened 2020-07-13
Update Alert banner message-body to match design system

Monkeychip:
The new design system has the alert banner text black.  This PR updates that. 

To have the least impact on other elements/components using the `.message-body` class I created a new .scss file for alert-banners and renamed the class it's using.  This keeps the change contained to the `alert-banner` component.

**Here are the design system pictures:**

**Here is the before the fix:**

**Here is the fix with the PR:**

### DIFF
--- a/ui/app/styles/core/alert-banner.scss
+++ b/ui/app/styles/core/alert-banner.scss
@@ -1,4 +1,3 @@
-// To match the updated design system, alert banner messages should have black text
 .alert-banner-message-body {
   border: 0;
   margin-top: $spacing-xxs;
--- a/ui/app/styles/core/alert-banner.scss
+++ b/ui/app/styles/core/alert-banner.scss
@@ -1,0 +1,7 @@
+// To match the updated design system, alert banner messages should have black text
+.alert-banner-message-body {
+  border: 0;
+  margin-top: $spacing-xxs;
+
+  color: $black;
+}
--- a/ui/app/styles/core.scss
+++ b/ui/app/styles/core.scss
@@ -15,6 +15,7 @@
 @import 'bulma/switch';
 
 // Override Bulma details where appropriate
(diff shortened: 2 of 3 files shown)


ui/app/styles/core/alert-banner.scss
```diff
@@ -0,0 +1,7 @@
+// To match the updated design system, alert banner messages should have black text
```
chelshaw, line 1: Personally I don't think this comment is needed, since it's clear what the styles are doing and one could look back at this PR for reasoning. But that's just a personal opinion/nit!

Monkeychip: Works for me!  I'll remove it.

ui/app/styles/core/alert-banner.scss
```diff
@@ -0,0 +1,6 @@
+.alert-banner-message-body {
+  border: 0;
+  margin-top: $spacing-xxs;
```
noelledaley, line 3: it looks like the `border` and `margin-top` here have the same values as the original `message-body` class. in that case, what do you think about extending `message-body` so these styles aren't duplicated? this way if we make any other styling changes to all messages (perhaps a new font or font weight, etc), they'll apply here too. i.e.:

```
.alert-banner-message-body {
@extend .message-body
color: $black;
}
```

Monkeychip: Ah, that's a great idea.  I didn't know you could do that.  Thanks!

Monkeychip: @noelledaley ... um. 🤔  this doesn't seem to work because `message-body` color is overriding the black here.  I would have to use `!important` after black to make it work, and that doesn't seem like an optimal solution.

PR #9465, opened 2020-07-13
Bug fix: PKI ca cert field set initial value to 720hr

Monkeychip:
We had originally added the `intiailvalue` on the `pki-certificate` file in this [PR](https://github.com/hashicorp/vault/pull/9405).  However, this set all initial values used in the PKI to 720hr, which was problematic when you created a certificate after you created your CA certificate; naturally, it outlived your CA certificate.

To fix this we are now only set the initial value of ca certs to 720h, and everything else to 30hr. 

**Note**: there is duplication between the `TtlPicker2` components displayed when `isCaCert` is true and when `isCaCert` is false.  We opted for the duplication instead of making a difficult to read inline conditional for the `intialValue` property, which is the only property that is different between the two.


ui/app/templates/partials/form-field-groups-loop.hbs
```diff
@@ -1,9 +1,10 @@
+do you see me: {{isCaCert}}
```
noelledaley, line 1: i think you'll want to remove this 😉

Monkeychip: Man, every time!  Thank you!

ui/lib/core/addon/templates/components/form-field.hbs
```diff
@@ -96,14 +96,25 @@
...
+      @initialValue='720h'
+    />
+  {{else}}
+    <TtlPicker2
+      @onChange={{action (action "setAndBroadcastTtl" valuePath)}}
+      @label={{labelString}}
+      @helperTextDisabled="Vault will use the default lease duration"
+      @helperTextEnabled="Lease will expire after"
+      @description={{attr.helpText}}
+      @initialValue={{or (get model valuePath) attr.options.setDefault}}
```
noelledaley, line 115: if the only thing that `isCaCert` changes is the `initialValue`, could `initialValue` be a computed property in the component? that'd keep this template a little cleaner. whaddya think?

ui/app/models/pki-certificate.js
```diff
@@ -53,7 +53,7 @@ export default DS.Model.extend({
   ttl: attr({
     label: 'TTL',
     editType: 'ttl',
-    defaultValue: '720h',
+    defaultValue: '30h',
```
noelledaley, line 56: another option for solving this would be to that would keep the `defaultValue`s set in the model and alleviate the need for computed properties in the component or conditionals in the template would be to overwrite `ttl` value that is extended inside the `pki-ca-certificate` model. i haven't tested it out but it's worth a shot! i think it's good to define values in the same source as much as possible. what are your thoughts?

ui/lib/core/addon/templates/components/form-field.hbs
noelledaley: heh, this idea is actually irrelevant given my other comment. i'll leave it here though just in case. :)

Monkeychip:
Closing for work that will change the workflow.

PR #9466, opened 2020-07-13
Fetch quota by name before updating it

vishalnayak:


vault/external_tests/quotas/quotas_test.go
```diff
@@ -126,6 +127,42 @@ func waitForRemovalOrTimeout(c *api.Client, path string, tick, to time.Duration)
...
+	require.NoError(t, err)
+	require.NotEmpty(t, s.Data)
+
+	// create a rate limit quota w/ empty path (same name)
+	_, err = client.Logical().Write("sys/quotas/rate-limit/secret-rlq", map[string]interface{}{
+		"rate":  7.7,
+		"burst": 8,
+		"path":  "",
+	})
+	require.NoError(t, err)
```
alexanderbez, line 157: So when I initially dug into this issue, it was my intuition that this should actually error -- i.e. it would seem you should only be able to mutate tunable fields of a resource quota, `path` not being one of them.

So in this example case, there is no longer a rate limit quota at the root level?

vishalnayak: There was no defined behavior in that case. We were allowing updates to all fields. I guess we can allow updates to the `path` field, which essentially means that we can "move" the quota around. I don't see any problems with this. Do you?

alexanderbez: Not inherently no, but we should make sure that's documented.

vishalnayak: Pushed a doc update.

website/pages/api-docs/system/lease-count-quotas.mdx
```diff
@@ -27,7 +27,10 @@ that can either be a namespace or mount.
 - `path` `(string: "")` - Path of the mount or namespace to apply the quota.
   A blank path configures a global lease count quota. For example `namespace1/`
   adds a quota to a full namespace, `namespace1/auth/userpass` adds a quota to
-  `userpass` in `namespace1`.
+  `userpass` in `namespace1`. Updating this field on an existing quota can have
+  "moving" effects. For example, updating `auth/userpass` to
+  `namespace1/auth/userpass` moves this quota from being a global mount quota to a
+  namespace specific mount quota.
```
briankassouf, line 33: Should this get the same "namespaces are an enterprise feature" message?

website/pages/api-docs/system/lease-count-quotas.mdx
```diff
@@ -27,7 +27,10 @@ that can either be a namespace or mount.
 - `path` `(string: "")` - Path of the mount or namespace to apply the quota.
   A blank path configures a global lease count quota. For example `namespace1/`
   adds a quota to a full namespace, `namespace1/auth/userpass` adds a quota to
-  `userpass` in `namespace1`.
+  `userpass` in `namespace1`. Updating this field on an existing quota can have
+  "moving" effects. For example, updating `auth/userpass` to
+  `namespace1/auth/userpass` moves this quota from being a global mount quota to a
```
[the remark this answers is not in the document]
vishalnayak: We have used "global" in the documentation. However, if you feel that can be confusing, let me know and I'll change it in all the places.

website/pages/api-docs/system/lease-count-quotas.mdx
vishalnayak: This does not need it because the whole page is marked as enterprise above in the file.

PR #9480, opened 2020-07-14
vault operator doc page example usage update for v1.4.2

assareh:
updating the vault operator example usage on the vault operator doc page for v1.4.2 which includes migrate and raft subcommands

### DIFF
--- a/website/pages/docs/commands/operator/index.mdx
+++ b/website/pages/docs/commands/operator/index.mdx
@@ -57,6 +57,8 @@ Subcommands:
     generate-root    Generates a new root token
     init             Initializes a server
     key-status       Provides information about the active encryption key
+    migrate          Migrates Vault data between storage backends
+    raft             Interact with Vault's raft storage backend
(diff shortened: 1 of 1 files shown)


aphorise:
Request is valid - however it's not clear how 1.4.x docs should be tagged or labelled according to the [current docs layout](https://www.vaultproject.io/docs) which now includes version and is different from the time of this PR.

aphorise:
Hey @assareh - thank you for reporting this.

This request is being closed without a merger due the following reasons:

  - Version 1.4.x included other exceptional operators that were beta such as `vault operator raft configuration` which existed briefly; therefore any backdated editions must then include additional note or explanations around these with reference to future that's also contradictory as those older versions are no longer supported nor could they be with the foresight of the future.

  - HashiCorp support scheme is for prior minor versions that are no farther than N-2 - so at the time of writing the versions: 1.11.x (latest), 1.10.x & 1.9.x are what's maintained.

PR #9483, opened 2020-07-15
Update date-fns

noelledaley:
# Update date-fns
Update date-fns to 2.0.0 which requires us update all of our date-related helpers. The most notable change we had to account for is that all of the date-fns functions no longer accept strings as arguments. You'll notice we also had to switch some of the formatting arguments from `YYYY` to `yyyy` as a result. See the full list of breaking changes when upgrading to date-fns 2.0.0 [here](https://github.com/date-fns/date-fns/blob/master/CHANGELOG.md#changed-4).

## Testing
This PR requires a considerable QA as it changes the behavior of our `date-from-now` and `date-format` helpers. We now return an empty string and console log an error if the formatting doesn't work, so there's a chance some dates may be missing in the UI.

Navigate to the various pages of the UI where `date-format` or `date-from-now` is used, such as the replication secondaries add modal which displays the expiration date. These pages should still display the date as expected.

Here's an example of the replication secondaries add modal:
image


ui/app/helpers/date-from-now.js
```diff
@@ -1,8 +1,17 @@
...
 
 export function dateFromNow([date], options = {}) {
-  return distanceInWordsToNow(date, options);
+  // debugger;
+  // const time = fromUnixTime(date / 1000);
+  let dateString;
+  try {
+    dateString = formatDistanceToNow(date, { ...options });
+  } catch (e) {
+    console.log(date);
```
Monkeychip, line 11: maybe also console log the error, or is this just for testing?

noelledaley: this is just for testing -- we haven't solved the issue yet 😭

ui/app/models/transit-key.js
```diff
@@ -43,7 +43,7 @@ const ACTION_VALUES = {
 
 export default DS.Model.extend({
   type: attr('string', {
-    defaultValue: 'aes256-gcm96',
+    defaultValue: 'aes128-gcm96',
```
Monkeychip, line 46: curious why we're lowing the bit count here (assuming the 256 stands for bits).  I'm also guessing, the higher the bit count the more secure?

ui/app/templates/components/auth-info.hbs
```diff
@@ -10,7 +10,7 @@
           <li class="action">
             <AlertBanner @type="warning" @message="We've stopped auto-renewing your token due to inactivity.
                 It will expire in {{date-from-now auth.tokenExpirationDate interval=1000 hideSuffix=true}}.
-                on {{date-format auth.tokenExpirationDate 'MMMM Do YYYY, h:mm:ss a'}}" />
+                on {{date-format auth.tokenExpirationDate 'MMMM Do yyyy, h:mm:ss a'}}" />
```
Monkeychip, line 13: MMMM is capitalized, but yyyy is not?

ui/app/models/transit-key.js
noelledaley: great question! in this case we're only changing the default value so it automatically selects the first item in the dropdown. the other option is still available. this way we're listing all options in alphabetic and numeric order:

image

ui/app/templates/components/auth-info.hbs
noelledaley: another excellent question! in the newest version of date-fns`yyyy` (lowercase) represents the calendar year, which is what we want, whereas `YYYY` represents the week numbering year. see the examples here: 

https://github.com/date-fns/date-fns/blob/master/docs/unicodeTokens.md#popular-mistakes

PR #9485, opened 2020-07-15
Add a lock to seal migration

mjarmy:
Add a lock to seal migration so we'll know if its already in progress

### DIFF
--- a/vault/core.go
+++ b/vault/core.go
@@ -46,6 +46,7 @@ import (
 	"github.com/hashicorp/vault/vault/quotas"
 	vaultseal "github.com/hashicorp/vault/vault/seal"
 	"github.com/patrickmn/go-cache"
+	uberAtomic "go.uber.org/atomic"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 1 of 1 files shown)


vault/core.go
```diff
@@ -236,6 +236,7 @@ type Core struct {
 	// about the seal we are migrating *from*.
 	migrationInfo *migrationInformation
 	sealMigrated  *uint32
+	migrationLock *int32
```
ncabatoff, line 239: May I suggest using Uber's go-atomic wrappers?  That way it's clear how this field is intended to be used.  I might also remove "lock" from the name, calling it "inSealMigration" or something like that, since it's not a mutex which is what we usually think of as locks.

ncabatoff: Oh, using the wrapper will also protect you against alignment issues on 32-bit platforms.

vishalnayak, line 239: I wonder if there is a reason why we always use `uint32` whenever we use this kind of locking.

mjarmy: I switched to uber-atomic

mjarmy: done

PR #9487, opened 2020-07-15
API and upgrade docs for global plugin reload

sgmiller:


website/pages/docs/upgrading/plugins.mdx
```diff
@@ -36,6 +36,7 @@ triggered, Vault will kill plugin_v1’s process and start a plugin_v2 process.
 -> **Important:** Plugin reload of a new plugin binary must be
 performed on each Vault instance. Performing a plugin upgrade on a single
```
sebinjohn, line 37: Is this still relevant if `scope` can be used to reload across all instances?

PR #8509, opened 2020-03-09
Add option allowed_domains_template enabling identity templating for issuing PKI certs

andrejvanderzee:
Implementation of identity templating when issuing a PKI cert using the exported sdk/framework merged in  #8088

Duplicate of #6558

Similar to  #7548

PR #9498, opened 2020-07-16
pki: Allow to use not only one variable during templating in allowed_domains

qk4l:
Improve #8509 PR. 
Now it's possible to use not only pure variable in allowed_domains but also combinations of other ones or  domain suffices.


andrejvanderzee:
> Improve #8509 PR.
> Now it's possible to use not only pure variable in allowed_domains but also combinations of other ones or domain suffices.

Looks good to me! Good improvement.

PR #9502, opened 2020-07-16
Resource Quotas: Remove 'burst' Param from Rate Limiter

alexanderbez:
Per internal discussions, remove `burst` as a configurable parameter for rate limit quotas.


mgritter:
Maybe change the PR title to say 'burst' instead of 'rate'?

alexanderbez:
haha @vishalnayak long day 😅

PR #9506, opened 2020-07-16
add upgrade sequence for DR and PR clusters to Upgrade Overview page

dedamico:
We were going to originally put this in the HC KB but thought it would be more appropriate to be in the Vault overview doc.  Thanks!

### DIFF
--- a/website/pages/docs/upgrading/index.mdx
+++ b/website/pages/docs/upgrading/index.mdx
@@ -103,3 +103,13 @@ Upgrading installations of Vault which participate in [Enterprise Replication](/
 - Verify functionality of each secondary instance after upgrading
 - When satisfied with functionality of upgraded secondary instances, upgrade
   the primary instance
+
+For Vault Enterprise customers who have both Disaster Recovery (DR) and Performance Replication (PR) clusters, the recommended sequence for upgrading all clusters is as follows:
+
+* Upgrade Disaster Recovery clusters
+
+* Upgrade Performance Replication secondary clusters
+
+* Upgrade Performance Replication primary clusters
+
+**NOTE**:  If the DR Primary is also a PR Primary, you will upgrade the DR secondary first then upgrade the PR secondary.  As the final step, upgrade the mixed DR+PR Primary. 
--- a/website/pages/docs/upgrading/index.mdx
+++ b/website/pages/docs/upgrading/index.mdx
@@ -103,3 +103,13 @@ Upgrading installations of Vault which participate in [Enterprise Replication](/
 - Verify functionality of each secondary instance after upgrading
 - When satisfied with functionality of upgraded secondary instances, upgrade
   the primary instance
+
+For Vault Enterprise customers who have both Disaster Recovery (DR) and Performance Replication (PR) clusters, the recommended sequence for upgrading is as follows:
+
+* Upgrade Disaster Recovery clusters
+
+* Upgrade Performance Replication secondary clusters
+
+* Upgrade Performance Replication primary clusters
+
+**NOTE**:  If the DR Primary is also a PR Primary, you will upgrade the DR secondary first then upgrade the PR secondary.  As the final step, upgrade the mixed DR+PR Primary. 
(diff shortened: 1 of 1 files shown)


website/pages/docs/upgrading/index.mdx
```diff
@@ -103,3 +103,13 @@ Upgrading installations of Vault which participate in [Enterprise Replication](/
 - Verify functionality of each secondary instance after upgrading
 - When satisfied with functionality of upgraded secondary instances, upgrade
   the primary instance
+
+For Vault Enterprise customers who have both Disaster Recovery (DR) and Performance Replication (PR) clusters, the recommended sequence for upgrading is as follows:
+
+* Upgrade Disaster Recovery clusters
```
mladlow, line 109: It seems weird to not have this broken out into secondary & primary the way performance replication is.

website/pages/docs/upgrading/index.mdx
```diff
@@ -103,3 +103,13 @@ Upgrading installations of Vault which participate in [Enterprise Replication](/
...
+
+* Upgrade Performance Replication secondary clusters
+
+* Upgrade Performance Replication primary clusters
+
+**NOTE**:  If the DR Primary is also a PR Primary, you will upgrade the DR secondary first then upgrade the PR secondary.  As the final step, upgrade the mixed DR+PR Primary.
```
mladlow, line 115: You may want the visual emphasis, but it also seems like it would be fine to just add another bullet of * Finally, upgrade any mixed DR and PR Primary clusters, or something like that.

mladlow:
It would be helpful to know what people found confusing about the original docs - I'm assuming it's that people wanted specific instructions for the case when they have DR & PR replication, but it would be nice to have that assumption not be an assumption.

dedamico:
> It would be helpful to know what people found confusing about the original docs - I'm assuming it's that people wanted specific instructions for the case when they have DR & PR replication, but it would be nice to have that assumption not be an assumption.

Yes, that is correct.  We have received support tickets related to this exact question so just wanted to encapsulate the guidance we have provided in the past and/or to get confirmation on what we should recommend.

website/pages/docs/upgrading/index.mdx
dedamico: I considered the pre-existing section immediately above my additions to address this, but let me know if you would prefer it to be emphasized again:

> - **Upgrade the replication secondary instances first** using appropriate
  guidance from the previous sections depending on whether each secondary
  instance is Non-HA or HA
> - Verify functionality of each secondary instance after upgrading
> - When satisfied with functionality of upgraded secondary instances, upgrade
  the primary instance

dedamico:
> I can't speak to correctness. This new section seems to slightly conflict with the above section, which says to upgrade secondary instances first, and would imply:
> 
> 1. DR Secondaries
> 2. PR Secondaries
> 3. DR Primaries
> 4. PR Primaries
> 5. Mixed clusters
> 
> So, I think we should get the section as a whole to be consistent.

I agree, and it would be great to get feedback from Eng on -- this is one of the reasons why we are wanting more prescriptive guidance in this update from Eng.  The existing section is not clear on this scenario, and honestly I'm not sure whether or not DR secondaries and PR secondaries should be upgraded BEFORE DR and PR primaries (as you mentioned), or if the order should go more along the lines of:

1. DR Secondaries
2. DR Primaries
3. PR Secondaries
4. PR Primaries

briankassouf:
If a cluster has PR & DR replication then some clusters will be acting in two modes. So it's a little more nuanced then the general order you outlined, though this order could be considered technically correct:

```
DR Secondaries
PR Secondaries
DR Primaries
PR Primaries
```
But DR Primaries will be another cluster type as well (either PR Secondary or PR Primary) and it's not necessary to upgrade them twice. 

I wonder if it would be more clear to add a diagram of a common 4 cluster (2 PR & 2 DR) setup, number them, and then give the correct order?

kalafut:

mladlow:
@dedamico I've pulled following up on this PR into an action item and ticket associated with a recent incident.

mladlow:
I think we want a diagram like https://learn.hashicorp.com/img/vault/replication/vault-dr-21.png - in this example you'd do B+D, then C, then A. I'm going to see if marketing design wants to modernize the diagram for the docs though.

PR #9514, opened 2020-07-17
replace .Data.[key] with .Data.data.[key]

oneacl:
Line 267 was correctly templating .Data.data.key but on the other lines the second .data was missing.


kalafut:

PR #9515, opened 2020-07-17
Get rid of sdk internal reloadutil package for now.

ncabatoff:
Motivated by this failure: https://circleci.hashicorp.engineering/gh/hashicorp/vault-enterprise/42316#tests/containers/1


calvn:
I'm a bit confused why the internal package can't be used by `stepwise/docker` since they are both rooted at `sdk/testing` which should work according to [the semantics](https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface) around internal packages.

PR #9518, opened 2020-07-17
docs: specify which JWT/OIDC role params support JSON pointer syntax in API docs

austingebauer:
This PR improves the JWT/OIDC docs by making it more obvious which role parameters support JSON pointer syntax.


website/pages/api-docs/auth/jwt/index.mdx
```diff
@@ -126,17 +126,20 @@ entities attempting to login. At least one of the bound values must be set.
   claim matches this value.
 - `bound_claims` `(map: <optional>)` - If set, a map of claims/values to match against.
   The expected value may be a single string or a list of strings. The interpretation of the bound
-  claim values is configured with `bound_claims_type`.
+  claim values is configured with `bound_claims_type`. Keys support [JSON pointer](/docs/auth/jwt#claim-specifications-and-json-pointer)
```
catsby, line 129: Nit: the `claim_mappings` below specifically mentions the map is pairing of keys and values, but `bound_claims` does not. Here we say "Keys support" which may not be totally clear to users what it is meant by "keys". What do you think about rewording the description of `bound_claims` to mention keys/values like `claim_mappings` has?

austingebauer: Good idea 👍 . Done in d18b02610fd472d21d7e3f187f2b04804b30f65d.

Issue #9509, opened 2020-07-16
Vault fails to validate certificate for MongoDB database connection (DocumentDB).


cyrus-mc:
**Describe the bug**

When configuring a database secret to mongoDB (in my case AWS DocumentDB) vault errors with "x509: certificate signed by unknown authority" even though I pass in via tls_ca the Certificate Authority used to sign the DocumentD certificates. 

**To Reproduce**

1. Run `vault write database/config/my-mongodb-database plugin_name=mongodb-database-plugin allowed_roles="my_role" connection_url="mongodb://{{username}}:{{password}}@unit-test01.cluster-cz6h536kutdi.xxxxxxx:27017/admin?tls=true" username="master" password="xxxx" tls_ca=@rds-ca-2019-root.pem`

Vault will respond with

```
Error writing data to database/config/my-mongodb-database: Error making API request.

URL: PUT http://localhost:8200/v1/database/config/my-mongodb-database
Code: 400. Errors:

* error creating database object: error verifying connection: server selection error: server selection timeout, current topology: { Type: Unknown, Servers: [{ Addr: unit-test01.cluster-cz6h536kutdi.xxxxxx.com:27017, Type: Unknown, State: Connected, Average RTT: 0, Last error: connection() : x509: certificate signed by unknown authority }, ] }
```

**Expected behavior**

The expectation is that the secret will be successfully added. According to the documentation the value of tls_ca is passed as option to `tlsCAFile` for Mongo. 

```
› mongo --tls --host unit-test01.cluster-cz6h536kutdi.xxxxx.com:27017 -u master -p xxxxx --tlsCAFile rds-ca-2019-root.pem
MongoDB shell version v4.2.8
connecting to: mongodb://unit-test01.cluster-cz6h536kutdi.xxxxxx.com:27017/?compressors=disabled&gssapiServiceName=mongodb
Implicit session: session { "id" : UUID("9058bb07-5d3a-42b6-938a-6ccb4d515a90") }
MongoDB server version: 3.6.0
WARNING: shell and server versions do not match

Warning: Non-Genuine MongoDB Detected

This server or service appears to be an emulation of MongoDB rather than an official MongoDB product.

Some documented MongoDB features may work differently, be entirely missing or incomplete, or have unexpected performance characteristics.

To learn more please visit: https://dochub.mongodb.org/core/non-genuine-mongodb-server-warning.

rs0:PRIMARY>
```

Which as shown above, works if I call mongo shell directly. 

**Environment:**

[...]

PR #9519, opened 2020-07-17
fix: Correctly handle TLS config when ssl=true or tls=true connection string option set.

cyrus-mc:
fix for #9509 

Merge writeOpts and authOpts after call too ApplyURI.


cyrus-mc:
Build is failing with error about using internal module. This PR doesn't touch any of that (it is a simple one line change) so not sure how this error is being caused.

cyrus-mc:
@ncabatoff Merged in latest master changes that fixed build. This PR is ready to go. I am using a locally build Vault, containing this fix, without issue.

ncabatoff:
Thanks @cyrus-mc!

pbar1:
Hey, FYI - this change seemed to break our Mongo db engine upon upgrading form 1.5 -> 1.6. When setting up the provider, we use client TLS _as well as_ the username/password authentication via SCRAM-SHA-256. Due to the order in which this change merges config, if client TLS configuration is passed at all, it will always win, making other auth mechanisms unusable.

PR #9521, opened 2020-07-17
Revert "Get rid of sdk internal reloadutil package for now."

jefferai:
Reverts hashicorp/vault#9515

Anything in internalshared is by definition being used by a non-Vault piece of code, which this just broke.


ncabatoff:
I didn't touch internalshared.

PR #9524, opened 2020-07-17
Ui/update datefns

noelledaley:
# Update date-fns
Update date-fns to 2.0.0 which requires us update all of our date-related helpers. The most notable change we had to account for is that all of the date-fns functions no longer accept strings as arguments. You'll notice we also had to switch some of the formatting arguments from `YYYY` to `yyyy` as a result. See the full list of breaking changes when upgrading to date-fns 2.0.0 [here](https://github.com/date-fns/date-fns/blob/master/CHANGELOG.md#changed-4).

## Testing
This PR requires a considerable QA as it changes the behavior of our `date-from-now` and `date-format` helpers. We now return an empty string and console log an error if the formatting doesn't work, so there's a chance some dates may be missing in the UI.

Navigate to the various pages of the UI where `date-format` or `date-from-now` is used, such as the replication secondaries add modal which displays the expiration date. These pages should still display the date as expected.

Here's an example of the replication secondaries add modal:
image


noelledaley:
@chelshaw @Monkeychip i'm closing this PR for now since it wasn't needed to fix the transit bug from last week and i don't see a huge need to update datefns just yet. if you encounter a reason to during the Ember upgrade you can re-open it, rebase and merge. :)

chelshaw:
Reopening this so we can take advantage of [RFC3339 formatting](https://date-fns.org/v2.8.0/docs/formatRFC3339) (available in >=2.8) which is the format used for Pricing Metrics api parameters.

aphorise:
@noelledaley @chelshaw is this PR still relevant?

hashishaw:
No longer relevant, updated in #10848 😄

PR #9525, opened 2020-07-17
Update documentation to support new OIDC mode parameter

broamski:
Relies on https://github.com/hashicorp/vault-plugin-auth-jwt/pull/125


broamski:
Abandoning as the related PR was never accepted.

PR #9529, opened 2020-07-20
approle: Include role_name in alias metadata

### LINKED ISSUE #8128: Add approle metadata to identity alias
Could we allows for writing identity token templates that include attribute:[role_name]?
(And including this attribute in path templates)
If Vault can add AppRole's role name to entity alias metadata, and then we can reference Entity metadata in our ACL templates directly.

Reference:
https://github.com/hashicorp/vault/pull/7985
https://github.com/hashicorp/vault/pull/2985
https://github.com/hashicorp/vault/issues/5916

endocrimes:
This change allows people who are using templated policies to use the role's name in their templates through `{{ identity.entity.aliases.approle.metadata.role_name }}`.

I'm not sure if there is a better place to implement this, but it seems like the most reliable bet to ensure that it gets populated on future logins?

Closes #8128.


endocrimes:
Test failure seems like a flake, but it looks like Circle doesn’t let external folks re-run tests :(

endocrimes:
@kalafut / @calvn any chance either of you could take a look? 😄

kalafut:
@calvn As this metadata is stable (i.e. not a volatile instance_id or the like) I think it would be OK to store by default. WDYT?

endocrimes:
@kalafut Yeah that's my understanding here too :) - and seems to hold up in my test environment with this patch deployed

calvn:
Merging this one. The race failure does not seem to be related.

endocrimes:
@calvn thanks! 🎉 ✨

PR #9530, opened 2020-07-20
Improve logging, and add polling to the post-stepdown leader check.

ncabatoff:


vault/external_tests/sealmigration/seal_migration_test.go
```diff
@@ -367,9 +367,16 @@ func migratePost14(t *testing.T, logger hclog.Logger, storage teststorage.Reusab
...
+	var leaderIdx int
+	for i := 0; i < 30; i++ {
+		leaderIdx, err = testhelpers.AwaitLeader(t, cluster)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if leaderIdx != 0 {
+			break
+		}
+		time.Sleep(1 * time.Second)
```
mjarmy, line 379: nitpick: I think you can just say `time.Second` here.

ncabatoff: Yeah, that's what I often do.  I just thought this would be clearer.

PR #9535, opened 2020-07-20
1.5 Downloads Page Updates

amanoske:
PR for updates to the downloads page


calvn:
Screen Shot 2020-07-20 at 12 31 09 PM

The downloads page should already have a banner for 1.5.0-rc. We don't update the reference on the stable version until after GA of a release.

amanoske:
@calvn Looking to add an explicit release notes to the page. I'm actually going to close this PR and just open the updated addition of a new .md under master

PR #9538, opened 2020-07-20
add approle auto_auth example

mikegreen:
Add an example config file for Vault Agent using approle and enabling auto-auth


mikegreen:
docs structure moved. made new PR 12297

PR #9540, opened 2020-07-21
docs: add release notes

mladlow:


website/pages/docs/release-notes/1.5.0.mdx
```diff
@@ -0,0 +1,64 @@
...
+layout: docs
+page_title: 1.5.0
+sidebar_title: 1.5.0
+description: |-
+  This page contains release notes for Vault 1.5.0.
+---
+
+# Vault 1.5.0
+
+## Vault 1.5 Release Highlights
```
calvn, line 11: Is there a particular sorting order for this list?

website/pages/docs/release-notes/1.5.0.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+**Resource Quotas:**  A new set of functionality called Resource Quotas that allows operators to define API quotas which impose usage limits at the system, endpoint, and namespace level (Enterprise Only) in order to ensure applications and users do not overuse Vault’s resources.  The following quota options will be supported:
+* **Lease Count Quotas (Enterprise Only):** Allows operators to specify lease count quotas. If the number of leases in the cluster hits the configured quota limits, additional lease creations will be forbidden for all clients until a lease has been revoked or has expired.
+* **Rate Limit Quotas (All versions):** Allows operators to specify request-per-second quotas. Rate limit quotas are applicable to every node in the Vault cluster, meaning each node will maintain separate counters to enforce rate limits. If the rate limit quota limit is hit on any of the nodes in the Vault cluster, additional requests will be canceled for all clients.
```
calvn, line 15: I think you could remove the “(All versions)” reference since no enterprise-only reference should imply that.

website/pages/docs/release-notes/1.5.0.mdx
```diff
@@ -0,0 +1,64 @@
...
+* Updated DR and performance primary dashboards that include a list of known secondaries and their corresponding UI URLs (for easier context-switching between the UIs)
+* Redesigned management workflows 
+
+**Integrated Storage for HA Coordination:** Vault now has the option to specify Integrated Storage as the HA coordination option for Vault, when any durable data storage backend (that may or may not support HA coordination) is used.  
+
+**Vault Monitor Command:** A new command, “vault monitor”, which allows users to stream logs of a running Vault server. The log level selected can be different from the log level used by the server logs.
+
+**Password policies:** Vault generates passwords as part of the dynamic secrets workflows. These passwords are created with high complexity for security reasons. However, certain organizations  have internal policies that are less stringent and restricted to usage of certain character sets. The new Password Policies feature in Vault 1.5 allows an operator to configure secret engines to generate passwords that comply with character set requirements of those systems.
```
calvn, line 31: I think we tend to capitalize each word in feature titles (here and below as well). Looks like we do in some of them but we should make it consistent.
```suggestion
**Password Policies:** Vault generates passwords as part of the dynamic secrets workflows. These passwords are created with high complexity for security reasons. However, certain organizations  have internal policies that are less stringent and restricted to usage of certain character sets. The new Password Policies feature in Vault 1.5 allows an operator to configure secret engines to generate passwords that comply with character set requirements of those systems.
```

website/pages/docs/release-notes/1.5.0.mdx
pbernal: No - i think this is just what Product wrote down in the draft.

mladlow: I don't know - a google doc went around for comments over the course of the last week but probably could have been more circulated.

website/pages/docs/release-notes/1.5.0.mdx
mladlow: I'm going to make these changes for consistency but not the other because these have already been somewhat reviewed.

PR #9342, opened 2020-06-29
Resource Quotas Doc

alexanderbez:
* API docs (config, rate limit and lease count)
* Concepts doc (rate limit)
* Enterprise doc (lease count)

PR #9544, opened 2020-07-21
docs: add pages to sidebar

alexanderbez:
Add quotas pages to sidebar. ref https://github.com/hashicorp/vault/pull/9342


website/data/api-navigation.js
```diff
@@ -111,6 +111,7 @@ export default [
       'internal-ui-mounts',
       'key-status',
       'leader',
+      'lease-count-quotas',
```
mladlow, line 114: I think on the actual website these should all be grouped together.

alexanderbez: `website/data/api-navigation.js` looks to be alphabetically ordered so I was just following convention. I can group these if my assumption is wrong.

mladlow: I did notice that, but I think the usability of the website should take precedence of the alphabetical ordering of the file? We could rename the file with some kind of prefix if we want to keep the ordering in this file alphabetical, or if there's a linter insisting on alphabetical order in the file.

alexanderbez: I don't think there is a linter? I'll group them 👍

PR #1375, opened 2016-05-04
Improve documentation around token TTLs.

BRMatt:
This change is sparked by a discussion[0] about vault not appearing to
respect the `increment` passed to `vault token-renew`. It turns out
this is expected behaviour if the token's new TTL would exceed the
Backend/System's Max Lease TTL.

[0]: https://github.com/hashicorp/vault/issues/1079

PR #9545, opened 2020-07-21
Use a large machine for go test races, and only single-package parall…

ncabatoff:
…elism.  This will hopefully alleviate the memory pressure that's causing timeouts/OOMs.


ncabatoff:
Closing this in favour of the ent PR #1375 which does the same thing, but only on the ent side.  Go race tests on the OSS side don't seem to be running out of resources.

PR #9546, opened 2020-07-21
Link to release notes on downloads page

mladlow:
Is it ok to add children to ProductDownloader so we can get this link to our release notes?


mladlow:
Also @pbernal I'm happy to change the language if needed - let me know.

mladlow:
That's a good point, we're only doing them for major releases, so I'm going to remove the dependency on VERSION and just link to the index.

PR #9551, opened 2020-07-21
Update retryablehttp to fix a data race

ncabatoff:
…with stdlib http.


ncabatoff:
The fix was in https://github.com/hashicorp/go-retryablehttp/pull/105.

mgritter:
Hooray! 🥇 💯 🎊

PR #9555, opened 2020-07-22
Allow non-string values in the seal stanza

ncabatoff:
With this fix, we still don't allow an unquoted `disabled` boolean value in the seal stanza (yields "unknown type for string *ast.LiteralType").  Not sure if that's something we should address now.

### DIFF
--- a/internalshared/configutil/kms.go
+++ b/internalshared/configutil/kms.go
@@ -68,51 +68,49 @@ func parseKMS(result *[]*KMS, list *ast.ObjectList, blockName string, maxKMS int
 			key = item.Keys[0].Token.Value().(string)
 		}
 
-		// We first decode into a map[string]interface{} because purpose isn't
-		// necessarily a string. Then we migrate everything else over to
-		// map[string]string and error if it doesn't work.
-		var m map[string]interface{}
-		if err := hcl.DecodeObject(&m, item.Val); err != nil {
-			return multierror.Prefix(err, fmt.Sprintf("%s.%s:", blockName, key))
-		}
-
+		var disabled bool
 		var purpose []string
 		var err error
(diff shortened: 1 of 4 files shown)


internalshared/configutil/kms.go
```diff
@@ -68,51 +68,49 @@ func parseKMS(result *[]*KMS, list *ast.ObjectList, blockName string, maxKMS int
...
-			if vs, ok := v.(string); ok {
-				strMap[k] = vs
-			} else {
-				return multierror.Prefix(fmt.Errorf("unable to parse 'purpose' in kms type %q: value could not be parsed as string", key), fmt.Sprintf("%s.%s:", blockName, key))
-			}
+		var cfg map[string]string
+		if err := hcl.DecodeObject(&cfg, item.Val); err != nil {
+			return multierror.Prefix(err, fmt.Sprintf("%s.%s:", blockName, key))
 		}
+		delete(cfg, "purpose")
```
jefferai, line 104: Don't you want these two deletions to happen prior to the DecodeObject call a few lines above?

ncabatoff: Prior to DecodeObject cfg is empty.

jefferai: Oh, I didn't realize you were decoding the same item twice, I figured the second time around you'd decode the first 'm' val. So this is probably not an issue so long as hcl doesn't choke on decoding a string array into a string.

ncabatoff: The test I have seems to show that it doesn't.  Are there other scenarios you think I should be testing too?

jefferai: Nope, not at this time!

PR #9557, opened 2020-07-22
plugin: fix process leak when Setup() fails

sorindumitru:
We've noticed some leftover processes from vault plugins on our boxes. Some
of them were even left over from previous instances of the service and reparented
to init.  This could cause issues if too many of them accumulate.

When running with TRACE logging the logs showed that there was an error return by
the call to Setup() the plugin. Looking through the code it looks like we do not
call Cleanup() in that case.


ncabatoff:
Sorry, still CLA issues it seems. I'm going to override it.

pierresouchay:
Thank you for solving this @sorindumitru !

PR #9559, opened 2020-07-22
Docs: Update MongoDB to support root credential rotation

pcman312:
Update MongoDB docs to show that it supports root credential rotation.


pcman312:
Good catch @austingebauer - updated

austingebauer:
LGTM 👍

pcman312:
FYI - forced pushed because the last commit message got garbled by my command line

Issue #9553, opened 2020-07-22
Vault 1.5 crashes on linux on arm if telmetry is enabled


joltedbot:
**Describe the bug**
After upgrading to Vault 1.5 Linux/Arm I enabled telemetry in the vault config file for the first time. When I do the active vault node crashes on restart with a go panic.  "panic: invalid argument to Intn".  See the bottom for the rest of the error.

I have a 3 node HA Cluster  so it just keeps setting a new node as active and that node fails and passes it to the next around in a circle.
If I remove the telemetry stanza it works again.  If I put back the telemetry stanza with any combination of parameters I get the same panic.

**To Reproduce**
1. Update Vault from 1.4.3 to 1.5 Linux Arm
2. Add any telemetry stanza to the config file
3. Restart Vault

**Expected behavior**
Vault should successfully start up and expose the Prometheus telemetry metrics at:  /v1/sys/metrics

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.5.0
* Vault CLI Version (retrieve with `vault version`): v1.5.0
* Server Operating System/Architecture: Raspbian (Kernel: 4.19.66-v7+   Arch: armv7l)

Vault server configuration file(s):

```hcl

ui = true

listener "tcp" {
  address     = "0.0.0.0:8200"
  cluster_address  = "0.0.0.0:8201"
  tls_cert_file = "xxxxxxxxxxxxx"
  tls_key_file = "xxxxxxxxxxxxx"
  tls_min_version = "tls12"
}

storage "consul" {
  address = "x.x.x.x:8501"
  path    = "vault/"
  scheme        = "https"
  token = "xxxxxxxxxxxxx"
  tls_ca_file   = "xxxxxxxxxxxxx"
  tls_cert_file = "xxxxxxxxxxxxx"
  tls_key_file  = "xxxxxxxxxxxxx"
}

api_addr = "https://x.x.x.x:8200"
cluster_addr = "https://x.x.x.x:8201"

log_level = "error"

seal "awskms" {
  region     = ""xxxxxxxxxxxxx"
  access_key = "xxxxxxxxxxxxx"
  secret_key = "xxxxxxxxxxxxx"
  kms_key_id = "xxxxxxxxxxxxx"
}

telemetry {
  prometheus_retention_time = "30s"
  disable_hostname = true
}

```

**Additional context**

[...]

PR #9563, opened 2020-07-22
Add upgrade note for #9553.

mgritter:
https://github.com/hashicorp/vault/issues/9553 reports that telemetry crashes Vault on 32-bit systems. Nick already created a fix, but this upgrade note explains the problem and suggests a workaround.

### DIFF
--- a/website/pages/docs/upgrading/upgrade-to-1.5.0.mdx
+++ b/website/pages/docs/upgrading/upgrade-to-1.5.0.mdx
@@ -28,7 +28,7 @@ the configuration value as necessary. For additional details, please see
 ## Known Issues
 
 Enabling telemetry on 32-bit systems will cause Vault to crash.  A workaround for this
-issue is to disable collection of usage gauges in the telemetry stanza of the configuration.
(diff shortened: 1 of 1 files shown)


website/pages/docs/upgrading/upgrade-to-1.5.0.mdx
```diff
@@ -0,0 +1,28 @@
...
+issue is to disable collection of usage gauges in the telemetry stanza of the configuration.
+
+```
+telemetry {
+  ...
+  usage_gauge_period = "none"
+}
+```
+
+This will suppress the metrics `vault.identity.entity.count`, `vault.identity.entity.alias.count`, `vault.token.count`, `vault.token.count.by_auth`, `vault.token.count.by_policy`, `vault.token.count.by_ttl` and `vault.secret.kv.count`, but all other Vault telemetry will remain available.
```
ncabatoff, line 27: Only thing I might add is to make explicit that this config setting won't impact any metrics that existed prior to 1.5.

mgritter: Good idea, added.

PR #9566, opened 2020-07-22
Add agent kerberos disable_fast_negotiation

sbutler:
The vault agent config is documented as supporting the kerberos disable_fast_negotiation option, however this was not implemented in code. This PR adds the kerberos disable_fast_negotiation config option.


sbutler:
Fixes #9567

jasonodonnell:
Fixed in https://github.com/hashicorp/vault/pull/9892 and back porting to 1.5.4 Vault.

PR #9569, opened 2020-07-22
Add upgrade note for #9555.

ncabatoff:


mgritter:
The links in the version that went live appear to be broken: https://www.vaultproject.io/docs/upgrading/upgrade-to-1.5.0, but it looks like your version used an absolute URL.

PR #9570, opened 2020-07-22
Fix links to documentation.

mgritter:
The links at https://www.vaultproject.io/docs/upgrading/upgrade-to-1.5.0 were interpreted relative to "upgrading".

However, I don't see 1.5.0 listed in the index on the right, is there some other step necessary to get it to appear?


mgritter:
Verified at https://5f18a4dde0a5480008659afe--vault-www.netlify.app/docs/upgrading/upgrade-to-1.5.0 that all the links work (and the sidebar is updated.)

PR #9572, opened 2020-07-22
Ui/ttl form

chelshaw:
This component will be used in upcoming work where a Time To Live (TTL) value is a required input. It has the same underlying functionality as the TtlPicker2 (minus the toggle) so I chose to move the base functionality for the unit and time into this this simpler TTL version, and extend the TtlPicker2 from that. 

ttl-form


ui/lib/core/addon/components/ttl-form.js
```diff
@@ -0,0 +1,104 @@
+/**
+ * @module TtlForm
+ * TtlForm components are used to enter a Time To Live (TTL) input.
+ * This component does not include a label and is designed to take
+ * a time and unit, and pass an object including seconds and
+ * timestring when those two values are changed.
+ *
+ * @example
+ * ```js
+ * <TtlForm @onChange={action handleChange} @unit={{m}}/>
```
Monkeychip, line 10: curious if the double curly brackets around m should be one?

ui/lib/core/addon/components/ttl-form.js
```diff
@@ -0,0 +1,104 @@
...
+    this.setProperties({
+      time: newTime,
+      unit: newUnit,
+    });
+  },
+  updateTime: task(function*(newTime) {
+    this.set('errorMessage', '');
+    let parsedTime;
+    parsedTime = parseInt(newTime, 10);
+    if (!newTime) {
```
Monkeychip, line 73: nice errorMessage handling!

ui/lib/core/addon/components/ttl-form.js
```diff
@@ -0,0 +1,104 @@
...
+      this.set('errorMessage', 'This field is required');
+      return;
+    } else if (Number.isNaN(parsedTime)) {
+      this.set('errorMessage', 'Value must be a number');
+      return;
+    }
+    this.set('time', parsedTime);
+    this.handleChange();
+    if (Ember.testing) {
```
Monkeychip, line 82: oh the fun bits of concurrency and testing; I love concurrency, I hate testing it.

ui/lib/core/addon/components/ttl-form.js
```diff
@@ -0,0 +1,104 @@
...
+      { label: 'days', value: 'd' },
+    ];
+  }),
+  handleChange() {
+    let { time, unit, seconds } = this.getProperties('time', 'unit', 'seconds');
+    const ttl = {
+      seconds,
+      timeString: time + unit,
+    };
+    this.onChange(ttl);
```
noelledaley, line 60: do you need a `return` here?

ui/lib/core/stories/ttl-form.md
```diff
@@ -0,0 +1,29 @@
+<!--THIS FILE IS AUTO GENERATED. This file is generated from JSDoc comments in lib/core/addon/components/ttl-form.js. To make changes, first edit that file and run "yarn gen-story-md ttl-form" to re-generate the content.-->
```
noelledaley, line 1: pro-tip: now that we've upgraded storybook you should be able to take advantage of Storybook Docs, which will generate these pages for you 👯‍♀️

https://github.com/storybookjs/storybook/tree/master/addons/docs

noelledaley:
i really like what you've done here by extending the `TTLForm` as a base class. out of curiosity, could you do the same thing with the old `TTLPicker`?

ui/lib/core/addon/components/ttl-form.js
chelshaw: Should actually just be quotes, good catch

ui/lib/core/addon/components/ttl-form.js
chelshaw: I don't think so -- it's not an async function and nothing is waiting on its value. It's essentially just calculating and triggering a side-effect.

chelshaw:
> i really like what you've done here by extending the `TTLForm` as a base class. out of curiosity, could you do the same thing with the old `TTLPicker`?

Good question! I decided against this because `TtlForm` and `TtlPicker2` diverge from the original `TtlPicker` in terms of expected inputs and outputs. It is my full intention to replace all of the TtlPicker instances with one of the updates ones, and then remove that component (and rename TtlPicker2)

PR #9577, opened 2020-07-23
Migrate to azure-storage-blob-go

elsesiy:
The azure sdk for go is maintenance-only for storage, see https://github.com/Azure/azure-sdk-for-go/tree/master/storage\#azure-storage-sdk-for-go-preview
Migrate to new azure-storage-blob-go SDK
Minor test improvements


elsesiy:
@briankassouf or @kalafut Can you please have a look? I left two `TODOs` open for discussion but would like to get some momentum going on this, thanks!

elsesiy:
@calvn Just rebased again. Can you have a look please?

physical/azure/azure_test.go
```diff
@@ -54,64 +56,52 @@ func TestAzureBackend(t *testing.T) {
...
+		blobService, _ := containerURL.GetProperties(ctx, azblob.LeaseAccessConditions{})
+		if blobService.StatusCode() == 200 {
+			_, _ = containerURL.Delete(ctx, azblob.ContainerAccessConditions{})
+		}
+	}()
 
-	cleanupEnvironment, err := environmentForCleanupClient(environmentName, environmentUrl)
-	if err != nil {
-		t.Fatalf("err: %s", err)
+	physical.ExerciseBackend(t, backend)
```
calvn, line 85: I'm experiencing an error in here claiming that the blob does not exist:
```
    azure_test.go:85: idempotent delete: -> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /Users/calvin/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.10.0/azblob/zc_storage_error.go:42
        ===== RESPONSE ERROR (ServiceCode=BlobNotFound) =====
        Description=The specified blob does not exist.
        RequestId:e723de6b-a01e-0074-8090-93ab1d000000
```

physical/azure/azure.go
```diff
@@ -198,15 +211,12 @@ func (a *AzureBackend) Get(ctx context.Context, key string) (*physical.Entry, er
...
-		Container: a.container,
-		Name:      key,
-	}
-
 	a.permitPool.Acquire()
 	defer a.permitPool.Release()
 
-	_, err := blob.DeleteIfExists(nil)
+	blobURL := a.container.NewBlockBlobURL(key)
+	_, err := blobURL.Delete(ctx, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
```
calvn, line 218: Does this return an error if the blob did not exist?

calvn (commented):
Thanks for the contribution! I left a few comments on the PR, mainly around the testing changes and error handling within them. I am still thinking about the proper context and retry values to use.

physical/azure/azure.go
calvn: Update: Seems like the idempotent delete test failure on `TestBackend` is due to this.

elsesiy: @calvn https://github.com/Azure/azure-storage-blob-go/blob/48358e1de5110852097ebbc11c53581d64d47300/azblob/zt_examples_test.go#L940 describes this behavior. We can change it to `DeleteSnapshotsOptionInclude`, what do you think?

physical/azure/azure_test.go
elsesiy: See comment on `Delete`

physical/azure/azure_test.go
```diff
@@ -3,49 +3,51 @@ package azure
...
 	return azure.EnvironmentFromName(name)
 }
 
-func TestAzureBackend(t *testing.T) {
-	if os.Getenv("AZURE_ACCOUNT_NAME") == "" ||
-		os.Getenv("AZURE_ACCOUNT_KEY") == "" {
-		t.SkipNow()
-	}
-
+func testFixture(t *testing.T) (azblob.ContainerURL, physical.Backend) {
```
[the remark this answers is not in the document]
elsesiy: I refactored the `testFixture` to return the `cleanup` function

physical/azure/azure.go
calvn: I think using `DeleteSnapshotsOptionInclude ` is fine.

The failure on the test is due to the deletion returning an error if the blob does not exist, unlike the old behavior by `DeleteIfExists`. This occurs on the latest commit (`997442d`) as well.

```
    azure_test.go:98: idempotent delete: -> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /Users/calvin/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.10.0/azblob/zc_storage_error.go:42
        ===== RESPONSE ERROR (ServiceCode=BlobNotFound) =====
        Description=The specified blob does not exist.
```

calvn: I don't see an equivalent method to `DeleteIfExist` on the new SDK, so you might have to do some error type checking for `azblob.ServiceCodeBlobNotFound`, similar to how you've done it on `Get`.

elsesiy: Sure I could catch that but the bigger question is why does the plugin tries to delete a blob that has already been deleted?

calvn: Storage backend implementations on the delete operation tends to no-op if the data is non-existent instead of returning an error. This is true for the current implementation of the azure storage backend via the `DeleteIfExists` method. It’s technically doing a read before a delete which is slightly different than doing a delete and ignoring the error if it’s a “not found”, but the end result should be the same — we’re not returning an error if the data didn’t exist.

There is no race in the test, it’s simply trying to delete some arbitrary key value that never got written/added.

physical/azure/azure.go
```diff
@@ -99,19 +99,37 @@ func NewAzureBackend(conf map[string]string, logger log.Logger) (physical.Backen
...
+		fmt.Sprintf("https://%s.blob.%s/%s", accountName, environment.StorageEndpointSuffix, name))
 	if err != nil {
-		return nil, errwrap.Wrapf(fmt.Sprintf("failed to create %q container: {{err}}", name), err)
+		return nil, errwrap.Wrapf("failed to create Azure client: {{err}}", err)
+	}
+
+	p := azblob.NewPipeline(credential, azblob.PipelineOptions{})
+
+	// TODO which context?
+	ctx := context.Background()
```
calvn, line 116: Let's use a context with a timeout of 5s, similar to [etcd3's](https://github.com/hashicorp/vault/blob/170521481d58961761bfd6877221b005cf0f4c7e/physical/etcd/etcd3.go#L130) behavior.

physical/azure/azure.go
```diff
@@ -167,22 +178,24 @@ func (a *AzureBackend) Get(ctx context.Context, key string) (*physical.Entry, er
...
-	if !exists {
-		return nil, nil
-	}
 
-	reader, err := blob.Get(nil)
-	if err != nil {
-		return nil, err
-	}
+	// TODO how many retries?
+	reader := res.Body(azblob.RetryReaderOptions{MaxRetryRequests: 3})
```
calvn, line 197: Was the old behavior retrying at all?

physical/azure/azure.go
elsesiy: Okay I'll update it but won't make it configurable for the time being

physical/azure/azure.go
elsesiy: No, it wasn't, see [here](https://github.com/Azure/azure-sdk-for-go/blob/master/storage/blob.go#L585)

physical/azure/azure.go
elsesiy: I looked at the old implementation and it was also returning an error that was swallow by the `DeleteIfExists` convenience function. Hence, I updated the snippet to catch it and return `nil` if we hit the `ServiceCodeBlobNotFound`.

physical/azure/azure.go
calvn: It should be fine to have no retry and signal failure if it fails to perform the operation the first time around, and this looks like the expectation on storage backend operations behavior in general. It feels a bit non-consistent to retry on Get but not for other operations.

elsesiy: Agreed, I removed the retries

elsesiy:
@calvn I believe all comments have been addressed

calvn (approved):
Ran acceptance tests and they passed. The listing test took a bit since it needs to create 5k+ keys on the storage container for that test case.

```
=== RUN   TestAzureBackend
--- PASS: TestAzureBackend (1.88s)
=== RUN   TestAzureBackend_ListPaging
--- PASS: TestAzureBackend_ListPaging (235.04s)
PASS
```

elsesiy:
> Ran acceptance tests and they passed. The listing test took a bit since it needs to create 5k+ keys on the storage container for that test case.
> 
> ```
> === RUN   TestAzureBackend
> --- PASS: TestAzureBackend (1.88s)
> === RUN   TestAzureBackend_ListPaging
> --- PASS: TestAzureBackend_ListPaging (235.04s)
> PASS
> ```

That's great!! Can we merge it then?

PR #9580, opened 2020-07-23
Upgrade to newer okta lib for pagination.  Fetch multiple group resul…

ncabatoff:
…tsets using it.


builtin/credential/okta/backend_test.go
```diff
@@ -13,17 +13,24 @@ import (
...
 	"github.com/hashicorp/vault/sdk/logical"
+	"github.com/okta/okta-sdk-golang/v2/okta"
+	"github.com/okta/okta-sdk-golang/v2/okta/query"
+	"github.com/stretchr/testify/require"
 )
 
 // To run this test, set the following env variables:
 // VAULT_ACC=1
 // OKTA_ORG=dev-219337
 // OKTA_API_TOKEN=<find in 1password>
```
jasonodonnell, line 24: I know it's not part of this PR but it's a bit misleading because the API tokens expire.  We might have a note here that a new one needs to be generated instead of having one in 1Password.

ncabatoff: Updated

builtin/credential/okta/backend_test.go
```diff
@@ -41,6 +48,7 @@ func TestBackend_Config(t *testing.T) {
 	username := os.Getenv("OKTA_USERNAME")
 	password := os.Getenv("OKTA_PASSWORD")
 	token := os.Getenv("OKTA_API_TOKEN")
+	createOktaGroups(t, username, token, os.Getenv("OKTA_ORG"))
```
jasonodonnell, line 51: Should we clean up these groups?  Looking in Okta I see 201 test groups now.

ncabatoff: What's the harm in leaving them?

jasonodonnell: No harm, just think it's good to clean up after a test.

jasonodonnell:
@ncabatoff FYI build errors:
```
1 errors occurred:
--> linux/amd64 error: exit status 1
Stderr: builtin/credential/okta/backend.go:12:2: cannot find package "github.com/okta/okta-sdk-golang/v2/okta" in any of:
	/go/src/github.com/hashicorp/vault/vendor/github.com/okta/okta-sdk-golang/v2/okta (vendor tree)
	/usr/local/go/src/github.com/okta/okta-sdk-golang/v2/okta (from $GOROOT)
	/go/src/github.com/okta/okta-sdk-golang/v2/okta (from $GOPATH)
```

builtin/credential/okta/backend_test.go
```diff
@@ -88,6 +97,57 @@ func TestBackend_Config(t *testing.T) {
...
+
+	users, _, err := client.User.ListUsers(ctx, &query.Params{
+		Q: username,
+	})
+	require.Nil(t, err)
+	require.Len(t, users, 1)
+	userID := users[0].Id
+	var groupIDs []string
+
+	for i := 0; i < 201; i++ {
```
kalafut, line 113: A comment linking to the docs or sdk that define the limit that this is derived from would be good.

builtin/credential/okta/path_config.go
```diff
@@ -310,7 +311,7 @@ func (new *oktaShimOld) Do(req *http.Request, v interface{}) (interface{}, error
 }
 
 // OktaClient creates a basic okta client connection
-func (c *ConfigEntry) OktaClient() (oktaShim, error) {
+func (c *ConfigEntry) OktaClient(ctx context.Context) (oktaShim, context.Context, error) {
```
calvn, line 314: Looks like the call to `OktaClient` doesn't use the returned context, and the context returned here could be accessed by `Client()`. I think we can simply not return that context here.

builtin/credential/okta/backend_test.go
ncabatoff: Done

builtin/credential/okta/path_config.go
ncabatoff: Fixed

PR #9581, opened 2020-07-23
Rate Limit Quota Headers

alexanderbez:
- Replace rate limit lib
- Remove old client/manager logic that is now handled for us by the new lib
- Add support for `enable_response_headers` to quota config
- Add support for custom `interval` (duration) option when creating a rate limit quota

Headers example:

```shell
➜ http GET http://127.0.0.1:8200/v1/sys/quotas/rate-limit/global X-Vault-Token:$VAULT_TOKEN
HTTP/1.1 429 Too Many Requests
Cache-Control: no-store
Content-Length: 88
Content-Type: application/json
Date: Mon, 27 Jul 2020 20:20:29 GMT
Retry-After: Mon, 27 Jul 2020 20:21:26 UTC
X-Ratelimit-Limit: 1
X-Ratelimit-Remaining: 0
X-Ratelimit-Reset: Mon, 27 Jul 2020 20:21:26 UTC

{
    "errors": [
        "request path \"sys/quotas/rate-limit/global\": rate limit quota exceeded"
    ]
}
```

Note, ENT will be handled in a separate PR.


alexanderbez:
Currently having trouble understanding how `reset` time works...looking into it. Otherwise, this is nearly r4r.

vault/quotas/quotas.go
```diff
@@ -177,6 +177,10 @@ type Response struct {
 	// Access is the handle to reach back into the quota rule that processed the
 	// quota request. This may not be set all the time.
 	Access Access
+
+	// Headers defines any optional headers that may be returned by the quota rule
+	// to clients.
+	Headers map[string]string
```
vishalnayak, line 183: I don't think we have the logic to append these headers to the actual API response yet. You might want to add that.

http/util.go
```diff
@@ -60,6 +60,10 @@ func rateLimitQuotaWrapping(handler http.Handler, core *vault.Core) http.Handler
 			return
 		}
 
+		for h, v := range quotaResp.Headers {
```
briankassouf, line 63: IIRC in the RFC we had discussed having headers behind a config flag to prevent information leakage. Do we still want to do so?

vault/quotas/quotas.go
alexanderbez: We do -- see changes to `http/util.go`.

vault/quotas/quotas_rate_limit.go
```diff
@@ -166,36 +130,19 @@ func (rlq *RateLimitQuota) initialize(logger log.Logger, ms *metricsutil.Cluster
...
-	rlq.staleAge = DefaultRateLimitStaleAge
-	rlq.rateQuotas = make(map[string]*ClientRateLimiter)
-
-	if !rlq.purgeEnabled {
-		rlq.purgeEnabled = true
-		rlq.closeCh = make(chan struct{})
-		go rlq.purgeClientsLoop()
+	rlStore, err := memorystore.New(&memorystore.Config{
+		Tokens:        uint64(math.Round(rlq.Rate)), // allow 'rlq.Rate' number of requests per 'Interval'
+		Interval:      time.Second,                  // time interval in which to enforce rate limiting
```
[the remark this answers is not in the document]
alexanderbez: Yeah, good idea. I'll tackle it as part of the PR since it should be straightforward.

http/util.go
alexanderbez: Sure, I'll add a config for this too.

vault/quotas/quotas.go
alexanderbez: Ohh, unless you mean for lease count quotas. No, we do not handle headers for lease count quotas.

vault/logical_system_quotas.go
```diff
@@ -20,6 +20,10 @@ func (b *SystemBackend) quotasPaths() []*framework.Path {
 					Type:        framework.TypeBool,
 					Description: "If set, starts audit logging of requests that get rejected due to rate limit quota rule violations.",
 				},
+				"enable_response_headers": {
```
vishalnayak, line 24: Maybe this is sufficient, but I wonder if we need type specific response header knobs. For example, `enable_rate_limit_response_headers` and `enable_lease_count_response_headers` and so on.

@briankassouf thoughts?

vault/logical_system_quotas.go
```diff
@@ -212,10 +232,11 @@ func (b *SystemBackend) handleRateLimitQuotasRead() framework.OperationFunc {
...
 		data := map[string]interface{}{
-			"type": qType,
-			"name": rlq.Name,
-			"path": nsPath + rlq.MountPath,
-			"rate": rlq.Rate,
+			"type":     qType,
+			"name":     rlq.Name,
+			"path":     nsPath + rlq.MountPath,
+			"rate":     rlq.Rate,
+			"interval": rlq.Interval.String(),
```
vishalnayak, line 239: We may not want to return `<number>s` format over the API. How about returning just the integer?

vault/logical_system_quotas.go
alexanderbez: That is an option, although I can't think of strong reasons why we'd want that, but perhaps flexibility triumphs simplicity?

vault/logical_system_quotas.go
alexanderbez: e.g. `100000` vs `1s`, the later is cleaner to me. Should the output not also be reflective of the input?

vishalnayak: We could return in the unit of second (just `1`) instead of `10000`.

`TypeDurationSecond` accepts both `1s` and `1`. The rationale behind the comment is that with input, we have control over how we parse it, and we do support parsing of both Go style durations and plain integers. Now assume that the input was just `1` then the output with the PR as it stands would still be `1s` which is inconsistent with the input.

Historically we have moved away from returning Go style durations because it puts more onus on output consuming non-go systems to parse it.

alexanderbez: > Now assume that the input was just 1 then the output with the PR as it stands would still be 1s which is inconsistent with the input.

If the user provides `1`, I would imagine this really means `1s`, in which case the output of `1s` would also make sense. Is this not the case? If not, how is the value `1` interpreted?

vishalnayak: Yep

vault/logical_system_quotas.go
vishalnayak: The reason I wondered about having separate knobs for these was that the security implication of revealing the rate limiting information and the lease count information could be very different. I am guessing that some folks may be okay to reveal the rate information but not of leases. WDYT?

alexanderbez: It can't hurt I suppose. It really depends on how many types of resource quotas we plan on supporting in the future. Perhaps we support separate knobs for now. Any objections @briankassouf?

alexanderbez: I'm going to go ahead and apply your suggestion @vishalnayak 👍

PR #9583, opened 2020-07-23
Implement the plugin reload status endpoint as a dummy return for OSS

sgmiller:
And modify the CLI to display error contents.

### DIFF
--- a/api/sys_plugins.go
+++ b/api/sys_plugins.go
@@ -2,8 +2,11 @@ package api
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/vault/sdk/logical"
 	"net/http"
 	"time"
 
@@ -299,6 +302,13 @@ func (c *Sys) ReloadPluginStatus(reloadStatusInput *ReloadPluginStatusInput) (*R
 
 	resp, err := c.c.RawRequestWithContext(ctx, req)
 	if err != nil {
+		defer resp.Body.Close()
+		var er logical.Response
+		dec := json.NewDecoder(resp.Body)
+		err2 := dec.Decode(&er)
+		if err2 == nil && er.Data != nil{
+			err = errwrap.Wrap(err, errors.New(er.Data["error"].(string)))
+		}
 		return nil, err
 	}
 	defer resp.Body.Close()
--- a/command/plugin_reload_status.go
+++ b/command/plugin_reload_status.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"fmt"
+	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/vault/api"
 	"github.com/mitchellh/cli"
 	"github.com/posener/complete"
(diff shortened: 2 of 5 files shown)


api/sys_plugins.go
```diff
@@ -299,6 +302,13 @@ func (c *Sys) ReloadPluginStatus(reloadStatusInput *ReloadPluginStatusInput) (*R
 
 	resp, err := c.c.RawRequestWithContext(ctx, req)
 	if err != nil {
+		defer resp.Body.Close()
```
alexanderbez, line 305: why not just have a single `defer resp.Body.Close()` call?

sgmiller: No reason!  Fixed.

vault/logical_system_helpers.go
```diff
@@ -91,6 +92,10 @@ var (
 		return nil
 	}
 
+	handlePluginReloadStatus = func(ctx context.Context, store logical.Storage, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
```
alexanderbez, line 95: Since the args aren't used...

```suggestion
	handlePluginReloadStatus = func(_ context.Context, _ logical.Storage, req *logical.Request, _ *framework.FieldData) (*logical.Response, error) {
```

command/plugin_reload_status.go
```diff
@@ -75,7 +76,13 @@ func (c *PluginReloadStatusCommand) Run(args []string) int {
 	})
 
 	if err != nil {
-		c.UI.Error(fmt.Sprintf("Error retrieving plugin reload status: %s", err))
+		if w, ok := err.(errwrap.Wrapper); ok {
```
alexanderbez, line 79: Just curious, what was the motivation behind this change? Is the error typically wrapped? Why the need to explicitly check and unwrap?

api/sys_plugins.go
```diff
@@ -298,10 +301,16 @@ func (c *Sys) ReloadPluginStatus(reloadStatusInput *ReloadPluginStatusInput) (*R
 	defer cancelFunc()
 
 	resp, err := c.c.RawRequestWithContext(ctx, req)
+	defer resp.Body.Close()
 	if err != nil {
+		var er logical.Response
```
alexanderbez, line 306: nit:

```suggestion
		var logicalResp logical.Response
```

command/plugin_reload_status.go
sgmiller: The error we want to get at, that it's an enterprise feature, is the inner not outer error.  If you call .Error() on a wrapped error you only get the outer message.

vault/logical_system_helpers.go
sgmiller: TIL

api/sys_plugins.go
```diff
@@ -298,10 +301,16 @@ func (c *Sys) ReloadPluginStatus(reloadStatusInput *ReloadPluginStatusInput) (*R
 	defer cancelFunc()
 
 	resp, err := c.c.RawRequestWithContext(ctx, req)
+	defer resp.Body.Close()
 	if err != nil {
+		var logicalResp logical.Response
+		dec := json.NewDecoder(resp.Body)
+		err2 := dec.Decode(&er)
+		if err2 == nil && er.Data != nil{
```
alexanderbez, line 309:
```suggestion
		if err2 == nil && logicalResp.Data != nil{
```

PR #9588, opened 2020-07-24
Fix navigation for the Sentinel config and clarify module usage

sgmiller:
This hopefully get's the Sentinel stanza docs into the sidebar, and tries to
address any confusion around which modules Vault generally supports and the 
justification for additonal_enabled_modules.

### DIFF
--- a/website/pages/docs/configuration/sentinel.mdx
+++ b/website/pages/docs/configuration/sentinel.mdx
@@ -9,7 +9,8 @@ description: |-
 
 # `sentinel` Stanza
 
-The sentinel stanza specifies configurations for Vault's Sentinel integration.
+The sentinel stanza specifies configurations for Vault's
+[Sentinel](https://docs.hashicorp.com/sentinel/) integration.
 
 ```hcl
 sentinel {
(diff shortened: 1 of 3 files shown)


website/pages/docs/configuration/sentinel.mdx
```diff
@@ -25,12 +26,18 @@ A valid Vault Enterprise license is required for use of Sentinel policies.
...
+- `additional_enabled_modules` `(string array: [])`` - By default Vault supports
+  all of Sentinel's [standard imports](https://docs.hashicorp.com/sentinel/imports/)
+  except "http", which has performance and security implications.
+
+  This parameter specifies a list of modules to allow.  At this time "http"
+  is the only candidate, but in the future other Sentinel modules that may not
+  be included could be overridden here.
+  
+
+  ~> **Warning**: Care should be taken when enabling modules which
```
mladlow, line 38: For some reason when I tested this out the warning didn't format with the fancy box.

sgmiller: Hmm.  How do we test this actually (ie render docs locally)

mladlow: See the `website/README.md`, but you can just `make` in that dir and then see it on `localhost:3000`.

rberlind:
I'm also going to do a PR to change https://docs.hashicorp.com/sentinel/extending/plugins/ to say that custom imports can be used with Nomad and Vault.

sgmiller:
> It's unfortunate that the property is called "additional_enabled_modules" when it should have been called "additional_enabled_imports". Modules in Sentinel are technically something different since they are written in Sentinel itself instead of Golang as is the case for imports (or "import plugins"). https://docs.hashicorp.com/sentinel/extending/internals/#sentinel-s-import-model shows that modules and import plugins can be considered as two types of "imports"; of course, both are integrated into Sentinel policies with the "import" statement. I should have noticed this bigger issue in addition to the doc issue. But let's put that aside for now and focus on the doc. I will submit a separate issue suggesting we rename the setting.
> 
> I have made an edit in this branch suggesting how this should be written. Apologies if you object to my having done it on your branch. Feel free to modify or revert my changes. But I do think they are better.
> 
> Note that in addition to discussing the http import, I added the important note about customers being able to create their own custom import plugins in Golang.

Does Vault support custom sentinel modules though?  I'm not sure it does...

rberlind:
@sgmiller : regarding your comment "Does Vault support custom sentinel modules though? I'm not sure it does...".
That had been my understanding, but I now wonder if it is wrong.  Can you check with engineering and @narayan-iyengar ?

sgmiller:
This feature for certain does not load any module.  It merely removes items from a blocklist in Vault core setup.  So this does not provide a route for adding custom plugins.

PR #9592, opened 2020-07-24
Add new page documenting limits.

mgritter:
Add some identity metrics not previously documented.

:mag: [Netlify preview](https://deploy-preview-9592--vault-www.netlify.app/docs/internals/limits)


website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+
+For the integrated storage backend, the default limit (introduced in
+Vault 1.5.0) is 1 MiB.  This may be configured via `max_entry_size` in
+the [storage stanza](/docs/configuration/storage/raft#max_entry_size).
+
+Many of the other limits within Vault derive from the maximum size of
+a storage entry, as described in the next sections. It is possible to
+recover from an error where a storage entry has reached its maximum
+size by reconfiguring Vault or Consul to a larger maximum storage
+entry.  However, using large storage entries also negatively affects
```
ncabatoff, line 38: Do we want to allude to this being an issue not just in terms of bigger writes to storage locally, but the fact that Raft doesn't like them?  e.g. from the raft repo:

    // Increasing beyond this risks RPC IO taking too long and preventing
    // timely heartbeat signals which are sent in serial in current transports,
    // potentially causing leadership instability.

website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+estimate for entities and groups. The number is slightly less than the
+amount that fits in one shard, to reflect the fact that the first
+shard to fill up will start inducing failures.  This maximum will
+decrease if each entity has a large amount of metadata, or if each
+group has a large number of members.
+
+|   | Consul default (512 KiB) | Integrated storage default (1 MiB) |
+|---|--------------------------|------------------------------------|
+| Maximum number of identity entities (best case, 200 bytes per entity) | ~610,000 | ~1,1250,000 |
+| Maximum number of identity entities (conservative case, 500 bytes per entity) | ~250,000 | ~480,000 |
```
ncabatoff, line 127: I wonder if we should be more pessemistic for the conservative case, since some orgs seem to want to put a lot of metadata into entities?

website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+point limit by the larger of of the number of auth mounts per namespace
+(including `ns_token`) and the number of secret mounts per namespace
+(including `identity` and `sys`.)
+
+### Entity and group limits
+
+The metadata that may be attached to an identity entity or an entity group
+has the following constraints:
+
+|            |      |
```
yhyakuna, line 94: It may look better with table header, I think.

```suggestion
|   Constraints         |   Limit   |
```

website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+### Request size
+
+The maximum size of an HTTP request sent to Vault is limited by
+the `maximum_request_size` option in the [listener stanza](/docs/configuration/listener/tcp). It defaults to 32 MiB.  This value, minus the overhead of
+the HTTP request itself, places an upper bound on any Transit operation,
+and on the maximum size of any key-value secrets.
+
+### Request duration
+
+The default maximum duration of a Vault operation is
```
ncabatoff, line 226: Do we want to remind people that they may have implicit client-side timeouts too, e.g. $VAULT_CLIENT_TIMEOUT?

website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+[`max_request_duration`](/configuration/listener/tcp), which defaults to
+90 seconds. If a particular secret engine takes longer than this to perform an
+operation on a remote service, the Vault client will see a failure.
+
+### Cluster and replication limits
+
```
ncabatoff, line 232: Maybe we could say something here about perf standby limits too?  Probably most people don't realize that they also stream wals.

website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+| Maximum number of DR replicas | no limit, up to active node capability |
+| Maximum number of performance replicas | no limit, up to actcive node capability |
+
+### Lease limits
+
+A systemwide [maximum TTL](/docs/configuration#max_lease_ttl), and a
+[maximum TTL per mount point](/api-docs/system/mounts#max_lease_ttl-1) can be
+configured.
+
+Although no technical maximum exists, high lease counts can cause
```
ncabatoff, line 255: Recommend monitoring num_leases?

website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+Instead, consider using [external groups](/docs/secrets/identity#external-vs-internal-groups) or split the group up into multiple sub-groups.
+
+### Token limits
+
+One storage entry is used per token; there is thus no
+upper bound on the number of active tokens.  There are no restrictions on
+the token metadata field, other than the entire token must fit into one
+storage entry:
+
+|                                       |          |
```
yhyakuna, line 150: Same here... The table may look better with some sort of header.

```suggestion
|            |   General limit   |
```

website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+entity belongs to, and recursively to any parents of that entity.
+Very large numbers of policies are possible, but can cause Vault’s
+response time to increase. You can monitor the
+[`vault.core.fetch_acl_and_token`](/docs/internals/telemetry#core-metrics)
+metric to determine if the time required to assemble an access control list
+is becoming excessive.
+
+### Versioned key-value store (kv-v2 secret engine)
+
+|                                       |          |
```
yhyakuna, line 181: Same here... The table may look better with some sort of header.

```suggestion
|            |   General limit   |
```

website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+or the maximum number of replicas associated with a primary.  However,
+each replica adds considerable overhead to the active node, as
+each write must be duplicated to all standbys. The overhead of
+resyncing multiple replicas at once is also high.
+
+Monitor the active Vault node's CPU and network utilization, as well as
+the lag between the last WAL and replica WAL, to determine if the
+maximum number of replicas has been exceeded.
+
+|                                       |          |
```
yhyakuna, line 243: Same here... The table may look better with some sort of header.

```suggestion
|            |   General limit   |
```

website/pages/docs/internals/limits.mdx
```diff
@@ -0,0 +1,271 @@
...
+A systemwide [maximum TTL](/docs/configuration#max_lease_ttl), and a
+[maximum TTL per mount point](/api-docs/system/mounts#max_lease_ttl-1) can be
+configured.
+
+Although no technical maximum exists, high lease counts can cause
+degradation in system performance.  We recommend short default
+time-to-live values on tokens and leases to avoid a large backlog of
+unexpired leases, or a large number of simultaneous expirations.
+
+|                                       |          |
```
yhyakuna, line 260: Same here... The table may look better with some sort of header.

```suggestion
|            |   General limit   |
```

website/pages/docs/internals/limits.mdx
tdsacilowski: My vote would be to definitely call this out. We don't want customers thinking that it's an "easy fix" to just increase the max storage entry size and the more technical justification we can provide, the better.

website/pages/docs/internals/limits.mdx
tdsacilowski: Maybe in addition to a conservative case, we also include an absolute worst case with maxed-out metadata for entities/groups? Since those are defined constants, it'll give customers an idea of how best-case, middle-ground case, worst-case differ from each other

website/pages/docs/internals/limits.mdx
tdsacilowski: If there are limits that are specific to perf standby that are in addition to what's already captured, my vote is to be explicit about that

website/pages/docs/internals/limits.mdx
mgritter: The numbers for the worst case (40960 bytes of incompressible metadata) are quite bad; I calculate them to be around 670 and 2400. 

The exact threshold depends on what you consider an acceptable probability of failure, as the buckets will fill up according to a binomial distribution.  But I don't have the right statistics math so I've been modelling them as independent, though this is of course incorrect.

website/pages/docs/internals/limits.mdx
mgritter: Added.

website/pages/docs/internals/limits.mdx
mgritter: Added a link.

website/pages/docs/internals/limits.mdx
mgritter: Added (I didnt know this was only 60 seconds!)

website/pages/docs/internals/limits.mdx
mgritter: I added a mention of perf standby; I wish we could be more specific here but there are too many variables.

website/pages/docs/internals/limits.mdx
mgritter: I added just "Limit" as a header.

website/pages/docs/internals/limits.mdx
ncabatoff: You added just what I was looking for.

PR #9598, opened 2020-07-25
Vault status version & storage listing in init too.

### LINKED ISSUE #9248: Vault Status missing Version prior to Init
Vault **does not** correctly **report** it's **version** prior to being initialised and this is also true for all cluster members that are in a similar state prior to being successfully peered. It can be bit of an annoyance when for example provisioning upgrades or introducing new nodes whose versions can only be determined with direct access to the binaries files on the host.

Using a newly configured and running instance (config below) both CLI & API similarly do not output the known Vault version: 

```bash
# // VIA CLI:
vault status
 # Key                      Value
 # ---                      -----
 # Recovery Seal Type       awskms
 # Initialized              false
 # Sealed                   true
 # Total Recovery Shares    0
 # Threshold                0
 # Unseal Progress          0/0
 # Unseal Nonce             n/a
 # Version                  n/a
 # HA Enabled               true

# // VIA API:
vault status -format=json | jq '.'
 # {
 #   "type": "awskms",
 #   "initialized": false,
 #   "sealed": true,
 #   "t": 0,
 #   "n": 0,
 #   "progress": 0,
 #   "nonce": "",
 #   "version": "",
 #   "migration": false,
 #   "recovery_seal": true,
 #   "storage_type": "raft"
 # }

vault operator init ... # …
# // after a successful int all versions now are reported
```

The **expected behaviour** is for **version** to **be reported** and not contigent to initilisation or anything else.

**Environment:**
* Vault CLI Version (retrieve with `vault version`): **`Vault v1.4.2`**
* Server Operating System/Architecture: **`uname -pisorv ; # Linux 4.19.0-9-amd64 #1 SMP Debian 4.19.118-2 (2020-04-29) GNU/Linux `**

```hcl
api_addr = "https://192.168.178.252:8200"
cluster_addr = "https://192.168.178.252:8201"
listener "tcp" {
  address       = "0.0.0.0:8200"
  cluster_address  = "192.168.178.252:8201"
  tls_cert_file = "/home/vagrant/vault2_certificate.crt"
  tls_key_file = "/home/vagrant/vault2_private.key"
}
storage "raft" {
        path            = "/vault/data"
#       node_id         = "vault2"
}
# // seal stanza eg: seal "awkskms" { ...
disable_mlock = true
log_level = "trace"
ui = true
raw_storage_endpoint = true
```

**PS** - May have some relevance to #9185 in so far as response formatting`"na" vs "" # empty string.`

aphorise:
Increased consistency with `vault status` providing version as well as storage type.
Fixes #9248

### DIFF
--- a/command/format.go
+++ b/command/format.go
@@ -335,6 +335,7 @@ func OutputSealStatus(ui cli.Ui, client *api.Client, status *api.SealStatusRespo
 	}
 
 	out = append(out, fmt.Sprintf("Version | %s", status.Version))
+	out = append(out, fmt.Sprintf("Storage Type | %s", status.StorageType))
 
 	if status.ClusterName != "" && status.ClusterID != "" {
 		out = append(out, fmt.Sprintf("Cluster Name | %s", status.ClusterName))
@@ -347,14 +348,14 @@ func OutputSealStatus(ui cli.Ui, client *api.Client, status *api.SealStatusRespo
 	if err != nil && strings.Contains(err.Error(), "Vault is sealed") {
 		leaderStatus = &api.LeaderResponse{HAEnabled: true}
 		err = nil
-	}
-	if err != nil {
+	} else if err != nil {
 		ui.Error(fmt.Sprintf("Error checking leader status: %s", err))
 		return 1
+	} else {
+		// Output if HA is enabled
+		out = append(out, fmt.Sprintf("HA Enabled | %t", leaderStatus.HAEnabled))
 	}
 
-	// Output if HA is enabled
-	out = append(out, fmt.Sprintf("HA Enabled | %t", leaderStatus.HAEnabled))
 	if leaderStatus.HAEnabled {
 		mode := "sealed"
 		if !status.Sealed {
--- a/command/format.go
+++ b/command/format.go
@@ -348,10 +348,12 @@ func OutputSealStatus(ui cli.Ui, client *api.Client, status *api.SealStatusRespo
 	if err != nil && strings.Contains(err.Error(), "Vault is sealed") {
 		leaderStatus = &api.LeaderResponse{HAEnabled: true}
 		err = nil
-	} else if err != nil {
+	} 
+	if err != nil {
 		ui.Error(fmt.Sprintf("Error checking leader status: %s", err))
 		return 1
-	} else {
+	}
+	if err == nil {
 		// Output if HA is enabled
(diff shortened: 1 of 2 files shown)


command/format.go
```diff
@@ -347,14 +348,14 @@ func OutputSealStatus(ui cli.Ui, client *api.Client, status *api.SealStatusRespo
 	if err != nil && strings.Contains(err.Error(), "Vault is sealed") {
 		leaderStatus = &api.LeaderResponse{HAEnabled: true}
 		err = nil
-	}
-	if err != nil {
+	} else if err != nil {
 		ui.Error(fmt.Sprintf("Error checking leader status: %s", err))
 		return 1
+	} else {
```
ncabatoff, line 354: I don't think this should be an else.  The change above to an `else if` is safe but gratuitous: it adds nothing, and any change in a PR that amounts to "I like it this way better" is likely to be contentious, so I would prefer you revert that one too.

ncabatoff:
Thanks for the PR @aphorise!  Small change requested but otherwise looks good.

command/format.go
aphorise: @ncabatoff thx for the review - I can revert it to retaining original (which is probably safer).

In my mind I thought it may be more contextually correct to append that when its applicable outside of any error conditions.

aphorise: @ncabatoff - I reverted back to original also adding an additional `if` in-line with above to only append HA status in the event of no errors.

command/format.go
```diff
@@ -347,14 +348,16 @@ func OutputSealStatus(ui cli.Ui, client *api.Client, status *api.SealStatusRespo
 	if err != nil && strings.Contains(err.Error(), "Vault is sealed") {
 		leaderStatus = &api.LeaderResponse{HAEnabled: true}
 		err = nil
-	}
+	} 
 	if err != nil {
 		ui.Error(fmt.Sprintf("Error checking leader status: %s", err))
 		return 1
 	}
+	if err == nil {
```
ncabatoff, line 356: We already return in the previous `if` when err is non-nil, so this change is again gratuitous.

aphorise: I'm blind and keep failing to register that... sorry its some personal bias and habit of mine with explicit conditioning even if too verbose - will revert back to original which was correct given the aforementioned `return` above. Thx again @ncabatoff

ncabatoff: No worries, we all have our style preferences and it's hard to relax them.

PR #9600, opened 2020-07-27
Increase expiration timeouts on leases to avoid races in NoopBackend

sgmiller:
NoopBackend is primed with a canned Response for many of the unit tests
in expiration_test.go.  If on a slow machine a lease expires, the revocation
will trigger a request to the backend, which may race on the population
of the same Response instance being used by the unit test.


ncabatoff:
Why such a modest increase?  I didn't look very closely, but I didn't see any behaviour in these tests that rely on it being a short expiry.  TestExpiration_RenewToken seems similar in spirit to TestExpiration_Renew and it uses an hour.  

I would go up to at least `1s`.

sgmiller:
Sure, can do.  I just assumed that if this was a rarely failing test, this would make it 2.5x less likely.  But happy to increase further since that doesn't actually slow the test.

PR #9605, opened 2020-07-27
Fix hyperlink incorrectly rendering

UpGado:
Originally this paragraph displayed the hyperlink improperly as raw markdown syntax. This fixes that so that it appears as text with a hyper link.


kitography:
Thanks for this PR.  Unfortunately PRs can't be merged until the CLA is signed by all parties, and this looks to have timed out - it seems to be fixed in all current branches and release branches.

PR #8105, opened 2020-01-07
Fix DB static role credential rotation replication issue

michelvocks:
This PR fixes a bug that prevents Vault performance secondaries and performance standby's to successfully forward a manual static role credential rotation request e.g. `vault write -f database/rotate-role/education`. 

To prevent the client to run in a timeout: https://github.com/hashicorp/vault/blob/31bbb813e7b8250a1e9ffe09ddc81e8f2653b476/http/handler.go#L642 

And the manual static role credential rotation method does not store anything in the backend (which usually triggers a guard check), we have to store a canary to make sure that a guard check is triggered.

PR #9606, opened 2020-07-27
Forward requests for rotate-root on Performance secondary / standbys

catsby:
In the AWS secret backend, rotating root credentials first creates a new key pair before attempting to store the updated key. 

If the rotation request is sent to a performance standby, the new key gets created but saving to Vault state will fail due to the read-only access to state. The request is forwarded to the primary, and at that point the AWS user has 2 keys created; the original it was configured with, and the new key that failed to save. 

When the primary attempts to complete the request, it also tries to create a new key before saving it to state, triggering an AWS error `LimitExceeded: Cannot exceed quota for AccessKeysPerUser: 2`. This then bubbles up in a confusing error message:

```
Errors:
2 errors occurred:
* errors from both primary and secondary; primary error was error calling CreateAccessKey: LimitExceeded: Cannot exceed quota for AccessKeysPerUser: 2
status code: 409, request id: [Request 1234]; secondary errors follow
* error calling CreateAccessKey: LimitExceeded: Cannot exceed quota for AccessKeysPerUser: 2
status code: 409, request id: [Request 1234]
```

In this PR we update the `config/rotate-root` path to use the update `OperationHandler` attribute, and designate this request should be forwarded before even attempting it in the secondary.

An alternate solution would be to delete the key that was created but which failed to save to state. A PR that does just that will likely follow this one, however due to the eventual consistency nature of IAM in AWS I feel it's better to immediately forward the request instead. 

Similar PRs:

- https://github.com/hashicorp/vault/pull/8105
- https://github.com/hashicorp/vault-plugin-secrets-ad/pull/66

### DIFF
--- a/builtin/logical/aws/path_config_rotate_root.go
+++ b/builtin/logical/aws/path_config_rotate_root.go
@@ -14,8 +14,12 @@ import (
 func pathConfigRotateRoot(b *backend) *framework.Path {
 	return &framework.Path{
 		Pattern: "config/rotate-root",
-		Callbacks: map[logical.Operation]framework.OperationFunc{
-			logical.UpdateOperation: b.pathConfigRotateRootUpdate,
+		Operations: map[logical.Operation]framework.OperationHandler{
+			logical.UpdateOperation: &framework.PathOperation{
+				Callback:                    b.pathConfigRotateRootUpdate,
+				ForwardPerformanceStandby:   true,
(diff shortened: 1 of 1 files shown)


catsby:
Backports:

- 1.5.x: https://github.com/hashicorp/vault/pull/9640
- 1.4.x: https://github.com/hashicorp/vault/pull/9639

PR #9609, opened 2020-07-28
pki: use revocationInfo.RevocationTimeUTC when revoking certs with ti…

calvn:
…dy_revoked_certs set to true

This PR changes the tidy logic for the case when [tidy_revoked_certs](https://www.vaultproject.io/api/secret/pki#tidy_revoked_certs) is set to `true`. When performing tidy, we were not taking the revocation time by a `pki/revoke` call into consideration if that certificate was revoked earlier than its NotAfter time, thus leaving dangling certificates in storage that were not cleanup with tidy even with the `tidy_revoked_certs` parameter specified.

### DIFF
--- a/builtin/logical/pki/path_tidy.go
+++ b/builtin/logical/pki/path_tidy.go
@@ -183,7 +183,8 @@ func (b *backend) pathTidyWrite(ctx context.Context, req *logical.Request, d *fr
 					// within the cert itself and the time from the revocation
 					// entry, and perform tidy if either one tells us that the
 					// certificate has already been revoked.
-					if time.Now().After(revokedCert.NotAfter.Add(bufferDuration)) || time.Now().After(revInfo.RevocationTimeUTC.Add(bufferDuration)) {
+					now := time.Now()
+					if now.After(revokedCert.NotAfter.Add(bufferDuration)) || now.After(revInfo.RevocationTimeUTC.Add(bufferDuration)) {
 						if err := req.Storage.Delete(ctx, "revoked/"+serial); err != nil {
 							return errwrap.Wrapf(fmt.Sprintf("error deleting serial %q from revoked list: {{err}}", serial), err)
 						}
--- a/builtin/logical/pki/path_tidy.go
+++ b/builtin/logical/pki/path_tidy.go
@@ -178,7 +178,12 @@ func (b *backend) pathTidyWrite(ctx context.Context, req *logical.Request, d *fr
 						return errwrap.Wrapf(fmt.Sprintf("unable to parse stored revoked certificate with serial %q: {{err}}", serial), err)
 					}
 
-					if time.Now().After(revokedCert.NotAfter.Add(bufferDuration)) {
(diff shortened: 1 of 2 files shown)


builtin/logical/pki/path_tidy.go
```diff
@@ -178,7 +178,12 @@ func (b *backend) pathTidyWrite(ctx context.Context, req *logical.Request, d *fr
 						return errwrap.Wrapf(fmt.Sprintf("unable to parse stored revoked certificate with serial %q: {{err}}", serial), err)
 					}
 
-					if time.Now().After(revokedCert.NotAfter.Add(bufferDuration)) {
+					// Remove the matched certificate entries from revoked/ and
+					// cert/ paths. We compare against both the NotAfter time
+					// within the cert itself and the time from the revocation
+					// entry, and perform tidy if either one tells us that the
+					// certificate has already been revoked.
+					if time.Now().After(revokedCert.NotAfter.Add(bufferDuration)) || time.Now().After(revInfo.RevocationTimeUTC.Add(bufferDuration)) {
```
seriousben, line 186: Should we use `time.UTC().Now()` on this line?

alexanderbez: I don't think the location matters when comparing time objects, but I would recommend that we store `time.Now()` into a variable since we're using it twice.

calvn: Yeah, time comparison via `After` should be based on absolute time so timezone does not really matter in this case. I'll store `time.Now()`, thanks for the suggestion!

alexk-jh:
Maybe I'm mistaken but this looks incorrect to me. First of all, this change also is triggered when using tidy_revocation_list due to the if branch condition:
`if tidyRevokedCerts || tidyRevocationList {`

Second of all, it is not clear to me why someone should ever want to remove revoked certs before they are expired. buildCRL() will no longer know about them, thus the formerly revoked cert becomes effectively unrevoked! Am I missing something? Can you explain what the intention behind this change is?

calvn:
`tidy_revocation_list` is a deprecated parameter which is now a synonym for `tidy_revoked_certs`, so although that conditional may look awkward, it really means that either flag will end up going into the same code block to process revoked certs.

The cert entry removal and its effect on `buildCRL` is indeed odd though. Thanks for letting us know, we'll look into that!

chair6:
This was a good catch, @alexk-jh. On investigation/review, Vault engineering confirmed that CRL creation was not occurring as intended. 

Vault releases today included a fix for this issue, and we've published a security bulletin at https://discuss.hashicorp.com/t/hcsec-2021-09-vault-s-pki-engine-crl-may-exclude-revoked-but-unexpired-certificates-after-tidy/23461/2.  Please let me know if you would like me to update the acknowledgement in the bulletin to refer to you by name.

alexk-jh:
No problem. Luckily a colleague of mine found something in the release notes that made us suspicious so we ran a few tests and found it before upgrading our vault instances.

I can confirm, our testing also showed that it is fixed with 1.7.1

PR #9611, opened 2020-07-28
Backport 1.5.x: Forward requests for rotate-root on Performance secondary / standbys

catsby:
Backport #9606; forward requests to AWS secret backend to rotate root config to primaries


catsby:
The branch name is incorrect here, re-making this to avoid confusion

PR #9625, opened 2020-07-29
Sidebranch: Transform Secret Engine Initial setup

Monkeychip:
Chelsea and I have paired through the initial setup of the Transform Secret Engine.  This PR get's us started.  We are starting with a sidebranch until the initial setup is ready to push towards master.

**In this PR we do the following:**

1. Setup the Transform Adapter and Model so that you can navigate to the first page --listing and then creating a Transformation.

2. Call `createOrUpdate` from the component and bubble it up to the Transform adapter.  We were able to do this following similar patterns for PKI, where we extended the EditRole component.   This is slightly different than how the SSH engine handles requests.  We wanted to stay away from partials and instead use nested components.

3. Create the transform/role model and adapter to populate the "Allowed Roles" `searchSelect` components model attr.  This component searches over a model to populate what a user can search.  The transform/role adapter makes a request to return a list of roles and returns it to the transform/model.

4. In the transform model, define the attrs of each type of form field so that when the `formField` component loops through it know how to display each type.  This is slightly different than how the SSH Secret Engine works, again because we wanted to avoid partials and have opted to not take advantage of openAPI.  Currently, openAPI does not give us enough flexibility to match the designs of this project.  We may come back after the fact add addresses this, but for now, we are moving forward without relying on openAPI to design the forms.

Here's a gif of the work so far.
transform

Another gif where we can create a Transformation (updated after we figured out the correct adapter/serializer pattern) 

create-transform

Here is an image with the `subText` and `subLabel` props.
image


ui/app/components/transform-edit-form.js
```diff
@@ -0,0 +1,20 @@
...
+export default Component.extend({
+  init() {
+    this._super(...arguments);
+    // TODO: don't need confirm?
+    this.set('backendType', 'transform');
+  },
+  store: service(),
+  actions: {
+    // TODO modify the parameters and potentially rename
+    createOrUpdate(type, event) {
```
noelledaley, line 13: you could rename this to `handleCreateOrUpdate` if you want to be clear that this action is an event handler calling the method on the adapter. in ember we see a lot of `handleFoo` where `foo` is either the event (submit, enter) or an action passed down from above.

ui/app/models/transform.js
```diff
@@ -0,0 +1,91 @@
...
+  },
+  {
+    value: 'internal',
+    displayName: 'internal',
+  },
+];
+
+export default DS.Model.extend({
+  // TODO: for now, commenting out openApi info, but keeping here just in case we end up using it.
+  // useOpenAPI: true,
```
noelledaley, line 40: you might've already explored this, but juuust in case: it's possible to combine the attrs from openApi with the ones defined in the model ([example here](https://github.com/hashicorp/vault/blob/24fad4f96177906150a1e417ce35e4ea3fa5547d/ui/app/models/auth-config/github.js#L16-L24)). this also allows you to separate attrs into field groups and control the order since they're rendered in the same order as the array (i think, not totally sure). this would give you the benefit of not having to update the model if/when a backend attribute changes while still giving you the flexibility of ultimately rendering the `transform-edit-component` you've built below. a bit of a hybrid approach, i suppose, but worth considering!

ui/app/components/transform-edit-form.js
Monkeychip: oh, good idea.

PR #8637, opened 2020-03-30
Add user configurable password policies available to secret engines

pcman312:
# Overview
Adds the concept of a Password Policy to Vault. These policies will enable secret engines to allow users to configure how a password should be generated. Several new endpoints under `/sys/` are created for creating/managing password policies, and a connection to plugins has been created so plugins can request a password from an existing policy.

Currently, how passwords are generated are hard coded to a few specific rules ultimately utilizing the [base62](https://github.com/hashicorp/vault/blob/cb178b7e4fd54369331000141fa7820cd94e6f29/sdk/helper/base62/base62.go) package. This has limitations when needing to integrate with systems that have unusual password requirements. For instance, the [SQLCredentialsProducer](https://github.com/hashicorp/vault/blob/942dd1ef9ec7396024866245c852520e16a56b52/sdk/database/helper/credsutil/sql.go#L76-L83) utilizes the base62 package and prepends the password with `A1a-` in order to satisfy the 4 common character sets that some systems require: uppercase, lowercase, numeric, and symbol. However, there are systems that have differing requirements from these that are not supported in the current system.

[Docs PR](https://github.com/hashicorp/vault/pull/8974)

## PR Size
The size of this PR is a bit misleading due to Vendoring & generated code.
Here is a breakdown of the parts of this PR:
<details>
<summary><b>Non-test files:</b> <code>11 files changed, 1007 insertions(+), 5 deletions(-)</code></summary>

Screen Shot 2020-05-06 at 2 00 58 PM
</details>

<details>
<summary><b>Test files:</b> <code>8 files changed, 2537 insertions(+), 4 deletions(-)</code></summary>

Screen Shot 2020-05-06 at 1 46 34 PM
</details>

<details>
<summary><b>Generated/Vendored files:</b> <code>31 files changed, 1505 insertions(+), 383 deletions(-)</code></summary>

Screen Shot 2020-05-06 at 2 03 17 PM
</details>

## Files to Review
- `sdk/helper/random/*` - Random string generation. This is where rules are defined and used. This is also where the passwords are actually generated.
- `sdk/logical/system_view.go` - Updated `SystemView` interface.
- `sdk/plugin/*` - Updated `gRPCSystemView` implementation of the `SystemView` interface
- `vault/dynamic_system_view*` - Updated `dynamicSystemView` implementation of the `SystemView` interface.
- `vault/logical_system*` - Added password policy endpoints.

# Design of Change
## Password Generation
[...]

PR #9641, opened 2020-07-30
[DBPW 1/X] Add Database v5 interface with gRPC client & server

pcman312:
# Overview

Creates a new `Database` interface for use in the combined database engine. This new interface will be v5 of the engine and is designed to support [password policies](https://github.com/hashicorp/vault/pull/8637). It also changes the pattern of the previous interface to a more gRPC style for better future compatibility. This is done so changes to the interface in the future are hopefully less painful.

Nothing is referencing this package yet. That will happen in a subsequent PR

# Files
`sdk/database/newdbplugin/` - New package for keeping all of the v5 database code. **This is a temporary location.** This package will replace the existing `sdk/database/dbplugin` package in a subsequent PR. The existing `sdk/database/dbplugin` package will be renamed to indicate it is deprecated.
`database.go` - New v5 interface & associated request/response objects
`proto/database.proto` - New v5 interface as defined in gRPC
`grpc_client.go` - gRPC client implementation of `proto/database.proto`. This does the conversion between Go types in `database.go` to protobuf-generated types.
`grpc_server.go` - gRPC server implementation of `proto/database.proto`. This does the conversion between protobuf-generated types and Go types in `database.go`.

# Related PRs
[3/X - Plugin handling](https://github.com/hashicorp/vault/pull/9745)


sdk/database/newdbplugin/database.go
```diff
@@ -0,0 +1,155 @@
+package newdbplugin
+
+import (
+	"context"
+	"time"
+)
+
+// Database to manipulate users within an external system (typically a database).
+type Database interface {
+	// Initialize the database plugin. This is the equivalent of a constructor for the
```
tvoran, line 10: `constructor for the` what?

pcman312: Fixed.

sdk/database/newdbplugin/database.go
```diff
@@ -0,0 +1,155 @@
...
+	// Username of the user created within the database.
+	// REQUIRED
+	Username string
+}
+
+// ///////////////////////////////////////////////////////
+// UpdateUser()
+// ///////////////////////////////////////////////////////
+
+type UpdateUserRequest struct {
```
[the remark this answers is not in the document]
pcman312: These are the statements that are currently supported and how they should be mapped to this interface:
`root_rotation_statements` -> `UpdateUserRequest.ChangePassword.Statements` (missing currently - will add)
`creation_statements` -> `NewUserRequest.Statements`
`revocation_statements` -> `DeleteUserRequest.Statements` (missing currently - will add)
`rollback_statements` -> `NewUserRequest.RollbackStatements` (missing currently - will add) - I hate this one because the description is specific to user creation but the name is general
`renew_statements` -> `UpdateUserRequest.ChangeExpiration.Statements`
`rotation_statements` -> `UpdateUserRequest.ChangePassword.Statements`

sdk/database/newdbplugin/database.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+type UpdateUserResponse struct{}
+
+// ///////////////////////////////////////////////////////
+// DeleteUser()
+// ///////////////////////////////////////////////////////
+
+type DeleteUserRequest struct {
+	Username string
```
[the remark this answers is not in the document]
pcman312: See above response re: statement mapping

sdk/database/newdbplugin/database.go
```diff
@@ -0,0 +1,155 @@
...
+// Initialize()
+// ///////////////////////////////////////////////////////
+
+// InitializeRequest contains all information needed to initialize a database plugin.
+type InitializeRequest struct {
+	// Config to initialize the database with. This can include things like connection details,
+	// a "root" username & password, etc. This will not include all configuration items specified
+	// when configuring the database. Some values will be stripped out by the database engine
+	// prior to being passed to the plugin.
+	Config map[string]interface{}
```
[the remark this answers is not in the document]
pcman312: Unfortunately `google.protobuf.Any` isn't really meant as a generic replacement for `interface{}` like we want to use here. Its main use case is to be any protobuf type, not primitive types: [google.protobuf.Any docs](https://pkg.go.dev/google.golang.org/protobuf/types/known/anypb?tab=doc).

If we use the protobuf types in the interface directly this results in a `map[string]Any` config in the request/response objects. This `Any` type will require plugin developers to handle the weirdness of type checking & converting types in their own code rather than having it done in a translation layer automatically. This would result in a lot of copy-pasted code across plugins as well as a poorer user experience especially for non-go developers.

I agree that having a translation layer isn't ideal, but I'd rather have us maintain that translation than put extra onus on plugin developers to understand the `Any` type. Plus I suspect that we won't be updating this interface *that* frequently, though that's just my opinion.

sdk/database/newdbplugin/database.go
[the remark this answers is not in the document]
calvn: It looks like you've added statements to `ChangePassword` and `ChangeExpiration`. I am not concerned that they are not in the parent struct as long as they are accessible through those fields.

sdk/database/newdbplugin/database.go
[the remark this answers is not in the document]
calvn: Looks like statements were added to `DeleteUserRequest` so this is good now.

sdk/database/newdbplugin/grpc_client.go
```diff
@@ -0,0 +1,249 @@
...
+	rpcResp, err := c.client.Initialize(ctx, rpcReq)
+	if err != nil {
+		return InitializeResponse{}, fmt.Errorf("unable to initialize: %s", err.Error())
+	}
+
+	return initRespFromProto(rpcResp)
+}
+
+func initReqToProto(req InitializeRequest) (*proto.InitializeRequest, error) {
+	config, err := json.Marshal(req.Config)
```
calvn, line 41: It seems that json encoding/decoding might not work properly across messages. Consider using [`protojson`](https://godoc.org/google.golang.org/protobuf/encoding/protojson).

pcman312: This won't work here as a drop-in replacement. `protojson.Marshal` takes in a `proto.Message` type, not `interface{}`.

What is the problem for encoding/decoding here? I'm not clear what the issue is.

sdk/database/newdbplugin/grpc_server_test.go
```diff
@@ -137,72 +129,28 @@ func TestCleanNumbers(t *testing.T) {
...
+				"foo": 42.2,
 			},
 			expected: map[string]interface{}{
-				"foo": map[string]interface{}{
-					"bar": float64(42.123),
-				},
+				"foo": 42.2,
 			},
 		},
-		"recursive no numbers": {
```
calvn: Should we keep non-numeric test cases to make sure that conversion only happens when appropriate?

calvn: Nevermind, didn't realize I was looking at a subset of changes.

sdk/database/newdbplugin/database.go
```diff
@@ -0,0 +1,177 @@
...
+// between protobuf types and request/response objects is handled by Vault developers
+// rather than needing to be handled by external plugin developers.
+// ///////////////////////////////////////////////////////////////////////////
+
+// ///////////////////////////////////////////////////////
+// Initialize()
+// ///////////////////////////////////////////////////////
+
+// InitializeRequest contains all information needed to initialize a database plugin.
+type InitializeRequest struct {
```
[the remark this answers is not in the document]
calvn: Discussed this with Brian and even though this was the original direction that we wanted to go with, it might make sense to keep this middle layer around to keep the networking abstracted away from the interface in case we ever decide to move away from gRPC as the networking layer (like we previously did by going from net/rpc to gRPC). There's a bit more overhead by keeping these structs around, but since the conversion is mostly handled by us (the sdk), this shouldn't be an issue.

sdk/database/newdbplugin/proto/database.proto
```diff
@@ -0,0 +1,92 @@
...
+/////////////////
+message UpdateUserRequest {
+    string username = 1;
+    ChangePassword password = 2;
+    ChangeExpiration expiration = 3;
+}
+
+message ChangePassword {
+    string new_password = 1;
+}
```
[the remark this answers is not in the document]
pcman312: Yes it should. Good catch!

sdk/database/newdbplugin/grpc_client_test.go
```diff
@@ -0,0 +1,537 @@
...
+			client: fakeClient{},
+			req: UpdateUserRequest{
+				Username: "user",
+				Expiration: &ChangeExpiration{
+					NewExpiration: time.Now(),
+				},
+			},
+			doneCtx:   runningCtx,
+			assertErr: assertErrNil,
+		},
```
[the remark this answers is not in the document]
pcman312: I don't think it would be valuable since the statements aren't actually used anywhere. The gRPC client is faked, so the statements wouldn't go anywhere.

sdk/database/newdbplugin/grpc_server_test.go
```diff
@@ -0,0 +1,692 @@
...
+				Username: "someuser",
+				Expiration: &proto.ChangeExpiration{
+					NewExpiration: ptypes.TimestampNow(),
+				},
+			},
+			expectedResp: &proto.UpdateUserResponse{},
+			expectErr:    false,
+			expectCode:   codes.OK,
+		},
+	}
```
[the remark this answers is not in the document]
pcman312: I don't think it would be valuable since the statements aren't actually used anywhere. The gRPC client is faked, so the statements wouldn't go anywhere.

sdk/database/newdbplugin/database.go
```diff
@@ -0,0 +1,177 @@
...
+	// UsernameConfig is metadata that can be used to generate a username
+	// within the database plugin
+	UsernameConfig UsernameMetadata
+
+	// Statements is an ordered list of commands to run within the database when
+	// creating a new user. This frequently includes permissions to give the
+	// user or similar actions.
+	Statements Statements
+
+	// RollbackSTatements is an ordered list of commands to run within the database
```
tvoran, line 88:
```suggestion
	// RollbackStatements is an ordered list of commands to run within the database
```

sdk/database/newdbplugin/grpc_client.go
```diff
@@ -0,0 +1,251 @@
...
+		UsernameConfig: &proto.UsernameConfig{
+			DisplayName: req.UsernameConfig.DisplayName,
+			RoleName:    req.UsernameConfig.RoleName,
+		},
+		Statements: &proto.Statements{
+			Commands: req.Statements.Commands,
+		},
+		Password:   req.Password,
+		Expiration: expiration,
+	}
```
tvoran, line 103: What about `RollbackStatements`?

sdk/database/newdbplugin/grpc_server.go
```diff
@@ -0,0 +1,180 @@
...
+		UsernameConfig: UsernameMetadata{
+			DisplayName: req.GetUsernameConfig().GetDisplayName(),
+			RoleName:    req.GetUsernameConfig().GetRoleName(),
+		},
+		Statements: Statements{
+			Commands: req.GetStatements().GetCommands(),
+		},
+		Password:   req.GetPassword(),
+		Expiration: expiration,
+	}
```
tvoran, line 71: What about `RollbackStatements`?

sdk/database/newdbplugin/grpc_server_test.go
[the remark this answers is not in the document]
tvoran: Ok, though I'm wondering if some of the conversion functions from struct <--> proto could use tests, since they seem to be ignoring some of the fields?

sdk/database/newdbplugin/grpc_server.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (g gRPCServer) UpdateUser(ctx context.Context, req *proto.UpdateUserRequest) (*proto.UpdateUserResponse, error) {
+	if req.GetUsername() == "" {
+		return &proto.UpdateUserResponse{}, status.Errorf(codes.InvalidArgument, "no username provided")
+	}
+
+	dbReq, err := getUpdateUserRequest(req)
+	if err != nil {
+		return &proto.UpdateUserResponse{}, status.Errorf(codes.InvalidArgument, err.Error())
```
austingebauer, line 91: Do we want to return `codes.InvalidArgument` for a call to `UpdateUser()` that doesn't change fields of the user? For example if `!hasChange(dbReq)` is true on line 131.

If this results in an eventual http response with a 400 status code to clients, we may want to consider if that's appropriate (usually malformed syntax, etc).

austingebauer: I'm thinking we should do whatever is consistent with other Vault APIs for updating a resource when nothing is actually being changed. I'm also not sure how these error codes map to the eventual http response status code that a client would receive. Something to consider :)

sdk/database/newdbplugin/grpc_server.go
```diff
@@ -0,0 +1,180 @@
...
+	return &proto.UpdateUserResponse{}, nil
+}
+
+func getUpdateUserRequest(req *proto.UpdateUserRequest) (UpdateUserRequest, error) {
+	var password *ChangePassword
+	if req.GetPassword() != nil && req.GetPassword().GetNewPassword() != "" {
+		password = &ChangePassword{
+			NewPassword: req.GetPassword().GetNewPassword(),
+		}
+	}
```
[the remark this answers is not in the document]
pcman312: Done.

sdk/database/newdbplugin/grpc_server.go
```diff
@@ -0,0 +1,180 @@
...
+	return false
+}
+
+func (g gRPCServer) DeleteUser(ctx context.Context, req *proto.DeleteUserRequest) (*proto.DeleteUserResponse, error) {
+	if req.GetUsername() == "" {
+		return &proto.DeleteUserResponse{}, status.Errorf(codes.InvalidArgument, "no username provided")
+	}
+	dbReq := DeleteUserRequest{
+		Username: req.GetUsername(),
+	}
```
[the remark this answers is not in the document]
pcman312: Done.

sdk/database/newdbplugin/database.go
pcman312: Done.

sdk/database/newdbplugin/grpc_client.go
pcman312: Done

sdk/database/newdbplugin/grpc_server.go
pcman312: Done

sdk/database/newdbplugin/database.go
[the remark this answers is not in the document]
pcman312: Done

sdk/database/newdbplugin/database.go
```diff
@@ -0,0 +1,177 @@
...
+// UsernameMetadata is metadata the database plugin can use to generate a username
+type UsernameMetadata struct {
+	DisplayName string
+	RoleName    string
+}
+
+// NewUserResponse returns any information Vault needs to know after creating a new user.
+type NewUserResponse struct {
+	// Username of the user created within the database.
+	// REQUIRED
```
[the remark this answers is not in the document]
pcman312: Done

sdk/database/newdbplugin/database.go
```diff
@@ -0,0 +1,177 @@
...
+type ChangePassword struct {
+	// NewPassword for the user
+	NewPassword string
+
+	// Statements is an ordered list of commands to run within the database
+	// when changing the user's password.
+	Statements Statements
+}
+
+// ChangeExpiration of a give user
```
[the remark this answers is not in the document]
pcman312: Done

sdk/database/newdbplugin/grpc_server.go
pcman312: I think it's appropriate to require a change when requesting to update a user. Since this isn't an "open" endpoint that can be called by anyone but should only be called by Vault itself, I see this as a notification that something in Vault went wrong.

It doesn't result in an HTTP status code since this is a gRPC code, however it is the gRPC equivalent of a 400 code.

austingebauer: That makes sense. I was thinking along the line that this could _eventually_ result in a 400 being returned to an actual Vault client, but that depends on how this gRPC response error is handled as you've mentioned.

No issue. Resolving.

sdk/database/newdbplugin/grpc_server_test.go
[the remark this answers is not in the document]
pcman312: I added a new series of tests in `conversions_test.go` that helps ensure we don't miss any fields.

sdk/database/newdbplugin/grpc_server.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (g gRPCServer) UpdateUser(ctx context.Context, req *proto.UpdateUserRequest) (*proto.UpdateUserResponse, error) {
+	if req.GetUsername() == "" {
+		return &proto.UpdateUserResponse{}, status.Errorf(codes.InvalidArgument, "no username provided")
+	}
+
+	dbReq, err := getUpdateUserRequest(req)
+	if err != nil {
+		return &proto.UpdateUserResponse{}, status.Errorf(codes.InvalidArgument, err.Error())
```
calvn, line 90: Instead of returning an error, we could have `UpdateUser` return a `bool` (or even better, within `UpdateUserResponse` itself) that indicates whether this operation resulted in a no-op or not.

calvn: Maybe this is not quite the proper approach, actually. I see why this check is implemented on the gRPC layer, but OTOH it feels that Vault usually allows for no-op updates to happen so we shouldn't be guarding by default. This could be handled in some other way, but I don't think it should be reflected as an error.

pcman312: I think this is more valuable as an error here because this should only happen if Vault is calling the plugin without any information to update. This only happens if the DB engine itself isn't providing any updated field information. I view this error as a programmer-time error, not a runtime error.

calvn: I misunderstood this, thinking that it was a comparison check against what was already in place, but it's more of an empty check to make sure that password or expiration is provided.

sdk/database/newdbplugin/grpc_server.go
```diff
@@ -0,0 +1,190 @@
...
+	_, err = g.impl.UpdateUser(ctx, dbReq)
+	if err != nil {
+		return &proto.UpdateUserResponse{}, status.Errorf(codes.Internal, "unable to update user: %s", err)
+	}
+	return &proto.UpdateUserResponse{}, nil
+}
+
+func getUpdateUserRequest(req *proto.UpdateUserRequest) (UpdateUserRequest, error) {
+	var password *ChangePassword
+	if req.GetPassword() != nil && req.GetPassword().GetNewPassword() != "" {
```
calvn, line 102: `GetNewPassword()` has a built-in nil check on the receiver so you shouldn't need ` req.GetPassword() != nil` to protect against potential panic, but it's good to be explicit.

PR #9642, opened 2020-07-30
[DBPW 2/X] Create middleware for Database v5

pcman312:
# Overview

Adds middleware for Database v5. This includes the same middleware as Database v4 (tracing, metrics, and error sanitization). The error sanitization has some bug fixes from the v4 version (like not panicking and only accepting a `map[string]string` instead of a `map[string]interface{}`)

Database v4 version: `sdk/database/dbplugin/databasemiddleware.go`

# Prerequisites
- [x] https://github.com/hashicorp/vault/pull/9641 merged
- [x] Change base branch to `master`

# Related PRs
[3/X - Plugin handling](https://github.com/hashicorp/vault/pull/9745)


sdk/database/newdbplugin/middleware.go
```diff
@@ -0,0 +1,273 @@
...
+		metrics.MeasureSince([]string{"database", mw.typeStr, "CreateUser"}, start)
+
+		if err != nil {
+			metrics.IncrCounter([]string{"database", "CreateUser", "error"}, 1)
+			metrics.IncrCounter([]string{"database", mw.typeStr, "CreateUser", "error"}, 1)
+		}
+	}(time.Now())
+
+	metrics.IncrCounter([]string{"database", "CreateUser"}, 1)
+	metrics.IncrCounter([]string{"database", mw.typeStr, "CreateUser"}, 1)
```
tvoran, line 139: I'm curious why these are using "CreateUser" instead of "NewUser"?

sdk/database/newdbplugin/middleware_test.go
```diff
@@ -0,0 +1,484 @@
...
+	"net/url"
+	"reflect"
+	"testing"
+
+	"github.com/hashicorp/go-hclog"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func TestDatabaseErrorSanitizerMiddleware(t *testing.T) {
```
calvn, line 15: Appreciate the addition of these tests!

sdk/database/newdbplugin/middleware.go
```diff
@@ -0,0 +1,273 @@
...
+func (mw DatabaseErrorSanitizerMiddleware) Type() (string, error) {
+	dbType, err := mw.next.Type()
+	return dbType, mw.sanitize(err)
+}
+
+func (mw DatabaseErrorSanitizerMiddleware) Close() (err error) {
+	return mw.sanitize(mw.next.Close())
+}
+
+// sanitize
```
calvn, line 246: Missing full godoc description here?

sdk/database/newdbplugin/middleware.go
```diff
@@ -0,0 +1,273 @@
...
+// DatabaseErrorSanitizerMiddleware wraps an implementation of Databases and
+// sanitizes returned error messages
+type DatabaseErrorSanitizerMiddleware struct {
+	next      Database
+	secretsFn secretsFn
+}
+
+type secretsFn func() map[string]string
+
+func NewDatabaseErrorSanitizerMiddleware(next Database, secrets secretsFn) DatabaseErrorSanitizerMiddleware {
```
calvn, line 210: I think I know why we're returning a concrete object here (not really modifying the object's fields after instantiation) and why we're not dealing with pointer receivers for its methods, but can you elaborate to confirm my assertions?

Since this is a middleware and likely to be passed around and reference as an interface, I am not sure if this can be problematic upstream at some point if the caller expects a pointer value.

sdk/database/newdbplugin/middleware.go
pcman312: Because I missed updating them to `NewUser`

pcman312: The alternative here is to keep it as `CreateUser` in order to keep consistency with the existing metrics.

sdk/database/newdbplugin/middleware_test.go
pcman312: :+1: Yay tests!

sdk/database/newdbplugin/middleware.go
pcman312: Originally copy-pasted from the old version of this middleware but I can flesh it out

sdk/database/newdbplugin/middleware.go
pcman312: You nailed it: the middleware isn't changing after it is initially created. Passing it around under an interface type won't change any behaviors here. All interface types eg `var db Database` are implicitly pointers so you can do things like nil checks, however the underlying object such as `DatabaseErrorSanitizerMiddleware` can be a concrete, non-pointer type without issue. Since the middleware type is never manipulating its own variables in any of its functions this won't cause any problems.

pcman312:
FYI: the force pushing is due to rebasing against the dependent PR #9641. Once that PR is merged & the base branch is updated to `master` the force pushing will cease.

sdk/database/newdbplugin/middleware.go
calvn: Are there any cases where you'd need to type assert on the interface to a typed pointer value followed by a nil check?

pcman312: None that I've found. All of the usages of this use this as the `Database` interface, not as the underlying middleware type.

sdk/database/newdbplugin/middleware.go
tvoran: Ok, I wondered if we needed to keep backwards compatibility with the old metrics, but since the functions are changing I'd say it makes sense to match the metrics with the new function names.

PR #9647, opened 2020-07-31
Ui/transform enable

chelshaw:
With this change, enterprise versions of Vault will show the transform secret engine option when adding a new secret engine. However, it will be disabled unless the appropriate module is licensed: 

enterprise-no-adp

If the correct license exists, transform is selectable and is can be enabled in the same way any other engine is enabled:
enterprise-with-adp


ui/app/components/mount-backend-form.js
```diff
@@ -56,11 +56,14 @@ export default Component.extend({
...
   engines: computed('version.features[]', function() {
+    let mountEngines = [...ENGINES];
     if (this.version.hasFeature('KMIP')) {
-      return ENGINES.concat([KMIP]);
-    } else {
-      return ENGINES;
+      mountEngines.push(KMIP);
     }
+    if (this.get('version.isEnterprise')) {
+      mountEngines.push(TRANSFORM);
```
chelshaw, line 64: As I understand it, KMIP and Transform are enabled under the same module. So, we could treat KMIP the same way (shown for enterprise, but disabled unless the feature exists). However, that wasn't outlined in the designs to I kept it this way for now.

ui/lib/core/addon/components/box-radio.js
```diff
@@ -0,0 +1,33 @@
+/**
```
Monkeychip, line 1: nice! Thanks for creating this.

ui/lib/core/addon/helpers/has-feature.js
```diff
@@ -12,10 +12,12 @@ const FEATURES = [
   'Seal Wrapping',
   'Control Groups',
   'Namespaces',
+  'KMIP',
```
Monkeychip, line 15: Nice!

chelshaw: It looks like it was an oversight, because the places where we check for KMIP prior to this were in the component files, not on the template. But that's a great reminder to double-check that this doesn't have unintended side-effects

chelshaw:
> This looks great. Quick question. If this is a new component, should there be a storybook md file with it?

Sure, I'll make one since Storybook-Docs isn't up and running yet

Issue #9618, opened 2020-07-29
vault is not initialized but is initialized


yoan-adfinis:
**Describe the bug**

A Vault cluster is setup with the Raft storage backend (using the vault-operator). The first node does the init and unseal. The second node, tells me that it is not initialized when the status says the contrary. Hence the vault-unsealer helper is stuck trying to init an already initialized vault.

```console
/ # vault operator init -status
Vault is not initialized
/ # vault status
Key                Value
---                -----
Seal Type          shamir
Initialized        true            <------------------------------ MEH!?
Sealed             true
Total Shares       4
Threshold          2
Unseal Progress    0/2
Unseal Nonce       n/a
Version            1.5.0
HA Enabled         true
```

**To Reproduce**
1. Setup a three node vault cluster with Raft
2. Init and unseal the first one
3. Boot the second one

**Expected behavior**

`vault operator init -status` to return true.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.5.0
* Vault CLI Version (retrieve with `vault version`): 1.5.0
* Server Operating System/Architecture: Linux / Minikube / Kubernetes 1.18 / containerd

Vault server configuration file(s):

```hcl
listener "tcp" {
  address = "0.0.0.0:8200"
  cluster_address = "0.0.0.0:8201"
  tls_cert_file = "/etc/vault/tls/tls.crt"
  tls_key_file  = "/etc/vault/tls/tls.key"
}


storage "raft" {
  path = "/mnt/vault/data"

  retry_join {
    leader_api_addr         = "https://vault-0.vault-internal:8200"
    leader_ca_cert_file     = "/etc/vault/tls/cacert.crt"
    leader_client_cert_file = "/etc/vault/tls/tls.crt"
    leader_client_key_file  = "/etc/vault/tls/tls.key"
  }

  retry_join {
    leader_api_addr         = "https://vault-1.vault-internal:8200"
    leader_ca_cert_file     = "/etc/vault/tls/cacert.crt"
    leader_client_cert_file = "/etc/vault/tls/tls.crt"
    leader_client_key_file  = "/etc/vault/tls/tls.key"
  }

  retry_join {
    leader_api_addr         = "https://vault-2.vault-internal:8200"
    leader_ca_cert_file     = "/etc/vault/tls/cacert.crt"
    leader_client_cert_file = "/etc/vault/tls/tls.crt"
    leader_client_key_file  = "/etc/vault/tls/tls.key"
  }
}

telemetry {
[...]

PR #9652, opened 2020-08-03
Fixed `vault operator init -status` which conflicted with `vault status`

aphorise:
when in raft join during init & not yet unsealed. Resolves: #9618


ncabatoff:
Hi @aphorise,

You had the right idea, so thanks for pointing us in the right direction.  A few tweaks were needed, which I've been working on in #10498.  I'm going to close this PR because it's now redundant.  Thanks again!

PR #9653, opened 2020-08-03
ci/packagespec

samsalisbury:
Adds packagespec integration. See ENGSRV-021 and cloudsearch "packagespec" for additional documentation and how-to guides.

This alters the structure of CI configuration, placing sections originally in `.circleci/config/@config.yml` into their own directories under `.circleci/config/`.

The majority of changes here are generated by the packagespec tool, and as such are not rendered by default.

Example workflows using this code, running in CircleCI (this is a demo 1.4.3 release)

- [Build workflow example](https://app.circleci.com/pipelines/github/hashicorp/vault/11122/workflows/3fb3f1f1-7607-4845-9db7-b4a6328bdb13)
- [Stage workflow example](https://app.circleci.com/pipelines/github/hashicorp/vault-release/8/workflows/e36a2124-fd05-4a8d-a7e9-5bc33e46ab15)
- [Publish workflow example](https://app.circleci.com/pipelines/github/hashicorp/vault-release/10/workflows/1380fa0e-88d8-4c59-9503-ad7f26e60158) (note: this is a "test publish" to the [old staging bucket](http://hc-releases-vault-stage-public.s3-website-us-east-1.amazonaws.com/vault/))


.circleci/config/@config.yml
```diff
@@ -1,52 +1,6 @@
 ---
 version: 2.1
 
-references:
```
samsalisbury, line 4: Note to reviewers: Deleted code in this file has mostly been migrated to separate files nested under the `.circleci/config/ directory`. This change was made to make generating the build-release config file easier.

.circleci/config/commands/@yarn-cache.yml
```diff
@@ -0,0 +1,10 @@
+restore_yarn_cache:
```
samsalisbury, line 1: Note to reviewers: This file contains code migrated from `.circleci/config/@config.yml` See [this comment](https://github.com/hashicorp/vault/pull/9653/files#r465849114) (above)

.circleci/config/executors/@executors.yml
```diff
@@ -0,0 +1,25 @@
+go-machine:
```
samsalisbury, line 1: Note to reviewers: This file contains code migrated from `.circleci/config/@config.yml` See [this comment](https://github.com/hashicorp/vault/pull/9653/files#r465849114) (above)

packages-oss.yml
```diff
@@ -0,0 +1,282 @@
...
+
+    # GO_VERSION is the version of the Go toolchain to use to compile this package.
+    GO_VERSION: 1.14.6
+
+    # YARN_VERSION is the version of Yarn to install for the UI layer.
+    YARN_VERSION: 1.19.1-1
+
+    # BUILD_TAGS are passed as-is to the -tag flag of 'go build'. It should be
+    # a comma or space separated list, depending on the GO_VERSION you are using.
+    BUILD_TAGS: vault
```
ncabatoff, line 47: I don't think comma separated was ever valid, as per https://go-review.googlesource.com/c/go/+/41951/.  Also, why are we specifying the build tag `vault`, which I've never seen before and doesn't seem to be referenced in any build pragmas?

packages-oss.yml
```diff
@@ -0,0 +1,282 @@
...
+
+    # Standard golang environment variables, passed to the 'go build' command.
+    # You can use any standard environment variables here, any that you omit
+    # will be ommitted from the go build command too, meaning to use the system
+    # default in the build container.
+    CGO_ENABLED: 0
+    GO111MODULE: "off"
```
ncabatoff, line 54: I thought we'd want GO111MODULE=on and GOFLAGS=-mod=vendor?

packages-oss.yml
```diff
@@ -0,0 +1,282 @@
...
+      RUN go-bindata-assetfs -o bindata_assetfs.go -pkg http -prefix pkg -modtime 1480000000 -tags ui ./pkg/web_ui/...
+      RUN mkdir -p http && mv bindata_assetfs.go http/
+      RUN goimports -w http/bindata_assetfs.go
+  # Up to now, the layers do not vary between packages. This means the layers above
+  # can be built prior to package builds in order to warm the cache and speed up
+  # builds.
+  #
+  # warm-go-build-vendor-cache does vary based on GOOS, GOARCH and CGO_ENABLED though.
+  # This means we should cache this in CI per package build.
+  - name: warm-go-build-vendor-cache
```
ncabatoff, line 266: What's the purpose of building the vendored packages?

packages-oss.yml
samsalisbury: Comma separated `-tags` support was added recently (in support of making the the `GOFLAGS` env var easier to use). See `go help build` for a recent version of Go; comma-separated is now recommended, at least as of go1.13.12.

packages-oss.yml
samsalisbury: This allows us to populate the build cache with the results of building the vendor packages. Whenever a new build happens that does not touch the vendor directory, the build is greatly speeded up as it only needs to rebuild our code.

packages-oss.yml
samsalisbury: We can change this, but this PR reflects what's currently on master, using GO111MODULE=off (See https://github.com/hashicorp/vault/blob/master/.circleci/config/%40config.yml#L30) I would rather introduce that change as a separate PR.

packages-oss.yml
samsalisbury: Re the `vault` build tag, that came from here: https://github.com/hashicorp/vault/blob/master/scripts/build.sh#L17 although I'm sure it prob isn't necessary really.

packages-oss.yml
ncabatoff: Couldn't we do that for the build as a whole, rather than targeting just vendor/ ?

packages-oss.yml
ncabatoff: I wasn't asking to change it, just trying to understand.  I see that mod=vendor is the default for go 1.14.  If this will be backported we might want to add that flag.

packages-oss.yml
ncabatoff: Yeah, it's not used now, if it ever was.  I'd remove it to avoid confusion.

packages-oss.yml
samsalisbury: In usual circumstances that would work, however the approach packagespec takes is to segregate the source code into discrete chunks. In this case, the layer being built does not contain the application code at all, it just contains the vendor directory. This is done to improve efficiency. Because we know that only the vendor directory is present, we can use the Git commit SHA of the last commit to have touched the `vendor/` tree as part of the cache key. Now, future builds that don't touch `vendor/` can be run in the docker image that already built the code in vendor.

samsalisbury: The structure of the cache reflects the structure of the repository, as defined by the git pathspec `source-include` definitions in each layer. Therefore different branches all just get the right version of the cache for each layer according to the git commit that touched that part of the tree last. (The entire contents of the rendered layer Dockerfile as well as the `source-include/source-exclude` is also taken into account in cache keys, of course.)

Upcoming plans for more intensive testing of our builds / performing release builds much more frequently than we do now should benefit significantly from this more granular caching.

packages-oss.yml
samsalisbury: Done

samsalisbury: FYI this was backported to the release/1.2.x branch as well.

PR #9667, opened 2020-08-05
Rate Limit Backoff

alexanderbez:
Add support for an optional `block` (nomenclature subject to change) field in rate limit quotas, where if set, will block clients from any further requests until that `block` interval has passed.


vault/logical_system_quotas.go
```diff
@@ -73,6 +73,11 @@ The 'rate' must be positive.`,
 					Type:        framework.TypeDurationSecond,
 					Description: "The duration to enforce rate limiting for (default '1s').",
 				},
+				"block": {
```
vishalnayak, line 76: This is feeling like a boolean knob. How about `block_interval`? That would indicate that it is a duration and consistent with the other param `interval`.

vault/quotas/quotas_rate_limit_test.go
```diff
@@ -116,3 +119,79 @@ func TestRateLimitQuota_Allow(t *testing.T) {
...
+		numFail := cr.atomicNumFail.Load()
+
+		// Since blocking is enabled, each client should only have 'rate' successful
+		// requests, whereas all subsequent requests fail.
+		require.Equal(t, int32(17), numAllow)
+		require.NotZero(t, numFail)
+	}
+
+	// allow enough time for all blocked clients to be purged
+	time.Sleep(rlq.purgeInterval * 2)
```
mgritter, line 192: I wonder if there's some way to speed this up so the test does not take 20 seconds.

I wrote some code to fake time.Now() and tickers in the gauge collection library, and Nick suggested I move that into a common library for re-use.

Not a blocker, though.

vault/quotas/quotas_rate_limit.go
```diff
@@ -181,24 +256,59 @@ func (rlq *RateLimitQuota) allow(req *Request) (Response, error) {
...
+
+	rlq.lock.Lock()
+	defer rlq.lock.Unlock()
+
+	// Check if the client is currently blocked and if so, deny the request. Note,
+	// we cannot simply rely on the presence of the client in the map as the timing
+	// of purging blocked clients may not yield a false negative. In other words,
+	// a client may no longer be considered blocked whereas the purging interval
+	// has yet to run.
+	if blockedAt, ok := rlq.blockedClients[req.ClientAddress]; ok {
```
mgritter, line 276: Using a sync.Map would mean not every request was contending for the same lock here, too.  We would only need rlq.lock further down, to check after we were disallowed.

(I think that's the case anyway-- memorystore.Store does its own locking, correct? And we don't need the mutex to access rlq.store?)

vault/quotas/quotas_rate_limit_test.go
alexanderbez: We just need to wait for the garbage collector to run. A multiple of two was added here for safety, but we can omit that entirely I think as just rely on a ticker that checks every second or timesout at `purgeInterval`. I've updated the test to reflect this. If you think there is a cleaner way, please lmk 👍

vault/quotas/quotas_rate_limit.go
alexanderbez: Correct, we don't need a lock for the `store`, the rate limiter is already thread-safe. See response in the first comment.

vault/quotas/quotas_rate_limit.go
```diff
@@ -150,10 +165,68 @@ func (rlq *RateLimitQuota) initialize(logger log.Logger, ms *metricsutil.Cluster
...
+// in which we stop the ticker and return.
+func (rlq *RateLimitQuota) purgeBlockedClients() {
+	rlq.lock.RLock()
+	ticker := time.NewTicker(rlq.purgeInterval)
+	rlq.lock.RUnlock()
+
+	for {
+		select {
+		case t := <-ticker.C:
+			rlq.lock.Lock()
```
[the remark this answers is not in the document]
alexanderbez: Technically, it could be unbounded, but I can't imagine it'll be that large in practice. We could use a `sync.Map`, but doesn't that come with relatively significant performance bottlenecks. I have no strong opinion here.

vault/quotas/quotas_rate_limit_test.go
```diff
@@ -188,10 +188,16 @@ func TestRateLimitQuota_Allow_WithBlock(t *testing.T) {
...
-	// allow enough time for all blocked clients to be purged
-	time.Sleep(rlq.purgeInterval * 2)
-
-	for addr := range results {
-		require.False(t, rlq.hasBlockedClient(addr), "expected blocked client to be purged")
-	}
+	func() {
+		for {
+			select {
+			case <-time.After(rlq.purgeInterval):
```
mgritter, line 194: I don't really understand whether this works, sorry.

If we get a new `time.Tick(time.Second)` every time through the for loop, then don't we also get a new `time.After(rlq.purgeInterval)` too, pushing the deadline back every time?

I have not used `case <-time.Tick(time.Second):` as an idiom before.

alexanderbez: This will tick every second checking attempting to see if all clients have been garbage collected or it will timeout.

vault/quotas/quotas_rate_limit.go
```diff
@@ -181,24 +255,59 @@ func (rlq *RateLimitQuota) allow(req *Request) (Response, error) {
...
 	}
 
-	if !resp.Allowed {
-		resp.Headers[httplimit.HeaderRetryAfter] = resp.Headers[httplimit.HeaderRateLimitReset]
-		rlq.metricSink.IncrCounterWithLabels([]string{"quota", "rate_limit", "violation"}, 1, []metrics.Label{{"name", rlq.Name}})
+	limit, remaining, reset, allow := rlq.store.Take(req.ClientAddress)
+	resp.Allowed = allow
+	resp.Headers[httplimit.HeaderRateLimitLimit] = strconv.FormatUint(limit, 10)
+	resp.Headers[httplimit.HeaderRateLimitRemaining] = strconv.FormatUint(remaining, 10)
+	resp.Headers[httplimit.HeaderRateLimitReset] = time.Unix(0, int64(reset)).UTC().Format(time.RFC1123)
```
vishalnayak, line 291: RateLimitReset should also be duration in seconds. Sorry for not noticing it in the previous PRs.

vault/quotas/quotas_rate_limit.go
```diff
@@ -181,24 +255,59 @@ func (rlq *RateLimitQuota) allow(req *Request) (Response, error) {
...
+	// a client may no longer be considered blocked whereas the purging interval
+	// has yet to run.
+	if blockedAt, ok := rlq.blockedClients[req.ClientAddress]; ok {
+		if time.Since(blockedAt) >= rlq.BlockInterval {
+			// allow the request and remove the blocked client
+			delete(rlq.blockedClients, req.ClientAddress)
+		} else {
+			// deny the request and return early
+			resp.Allowed = false
+			retryAfter = blockedAt.Add(rlq.BlockInterval).UTC().Format(time.RFC1123)
```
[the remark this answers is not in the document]
alexanderbez: Yup, not a problem. This was modeled after the rate limiter library FYI.

vault/quotas/quotas_rate_limit_test.go
mgritter: Sure, but after you tick for one second, doesn't time.After(rlq.purgeInterval) also get re-run, extending the timeout?

vault/quotas/quotas_rate_limit.go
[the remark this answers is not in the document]
mgritter: I don't know for sure-- we could measure it-- but my guess is that it would be less of a bottleneck than every request covered by a quota contending for the same lock.

However, making that a RLock instead of a Lock might be sufficient, I'll put a comment on that? But that introduces the same problem because we need to switch to a write-lock in order to update blockedClients.

vault/quotas/quotas_rate_limit.go
```diff
@@ -181,24 +255,59 @@ func (rlq *RateLimitQuota) allow(req *Request) (Response, error) {
...
+	var retryAfter string
+
+	defer func() {
+		if !resp.Allowed {
+			resp.Headers[httplimit.HeaderRetryAfter] = retryAfter
+			rlq.metricSink.IncrCounterWithLabels([]string{"quota", "rate_limit", "violation"}, 1, []metrics.Label{{"name", rlq.Name}})
+		}
+	}()
+
+	rlq.lock.Lock()
```
mgritter, line 267: I think it's important to make this a read lock instead of a write lock, because quite a lot of different requests could be contending on it.  But then you have to acquire a write-lock to modify rlq.blockedClients.  Making it a sync.Map instead would remove that as a problem.

vault/quotas/quotas_rate_limit.go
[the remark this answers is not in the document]
alexanderbez: Actually, to do this, we'll have to perform a manual `now - reset`, which will technically produce a slightly off delta if the interval is small (e.g. a few seconds or less). Do you still want to take this approach?

vault/quotas/quotas_rate_limit.go
alexanderbez: So after making those tweaks, `sync.Map` and the tweaked version perform nearly identical in benchmarks (1000 unique clients for executed for 1m). Being that `sync.Map` requires less LOC, I'll go that route.

vault/quotas/quotas_rate_limit.go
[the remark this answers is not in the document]
vishalnayak: I don't have a strong opinion. I wonder if the library took the time approach for this exact reason. We may want to check what the industry "standard" is and follow that. If there isn't one, then we could very well go with timestamp.

vault/quotas/quotas_rate_limit_test.go
alexanderbez: I've updated the function to create variables for the timeout and ticker. I've tested this by explicitly always returning non-zero in `numBlockedClients`. Feel free to unresolve this if you wish 👍

website/pages/docs/concepts/resource-quotas.mdx
```diff
@@ -40,6 +40,10 @@ rate limit quota, and a rate limit quota defined for a mount takes precedence ov
 the global and namespace rate limit quotas. In other words, the most specific
 quota rule will be applied.
 
+A rate limit can be created with an optional `block_interval`, such that when set
+to a non-zero value, any client that hits a rate limit threshold will be blocked
+from all subsequent requests for a duration of `block` seconds.
```
vishalnayak, line 45:
```suggestion
from all subsequent requests for a duration of `block_interval` seconds.
```

vault/quotas/quotas_rate_limit.go
alexanderbez: Updated.

vault/quotas/quotas_rate_limit.go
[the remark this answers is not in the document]
alexanderbez: Checked the RFC and delta seconds is the preferred option. I've updated the headers to reflect that.

PR #9668, opened 2020-08-05
Ui/create edit transformation fixes

Monkeychip:
This PR:

- fixes the create transformation issue with the `stringArray` error
- conditionally shows with the `tweak_source` or `masking field` based on the type selected.
- serializes the masking data such that instead of displaying a character code, which is what is returned by the API, we display the character (e.g. we now show * and not 42).
- Starts to layout the "show" display of a transformation.  My next PR will finish this.
- Fixes issue with template vs templates on the POST vs GET requests, respectfully.  These will likely be changed again once the BE fixes the issue.

Here is a gif run through:

transformation


ui/app/serializers/transform.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default ApplicationSerializer.extend({
+  normalizeResponse(store, primaryModelClass, payload, id, requestType) {
+    if (payload.data.masking_character) {
+      payload.data.masking_character = String.fromCharCode(payload.data.masking_character);
+    }
+
+    if (payload.data.templates && payload.data.templates.length === 1) {
+      // add space after comma in returned array length one of a string of templates
+      payload.data.templates[0] = payload.data.templates[0].replace(/,/g, ', ');
```
chelshaw, line 11: If we're adding a comma for display purposes I would expect the controller to add that in, not the serializer. I'm also a little confused why we're adding it only to the first template? Maybe an example of the before and after here would be helpful.

ui/app/styles/core/forms.scss
```diff
@@ -21,7 +21,7 @@ label {
 .b-checkbox .is-label {
   color: $grey-darker;
   display: inline-block;
-  font-size: $size-small;
+  font-size: 14px;
```
chelshaw, line 24: Why are we deviating from the variables here?

ui/app/styles/core/forms.scss
```diff
@@ -73,6 +73,7 @@ label {
 .sub-text {
   color: $grey;
   margin-bottom: 0.25rem;
+  font-size: 12px;
```
chelshaw, line 76: Here too. If the designs don't match the sizing variables we have we need to revisit with the design team

ui/app/templates/components/transform-edit-form.hbs
```diff
@@ -10,7 +10,7 @@
         @model={{model}}
       />
     {{/each}}
-    <FormFieldGroups @model={{model}} @mode="create" />
+    {{!-- <FormFieldGroups @model={{model}} @mode="create" /> --}}
```
chelshaw, line 13: We can probably go ahead and remove this

ui/app/templates/components/transform-show-transformation.hbs
```diff
@@ -0,0 +1,44 @@
...
+  <div class="has-bottom-margin-s">
+    <h2 class="title is-6">Encode</h2>
+    <div class="has-bottom-margin-s">
+      <span class="helper-text has-text-grey">
+        To test the encoding capability of your transformation, use the following command. It will output an encoded_value.
+      </span>
+    </div>
+    <div class="copy-text level">
+      <code>vault write transform/encode/payments value=&lt;enter your value here&gt;</code>
+      <CopyButton class="button is-transparent level-right" @clipboardText='blahblah'
```
chelshaw, line 23: The `@clipboardText` should probably be the command above, which makes me think this is a perfect use case for the handlebars {{let}} helper

ui/app/templates/components/transform-show-transformation.hbs
```diff
@@ -0,0 +1,44 @@
...
+  <div>
+    <h2 class="title is-6">Decode</h2>
+    <div class="has-bottom-margin-s">
+      <span class="helper-text has-text-grey">
+        To test decoding capability of your transformation, use the encoded_value in the following command. It should return your original input.
+      </span>
+    </div>
+    <div class="copy-text level">
+      <code>vault write transform/decode/payments value=&lt;enter your value here&gt;</code>
+      <CopyButton class="button is-transparent level-right" @clipboardText='blahblah'
```
chelshaw, line 38: Here too! :D

ui/app/templates/components/transformation-edit.hbs
```diff
@@ -24,34 +24,19 @@
...
           @buttonClasses="toolbar-link"
           @onConfirmAction={{action "delete"}}
+          @confirmTitle="Are you sure?"
+          @confirmMessage="This transformation is not in use by a role and can be deleted."
         >
-          Delete role
+          Delete transformation
         </ConfirmAction>
       {{/if}}
       {{#if (or model.canUpdate model.canDelete)}}
```
chelshaw, line 42: Someone with only delete capabilities can edit the transformation, is that right?

ui/app/templates/partials/secret-list/transform-transformation-item.hbs
```diff
@@ -0,0 +1,62 @@
+{{#linked-block
+  (concat
```
chelshaw, line 2: Probably don't need concat if it's all magic strings

ui/app/templates/partials/secret-list/transform-transformation-item.hbs
```diff
@@ -0,0 +1,62 @@
...
+      <nav class="menu">
+        <ul class="menu-list">
+          {{#if (or item.versionPath.isLoading item.secretPath.isLoading)}}
+            <li class="action">
+              <button disabled type="button" class="link button is-loading is-transparent">
+                loading
+              </button>
+            </li>
+          {{else}}
+            {{#if item.updatePath.canRead}}
```
chelshaw, line 36: I think you followed the general item pattern here so not a blocker, but if we check `canRead` for the details dropdown link, should we check it for the main item link as well?

ui/app/serializers/transform.js
Monkeychip: Let's discuss during our sync.  This was a request by design, but becomes a problem with Roles because the data is structured differently.  And yeah, I originally thought of using a serializer because it was changing the data, but it's more of a display thing.  It should go in a controller.

ui/app/styles/core/forms.scss
Monkeychip: Yep, we're deviating.  I walked through Ivana on this as I noticed her designs were different than what we currently have and. she asked me to update all.

ui/app/styles/core/forms.scss
Monkeychip: See comment above.

ui/app/templates/components/transform-show-transformation.hbs
Monkeychip: You may have to walk me through that one.  Let's touch base in the sync later today

ui/app/templates/components/transformation-edit.hbs
Monkeychip: Ah, good catch. I was using a pattern from the template earlier, but they to have canUpdate in order to edit, not canDelete.  I'll amend.

ui/app/templates/partials/secret-list/transform-transformation-item.hbs
Monkeychip: Good catch, another carry over from copying from a different template.

ui/app/serializers/transform.js
Monkeychip: Removing for now based on discussion

ui/app/templates/components/transform-show-transformation.hbs
Monkeychip: That's sweet!!

Issue #9306, opened 2020-06-24
Vault agent not setting exit code to 1 on templating error


omerlh:
**Describe the bug**
I had a bug with the template I am passing to the vault agent, running as a Kubernetes init container. It took my some time to figure there was an issue because the agent existed without errors although it failed to parse the template:
```
==> Vault server started! Log data will stream in below:

==> Vault agent configuration:

                     Cgo: disabled
               Log Level: info
                 Version: Vault v1.4.2

2020-06-24T05:23:17.399Z [INFO]  sink.file: creating file sink
2020-06-24T05:23:17.400Z [INFO]  sink.file: file sink configured: path=/home/vault/.vault-token mode=-rw-r-----
2020-06-24T05:23:17.402Z [INFO]  template.server: starting template server
2020/06/24 05:23:17.402920 [INFO] (runner) creating new runner (dry: false, once: false)
2020/06/24 05:23:17.403710 [INFO] (runner) creating watcher
2020-06-24T05:23:17.403Z [INFO]  auth.handler: starting auth handler
2020-06-24T05:23:17.403Z [INFO]  auth.handler: authenticating
2020-06-24T05:23:17.404Z [INFO]  sink.server: starting sink server
2020-06-24T05:23:17.918Z [INFO]  auth.handler: authentication successful, sending token to sinks
2020-06-24T05:23:17.918Z [INFO]  auth.handler: starting renewal process
2020-06-24T05:23:17.919Z [INFO]  sink.file: token written: path=/home/vault/.vault-token
2020-06-24T05:23:17.919Z [INFO]  sink.server: sink server stopped
2020-06-24T05:23:17.919Z [INFO]  sinks finished, exiting
2020-06-24T05:23:17.919Z [INFO]  template.server: template server received new token
2020/06/24 05:23:17.919867 [INFO] (runner) stopping
2020/06/24 05:23:17.920087 [INFO] (runner) creating new runner (dry: false, once: false)
2020/06/24 05:23:17.920450 [INFO] (runner) creating watcher
2020/06/24 05:23:17.920516 [INFO] (runner) starting
2020-06-24T05:23:17.920Z [ERROR] template.server: template server error: error="(dynamic): parse: template: :4: bad character U+002D '-'"
2020-06-24T05:23:17.920Z [INFO]  template.server: template server stopped
```

**To Reproduce**
1. Use the following cronjob manifest:
```yaml
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: test
[...]

PR #9670, opened 2020-08-05
agent: return a non-zero exit code on error

calvn:
This PR introduces changes around agent to properly return a non-zero exit code on error. Particularly, we can now return a proper error if agent template fails to render any templates due to a non-existent secret path or a non-existent key within a secret path. Using `error_on_missing_key` ([same as the config](https://github.com/hashicorp/consul-template#secret) within consul-template) on any of the templates stanza will result in agent returning immediately if a template-related error is encountered. 

In order for agent to properly exit for cases such as an error due to a non-existent secret path, the current behavior of indefinite retry is changed to exit after it's max [default retry of 12 attempts](https://github.com/hashicorp/consul-template#configuration-file-format) (with an exponential backoff of 250ms increase per try, up to 1m). This still results in a reasonable number of attempts before agent gives up and exits.

I also took the opportunity to eliminate possible panics by changing them to errors since we can now gracefully handle errors returned by the subsystems' Run function within the agent's main Run function.

Fixes https://github.com/hashicorp/vault/issues/9306


command/agent/template/template.go
```diff
@@ -142,6 +151,14 @@ func (ts *Server) Run(ctx context.Context, incoming chan string, templates []*ct
 						Token: latestToken,
 					},
 				}
+
+				// If we're testing, limit retries to 3 attempts to avoid
```
pcman312, line 155: Instead of putting test-only logic, what about specifying something like `retryAttempts` and then doing a conditional off of that?

```go
if ts.retryAttempts >= 0 {
  ctv.Vault.Retry = &ctconfig.RetryConfig{Attempts: &ts.retryAttempts}
}
```

That way it's more flexible and isn't a test-only path in production code.

command/agent/template/template_test.go
```diff
@@ -127,13 +160,27 @@ func TestServerRun(t *testing.T) {
...
+			errCh := make(chan error, 1)
+			go server.Run(ctx, templateTokenCh, templatesToRender, errCh)
 
 			// send a dummy value to trigger the internal Runner to query for secret
 			// info
 			templateTokenCh <- "test"
-			<-server.DoneCh
+
+			select {
+			case <-server.DoneCh:
```
pcman312, line 173: Can `DoneCh` be closed/have a value at the same time the error channel has one? Could this cause flakiness?

command/agent/cache_end_to_end_test.go
```diff
@@ -199,9 +199,14 @@ func TestCache_UsingAutoAuthToken(t *testing.T) {
...
 	}
 	ah := auth.NewAuthHandler(ahConfig)
-	go ah.Run(ctx, am)
+	errCh := make(chan error, 2)
+	go ah.Run(ctx, am, errCh)
 	defer func() {
-		<-ah.DoneCh
+		select {
+		case <-ah.DoneCh:
+		case err := <-errCh:
```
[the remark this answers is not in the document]
calvn: We're intentionally blocking until we receive or close on either ah.DoneCh or errCh in here (the former which gets closed if the timer cancels the context). Having a default case or closing the errCh would mean that we would always return immediately which is not what we want if we're waiting for tasks to finish.

command/agent/template/template.go
calvn: Yep, I made this change earlier but forgot to push.

command/agent/template/template_test.go
calvn: I don't think so since the error is sent back before the DoneCh is closed on a defer.

command/agent/auth/auth.go
```diff
@@ -74,9 +75,12 @@ func backoffOrQuit(ctx context.Context, backoff time.Duration) {
 	}
 }
 
-func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) {
+func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod, errCh chan<- error) {
 	if am == nil {
-		panic("nil auth method")
+		err := errors.New("nil auth method")
+		ah.logger.Error("auth handler internal error", "error", err)
```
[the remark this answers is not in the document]
calvn: Emitting the error in `AgentCommand.Run` means that you lose the context around who emitted the error since loggers are named and the initial description in the log message can provide richer information around where the error happened. 

I wonder if it's better to simply not call `c.UI.Error` in `AgentCommand`.

command/agent.go
```diff
@@ -559,9 +560,11 @@ func (c *AgentCommand) Run(args []string) int {
 		})
 		tsDoneCh = ts.DoneCh
 
-		go ah.Run(ctx, method)
-		go ss.Run(ctx, ah.OutputCh, sinks)
-		go ts.Run(ctx, ah.TemplateTokenCh, config.Templates)
+		errCh = make(chan error, 3)
```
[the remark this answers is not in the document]
calvn: I've changed this to be unbuffered now.

command/agent/auth/auth.go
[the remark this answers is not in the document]
calvn: I wanted to re-visit this, and I think it makes sense for fatal errors (i.e. errors that would cause the program to terminate) to be emitted directly out for the UI output. This mainly affects template's `Server.Run` log output on previous errors that were being emitted through its logger but no longer do so, but this might be fine.

command/agent/template/template_test.go
briankassouf: I think @pcman312 is right, if there are values on both `DoneCh` and `errCh` then go will randomly choose one of them. You can't be sure that select will choose the expected thing here.

command/agent.go
```diff
@@ -593,6 +594,9 @@ func (c *AgentCommand) Run(args []string) int {
 	}()
 
 	select {
+	case err := <-errCh:
+		c.UI.Error(fmt.Sprintf("Error encountered: %s", err))
```
ncabatoff, line 598: I think it's a mistake to emit runtime errors (as opposed to startup errors) this way.  It makes sense to use UI when errors occur at startup, e.g. bad arguments or config, because typically that happens when a user is experimenting manually.  Runtime errors, on the other hand, will normally not be encountered by the user.  They should go into a log, formatted as a log entry, and be tagged as errors such that log scrapers and aggregators will report them appropriately.

calvn: That's fair. It was suggested that error emitted this way to be logged as a UI error, and I did observe this pattern on non-zero exit code runs. Would you suggest to still return a generic error in the UI output?

ncabatoff: I'm okay with that I guess, though I think it sets a bad precedent.  If we start logging *and* emitting UI errors here, then subsequently whenever a dev wants to log an error they're going to ask themselves whether they also need to do a UI output.

calvn: Updated in child PR #9761, lmk what you think.

command/agent/template/template.go
```diff
@@ -130,36 +142,49 @@ func (ts *Server) Run(ctx context.Context, incoming chan string, templates []*ct
...
 					Vault: &ctconfig.VaultConfig{
 						Token: latestToken,
 					},
 				}
+
+				// If we're testing, limit retries to 3 attempts to avoid
+				// long test runs from exponential back-offs
+				if ts.testingLimitRetry != 0 {
+					ctv.Vault.Retry = &ctconfig.RetryConfig{Attempts: &ts.testingLimitRetry}
+				}
```
pcman312, line 171: I don't think you need the conditional:
https://github.com/hashicorp/consul-template/blob/7a4683109607e642f014f29898b1acc797efb890/config/retry.go#L46
https://github.com/hashicorp/consul-template/blob/master/config/vault.go#L196-L198

```suggestion
				ctv.Vault.Retry = &ctconfig.RetryConfig{Attempts: &ts.testingLimitRetry}
```

pcman312 (commented):
I realize this isn't part of your PR, but I was looking to see where `ctx` was set in the end to end tests and [came across this](https://github.com/hashicorp/vault/blob/master/command/agent/alicloud_end_to_end_test.go#L69-L73):
```go
ctx, cancelFunc := context.WithCancel(context.Background())
timer := time.AfterFunc(30*time.Second, func() {
	cancelFunc()
})
defer timer.Stop()
```

Could you update this to:
```go
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
```

(also in the other end-to-end tests, not just the one I linked above)

Thanks!

command/agent/template/template.go
calvn: The default value prior to this change is based off `DefaultRetryAttempts`, which is 12 retries and set by [Finalize](https://github.com/hashicorp/consul-template/blob/7a4683109607e642f014f29898b1acc797efb890/config/retry.go#L137). We don't want to change this to be unlimited by default.

jasonodonnell (approved):
LGTM and works as described in K8s.

pcman312 (approved):

PR #9674, opened 2020-08-06
Trail of bits 018

sgmiller:
This addresses two items from Trail-of-Bits 018.  First, we use a different
scalar multiplication for the DH params that returns an error if a low-order
curve is used producing an all-zero key result.

Second, we introduce an optional config flag for Vault Agent, "derive_key",
which additionally applies the HKDF-SHA256 key derivation function to the
shared secret and both public keys.


helper/dhutil/dhutil.go
```diff
@@ -36,21 +39,47 @@ func GeneratePublicPrivateKey() ([]byte, []byte, error) {
...
+}
+
+// DeriveSharedKey uses HKDF to derive a key from a shared secret and public keys
+func DeriveSharedKey(secret, ourPublic, theirPublic []byte) ([]byte, error) {
+	// Derive the final key from the HKDF of the secret and public keys.
+
+	//These must be consistently ordered we're on "our" side of key negotiation or the other.
+	var pub1 []byte
+	var pub2 []byte
+	switch bytes.Compare(ourPublic, theirPublic) {
```
ncabatoff, line 60: Is `subtle.ConstantTimeCompare` warranted here?

helper/dhutil/dhutil.go
```diff
@@ -36,21 +39,47 @@ func GeneratePublicPrivateKey() ([]byte, []byte, error) {
 
 // generateSharedKey uses the private key and the other party's public key to
```
ncabatoff, line 40: stale godoc

helper/dhutil/dhutil.go
```diff
@@ -36,21 +39,47 @@ func GeneratePublicPrivateKey() ([]byte, []byte, error) {
...
-	copy(scalar[:], ourPrivate)
-	copy(pub[:], theirPublic)
+	return curve25519.X25519(ourPrivate, theirPublic)
+}
+
+// DeriveSharedKey uses HKDF to derive a key from a shared secret and public keys
+func DeriveSharedKey(secret, ourPublic, theirPublic []byte) ([]byte, error) {
+	// Derive the final key from the HKDF of the secret and public keys.
+
+	//These must be consistently ordered we're on "our" side of key negotiation or the other.
```
ncabatoff, line 57: Can you explain a bit more (or link to some reading material) why the result of the comparison determines the order of arguments?

helper/dhutil/dhutil.go
```diff
@@ -36,21 +39,47 @@ func GeneratePublicPrivateKey() ([]byte, []byte, error) {
...
-	curve25519.ScalarMult(&secret, &scalar, &pub)
+	kio := hkdf.New(sha256.New, secret, pub1, pub2)
 
-	return secret[:], nil
+	var key [32]byte
+	n, err := io.ReadFull(kio, key[:])
+	if n != 32 {
+		return nil, errors.New("short read from hkdf")
+	}
+	if err != nil {
```
ncabatoff, line 78: Maybe move this before the size test, in case there's a more meaningful error?

ncabatoff:
Looks like there's already a test (TestCertWithNameEndToEnd) that exercises encrypted sink tokens, can you copy or extend it to cover the new option?

ncabatoff:
> Looks like there's already a test (TestCertWithNameEndToEnd) that exercises encrypted sink tokens, can you copy or extend it to cover the new option?

Ah, I see you found the test since you modified it.  Looking at the change, it seems that this is going to be a breaking change even without the new option?

helper/dhutil/dhutil.go
sgmiller: Not for a public key, no.

helper/dhutil/dhutil.go
sgmiller: thx

helper/dhutil/dhutil.go
sgmiller: Internally, HKDF hashes the secret and two public keys.  If Alice and Bob are doing DH key exchange, Alice calculates:

HKDF(secret, A, B)  since ourPublic is A.

Bob calculates HKDF(secret, B, A), since Bob's ours is B.  That produces a different value.  Now we only care that both public keys participate in the derivation, so simply sorting them so they are in a consistent numerical order (either one would do) arrives at an agreed value.

helper/dhutil/dhutil.go
sgmiller: Good call.

sgmiller:
> > Looks like there's already a test (TestCertWithNameEndToEnd) that exercises encrypted sink tokens, can you copy or extend it to cover the new option?
> 
> Ah, I see you found the test since you modified it. Looking at the change, it seems that this is going to be a breaking change even without the new option?

I'm hoping it's a non-breaking change unless you set derive_key to true.  Am I wrong?

helper/dhutil/dhutil.go
ncabatoff: Mind adding that as a comment?

ncabatoff:
> > > Looks like there's already a test (TestCertWithNameEndToEnd) that exercises encrypted sink tokens, can you copy or extend it to cover the new option?
> > 
> > 
> > Ah, I see you found the test since you modified it. Looking at the change, it seems that this is going to be a breaking change even without the new option?
> 
> I'm hoping it's a non-breaking change unless you set derive_key to true. Am I wrong? You're right that I blind edited the unit tests always to derive.

I don't know, can we add a test variant that doesn't derive to reassure me?  And I don't know enough about the multiplication change to say whether this is warranted, but do we also want a test where the client uses the old multiplier curve25519.ScalarMult?

sgmiller:
For certain we don't need to worry about the difference between ScalarMult and X25519:

"// Deprecated: when provided a low-order point, ScalarMult will set dst to all
// zeroes, irrespective of the scalar. Instead, use the X25519 function, which
// will return an error."

So they're mathematically equivalent.  But I will add a test case without derived keys.

command/agent/cert_with_name_end_to_end_test.go
```diff
@@ -3,24 +3,24 @@ package agent
 import (
 	"context"
 	"encoding/pem"
+	"github.com/hashicorp/vault/api"
```
calvn, line 6: Import diff block here should belong down below right?

PR #9688, opened 2020-08-07
Fix cassandra data race.

ncabatoff:


ncabatoff:
Upstream fix that's in the fork we're using here: https://github.com/gocql/gocql/pull/1474

PR #9693, opened 2020-08-09
fixing a spelling error

yaleman:


aphorise:
I make this mistake a lot in my writing as well - which is a natural symptom of the lingo (not me) where every word has at least a phonetic other or more 😝

PR #9695, opened 2020-08-09
feat(pg): allow mutlilines statements on SetCredentials

rasta-rocket:
Enabling multilines statement parsing on **SetCredentials** in the PG secret engine will allow vault user to init the DB in the right state in order to use vault static-roles.

Example: 
1. I create a new static role in vault bound to a PG user called foo
2. If the user foo is not created in advance on the DB the static role cannot be created in vault

So the first idea was to add one extra statement that is created the given user in the **rotation_statements** list attribute.
Unfortunately since the _user creation statement is not idempotent_ in postgres (with vanilla SQL statements) the **rotation_statements** will work the first time when the static role is created but it will not work when it arrived at the rotation period because it will rerun the user creation statement and the user already exists which throw an error.

The only way to make user creation idempotent in postgres is by using PL/PGSQL command which are **multilines** ^^
Example:
```
DO $$ BEGIN IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname='{{username}}') THEN CREATE ROLE "{{username}}"; END IF; END $$
```

So the goal of this patch is to allow to parse these kind of statements in the chain of functions of static-roles

_Note: This way of initiating the DB is a good way to avoid relying on any further tool and it will ease the workflow of integrating vault with a newly fresh DB added in the infra._

Cheers ;)


plugins/database/postgresql/postgresql_test.go
```diff
@@ -275,6 +275,22 @@ func TestUpdateUser_Password(t *testing.T) {
 			expectErr:      true,
 			credsAssertion: assertCredsDoNotExist,
 		},
+		"name creation and rotation": {
+			statements: []string{
+				`DO $$ BEGIN IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname='{{name}}') THEN CREATE ROLE "{{name}}"; END IF; END $$`,
+				`ALTER ROLE "{{name}}" WITH PASSWORD '{{password}}';`,
+			},
+			expectErr:      false,
+			credsAssertion: assertCredsDoNotExist,
```
pcman312, line 284: Should this be `assertCredsExist`? (and the same in the next test case)

pcman312 (commented):
Thanks for the PR @rasta-rocket ! I apologize for us not getting to this sooner.

Looking at this it occurred to me that we should probably include the multi-line statement detection logic in the update expiration and delete user logic so we have the same behavior across all of the operations.

The logic in each of the functions is essentially the same: loop over the statements, detect if they are multi-line, split the statements if they are not multiline, and then execute them individually. I suggest we take the statement parsing logic out into its own function that takes in the list of statements from the user, detects multi-line statements, splits them out if they are not multi-line, removes empty statements, and returns the list of statements that should be executed. Then each of the operations can use that list as-is without having to do any checking or other logic on the list.

I threw this together to see if it was a reasonable function:
```go
// splitStatements into individual statements that can be run against PostgreSQL. If it detects a multiline statement,
// that will remain as a single statement. Any empty (or whitespace only) statements will be removed.
func splitStatements(stmts []string) (result []string) {
	for _, stmt := range stmts {
		if containsMultilineStatement(stmt) {
			result = append(result, stmt)
			continue
		}
		split := strutil.ParseArbitraryStringSlice(stmt, ";")
		for _, str := range split {
			str = strings.TrimSpace(str)
			if str == "" {
				continue
			}
			result = append(result, str)
		}
	}
	return result
}
```

Then we can use it in `NewUser`, `changeUserPassword`, `changeUserExpiration`, and `DeleteUser`:
```go
// ...
stmts := splitStatements(req.Statements.Commands)

for _, stmt := range stmts {
	m := map[string]string{
		"name":     username,
		"username": username,
		"password": password,
	}
	if err := dbtxn.ExecuteTxQuery(ctx, tx, m, query); err != nil {
		return fmt.Errorf("failed to execute query: %w", err)
	}
}
```

pcman312:
Additional note: this PR won't solve the problem of being able creating a user during a call to `UpdateUser` because there's a check to see if the user exists first: https://github.com/hashicorp/vault/blob/master/plugins/database/postgresql/postgresql.go#L149-L154. I'm very hesitant to add this functionality (and remove that check) in static credential rotation because the purpose of static credential rotation is to take ownership of a user that already exists and the service(s) using it can't be easily updated to use dynamic users. I do think this PR is valuable for adding multi-line support to the update and deletion usages though.

Can you elaborate on why you want Vault to create the user for static credential rotation? Can you use dynamic users instead?

PR #9696, opened 2020-08-10
adding new config flag disable_sentinel_trace

bruj0:


aphorise:
Any chance of re-naming it to **`log_disable_sentinel_trace`** this would inline it with the other `log...` parameter(s).

bruj0:
> Any chance of re-naming it to **`log_disable_sentinel_trace`** this would inline it with the other `log...` parameter(s).
> This could also be extended for any other component that could be disabled as well (if any at all).

The flag doesnt modify anything related to logs.

aphorise:
> The flag doesnt modify anything related to logs.

It is not related to sentinel log entries in Vault?

PR #9702, opened 2020-08-10
Dev Message after log quiescence

sgmiller:
Register a log sink that delays the printing of the big dev warning until logs have settled
down.


sgmiller:
> I'm uncomfortable-ish with changing the type of Logger from an interface to an implementation type. I could suggest ways of working around it, but they would all be sort of weird and artificial.

Yeah, didn't love that either, but since that field is only ever an InterceptorLogger it seemed safe and that we could engineer around if we needed to go back to generic Logger.
> 
> The other concern I had was whether all the UI output would be interrupted by a log message, which would almost be worse than having it at the top. I don't see any way to tell go-hclog to stop. You _might_ be able to hold things up in quiescenceSink.Accept during the interval where you're printing the message, but that doesn't seem worth it (or necessarily robust, I don't think the delivery order is predictable.)

This I'm sort of okay with.  The user experience is still way better than it was where the dev message is buried two pages back.  A spurious, differently colored log line interspersed doesn't feel like the end of the world, and would be rare anyway.

PR #9726, opened 2020-08-12
ssh backend: Change the default signature algorithm to rsa-sha2-256

thomas-maurice:
The current default rsa-sha1 signature algorithm for SSH certificates is now no longer accepted by modern versions of SSH (> 8.2).

It would make sense to update it to another algorithm that works out of the box and provides an increased level of security.

### DIFF
--- a/builtin/logical/ssh/path_sign.go
+++ b/builtin/logical/ssh/path_sign.go
@@ -531,7 +531,7 @@ func (b *creationBundle) sign() (retCert *ssh.Certificate, retErr error) {
 
 	algo := b.Role.AlgorithmSigner
 	if algo == "" {
-		algo = ssh.SigAlgoRSA
+		algo = ssh.SigAlgoRSASHA2256
 	}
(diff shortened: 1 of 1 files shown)


ncabatoff:
Hi @thomas-maurice ,

Thanks for the PR.  My concern with this change is that it may break existing deployments.  I agree it's unfortunate that the default is no longer supported by newer OSes, but I'm not sure that's yet reason enough for this change.

thomas-maurice:
@ncabatoff this change only affects the setups that have not forced the algorithm to be rsa-sha1. Also I am not sure how it would break existing deployments as the rsa-sha1 signature algorithm is now deprecated anyways and the issued certs just won't work

Blackclaws:
I just ran into this exact same issue. Having the default be something that is still accepted in modern openssh versions would really help a lot. Or at least put a disclaimer onto the documentation page and an example of how to manually change this

calvn:
I believe that this is (at least partially) addressed by https://github.com/hashicorp/vault/pull/9824. The backend will use the key's signing algorithm if none is provided so if a cert has been signed with rsa-sha2-256, it will use that as the `algorithm_signer`.

PR #9727, opened 2020-08-12
Migrate internalshared out

jefferai:
Note: there are no code changes here. There are a couple of minor dep updates (mitchellh/cli, prometheus) and outside of that it's only reorganizing import paths.


ncabatoff:
This doesn't seem to build.  Also, it looks like it deletes ClusterMetricSink?

jefferai:
It builds fine, I thought Circle was off of using `vendor/` but apparently not. ClusterMetricsSink was migrated out to the shared repo.

briankassouf (approved):

PR #9729, opened 2020-08-13
Conditionally overwrite TLS parameters for MySQL secrets engine

0x63lv:
Changes to how connections to MySQL secrets engine are set up, which were introduced with https://github.com/hashicorp/vault/pull/9181 and released with Vault 1.5.0, broke a working setup in Vault versions prior to 1.5.0, where a valid TLS parameter was set in MySQL DSN (e.g. `...?tls=true`).

If none of the new parameters introduced with 1.5.0 (`tls_ca` or `tls_certificate_key`) would be set, the TLS configuration would be empty, and it would be written over the existing TLS  parameters in the DSN (e.g. `tls=true`), resulting in a non-TLS connection attempt to MySQL.

This PR would change that behaviour, and only overwrite the TLS parameters in the DSN, if at least one of the `tls_ca` or `tls_certificate_key` parameters are set. Otherwise it would leave the TLS configuration set in DSN as-is.

Tests also updated with this case.


0x63lv:
Would it be possible to know, when this change will be in a public release?

Valarissa:
It will be released as part of the 1.6 release which should be released later this month (October)

Valarissa:
We'll be releasing it as part of the 1.5.5 release which should be in the next couple weeks. Once again, apologies that this wasn't released sooner.

PR #9735, opened 2020-08-15
TLS Cert Authentication example updates

methridge:
- Updated the Cert Auth example description
- Removed `-ca-cert` parameter from examples as this caused
confusion.  Is this the auth CA or the CA of the listener?


jescalan:
Removing myself as reviewer here - I work on the website structure rather than the docs content. Someone else more familiar with vault, the tool, can probably be of more use!

website/pages/docs/auth/cert.mdx
```diff
@@ -64,17 +64,15 @@ it is up to the administrator to remove it from the method.
...
+The below authenticates against the `web` cert role by presenting a certificate
+(`cert.pem`) and key (`key.pem`) signed by the CA associated with the `web` cert
+role. Note that the name `web` ties to the configuration example below writing
+to a path of `auth/cert/certs/web`. If a certificate role name is not specified,
+the auth method will try to authenticate against all trusted certificates.
 
 ```shell-session
 $ vault login \
     -method=cert \
-    -ca-cert=ca.pem \
```
calvn, line 77: This flag is necessary on the CLI login command since the cert auth method needs [TLS to be enabled](https://www.vaultproject.io/docs/auth/cert#tls-certificates-auth-method) on Vault's listener.

> Please note that to use this auth method, tls_disable must be false in the Vault configuration. This is because the certificates are sent through TLS communication itself.

The CA cert specified here is the cert used by the client to verify the authenticity of the client. The help output from login, `vault login -h`, has a more detailed description around this:

```
  -ca-cert=<string>
      Path on the local disk to a single PEM-encoded CA certificate to verify
      the Vault server's SSL certificate. This takes precedence over -ca-path.
      This can also be specified via the VAULT_CACERT environment variable.
```

The API/cURL example below can technically omit the flag if the CA is added to the system's cert store since cURL falls back to looking up certs installed on the system if this is not provided explicitly. However, for consistency we should keep the flags in both examples. 

The CA that signed `cert.pem` (and provided to the cert role) may or may not have been signed by `ca.pem`, but in this case it is the CA cert that signed the cert for the TLS configuration on the server side and thus required in order to establish communication (unless explicitly ignored by `-tls-skip-verify`). 

My recommendation would be to re-word the paragraph above, but keep the flags in place. We should also consider updating the [example](https://www.vaultproject.io/api/auth/cert#sample-request-8) in the API docs to include the flag.

PR #9738, opened 2020-08-17
Updated OpenShift version

jweissig:
Helm 3 support was added in OpenShift 4.3 [1]. So, we might support 4.X but if we require Helm 3 then we need to start at 4.3. Just didn't want someone to try 4.0-4.2 and wonder why it wasn't working because Helm 3 isn't supported.

[1] https://www.openshift.com/blog/openshift-4-3-deploy-applications-with-helm-3


aphorise:
Request is still valid and the documentation would do well with this note. @jasonodonnell @tomhjp WDYT?

tomhjp (approved):

PR #9746, opened 2020-08-17
[WIP] Create external groups automatically

vishalnayak:
Write tests.
Think about possibilities of assigning policies to those external groups.
Fix the issue of duplicate member entity IDs added to the same group.


mladlow:
@vishalnayak I'm going to un-milestone this until we prioritize it.

PR #9761, opened 2020-08-18
agent: use oklog's run.Group to schedule subsystem runners

calvn:
Use run.Group mechanism to schedule and manage the lifecycle of different subsystems' runners within agent. This mostly gets rid of the need for a `DoneCh` and `errCh`, since each Run funcs can return an error and goroutine termination is properly handled by by `oklog/run`'s Group.Run. 

The exception to this is `template.Server` which still requires a `DoneCh` as the signal to block the termination of the sink and prevent the actor from potentially triggering its interrupt first. More specifically, in the case that `exit_after_auth = true`, we would still want the template server to finish rendering any pending templates before exiting so we need to utilize its `DoneCh` to block on the sink's actor.

### DIFF
--- a/command/agent.go
+++ b/command/agent.go
@@ -15,8 +15,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/oklog/run"
-
 	"github.com/hashicorp/errwrap"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault/api"
@@ -44,6 +42,7 @@ import (
 	"github.com/hashicorp/vault/sdk/version"
 	"github.com/kr/pretty"
 	"github.com/mitchellh/cli"
+	"github.com/oklog/run"
 	"github.com/posener/complete"
 )
 
@@ -533,11 +532,12 @@ func (c *AgentCommand) Run(args []string) int {
 
 	var g run.Group
 
-	// This run group watches for
+	// This run group watches for signal termination
 	g.Add(func() error {
 		for {
 			select {
 			case <-c.ShutdownCh:
+				c.UI.Output("==> Vault agent shutdown triggered")
 				return nil
 			case <-ctx.Done():
 				return nil
--- a/command/agent.go
+++ b/command/agent.go
@@ -15,6 +15,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/oklog/run"
+
 	"github.com/hashicorp/errwrap"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault/api"
@@ -529,8 +531,20 @@ func (c *AgentCommand) Run(args []string) int {
 	// TODO: implement support for SIGHUP reloading of configuration
 	// signal.Notify(c.signalCh)
 
-	var ssDoneCh, ahDoneCh, tsDoneCh chan struct{}
-	errCh := make(chan error)
+	var g run.Group
+
+	// This run group watches for
+	g.Add(func() error {
+		for {
+			select {
+			case <-c.ShutdownCh:
+				return nil
+			case <-ctx.Done():
+				return nil
+			}
+		}
+	}, func(error) {})
+
 	// Start auto-auth and sink servers
 	if method != nil {
 		enableTokenCh := len(config.Templates) > 0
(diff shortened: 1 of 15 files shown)


command/agent/sink/sink.go
```diff
@@ -140,14 +136,14 @@ func (ss *SinkServer) Run(ctx context.Context, incoming chan string, sinks []*Si
 				ss.logger.Trace("no sinks, ignoring new token")
 				if ss.exitAfterAuth {
 					ss.logger.Trace("no sinks, exitAfterAuth, bye")
-					return
+					return nil
```
ncabatoff, line 139: Don't you need a `<-ctx.Done()` here to prevent the rungroup from terminating?

command/agent/template/template.go
```diff
@@ -196,12 +198,19 @@ func (ts *Server) Run(ctx context.Context, incoming chan string, templates []*ct
...
 				ts.runner.Stop()
-				return
+				return nil
 			}
 		}
 	}
 }
 
+func (ts *Server) Stop() {
+	if !ts.stopped.Load() {
```
ncabatoff, line 208: You probably want a CompareAndSwap here to ensure we only call close once.

calvn: Good call, done!

command/agent/alicloud_end_to_end_test.go
```diff
@@ -100,13 +100,9 @@ func TestAliCloudEndToEnd(t *testing.T) {
...
 	}
 
-	errCh := make(chan error)
 	ah := auth.NewAuthHandler(ahConfig)
-	go ah.Run(ctx, am, errCh)
-	defer func() {
-		select {
-		case <-ah.DoneCh:
-		case err := <-errCh:
+	go func() {
```
briankassouf, line 104: IIRC you can't t.Fatal() in a go routine, maybe for tests you can send the error down an errCh to avoid this?

calvn: :( Ah right, I forgot..

ncabatoff: You could also use t.Run() with t.Parallel(), kinda hacky though.

command/agent.go
```diff
@@ -558,11 +570,31 @@ func (c *AgentCommand) Run(args []string) int {
...
+			return ah.Run(ctx, method)
+		}, func(error) {
+			cancelFunc()
+		})
+
+		g.Add(func() error {
+			err := ss.Run(ctx, ah.OutputCh, sinks)
+
+			// Wait until templates are rendered
+			<-ts.DoneCh
```
briankassouf, line 584: Now that `ts.DoneCh` is closed in the `ts.Stop()` function i'm not sure this is truly waiting for `ts` to be done rendering? maybe you should go back to having the DoneCh closed on defer of the server?

calvn: `ts.Stop()` is triggered by the interrupt func, which only happens once the actor exits. That should only happen if the render is finished (or if the render errors).

briankassouf: > which only happens once the actor exits.

I don't think that's true. oklog/run runs the interrupt func if any of the actors exit.

briankassouf: Though, perhaps we don't really care if the templates are rendered if there is another error. We are doing that because the template server can return after a single render, right?

calvn: Right, but on the case where `exit_after_auth = true`, only `ss.Run`'s or `ts.Run`'s actor can exit right away, and we block on the former in here. The latter only exits after a successful render which then triggers its interrupt func, calling `ts.Stop()`, and unblocking this actor (`ss.Run`).

command/agent/alicloud_end_to_end_test.go
calvn: Ran with Brian's suggestion, should be addressed now.

calvn:
Related to #9670

command/agent.go
calvn: Actually this is not quite right. I think this will deadlock since `ts.Run` is blocking on the return until the context is canceled, which won't happen since we're blocking the actor from returning here (and auth handler's actor won't return since it's not aware of this bool).

calvn: This was causing `TestAgent_ExitAfterAuth` to fail, but should now be addressed by 4d73df2. Note that this makes `ss` and `ts` pretty tightly coupled so there's room for improvement, but it should get us through this hoop for the moment.

command/agent.go
```diff
@@ -558,11 +570,34 @@ func (c *AgentCommand) Run(args []string) int {
...
+		}, func(error) {
+			cancelFunc()
+		})
+
+		g.Add(func() error {
+			err := ss.Run(ctx, ah.OutputCh, sinks)
+			c.logger.Info("sinks finished, exiting")
+
+			// Wait until templates are rendered
+			if len(config.Templates) > 0 {
```
briankassouf, line 585: Should we do && ts.exitAfterAuth?

calvn: We shouldn't need to in this case. We are only concerned about it being blocked by this https://github.com/hashicorp/vault/pull/9761/files#diff-0295527e0c39f993dedbf055a88e7b31R104.
If the there are templates to be rendered, `ts.Run` should correctly return and we would need to block here regardless of whether exit after auth is set.

PR #9762, opened 2020-08-18
Adding leader status to the sys/health endpoint

bruj0:
This PR adds 2 new fields to the `sys/health` endpoint:
`is_leader`: boolean, true if the local node is the cluster leader
`leader_address`: current leader API address

The reason for adding this field is because when creating a system to detect leadership changes and possibly reroute client traffic it becomes cumbersome to create the logic.
To make the decision if the traffic should be routed it would involve the output of the `sys/leader` and `sys/health` endpoints and require custom logic.
This simplifies the process by combining the important parts of `sys/leader` in `sys/health`.

### DIFF
--- a/api/sys_health.go
+++ b/api/sys_health.go
@@ -38,4 +38,6 @@ type HealthResponse struct {
 	ClusterName                string `json:"cluster_name,omitempty"`
 	ClusterID                  string `json:"cluster_id,omitempty"`
 	LastWAL                    uint64 `json:"last_wal,omitempty"`
+	IsLeader                   bool   `json:"is_leader"`
+	LeaderAddress              string `json:"leader_address"`
 }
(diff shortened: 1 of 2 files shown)


jefferai:
Closing for now. This is not as simple as it might first seem because in a case where we are failing to have a leader we may well have out of date or wildly inaccurate information about the actual nodes that are there. As one example, a node failing to connect to other nodes may think there is no leader, or may have an outdated idea of which node is the leader, when in fact the other nodes have elected a new leader and are just fine.

jefferai:
Just a note: we may want something like this anyways eventually but we need a better idea of the impetus. Calling Leader() does a lot of locks and a lot of extra work that we may not want on every health call -- including locking the same lock a few times, something that could be cleaned up on the Health side as well. We may want this to be a flag to sys/health, or add extra health info to sys/leader, depending on what will solve the specific problem this is trying to solve. And either way we should probably do some refactoring on the locking.

PR #9766, opened 2020-08-18
Changes the Vault address in the example config

jrx:
The example configuration of the Vault Agent shouldn't point to a local address "127.0.0.1" when referring to the address of the Vault server since it's not a realistic endpoint, so people might get confused.  

```
vault {
        address = "https://127.0.0.1:8200"
}
``` 

This PR changes the address to a generic example address: `https://vault.company.internal:8200`

### DIFF
--- a/website/pages/docs/agent/index.mdx
+++ b/website/pages/docs/agent/index.mdx
@@ -61,7 +61,7 @@ There can at most be one top level `vault` block and it has the following
 configuration entries:
 
 - `address` `(string: <optional>)` - The address of the Vault server. This should
-  be a complete URL such as `https://127.0.0.1:8200`. This value can be
+  be a complete URL such as `https://vault.company.internal:8200`. This value can be
   overridden by setting the `VAULT_ADDR` environment variable.
 
 - `ca_cert` `(string: <optional>)` - Path on the local disk to a single PEM-encoded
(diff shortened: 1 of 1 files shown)


Godwottery:
[RFC2606](https://tools.ietf.org/html/rfc2606), [RFC6716](https://tools.ietf.org/html/rfc6761) specify example.com, (or .example) for example domain. Perhaps it would be more suitable to choose https://vault.example.org:8200 instead?

jrx:
> [RFC2606](https://tools.ietf.org/html/rfc2606), [RFC6716](https://tools.ietf.org/html/rfc6761) specify example.com, (or .example) for example domain. Perhaps it would be more suitable to choose https://vault.example.org:8200 instead?

Thanks, @Godwottery, great suggestion. I've adjusted the example address in the PR.

aphorise:
This request is reasonable - particular given the typical use case where the agent will not connect to Vault on localhost / loopback. The current [Agent docs have three (3x) references to loopback](https://www.vaultproject.io/docs/agent) once in the text and twice in the examples.

However - the address `vault.example.org` that's proposed in this PR is actually a valid FQDN on a TLD and therefore it would be best to avoid use of that or anything potentially resolvable beyond what the user has configured.

What may be better is to stipulate:

```
- `address` `(string: <optional>)` - The address of the Vault server. This should
  be a FQDN or IP such as `https://vault-fqdn:8200` or `https://172.16.9.8:8200`. 
  This value can be overridden by setting the `VAULT_ADDR` environment variable.
```

@yhyakuna @brianshumate your inputs welcome.

yhyakuna:
Folder structure has changed since this PR was originally opened.  I created [PR 16751](https://github.com/hashicorp/vault/pull/16751) to accomplish the fixes, and closing this PR.

PR #9777, opened 2020-08-19
Make the core of agent command usable in tests.

ncabatoff:
Proposed refactoring to ensure we're actually testing the same thing that the agent does.  I didn't bother modifying the other tests, thought I'd see what appetite there was for this change first.


command/agent/cert_end_to_end_test.go
```diff
@@ -580,4 +543,9 @@ func TestCertEndToEnd_CertsInConfig(t *testing.T) {
 
 		time.Sleep(250 * time.Millisecond)
 	}
+
+	err = <-errchan
```
calvn, line 547: The tests are passing, but for the wrong reasons. The tests are getting context canceled after their 30s timeout when ran locally (taking ~33s per case), but we don't report this as a test failure. The same tests on the base branch takes about 4s per test case.

command/agent/cert_end_to_end_test.go
```diff
@@ -301,6 +282,11 @@ func testCertEndToEnd(t *testing.T, withCertRoleName, ahWrapping bool) {
 		}
 	}
 	checkToken()
+
+	err = <-errchan
```
calvn, line 286: The tests are passing, but for the wrong reasons. The tests are getting context canceled after their 30s timeout when ran locally (taking ~33s per case), but we don't report this as a test failure. The same tests on the base branch takes about 4s per test case.

ncabatoff:
This has grown stale.  I'm kind of inclined now to say that we should just use agent.Run directly in e2e tests, as we're doing in a couple of places already.

PR #9778, opened 2020-08-19
Sidebranch: UI transform create and edit clean up

Monkeychip:
**This PR addresses the following:**

* Removes/addresses the TODOs that can be addressed.  Some still remain due to potential API changes.
* Sets up `readOnly` attributes and styling when editing the transformation.  This will be the same pattern for roles and alphabets.  You cannot edit the name or the template it is using once created.
* Adds new properties to the `SearchSelect` component that allows you to: 1. pass in a default set of already selected options (needed for the edit view), and 2. set a `selectLimit` which limits the number of items you can select.  This is needed for templates because you can only select 1 template.
* Hides the `#power-select-with-create` part of the `SearchSelect` component when the limit is reached.  This was how design wanted this to be disabled.
* hides the `templates` model property from edit and create views (This property is needed because of the whole templates vs template return property from the API).

Here is a gif of the changes.
changes

### DIFF
--- a/ui/lib/core/addon/components/search-select.js
+++ b/ui/lib/core/addon/components/search-select.js
@@ -132,10 +132,7 @@ export default Component.extend({
     },
     selectOption(option) {
       if (this.selectLimit) {
-        if (
-          this.selectLimit === this.selectedOptions.length ||
-          this.selectLimit < this.selectedOptions.length
-        ) {
+        if (this.selectLimit <= this.selectedOptions.length) {
           return;
         }
       }
--- a/ui/lib/core/addon/components/search-select.js
+++ b/ui/lib/core/addon/components/search-select.js
@@ -9,7 +9,7 @@ import layout from '../templates/components/search-select';
  * @module SearchSelect
  * The `SearchSelect` is an implementation of the [ember-power-select-with-create](https://github.com/poteto/ember-cli-flash) used for form elements where options come dynamically from the API.
  * @example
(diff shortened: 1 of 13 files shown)


ui/lib/core/addon/components/search-select.js
```diff
@@ -126,6 +131,14 @@ export default Component.extend({
       this.handleChange();
     },
     selectOption(option) {
+      if (this.selectLimit) {
+        if (
+          this.selectLimit === this.selectedOptions.length ||
```
chelshaw, line 136: This can be simplified to `if (this.selectLimit <= this.selectedOptions.length) ...`

ui/lib/core/addon/templates/components/form-field.hbs
```diff
@@ -71,7 +71,10 @@
       @onChange={{action (action "setAndBroadcast" valuePath)}} @inputValue={{get model valuePath}}
       @helpText={{attr.options.helpText}} @subText={{attr.options.subText}} @label={{labelString}}
       @subLabel={{attr.options.subLabel}}
-      @fallbackComponent={{attr.options.fallbackComponent}} />
+      @fallbackComponent={{attr.options.fallbackComponent}}
+      @initialSelected={{initialSelected}}
```
chelshaw, line 75: I wonder if we should pass the initialSelected and selectLimit on the attr so it follows convention for how this component is used. Thoughts?

ui/lib/core/addon/components/search-select.js
Monkeychip: You're a 🧙‍♂️ when it comes to simplification of conditionals.  Thanks!

ui/lib/core/addon/templates/components/form-field.hbs
Monkeychip: The problem with that is initialSelected isn't known on the model (which is where attr.options comes from).  SelectLimit can be though.

Issue #9575, opened 2020-07-23
vault 1.5.0 go mod error


sdeoras:
**Describe the bug**
Error upgrading go mod dependency from vault 1.4.3 to vault 1.5.0

**To Reproduce**
Write simple code that calls vault api at v1.5.0. You may likely run into this issue during go mod tidy.

**Expected behavior**
Expected `go mod tidy` command to go through fine. I got the following error:

```
go mod tidy
go: github.com/hashicorp/vault@v1.5.0 requires
	github.com/hashicorp/vault/api@v1.0.5-0.20200630205458-1a16f3c699c6: invalid version: unknown revision 1a16f3c699c6
```

The error goes away if I use following replace statements in `go.mod` file:
```
require (
	github.com/hashicorp/vault v1.5.0
	github.com/hashicorp/vault/api v1.0.5-0.20200630205458-1a16f3c699c6
)
replace github.com/hashicorp/vault/api => github.com/hashicorp/vault/api v0.0.0-20200718022110-340cc2fa263f
```

**Environment:**
* Vault Server Version (retrieve with `vault status`):
* Vault CLI Version (retrieve with `vault version`):
* Server Operating System/Architecture:

Vault server configuration file(s):

```hcl
# Paste your Vault config here.
# Be sure to scrub any sensitive values
```

**Additional context**

alexanderbez:
Yeah doesn't look like commit 1a16f3c699c6 is on the 1.5 tag HEAD.

zaunist:
I met the same error,how to solve the problem？

ncabatoff:
Hi @sdeoras, @zaunist,

No one should be importing `github.com/hashicorp/vault`, only the `api` and `sdk` modules. The top-level `vault` module is not intended to be a dependency for other projects, we're using go.mod solely to manage our own dependencies here.  There's some discussion of this in https://github.com/hashicorp/vault/issues/7848.

Since we haven't been creating new version tags for api and sdk very often, I suggest using the tip of the branch if you want to use the latest, i.e.

```
go get github.com/hashicorp/vault/api@master
go get github.com/hashicorp/vault/sdk@master
```

greut:
E.g. the `DefaultTokenHelper` is used in the `terraform-provider-nomad` which requires to import everything.

https://github.com/hashicorp/vault/blob/124572bf80150689e747dbe84db0b76f4a1a6ad0/command/config/util.go

unflag:
[...]

PR #4055, opened 2018-02-28
Create a new command/config subpackage.

paddycarver:
This PR extracts the functions associated with loading and parsing
configs, and the DefaultTokenHelper, into a command/config subpackage,
just like TokenHelpers are in the command/token subpackage. The goal is
to allow other clients (in this case, the Vault and Nomad Terraform
providers, but in theory any client that wants to lean on Vault's
default behaviour) to reuse this logic and not drift from Vault, without
vendoring the entirety of Vault.

To retain backwards compatibility, I didn't remove any functions from
the command package; I just copied them into the command/config package,
and update the functions in the command package to call through to the
config package.

PR #9786, opened 2020-08-20
Fix: move DefaultTokenHelper to vault/sdk

greut:
Some projects, such as `terraform-provider-nomad` are relying on the `DefaultTokenHelper` which breaks the idea of only using `vault/api`, or `vault/sdk`. This moves those to a new location.

As vault 1.5.0 cannot be used via go.mod, it's shouldn't be adding more breakage. cf. https://github.com/hashicorp/vault/issues/9575


jmthvt:
Facing the exact same case 2+ years later.
Importing `vault` is unsupported and discouraged, wouldn't this PR make sense?

greut:
@jmthvt maybe. I've updated the branch to resolve the conflicts they were.

maxb:
Wouldn't it make more sense to put this in `vault/api` rather than `vault/sdk` ?

After all, `vault/api` is (mostly) about writing code that talks to the Vault API, whereas `vault/sdk` is (mostly) about implementing Vault plugins.

greut:
> Wouldn't it make more sense to put this in `vault/api` rather than `vault/sdk` ?

One if full with _helpers_ already.

https://github.com/hashicorp/vault/tree/main/sdk/helper

I didn't see where I could put it in the other:

https://github.com/hashicorp/vault/tree/main/api

ncabatoff:
Before we worry about api vs sdk, can someone explain what the purpose behind making this available to other projects is?  What problem does it solve?

ncabatoff:
> Before we worry about api vs sdk, can someone explain what the purpose behind making this available to other projects is? What problem does it solve?

Ah nevermind, I looked at #4055, as I should've before commenting.  I get it now.  I prefer sdk over api, as this isn't something that a typical vault client should care about.

ncabatoff:
I dug a bit more into this and I have a different proposal.  Instead of making DefaultTokenHelper available to other repos like these terraform providers, how about we add the desired functionality into api.Client itself?  In other words, these projects are using DefaultTokenHelper because they want to behave like Vault CLI commands, in that $HOME/.vault-token gets used if $VAULT_TOKEN isn’t set.  So let's make that easy by adding an option to api.Config, say a bool `UseTokenFile`, and when that's set NewClient will read `$HOME/.vault-token`.  This will be less disruptive than moving these packages to the sdk, and will make it easier for other projects to take advantage of this functionality without adding another dependency.  It will also make it easier to discover.

jmthvt:
For clarity, [DefaultTokenHelper](https://github.com/hashicorp/vault/blob/43a5fcc9f42f0ca99e434c0058349d9bb3d776fd/command/config/util.go#L9) looks for a token in `$HOME/.vault-token` **and** from external token helpers, if a token_helper path is defined.

james0209:
Is there any movement on this PR? Did we come to a consensus? I am facing the same issue regarding the discouraged import for `DefaultTokenHelper` almost a year after the last comment  on this PR.

banks:
@ncabatoff and I discussed this PR again last week and came away with the following thoughts:

We still somewhat like the idea of this being more generally the behaviour of the API package that Nick considered before, however that has a few shortcomings too:
 1. It's a larger change that requires more thought
 2. As the comments on the helper itself point out, this isn't _always_ the ideal behaviour. It's mainly useful for short-lived scripts or CLI commands since there is no provision for rotation of tokens, so making it the default API client behaviour is risky and probably incorrect for a large portion of clients that would be embedded in long-running services.

On that basis, it seems perfectly reasonable and more pragmatic to merge this PR essentially as it is (once it's been rebased).

There was a question raised about whether this should live in `api` instead of `sdk`. While that does make some sense on the current usage of `sdk`, we prefer to keep this in `sdk` for now because:
 * Long term we hope to move to a different API package that is generated code. Having extra helpers like this in `api` would then need to be duplicated there or involve needing to import two different versions of the `api` package. We might review this later.
 * It's useful but it's not universally needed by most `api` clients currently so it feels reasonable to import the more complete `sdk` if you want this additional functionality.

We may review either in the future, but in the interest of getting this merged it seems like the best near-term solution is to merge it pretty much as is!

Thanks @greut for the contribution and sorry it got stuck in decision mode for so long. If you would like to retain all the commit credit please feel free to rebase/merge latest changes here so we can get it merged. But equally happy to pick it up since it's been a while!

tomhjp:
I've just pushed `api` tag v1.12.1 with the `DefaultTokenHelper()` function exported from the `cliconfig` package, so I'll close this now. Sorry we didn't get this merged but thank you very much for the contribution ❤️

greut:
@tomhjp thanks a bunch! The Nomad provider can get rid of a ton of dependencies now, https://github.com/hashicorp/terraform-provider-nomad/pull/445

PR #9789, opened 2020-08-20
Ui/transform cleanup 2

Monkeychip:
This PR address clean up issues needed before merge to master

* Remove passing properties of `selectLimit` that are no longer needed.
* Prevents masking input field from taking more than 1 character.
* changes the code copy for the CLI commands.  These will be expanded up in a future PR, but this gets us to an okay place for a master merge.

### DIFF
--- a/ui/app/templates/components/transform-show-transformation.hbs
+++ b/ui/app/templates/components/transform-show-transformation.hbs
@@ -18,8 +18,8 @@
       </span>
     </div>
     <div class="copy-text level">
-      {{#let "vault write transform/encode/<your role name> value=<enter your value here> tweak=<base-64-string>" as |copyEncodeCommand|}}
-        <code>vault write transform/encode/&lt;your role name&gt; value=&lt;enter your value here&gt; tweak=&lt;base-64 string&gt;</code>
+      {{#let "vault write <backend>/encode/<your role name> value=<enter your value here> tweak=<base-64-string>" as |copyEncodeCommand|}}
+        <code>vault write &lt;backend&gt;/encode/&lt;your role name&gt; value=&lt;enter your value here&gt; tweak=&lt;base-64 string&gt;</code>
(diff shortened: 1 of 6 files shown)


ui/app/templates/components/transform-show-transformation.hbs
```diff
@@ -18,8 +18,8 @@
       </span>
     </div>
     <div class="copy-text level">
-      {{#let "vault write transform/encode/payments value=<enter your value here>" as |copyEncodeCommand|}}
-        <code>vault write transform/encode/payments value=&lt;enter your value here&gt;</code>
+      {{#let "vault write transform/encode/<your role name> value=<enter your value here> tweak=<base-64-string>" as |copyEncodeCommand|}}
```
chelshaw, line 21: should also replace `transform` with `backend`

ui/lib/core/addon/components/search-select.js
```diff
@@ -131,11 +131,6 @@ export default Component.extend({
       this.handleChange();
     },
     selectOption(option) {
-      if (this.selectLimit) {
```
chelshaw, line 134: Does selectLimit do anything if this is removed?

ui/app/templates/components/transform-show-transformation.hbs
Monkeychip: Ah, good point!

ui/lib/core/addon/components/search-select.js
Monkeychip: Yes.  selectLimit now prevents the select from ever showing, so the handleChange never comes into play.  It goes from model to template, no need for it to hit the component js file.

Monkeychip: It's hard to see in this PR, because it was done in a previous one, but the template `search-select.hbs` has a line on row 22: `{{#unless (gte selectedOptions.length selectLimit)}}...` that provides the conditional.

PR #9793, opened 2020-08-20
Updates for 1.5.1, 1.4.4, 1.3.8, and 1.2.5

mladlow:
The upgrading pages look like:
Screen Shot 2020-08-20 at 15 39 38


calvn:
Should the upgrades be for 1.5.2, 1.4.5, 1.3.8, and 1.2.5 now?

mladlow:
In the past we've kept the known issues on the version with the known issues, but I've added a note to recommend against using the versions with known issues.

PR #9794, opened 2020-08-21
Add logging during awskms auto-unseal

tvoran:
Adds debug and warn logging around AWS credential chain generation, specifically to help users understand why auto-unseal isn't working on AWS, by logging which role is being used in the case of a webidentity token.

Also in the case of webidentity, this tests retrieving credentials, and logs a warning if it has trouble, in which case it falls through to using creds from the instance metadata, which may still fail, but with a warning logged at least users can tell why the webidentity auth failed.

Depends on https://github.com/hashicorp/go-kms-wrapping/pull/22

Example logging with level=trace when auto-unseal fails:

before:

```

Error parsing Seal configuration: error fetching AWS KMS wrapping key information: AccessDeniedException: User: arn:aws:sts::000000000000:assumed-role/eksctl-tvoran-eks-dev-nodegroup-s-NodeInstanceRole-000000000000/i-0000000000000 is not authorized to perform: kms:DescribeKey on resource: arn:aws:kms:us-west-2:000000000000:key/00000000-00000000-00000000-00000000
    status code: 400, request id: 00000000-00000000-00000000-00000000
stream closed
```

after:

```

2020-08-20T20:33:47.809Z [INFO]  proxy environment: http_proxy= https_proxy= no_proxy=
2020-08-20T20:33:47.812Z [DEBUG] service_registration.kubernetes: "namespace": "default"
2020-08-20T20:33:47.812Z [DEBUG] service_registration.kubernetes: "pod_name": "vault-0"
2020-08-20T20:33:47.812Z [DEBUG] seal.awskms: adding web identity provider: roleARN=arn:aws:iam::000000000000:role/eksctl-my-cluster-addon-iamserviceaccoun-Role1-0000000000000
2020-08-20T20:33:48.798Z [WARN]  seal.awskms: error assuming role: roleARN=arn:aws:iam::000000000000:role/eksctl-my-cluster-addon-iamserviceaccoun-Role1-0000000000000 tokenPath=/var/run/secrets/eks.amazonaws.com/serviceaccount/token sessionName= err="WebIdentityErr: failed to retrieve credentials
caused by: AccessDenied: Not authorized to perform sts:AssumeRoleWithWebIdentity
    status code: 403, request id: 00000000-00000000-00000000-00000000"
Error parsing Seal configuration: error fetching AWS KMS wrapping key information: AccessDeniedException: User: arn:aws:sts::000000000000:assumed-role/eksctl-tvoran-eks-dev-nodegroup-s-NodeInstanceRole-0000000000000/i-0000000000000000 is not authorized to perform: kms:DescribeKey on resource: arn:aws:kms:us-west-2:000000000000:key/00000000-00000000-00000000-00000000
    status code: 400, request id: 00000000-00000000-00000000-00000000
stream closed
```


sdk/helper/awsutil/generate_credentials.go
```diff
@@ -51,6 +55,8 @@ func (c *CredentialsConfig) GenerateCredentialChain() (*credentials.Credentials,
 				SecretAccessKey: c.SecretKey,
 				SessionToken:    c.SessionToken,
 			}})
+		c.Logger.Debug("added static credential provider", "AccessKey", c.AccessKey)
```
calvn, line 58: `c.Logger` could potentially be nil if none is passed into `CredentialsConfig` so these calls would panic. Let's wrap logger calls with a nil check.

builtin/credential/aws/cli.go
```diff
@@ -98,7 +98,7 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
 		headerValue = ""
 	}
 
-	creds, err := RetrieveCreds(m["aws_access_key_id"], m["aws_secret_access_key"], m["aws_security_token"])
+	creds, err := RetrieveCreds(m["aws_access_key_id"], m["aws_secret_access_key"], m["aws_security_token"], hclog.Default())
```
calvn, line 101: This does not derive any of the logging properties from the main command, so it might result in different format, log level, etc.

tvoran: How would we get the logging properties from the main command here? I haven't found a way to do that.

calvn: The log would be from the client (result from a `vault login` command) for this particular case so it might be fine, though `hclog.Default()` logs up to `Info` only, so it would skip any `Debug` output.

tvoran: I'm not seeing anywhere else in the vault login command that verbosity levels are set, so in 2271d6c I added a `log_level` option to the aws cli login command, and used that to configure the logger passed to `RetrieveCreds()`. Or maybe it's preferable to just pass a null logger here?

stephane439 (approved):

calvn (approved):

go.mod
```diff
@@ -58,7 +58,7 @@ require (
 	github.com/hashicorp/go-cleanhttp v0.5.1
 	github.com/hashicorp/go-gcp-common v0.6.0
 	github.com/hashicorp/go-hclog v0.14.1
-	github.com/hashicorp/go-kms-wrapping v0.5.12
+	github.com/hashicorp/go-kms-wrapping v0.5.15
```
calvn, line 61: We should be using v0.5.16 right?

tvoran: Yep, just fixed in e4f2e2b

Issue #5893, opened 2018-12-04
Unwrap tokens from the UI login page


khicks:
I really enjoy how easy it is to wrap existing secrets now in 1.0.0. Retrieval however, seems to be the same tedious process for users who only use the UI: log in, go to Tools -> Unwrap, and paste in the response-wrapping token. This method of retrieval is, of course, only for users who already have a way to otherwise access Vault. Users without any other Vault credentials must retrieve the response-wrapped secret via the CLI or API.

I think it would be helpful to be able to use response-wrapping tokens directly from the UI login screen. A possible use case would be when sharing a one-time secret with a colleague who may not have Vault credentials of their own or the CLI tool installed.

For example, it could look like this:
![image](https://user-images.githubusercontent.com/3689108/49418844-c1268300-f749-11e8-9c72-6f7a1b3b7cbf.png)

Or perhaps it could be an extra checkbox option underneath the normal Token login field.

Thanks for reading and for all the hard work!

avoidik:
you may try with
http://<hostname>:8200/ui/vault/auth?with=token&wrapped_token=<wrapped_token>

Garath620:
@rotorcowboy
It would be a very good feature!
+1

Garath620:
An alternative could be to access directly without authentication to : `/ui/vault/tools/unwrap`
and pass `wrapped_token` in url param.
`http://<hostname>:8200/ui/vault/tools/unwrap?wrapped_token=<wrapped_token>`

meirish:
@Garath620 those routes are gated on authenticated currently, so the suggestion of using `wrapped_token` will work the same way I described above here: https://github.com/hashicorp/vault/issues/5893#issuecomment-445286965 - you'll likely just lose whatever data was wrapped as the UI would not display it.

SimonGurney:
Has there been any progress on this, it would be a very useful feature

aphorise:
I believe that this has already been delivered as described in the ***UI*** tabs of the [learn guide for example](https://learn.hashicorp.com/tutorials/vault/cubbyhole-response-wrapping#step-2-unwrap-the-secret). 

Issue should be closed - or am I miss-understanding something?

![](https://learn.hashicorp.com/img/vault-cubbyhole-5.png)

khicks:
[...]

PR #9799, opened 2020-08-22
UI - Unwrap Wrapping token from Login page. Resolves: #5893.

aphorise:
Provided new feature for unwrapping data on login screen as per the request on GH-Issue: #5893 

Re-used existing unwrap templates which were added to login. Credentials or other pertinent data can be unwrapped by end-users via UI without the need for CLI or API specific access.

PS - advance apologies for missing tests - kindly help with my inadequacies there.

90963669-f5ac1680-e4b9-11ea-82e9-5d55d9050760

### DIFF
--- a/ui/app/components/auth-form.js
+++ b/ui/app/components/auth-form.js
@@ -9,6 +9,8 @@ import { supportedAuthBackends } from 'vault/helpers/supported-auth-backends';
 import { task } from 'ember-concurrency';
 const BACKENDS = supportedAuthBackends();
 
+import { toolsActions } from 'vault/helpers/tools-actions';
+
 /**
  * @module AuthForm
  * The `AuthForm` is used to sign users into Vault.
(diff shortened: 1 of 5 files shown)


aphorise:
Any chance of getting some input here? - it would be nice to know that this feature would also be offered nearing to 1.10 (only been 15 months 😛)

aphorise:
Hey guys this request is approaching its 2 year mark (in a months time) from the time I've opened it. What is the status here?

Please state alternatives and close if you are not likely to merge it - it's not clear what's holding this back or what may be next.

pree:
I really like this change. The only issue I see is the audibility. It might need an extra flag for wrapped tokens, to allow unauthenticated unwrapping.

zofskeez:
Thanks for opening this @aphorise. This was discussed by the UI team and it was determined that it would be a good feature but one that needs further consideration. The feature has been added to the list of potential projects for future prioritization.

Issue #7028, opened 2019-07-01
Implement systemd notification


astorath:
**Is your feature request related to a problem? Please describe.**
Using vault binary as a systemd unit have some control problem:
- vault does not have daemon mode
- vault does not support systemd notification system

So when service starts unsuccessfully (due to config error for instance), systemctl can't report the problem right away and systemd keeps restarting failing unit forever.

**Describe the solution you'd like**
I think the best solution is to implement systemd notification when vault (or vault agent) is ready to accept requests.

Example from consul: [agent.go](https://github.com/hashicorp/consul/blob/33a7df333018c79087f1be092297402a312c571d/agent/agent.go#L1763)

**Describe alternatives you've considered**
I've implemented a simple bash wrapper to check for active port after vault startup. This is a dirty solution:
- it makes startup take longer
- checking port availability is not the same as checking service is ready
- requires some unnecessary intermediate script

**Explain any additional use-cases**
You can implement [WATCHDOG](https://www.freedesktop.org/software/systemd/man/sd_notify.html#WATCHDOG=1) notification to make systemd monitor unit health.

**Additional context**
I see that `github.com/coreos/go-systemd` module is in https://github.com/hashicorp/vault/blob/ceb4aeadcb793a2355208feb5fce6fd06aa92dbe/vendor/modules.txt#L185 and https://github.com/hashicorp/vault/blob/0a58f3077b455dc7eb71fecc6a999c4959b61a4f/go.sum#L104, maybe there were some attempts to implement it already?

aphorise:
Should be implemented in above PR (make a note for my own reference).

sgmiller:
Implemented using a cleaner dep, and also notifying on shutdown/config reload: https://github.com/hashicorp/vault/pull/11517

aphorise:
@astorath can you confirm if the merged changes already accommodate for what you were after within the latest 1.7.x or higher releases?

ncabatoff:
This was fixed in #11517 and should've been closed then.  Closing.

PR #9802, opened 2020-08-23
Server systemd notify added similar to Consul. Resolves: #7028

aphorise:
Not sure if there is a more elegant way of reading Systemd configuration so as to only display message conditional to Vault unit file reads:

```
[Service]
Type=notify
```

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -20,6 +20,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/hashicorp/consul/agent/systemd"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-hclog"
 	log "github.com/hashicorp/go-hclog"
@@ -73,8 +74,16 @@ const (
 	storageTypeConsul = "consul"
 )
 
+// notifier is called after a successful JoinLAN.
+type notifier interface {
+	Notify(string) error
+}
+
 type ServerCommand struct {
 	*BaseCommand
+
+	// joinLANNotifier is called after a successful JoinLAN.
+	NotifierSystemd notifier
 
 	AuditBackends      map[string]audit.Factory
 	CredentialBackends map[string]logical.Factory
@@ -798,6 +807,9 @@ func (c *ServerCommand) processLogLevelAndFormat(config *server.Config) (log.Lev
 }
 
 func (c *ServerCommand) Run(args []string) int {
+
+	c.NotifierSystemd = &systemd.Notifier{}
+
 	f := c.Flags()
 
 	if err := f.Parse(args); err != nil {
@@ -1548,6 +1560,9 @@ CLUSTER_SYNTHESIS_COMPLETE:
 			for {
 				err := core.UnsealWithStoredKeys(context.Background())
 				if err == nil {
+					if notifyErr := c.NotifierSystemd.Notify(systemd.Ready); notifyErr != nil {
+						c.logger.Debug("server: systemd notify failed", notifyErr)
+					}
 					return
 				}
 


command/agent.go
```diff
@@ -15,6 +15,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/hashicorp/consul/agent/systemd"
```
ncabatoff, line 18: I don't think we should be depending on an internal Consul package.  Can it be factored out somewhere instead, or if it's very small maybe we can duplicate it?

aphorise: It's rather small & I'd agree to making a copy local to this repo retaining some header of where its taken from :-) - let me know what to do.

aphorise: @ncabatoff should I localise the `systemd` dependency? - kindly suggest how to proceed with this PR.

ncabatoff: Yes please

command/agent.go
```diff
@@ -862,12 +860,11 @@ func (c *AgentCommand) Run(args []string) int {
...
 		c.UI.Error(fmt.Sprintf("Error storing PID: %s", err))
 		return 1
-	} else {
-		if notifyErr := c.NotifierSystemd.Notify(systemd.Ready); notifyErr != nil {
-			c.logger.Debug("server: systemd notify failed", notifyErr)
-		}
 	}
 
+	// Notify systemd that the server is ready (if applicable)
+	c.notifySystemd(systemd.SdNotifyReady)
```
aphorise, line 866: I tried adding this a bit later after the run condition which may be more valid to NotifyReady as and when it's actually ready - however since the time it may take for that is not as easily determinable then getting an earlier start notification coupled with a followed up by any stop is probably better than just getting NotifyStopping without ever seeing NotifyReady

ncabatoff:
Changes seem fine.  Since this isn't practical to test using Go tests, can you summarize what manual testing you did?

aphorise:
@ncabatoff  thank you for the review.

The manual tests performed were to provision a Vault-agent and to confirm the expected emittance - observing the following tests:

1. Vault agent being started in a valid / proper sense where only `NotifyReady` is expected.
2. Vault agent - later being stopped where only `NotifyStopping` is received.
3. Vault agent being started / with some erroneous condition where it emits ``NotifyReady` attempting to progress as far as possible before crashing / stopping when `NotifyStopping` is then emitted.

ncabatoff (approved):

ncabatoff:
Thanks!  Could you merge in latest main please?

jdoss:
Any chance to get this rebased and merged?

ncabatoff:
Hi @aphorise, sorry I forgot about this.  Could you merge in main again please, then I'll merge this?

aphorise:
hey @ncabatoff done - hope this is okay. Thanks.

aphorise:
@ncabatoff will this make it into the next release?

command/agent.go
```diff
@@ -794,13 +795,16 @@ func (c *AgentCommand) Run(args []string) int {
...
+				// Notify systemd that the server is shutting down
+				c.notifySystemd(systemd.SdNotifyStopping)
 				// Let the lease cache know this is a shutdown; no need to evict
 				// everything
 				if leaseCache != nil {
 					leaseCache.SetShuttingDown(true)
 				}
 				return nil
 			case <-ctx.Done():
+				c.notifySystemd(systemd.SdNotifyStopping)
```
ncabatoff, line 807: Shouldn't we also do this in the case above, when we were asked to shutdown?

aphorise: hey @ncabatoff I'm not comprehending - does `systemd.SdNotifyStopping` need to be included in any other case?

aphorise: If any **one** of the two signals are the only mode of exit (outside Windows) then I think these two later here would suffice and void any repeats - but I'm not entirely sure TBH.

ncabatoff: Maybe it's simpler just to add the notifySystemd call to the end of the func, or to a defer.

aphorise: Is there any chance that a end or defer function may take longer or not ever reached? - just thinking aloud in case of processor monitors or other cases an agent may be forced quit and never actually ending.

Anyway @ncabatoff any chance you can do a PR if you think a defer may be better.

ncabatoff: Defer will always fire when the func returns, and it shouldn't be any slower than doing it just before a return statement.

ncabatoff (approved):

PR #8499, opened 2020-03-06
Unique lock key based on database and table name

pandeybk:
Signed-off-by: Balkrishna Pandey <sachit.nep@gmail.com>

Issue #8458, opened 2020-03-03
HA vault cluster failing to communicate with MySQL db endpoint from another cluster


pkoritala:
**Describe the bug**
I'm currently using [hashicorp vault chart](https://github.com/hashicorp/vault-helm) which has image 1.3.2 to install in kubernetes cluster(v1.13.10) using helm (v3.0.3) for HA installation. I created a new amazon rds aurora db and when installed with vault chart, it worked fine.
But when I used the same amazon rds aurora db endpoint from a different kubernetes cluster using different table name, 
`vault operator init` - successful
 `vault login <roottoken> - failed with the following error. This bug is consistent.

```
Error authenticating: error looking up token: Error making API request.

URL: GET https://127.0.0.1:8200/v1/auth/token/lookup-self
Code: 500. Errors:

* sql: no rows in result set
```
I tried to debug, by logging into the mysql database.
Has no records in the <tableName>_lock table.

**To Reproduce**
1. Download [hashicorp vault chart](https://github.com/hashicorp/vault-helm)

2. Package the chart.

3. Create values.yaml and configure as per the below content.
```
server:
  image:
    repository: "vault"
    tag: "1.3.2"
    pullPolicy: IfNotPresent

ha:
    enabled: true
    replicas: 3
    config: |
      ui = true
      listener "tcp" {
          address = "[::]:8200"
          cluster_address = "[::]:8201"
          tls_cert_file = "/vault/userconfig/vault-server-tls/tls.crt"
          tls_key_file  = "/vault/userconfig/vault-server-tls/tls.key"
          tls_client_ca_file = "/vault/userconfig/vault-server-tls/ca.crt"
        }
 
      storage "mysql" {
          ha_enabled = "true"
          address = "{{ .Values.mysql.address }}"
          username = "{{ .Values.mysql.username }}"
          password = "{{ .Values.mysql.password }}"
          database = "{{ .Values.mysql.database }}"
          table = "{{ .Values.mysql.table }}"
      }
 
      seal "awskms" {
        region     = "{{ .Values.awskms.region }}"
        kms_key_id = "{{ .Values.awskms.kms_key_id }}"
      }
```
3. Install chart by overriding the values.yaml
helm install vault <package_name> --values values.yaml

For the 1st cluster with the db endpoint.
1. Run `vault operator init
2. Run `vault login <roottoken>` (works fine)

For the 2nd cluster with the db endpoint with different table name.
1. Run `vault operator init
[...]

PR #9807, opened 2020-08-24
Creating unique lock key for MySQLHALock in v1.5.0 (#8458/8499)

cjbaar:
This is an update to the ongoing discussions for issue #8458/PR #8499, regarding the locking mechanism when using MySQL as the HA backend. In short, the current lock key is set as a constant ('core/lock'), which means multiple vault installations are not able to share the same MySQL instance.

This is a re-write of the solution proposed by @pandeybk on 8499, and was branched from v1.5.0. It uses the database+table name to create a unique string for the lock. This implementation makes two changes from the previous version.

1) The unique string is generated on the &MySQLHALock struct, instead of the &MySQLLock struct. We found issues running the first version which would lose HA cohesion after 8 hours (the default MySQL session timeout in our environment).

2) The db+table string is hashed, to make sure the string does not overrun the 64-character limit for lock names (starting with 5.7)

### DIFF
--- a/physical/mysql/mysql.go
+++ b/physical/mysql/mysql.go
@@ -2,6 +2,7 @@ package mysql
 
 import (
 	"context"
+	"crypto/sha1"
 	"crypto/tls"
 	"crypto/x509"
 	"database/sql"
@@ -450,9 +451,17 @@ func (m *MySQLBackend) List(ctx context.Context, prefix string) ([]string, error
 
 // LockWith is used for mutual exclusion based on the given key.
 func (m *MySQLBackend) LockWith(key, value string) (physical.Lock, error) {
+	// If multiple instances of Vault deployment trying to use the same MySQL
+	// DB endpoint, the first deployment will hold lock forever and others
+	// deployment will fail to insert record in the lock table.
+	// Use of a unique lock key based on db/table will resolve this problem.
+	// The db/table is hashed to ensure the lock string is less than 64 chars.
+	lockData := []byte(m.conf["database"] + "/" + m.conf["table"])
+	lockKey := key + "/" + fmt.Sprintf("%x", sha1.Sum(lockData))
+
 	l := &MySQLHALock{
 		in:     m,
-		key:    key,
+		key:    lockKey,
 		value:  value,
 		logger: m.logger,
 	}


jzbruno:
Can we get someone to look at this? We would like to be able to run current versions of Vault but in our environment we need this fix.

llivingstone:
+1 to this. 

Having one single MySQL instance w/ a database for each vault cluster is a cost effective solution for specific environments.

If i can help in anyway, please let me know.

edshelton:
Same use case as @llivingstone here - would be great to see some traction on this!

cjbaar:
@sgmiller Yes; as noted above, the hash ensures that the key itself will not exceed the limit of 64 characters in v5.7. If a user has long database and table names, it is possible the concatenation would go over 64 characters, which is why we decided to incorporate the hash.

nandudheeraj:
+1

Very Much needed

aftershockdan:
+1 we need this fix as well. Can we get this merged in and released soon please?

banks:
Hi folks. I'm fully aware of how old this PR is but I'm trying to help get some of our backlog of things addressed.

I think I understand it all. I understand the hashing to avoid length issues. One question: is the fact the lock name is now opaque likely to be an issue? Like would it make it harder for a DB admin to work out which lock corresponded to which Vault instance and would that matter? I've not used locks in MySQL like this before so don't have a good intuition for that so any feedback from those of you here that do would be welcome!

If it is an issue, there may be some hybrid where you could do something like first and last 4 chars of both DB and table followed by random hash to avoid collisions and then most folks would get something partially human readable?

But happy to take your expert guidance on if that is a real issue.

I also think Brian's [concern on the original PR](https://github.com/hashicorp/vault/pull/8499#issuecomment-596035936) is still valid and will require more siginficant changes. It's not necessarily a reason to not merge this but I think it will need further work to avoid breaking every Vault+MySQL HA cluster on upgrade!

As far as I can tell there is no safe way to do a standard in-place rolling upgrade of a Vault cluster using MySQL storage if we merge this change. Any new clusters will be fine but upgrades would be very likely to end up split-brain (i.e. two active nodes at the same time) which can corrupt storage and even loose data 😱 . The only way I can see to make that 100% safe would be if we did a much more elaborate thing here where we introduce a new config to say "keep checking the old lock key too" and if that is set then we have to acquire _both_ keys to become active node.

With that it would still require operators to explicitly follow a correct procedure during upgrade:
 1. reconfigure each vault server with the new `also_check_old_lock_key=true` (or whatever we call it) setting.
 2. upgrade binary and restart
 3. complete process on each server in turn
 4. go back and remove `also_check_old_lock_key=true` and then restart each server in turn again.

This ensures that the first server you upgrade doesn't immediately grab the _new_ lock key and become active even though the previous active node is still leader under the old lock key.

The above would still be dangerous in that any users who didn't carefully read the upgrade notes would likely have their cluster split-brain during upgrade which they may or may not notice but could lead to corruption and data loss. For that reason I think it would need to be merged as opt-in behaviour only (i.e. by default we still use the old lock key so no surprises on upgrade) but only if you configure `instance_specific_lock_key=true` do we do the new behaviour in this PR. We'd _still_ need to have the "claim both locks" thing mentioned above to allow folks to migrate an existing cluster safely but at least those who don't need this won't have a nasty surprise. Or a lower-effort option would be to make it opt-in only but not bother with the double-locking thing and then document with huge warnings that it is never safe to change a cluster from opt-out to opt-in without taking it entirely offline (i.e. stopping all instances of Vault before changing the config and restarting)? Then folks with existing clusters would remain opted out and use the current Key safely, while new clusters could opt-in and share the same DB with new keys?

Is anyone here still interested in picking up that extra bit of work to make this safe to merge? If not for now perhaps we can document the requirements in the original issue and close this until there is someone who would like to take that on?

Thanks (belatedly 😢 ) @cjbaar for this great contribution and to all who have commented. I hope we can get it resolved one way or another soon!

hsimon-hashicorp:
@cjbaar it has been several months since we requested more information - are you able to continue this effort? Would you like someone to get it over the finish line? Let me know. Thanks!

cjbaar:
@hsimon-hashicorp  I haven't worked with this for over 3 years now. It was my understanding the PR was closed.

hsimon-hashicorp:
> @hsimon-hashicorp I haven't worked with this for over 3 years now. It was my understanding the PR was closed.

Certainly understood - one of my main goals in my role is to prevent this sort of "PR decay" from occurring. I will check with the team on how they wish to handle this - sometimes we will fork a PR to finish it, while retaining credit for the original contributor. If that's not something you want, please do let me know. And thank you for the fast response!

cjbaar:
Thanks. Whatever is easier for you is fine; I am not necessarily concerned with the credit.

hsimon-hashicorp:
After discussion with engineering, I'm going to go ahead and close this PR out now. If you or another community member is interested in completing this work, please feel free to open a new PR. Thanks!

Issue #7304, opened 2019-08-13
The UI Secrets page should display the mount description


nickwales:
It would provide useful information to customers if the secrets page that shows all the current mounts could also display the description. 

**Describe the solution you'd like**
Displaying the description `/ui/vault/secrets` page could have the configured mount description displayed to the right of the mount name + ID 

**Describe alternatives you've considered**
The description is available when looking at the mount settings but customers are unlikely to go and investigate there.

**Additional context**
Here's a very high quality mockup of how it _could_ look.
![Screen Shot 2019-08-13 at 10 26 29 AM](https://user-images.githubusercontent.com/588472/62954824-a8c93680-bdb5-11e9-82d9-b46d4ebd23f7.png)

lpar:
I'd really like to be able to have a displayed description at secret level as well. My current kv secrets store has one, and I've found it very useful when I can't remember which remote database system is associated with which business process, etc.

Failing that, is there a performance implication to having a separate kv2 store mounted for every application, so that at least each application can have a description?

aphorise:
Above PR should address built-in secret engines with a description detail where none are set by the user.

![Screenshot 2020-08-25 at 00 41 30](https://user-images.githubusercontent.com/974854/91103769-917d8400-e66c-11ea-9365-550a93401457.png)

aphorise:
@nickwales - have you tried the latest 1.8.0 release? - please consider closing this ticket if it fits the bill.

<img width="1072" alt="Screenshot 2021-08-05 at 00 16 58" src="https://user-images.githubusercontent.com/974854/128262730-5e180707-a6bd-4e9a-9401-6575f47ae3e3.png">

aphorise:
Closing as there's been no follow-up & original request made is already in place.

PR #9814, opened 2020-08-24
UI - Added default text description where none are set for builtin secret engines. Resolves: #7304

aphorise:
Provided new feature for displaying mount descriptions in the listing view - as per the request on GH-Issue: #7304  

Feature request is reasonable - given that multiple mounts or other's that are similar (eg: `pki1/`, `pki2/`, `pk3/`) are not distinguishable at present due to lacking descriptives as part of the secrets mount-listing view. Presently this data is collected by not displayed unless going much further into the configuration view of the mount.

PS - advance apologies for missing tests - kindly help with my inadequacies there.

Screenshot 2020-08-25 at 00 41 30

### DIFF
--- a/ui/app/templates/vault/cluster/secrets/backends.hbs
+++ b/ui/app/templates/vault/cluster/secrets/backends.hbs
@@ -64,6 +64,9 @@
           <code class="has-text-grey is-size-8">
             {{backend.accessor}}
           </code>
+          <font class="has-text-grey is-size-8" style="margin-left:10px;">
+            {{backend.description}}
+          </font>
         </div>
       </div>
       <div class="level-right is-flex is-paddingless is-marginless">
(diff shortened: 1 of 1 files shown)


ui/app/templates/vault/cluster/secrets/backends.hbs
```diff
@@ -130,6 +133,31 @@
           <code class="has-text-grey is-size-8">
             {{backend.accessor}}
           </code>
+          <font class="has-text-grey is-size-8" style="margin-left:10px;">
+            {{#if (eq backend.description "")}}
+              {{#if (eq "ad" backend.type)}}Active Directory credentials & service accounts.{{/if}}
```
chelshaw, line 138: We should store these strings somewhere centralized (such as `options-for-backend`) and then get them here instead of doing all these if statements

ui/app/templates/vault/cluster/secrets/backends.hbs
```diff
@@ -64,6 +64,9 @@
           <code class="has-text-grey is-size-8">
             {{backend.accessor}}
           </code>
+          <font class="has-text-grey is-size-8" style="margin-left:10px;">
```
chelshaw, line 67: Getting linter errors that `style` on templates (hbs files) is not allowed, so we should instead use a class for the left margin (can add to `helpers.scss` and follow the pattern there)

ui/app/templates/vault/cluster/secrets/backends.hbs
chelshaw: Save each string as `description` key to its corresponding engine type in `SECRET_BACKENDS` and then use the helper like 
```hbs
{{#unless backend.description}}
  {{get (options-for-backend backend.type) "description"}}
{{else}} ...
```

chelshaw (changes requested):
A couple pointers to help clean this up (and get tests passing). A test that checks for the both the default and custom-set description to show up would be great too!

aphorise:
Redundant and likely addressed in #11995 already.

PR #9817, opened 2020-08-25
Add date/time argument type.

mgritter:
This lets us create commands that take an absolute time.


ncabatoff:
My only concern with this new timevalue is that as implemented it permits partially specified times.  I understand that's desirable for the feature at hand, but I'm not sure that's the right general behaviour for a time input, and it will be hard to change later if clients start depending on it.  Should we choose a less general name for this time input that allows partial time values?

mgritter:
> My only concern with this new timevalue is that as implemented it permits partially specified times. I understand that's desirable for the feature at hand, but I'm not sure that's the right general behaviour for a time input, and it will be hard to change later if clients start depending on it. Should we choose a less general name for this time input that allows partial time values?

I am happy to change the name. However, I'm pretty sure the next developer who goes looking for a time parser for arguments is just going to re-use this one no matter what it's named.  :)

I think the only case where the partial-time parsing might go further than similar conventions used elsewhere is support for specifying just the month. We could make that opt-in behavior via a flag in TimeFlag

ncabatoff:
> We could make that opt-in behavior via a flag in TimeVar

Sounds good.

command/base_flags.go
```diff
@@ -749,6 +750,134 @@ func (f *FlagSet) Var(value flag.Value, name, usage string) {
...
+			return t, err
+		}
+	}
+
+	if allowedFormats&TimeVar_EpochSecond != 0 {
+		i, err := strconv.ParseInt(input, 10, 64)
+		if err == nil {
+			// Four-digit numbers are ambiguous, don't
+			// parse them as a time in 1970.
+			if i > 10000 {
```
ncabatoff, line 819: Do we really care to accept anything older than say 2020?  What if someone decides to leave off date separators and gives 20200830 as input?

mgritter: That's a good use case to consider. We could eliminate everything below about 946000000.

mgritter: Changed bound to 946000000 (2000-ish)

PR #9818, opened 2020-08-25
Fixes for AWS sts allowed header values

sgmiller:
- Only validates AWS related (X-Amz-) headers.
- Changes the aws_sts_allowed_header_values parameter to be additive to the default list.
- Changes parsing of that config entry to a comma separated string for ease of use.
- Adds that config entry to output of reads to an aws client config.

### DIFF
--- a/builtin/credential/aws/backend.go
+++ b/builtin/credential/aws/backend.go
@@ -19,7 +19,7 @@ import (
 )
 
 var defaultAllowedSTSRequestHeaders = []string{"Authorization", "Content-Length", "Content-Type", "Host", "User-Agent",
-	"X-Amz-Date", "X-Amz-Security-Token", "X-Vault-Aws-Iam-Server-Id",textproto.CanonicalMIMEHeaderKey(iamServerIdHeader)}
+	"X-Amz-Date", "X-Amz-Security-Token", textproto.CanonicalMIMEHeaderKey(iamServerIdHeader)}
 
 func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend, error) {
 	b, err := Backend(conf)
--- a/builtin/credential/aws/backend.go
+++ b/builtin/credential/aws/backend.go
@@ -3,10 +3,6 @@ package awsauth
 import (
 	"context"
 	"fmt"
-	"github.com/hashicorp/vault/sdk/framework"
-	"github.com/hashicorp/vault/sdk/helper/awsutil"
-	"github.com/hashicorp/vault/sdk/helper/consts"
-	"github.com/hashicorp/vault/sdk/logical"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


mladlow:
Shouldn't this go into master and then get backported to everything else?

kalafut:
This LGTM, but I'd like @calvn to sign off.

builtin/credential/aws/backend.go
```diff
@@ -3,22 +3,23 @@ package awsauth
...
-	"github.com/hashicorp/vault/sdk/framework"
-	"github.com/hashicorp/vault/sdk/helper/awsutil"
-	"github.com/hashicorp/vault/sdk/helper/consts"
-	"github.com/hashicorp/vault/sdk/logical"
 	cache "github.com/patrickmn/go-cache"
 )
 
-var defaultAllowedSTSRequestHeaders = []string{"Authorization", "Content-Length", "Content-Type", "User-Agent", "X-Amz-Date", textproto.CanonicalMIMEHeaderKey(iamServerIdHeader)}
+const amzHeaderPrefix = "X-Amz-"
+var defaultAllowedSTSRequestHeaders = []string{"X-Amz-Date", "X-Amz-Security-Token", "X-Amz-Algorithm", "X-Amz-Signature",
```
calvn, line 21: Nit: Can we have this be one line per entry so that it's a bit easier to read/count?

sgmiller: Sure.

builtin/credential/aws/backend.go
```diff
@@ -3,22 +3,23 @@ package awsauth
 import (
 	"context"
 	"fmt"
-	"net/textproto"
+	"github.com/hashicorp/vault/sdk/framework"
```
calvn, line 6: Imports seems to be out of order here

sgmiller:
> Shouldn't this go into master and then get backported to everything else?

Master still doesn't have the 1.5.x point fixes in it, so it makes a huge mess of conflicts atm.

builtin/credential/aws/backend.go
sgmiller: Yep, Goland likes to do that.

builtin/credential/aws/backend.go
```diff
@@ -18,7 +17,13 @@ import (
 	cache "github.com/patrickmn/go-cache"
 )
 
-var defaultAllowedSTSRequestHeaders = []string{"Authorization", "Content-Length", "Content-Type", "User-Agent", "X-Amz-Date", textproto.CanonicalMIMEHeaderKey(iamServerIdHeader)}
+const amzHeaderPrefix = "X-Amz-"
+var defaultAllowedSTSRequestHeaders = []string{
+	"X-Amz-Date",
```
calvn, line 22: Should we also include `X-Amz-Credential`, based on the common params for STS from [this page](https://docs.aws.amazon.com/STS/latest/APIReference/CommonParameters.html)?

PR #9825, opened 2020-08-25
Aws auth fixes

sgmiller:
Rebuilt changes by cherry-picking from the two major branches from the point releases.  Should be identical to what was previously reviewed.


builtin/credential/aws/path_config_client.go
```diff
@@ -2,6 +2,11 @@ package awsauth
 
 import (
 	"context"
+	"errors"
+	"github.com/hashicorp/vault/sdk/helper/strutil"
```
calvn, line 6: Move down?

sgmiller: <shakes fist at Goland>

PR #9838, opened 2020-08-26
Update template/templates data on transformation

chelshaw:
This PR essentially normalizes the data coming back from the API to match what gets set on the API. In addition to simplifying the model it fixes an issue with the Search Select component, shown on the gif below, where the template data on the transformation is not set correctly on the edit form.

**BEFORE:** 
before-templates-on-transformation

**AFTER**
after-template-transformations

### DIFF
--- a/ui/app/serializers/transform.js
+++ b/ui/app/serializers/transform.js
@@ -8,10 +8,10 @@ export default ApplicationSerializer.extend({
     return this._super(store, primaryModelClass, payload, id, requestType);
   },
 
-  serialize(snapshot, options) {
+  serialize() {
     let json = this._super(...arguments);
     if (json.template && Array.isArray(json.template)) {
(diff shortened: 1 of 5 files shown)


ui/app/serializers/transform.js
```diff
@@ -5,7 +5,15 @@ export default ApplicationSerializer.extend({
...
       payload.data.masking_character = String.fromCharCode(payload.data.masking_character);
     }
-    // TODO: the BE is working on a ticket to amend these response, so revisit.
     return this._super(store, primaryModelClass, payload, id, requestType);
   },
+
+  serialize(snapshot, options) {
+    let json = this._super(...arguments);
+    if (json.template && Array.isArray(json.template)) {
+      //   // Transformations should only ever have one template
```
Monkeychip, line 14: Not a blocker, but looks like double comment lines here.

PR #9839, opened 2020-08-26
Import vault-plugin-mock

kalafut:
Support testing of CI and GitHub actions by creating a real dependency
between Vault and a plugin. The plugin itself is a no-op.


ncabatoff:
No vendoring?

kalafut:
@ncabatoff Good call. This also revealed a not ideal go.sum in the mock plugin, which I fixed as well.

PR #9842, opened 2020-08-26
Ui/transform delete v2

Monkeychip:
This PR addresses the following:

* Modifies the "show" route to handle edit and delete functionality of a transformation.
image

* implements the delete modal and handles any edge case errors.
delete

* will concat the view when there are greater than 15 roles (or 15 items in an array passed to the item-list). If there are more than 15 items, it will display the first 5, and then tell how many more there are. This only happens if you pass in the new property `type` to the `info-table-rows` component and the type is `array`. 

image

* dynamically fill in the cli copy clipboard commands for encode and decode for a transformation. (this follows the google doc ivana created for the varations of cli commands based on the type of transformation).  Below is an example when I remove any allowed_roles and when I change the tweak source.  There are other variations such as when it's type=masking that I have not demoed.

cli

### DIFF
--- a/ui/app/templates/components/transform-show-transformation.hbs
+++ b/ui/app/templates/components/transform-show-transformation.hbs
@@ -18,8 +18,8 @@
       </span>
     </div>
     <div class="copy-text level">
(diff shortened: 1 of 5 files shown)


Monkeychip:
> transform-show-transformations

I'll hit this up in the next PR.  Good point!

PR #9852, opened 2020-08-28
Ui/transform roles list create

chelshaw:
This PR sets up the subtype models for concepts under the transform secret engine. 

- Can see list of roles, templates, and alphabets when you click on corresponding tab inside a transform secrets engine
- Cannot click on items in list other than transformations
- Can create a new transform role from the empty state or toolbar
- Creating a role redirects to the view of that role
- Breadcrumb links on transform roles work
- Role create form handles error

Needs: 

- [ ] Tests for new component updated

role-empty-error
create-role


chelshaw:
> This looks great! nice work. I qa'd the branch and everything looked good to me.
> 
> EDITED: One thing, they're's a failing ui test.

As we discussed offline, I went ahead and skipped the test in question so we can merge this and unblock the other work. The tests will be included on the next chunk of roles work (click from list to show)

PR #9856, opened 2020-08-31
docs: add ldap ppolicy to enforce password hashing

jasonodonnell:
This adds a note to the OpenLDAP doc about using an LDAP ppolicy to the LDAP server for enforcing plaintext passwords to be hashed by the server. The consequence of not using this results in plaintext passwords being stored in LDAP.

### DIFF
--- a/website/pages/docs/secrets/openldap/index.mdx
+++ b/website/pages/docs/secrets/openldap/index.mdx
@@ -69,8 +69,8 @@ values in LDAP. This behavior can cause plaintext passwords to be stored in LDAP
 To avoid having plaintext passwords stored an LDAP password policy (ppolicy) should 
 be configured on the LDAP server to enforce hashing or encryption by default.
 
-The following is an example of a password policy to enforce hashing on data information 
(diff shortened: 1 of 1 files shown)


website/pages/docs/secrets/openldap/index.mdx
```diff
@@ -61,6 +61,34 @@ This plugin currently supports LDAP v3.
     $ vault read openldap/static-role/hashicorp
     ```
 
+## LDAP Password Policy
+
+The OpenLDAP secret engine does not hash or encrypt passwords prior to modifying 
+values in LDAP. This behavior can cause plaintext passwords to be stored in LDAP.
+
+To avoid having plaintext passwords stored an LDAP password policy (ppolicy) should
```
pcman312, line 69: Can you add a blurb explicitly stating this is referring to the LDAP password policies and not the Vault password policies? Maybe something like "... not to be confused with Vault password policies..."

jasonodonnell: Done.

PR #9869, opened 2020-09-01
Add the ability to export TOTP secret key

calj:
This PR expose the TOTP secret on a new path /totp/export/{key_name}
This gives the ability to export TOTP (and thus import).


jsalatiel:
Can this be merged ?

ipaqmaster:
Any chance of this one seeing the light? Sucks to have Vault absorb totp keys but no way to export them or move them to a different secret engine path given the API doesn't provide a method to read out at all. Only hope right now is finding the write in shell history...

PR #9871, opened 2020-09-01
docs: add injector tls setup

jasonodonnell:
This adds instructions on installing the Vault Agent Injector using user generated certificates. Out of the box the injector supports auto-tls where it generates these for the user, however, it also supports manual certificate injection (which this document is describing).

### DIFF
--- a/website/pages/docs/platform/k8s/helm/examples/injector-tls.mdx
+++ b/website/pages/docs/platform/k8s/helm/examples/injector-tls.mdx
@@ -60,7 +60,7 @@ $ openssl req \
    -subj "/C=US/ST=CA/L=San Francisco/O=HashiCorp/CN=vault-agent-injector-svc"
 ```
 
-After creating the CSR, configure an extension file to configure additional parameters for signing 
+After creating the CSR, create an extension file to configure additional parameters for signing 
 the certificate.
 
 ~> **Important Note:** The alternative names for the certificate must be configured to use the name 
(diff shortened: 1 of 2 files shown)


website/pages/docs/platform/k8s/helm/examples/injector-tls.mdx
```diff
@@ -0,0 +1,132 @@
...
+```shell
+$ export CA_BUNDLE=$(cat injector-ca.crt | base64)
+```
+
+Finally, install the Vault Agent Injector with the following custom values:
+
+```shell
+$ helm install vault hashicorp/vault \
+  --set="injector.certs.secretName=injector-tls" \
+  --set="injector.certs.caBundle=${CA_BUNDLE?}"
```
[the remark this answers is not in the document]
jasonodonnell: This is a good idea, thanks!

website/pages/docs/platform/k8s/helm/examples/injector-tls.mdx
```diff
@@ -0,0 +1,133 @@
...
+  -CA injector-ca.crt \
+  -CAkey injector-ca.key \
+  -CAcreateserial \
+  -out tls.crt \
+  -days 1825 \
+  -sha256 \
+  -extfile csr.conf
+```
+
+~> **Important Note:** Values such as days (how long the certificate is valid for) should be configured for your environment.
```
catsby, line 102: In the previous example where this `Note` is given, it's placed before the command example, but here we place it afterwards. Should we make them consistent and either both before or both after?

jasonodonnell: Good point, done!

PR #9873, opened 2020-09-02
Retry a KV put command if the message indicates kv-v2 upgrade.

mgritter:
Tested via causing upgrade to delay five seconds:

```
diff --git a/go.mod b/go.mod
index 78d4cfd3b..c20d1c59c 100644
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,8 @@ replace github.com/hashicorp/vault/api => ./api
 
 replace github.com/hashicorp/vault/sdk => ./sdk
 
+replace github.com/hashicorp/vault-plugin-secrets-kv => ./vendor/github.com/hashicorp/vault-plugin-secrets-kv
+
 require (
        cloud.google.com/go v0.56.0
        cloud.google.com/go/spanner v1.5.1
diff --git a/vendor/github.com/hashicorp/vault-plugin-secrets-kv/upgrade.go b/vendor/github.com/hashicorp/vault-plugin-secrets-kv/upgrade.go
index 35893921e..36c9a92f6 100644
--- a/vendor/github.com/hashicorp/vault-plugin-secrets-kv/upgrade.go
+++ b/vendor/github.com/hashicorp/vault-plugin-secrets-kv/upgrade.go
@@ -85,6 +85,7 @@ func (b *versionedKVBackend) Upgrade(ctx context.Context, s logical.Storage) err
                return errors.New("upgrade already in process")
        }
 
+       time.Sleep(5 * time.Second)
        // If we are a replication secondary or performance standby, wait until the primary has finished
        // upgrading.
        if b.perfSecondaryCheck() {
```

### DIFF
--- a/command/kv_test.go
+++ b/command/kv_test.go
@@ -21,6 +21,33 @@ func testKVPutCommand(tb testing.TB) (*cli.MockUi, *KVPutCommand) {
 	}
 }
 
+func retryKVCommand(t *testing.T, client *api.Client, args []string) (code int, combined string) {
+	t.Helper()
+
+	// Loop until return message does not indicate upgrade, or timeout.
+	timeout := time.After(20 * time.Second)
+	for true {
+		ui, cmd := testKVPutCommand(t)
+		cmd.client = client
+		code = cmd.Run(args)
+		combined = ui.OutputWriter.String() + ui.ErrorWriter.String()
+
+		// This is an error if a v1 mount, but test case case doesn't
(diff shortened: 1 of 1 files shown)


mgritter:
I notice that the other functions in the same file all have a call to time.Sleep(time.Second), which will almost always work.

We could do something similar in those cases in the initial Write and remove the mandatory sleep.

sgmiller:
Like the retry better to keep the unit tests running as quickly as possible.  Approved.

PR #9876, opened 2020-09-02
Consolidate locking for sys/health

jefferai:
This avoids a second state lock read-lock on every sys/health hit


vault/ha.go
```diff
@@ -72,6 +72,16 @@ func (c *Core) PerfStandby() bool {
 	return perfStandby
 }
 
+// HealthParams is meant as a way to avoid some extra locking on the very
+// common sys/health check
+func (c *Core) HealthParams() (standby, perfStandby bool) {
```
calvn, line 77: Nit: maybe rename to `StandbyStates`?

vishalnayak (approved):

PR #9878, opened 2020-09-02
[DBPW 4/X] Update DB engine to support v4 and v5 interfaces with password policies

pcman312:
# Overview

Updates the combined database secrets engine to serve both Database v5 and v4 interfaces. Most of the logic is located in `versioning.go`. This adds a new type: `databaseVersionWrapper` that is essentially a one-of type. It can either contain the v4 Database type or a v5 Database type. It should never contain both. This is then used everywhere the v4 Database was previously referenced in order to switch on the version of the interface. This was the simplest approach I could come up with since the two Database interfaces don't have useful overlap that would allow us to define a shared interface.

If using the v5 interface, this also supports using password policies for password generation. Setting `password_policy=<string>` when configuring the database will use the specified password policy for password generation. This is a change in current behavior. In the v4 Database, the database plugin is responsible for generating passwords for dynamic & static users, as well as the root user. In the v5 Database, this responsibility is handled by Vault such that passwords are generated in Vault & passed into the plugin's functions.

- [Database v4](https://github.com/hashicorp/vault/blob/master/sdk/database/dbplugin/)
- [Database v5](https://github.com/hashicorp/vault/blob/master/sdk/database/newdbplugin/)

# PR Size
This PR looks huge (20k+ lines!) but it's actually much smaller. Most of those lines are from vendoring:
```
$ git diff --stat dbpw-03-plugin dbpw-04-engine -- :/\!vendor\/\*
 builtin/logical/database/backend.go                 |  38 +++---
 builtin/logical/database/backend_test.go            |   6 +-
 builtin/logical/database/mocks_test.go              | 101 +++++++++++++++
 builtin/logical/database/path_config_connection.go  |  49 ++++---
 builtin/logical/database/path_creds_create.go       |  16 +--
 builtin/logical/database/path_rotate_credentials.go |  59 +++++----
 builtin/logical/database/rollback.go                |  26 +---
 builtin/logical/database/rollback_test.go           |  16 +--
 builtin/logical/database/rotation.go                |  19 +--
 builtin/logical/database/secret_creds.go            |  40 +++---
 builtin/logical/database/versioning.go              | 292 ++++++++++++++++++++++++++++++++++++++++++
 builtin/logical/database/versioning_large_test.go   | 310 +++++++++++++++++++++++++++++++++++++++++++++
 builtin/logical/database/versioning_test.go         | 666 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 go.mod                                              |   2 +-
 go.sum                                              |   4 +
 vault/plugin_catalog.go                             |  18 ++-
 16 files changed, 1520 insertions(+), 142 deletions(-)
```

# Prerequisites
- [ ] https://github.com/hashicorp/vault/pull/9745 merged
- [ ] Change base branch to `master`
- [ ] Rebased against `master`

# Related PRs


builtin/logical/database/backend.go
```diff
@@ -30,7 +29,7 @@ const (
 
 type dbPluginInstance struct {
 	sync.RWMutex
-	dbplugin.Database
+	database databaseVersionWrapper
```
calvn, line 32: Wouldn't this make `dbPluginInstance` lose all the methods that it inherited from the embedded interface before this change?

builtin/logical/database/versioning.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/hashicorp/vault/helper/random"
+	"github.com/hashicorp/vault/sdk/database/dbplugin"
+	"github.com/hashicorp/vault/sdk/database/newdbplugin"
+	"github.com/hashicorp/vault/sdk/helper/pluginutil"
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+type databaseVersionWrapper struct {
+	database       newdbplugin.Database
+	legacyDatabase dbplugin.Database
```
calvn, line 19: Even though there's the intent to leave the v4 nomenclature out out of code references, I think that terms like legacy, old, new, or similar have its own set of problems too. If we were to introduce the next iteration of database plugins (which is unlikely but not impossible), this terminology might make things confusing as it would be come the "legacy" database, and the `legacyDatabase` is really a `oldLegacyDatabase`, which would be just as ambiguous if not more so.

builtin/logical/database/backend.go
calvn: Ah, we expose equivalent helpers for these methods instead of calling them directly.

builtin/logical/database/versioning.go
```diff
@@ -0,0 +1,292 @@
...
+		return defaultPasswordGenerator.Generate(ctx, rand.Reader)
+	}
+	return generator.GeneratePasswordFromPolicy(ctx, passwordPolicy)
+}
+
+// /////////////////////////////////////////////////////////////////////////////////
+// Change new user
+// /////////////////////////////////////////////////////////////////////////////////
+
+func createUser(ctx context.Context, dbw databaseVersionWrapper, pg passwordGenerator, statements dbplugin.Statements, displayName, roleName string, expiration time.Time, passwordPolicy string) (username, password string, err error) {
```
calvn, line 112: Can we make these funcs be part of `dbPluginInstance's` methods?

builtin/logical/database/versioning.go
pcman312: Unfortunately I don't have a good way of differentiating between versions other than referring to the gRPC version number. I've switched this over to using that nomenclature within the version wrapping struct since it has more forwards compatibility.

builtin/logical/database/versioning.go
pcman312: This has been put into a separate wrapper struct in `version_wrapper.go` rather than on the `dbPluginInstance`.

vault/plugin_catalog.go
```diff
@@ -66,7 +66,19 @@ func (c *Core) setupPluginCatalog(ctx context.Context) error {
...
 	{
-		// Attempt to run as database plugin
+		// Attempt to run as database V5 plugin
+		client, err := newdbplugin.NewPluginClient(ctx, nil, plugin, log.NewNullLogger(), true)
+		if err == nil {
+			// Close the client and cleanup the plugin process
+			client.Close()
+			return consts.PluginTypeDatabase, nil
+		} else {
+			logger.Warn(fmt.Sprintf("received %s attempting as db plugin, attempting as auth/secret plugin", err))
```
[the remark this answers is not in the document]
pcman312: 👍 Moved Database type checking into its own function

builtin/logical/database/version_wrapper.go
```diff
@@ -0,0 +1,272 @@
...
+		return nil, fmt.Errorf("no underlying database specified")
+	}
+
+	// v5 Database
+	if d.isNewDB() {
+		_, err := d.v5.UpdateUser(ctx, req)
+		return nil, err
+	}
+
+	// /////////////////////////////////////////////
```
calvn, line 119:
```suggestion
```

vault/plugin_catalog.go
```diff
@@ -104,6 +97,28 @@ func (c *PluginCatalog) getPluginTypeFromUnknown(ctx context.Context, logger log
...
+
+	v4Client, err := dbplugin.NewPluginClient(ctx, nil, plugin, log.NewNullLogger(), true)
+	if err == nil {
+		// Close the client and cleanup the plugin process
+		v4Client.Close()
+		return nil
+	}
+	merr = multierror.Append(merr, err)
+
+	return merr
```
calvn, line 119: This ensures that we return a `nil` value and not an interface that's `*multierror.Error` with a nil value.

```suggestion
	return merr.ErrorOrNil()
```

vault/plugin_catalog.go
```diff
@@ -104,6 +97,28 @@ func (c *PluginCatalog) getPluginTypeFromUnknown(ctx context.Context, logger log
...
+func isDatabasePlugin(ctx context.Context, plugin *pluginutil.PluginRunner, logger log.Logger) error {
+	merr := &multierror.Error{}
+	// Attempt to run as database V5 plugin
+	v5Client, err := newdbplugin.NewPluginClient(ctx, nil, plugin, log.NewNullLogger(), true)
+	if err == nil {
+		// Close the client and cleanup the plugin process
+		v5Client.Close()
+		return nil
+	}
+	merr = multierror.Append(merr, err)
```
calvn, line 109: Does the error from there indicate which database version was attempted? If not, might be useful to wrap the error message with this bit of info.

vault/plugin_catalog.go
```diff
@@ -64,18 +64,11 @@ func (c *Core) setupPluginCatalog(ctx context.Context) error {
...
-			client.Close()
-			return consts.PluginTypeDatabase, nil
-		} else {
-			logger.Warn(fmt.Sprintf("received %s attempting as db plugin, attempting as auth/secret plugin", err))
-		}
+	err := isDatabasePlugin(ctx, plugin, logger)
+	if err == nil {
+		return consts.PluginTypeDatabase, nil
 	}
+	logger.Warn(fmt.Sprintf("received %s attempting as db plugin, attempting as auth/secret plugin", err))
```
calvn, line 71: The error might look a bit off in here now since it can be a multierror, so consider appending error to the end of the string. https://play.golang.org/p/rOofMYebB2d

builtin/logical/database/versioning_large_test.go
```diff
@@ -0,0 +1,317 @@
...
+				Storage:   config.StorageView,
+			})
+
+			// Generate credentials
+			req = &logical.Request{
+				Operation: logical.ReadOperation,
+				Path:      fmt.Sprintf("creds/%s", dynamicRoleName),
+				Storage:   config.StorageView,
+			}
+			ctx, cancel = context.WithTimeout(context.Background(), 2*time.Second)
```
calvn, line 159: Could context cancelation due to timeout mask a request/operation that would have failed otherwise? https://play.golang.org/p/5iqBODFFFoC

builtin/logical/database/path_rotate_credentials.go
```diff
@@ -106,49 +103,48 @@ func (b *databaseBackend) pathRotateCredentialsUpdate() framework.OperationFunc
...
+		if err != nil {
+			return nil, err
 		}
 
+		b.deleteWal(ctx, req.Storage, walID)
 		return nil, nil
 	}
 }
 
+func (b *databaseBackend) deleteWal(ctx context.Context, storage logical.Storage, walID string) {
```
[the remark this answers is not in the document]
pcman312: Done. I think at one point I had this doing more but it's not useful as a separate function any more.

builtin/logical/database/rollback.go
```diff
@@ -73,13 +73,13 @@ func (b *databaseBackend) walRollback(ctx context.Context, req *logical.Request,
...
 // rollbackDatabaseCredentials rolls back root database credentials for
-// the connection associated with the passed WAL entry. It will creates
+// the connection associated with the passed WAL entry. It will create
 // a connection to the database using the WAL entry new password in
 // order to alter the password to be the WAL entry old password.
 func (b *databaseBackend) rollbackDatabaseCredentials(ctx context.Context, config *DatabaseConfig, entry rotateRootCredentialsWAL) error {
 	// Attempt to get a connection with the WAL entry new password.
 	config.ConnectionDetails["password"] = entry.NewPassword
-	dbc, err := b.GetConnectionWithConfig(ctx, entry.ConnectionName, config)
+	db, err := b.GetConnectionWithConfig(ctx, entry.ConnectionName, config)
```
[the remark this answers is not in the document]
pcman312: Done

builtin/logical/database/version_wrapper.go
```diff
@@ -0,0 +1,272 @@
+package database
+
+import (
+	"context"
+	"crypto/rand"
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
```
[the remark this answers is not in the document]
pcman312: Done

builtin/logical/database/version_wrapper.go
pcman312: Done

vault/plugin_catalog.go
pcman312: Fixed

vault/plugin_catalog.go
pcman312: I ended up keeping this warning since it was there in the previous version, however I don't think this message makes sense as a warning. Every plugin goes through this path, and if the plugin isn't a database plugin you're going to get a warning saying the plugin isn't a database plugin. What do you think of changing this to a debug (or trace?) log? I'm also tempted to not include the error message at all as I'm skeptical that it's useful in this context. Thoughts?

vault/plugin_catalog.go
pcman312: It does not but that's because the error will only be returned if both v5 and v4 fail.

Holding off on doing anything with this until we figure out what to do with the log message that this error is used in.

builtin/logical/database/versioning_large_test.go
pcman312: As far as I know, a timeout could occur prior to a non-timeout error being returned. In this case, the `err` from `HandleRequest` will be non-nil and be caught by `assertErrIsNil`. I can't think of a way that a timeout would occur that we don't get an error back. The example you have requires that the context cancellation (in this case due to the timeout) doesn't return an error which as far as I can tell isn't the behavior of any of the handlers. Ultimately the timeouts here are to help ensure the test doesn't sit there and do nothing for an extended period of time.

builtin/logical/database/version_wrapper.go
calvn: Did you end up pushing new changes? I still see this on the latest commit.

vault/plugin_catalog.go
calvn: Did you end up pushing new changes? I still see this on the latest commit.

builtin/logical/database/version_wrapper.go
pcman312: Just got pushed.

vault/plugin_catalog.go
pcman312: Just got pushed

vault/plugin_catalog.go
calvn: `getPluginTypeFromUnknown` shouldn't be called that often -- once when setting up the plugin catalog if a plugin has not been upgraded yet, and when a plugin with unknown type is added to the catalog (which shouldn't happen since we're now [enforcing the type](https://www.vaultproject.io/api-docs/system/plugins-catalog#register-plugin) to be provided via the API). The latter is more of a backward compat behavior, so new plugin registration operation shouldn't be hitting this path.

pcman312: For posterity:
@calvn & I talked offline and agreed that we should keep this as a warning, but move it to the end of the function if we are unable to determine the type of the plugin. I also changed it so the message is a bit more descriptive differentiating between an unknown type vs one that isn't recognized by this particular function. That will help guard against adding new types to the `logical.BackendType` enum.

vault/plugin_catalog.go
pcman312: Wrapped the error to indicate which error failed on which database version. This will bubble up to the log message referenced in the next comment thread.

builtin/logical/database/version_wrapper.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return resp, nil
+}
+
+// NewUser in the database. This is different from the v5 Database in that it returns a password as well.
+// This is done because the v4 Database is expected to generate a password and return it. The NewUserReponse
+// does not have a way of returning the password so this function signature needs to be different.
+// The password returned here should be considered the source of truth, not the provided password.
+// Errors if the wrapper does not contain an underlying database.
+func (d databaseVersionWrapper) NewUser(ctx context.Context, req newdbplugin.NewUserRequest) (resp newdbplugin.NewUserResponse, password string, err error) {
```
calvn, line 72: This probably okay as-is, but I also wanted to bring up a suggestion that was tossed around, which is the idea of having the password be part of `newdbplugin.NewUserResponse` (and its proto equivalent), but marking that field as being there solely for backward compatibility purposes. This should address the issue of `NewUser` being unable to return the password string if we had a wrapper around v4. We could probably do something in the same vein (e.g. an `UpdateType` or similar on the request payload) for `UpdateUser` so that the v4 wrapper knows what method to call.

builtin/logical/database/version_wrapper.go
```diff
@@ -0,0 +1,268 @@
...
+		return dbw, nil
+	}
+
+	return dbw, fmt.Errorf("invalid database version")
+}
+
+// Initialize the underlying database. This is analogous to a constructor on the database plugin object.
+// Errors if the wrapper does not contain an underlying database.
+func (d databaseVersionWrapper) Initialize(ctx context.Context, req newdbplugin.InitializeRequest) (newdbplugin.InitializeResponse, error) {
+	if !d.isNewDB() && !d.isLegacyDB() {
```
calvn, line 47: `newDatabaseWrapper` should disallow this from happening right?

builtin/logical/database/backend_test.go
```diff
@@ -289,6 +289,7 @@ func TestBackend_config_connection(t *testing.T) {
 			},
 			"allowed_roles":                      []string{"*"},
 			"root_credentials_rotate_statements": []string{},
+			"password_policy":                    "",
```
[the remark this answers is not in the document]
pcman312: No, password policies are optional.

builtin/logical/database/version_wrapper.go
pcman312: I don't think this is a good idea for three reasons:
1) It conflates the two interfaces. We'd be stuck with the new interface returning a password even if we removed the old interface.
2) It confuses plugin authors. `NewUser` is receiving a password and returning one?
3) We're explicitly moving password generation out of plugins and into Vault. By having the response return a password, we revert that decision.

Unfortunately, we cannot make the version selection code simpler because we are dramatically changing what parts of the system are responsible for what (in particular password generation).

builtin/logical/database/version_wrapper.go
pcman312: Yes, but this is protecting against creating a `databaseVersionWrapper` instance directly without calling `newDatabaseWrapper`.

builtin/logical/database/version_wrapper.go
calvn: That's right, I'm satisfied with the current approach since there are quite a few downsides for little gain going with the other suggestion. Documentation around fields should hopefully help alleviate with the confusion on the return value. However, leaking some of the old behavior that will be unused in v5 and having dangling fields when v4 is removed is certainly non-ideal.

calvn (approved):

builtin/logical/database/version_wrapper.go
```diff
@@ -0,0 +1,268 @@
...
+		return newdbplugin.InitializeResponse{}, fmt.Errorf("no underlying database specified")
+	}
+
+	// v5 Database
+	if d.isNewDB() {
```
calvn, line 52: More of a nit since this is an internal struct, but why not have these methods be called `isV4` and `isV5` since we're referencing by version on the fields already?

pcman312: Good catch - I originally was using `legacy` etc. terminology but missed changing these. Updated.

builtin/logical/database/versioning_large_test.go
calvn: Context cancelation due to timeout could result from an error within the operation handler's storage calls (if not by something else before that). The failure wouldn't really be telling us that it's a failure of the database implementation itself, but rather something related to storage calls being inaccessible. That's less likely to happen in this case since we're using inmem storage, so it's unlikely that we will hit this timeout even on CI.

pcman312: Agreed. This should really only happen in two scenarios: CI is *really* slow or there's something wrong in the handler(s) that needs to be investigated.

builtin/logical/database/backend_test.go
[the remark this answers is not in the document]
calvn: Is there a reason why we're explicitly setting this to `""` on every case? Not sure if I missed this test case, but we could test various permutation of how this value is passed in (missing, empty, valid, invalid).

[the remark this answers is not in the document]
calvn: Nvm, this is the expected value returned!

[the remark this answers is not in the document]
pcman312: The key needs to exist because the `DatabaseConfig` doesn't have `omitempty` on this field.

builtin/logical/database/version_wrapper.go
calvn: Fair enough. I always go back and forth between nil checking or letting the constructor handle sane defaults and proper guarantees (more applicable around the receiver itself when it's a pointer value and not on its fields, but in similar spirit).

builtin/logical/database/version_wrapper_test.go
```diff
@@ -0,0 +1,963 @@
...
+			}
+
+			passGen := fakePasswordGenerator{
+				password: test.passwordPolicyPassword,
+				err:      test.passwordPolicyErr,
+			}
+
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+			defer cancel()
+			password, err := dbw.GeneratePassword(ctx, passGen, "test_policy")
```
calvn, line 303: Looks like this is always testing for a password to be provided at the wrapper layer. Can we also test for `""` so that we make sure the call to `defaultPasswordGenerator.Generate(ctx, rand.Reader)` is exercised?

builtin/logical/database/version_wrapper.go
```diff
@@ -0,0 +1,268 @@
...
+
+type databaseVersionWrapper struct {
+	v4 dbplugin.Database
+	v5 newdbplugin.Database
+}
+
+// newDatabaseWrapper figures out which version of the database the pluginName is referring to and returns a wrapper object
+// that can be used to make operations on the underlying database plugin.
+func newDatabaseWrapper(ctx context.Context, pluginName string, sys pluginutil.LookRunnerUtil, logger log.Logger) (dbw databaseVersionWrapper, err error) {
+	newDB, err := newdbplugin.PluginFactory(ctx, pluginName, sys, logger)
```
briankassouf, line 25: @pcman312 @calvn Since we are switching to a completely new plugin client this would be a perfect time to switch from the TLSProvider to AutoMTLS config on go plugin!

builtin/logical/database/version_wrapper.go
stephane439: @Il Il faut le faire @pcman312

builtin/logical/database/version_wrapper.go
pcman312: Will attempt! This will end up being in a separate PR though since it will be a part of the SDK code rather than the engine.

builtin/logical/database/version_wrapper_test.go
pcman312: Added this test. It was being exercised in another test, but it wasn't explicit.

Jamesedwand0147 (approved):

Issue #6427, opened 2019-03-18
Storage consul: upper case letters in service_tags are lowered


caparisi:
**Describe the bug**
Value set in consul tags should been stricly the same that defined in `service_tags` consul storage config.

**To Reproduce**

1. Setup a consul backend storage with upper case : 
```hcl
storage "consul" {
  address = "<...>:8500"
  path    = "vault"
  service = "vault"
  service_tags = "traefik.enable=true,traefik.frontend.rule=Host:test-vault.mydomain.com"
  token   = "<...>"
}
```
2. Restart / Unseal your vault
3. Check on Consul UI your `vault` service, the `Host` define in vault configuration became `host`

**Expected behavior**
All upper case in `service_tags` should not been converted in low case.

**Environment:**
* Vault Server Version (retrieve with `vault status`):
```
Key                    Value
---                    -----
Seal Type              shamir
Initialized            true
Sealed                 false
Total Shares           5
Threshold              3
Version                1.0.3
Cluster Name           vault-cluster-<...>
Cluster ID             <...>
HA Enabled             true
HA Cluster             https://<...>:8201
HA Mode                standby
Active Node Address    http://<...>:8200
```

* Vault CLI Version (retrieve with `vault version`):
```
Vault v1.0.3 ('85909e3373aa743c34a6a0ab59131f61fd9e8e43')
```

* Server Operating System/Architecture:
```
Centos 7
Docker 18.09.3 
Official vault image vault:1.0.3
```

Vault server configuration file(s):

```hcl
storage "consul" {
  address = "<...>:8500"
  path    = "vault"
  service = "vault"
  service_tags = "traefik.enable=true,traefik.frontend.rule=Host:test-vault.mydomain.com"
  token   = "<...>"
}
listener "tcp" {
  address           = "0.0.0.0:8200"
  cluster_address   = "0.0.0.0:8201"
  tls_disable       = true
}
default_lease_ttl = "1h"
ui = true
api_addr = "http://<...>:8200"
cluster_addr = "https://<...>:8201"
```

**Additional context**
Theses tags are used by a reverse proxy / load balancer [TRAEFIK](https://traefik.io/) which is case sensitive.

deniszh:
👍 
Faced exact same issue, also with Traefik.

luhhujbb:
Seems to be here:

https://github.com/hashicorp/vault/blob/f06bfa3087ab51bf7a12d339fc744e453546aad8/physical/consul/consul.go#L305

[...]

PR #6483, opened 2019-03-27
Support mixed case consul service tags on consul storage engine

ncabatoff:
When support for service tags was added, the only way we had to parse
and dedup a list of strings also forced them to be lowercase.  Now there's
another helper func that doesn't smash the case so use that instead.

Fixes #6427.

PR #9883, opened 2020-09-02
issue-6427

GabeL7r:
Fix https://github.com/hashicorp/vault/issues/6427

Same function minus the lower case functionality: https://godoc.org/github.com/hashicorp/vault/sdk/helper/strutil#ParseDedupAndSortStrings


banks:
Hey, sorry it's been so long! I'm cleaning up old issues and reaslised this PR duplicates #6483 which has some additional context as well as fixing up a test this change would break. I'm going to close in favour of that other PR which has also been open for many years 😬 .

Thanks so much for taking the initiative to add a PR for this though @GabeL7r. We'll work out on that other PR how to resolve this once and for all!

PR #9890, opened 2020-09-04
Validate to/from parameters when remounting a backend

### LINKED ISSUE #9889: Unable to delete vault mount due to HTTP 301 redirect
We ran into a situation where we had a Vault PKI and KV mounts which had a double slash (`//`) which resulted in not being able to manage those mounts appropriately. Specifically, we were unable to delete the mounts after they contained the double slash.

These mounts were created using Terraform's [`vault_mount`](https://registry.terraform.io/providers/hashicorp/vault/latest/docs/resources/mount) and later renamed which resulted in the Terraform provider correctly [moving/remounting](https://github.com/terraform-providers/terraform-provider-vault/blob/0118a1951317b103e33dfa23856d17120a2e4864/vault/resource_mount.go#L147-L159) the backend. Unfortunately, our Terraform configuration was buggy which resulted in our mount path going from `foo/bar` to `pki//foo/bar`.

`vault secrets move <from> <to>` does not validate the `from`/`to` parameters. However, Vault uses Golang's [`http.ServeMux` which will HTTP 301 redirect any double slash](https://github.com/golang/go/blob/93810ac1f4574e1e2a79ea156781bafaf8b8ebe0/src/net/http/server.go#L2383-L2388) to the request path which does not contain a double slash (e.g. `a//b` -> `a/b`). As a result, it is possible to end up in a scenario where a mount is not usable if the mount is moved to an invalid path after it has already been created.

**To Reproduce**
1. Run `vault server -dev -dev-root-token-id=root` in a separate terminal or in the background.
1. Run `export VAULT_ADDR='http://127.0.0.1:8200'`
1. Run `export VAULT_TOKEN=root`
1. Create the mount `vault secrets enable -path=pki/test pki`
1. Move the mount `vault secrets move pki/test pki//test`
1. Tune the mount with `vault read sys/mounts/pki/test/tune` results in 400 error `cannot fetch sysview for path "pki/test/"`
1. Tune the mount with `vault read sys/mounts/pki//test/tune` results in 400 error `cannot fetch sysview for path "pki/test/"`
1. Delete the mount with `vault secrets disable pki/test`
1. Delete the mount with `vault secrets disable pki//test`
1. View the secret backends `vault secrets list` and see the mount is still there

**Workaround**
To work around this you can move the mount back to the original name and then access it:
1. Move the mount back `vault secrets move pki//test pki/test`
1. Delete the mount with `vault secrets disable pki/test`
1. View the secret backend was deleted with `vault secrets list`

**Expected behavior**
A mount should not be allowed to move to a mount path that cannot be accessed. This includes both double slashes and non-printable characters. There are some other edge cases which [`path.Clean()`](https://github.com/golang/go/blob/93810ac1f4574e1e2a79ea156781bafaf8b8ebe0/src/path/path.go#L55-L69) takes care of as well.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.5.3
* Vault CLI Version (retrieve with `vault version`): 1.5.3

mdgreenfield:
Vault uses http.ServeMux which issues an HTTP 301 redirect if the
request path contains a double slash (`//`). Additionally, vault
handles all paths to ensure that the path only contains printable
characters. Therefore use the same validation on the to/from parameters
for remounting.

Not doing this can result in a Vault mount that was originally mounted
at `pki/foo` to being remounted at `pki/foo//bar` resulting in mounts
that cannot be accessed.

fixes #9889

/cc @byxorna @ianferguson


vault/logical_system.go
```diff
@@ -1025,6 +1045,10 @@ func (b *SystemBackend) handleRemount(ctx context.Context, req *logical.Request,
 			logical.ErrInvalidRequest
 	}
 
+	if err = validateMountPath(toPath); err != nil {
+		return handleError(fmt.Errorf("'to' %v", err))
+	}
```
mdgreenfield, line 1050: Note that I did not validate the `fromPath` because if we validate it then any Vault users would not be able to move their mounts back to the original name. Also, because the `fromPath` is the path that was already cleaned by `http.ServeMux` and the non-printable characters handler, `fromPath` should only ever be a valid path when a mount is created.

vishalnayak (approved):
Thanks for submitting the PR!

vault/logical_system.go
```diff
@@ -1015,6 +1017,24 @@ func (b *SystemBackend) handleUnmount(ctx context.Context, req *logical.Request,
 	return nil, nil
 }
 
+func validateMountPath(p string) error {
+	s := path.Clean(p)
+	if p != s {
```
vishalnayak, line 1022: One wrinkle here is that "foo/" which is a valid input will become "foo" by clean which will be thrown as an error.

vishalnayak: This is caught by the tests as well. Can be fixed by retaining the "/" suffix on `s` if it was present in `p`.

vishalnayak: In the interest of time, in order to get this fix in sooner. I'll be merging this PR and fixing the breaking test in a separate PR.

Issue #9567, opened 2020-07-22
Agent ignores kerberos disable_fast_negotiation


sbutler:
**Describe the bug**
The vault agent documentation for the kerberos auto_auth method shows a disable_fast_negotiation option. However, setting this option in the config has no effect and fast negotiation is always attempted.

**To Reproduce**
1. Create a vault agent config
2. Configure auto_auth with disabled kerberos fast negotiation:
```
auto_auth {
        method "kerberos" {
                mount_path = "auth/kerberos"
                config = {
                        krb5conf_path            = "/etc/krb5.conf"
                        keytab_path              = "/etc/krb5.keytab"
                        username                 = "my-principal"
                        service                  = "http/my-server"
                        realm                    = "EXAMPLE.ORG"
                        disable_fast_negotiation = true
                }
        }
}
```
3. Run the agent: `vault agent -config /path/to/config.hcl`

**Expected behavior**
Fast negotiation should be disabled (but it is not).

**Environment:**
* Vault Server Version: v1.5.0
* Vault CLI Version: v1.5.0
* Server Operating System/Architecture: amd64 docker container
* Client Operating System/Architecture: macOS 10.15.6

PR #9892, opened 2020-09-04
agent/auth/kerberos: add disable_fast_negotiation

jasonodonnell:
A documented parameter `disable_fast_negotiation` wasn't being configured on the Kerberos auth-method, resulting in the boolean always being false.  This adds proper support to override the value.

Fixes https://github.com/hashicorp/vault/issues/9567

### DIFF
--- a/command/agent/auth/kerberos/kerberos_test.go
+++ b/command/agent/auth/kerberos/kerberos_test.go
@@ -51,9 +51,8 @@ func TestNewKerberosAuthMethod(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation != false {
-		t.Fatalf("disable_fast_negotation should be false, it wasn't: %t",
-			authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation)
+	if actual := authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation; actual != false {
+		t.Fatalf("disable_fast_negotation should be false, it wasn't: %t", actual)
 	}
 
 	authConfig.Config["disable_fast_negotiation"] = "true"
--- a/command/agent/auth/kerberos/kerberos_test.go
+++ b/command/agent/auth/kerberos/kerberos_test.go
@@ -51,7 +51,8 @@ func TestNewKerberosAuthMethod(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if actual := authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation; actual != false {
+	// False by default
+	if actual := authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation; actual {
 		t.Fatalf("disable_fast_negotation should be false, it wasn't: %t", actual)
 	}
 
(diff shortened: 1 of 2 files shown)


command/agent/auth/kerberos/kerberos_test.go
```diff
@@ -57,11 +79,12 @@ func simpleAuthConfig() *auth.AuthConfig {
...
-			"service":       "HTTP/05a65fad28ef.matrix.lan:8200",
-			"realm":         "MATRIX.LAN",
-			"keytab_path":   "grace.keytab",
-			"krb5conf_path": "krb5.conf",
+			"username":                 "grace",
+			"service":                  "HTTP/05a65fad28ef.matrix.lan:8200",
+			"realm":                    "MATRIX.LAN",
+			"keytab_path":              "grace.keytab",
+			"krb5conf_path":            "krb5.conf",
+			"disable_fast_negotiation": "true",
```
calvn, line 87: Nit: this seems to be an optional parameter, so I'd think that `simpleAuthConfig` shouldn't have this by default. Maybe we could switch the test logic around, and first test for the absence of this param, and then test for it being set like so:

```
...
authConfig.Config["disable_fast_negotiation"] = "true"
authMethod, err = NewKerberosAuthMethod(authConfig)
...
```

calvn (approved):
One small comment, otherwise LGTM!

calvn:
I was looking at the docs on https://www.vaultproject.io/docs/agent/autoauth/methods/kerberos, and it seems that the params are missing indicators on the type and whether the param is optional. Might be a good opportunity to add them in as well, e.g. `disable_fast_negotiation (bool: optional) - ...`.

command/agent/auth/kerberos/kerberos_test.go
jasonodonnell: @calvn It is optional but the Kerberos auth method defines this as a bool, so it will always have a default of `false`: https://github.com/hashicorp/vault-plugin-auth-kerberos/blob/master/cli.go#L135.  So I was testing the override and then removing it to ensure it's false by default.  Thoughts?

calvn: Yep, I was suggesting not setting this in `simpleAuthConfig` so that we can test on the default right away without having to `delete` the value.

If `simpleAuthConfig` didn't have `disable_fast_negotiation` set, we could do something like this for the test which should look a bit simpler:
```go
	authConfig = simpleAuthConfig()
	if authMethod, err := NewKerberosAuthMethod(authConfig); err != nil {
		t.Fatal("err shouldn't be returned for missing disable_fast_negotiation")
	}

	if authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation != false {
		t.Fatalf("disable_fast_negotation should be false, it wasn't: %t", authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation)
	}

	authConfig.Config["disable_fast_negotiation"] = "true"
	authMethod, err := NewKerberosAuthMethod(authConfig)
	if err != nil {
		t.Fatal(err)
	}

	if authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation != true {
		t.Fatalf("disable_fast_negotation should be true, it wasn't: %t",
			authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation)
	}
```

jasonodonnell: Done!

jasonodonnell:
@calvn I'll follow up with a PR to fix the documentation.  Good catch!

command/agent/auth/kerberos/kerberos_test.go
```diff
@@ -46,9 +46,26 @@ func TestNewKerberosAuthMethod(t *testing.T) {
...
-	if _, err := NewKerberosAuthMethod(authConfig); err != nil {
+	authMethod, err := NewKerberosAuthMethod(authConfig)
+	if err != nil {
 		t.Fatal(err)
 	}
+
+	if authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation != false {
+		t.Fatalf("disable_fast_negotation should be false, it wasn't: %t",
+			authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation)
+	}
```
calvn, line 57:
```suggestion
	if actual := authMethod.(*kerberosMethod).loginCfg.DisableFASTNegotiation; actual != false {
		t.Fatalf("disable_fast_negotation should be false, it wasn't: %t", actual)
	}
```

calvn: You could do the same trick below as well :)

PR #9894, opened 2020-09-04
Don't open a raft snapshot file until we have a successful snapshot response.

sgmiller:
Implement a lazy writer which only opens the file once data starts being
written to it, e.g. the snapshot request has succeeded.  This will prevent
opening 0 byte files or clobbering existing files on error.


command/operator_raft_snapshot_save.go
```diff
@@ -71,24 +72,48 @@ func (c *OperatorRaftSnapshotSaveCommand) Run(args []string) int {
...
-	if err != nil {
-		c.UI.Error(fmt.Sprintf("Error opening output file: %s", err))
-		return 2
+	w := &lazyOpenWriter{
+		openFunc: func() (io.WriteCloser, error) {
+			return os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
+		},
 	}
-	defer snapFile.Close()
+	defer w.Close()
```
ncabatoff, line 80: I know this was in the original code, but isn't this a case where we really should be checking the result of the `Close` call?

command/operator_raft_snapshot_save.go
```diff
@@ -71,24 +72,48 @@ func (c *OperatorRaftSnapshotSaveCommand) Run(args []string) int {
...
 		c.UI.Error(fmt.Sprintf("Error taking the snapshot: %s", err))
 		return 2
 	}
 
 	return 0
 }
+
+type lazyOpenWriter struct {
+	openFunc func() (io.WriteCloser, error)
+	writer io.WriteCloser
```
ncabatoff, line 99: It makes me a little uncomfortable that this isn't thread/data-race safe, but it looks like it won't break anything in the existing Sys.RaftSnapshot implementation.

ncabatoff (approved):
Approved, but note Brian's new comment in the jira, which may require more work.

command/operator_raft_snapshot_save.go
sgmiller: Yeah, my thought was this local, unexportable writer is only used in the one case.  If it were reusable I'd stick a mutex in it.

command/operator_raft_snapshot_save.go
sgmiller: Sure?  It makes the code a little uglier but what the heck.

ncabatoff:
Looks good!

PR #9895, opened 2020-09-04
Lower the interval for rotation during tests, to make it more likely that our five second grace period is sufficient.

mgritter:
The unit test was flaky, it looked like adding 5 seconds to the test to accommodate the 5-second rotation ticker was cutting it too close.


mgritter:
Hm, test is failing due to a nil map, but I included the change to sdk/local/testing.go that should initialize it for all these tests.

alexanderbez (approved):

PR #9909, opened 2020-09-08
Fix is-wildcard-string helper to check for classes

Monkeychip:
The wildcard helper was being used on both typed inputs and a select dropdown. The later returned a class that was causing it to error.  It now checks for the class object and returns false in that situation.  The test has also been updated.

### DIFF
--- a/ui/app/helpers/is-wildcard-string.js
+++ b/ui/app/helpers/is-wildcard-string.js
@@ -6,7 +6,11 @@ export function isWildcardString([string]) {
   }
   // string is actually an object which is what comes in in the searchSelect component
   if (typeof string === 'object') {
-    string = Object.values(string)[0];
+    // if the dropdown is used the object is a class which cannot be converted into a string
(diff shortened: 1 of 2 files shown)


ui/app/helpers/is-wildcard-string.js
```diff
@@ -6,7 +6,11 @@ export function isWildcardString([string]) {
   }
   // string is actually an object which is what comes in in the searchSelect component
   if (typeof string === 'object') {
-    string = Object.values(string)[0];
+    // if the dropdown is used the object is a class which cannot be converted into a string
```
chelshaw, line 9: We may want to also mention that we know it is not a wilcard, since being a model means it is a single result? Up to you. Maybe this comment on the PR will suffice :D

chelshaw (approved):
LGTM

ui/app/helpers/is-wildcard-string.js
Monkeychip: Good point.  We'll keep this comment as a reminder.

PR #6294, opened 2019-02-26
UI - fix encoding for user-entered paths

meirish:
By default ember will encode and decode dynamic route segments for you, but there are several places where we're using star dynamic segments which _don't_ come with this behavior so we need to do that encoding and decoding ourselves.

This is also true in ember data - model ids are encoded when you pass them to ember data's `buildUrl` method. There are several places where we're constructing api URLs via string concatenation, and not properly encoding the user-entered items. 

Additionally, when passing the path into the filter box on the secret list page, we were not escaping the path so that we could create a valid Regex object. This meant that any characters that were valid in Regex pattern matching being present in the paths could potentially cause an error.

This PR changes the app to make sure we're encoding when passing elements to star segment urls, decoding when we're pulling params from those same routes, and properly escaping strings before passing them to the `Regex` constructor in the list-controller mixin.

Fixes #6282 .

PR #9910, opened 2020-09-08
Ui/transform role create transformations

chelshaw:
When a role is edited or created, the UI attempts to also attach the role to each transformation that was added, and remove from those that were removed. If the active policy results in failure for those calls, it will flash a message relevant to the action the user had with the transformations (created, added, removed, both). Within this PR there are also a number of updates that ensures the backend is loaded onto the transform models so that we can appropriately fetch relevant capabilities.

Creating a role and adding a transformation that previously had zero roles:
automatic-add-role

Editing the role to add a different transformation that had zero roles, and removing the old one:
role-add-delete-transformation

Creating a role when the policy disallows updating transformations in the secret engine:
policy-create-role

Editing a role when the policy disallows updating transformations in the secret engine:
policy-add-remove-role


ui/app/adapters/transform/base.js
```diff
@@ -42,7 +42,10 @@ export default ApplicationAdapter.extend({
   fetchByQuery(query) {
     const { backend, modelName, id } = query;
     return this.ajax(this.url(backend, modelName, id), 'GET').then(resp => {
-      return resp;
+      return {
+        ...resp,
+        backend,
```
chelshaw, line 47: This is required to attach the backend to a transform/role (or any other nested transform object)

ui/app/adapters/transform.js
```diff
@@ -66,22 +66,17 @@ export default ApplicationAdapter.extend({
 
       results.forEach(result => {
         if (result.value) {
-          if (result.value.data.roles) {
```
chelshaw, line 69: Removed this if-block, the rest is whitespace

ui/app/components/transform-edit-base.js
```diff
@@ -54,45 +40,33 @@ export default Component.extend(FocusOnInsertMixin, {
...
+  persist(method, successCallback) {
+    const model = get(this, 'model');
+    return model[method]().then(() => {
+      successCallback(model);
+    });
+  },
 
-  onEscape(e) {
-    if (e.keyCode !== keys.ESC || this.get('mode') !== 'show') {
+  applyChanges(type, callback = () => {}) {
```
chelshaw, line 50: By moving the createOrUpdate logic out of actions, we can access it from any component that extends this one, which will ensure our behavior is consistent across transform items

ui/app/routes/vault/cluster/secrets/backend/list.js
```diff
@@ -54,7 +54,7 @@ export default Route.extend({
   beforeModel() {
     let secret = this.secretParam();
     let backend = this.enginePathParam();
-    let { tab } = this.paramsFor('vault.cluster.secrets.backend');
+    let { tab } = this.paramsFor('vault.cluster.secrets.backend.list-root');
```
chelshaw, line 57: We should assign this to a separate variable since it was previously used in [PR 6294](https://github.com/hashicorp/vault/pull/6294/files)

chelshaw: Checked this, and getting `tab` from `list-root` still works for the other scenario it's used in, which is PKI

ui/app/components/transform-role-edit.js
```diff
@@ -1,3 +1,130 @@
...
+    createOrUpdate(type, event) {
+      event.preventDefault();
+
+      this.applyChanges('save', () => {
+        const roleId = this.get('model.id');
+        const newModelTransformations = this.get('model.transformations');
+
+        if (!this.initialTransformations) {
+          this.handleUpdatedTransformations(
+            newModelTransformations.map(t => ({
```
Monkeychip, line 99: nice!

Monkeychip (approved):
This looks awesome.  Thanks for tackling and looking into the permissions.  Is the next step to create a role from the transformation when added to allowed_roles?

chelshaw:
> This looks awesome. Thanks for tackling and looking into the permissions. Is the next step to create a role from the transformation when added to allowed_roles?

Yes, which is a different ticket under the `transformations` epic IIRC. It will also be slightly more complicated since the roles might not exist, so I went ahead and deferred it to the next PR :)

PR #9911, opened 2020-09-08
Add a time type for use in APIs.

mgritter:
Time format must be RFC3339, or an epoch timestamp.


mgritter:
Dangit, I need to vendor these changes.

alexanderbez (approved):

raskchanky (approved):

vendor/github.com/hashicorp/vault/sdk/helper/parseutil/parseutil.go
```diff
@@ -67,6 +67,54 @@ func ParseDurationSecond(in interface{}) (time.Duration, error) {
...
+			break
+		}
+		return t, errors.New("could not parse string as date and time")
+	case json.Number:
+		epochTime, err := inp.Int64()
+		if err != nil {
+			return t, err
+		}
+		t = time.Unix(epochTime, 0)
+	case int:
```
swayne275, line 100: nitpick: you could have something like `case int, int32, uint, uint32, uint64` instead of so many switch cases doing the same thing?

also i guess the other weird thing is going from something close to a maxed out `uint64` would overflow an `int64`, but that's also not really a realistic time

swayne275: ignore this, didn't notice it was already merged

PR #9924, opened 2020-09-10
Convert time to UTC, add to OpenAPI

mgritter:


sdk/framework/field_data.go
```diff
@@ -231,7 +231,7 @@ func (d *FieldData) getPrimitive(k string, schema *FieldSchema) (interface{}, bo
 			if err != nil {
 				return nil, false, err
 			}
-			return time, true, nil
+			return time.UTC(), true, nil
```
swayne275, line 234: does changing this have any big implications for calling code?

mgritter: Well, fortunately nobody uses it yet.  :)

Jim and I had a Slack thread about it, he suggested it would be better to be consistent/conservative in what plugins and storage engines get for this data type.  I don't think anybody should be changing the API semantics based on which time zone appears in the call, so the conversion seems like it will be OK.

swayne275 (approved):

alexanderbez (approved):

PR #9931, opened 2020-09-11
Add disable_mlock env support

jasonodonnell:
This adds environment variable support for setting `disable_mlock`.  Having environment variable support is useful for Kubernetes where Kubelet requires swap to be disabled, so mlock isn't explicitly needed.  We can set this environment variable on the Vault pod and have one less setting in the config.

IDE removed an extra newline as well.

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -801,7 +801,6 @@ type quiescenceSink struct {
 	t *time.Timer
 }
 
-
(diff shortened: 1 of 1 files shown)


raskchanky (approved):
Seems reasonable to me

briankassouf (approved):

PR #9932, opened 2020-09-11
Replace GF(256) division with a constant time impl

sgmiller:
The implementation of division in the finite field GF(2^8) used by the shamir code had a test in order to correct modulo division on a uint8 that made it not constant time, as well as intermediate assignments that the Go compiler was optimizing away.

This removes the conditional modulo in div() and replaces the assignments and branches with ConstantTimeSelects.

Addresses ToB-004.


sgmiller:
WIP, according to the issue, mul isn't constant time either.  Need to look more closely at that side.

sgmiller:
Okay, mul/div/invert all good to go.

shamir/shamir.go
```diff
@@ -88,57 +88,32 @@ func div(a, b uint8) uint8 {
 		panic("divide by zero")
```
briankassouf, line 88: Should this if statement use subtle.ConstantTimeByteEq?

sgmiller: No, that case should never get hit, so having it be a different timing in the panic case from the normal case is fine.  I'll add a comment about it though.

briankassouf (approved):
Looks good!

PR #9942, opened 2020-09-14
Ui/transform info table row roles

Monkeychip:
This PR adds links to arrays passed into the `info-table-row` component.  (See gif at end of PR)

If a wildcard is used as an `allowed_roles` then a count and link to "View all roles" is displayed. (I say `allowed_roles` here for example, but this is a component, so any array passed into this component follows this logic).

If the array of `allowed_roles.length` is >= 10, then the first 5 `allowed_roles` are displayed and a link next to it that says how many more there are and a link to "View all roles."

To accomplish the above, a new component called `info-table-item-array` was created that handles arrays inside of the `info-table-row` component.  

**Note**: there are a lot of situations that I did my best to handle, but I was not able to handle every situation.  I gave preference to the situations I thought more common.  An example situation I did handle is the following:  you have a wild card and an array greater than 10.  The View all roles would normally show up twice in this situation, but I prevent this by checking first if a wildcard string exists in the truncated array 5 length array.   In the following screenshot, "View all roles" shows at the end and not _also_ next to the wildcard.

An example of a situation I do _not_ handle is when you do not have a truncated array but you have a wildcard.  In this situation, I do not include a link to "View all roles."  This is because the conditional flow is as follows: is there a wildcard in the truncated array AND it's the last object of the truncated array? if yes, show "View all roles."  The second conditional check, is the item in the array the last in the array AND the array length is greater than 10, then show "View all roles."  In this situation, none of these conditions are truthy, so the "View all roles" does not show.  Trying to accommodate for this particular situation was difficult because I did not want to show "View all roles" twice, which would have happened if I showed "View all roles" if there were a wildcard and no other checks.

Though, this situation is handled if the wildcard is the last item in the truncated array.  

Here is a walk through gif, where you can link back and forth between roles and transformations.
transform

### DIFF
--- a/ui/app/styles/components/info-table-row.scss
+++ b/ui/app/styles/components/info-table-row.scss
@@ -42,6 +42,10 @@
   .icon-false {
     color: $ui-gray-300;
   }
+
+  a {
+    text-decoration: none;
+  }
 }
 
 .info-table-row:not(.is-mobile) .column {
--- a/ui/lib/core/addon/components/info-table-row.js
+++ b/ui/lib/core/addon/components/info-table-row.js
@@ -19,6 +19,8 @@ import layout from '../templates/components/info-table-row';
  * @param helperText=null {string} - Text to describe the value displayed beneath the label.
  * @param alwaysRender=false {Boolean} - Indicates if the component content should be always be rendered.  When false, the value of `value` will be used to determine if the component should render.
  * @param [type=array] {string} - The type of value being passed in.  This is used for when you want to trim an array.  For example, if you have an array value that can equal length 15+ this will trim to show 5 and count how many more are there
+ * @param [isLink=true] {Boolean} - Indicates if the item should contain a link-to component.  Only setup for arrays, but this could be changed if needed.
+ * @param [modelType=null] {string} - Tells what model you want data for the allOptions to be returned from.  Used in conjunction with the the isLink.
  */
 export default Component.extend({
   layout,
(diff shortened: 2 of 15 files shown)


ui/lib/core/addon/components/info-table-row.js
```diff
@@ -19,7 +19,11 @@ import layout from '../templates/components/info-table-row';
  * @param helperText=null {string} - Text to describe the value displayed beneath the label.
  * @param alwaysRender=false {Boolean} - Indicates if the component content should be always be rendered.  When false, the value of `value` will be used to determine if the component should render.
  * @param [type=array] {string} - The type of value being passed in.  This is used for when you want to trim an array.  For example, if you have an array value that can equal length 15+ this will trim to show 5 and count how many more are there
+ * @param [isLink=true] {Boolean} - Passed through to InfoTableItemArray. Indicates if the item should contain a link-to component.  Only setup for arrays, but this could be changed if needed.
```
Monkeychip, line 22: Though these properties are really for the dependent `info-table-item-array` they are passed through in this component first.  Not sure our thoughts on adding documentation for an essentially nested component in the parent.  My thought is it's helpful, so I added it to this component as well.  Open to feedback.

chelshaw: For the other two specifically this makes sense. My only question is, if we're showing an array (which is a new type on info-table) can we assume isLink is true for now? And if we run into a future case where it's not, we abstract it out in this way?

ui/app/styles/components/info-table-row.scss
```diff
@@ -42,6 +42,10 @@
   .icon-false {
     color: $ui-gray-300;
   }
+
+  a {
```
chelshaw, line 46: Do we want _all_ links to have this styling? If not we should add a class to the array links

chelshaw (approved):
Just a couple minor comments, overall looks good

ui/lib/core/addon/components/info-table-row.js
Monkeychip: Arrays are actually used elsewhere in the app.  The originally if array conditional on info-table-row was there before I changed things in there (See row 32 of `info-table-row`).  As such, I made isLink false by default so I don't impact these other array types.

ui/app/styles/components/info-table-row.scss
Monkeychip: Yeah, I checked with Ivana on this.  I think it's another situation where --I think-- the design system is ahead of our code base.  Ivana mentioned that we don't underline links, but having to add this contradicts that.

glenda32681 (approved):

chelshaw (approved):

PR #9948, opened 2020-09-14
DBPW - Add test helpers for DB v5 interface

pcman312:
# Overview

Adds some test helper functions to call the various functions on the v5 `Database` interface with common assertions (no error, `NewUser` gives a non-empty username). These do not preclude calling the functions directly in a test if needed. These are made available for the common testing use cases.

# Related PRs
[4/X - Database engine](https://github.com/hashicorp/vault/pull/9878)

### DIFF
--- a/sdk/database/newdbplugin/testing/test_helpers.go
+++ b/sdk/database/newdbplugin/testing/test_helpers.go
@@ -2,16 +2,30 @@ package dbtesting
 
 import (
 	"context"
+	"os"
 	"testing"
 	"time"
 
 	"github.com/hashicorp/vault/sdk/database/newdbplugin"
 )
 
+func getRequestTimeout(t *testing.T) time.Duration {
+	rawDur := os.Getenv("VAULT_TEST_DATABASE_REQUEST_TIMEOUT")
+	if rawDur == "" {
+		return 2 * time.Second
+	}
+
+	dur, err := time.ParseDuration(rawDur)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


sdk/database/newdbplugin/testing/test_helpers.go
```diff
@@ -0,0 +1,71 @@
...
+	"testing"
+	"time"
+
+	"github.com/hashicorp/vault/sdk/database/newdbplugin"
+)
+
+func AssertInitialize(t *testing.T, db newdbplugin.Database, req newdbplugin.InitializeRequest) newdbplugin.InitializeResponse {
+	t.Helper()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
calvn, line 14: Can we make this a tunable value? It may be sufficient for certain cases such as mocked databases, but 2 seconds might not be enough if operations end up taking longer than this duration  for certain databases which can happen on CI (depending on load, machine type/size, etc.) and result in flaky tests.

pcman312: I'm hesitant to add too much configurability to these as these are meant for the common use case. If they don't work for a specific database, those tests can fall back to calling the database functions directly.

pcman312: I added the ability to change the timeouts via env variable with a default of 2s.

calvn (approved):

PR #9955, opened 2020-09-15
Ui/transformation edit update roles

chelshaw:
When a transformation is created or updated with `allowed_roles`, after saving the transformation the UI attempts to update each of the roles added or removed. If a role is not found from querying, it will be created with the transformation in `role.transformations`. Note: any roles including a wildcard (`*`) will not trigger the create or update behavior on the role. 

create-role-from-transformation

Generic message that shows when a role is removed that doesn't have any other transformations attached to it (cannot save role without any transformations)
info-message


ui/app/components/transformation-edit.js
```diff
@@ -1,3 +1,114 @@
...
+        if (role.action === 'ADD') {
+          transformations = addToList(transformations, transformationId);
+        } else if (role.action === 'REMOVE') {
+          transformations = removeFromList(transformations, transformationId);
+        }
+        roleStore.setProperties({
+          backend,
+          transformations,
+        });
+        console.log(`Saving ${role.id} with transformations`, transformations);
```
chelshaw, line 29: need to remove

Monkeychip:
This looks great so far.  Ping me when the tests are ready and I'll give a final review.

Monkeychip (approved):
This looks great, I would suggest merging in master as I just merged in my branch.

PR #9958, opened 2020-09-15
Fix identity entity capabilities path

chelshaw:
This issue was causing a bug where the capabilities on an `identity/entity` model had the incorrect `updatePath` due to `identityType` not properly being passed into the apiPath method. This change hardcodes the path for the entity model, since by the model's own definition the type is `entity`. 

Investigation is still needed for why the computed property was not being passed into the apiPath method properly, but this should fix adverse side affects such as tokens with policy including the path `identity/entity/id/*` not having the correct options on an entity dropdown.

Sample policy: 
```
path "identity/entity/id/*"
{
  capabilities = ["list", "read", "delete", "sudo", "create", "update"]
}
```
has-sample-policy


Monkeychip (approved):
Looks good!  Thanks for the quick fix.

PR #9959, opened 2020-09-15
Ui/ember upgrade

Monkeychip:
This PR upgrades Ember from 3.8 to 3.16, effectively bringing us up to Octane.  However, we do not octanify anything in this PR, simply run the upgrade.  This will allow us to develop/test before the 1.6 release to make sure no issues --most likely to occur in dependent libraries-- come up.  

I've done QA and run through the tests and have not hit any errors.


Monkeychip:
Closing. Ran into issues on my branch, Chelsea was able to get it to work on hers and we'll be using that one instead.

PR #9965, opened 2020-09-15
Update vault-plugin-secrets-gcp to v0.6.5

Valarissa:
Just backporting a version update of vault-plugin-secrets-gcp here.


jasonodonnell (approved):
LGTM! :shipit:

PR #9966, opened 2020-09-15
Disable usage metrics on performance standby nodes.

mgritter:
Thought about a unit test but it would have to live in ent?


briankassouf (approved):
Looks good

PR #9967, opened 2020-09-15
Backport 1.5: Disable usage metrics on performance standby nodes. (#9966)

mgritter:


vault/core_metrics.go
```diff
@@ -27,10 +27,13 @@ func (c *Core) metricsLoop(stopCh chan struct{}) {
 		select {
 		case <-emitTimer:
 			c.metricsMutex.Lock()
-			if c.expiration != nil {
+
+			// Emit on active node only
+			if c.expiration != nil && !c.perfStandby {
```
ncabatoff, line 32: Suggest `c.PerfStandby()` instead for locking.

vault/core_metrics.go
```diff
@@ -54,6 +57,11 @@ func (c *Core) metricsLoop(stopCh chan struct{}) {
 			}
 			c.stateLock.RUnlock()
 		case <-identityCountTimer:
+			// Only emit on active node
+			if c.perfStandby {
```
ncabatoff, line 61: Ditto.

vault/core_metrics.go
mgritter: I think that some care is needed because c.stateLock can be held while acquiring c.metricsLock, so we shouldn't acquire them in the reverse order here.

ncabatoff (approved):

PR #9972, opened 2020-09-16
Ember-cli upgrade from ~3.8 to ~3.20

Monkeychip:
This PR upgrades Ember from 3.8 to 3.22, effectively bringing us up to Octane. However, we do not octanify much in this PR other than the things required by the upgrade. This will allow us to develop/test before the 1.7 release to make sure no issues come up.

Used ember-cli-upgrade to run the update:
1. `npm install -g ember-cli-update`
2.  `ember-cli-update --to=~3.20`

Note: Because the test suite appears to run faster we ran into a lot of race condition issues in the tests.  Where we could solve for them we did, mostly adding in `settled()`.  However, in the cases where we couldn't, we confirmed on the application they were not an issue and that individually the tests would pass.  We then skipped the tests with notes/todos about coming back.

Additionally, this upgrade upgraded a lot of other `devDependencies` per the recommendation of Embermaps upgrade article [here](https://embermap.com/video/how-to-upgrade-an-ember-app-or-addon).

**TODO's from within the code:**
- [ ] revisit once figure out how to replace, added during upgrade to 3.20 (`ember/no-new-mixins` & `ember/no-mixins`) **No fix for now**
- [x] double check for linter removed size, page, responsePath, pageFilter as unused
- [x] handle withTestWaiter() which no longer works in upgrade, remove package from package.json **ARG**
- [x] double check: remove get from import but used in computed 
- [x] double check used brace expansion
- [x] something here fix
- [x] removed .volatile() but may have a consequence https://deprecations.emberjs.com/v3.x/#toc_computed-property-volatile **CBS**
- [x] handle clearNonGlobalModels() either refresh or find all models **COLLAB**
- [x] Check backend/credentials.hbs "if actions" (post to frontend channel)
- [x] Try bracket syntax on policies/index `{{navigate-input}}` **CBS**
- [x] handle `no-private-routing-service` **CBS**
- [x] secrets-details.hbs `@secretType` is not defined **ARG**
- [x] Review skipped tests **BOTH**


ui/app/components/flash-message.js
```diff
@@ -1,12 +1,12 @@
...
+import { computed } from '@ember/object';
 import FlashMessage from 'ember-cli-flash/components/flash-message';
 
 export default FlashMessage.extend({
   // override alertType to get Bulma specific prefix
   //https://github.com/poteto/ember-cli-flash/blob/master/addon/components/flash-message.js#L35
   alertType: computed('flash.type', {
     get() {
-      const flashType = getWithDefault(this, 'flash.type', '');
+      const flashType = this.flash.type === undefined ? '' : this.flash.type;
```
chelshaw, line 9: Could simplify to `this.flash.type || ''`

ui/app/components/pricing-metrics-dates.js
```diff
@@ -69,7 +69,7 @@ export default Component.extend({
     }
   }),
 
-  showResultsWarning: computed('resultStart', 'resultEnd', function() {
+  showResultsWarning: computed('queryEnd', 'queryStart', 'resultEnd', 'resultStart', function() {
```
chelshaw, line 72: This might be unintended behavior: when you change the query and it doesn't match the result, one would _not_ expect the warning to show up before query is processed

ui/app/components/status-menu.js
```diff
@@ -23,9 +23,7 @@ export default Component.extend({
     return glyphs[this.type];
   }),
   activeCluster: computed('auth.activeCluster', function() {
-    return this.get('store').peekRecord('cluster', this.get('auth.activeCluster'));
-  }),
-  currentToken: computed('auth.currentToken', function() {
-    return this.get('auth.currentToken');
+    return this.store.peekRecord('cluster', this.auth.activeCluster);
   }),
+  currentToken: reads('auth.currentToken'),
```
chelshaw, line 28: Double-check: does this do the same thing as the old code?

ui/app/components/transit-key-actions.js
```diff
@@ -65,37 +65,40 @@ export default Component.extend(TRANSIT_PARAMS, {
   onRefresh() {},
   init() {
     this._super(...arguments);
+
+    // eslint-disable-next-line ember/no-get
     if (get(this, 'selectedAction')) {
```
chelshaw, line 70: We should probably figure out why this needs the `get` at some point

ui/app/controllers/vault/cluster.js
```diff
@@ -24,19 +25,18 @@ export default Controller.extend({
 
   namespaceQueryParam: '',
 
-  /* eslint-disable-next-line ember/no-observers */
   onQPChange: observer('namespaceQueryParam', function() {
```
chelshaw, line 28: This only needs to be an observer because it's based on a computed property, so we could potentially get away with computing on `clusterController.namespaceQueryParam` which is what namespaceQueryParam is an alias of

ui/codemods.log
```diff
@@ -0,0 +1,505 @@
+2020-09-24T17:50:46.692Z [warn] WARNING: {{capitalize}} was not converted as it has positional parameters which can't be automatically converted. Source: app/templates/partials/form-field-from-model.hbs
```
chelshaw, line 1: I feel like we could probably delete this file, but @Monkeychip you know more about the codemod process than I do

Monkeychip: Yes, it's safe to delete. I'll go ahead and remove.

ui/app/components/transit-key-actions.js
Monkeychip: I'll go ahead and add a comment.

ui/app/components/status-menu.js
chelshaw: answer: yes, I was reading the diff wrong

chelshaw (approved):

PR #9973, opened 2020-09-16
Fix flaky TestCluster_ForwardRequest tests. 

ncabatoff:
We wait to see that clients are aware of a new leader before we hit them, and we wait to see that the leader has setup the forwarding handler.

### DIFF
--- a/vault/cluster_test.go
+++ b/vault/cluster_test.go
@@ -289,6 +289,9 @@ func testCluster_Forwarding(t *testing.T, cluster *TestCluster, oldLeaderCoreIdx
 
 		time.Sleep(100 * time.Millisecond)
 	}
+	if ready != 2 {
+		t.Fatal("standbys have not discovered the new active node in time")
+	}
 
 	for i := 0; i < 3; i++ {
 		if i != newLeaderCoreIdx {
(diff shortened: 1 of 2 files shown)


vault/cluster_test.go
```diff
@@ -232,107 +232,77 @@ func testCluster_ForwardRequestsCommon(t *testing.T, clusterOpts *TestClusterOpt
...
+		if i != oldLeaderCoreIdx && i != newLeaderCoreIdx {
+			_ = cluster.Cores[i].StepDown(context.Background(), &logical.Request{
+				Operation:   logical.UpdateOperation,
+				Path:        "sys/step-down",
+				ClientToken: rootToken,
+			})
+			time.Sleep(clusterTestPausePeriod)
+		}
+	}
+
```
mgritter, line 271: This looks good, but because we now lack a line number, could we include a t.Log() indicating who the leader is supposed to be, or some other indication of how far in the test case we are?

mgritter (approved):

vault/cluster_test.go
ncabatoff: Sounds reasonable.  Does the new commit satisfy your request?

mgritter: Yup!

PR #9975, opened 2020-09-16
Move the code that creates Consul containers out of teststorage. 

ncabatoff:
This allows importers of teststorage that don't need consul to run as a non-docker test.


HridoyRoy (approved):
👍 LGTM

PR #9989, opened 2020-09-18
CRUD for transform alphabets

chelshaw:
Adds CRUD actions in the UI for Transform Alphabets. Also adds error message if someone tries to delete a template OR an alphabet that is currently in use. 

Create and edit an alphabet
create-edit-alphabet

Error shown after attempting delete on an alphabet in use. Similar message shown for templates in use. 
error-alphabet

Tooltip shown on list for built-in alphabets. Similar one is shown for built-in templates. 
tooltip-alphabet


Monkeychip (approved):
Great work on this!  I got to play around a lot on this branch for the demo prep and things worked so smoothly.

PR #9990, opened 2020-09-18
Update AD secret engine docs for root cred rotation

Valarissa:
This is a PR to update documents for changes made in https://github.com/hashicorp/vault-plugin-secrets-ad/pull/70


austingebauer (approved):
LGTM!

PR #9995, opened 2020-09-21
Packagespec v0.1.2/master

samsalisbury:
**Concurrent with merging this PR, all internal Vault engineers should update to [packagespec v0.1.2](https://github.com/hashicorp/packagespec/releases/tag/v0.1.2) (download binary for Mac or Linux there).**

Updates to packagespec v0.1.2. See packagespec changelog (internal only) for details. Notable changes visible in this PR:

- Packagespec-specific code from the main Makefile has been moved out to its own `packagespec.mk` which counts as generated code. This file is managed by packagespec.
- The Dockerfiles in `packages-oss.lock/layers` are gone. These files are now written on-the-fly as needed, to a temp directory. This means a lot of deletions in this PR, but will result in much smaller diffs in future.
- No changes to the spec file `packages-oss.yml`
- The file `.circleci/config/@build-release.yml.tpl` is gone. This template is now owned by packagespec.
- One additional/unrelated tidy-up in `Makefile`: the check for and removal of old husky git hooks is removed in this PR, it has been in place long enough that we can assume that no one still has them installed.

Reviewers: The main file to review is the `Makefile` (most of which is deletions) and `packagespec.mk` which contains the migrated code. In the CircleCI config, you will notice that the cache prefix `cache002` has been rolled back to `cache001`. This is safe, because all layerIDs and packagespecID checksums are also updated, invalidating all older caches. That cache prefix is now owned by code in packagespec rather than this repository.


sarahethompson (approved):
LGTM

Issue #5916, opened 2018-12-07
New Templating Parameters for approle/ldap Auth Methods


lcgkm:
How to write the templated ACL policies for AppRole?
Current Available Templating Parameters don't have role_name... 
https://www.vaultproject.io/guides/identity/policy-templating.html

For approle token, we can find following info:
token_meta_role_name=\<\<approle name\>\>

For ldap token, we can find following info:
token_meta_username=\<\<ldap username\>\>

So, can we use token meta data as policy templating parameter? For example:
token.metadata.\<\<metadata key\>\>

Reference:
https://learn.hashicorp.com/vault/identity-access-management/policy-templating#available-templating-parameters

chrishoffman:
Token metadata will likely not be added to the available template parameters.  For both of the use cases above, you can leverage the identity system.  Each AppRole will create an entity in the identity system with a mapping to the role id.  LDAP has a mapping to the username.

Check out these guides for more information:
https://learn.hashicorp.com/vault/identity-access-management/iam-identity
https://learn.hashicorp.com/vault/identity-access-management/policy-templating

lcgkm:
Hi @chrishoffman ,

Sorry for bother you.

> Each AppRole will create an entity in the identity system with a mapping to the role id. 
Could you provide more details about this? How about the entity lifecycle?
And I can find that AppRole will create an entity and entity alias in identity system, the entity alias as same as the AppRole's role id.
But how about AppRole's role name?
Can we add AppRole's role name to entity alias meta?

Best Regards,
CG Li

chrishoffman:
@sielaq Yes, that is true but these have to be encoded into each backend and do not provide centralized management available to all backends.  We have seen requests in the past to extend token metadata to additional backends but have decided not to pursue it because it has to be customized per backend.  The identity system provides the same functionality but centralizes all the functionality and makes it available to all auth methods.

sielaq:
@chrishoffman thanks for clarification. 
The only problem I have (probably different from @lcgkm, but still around this issue)
 is: that we use very cool project
https://github.com/nemosupremo/vault-gatekeeper
to let application accessing properties/secrets...

Using vault-gatekeeper you can match the app name (that runs in mesos) with AppRole (role_name),
and you get temporal token back for accessing own secrets.  In that moment app, that want to access own-namebased-secret, require a policy per application. I cannot create a simple one policy like:
```
path "secrets/{{identity.entity.name}}" {
    capabilities = ["read"]
}
```
[...]

PR #10000, opened 2020-09-21
Allow token metadata for templating

jramosf:
This pull request adds functionality of using token metadata for templated policies, similarly to the current available templates for metadata in stored in entities.

**RATIONALE:**

I am aware this was discussed in the past (https://github.com/hashicorp/vault/issues/5916#issuecomment-448298263) and HashiCorp's response was to keep templated policies centralized with the Identity secrets engine, the solution Vault provides for identities.

However, Identity does not provide **Implicit Groups**, or automated group sync, just [Implicit Entities](https://www.vaultproject.io/docs/secrets/identity#implicit-entities). The rationale for this change is substantially simplifying many workflows (usually cronjobs) that need to sync external groups into "Identity" groups via leveraging more metadata than just the one coming from the Entity.

Furthermore, when setting up an auth backend we need to specify both the `metadata` **and** `policies`. Therefore, we cannot have confusion here, since prior to using the policy in an Auth backend, we will know if metadata needs to be supplied as well. The same goes when using metadata in from an Entity, such as `{{identity.entity.metadata.color}}`

### DIFF
--- a/sdk/helper/identitytpl/templating.go
+++ b/sdk/helper/identitytpl/templating.go
@@ -29,6 +29,7 @@ type PopulateStringInput struct {
 	ValidityCheckOnly bool
 	Entity            *logical.Entity
 	Groups            []*logical.Group
+	TokenMeta         map[string]string
 	NamespaceID       string
 	Mode              int       // processing mode, ACLTemplate or JSONTemplating
 	Now               time.Time // optional, defaults to current time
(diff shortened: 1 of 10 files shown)


canidam:
Any chance this gets merged? It would be very convenient to manage approle policies in such way.

jramosf:
> Any chance this gets merged? It would be very convenient to manage approle policies in such way.

I've rebased the merge request to address the conflict (my original commit was 3 months ago) and the pipeline is passing, as for the approval I'm afraid I'm not part of the maintainers.

jramosf:
I'm closing the PR as it's been over a year without activity. Please feel free to reopen if this has a chance to get merged. Thanks!

PR #10007, opened 2020-09-21
Integrate Allowed and Denied Parameters With KV2 [VAULT-616]

HridoyRoy:
Jira ticket: https://hashicorp.atlassian.net/browse/VAULT-616

Description: Vault requests to the KV2 secrets engine contain an extra "data" field, such that request.Data has a field "data" within which the secret key/values are stored, as we can see here: https://github.com/hashicorp/vault/blob/7aa1ffa92ee61b977efad1488b8f309b1e2136df/command/kv_get.go#L128 

Allowed and denied parameters are put within that data block, and therefore are not found up by acl.go, which assumes the format of request.Data is compatible with KV1. 

This PR provides a fix, and adds tests to the current test suite to include KV2-integrable tests.

### DIFF
--- a/vault/acl.go
+++ b/vault/acl.go
@@ -318,6 +318,13 @@ func (a *ACL) Capabilities(ctx context.Context, path string) (pathCapabilities [
 func (a *ACL) AllowOperation(ctx context.Context, req *logical.Request, capCheckOnly bool) (ret *ACLResults) {
 	ret = new(ACLResults)
 
+	// check if format of kv request is a v1 or v2 request:
+	// https://github.com/hashicorp/vault/blob/7aa1ffa92ee61b977efad1488b8f309b1e2136df/command/kv_get.go#L128
+	requestData := req.Data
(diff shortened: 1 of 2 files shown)


vault/acl.go
```diff
@@ -318,6 +318,13 @@ func (a *ACL) Capabilities(ctx context.Context, path string) (pathCapabilities [
 func (a *ACL) AllowOperation(ctx context.Context, req *logical.Request, capCheckOnly bool) (ret *ACLResults) {
 	ret = new(ACLResults)
 
+	// check if format of kv request is a v1 or v2 request:
```
ncabatoff, line 321: I don't think we should put secret-engine-specific code into generic core code like acl.go.

HridoyRoy:
Closing this PR after discussing the design decisions this change would impact. Kv secrets engine formatting decisions should be encapsulated in the http layer and the kv secrets engine. Vault should the request data map abstractly, instead of making assumptions about the format of the request based on the secrets engine output.

PR #10008, opened 2020-09-21
[DBPW 5/X] Use AutoMTLS with DB plugins

pcman312:
# Overview

Uses [go-plugin's](https://godoc.org/github.com/hashicorp/go-plugin#ClientConfig) `AutoMTLS` feature for DB v5 plugins. This adds a new function `RunConfig` to the `pluginutil` package alongside `Run` and `RunMetadataMode`. This function follows the [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) pattern which will allow us to specify future changes to the `RunConfig` function (such as allowing the user to configure other fields) without actually changing the function signature.

# Related PRs
[4/X - DB engine](https://github.com/hashicorp/vault/pull/9878)

### DIFF
--- a/sdk/helper/pluginutil/run_config.go
+++ b/sdk/helper/pluginutil/run_config.go
@@ -29,7 +29,7 @@ type runConfig struct {
 	autoMTLS       bool
 }
 
-func (rc runConfig) run(ctx context.Context) (*plugin.Client, error) {
+func (rc runConfig) makeConfig(ctx context.Context) (*plugin.ClientConfig, error) {
 	cmd := exec.Command(rc.command, rc.args...)
 	cmd.Env = append(cmd.Env, rc.env...)
 
@@ -39,11 +39,14 @@ func (rc runConfig) run(ctx context.Context) (*plugin.Client, error) {
 	}
 	cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", PluginVaultVersionEnv, version.GetVersion().Version))
 
-	var clientTLSConfig *tls.Config
-	if !rc.isMetadataMode {
-		// Add the metadata mode ENV and set it to false
-		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", PluginMetadataModeEnv, "false"))
+	if rc.isMetadataMode {
+		rc.logger = rc.logger.With("metadata", "true")
+	}
+	metadataEnv := fmt.Sprintf("%s=%t", PluginMetadataModeEnv, rc.isMetadataMode)
+	cmd.Env = append(cmd.Env, metadataEnv)
 
+	var clientTLSConfig *tls.Config
+	if !rc.autoMTLS && !rc.isMetadataMode {
 		// Get a CA TLS Certificate
 		certBytes, key, err := generateCert()
 		if err != nil {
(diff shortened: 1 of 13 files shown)


sdk/helper/pluginutil/run_config.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", PluginVaultVersionEnv, version.GetVersion().Version))
+
+	var clientTLSConfig *tls.Config
+	if !rc.isMetadataMode {
+		// Add the metadata mode ENV and set it to false
+		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", PluginMetadataModeEnv, "false"))
+
+		// Get a CA TLS Certificate
+		certBytes, key, err := generateCert()
```
briankassouf, line 48: I think we want to avoid creating these certs and then response wrapping them if we are using AutoMTLS, right?

pcman312: It doesn't seem to make a difference in terms of using the plugin. I can certainly make this conditional but I defer that decision that knows go-plugin better than I do.

For what it's worth, this is a copy-paste of the previous `runCommon` function, so nothing is changed here other than `AutoMTLS` and how it gets its values.

briankassouf: https://github.com/hashicorp/go-plugin/blob/4c2f0639bfb69d99f6f21b350e87639b4df32df2/client.go#L184-L205

> The client should not set TLSConfig, nor should the server set a TLSProvider, because AutoMTLS implies that a new certificate and tls configuration will be generated at startup.

pcman312: Done. I also included some tests around the ClientConfig creation.

briankassouf (approved):
Looks good to me!

sdk/helper/pluginutil/run_config.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func AutoMTLS(autoMTLS bool) RunOpt {
+	return func(rc *runConfig) {
+		rc.autoMTLS = autoMTLS
+	}
+}
+
+func (r *PluginRunner) RunConfig(ctx context.Context, opts ...RunOpt) (*plugin.Client, error) {
```
calvn, line 148: Nit: This might have been derived from the `runCommon` terminology and to keep things as close as it was before, but it's really creating a plugin client so something like `NewPluginClient` would be more appropriate.

sdk/helper/pluginutil/run_config.go
```diff
@@ -0,0 +1,161 @@
...
+	clientConfig, err := rc.makeConfig(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	client := plugin.NewClient(clientConfig)
+	return client, nil
+}
+
+type RunOpt func(*runConfig)
```
calvn, line 104: Nit: Can we rename this to `RunnerOpt` or `Options` (or `ClientOpt`, see below)? "Run" makes it sound like this is some sort of execution func, especially since this is a typed function.

PR #10010, opened 2020-09-22
Rate Limit Quotas: Allow Exempt Paths to be Configurable

alexanderbez:
Rate limit resource quotas exempt paths can now be configurable. Being that we do not want to introduce any breaking changes since 1.5, Vault uses a default set of paths that exist in 1.5.

/cc @briankassouf @vishalnayak


vault/quotas/quotas.go
```diff
@@ -755,7 +792,10 @@ func (m *Manager) Setup(ctx context.Context, storage logical.Storage, isPerfStan
 	if err != nil {
 		return err
 	}
+
 	m.SetEnableRateLimitAuditLogging(config.EnableRateLimitAuditLogging)
+	m.SetEnableRateLimitResponseHeaders(config.EnableRateLimitResponseHeaders)
```
alexanderbez, line 797: bug fix 🎉

alexanderbez:
Unrelated to this PR, I believe there are race conditions hit in the `Core`'s use of the internal `Manager`. Unsure if this is just related to the tests, but it can't hurt to add a mutex around the Manager/Config calls.

vault/quotas/quotas.go
```diff
@@ -543,6 +552,20 @@ func (m *Manager) SetEnableRateLimitResponseHeaders(val bool) {
...
+// configuration in addition to updating the path manager. Every call to
+// SetRateLimitExemptPaths will wipe out the existing path manager and set the
+// paths based on the provided argument.
+func (m *Manager) SetRateLimitExemptPaths(vals []string) {
+	if vals == nil {
+		vals = []string{}
+	}
+
+	m.config.RateLimitExemptPaths = vals
+	m.rateLimitPathManager = pathmanager.New()
```
mgritter, line 565: I'm not sure if this is the race condition you referred to already, but it seems like this change can occur concurrently with requests coming in. Does pathmanager have its own internal mutex?

Might be slightly better to create the manager locally, add the paths to it, and then swap it into rateLimitPathManager.

mgritter: I'm not able to find how any of the manager's attributes are protected against concurrent access? It seems like SetEnableRateLimitResponseHeaders and SetEnableRateLimitAuditLogging have the same problem, or something is protecting all three that I'm not seeing.

alexanderbez: > I'm not sure if this is the race condition you referred to already, but it seems like this change can occur concurrently with requests coming in. Does pathmanager have its own internal mutex?
> 
> Might be slightly better to create the manager locally, add the paths to it, and then swap it into rateLimitPathManager.

No, it's not the race condition I was referring to based on what I see in the unit tests. However, the point you raise is not specific to the path manager. It's applicable to all three configurations (e.g. enabling/disabling response headers can happen mid-request). This is why I was considering adding a lock around the entire manager's use in `Core`, not just for the path manager.

alexanderbez: > I'm not able to find how any of the manager's attributes are protected against concurrent access?

They're not! I think we need to dig further into this. cc @vishalnayak

vishalnayak: Hey there, pathmanager does have an internal lock of its own. Quota manager also have a lock of its own to protect its attributes. Some functions don't use it directly because the functions from where they are called from are holding the lock needed to protect them.

We might want to look at all the callers and see if they are holding locks properly (in this PR), and probably update the comment on the functions that the callers should hold the lock.

alexanderbez: Taking a quick look I don't see any protection against rate limit application during a config or quota update.

i.e. `ApplyRateLimitQuota` can be called at the same time as `handleRateLimitQuotasUpdate`/`handleQuotasConfigUpdate`. Since the `Manager` has a RW mutex, why not just use that for _all_ methods @vishalnayak?

alexanderbez: Going to resolve this. Spoke with @vishalnayak and it seems this isn't an issue.

vault/quotas/quotas.go
```diff
@@ -543,6 +569,20 @@ func (m *Manager) SetEnableRateLimitResponseHeaders(val bool) {
...
+	m.rateLimitPathManager.AddPaths(vals)
```
mgritter, line 583: I think that you should AddPaths to the new manager before moving it "into production". Even if the write to rateLimitPathManager itself is not a race condition (or is resolved by making it atomic), the following interleaved sequence is possible:

goroutine A: m.rateLimitPathManager = pathManager.New()
goroutine B: ApplyRateLimitQuota to an exempt path, have it rejected because pathManager is empty
goroutine A: AddPaths()

mgritter (changes requested):

vault/quotas/quotas.go
alexanderbez: I don't think this is necessary as I've added a write-lock to `SetRateLimitExemptPaths`.

vault/core.go
```diff
@@ -2522,14 +2522,26 @@ func (c *Core) setupQuotas(ctx context.Context, isPerfStandby bool) error {
...
 	req.Type = quotas.TypeRateLimit
+
+	resp := quotas.Response{
+		Allowed: true,
+		Headers: make(map[string]string),
+	}
+
 	if c.quotaManager != nil {
+		// skip rate limit checks for paths that are exempt from rate limiting
+		if c.quotaManager.RateLimitPathExempt(req.Path) {
```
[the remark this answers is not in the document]
alexanderbez: I've added read and write locks to the config-based getters and setters as well as `SetRateLimitExemptPaths`.

mgritter (approved):

vishalnayak (approved):

PR #10017, opened 2020-09-22
Add Documentation Specifying that Allowed, Denied, and Required Parameters only Work With Version 2 of The K/V Secrets Engine

HridoyRoy:
Due to Request.Data for requests to get/put secrets from version 2 of the K/V secrets engine being of the form map[data:[map[bar:[1]]]], in comparison to this struct being of the form [map[bar:[1]]] for requests to version 1 of the K/V secrets engine, acl.go does not properly compare secret request parameters and values with allowed, denied, and required parameters in the associated token policy in acl.go. 

Detailed testing of this can be found in the comments [here](https://hashicorp.atlassian.net/browse/VAULT-616). 

Additional / related links:

Code verifying the formatting difference for KV1 and KV2:
https://github.com/hashicorp/vault/blob/7aa1ffa92ee61b977efad1488b8f309b1e2136df/command/kv_test.go#L415 
https://github.com/hashicorp/vault/blob/7aa1ffa92ee61b977efad1488b8f309b1e2136df/command/kv_get.go#L128

Code demonstrating the assumptions that acl.go would need to make in order to function with KV2, along with the documentation in comments as to why we think this would be an anti pattern. 
https://github.com/hashicorp/vault/pull/10007

### DIFF
--- a/website/pages/docs/secrets/kv/kv-v2.mdx
+++ b/website/pages/docs/secrets/kv/kv-v2.mdx
@@ -151,6 +151,10 @@ path "secret/metadata/dev/team-1/*" {
 }
 ```
 
+The `allowed_parameters`, `denied_parameters`, and `required_parameters` fields are 
+not supported for policies used with the version 2 kv store. See the [Policies Concepts](/docs/concepts/policies)
+for a description of these parameters.
+
 See the [API Specification](/api/secret/kv/kv-v2) for more
 information.
 


ncabatoff:
Since this is a generic concepts page, maybe instead we should note these restrictions in https://www.vaultproject.io/docs/secrets/kv/kv-v2 ?

HridoyRoy:
> Since this is a generic concepts page, maybe instead we should note these restrictions in https://www.vaultproject.io/docs/secrets/kv/kv-v2 ?

Sounds good. I moved the documentation and added a link to the concepts page for more information on these policy parameters for further clarification!

vishalnayak (approved):

ncabatoff (approved):

PR #10019, opened 2020-09-22
Auto renew token leases

shishirkh:
- Tokens other than the root tokens have a maximum ttl of about 755 hours before which they must be renewed.
- This feature automatically renews the tokens through a cron job.
- Also gives option to specify duration for which lease has to be renewed.

### DIFF
--- a/auto-renewer/Dockerfile
+++ b/auto-renewer/Dockerfile
@@ -1,0 +1,8 @@
+FROM ubuntu:latest
+RUN apt-get update 
+RUN apt-get install -y curl
+COPY script.sh /script.sh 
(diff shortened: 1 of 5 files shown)


hsimon-hashicorp:
Hello! Welcome to the HashiCorp Vault community! Our contributing guidelines state that issues should be opened prior to all PRs, so the contributors, community, and our engineering team can discuss the issue and possible solutions and implementations. Vault does not rotate tokens automatically, and instead requires re-authentication. As a result, I'll be closing this PR now. We invite you to open other issues and PRs associated with them in the future. Thanks!

PR #10027, opened 2020-09-23
Transit: update documentation strings

brianshumate:
Update descriptions to match field content (the field is actually key name, not policy name).

As shown in the attached screen shot of API Explorer for Transit, the encrypt endpoint takes a key name, and the UI text appears to be pulling the **Description** field as  "Name of the policy" from the secrets engine instead of "Name of the key", which is what really goes in the field (i.e. "my-key").

Screen Shot 2020-09-23 at 12 18 40 PM


aphorise:
Request is valid and still relevant.

taoism4504 (approved):

PR #10028, opened 2020-09-23
Logging tweaks, mostly for tests

ncabatoff:
Fix some places where raft wasn't hooking into the core logger as it should.

Revisited the code that was setting the log level to Error during test cluster cleanup: it's normal for there to be a bunch of errors then, which makes it harder to see what went wrong up to the point where the test was deemed to have failed.  So now, instead of setting log level to Error, we actually stop logging altogether.  This only applies if the test didn't pass in its own logger during cluster creation, but we should be moving away from that anyway.

### DIFF
--- a/vault/testing.go
+++ b/vault/testing.go
@@ -896,26 +896,10 @@ func (c *TestClusterCore) stop() error {
 	return nil
 }
 
-func CleanupClusters(clusters []*TestCluster) {
-	wg := &sync.WaitGroup{}
-	for _, cluster := range clusters {
-		wg.Add(1)
-		lc := cluster
-		go func() {
-			defer wg.Done()
-			lc.Cleanup()
-		}()
-	}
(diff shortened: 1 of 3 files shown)


vault/testing.go
```diff
@@ -926,6 +910,7 @@ func (c *TestCluster) Cleanup() {
 		go func() {
 			defer wg.Done()
 			if err := lc.stop(); err != nil {
+				// TODO do we care about this?
```
vishalnayak, line 913: We might not care about it, but I'd lean towards leaving it there, coz we never know when we begin to care :-)

vishalnayak (approved):
Nice!

vault/testing.go
ncabatoff: What I meant was, do we care about not seeing this error any more (now that we're stopping all output)?

PR #10035, opened 2020-09-24
Ui/transform handle delete

chelshaw:
When a role is deleted, the transformations attached to it should be handled in a similar way as if the role had removed all of the transformations. This PR accomplishes that, removing the deleted role from each of the attached transformations' list of `allowed_roles`. 

There is an error in the console that is not breaking, but reads:
`Assertion Failed: Cannot create a new tag for `<vault@model:capabilities::ember1046:transform/role/role-name>` after it has been destroyed`
after a transition is attempted. Made a to-do in the code to address this at a future time.


ui/app/components/transform-edit-base.js
```diff
@@ -46,7 +46,7 @@ export default Component.extend(FocusOnInsertMixin, {
   },
 
   transitionToRoute() {
-    this.get('router').transitionTo(...arguments);
```
Monkeychip, line 49: 🙂

ui/app/components/transform-edit-base.js
```diff
@@ -78,11 +97,16 @@ export default Component.extend(FocusOnInsertMixin, {
...
 
     this.persist('save', () => {
+      this.hasDataChanges();
       callback();
       this.transitionToRoute(SHOW_ROUTE, `${modelPrefix}${modelId}`);
     });
   },
 
+  hasDataChanges() {
+    this.onDataChange(this.get('model.hasDirtyAttributes'));
```
Monkeychip, line 107: this will eventually change `this.model?.hasDirtyAttributes`

ui/app/components/transform-role-edit.js
```diff
@@ -103,5 +102,16 @@ export default TransformBase.extend({
         this.handleUpdateTransformations(updateTransformations, roleId);
       });
     },
+
+    delete() {
+      const roleId = this.get('model.id');
+      const roleTransformations = this.get('model.transformations') || [];
```
Monkeychip, line 108: same here this.model?.transformations

Monkeychip (approved):
A couple of comments re: the ember upgrade but nothing blocking

PR #10036, opened 2020-09-24
Improve errors for aws login with an unbound ARN

mikegrass:
During AWS login / renew, vault has logic to try to match the caller's identity against one of the `BoundIamPrincipalARNs` associated with the vault role.

The third step of this logic is to try to match any wildcard entries in the `BoundIamPrincipalARNs` for the vault role to the fully-resolved ARN of the caller's identity. 

In some environments, especially with cross-AWS account logins, resolving the full ARN is unlikely to succeed, which leads to rather confusing error messages:

```
2020-08-24T01:46:41.442Z [ERROR] auth.handler: error authenticating: error="Error making API request.
URL: PUT https://vault.company.com/v1/auth/aws/login
Code: 400. Errors:
* error looking up full ARN of entity &{aws 123456789012 assumed-role  my-role some-session}: error fetching role "my-role": NoSuchEntity: The role with name my-role cannot be found.
        status code: 404, request id: f0256e67-fd5f-4e45-b457-f456570bf6bb" backoff=2.852132462
```

However, resolving the full ARN only needs to be done if there are any wildcard entries in `BoundIamPrincipalARNs`, since the full ARN is only used in that path.

This PR:
1. Checks for wildcard entries in `BoundIamPrincipalARNs` before trying to resolve the full ARN. If there are none, it immediately returns with an error indicating the vault role is not bound to the caller's ARN, shortcutting the logic to resolve the full ARN (which wouldn't have been used anyway since there are no wildcard entries).
1. Augments error messages returned in this path to include the vault role, which can be helpful information for support teams when customers copy & paste the error message from the vault login call (currently a frequently-asked follow-up question is, "which vault role were you trying to log in with?").


builtin/credential/aws/path_login.go
```diff
@@ -1317,15 +1330,27 @@ func (b *backend) pathLoginUpdateIam(ctx context.Context, req *logical.Request,
...
+			hasWildcardBind := false
+			for _, principalARN := range roleEntry.BoundIamPrincipalARNs {
+				if strings.HasSuffix(principalARN, "*") {
+					hasWildcardBind = true
+					break
+				}
+			}
+
+			if !hasWildcardBind {
+				return logical.ErrorResponse(fmt.Sprintf("IAM Principal %q does not belong to the role %q", callerID.Arn, roleName)), nil
```
kalafut, line 1343:
```suggestion
				return logical.ErrorResponse("IAM Principal %q does not belong to the role %q", callerID.Arn, roleName), nil
```

builtin/credential/aws/path_login.go
```diff
@@ -1047,18 +1047,31 @@ func (b *backend) pathLoginRenewIam(ctx context.Context, req *logical.Request, d
 		case !roleEntry.ResolveAWSUniqueIDs && strutil.StrListContains(roleEntry.BoundIamPrincipalARNs, canonicalArn): // check 2 passed
 		default:
 			// check 3 is a bit more complex, so we do it last
+			// only try to look up full ARNs if there's a wildcard ARN in BoundIamPrincipalIDs.
+			hasWildcardBind := false
+			for _, principalARN := range roleEntry.BoundIamPrincipalARNs {
```
kalafut, line 1052: Since this loop is used in a couple of places and the function is already complex, it would be good to make it into a small helper function that does this check and returns a `bool`. Then the use here and below can reduce to:

```go
if !hasWildcardBind(roleEntry.BoundIamPrincipleARNs) {
   return nil, fmt.Errorf(...)
}
```

mikegrass: Good call, addressed in https://github.com/hashicorp/vault/pull/10036/commits/50fbee14ca914c23f22170b0b86b2430142655af

builtin/credential/aws/path_login.go
mikegrass: Thanks @kalafut! I didn't realize that `logical.ErrorResponse` accepted format strings. Addressed in https://github.com/hashicorp/vault/pull/10036/commits/50fbee14ca914c23f22170b0b86b2430142655af.

builtin/credential/aws/path_login.go
```diff
@@ -1047,18 +1047,23 @@ func (b *backend) pathLoginRenewIam(ctx context.Context, req *logical.Request, d
 		case !roleEntry.ResolveAWSUniqueIDs && strutil.StrListContains(roleEntry.BoundIamPrincipalARNs, canonicalArn): // check 2 passed
 		default:
 			// check 3 is a bit more complex, so we do it last
+			// only try to look up full ARNs if there's a wildcard ARN in BoundIamPrincipalIDs.
+			if !hasWildcardBind(roleEntry.BoundIamPrincipalARNs) {
+				return nil, fmt.Errorf("role %q no longer bound to ARN %q", roleName, canonicalArn)
```
calvn, line 1052: @kalafut Is it okay to return the role name as part of the error? This is based off the note above on information leakage, so wondering if where's trying to avoid this explicitly.

calvn: Discussed with this with a few parties internally and we're good with this change. Returning the role, though it's an additional bit of information, seems to be okay since a valid token with additional permissions is required to query information about the role.

calvn (approved):

PR #10037, opened 2020-09-24
Changed behavior of LDAP helper to have a 10s timeout per url

Valarissa:
This involved updating the Dial and DialTLS methods to use the `net.DialUrl` method. This also allowed us to move away from deprecated methods.

### DIFF
--- a/sdk/helper/ldaputil/ldap.go
+++ b/sdk/helper/ldaputil/ldap.go
@@ -15,18 +15,18 @@ func NewLDAP() LDAP {
 // LDAP provides ldap functionality, but through an interface
 // rather than statically. This allows faking it for tests.
 type LDAP interface {
-	Dial(addr string) (Connection, error)
-	DialTLS(addr string, config *tls.Config) (Connection, error)
+	Dial(network string, addr string) (Connection, error)
+	DialTLS(network string, addr string, config *tls.Config) (Connection, error)
 }
 
 type ldapIfc struct{}
 
-func (l *ldapIfc) Dial(addr string) (Connection, error) {
+func (l *ldapIfc) Dial(_network string, addr string) (Connection, error) {
 	dialer := ldap.DialWithDialer(&net.Dialer{Timeout: 10 * time.Second})
 	return ldap.DialURL(addr, dialer)
 }
 
-func (l *ldapIfc) DialTLS(addr string, config *tls.Config) (Connection, error) {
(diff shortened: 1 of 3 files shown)


sdk/helper/ldaputil/ldap.go
```diff
@@ -13,16 +15,19 @@ func NewLDAP() LDAP {
 // LDAP provides ldap functionality, but through an interface
 // rather than statically. This allows faking it for tests.
 type LDAP interface {
-	Dial(network, addr string) (Connection, error)
-	DialTLS(network, addr string, config *tls.Config) (Connection, error)
+	Dial(addr string) (Connection, error)
```
jasonodonnell, line 18: The changes to the interface are probably not needed and potentially could cause issues outside of the project (OpenLDAP and AD engines come to mind).  We can just ignore `network` and use `DialURL` instead of `Dial`.

sdk/helper/ldaputil/ldap.go
```diff
@@ -13,16 +15,19 @@ func NewLDAP() LDAP {
...
+	DialTLS(addr string, config *tls.Config) (Connection, error)
 }
 
 type ldapIfc struct{}
 
-func (l *ldapIfc) Dial(network, addr string) (Connection, error) {
-	return ldap.Dial(network, addr)
+func (l *ldapIfc) Dial(addr string) (Connection, error) {
+	dialer := ldap.DialWithDialer(&net.Dialer{Timeout: 10 * time.Second})
```
jasonodonnell, line 25: nit: Can we make this timeout a constant? It might also make a good configurable.

sdk/helper/ldaputil/ldap.go
```diff
@@ -13,16 +15,19 @@ func NewLDAP() LDAP {
...
+	return ldap.DialURL(addr, dialer)
 }
 
-func (l *ldapIfc) DialTLS(network, addr string, config *tls.Config) (Connection, error) {
-	return ldap.DialTLS(network, addr, config)
+func (l *ldapIfc) DialTLS(addr string, config *tls.Config) (Connection, error) {
+	tlsOpts := ldap.DialWithTLSConfig(config)
+	dialer := ldap.DialWithDialer(&net.Dialer{Timeout: 10 * time.Second})
```
jasonodonnell, line 31: Same as above.

Valarissa:
Other approaches were recommended to remediate the customer's issue before we attempted a code-based solution.

PR #10039, opened 2020-09-24
Retry Logic to Mssql Tests [VAULT-637]

HridoyRoy:
Associated Jira: https://hashicorp.atlassian.net/browse/VAULT-637

Bug Description: Mssql is unreachable in docker containers intermittently when running in parallel. Containers seem otherwise healthy, and mssql is not in the process of starting during these test errors. Instead, mssql has already started, and the processes exist, but mssql commands do not run in the containers due to timeouts. 

Resolution: Wrapping the mssql docker container creation within simple retry logic. Observed behavior from local testing is that the number of retries corresponds roughly to the number of containers being spun up in parallel (up to a point, before the docker daemon crashes), but this is not representative of CI. 

Cost: At most each mssql test will take 5 minutes to fail as opposed to 1 minute to fail.


helper/testhelpers/mssql/mssqlhelper.go
```diff
@@ -13,28 +13,37 @@ import (
...
+// This constant is used in retrying the mssql container restart, since
+// intermittently the container starts but mssql within the container
+// is unreachable.
+const numRetries = 3
+
 func PrepareMSSQLTestContainer(t *testing.T) (cleanup func(), retURL string) {
 	if os.Getenv("MSSQL_URL") != "" {
 		return func() {}, os.Getenv("MSSQL_URL")
 	}
+	var svc *docker.Service
```
ncabatoff, line 25: May as well keep these loop-scoped since they're not used outside of it.

helper/testhelpers/mssql/mssqlhelper.go
```diff
@@ -13,28 +13,37 @@ import (
...
+	var err error
+	for i := 0; i < numRetries; i++ {
+		runner, err := docker.NewServiceRunner(docker.RunOptions{
+			ContainerName: "sqlserver",
+			ImageRepo:     "mcr.microsoft.com/mssql/server",
+			ImageTag:      "2017-latest-ubuntu",
+			Env:           []string{"ACCEPT_EULA=Y", "SA_PASSWORD=" + mssqlPassword}, // "MSSQL_PID=Express" does nothing
```
ncabatoff, line 32: I don't think this comment adds anything.

ncabatoff (approved):
Minor suggestions.

helper/testhelpers/mssql/mssqlhelper.go
HridoyRoy: Oops, 💯

helper/testhelpers/mssql/mssqlhelper.go
HridoyRoy: Makes sense, will move those inside the loop!

PR #10045, opened 2020-09-25
Make raft clusters in tests come up quicker

ncabatoff:
…  via a smaller performance multiplier and thus quicker elections.

### DIFF
--- a/helper/testhelpers/teststorage/teststorage.go
+++ b/helper/testhelpers/teststorage/teststorage.go
@@ -95,7 +95,7 @@ func MakeRaftBackend(t testing.T, coreIdx int, logger hclog.Logger) *vault.Physi
 	conf := map[string]string{
 		"path":                   raftDir,
 		"node_id":                nodeID,
-		"performance_multiplier": "8",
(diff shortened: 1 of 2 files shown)


helper/testhelpers/teststorage/teststorage.go
```diff
@@ -95,7 +95,7 @@ func MakeRaftBackend(t testing.T, coreIdx int, logger hclog.Logger) *vault.Physi
 	conf := map[string]string{
 		"path":                   raftDir,
 		"node_id":                nodeID,
-		"performance_multiplier": "8",
```
vishalnayak, line 98: I didn't know that it could affect bootup times.

calvn:
Are there any repercussions by lowering this value?

ncabatoff:
> Are there any repercussions by lowering this value?

Brian points out that it may increase test flakiness, and indeed I see some raft race test failures that I don't recall seeing much before.  I may make this a CI-dependent (or CI-race-dependent) setting.

ncabatoff:
Closing, see #10945 for a more complete & current version.

PR #10047, opened 2020-09-26
Minor fix to audit documentation

torinthiel:
Since the sentence references more than one audit device,
it makes sense to continue in this style, instead of using 'both.


raskchanky (approved):
Looks great, thank you!

kqtqk88:
@torinthiel

PR #10049, opened 2020-09-28
Ui/pricing metrics page setup

chelshaw:
This PR sets up the Pricing Metrics UI page (currently based on mock data from the endpoint). This is the bare-bones version of this page, without date filtering or namespaces. Metrics are seen at the route `/ui/vault/metrics` while the config is at `/ui/vault/metrics-config`.

metrics-with-result-dates

When disabled but there is data:
metrics-enabled-off

Config: 
metrics-page

TODOS: 

- [ ] Remove http-entities and other old models for metrics


ui/app/controllers/vault/cluster/metrics-config.js
```diff
@@ -0,0 +1,24 @@
...
+  infoRows: computed(function() {
+    return [
+      {
+        label: 'Usage data collection',
+        helperText: 'Enable or disable collecting data to track clients.',
+        valueKey: 'enabled',
+      },
+      {
+        label: 'Retention period',
+        helperText: 'The number of months of activity logs to maintain for  client tracking.',
```
Monkeychip, line 14: looks like maybe an extra space between for and client

ui/tests/unit/models/metrics/activity-test.js
```diff
@@ -0,0 +1,13 @@
+import { module, test } from 'qunit';
```
Monkeychip, line 1: I'm guessing there's some test that can be removed as well when this goes live.

Monkeychip (approved):
Looks like a test failure.  But this looks good.  Same with the naming.  Are we going to go live with this before the API endpoint is ready?

chelshaw:
> Looks like a test failure. But this looks good. Same with the naming. Are we going to go live with this before the API endpoint is ready?

Good question, we can either merge this (and all forthcoming work) into a sidebranch until we know the API will be ready for release, or merge now and be prepared to revert later. My concern with a sidebranch is that when the ember upgrade is finished we'll have to do that process again for this new work, but hopefully since it's a small section it won't be too painful. Thoughts @Monkeychip ?

PR #10050, opened 2020-09-28
Make username generation in SQLCredentialsProducer available without an instance

pcman312:
# Overview
Moves the logic of username generation in the `SQLCredentialsProducer` into its own standalone function. Also gives it the ability to uppercase usernames.

**Note**: I did test this against the existing `SQLCredentialsProducer` to ensure the usernames that are being generated match the same patterns as they have been. Since the producer has been updated to use the new function, doing any tests to ensure they match seemed pointless to include, but I did have them prior to moving the producer over to the new function.

### DIFF
--- a/sdk/database/helper/credsutil/usernames.go
+++ b/sdk/database/helper/credsutil/usernames.go
@@ -1,0 +1,143 @@
+package credsutil
+
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+type CaseOp int
+
+const (
+	KeepCase CaseOp = iota
+	Uppercase
+	Lowercase
+)
+
+type usernameBuilder struct {
+	displayName string
+	roleName    string
+	separator   string
+
+	maxLen        int
+	caseOperation CaseOp
+}
+
+func (ub usernameBuilder) makeUsername() (string, error) {
+	userUUID, err := RandomAlphaNumeric(20, false)
+	if err != nil {
+		return "", err
+	}
+
+	now := fmt.Sprint(time.Now().Unix())
+
+	parts := []string{
+		"v",
+		ub.displayName,
+		ub.roleName,
+		userUUID,
+		now,
+	}
+	username := joinNonEmpty(ub.separator, parts...)
+	if ub.maxLen > 0 {
+		username = trunc(username, ub.maxLen)
+	}
+	switch ub.caseOperation {
+	case Lowercase:
+		username = strings.ToLower(username)
+	case Uppercase:
+		username = strings.ToUpper(username)
+	}
+
+	return username, nil
+}
+
+type UsernameOpt func(*usernameBuilder)
+
+func DisplayName(dispName string, maxLength int) UsernameOpt {
+	return func(b *usernameBuilder) {
+		if maxLength > 0 {
+			dispName = trunc(dispName, maxLength)
+		}
+		b.displayName = dispName
+	}
+}
+
+func RoleName(roleName string, maxLength int) UsernameOpt {
+	return func(b *usernameBuilder) {
+		if maxLength > 0 {
+			roleName = trunc(roleName, maxLength)
+		}
+		b.roleName = roleName
+	}
+}
+
+func Separator(sep string) UsernameOpt {
+	return func(b *usernameBuilder) {
+		b.separator = sep
+	}
+}
+
+func MaxLength(maxLen int) UsernameOpt {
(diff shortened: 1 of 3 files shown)


tvoran (approved):

sdk/database/helper/credsutil/usernames.go
```diff
@@ -0,0 +1,143 @@
...
+	now := fmt.Sprint(time.Now().Unix())
+
+	parts := []string{
+		"v",
+		ub.displayName,
+		ub.roleName,
+		userUUID,
+		now,
+	}
+	username := joinNonEmpty(ub.separator, parts...)
```
calvn, line 41: Any reason not to have the signature be the same as `strings.Join`, `joinNonEmpty(parts, ub.separator)`?

sdk/database/helper/credsutil/usernames.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return username, nil
+}
+
+type UsernameOpt func(*usernameBuilder)
+
+func DisplayName(dispName string, maxLength int) UsernameOpt {
+	return func(b *usernameBuilder) {
+		if maxLength > 0 {
```
calvn, line 59: I think it's missing the same where the length is `NoneLength` (-1), which would entirely skip including the display name regardless of what display name was provided. Same for RoleName.

sdk/database/helper/credsutil/usernames.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return b.makeUsername()
+}
+
+func trunc(str string, l int) string {
+	if len(str) < l {
+		return str
+	}
+	return str[:l]
```
calvn, line 119: This could be out of bound if `l` is negative.

calvn (commented):
Left a few comment, but looks good otherwise!

sdk/database/helper/credsutil/usernames.go
pcman312: Because you can't put a vararg at the beginning of the signature.

sdk/database/helper/credsutil/usernames.go
pcman312: Ah good catch. I missed that. Fixed.

sdk/database/helper/credsutil/usernames.go
pcman312: True, but the checks for this are happening above in `makeUsername`, `DisplayName` and `RoleName`. I wanted to keep this function cleaner by not returning an error.

sdk/database/helper/credsutil/usernames.go
calvn: You can change the signature to accept a slice

pcman312: Accepting a slice is a lot less flexible than accepting a vararg.

calvn: Since the separator is a string value, having the input values be passed in as a slice is a bit easier to read IMO

pcman312: Admittedly it's a small issue, but this approach makes it so you don't have to allocate a slice if all of the values are already in variables (or in-line strings). I've put up a commit that shows this (since I neglected to remove the slice from a previous iteration of this function): https://github.com/hashicorp/vault/pull/10050/commits/b6a431d265448c262bef011f4abeb7f5d6bcf28c

calvn (approved):

sdk/database/helper/credsutil/usernames.go
calvn: Yeah, this is purely internal so I'll leave it up to you.

sdk/database/helper/credsutil/usernames.go
calvn: Can we add a comment on trunc to explicitly state this? I'm more concerned about future usage of this func without proper bound check leading to panics. Alternatively, we could handle the negative value by returning an empty string if that's provided (effectively doing a `max(l, 0)` on the length).

pcman312: Looking at this again I think we can do:
If `l == 0` - return the unmodified string
if `l == NoneLength` - return `""`
if `l < 0` - return `""` (in principle we could error in this case, but I think just returning an empty string is fine too)

PR #10059, opened 2020-09-30
Port OSS changes from #1497

alexanderbez:


vault/logical_system_helpers.go
```diff
@@ -139,3 +139,7 @@ func (b *SystemBackend) tuneMountTTLs(ctx context.Context, path string, me *Moun
 
 	return nil
 }
+
+func (b *SystemBackend) wrapHandleRaftRemovePeer() framework.OperationFunc {
+	return b.handleRaftRemovePeerUpdate()
+}
```
briankassouf, line 145: Checkout the var definitions at the top of this file, i think you'll want to do something similar so you can override the function on ent.

briankassouf (approved):

vault/logical_system_helpers.go
```diff
@@ -92,6 +92,10 @@ var (
 	}
 
 	checkRaw = func(b *SystemBackend, path string) error { return nil }
+
+	wrapHandleRaftRemovePeer = func(b *SystemBackend) framework.OperationFunc {
```
vishalnayak, line 96: @alexanderbez I think this is causing oss to ent merge conflict. You may want to fix the signature here.

PR #10064, opened 2020-09-30
Reworded disable_mlock docs

TroyFluegge:
Reworded disable_mlock to remove confusion regarding what is acceptable for production deployments.  Disabling mlock is alright for production given the additional security recommendations are implemented.  Disabling mlock is also recommended for integrated storage


briankassouf (approved):
Thanks!

PR #10070, opened 2020-10-01
Test Fix: test for default TTL instead of date timestamp

Monkeychip:
Originally we were testing for the date timestamp on the replication modal.  The problem is there is an off chance that the timestamp could be off by 1 second.  To fix this we are instead checking for the default TTL of 1800.  We no longer check date formatting but are still checking for the correct default TTL, which was the core of this test in the first place.

image


ui/tests/acceptance/enterprise-replication-test.js
```diff
@@ -235,16 +234,8 @@ module('Acceptance | Enterprise | replication', function(hooks) {
...
-    today.setMinutes(today.getMinutes() + 30); // add default 30 min TTL to current date to return expires
-    const dateFormatted = dateFormat([today, 'MMM DD, YYYY hh:mm:ss A']);
-    const modalDate = document.querySelector('[data-test-row-value="Expires"]').innerText;
-    // because timestamp might be off by a 1 sec due to different in exp from token vs. adding 30 min, checking the string before the seconds.
-    assert.equal(
-      dateFormatted.slice(0, 18),
-      modalDate.slice(0, 18),
-      'shows the correct expiration date and in the correct format'
-    );
+    const modalDefaultTtl = document.querySelector('[data-test-row-value="TTL"]').innerText;
```
chelshaw, line 237: Should we do one more test scenario where we don't use the default (set it to something specific) and make sure the TTL reflects that?

chelshaw (approved):
Other than another test possibly, this change looks great!

ui/tests/acceptance/enterprise-replication-test.js
Monkeychip: Good call.  Just pushed a commit that checks for when it's set at 3 minutes as well.

Issue #9936, opened 2020-09-12
Incorrect documentation on disabling max_request_size


somethingnew2-0:
**Describe the bug**
The current [documentation for `max_request_size`](https://www.vaultproject.io/docs/configuration/listener/tcp#max_request_size) states:
> `max_request_size` (int: 33554432) – Specifies a hard maximum allowed request size, in bytes. Defaults to 32 MB. Specifying a number less than or equal to `0` turns off limiting altogether.

However when setting `max_request_size` to 0, according to the code Vault currently sets it to the default of 32 MB instead of disabling the request size limiting feature.
https://github.com/hashicorp/vault/blob/v1.5.3/http/handler.go#L266

Also when setting `max_request_size` to -1 or any negative number, according to the code an error is thrown "max_request_size cannot be negative" which halts Vault from starting up
https://github.com/hashicorp/vault/blob/v1.5.3/internalshared/configutil/listener.go#L151

Please correct the documentation, correct the code, or both as currently this `max_request_size` limiting feature cannot be disabled at all.

**To Reproduce**
Set a vault server setup with Vault config such as demonstrated below 
- if max_request_size is set to 0, then it is not disabled, but set to the default of 32 MB.
- if max_request_size is set to less than 0, then it is not disabled, but Vault does not correctly startup and instead throws an error of an invalid config as the value is incorrectly not allowed to be negative.

**Expected behavior**
Setting `max_request_size` to 0 or a negative number should disable limiting request sizes.

**Environment:**
* Vault Server Version: 1.5.3
* Vault CLI Version: N/A
* Server Operating System/Architecture: N/A

Vault server configuration file(s):

Setting max_request_size to 0 should disable max request size limiting, but it is set to the default of 32MB
```hcl
...
listener "tcp" {
  ...
  max_request_size = 0
}
```

Setting max_request_size to a negative should disable max request size limiting, but Vault fails to boot and throws this error

> error loading configuration from /secret/config.hcl: error parsing 'listener': listeners.0 max_request_size cannot be negative

```hcl
...
listener "tcp" {
  ...
  max_request_size = -1
}
```

jay-dee7:
@guacamole want to take this one up?

guacamole:
Hey @raskchanky , I would like to give it a shot.

guacamole:
@raskchanky I went through the code and i think the conditions where-all  max_request_size <= 0 occurs can be removed without breaking anything. Let me know what you think!

[...]

PR #10072, opened 2020-10-01
Fix: handle max_request_size<=0

guacamole:
Signed-off-by: guacamole <gunjanwalecha@gmail.com>

This PR tries to fix the issue #9936 
Setting `max_request_size` to 0 or a negative number should disable request size limiting 

### Motivation

As per the documentation: `max_request_size` is the maximum allowed request size, in bytes. Defaults to 32 MB. Specifying a number less than or equal to 0 should turn off limiting altogether.

#### Implementation details

Old implementation:
* when `max_request_size` is 0, it gets set to a default size of 32 MB
* when  `max_request_size` is negative, it throws an error saying: " max_request_size cannot be negative"

New implementation:
* when `max_request_size`  set to less than or equal to 0, the limiting is disabled


guacamole:
@raskchanky Would you please have a look at this ?

raskchanky (changes requested):
One issue with the current approach is that 0 is Go's default value for integers, so if no max request size is specified, it will be 0, in which case the default _should_ be used. Otherwise, it's kinda pointless to have a default value 😁.

In my opinion, the right fix here is to:
* remove the error saying max request size can't be negative
* if max request size is negative, then disable it
* update the docs to reflect this change

Also, in addition to fixing the existing CI failure, it'd be great to have some additional tests showing that:
* the default max request size is still in place if it's not overridden
* it's properly disabled when set to a negative value
* the config is parsed correctly with negative values in it

Thanks so much for working on this!

guacamole:
@raskchanky Thanks for the input, I'm working on it!

HridoyRoy:
Hi @guacamole ,
It's been a while since we've heard from you. Wondering if you were still working on this PR? 

Thanks so much!

guacamole:
Hello @HridoyRoy  
 sorry for the long time, I'll get to it today and have an update for you soon:)

guacamole:
Hi @raskchanky , I just rebased my PR from master and `test-go` is failing on circle-ci . Would you please check the failing test cases as the http module did not fail on my local.

HridoyRoy (approved):
@guacamole lgtm!

raskchanky (approved):

HridoyRoy:
I've created a docs PR to go with this change[ here](https://github.com/hashicorp/vault/pull/10723/files)

PR #10073, opened 2020-10-01
Backport 1.5.x - Conditionally overwrite TLS parameters for MySQL secrets engine (#9729)

Valarissa:
* Conditionally overwrite TLS parameters in MySQL DSN

Overwrite MySQL TLS configuration in MySQL DSN only if have `tls_ca` or `tls_certificate_key` set
Current logic always overwrites it

* Add test for MySQL DSN with a valid TLS parameter in query string


jasonodonnell (approved):
LGTM

PR #10076, opened 2020-10-01
Expose generic versions of KDF and symmetric crypto

sgmiller:
This allows other systems to invoke the vetted versions of these in a
generic manner that doesn't involve the details of Transit.


sdk/helper/keysutil/policy.go
```diff
@@ -692,16 +692,21 @@ func (p *Policy) Upgrade(ctx context.Context, storage logical.Storage, randReade
...
+func (p *Policy) GetKey(context []byte, ver, numBytes int) ([]byte, error) {
 	// Fast-path non-derived keys
 	if !p.Derived {
 		return p.Keys[strconv.Itoa(ver)].Key, nil
 	}
 
+	return p.DeriveKey(context, nil, ver, numBytes)
+}
+
+// DeriveKey is used to derive a symmetric key given a context and salt.
```
briankassouf, line 708: Can you add a blurb to the comment about how this does not check the p.Derived flag?

sdk/helper/keysutil/policy.go
```diff
@@ -843,65 +847,10 @@ func (p *Policy) Encrypt(ver int, context, nonce []byte, value string) (string,
...
-			}
-		}
-
-		// Encrypt and tag with AEAD
-		ciphertext = aead.Seal(nil, nonce, plaintext, nil)
-
-		// Place the encrypted data after the nonce
-		if !p.ConvergentEncryption || p.convergentVersion(ver) > 1 {
-			ciphertext = append(nonce, ciphertext...)
+		ciphertext,err = p.SymmetricEncryptRaw(ver, p.ConvergentEncryption, encKey, nil, hmacKey, plaintext, nil)
```
briankassouf, line 850:
```suggestion
		ciphertext, err = p.SymmetricEncryptRaw(ver, p.ConvergentEncryption, encKey, nil, hmacKey, plaintext, nil)
```

sdk/helper/keysutil/policy.go
```diff
@@ -992,49 +939,7 @@ func (p *Policy) Decrypt(context, nonce []byte, value string) (string, error) {
...
-			nonce = decoded[:aead.NonceSize()]
-			ciphertext = decoded[aead.NonceSize():]
-		}
-
-		// Verify and Decrypt
-		plain, err = aead.Open(nil, nonce, ciphertext, nil)
-		if err != nil {
-			return "", errutil.UserError{Err: "invalid ciphertext: unable to decrypt"}
-		}
+		plain, err = p.SymmetricDecryptRaw(p.ConvergentEncryption, convergentVersion, encKey, decoded, nil)
```
briankassouf, line 942: Looks like you're not checking the error here

briankassouf (commented):
Also, consider if a struct argument would make more sense `SymmetricEncryptRaw`/`SymmetricDecryptRaw`

sdk/helper/keysutil/policy.go
```diff
@@ -1502,10 +1511,19 @@ func (p *Policy) getVersionPrefix(ver int) string {
 	return prefix
 }
 
+type SymmetricOpts struct {
+	Convergent bool
+	ConvergentVersion int
+	Nonce []byte
+	AdditionalData []byte
+	HMACKey []byte
+
```
briankassouf, line 1520:
```suggestion
```

sdk/helper/keysutil/policy.go
```diff
@@ -1502,10 +1511,19 @@ func (p *Policy) getVersionPrefix(ver int) string {
 	return prefix
 }
 
+type SymmetricOpts struct {
```
briankassouf, line 1514: Can you comment the struct and each member?

sdk/helper/keysutil/policy.go
```diff
@@ -1596,3 +1510,129 @@ func (p *Policy) getVersionPrefix(ver int) string {
...
+	Convergent bool
+	ConvergentVersion int
+	Nonce []byte
+	AdditionalData []byte
+	HMACKey []byte
+
+}
+// Symmetrically encrypt a plaintext given the convergence configuration and appropriate keys
+func (p *Policy) SymmetricEncryptRaw(ver int,  encKey, plaintext []byte, opts SymmetricOpts) ([]byte, error) {
```
briankassouf, line 1523: Here and decrypt- since we are being handed an encryption key, should we validate its length?

sgmiller: aes.NewCipher and chacha's New both do this already.

sdk/helper/keysutil/policy.go
```diff
@@ -1596,3 +1514,136 @@ func (p *Policy) getVersionPrefix(ver int) string {
...
+				return nil, errutil.InternalError{Err: fmt.Sprintf("invalid hmac key length of zero")}
+			}
+			nonceHmac := hmac.New(sha256.New, opts.HMACKey)
+			nonceHmac.Write(plaintext)
+			nonceSum := nonceHmac.Sum(nil)
+			nonce = nonceSum[:aead.NonceSize()]
+		default:
+			return nil, errutil.InternalError{Err: fmt.Sprintf("unhandled convergent version %d", convergentVersion)}
+		}
+	} else if nonce == nil {
```
briankassouf, line 1582: Can we do a length check here instead? Just to be sure we aren't being tricked by typed nils?

briankassouf (approved):
One more thing, otherwise looks good!

sdk/helper/keysutil/policy.go
sgmiller: Yep.

PR #10080, opened 2020-10-02
Add a helper function for safely grabbing a keyEntry by version

sgmiller:


sdk/helper/keysutil/policy.go
```diff
@@ -770,6 +780,15 @@ func (p *Policy) DeriveKey(context, salt []byte, ver int, numBytes int) ([]byte,
 	}
 }
 
+func (p *Policy) safeGetKeyEntry(ver int) (*KeyEntry, error) {
+	keyVerStr := strconv.Itoa(ver)
+	keyEntry, ok := p.Keys[keyVerStr]
+	if !ok {
+		return nil, errutil.UserError{Err: "no such key version"}
+	}
+	return &keyEntry, nil
```
briankassouf, line 789: Can we go back to returning by value here?

sgmiller: There was some reason it didn't work, but I'll take another look.

sgmiller: Nm, works just fine.  Changed.

briankassouf (approved):

PR #10083, opened 2020-10-02
Ui/pricing metrics params

chelshaw:
This PR adds the date inputs for pricing metrics, with date form validation. The query button links to the metrics page with start and end parameters, which get consumed and displayed by the inputs.

When the date inputs match the result dates, this is how the page should look. 
metrics-no-warning

If the result dates are off of the query dates by more than 1 day (due to timezones) a warning will show. The "from" input has an implicit "first of month" date and the "through" input has an implicit "end of month" date.
metrics-date-warning
 
### Form Validation
Form errors show when the date is invalid (eg. month value of > 12), from is after through, or input is incomplete/formatted incorrectly. 
metrics-date-invalid
metrics-start-date-after-end
metrics-bad-month

### Namespace Navbar link
There is no "status" nav dropdown when you are within a namespace, so you instead get to a namespace pricing metrics page through this `Metrics` link, which will automatically query only for that namespace's activity
namespace-metrics


ui/app/templates/vault/cluster/metrics.hbs
```diff
@@ -37,34 +37,53 @@
...
     />
+
+    {{#unless model.activity.total}}
+      <EmptyState @title="No data found" @message="No data exists for that query period. Try searching again." />
+    {{else}}
+      <div class="selectable-card-container">
+        <SelectableCard
+          @cardTitle="Active clients"
+          @total={{model.activity.total.clients}}
+          @subText="Current namespace"
```
Monkeychip, line 74: Should this pull in the actual namespace?  And if no namespace, should it be root (or something similar)?

ui/lib/core/addon/components/form-error.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @module FormError
+ * FormError components are used to...
+ *
+ * @example
+ * ```js
+ * <FormError @requiredParam={requiredParam} @optionalParam={optionalParam} @param1={{param1}}/>
+ * ```
+ * @param {object} requiredParam - requiredParam is...
```
Monkeychip, line 9: should these sentences be finished?

ui/app/components/pricing-metrics-dates.js
```diff
@@ -0,0 +1,131 @@
+/**
+ * @module PricingMetricsDates
+ * PricingMetricsDates components are used on the Pricing Metrics page to handle queries related to pricing metrics.
+ * This component assumes that query parameters (as in, from route params) are being passed in with the format MM-YYYY,
+ * while the inputs expect a format of MM/YYYY.
+ *
+ * @example
+ * ```js
+ * <PricingMetricsDates @requiredParam={requiredParam} @optionalParam={optionalParam} @param1={{param1}}/>
```
Monkeychip, line 9: should this be filled in with example params from below?

ui/app/components/pricing-metrics-dates.js
```diff
@@ -0,0 +1,131 @@
...
+  resultEnd: null,
+
+  start: null,
+  end: null,
+
+  defaultSpan: 12,
+  retentionMonths: 24,
+
+  startDate: computed('start', function() {
+    if (!this.start) return null;
```
Monkeychip, line 46: nice - I didn't know you didn't have to use function curly braces in this situation?

ui/app/helpers/parse-date-string.js
```diff
@@ -0,0 +1,20 @@
+import { helper } from '@ember/component/helper';
```
Monkeychip, line 1: really nice job on this helper.  I tried to break it and I couldn't 🙂

ui/app/routes/vault/cluster/metrics.js
```diff
@@ -1,14 +1,52 @@
...
+    let startDate = parseDateString(start);
+    if (startDate) {
+      // TODO: Replace with formatRFC3339 when date-fns is updated
+      params.start_time = Math.round(startDate.getTime() / 1000);
+    }
+  }
+  if (end) {
+    let endDate = parseDateString(end);
+    if (endDate) {
+      // TODO: Replace with formatRFC3339 when date-fns is updated
```
Monkeychip, line 21: if you don't already have a JIRA ticket for this might be a good idea to create one so we don't forget.  And have it dependent on any relevant API ticket from the backend.

ui/app/templates/components/pricing-metrics-dates.hbs
```diff
@@ -0,0 +1,52 @@
...
+    {{date-format resultStart "MMM DD, YYYY"}} through {{date-format resultEnd "MMM DD, YYYY"}}
+  </h2>
+  {{#if showResultsWarning}}
+  <div class="access-information" data-test-results-date-warning>
+    <Icon @glyph="info-circle-fill" class="has-text-info"/>
+    {{!-- TODO: Add "Learn more here." link --}}
+    <p>This data may not reflect your search exactly. This is because Vault will only show data for contiguous blocks of time during which tracking was on. </p>
+  </div>
+  {{/if}}
+</div>
```
Monkeychip, line 52: extra line at the end.

ui/lib/core/addon/templates/components/form-error.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="is-flex is-flex-center">
```
Monkeychip, line 1: maybe check with Ivana if there should be any padding/margin on this component.  Seems a little close to the text. 
image

ui/tests/integration/components/pricing-metrics-dates-test.js
```diff
@@ -0,0 +1,85 @@
...
+      <PricingMetricsDates
+        @queryStart="2-2020"
+        @queryEnd="10-2020"
+        @resultStart={{resultStart}}
+        @resultEnd={{resultEnd}}
+      />
+    `);
+    assert.dom('[data-test-results-date-warning]').exists('shows states warning');
+  });
+});
```
Monkeychip, line 85: nice coverage on this test!

Monkeychip (changes requested):
This looks great.  A couple of questions/comments:
1. Right now we have fake data.  If we push this to master there's no way folks would know that the data is fake.  Unsure how to handle that. 
2. The old code for metrics (ex: models/metrics/tokens) still exists.  Maybe once this is released we can remove all the old code.  I checked and didn't see a ticket for it, but might be something we create one for.
3. There's a test failure, and I'm not entirely sure what it's all about.  Happy to help problem-solve if it doesn't resolve itself with a master merge.

Monkeychip:
Also, I checked with Ivana on this --this can be a separate ticket--, but the card components should not have a hover state (especially because they are not clickable). 
hover

ui/app/templates/vault/cluster/metrics.hbs
chelshaw: The design says "current namespace" and the total clients on the model will reflect whatever namespace they are currently in, so this works as currently designed

ui/lib/core/addon/components/form-error.js
chelshaw: yeah... 😂

ui/app/routes/vault/cluster/metrics.js
chelshaw: Good call, done!

ui/tests/integration/components/pricing-metrics-dates-test.js
chelshaw: Still not a bad idea, since a unit test is easier to really see what's going wrong than a component test that consumes it. I will add one before I merge!

chelshaw:
@Monkeychip I've made updates that address most of your comments, including removing the old metrics files. I also realized I didn't add a screenshot of the namespace navbar link, so I added that to the description.

Monkeychip (approved):
Looks great!

PR #10087, opened 2020-10-05
Mask LDAP bindpass while typing

Monkeychip:
This PR addresses a request to mask the LDAP bindpass field while typing.  This field is generated by the `masked-input` component, which uses a `textArea` html element.  However, to effectively mask the field while typing we needed to change the element to `input` and add `type=password` (see screenshot and MDN docs [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input)).

image

Before -with the `textArea` element- you could enter a multi-line password for bindpass.  However, given that bindpass is defined as a password field in the LDAP [API docs ](https://www.vaultproject.io/docs/auth/ldap) we felt it safe to go ahead and change this to a single-line input field. 

There are similar requests for the value field in the KV secret engine.  Right now, this component will only show the input field when the name is `bindpass` but if we later determine to add other fields we can add a list of them in the `masked-input` javascript file and pass a computed property to the new `maskWhileTyping` property.

See gif
ldap

*You'll notice the browser warning in the gif.  I tried a variety of suggestions for preventing this, but ultimately this is a browser setting and not something we can safely prevent when we use `type=password`.  I tried as[ suggested here](https://stackoverflow.com/questions/41217019/how-to-prevent-a-browser-from-storing-passwords) turning the input field to readonly and readonly=false onfocus, but that did not work.

### DIFF
--- a/ui/lib/core/addon/templates/components/form-field.hbs
+++ b/ui/lib/core/addon/templates/components/form-field.hbs
@@ -129,8 +129,7 @@
   }}
 {{else if (eq attr.options.sensitive true)}}
   <MaskedInput @value={{or (get model valuePath) attr.options.defaultValue}} @placeholder="" @allowCopy="true"
-    @onChange={{action (action "setAndBroadcast" valuePath)}} />
-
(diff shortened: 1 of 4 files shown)


chelshaw (approved):
:chef-kissing-fingers: looks great!

PR #10095, opened 2020-10-06
Integrated Storage Cloud Auto-Join

alexanderbez:


vault/raft.go
```diff
@@ -739,14 +740,14 @@ func (c *Core) JoinRaftCluster(ctx context.Context, leaderInfos []*raft.LeaderJo
...
+	// Disallow leader API address to be provided if we're using raft for HA-only.
 	// The leader API address is obtained directly through storage. This serves
 	// as a form of verification that this node is sharing the same physical
 	// storage as the leader node.
 	if isRaftHAOnly {
 		for _, info := range leaderInfos {
-			if info.LeaderAPIAddr != "" {
-				return false, errors.New("leader API address must be unset when raft is used exclusively for HA")
+			if info.LeaderAPIAddr != "" || info.AutoJoin != "" {
+				return false, errors.New("leader API address or auto-join metadata must be unset when raft is used exclusively for HA")
```
ncabatoff, line 750: Should be 'and' rather than 'or' I think.

vault/raft.go
```diff
@@ -1116,3 +1151,14 @@ type answerResp struct {
 	Peers      []raft.Peer      `json:"peers"`
 	TLSKeyring *raft.TLSKeyring `json:"tls_keyring"`
 }
+
+func newDiscover() (*discover.Discover, error) {
```
ncabatoff, line 1155: Is this helper still needed now that we're not adding k8s?

ncabatoff (approved):
Minor comments, overall looks good!

ncabatoff:
Oh, except test-go is failing for some reason.  Better fix that before merging.

swayne275:
1025 files changed haha, which ones should I care about?

alexanderbez:
> 1025 files changed haha, which ones should I care about?

All the non-vendor ones ;-p

vault/raft.go
alexanderbez: Yeah we can safely remove this 👍

alexanderbez: But on second thought, just in case the go-discover API changes, I figure it's safe to be explicit. So I'm leaning towards keeping this.

command/operator_raft_join.go
```diff
@@ -143,14 +148,21 @@ func (c *OperatorRaftJoinCommand) Run(args []string) int {
...
+	joinReq := &api.RaftJoinRequest{
 		LeaderCACert:     leaderCACert,
 		LeaderClientCert: leaderClientCert,
 		LeaderClientKey:  leaderClientKey,
 		Retry:            c.flagRetry,
 		NonVoter:         c.flagNonVoter,
-	})
+	}
+
+	if strings.HasPrefix(leaderInfo, "provider=") {
```
HridoyRoy, line 159: Do we want provider to always be the first word in this string?

vault/external_tests/raft/raft_test.go
```diff
@@ -40,6 +41,56 @@ func raftCluster(t testing.TB) *vault.TestCluster {
...
+	{
+		testhelpers.EnsureCoreSealed(t, leaderCore)
+		leaderCore.UnderlyingRawStorage.(*raft.RaftBackend).SetServerAddressProvider(addressProvider)
+		cluster.UnsealCore(t, leaderCore)
+		vault.TestWaitActive(t, leaderCore.Core)
+	}
+
+	leaderInfos := []*raft.LeaderJoinInfo{
+		{
+			AutoJoin:  "provider=aws region=eu-west-1 tag_key=consul tag_value=tag access_key_id=a secret_access_key=a",
```
HridoyRoy, line 74: Are there a set of params that the user has to specify to use auto join? Like, provider and region make sense, but are there other fields that can be used instead of the fields here that specify, as I understand, an IAM user? If so, would it be helpful to have a list of fields that autoJoin can use to join a cluster? 

I might also just have this question because I'm new to the raft library :).

command/operator_raft_join.go
alexanderbez: Yes, it's the format for cloud auto-join. See [go-discover](https://github.com/hashicorp/go-discover).

vault/external_tests/raft/raft_test.go
alexanderbez: Good question. It's entirely ambiguous and dependent on the cloud provider and operator infrastructure. See [go-discover](https://github.com/hashicorp/go-discover).

command/operator_raft_join.go
ncabatoff: Is it though?  Roy raises a good point: there's nothing in the docs that say that the provider pair must come first in the list of kv pairs.  And the very fact that consul uses `Contains` suggests that they support it anywhere in the list.

alexanderbez: You're absolutely right. I thought I recalled reading the Parse logic and that this was expected.

swayne275 (approved):
Looks good!

Issue #9930, opened 2020-09-11
Vault - Panic after Fictitious Verify Rekey


aphorise:
An unexpected panic is encountered when attempting to cancel a **rekey** that's **not in progress** using the latest **1.5.3**

**To Reproduce**
Run a new vault instance unsealed and ready for consuming - and then immediatly request cancellation of rekey verification using either CLI or API:

```bash
vault operator rekey -verify -cancel ;
curl -X DELETE -H "X-Vault-Token: ${VAULT_TOKEN}" ${VAULT_ADDR}/v1/sys/rekey/verify
# // api or curl responds with:
  # curl: (92) HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR (err 2)
```

```
# // … denotes date omission.
… 2020-09-11T09:53:09.307Z [INFO]  http2: panic serving 192.168.178.253:36410: runtime error: invalid memory address or nil pointer dereference
… goroutine 237 [running]:
… net/http.(*http2serverConn).runHandler.func1(0xc000581370, 0xc0042d5f8e, 0xc00416a480)
…         /goroot/src/net/http/h2_bundle.go:5713 +0x16b
… panic(0x35bbda0, 0x67f4670)
…         /goroot/src/runtime/panic.go:969 +0x166
… github.com/hashicorp/vault/vault.(*Core).RekeyVerifyRestart(0xc0004ac000, 0x2c44f00, 0x0, 0x0)
…         /gopath/src/github.com/hashicorp/vault/vault/rekey.go:957 +0x29e
… github.com/hashicorp/vault/http.handleSysRekeyVerifyDelete(0x452b9a0, 0xc000aa40c0, 0xc0004ac000, 0xc000a29b00, 0x4504ba0, 0xc000581370, 0xc000205b00)
…         /gopath/src/github.com/hashicorp/vault/http/sys_rekey.go:298 +0x38
… github.com/hashicorp/vault/http.handleSysRekeyVerify.func1(0x4504ba0, 0xc000581370, 0xc000205b00)
…         /gopath/src/github.com/hashicorp/vault/http/sys_rekey.go:250 +0x1b8
… net/http.HandlerFunc.ServeHTTP(0xc0001265c0, 0x4504ba0, 0xc000581370, 0xc000205b00)
…         /goroot/src/net/http/server.go:2041 +0x44
… github.com/hashicorp/vault/http.handleRequestForwarding.func1(0x4504ba0, 0xc000581370, 0xc000205b00)
[...]

PR #10099, opened 2020-10-06
Fix panic in RekeyVerifyRestart (#9930)

ivanlemeshev:
A panic occurs if recoveryRekeyConfig or barrierRekeyConfig is a nil pointer in RekeyVerifyRestart. Added a small nil check.


raskchanky (approved):
Looks good, thank you!

PR #10102, opened 2020-10-06
Ui/pricing metric config

chelshaw:
This PR adds a route for editing the pricing metrics config, and implements the edit functionality. 

**The edit configuration button is available from the config route**
metrics-config-edit-button

**Config form:**
metrics-config-form

**If the enabled value is being changed to on, this message shows when user clicks "Submit"**
metrics-enable-warning

**If the enabled value is being changed to off, this message shows when user clicks "Submit"**
metrics-disable-warning

**Errors, including if the input is not a valid number, show on the form**
invalid-error


ui/app/components/pricing-metrics-config.js
```diff
@@ -0,0 +1,76 @@
...
+    try {
+      yield model.save();
+    } catch (err) {
+      this.set('error', err.message);
+      return;
+    }
+    this.router.transitionTo('vault.cluster.metrics.config');
+  })
+    .drop()
+    .withTestWaiter(),
```
Monkeychip, line 63: withTestWaiter won't work in the new ember upgrade.   I haven't worked through a replacement, as it looks like the replacement is to refactor the test.   So for now --in the upgrade-- I've been commenting these out.  If there's any way not to use this guy, that would be helpful.  This isn't a blocking comment.

ui/app/models/metrics/config.js
```diff
@@ -1,10 +1,32 @@
 import DS from 'ember-data';
+import { computed } from '@ember/object';
+// import { apiPath } from 'vault/macros/lazy-capabilities';
```
Monkeychip, line 3: Should these be removed?

ui/app/serializers/metrics/config.js
```diff
@@ -11,4 +15,19 @@ export default ApplicationSerializer.extend({
     };
     return this._super(store, primaryModelClass, normalizedPayload, id, requestType);
   },
+
+  serialize() {
```
Monkeychip, line 19: nice!

ui/app/styles/components/selectable-card-container.scss
```diff
@@ -15,13 +15,6 @@
     grid-template-columns: 2fr;
   }
 
-  .grid-item-http {
```
Monkeychip, line 18: thanks for making sure to clean up the scss.

ui/app/templates/components/pricing-metrics-config.hbs
```diff
@@ -0,0 +1,73 @@
...
+        {{#link-to
+          "vault.cluster.metrics.config"
+          class="button"
+        }}
+          Cancel
+        {{/link-to}}
+      </div>
+    </div>
+  </form>
+<Modal
```
Monkeychip, line 32: is this within the if statement?  if so, looks like it needs to be indented.

ui/app/templates/vault/cluster/metrics/config.hbs
```diff
@@ -37,12 +37,15 @@
...
-    {{info-table-row
-      label=item.label
-      helperText=item.helperText
-      value=(get model item.valueKey)
-    }}
-  {{/each}}
-</div>
+<Toolbar>
+  <ToolbarActions>
+    {{!-- {{#if capabilities.canUpdate }} --}}
```
Monkeychip, line 42: remove the commented if statement?

ui/app/templates/vault/cluster/metrics/edit.hbs
```diff
@@ -0,0 +1,9 @@
+<PageHeader as |p|>
+  <p.levelLeft>
+    <h1 class="title is-3">
+      Configure usage tracking
+    </h1>
+  </p.levelLeft>
+</PageHeader>
+{{!-- CBS TODO: Move to component --}}
+<PricingMetricsConfig @model={{model}} @mode="edit" />
```
Monkeychip, line 9: missing space at end of file

Monkeychip (approved):
When trying to edit the config nothing actually happens. Confirming that is because we're still on a mock API? https://share.getcloudapp.com/nOunNP88

Also, something I can discuss with Ivana/design but because the retention and default display are number fields maybe we should consider replace the input text field for numbers fields?

ui/app/components/pricing-metrics-config.js
chelshaw: Yeah, i was leaving it in ahead of writing tests but I can just avoid that function in the test

ui/app/models/metrics/config.js
chelshaw: Yes, but this is a good reminder to make another follow-up ticket for handling non-operator policies

chelshaw:
> When trying to edit the config nothing actually happens. Confirming that is because we're still on a mock API? https://share.getcloudapp.com/nOunNP88
> 
> Also, something I can discuss with Ivana/design but because the retention and default display are number fields maybe we should consider replace the input text field for numbers fields?

Yes re: nothing happens because it's a mock API, so the changes don't actually post. I'm expecting to do at least a little bit of cleanup once the real API is available, but this should generally set us up for success based on the documentation so far. 

I think it's worth asking Ivana about, and if we decide the native html number input is too clunky we could do some custom validation on a text input to make sure the output is a number instead of doing that in the serializer like I did here. Would be easier with the validation library, of course .... 😄 but all in due time.

Issue #9310, opened 2020-06-24
Vault crashes when attempting to generate-root while Vault is sealed


thomasmitchell:
**Describe the bug**
If you attempt to initialize a new generate root token session but the Vault is sealed, the Vault server crashes.

**To Reproduce**
1. Start a `vault server -dev`
2. Seal the vault server
3. run `vault operator generate-root -generate-otp`
2. `vault operator generate-root -init -otp 'yourotp'`
3. See that the Vault has crashed

**Expected behavior**
I expected a returned error code telling me that the Vault is sealed.
**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.4.2
* Vault CLI Version (retrieve with `vault version`): 1.4.2
* Server Operating System/Architecture: linux/amd64 under WSL2 on Windows

Vault server configuration file(s):

`vault server -dev`

**Additional context**

```
nil pointer dereference
goroutine 374 [running]:
net/http.(*conn).serve.func1(0xc000a228c0)
        /usr/local/go/src/net/http/server.go:1767 +0x139
panic(0x30582c0, 0x5f6dcc0)
        /usr/local/go/src/runtime/panic.go:679 +0x1b2
sync.(*RWMutex).RLock(...)
        /usr/local/go/src/sync/rwmutex.go:48
github.com/hashicorp/vault/vault.(*AuditBroker).LogRequest(0x0, 0x3e11320, 0xc0007efe30, 0xc00028b5e0, 0xc000944150, 0x0, 0x0)
        /go/src/github.com/hashicorp/vault/vault/audit_broker.go:92 +0x117
github.com/hashicorp/vault/vault.(*basicAuditor).AuditRequest(0xc0005f0340, 0x3e11320, 0xc0007efe30, 0xc00028b5e0, 0xc000a2a640, 0x0)
        /go/src/github.com/hashicorp/vault/vault/audit.go:519 +0x5f
github.com/hashicorp/vault/http.handleAuditNonLogical.func1(0x3defae0, 0xc0009d8540, 0xc000967000)
        /go/src/github.com/hashicorp/vault/http/handler.go:236 +0x2d0
net/http.HandlerFunc.ServeHTTP(0xc000b8b100, 0x3defae0, 0xc0009d8540, 0xc000967000)
        /usr/local/go/src/net/http/server.go:2007 +0x44
github.com/hashicorp/vault/http.handleRequestForwarding.func1(0x3defae0, 0xc0009d8540, 0xc000967000)
[...]

PR #10103, opened 2020-10-06
Fix panic in handleAuditNonLogical if vault is sealed (#9310)

ivanlemeshev:
A panic occurs if auditBroker is a nil pointer when vault is sealed. Added a nil check and return of ErrSealed if it is nil.


ncabatoff (approved):

ncabatoff:
Thanks @ivanlemeshev !

PR #10116, opened 2020-10-08
Update ad plugin to v0.6.7

jasonodonnell:
A small bug was found in `ad/config` where `password_policy` had the wrong key name when the config was read. https://github.com/hashicorp/vault-plugin-secrets-ad/pull/71

`go mod vendor` pulled in other things which seem to be missing from the repo.


calvn (approved):
Looks like there were other changes aside from the `password_policy` change by going from 0.6.6 to 0.6.7. This is fine, but we'd probably want to include all of the changes in the changelog.

PR #10118, opened 2020-10-08
DBPW - Update InfluxDB to adhere to v5 Database interface

pcman312:
# Overview
Updates the InfluxDB Database plugin to adhere to the v5 Database interface.

### DIFF
--- a/plugins/database/influxdb/influxdb.go
+++ b/plugins/database/influxdb/influxdb.go
@@ -2,14 +2,15 @@ package influxdb
 
 import (
 	"context"
+	"fmt"
 	"strings"
-	"time"
+
+	"github.com/hashicorp/vault/sdk/database/helper/credsutil"
 
 	multierror "github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/vault/api"
-	"github.com/hashicorp/vault/sdk/database/dbplugin"
-	"github.com/hashicorp/vault/sdk/database/helper/credsutil"
 	"github.com/hashicorp/vault/sdk/database/helper/dbutil"
+	"github.com/hashicorp/vault/sdk/database/newdbplugin"
 	"github.com/hashicorp/vault/sdk/helper/strutil"
 	influx "github.com/influxdata/influxdb/client/v2"
 )
@@ -21,18 +22,17 @@ const (
 	influxdbTypeName                  = "influxdb"
 )
 
-var _ dbplugin.Database = &Influxdb{}
+var _ newdbplugin.Database = &Influxdb{}
 
 // Influxdb is an implementation of Database interface
 type Influxdb struct {
 	*influxdbConnectionProducer
-	credsutil.CredentialsProducer
 }
 
 // New returns a new Cassandra instance
 func New() (interface{}, error) {
 	db := new()
-	dbType := dbplugin.NewDatabaseErrorSanitizerMiddleware(db, db.secretValues)
+	dbType := newdbplugin.NewDatabaseErrorSanitizerMiddleware(db, db.secretValues)
 
 	return dbType, nil
 }
@@ -41,16 +41,8 @@ func new() *Influxdb {
 	connProducer := &influxdbConnectionProducer{}
 	connProducer.Type = influxdbTypeName
 
-	credsProducer := &credsutil.SQLCredentialsProducer{
-		DisplayNameLen: 15,
-		RoleNameLen:    15,
-		UsernameLen:    100,
-		Separator:      "_",
-	}
-
 	return &Influxdb{
 		influxdbConnectionProducer: connProducer,
-		CredentialsProducer:        credsProducer,
 	}
 }
 
@@ -61,7 +53,7 @@ func Run(apiTLSConfig *api.TLSConfig) error {
 		return err
 	}
 
-	dbplugin.Serve(dbType.(dbplugin.Database), api.VaultPluginTLSProvider(apiTLSConfig))
+	newdbplugin.Serve(dbType.(newdbplugin.Database), api.VaultPluginTLSProvider(apiTLSConfig))
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


Valarissa:
Tests pass locally, other than the comment nit, it all LGTM

plugins/database/influxdb/influxdb.go
```diff
@@ -80,64 +72,65 @@ func (i *Influxdb) getConnection(ctx context.Context) (influx.Client, error) {
 	return cli.(influx.Client), nil
 }
 
-// CreateUser generates the username/password on the underlying Influxdb secret backend as instructed by
+// NewUser generates the username/password on the underlying Influxdb secret backend as instructed by
 // the CreationStatement provided.
```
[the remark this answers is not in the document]
pcman312: Done

plugins/database/influxdb/influxdb_test.go
```diff
@@ -192,112 +331,104 @@ func TestInfluxdb_RevokeDeletedUser(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("err: %s", err)
-	}
-
-	if !connProducer.Initialized {
-		t.Fatal("Database should be initialized")
-	}
-
-	newConf, err := db.RotateRootCredentials(context.Background(), nil)
+func assertCredsExist(t testing.TB, address, username, password string) {
```
tvoran, line 409:
```suggestion

func assertCredsExist(t testing.TB, address, username, password string) {
```

plugins/database/influxdb/influxdb_test.go
```diff
@@ -192,112 +331,104 @@ func TestInfluxdb_RevokeDeletedUser(t *testing.T) {
...
-	username, password, err := db.CreateUser(context.Background(), statements, usernameConfig, time.Now().Add(time.Minute))
-	if err != nil {
-		t.Fatalf("err: %s", err)
-	}
+	assertCredsExist(t, config.URL().String(), newUserResp.Username, initialPassword)
 
-	config.Username, config.Password = username, password
-	if err = testCredsExist(t, config); err != nil {
-		t.Fatalf("Could not connect with new credentials: %s", err)
+	// attempt to revoke the user after database is gone
```
tvoran, line 399: I don't think the database is gone at this point?

plugins/database/influxdb/influxdb_test.go
```diff
@@ -192,112 +331,104 @@ func TestInfluxdb_RevokeDeletedUser(t *testing.T) {
...
+	delReq := newdbplugin.DeleteUserRequest{
+		Username: newUserResp.Username,
 	}
-
-	// Test default revoke statements
-	err = db.RevokeUser(context.Background(), dbplugin.Statements{}, username)
+	_, err := db.DeleteUser(context.Background(), delReq)
```
tvoran, line 403: Would `dbtesting.AssertDeleteUser()` work here?

tvoran (commented):
Just a few niggles.

plugins/database/influxdb/influxdb.go
```diff
@@ -196,51 +178,60 @@ func (i *Influxdb) RevokeUser(ctx context.Context, statements dbplugin.Statement
...
+	}
+	return newdbplugin.DeleteUserResponse{}, nil
 }
 
-// RotateRootCredentials is useful when we try to change root credential
-func (i *Influxdb) RotateRootCredentials(ctx context.Context, statements []string) (map[string]interface{}, error) {
-	// Grab the lock
+func (i *Influxdb) UpdateUser(ctx context.Context, req newdbplugin.UpdateUserRequest) (newdbplugin.UpdateUserResponse, error) {
 	i.Lock()
 	defer i.Unlock()
```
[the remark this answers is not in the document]
pcman312: Added, good catch

plugins/database/influxdb/influxdb_test.go
```diff
@@ -96,8 +100,73 @@ func TestInfluxdb_Initialize(t *testing.T) {
...
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			db := new()
+			defer dbtesting.AssertClose(t, db)
+
+			req := newdbplugin.InitializeRequest{
+				Config:           config.connectionParams(),
+				VerifyConnection: true,
+			}
+			resp, err := db.Initialize(context.Background(), req)
```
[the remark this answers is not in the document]
pcman312: Yes it should. Let this be a lesson: Don't put up a PR when you're very tired.

plugins/database/influxdb/influxdb_test.go
```diff
@@ -113,75 +182,145 @@ func TestInfluxdb_Initialize(t *testing.T) {
...
 	connectionParams["port"] = strconv.Itoa(connectionParams["port"].(int))
-	_, err = db.Init(context.Background(), connectionParams, true)
+	req = newdbplugin.InitializeRequest{
+		Config:           connectionParams,
+		VerifyConnection: true,
+	}
+	_, err = db.Initialize(context.Background(), req)
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
```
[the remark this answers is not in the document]
pcman312: They are - I neglected to remove them. Removed now.

plugins/database/influxdb/influxdb_test.go
pcman312: 👍

plugins/database/influxdb/influxdb_test.go
pcman312: The call to `cleanup()` above will destroy the running container, however I just realized that this test will only work properly if using a container rather than a currently running Influx instance since that path doesn't do anything when calling `cleanup`.

plugins/database/influxdb/influxdb_test.go
pcman312: Also see below: after calling `cleanup()` the container is destroyed. However I realized the test can run against an already running Influx instance and then `cleanup()` does nothing and the test would fail. I've updated it to attempt to delete a user that doesn't exist.

plugins/database/influxdb/influxdb_test.go
```diff
@@ -96,8 +100,73 @@ func TestInfluxdb_Initialize(t *testing.T) {
 	cleanup, config := prepareInfluxdbTestContainer(t)
 	defer cleanup()
 
+	type testCase struct {
+		req               newdbplugin.InitializeRequest
+		expectedResponse  newdbplugin.InitializeResponse
+		expectErr         bool
+		expectInitialized bool
```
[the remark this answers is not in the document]
pcman312: I meant to add additional tests to this but forgot to. I've added some more that take advantage of these fields.

plugins/database/influxdb/influxdb_test.go
tvoran: That makes sense for `TestInfluxdb_RevokeDeletedUser()`, but this is `TestInfluxdb_RevokeUser()`, which I think is just testing the delete user happy path.

plugins/database/influxdb/influxdb_test.go
tvoran: That makes sense for `TestInfluxdb_RevokeDeletedUser()`, but this is `TestInfluxdb_RevokeUser()`, which I think is just testing the delete user happy path.

plugins/database/influxdb/influxdb_test.go
pcman312: Oh, you're correct. The comment is copy-pasted and needs to be removed.

plugins/database/influxdb/influxdb_test.go
pcman312: That's what happens when I put my comments in the wrong text box. You are correct: this can use `dbtesting.AssertDeleteUser`

tvoran (approved):

PR #10119, opened 2020-10-08
Ui/pricing metric sparkle

chelshaw:
This PR adds some more safeguards against pricing metrics date inputs, as well as capabilities checking

Config tab is hidden when a policy denies access to metrics config, such as: 
```hcl
path "sys/internal/counters/*" {
    capabilities = ["read"]
}

path "sys/internal/counters/config" {
    capabilities = ["deny"]
}
```
policy-no-config-update

Config edit button is hidden when a policy allows only read access to metrics config, such as: 
```hcl
path "sys/internal/counters/*" {
    capabilities = ["read"]
}
```
policy-no-config-update

Error message appears when the start date is father than the retained months setting
start-date-error

Error message appears when the end date is the current month or later
end-date-error

### DIFF
--- a/ui/app/templates/vault/cluster/metrics/index.hbs
+++ b/ui/app/templates/vault/cluster/metrics/index.hbs
@@ -21,18 +21,20 @@
           Vault usage
         {{/link-to}}
       {{/link-to}}
-      {{#link-to
-        "vault.cluster.metrics.config"
-        tagName="li"
-        activeClass="is-active"
-      }}
+      {{#if model.config.configPath.canRead}}
         {{#link-to
           "vault.cluster.metrics.config"
-          data-test-configuration-tab=true
+          tagName="li"
+          activeClass="is-active"
         }}
-          Configuration
+          {{#link-to
+            "vault.cluster.metrics.config"
+            data-test-configuration-tab=true
+          }}
+            Configuration
+          {{/link-to}}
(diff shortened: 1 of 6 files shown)


ui/app/templates/vault/cluster/metrics/index.hbs
```diff
@@ -21,18 +21,20 @@
           Vault usage
         {{/link-to}}
       {{/link-to}}
-      {{#link-to
-        "vault.cluster.metrics.config"
-        tagName="li"
-        activeClass="is-active"
-      }}
+      {{#if model.config.configPath.canRead}}
         {{#link-to
```
Monkeychip, line 25: link-to within another link-to, is that right, or am I reading it wrong?

Monkeychip (approved):
Nice work!

ui/app/templates/vault/cluster/metrics/index.hbs
chelshaw: This was copypasta but you're right, it doesn't look like it needs to have both nested 🤔

chelshaw: Just confirmed, this is how the tabs are in `secret-list-header` component, and removing the inner one causes strange styling issues. Gonna leave it for now, but worth looking into in the future

PR #10122, opened 2020-10-09
fix by returning apiPath from the model

Monkeychip:
This PR fixes an issue where on route change the model is not refreshed and therefore the path to the auth method for listing roles was not being called properly. Originally this was reported in an Asana ticket for Kubernetes, but it's an issue with all auth methods.  

The fix was to use the store to return the `generated-item-list` adapter that is then passed to the `path-help` service.

Below you can see the change.  Where before you'd see the same role list from the previous auth method selected.  It would only change on refresh which would cause the model to refresh.

auth

Additionally, when I was writing a test I noticed that in certain cases if you transitioned between auth methods and created a role or user you would create that user at the endpoint of the previous route you had visited.  Here's a video of that happening.  The test I wrote checks both for this and the original fix.  This can still happen if you copy/paste to the create link, but that does not follow the regular flow of a UI user.  However, I thought it was worthing pointing out.  It has something to do with how the path is first created and not refreshed, but I was unable to get to the meat of the problem.  

Screen Recording 2020-10-13 at 12 01 02 PM

### DIFF
--- a/ui/app/services/path-help.js
+++ b/ui/app/services/path-help.js
@@ -232,8 +232,8 @@ export default Service.extend({
         let url;
         id = encodePath(id);
         // the apiPath changes when you switch between routes but the apiPath variable does not unless the model is reloaded
-        // check first for the dynamicApiPath which is passed through from the model->adapter
-        // If no dynamicApiPath, then use apiPath set higher up in function
+        // overwrite apiPath if dynamicApiPath exist.
+        // dynamicApiPath comes from the model->adapter
         if (dynamicApiPath) {
           apiPath = dynamicApiPath;
         }
--- a/ui/app/services/path-help.js
+++ b/ui/app/services/path-help.js
@@ -226,15 +226,18 @@ export default Service.extend({
     const deletePath = paths.find(path => path.operations.includes('delete'));
 
     return generatedItemAdapter.extend({
-      urlForItem(id, isList) {
+      urlForItem(id, isList, dynamicApiPath) {
         const itemType = getPath.path.slice(1);
         let url;
         id = encodePath(id);
-
+        // the apiPath changes when you switch between routes but the apiPath variable does not unless the model is reloaded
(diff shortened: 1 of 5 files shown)


ui/app/services/path-help.js
```diff
@@ -23,6 +23,7 @@ export function sanitizePath(path) {
 
 export default Service.extend({
   attrs: null,
+  dynamicApiPath: '',
```
chelshaw, line 26: We don't need this set here if it's just being passed into `urlForItem` right?

ui/app/services/path-help.js
```diff
@@ -226,11 +227,16 @@ export default Service.extend({
...
     return generatedItemAdapter.extend({
-      urlForItem(id, isList) {
+      urlForItem(id, isList, dynamicApiPath) {
         const itemType = getPath.path.slice(1);
         let url;
         id = encodePath(id);
-
+        // the apiPath changes when you switch between routes but the apiPath variable does not unless the model is reloaded
+        // check first for the dynamicApiPath which is passed through from the model->adapter
+        // If no dynamicApiPath, then use apiPath set higher up in function
```
chelshaw, line 236: nit: I find the wording here a bit confusing. It's not exactly checking for dynamicApiPath "first", it's just overwriting `apiPath` if the value exists

ui/tests/acceptance/auth-list-test.js
```diff
@@ -0,0 +1,59 @@
+import { click, fillIn, settled, visit } from '@ember/test-helpers';
```
chelshaw, line 1: 👏

chelshaw (approved):
Just a couple minor things! Great work on this

ui/app/services/path-help.js
Monkeychip: When writing a test, I noticed that it needs to be set as an empty string because it will hit getDynamicApiPath before result.apiPath is set.  It then hits it again when it is set.

PR #10131, opened 2020-10-12
Aerospike storage backend

reugn:
This adds support for [Aerospike](https://www.aerospike.com/) as a (non-HA) storage backend for Vault.


reugn:
@jasonodonnell , @calvn , @catsby , @Valarissa , @vishalnayak , @kalafut could you please review this PR.

kalafut:
@reugn Thanks for the PR! Yes, we'll be reviewing this soon.

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := aero.NewClientWithPolicy(policy, conf["hostname"], port)
+	if err != nil {
+		return nil, err
+	}
+
+	return &AerospikeBackend{
```
kalafut, line 50: Since (per the docs), `set`/`namespace`/`hostname` are all required, should we be validating `!= ""` before successfully returning?

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// Put is used to insert or update an entry.
+func (a *AerospikeBackend) Put(ctx context.Context, entry *physical.Entry) error {
+	aeroKey, err := a.key(entry.Key)
+	if err != nil {
+		return err
+	}
+
+	writePolicy := aero.NewWritePolicy(0, 0)
```
kalafut, line 69: I comment or docs link that has some details about this would be helpful.

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+// Get is used to fetch an entry.
+func (a *AerospikeBackend) Get(ctx context.Context, key string) (*physical.Entry, error) {
+	aeroKey, err := a.key(key)
+	if err != nil {
+		return nil, err
+	}
+
+	record, err := a.client.Get(nil, aeroKey)
+	if err != nil {
+		if err.Error() == "Key not found" {
```
kalafut, line 88: Is the compared-to error exported by chance?

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	return &physical.Entry{
+		Key:   key,
+		Value: record.Bins[valueBin].([]byte),
```
kalafut, line 96: Considering indexing against `valueBin` in advance so we can check for the presence of the key and error (otherwise we'll panic on nil)

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+		logger:    logger,
+	}, nil
+}
+
+func (a *AerospikeBackend) key(userKey string) (*aero.Key, error) {
+	return aero.NewKey(a.namespace, a.set, hash(userKey))
+}
+
+// Put is used to insert or update an entry.
+func (a *AerospikeBackend) Put(ctx context.Context, entry *physical.Entry) error {
```
kalafut, line 63: s/ctx/_  on these unused `ctx` parameters

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	var keyList []string
+	for res := range recordSet.Results() {
+		if res.Err != nil {
+			return nil, res.Err
+		}
+		recordKey := res.Record.Bins[keyBin].(string)
+		if strings.HasPrefix(recordKey, prefix) {
+			trimPrefix := strings.Replace(recordKey, prefix, "", 1)
```
kalafut, line 126: `strings.TrimPrefix` is a little simpler.

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+			keys := strings.Split(trimPrefix, "/")
+			if len(keys) == 1 {
+				keyList = append(keyList, keys[0])
+			} else {
+				withSlash := keys[0] + "/"
+				if !listContains(keyList, withSlash) {
```
kalafut, line 132: We have a helper you can use: https://github.com/hashicorp/vault/blob/8af1f3b57710660b6b5c7abd6966ebc0e3e2da51/sdk/helper/strutil/strutil.go#L26

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+				}
+			}
+		}
+	}
+
+	return keyList, nil
+}
+
+func hash(s string) string {
+	h := fnv.New32a()
```
kalafut, line 143: Any reason this was chosen in particular? 32-bits is a fairly small hash size. We'd typically use sha256 to satisfy both irreversibility and collision avoidance.

website/pages/docs/configuration/storage/aerospike.mdx
```diff
@@ -0,0 +1,50 @@
...
+page_title: Aerospike - Storage Backends - Configuration
+sidebar_title: Aerospike
+description: |-
+  The Aerospike storage backend is used to persist Vault's data in Aerospike
+  cluster.
+---
+
+# Aerospike Storage Backend
+
+The Aerospike storage backend is used to persist Vault's data in
```
kalafut, line 12:
```suggestion
The Aerospike storage backend is used to persist Vault's data in an
```

website/pages/docs/configuration/storage/aerospike.mdx
```diff
@@ -0,0 +1,50 @@
+---
+layout: docs
+page_title: Aerospike - Storage Backends - Configuration
+sidebar_title: Aerospike
+description: |-
+  The Aerospike storage backend is used to persist Vault's data in Aerospike
```
kalafut, line 6:
```suggestion
  The Aerospike storage backend is used to persist Vault's data in an Aerospike
```

kalafut (changes requested):
Looks pretty good! Some initial feedback, mostly minor.

website/pages/docs/configuration/storage/aerospike.mdx
```diff
@@ -0,0 +1,50 @@
+---
```
kalafut, line 1: Need to update the JSON to reference this as well: https://github.com/hashicorp/vault/blob/f66c721e79fec9e51bd76841969db92a830cc6e7/website/data/docs-navigation.js#L67

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+// NewAerospikeBackend constructs an AerospikeBackend backend.
+func NewAerospikeBackend(conf map[string]string, logger log.Logger) (physical.Backend, error) {
+	namespace := conf["namespace"]
+	set := conf["set"]
+
+	policy := aero.NewClientPolicy()
+	policy.User = conf["username"]
+	policy.Password = conf["password"]
+
+	port, err := strconv.Atoi(conf["port"])
```
calvn, line 40: If `port` is not provided, the error returned does not make it obvious that this is the case. Consider checking the value beforehand separately.

```
strconv.Atoi: parsing "": invalid syntax
```

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	_, err = a.client.Delete(nil, aeroKey)
+	return err
+}
+
+// List is used to list all the keys under a given
+// prefix, up to the next prefix.
+func (a *AerospikeBackend) List(ctx context.Context, prefix string) ([]string, error) {
+	recordSet, err := a.client.ScanAll(nil, a.namespace, a.set)
```
calvn, line 114: Client calls does not seem to provide a way to cancel an inflight request through context cancelation. It might be worth exposing `ClientPolicy.Timeout` to be able to tune this value, even if it's a the storage-level.

physical/aerospike/aerospike_test.go
```diff
@@ -0,0 +1,88 @@
...
+	set       string
+}
+
+func prepareAerospikeContainer(t *testing.T) (func(), *aerospikeConfig) {
+	runner, err := docker.NewServiceRunner(docker.RunOptions{
+		ImageRepo:       "aerospike/aerospike-server",
+		ContainerName:   "aerospikedb",
+		ImageTag:        "latest",
+		Ports:           []string{"3000/tcp", "3001/tcp", "3002/tcp", "3003/tcp"},
+		DoNotAutoRemove: true,
```
calvn, line 49: Is this strictly necessary? This leaves dangling containers on the host, and may result in errors if tests are ran multiple times due to colliding container name.

website/pages/docs/configuration/storage/aerospike.mdx
calvn, line 1: It also needs an entry in [this file](https://github.com/hashicorp/vault/blob/master/website/data/docs-navigation.js#L67) so that it can be included in the sidebar.

calvn: Looks like Jim was referring to the same thing in the comment above, resolving this one!

reugn:
@kalafut , @calvn thanks for the review. Fixes and improvements were done.

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/SAP/go-hdb v0.14.1
 	github.com/Sectorbob/mlab-ns2 v0.0.0-20171030222938-d3aa0c295a8a
 	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d // indirect
+	github.com/aerospike/aerospike-client-go v3.1.0+incompatible
```
calvn, line 20: I'm a bit curious as to why this particular pseudo-version was chosen since v3.1.1 is available.

physical/aerospike/aerospike.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/hashicorp/vault/sdk/physical"
+)
+
+const (
+	keyBin   = "keyBin"
+	valueBin = "valueBin"
+
+	defaultNamespace = "test"
+
+	defaultHost = "127.0.0.1"
```
calvn, line 23: Nit: Since this is assigned as the default hostname, can we rename this to `defaultHostname`?

website/pages/docs/configuration/storage/aerospike.mdx
```diff
@@ -0,0 +1,66 @@
...
+  cluster. Access Control is available only in Aerospike Server Enterprise Edition.
+  Leave empty for clusters running without restricted access.
+
+- `cluster_name` `(string: "")` - Specifies the expected Aerospike cluster name. If set, server nodes must return this cluster ID in order to
+	join the client's view of the cluster. Should only be set when connecting to servers that
+	support the "cluster-name" info command. (v3.10+)
+
+- `auth_mode` `(string: "INTERNAL")` - Specifies the authentication mode when user/password is defined (INTERNAL/EXTERNAL).
+
+- `timeout` `(int: 30000)` - Initial host connection timeout duration in milliseconds.
```
calvn, line 60: Where does the `timeout` and `idle_timeout` values come from?

physical/aerospike/aerospike.go
calvn: Looks like this is now handled via `timeout` and `idle_timeout`.

go.mod
calvn: Interesting, the `+incompatible` label is actually how go/mod sees the release: https://pkg.go.dev/github.com/aerospike/aerospike-client-go@v3.1.0+incompatible?tab=versions

Since `v3.1.1+incompatible` has been released, we should consider bumping it to that version (unless there's a particular reason not to)

physical/aerospike/aerospike.go
calvn: Looks like `set` is optional, and both `namespace` and `hostname` now have default values so we should be good.

calvn (approved):
Some minor suggestions, otherwise looks good!

reugn:
@calvn for some reason, it doesn't pass the `test-go-race-remote-docker` with the aerospike client v3.1.1 .

calvn:
Can you give v3.1.1 another try? The failure seems to be from a non-related test:
```
Failed
=== RUN   TestRaft_SnapshotAPI_RekeyRotate_Backward
--- FAIL: TestRaft_SnapshotAPI_RekeyRotate_Backward (0.00s)
```

There also seems to be a conflict with go.sum since your latest push. It's looking ready though!

reugn:
Still failed.

calvn:
Looks like that's a known flaky test. I'm kicking off CI again to see if we can get a passing run.

calvn:
@reugn can you merge master into this PR to get the latest set of changes and ensure that things are still working?

calvn:
Can you move the docs page to `website/content/docs/configuration/storage/`? We recently changed where these files live.

PR #10132, opened 2020-10-12
docs: add missing "_prefix" to Consul 1.4+ policy "service" rule

akurz:
Hello,

just noticed that missing "_prefix" in the example Consul policy rules when using it as Vault storage backend.

Cheers,
Andreas


akurz:
Sorry for the noise, just realized that prefix is not wanted.

PR #10138, opened 2020-10-13
DBPW - Update docs with password policies & new Database interface

pcman312:
# Overview
Web docs updates:
  - Support for password policies
  - New Database interface for custom plugins


website/pages/docs/secrets/databases/cassandra.mdx
```diff
@@ -74,8 +74,8 @@ the proper permission, it can generate credentials.
     lease_id           database/creds/my-role/2f6a614c-4aa2-7b19-24b9-ad944a8d4de6
     lease_duration     1h
     lease_renewable    true
-    password           8cab931c-d62e-a73d-60d3-5ee85139cd66
-    username           v-vaultuser-e2978cd0-
+    password           7iO-FxlVCdj8szsgGgsL
+    username           v_vaultuser_my_role_6zjahr4mjn5c9prgucy1_1602541475
```
calvn, line 78: Not related to this PR in particular, but did we change the separator to `_` recently?

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -12,81 +12,164 @@ description: |-
...
+| API Argument               | Request Object                                     |
+| -------------------------- | -------------------------------------------------- |
+| `creation_statements`      | `NewUserRequest.Statements.Commands`               |
+| `revocation_statements`    | `DeleteUserRequest.Statements.Commands`            |
+| `rollback_statements`      | `NewUserRequest.RollbackStatements.Commands`       |
+| `renew_statements`         | `UpdateUserRequest.Expiration.Statements.Commands` |
+| `rotation_statements`      | `UpdateUserRequest.Password.Statements.Commands`   |
+| `root_rotation_statements` | `UpdateUserRequest.Password.Statements.Commands`   |
+
+In many of the built in plugins, they replace `{{name}}` (or `{{username`), `{{password}}`,
```
calvn, line 85:
```suggestion
In many of the built in plugins, they replace `{{name}}` (or `{{username}}`), `{{password}}`,
```

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -12,81 +12,164 @@ description: |-
...
+function. Use the `Serve` function within `sdk/database/dbplugin/v5` to serve your plugin. You
+will also need to pass some TLS configuration information that Vault uses when initializing the
+plugin. Below is an example setup:
 
 ```go
 package main
 
 import (
-    "github.com/hashicorp/vault/api/plugins"
+	"github.com/hashicorp/vault/api/plugins"
```
calvn, line 114: Not sure if intentional, but spacing is set to 1 space on per tab in here.

website/pages/docs/secrets/databases/oracle.mdx
```diff
@@ -118,8 +118,8 @@ the proper permission, it can generate credentials.
     lease_id           database/creds/my-role/2f6a614c-4aa2-7b19-24b9-ad944a8d4de6
     lease_duration     1h
     lease_renewable    true
-    password           8cab931c-d62e-a73d-60d3-5ee85139cd66
-    username           v-root-e2978cd0-
+    password           yRUSyd-vPYDg5NkU9kDg
+    username           V_VAULTUSE_MY_ROLE_SJJUK3Q8W3BKAYAN8S62_1602543009
```
calvn, line 122: Why is the formatting changed for usernames on some of the DB's, or were the previous examples incorrect?

website/pages/docs/secrets/databases/index.mdx
```diff
@@ -124,25 +126,27 @@ the proper permission, it can generate credentials.
...
-    username           v-vaultuser-e2978cd0-
+    password           FSREZ1S0kFsZtLat-y94
+    username           v-vaultuser-e2978cd0-ugp7iqI2hdlff5hfjylJ-1602537260
     ```
 
 ## Database Capabilities
+As of Vault 1.6, all databases support dynamic roles, static roles, and rotating the root user's credentials.
+
 | Database | Root Credential Rotation | Dynamic Roles | Static Roles |
 | --- | --- | --- | --- |
```
calvn, line 137: Might also want to format/extend these two lines like you're doing with the other tables.

website/pages/docs/secrets/databases/cassandra.mdx
pcman312: No, it was always using underscores. However many of the docs were copy-pasted and had inaccurate username structures. I updated all of them to accurately represent what usernames will actually be generated.

website/pages/docs/secrets/databases/custom.mdx
pcman312: 👍

website/pages/docs/secrets/databases/oracle.mdx
pcman312: The previous examples were incorrect. Most of them were copy-pasted and never a proper representation of the usernames generated.

website/pages/docs/secrets/databases/custom.mdx
pcman312: That's a tab in every other view except github's diff view 0.o

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -12,81 +12,165 @@ description: |-
...
+The `InitializeRequest` object contains a map of keys to values. This data is what the
 user specified as the configuration for the plugin. Your plugin should use this
-data to make connections to the database. It is also passed a boolean value
-specifying whether or not your plugin should return an error if it is unable to
-connect to the database.
+data to make connections to the database. The response object contains a similar configuration
+map. The response object should contain the configuration map that should be saved within Vault.
+This allows the plugin to manipulate the configuration prior to saving it.
+
+It is also passed a boolean value indicating if your plugin should initialize a connection to
```
calvn, line 97: I think we should rephrase this sentence a bit and directly mention `IntializeRequest.VerifyConnection` as the boolean value that controls this behavior.

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -101,17 +185,91 @@ You should now be able to register your plugin into the vault catalog. To do
...
+   three functions to keep in mind: In the version 4 interface, there was `Initialize` and
+   `Init`. The `Initialize` function was replaced with the `Init` function and was not used
+   some time before Vault 1.6. It was kept in the interface originally for backwards
+   compatibility. The version 5 interface has an `Initialize` function that will need to replace
+   the `Init` function. The two arguments passed into `Init` (`config map[string]interface{}`
+   and `verifyConnection bool`) are included in the `InitializeRequest` object and should be
+   referenced accordingly. The verison 4 `Init` function returned a `map[string]interface{}`
+   indicating the configuration that should be saved in Vault. This is wrapped inside the
+   `InitializeResponse` object within the version 5 interface.
+4. Update `CreateUser` to `NewUser`. The `NewUserRequest` object contains the username and
```
calvn, line 261: Might be worth explicitly calling out this change in behavior from v4 to v5. As in, v4 used to leave password generation up to the plugin which then gets returned back to Vault, but v5 now accepts one from Vault/the request object which then consumes it.

website/pages/docs/secrets/databases/index.mdx
```diff
@@ -13,7 +13,7 @@ description: |-
 
 The database secrets engine generates database credentials dynamically based on
 configured roles. It works with a number of different databases through a plugin
-interface. There are a number of builtin database types and an exposed framework
+interface. There are a number of builtin database types, and an exposed framework
```
austingebauer, line 16: Perhaps `builtin` -> `built in` to match what's in the rest of the docs? Or vice versa.

website/pages/docs/secrets/databases/custom.mdx
pcman312: Good call. I'll put it as a notification block above the steps as it applies to both `NewUser` and `UpdateUser`.

website/pages/docs/secrets/databases/custom.mdx
pcman312: :+1:

website/pages/docs/secrets/databases/index.mdx
pcman312: :+1:

website/pages/docs/secrets/databases/index.mdx
```diff
@@ -69,32 +69,34 @@ management tool.
...
 
    ```shell
    $ vault write -force database/rotate-root/my-database
    ```
 
-   When this is done, the password for the user specified in the previous step
-   is no longer accessible. Because of this, it is highly recommended that a
-   user is created specifically for Vault to use to manage database
-   users.
+   !> When this is done, the password for the user specified in the previous step
```
Valarissa, line 93: Should this be `~>` or am I missing something? I'm fine if I'm missing something.

pcman312: `!>` is an error (red background) whereas `~>` is a warning (yellow background). Given how important this information is, I decided to make it red to grab people's attention more.

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -12,81 +12,166 @@ description: |-
...
+	UpdateUser(ctx context.Context, req UpdateUserRequest) (UpdateUserResponse, error)
+
+	// DeleteUser from the database. This should not error if the user didn't
+	// exist prior to this call.
+	DeleteUser(ctx context.Context, req DeleteUserRequest) (DeleteUserResponse, error)
+
+	// Type returns the Name for the particular database backend implementation.
+	// This type name is usually set as a constant within the database backend
+	// implementation, e.g. "mysql" for the MySQL database backend. This is used
+	// for things like metrics and logging. No behavior is switched on this.
```
calvn, line 60:
```suggestion
	// for things like metrics and logging.
```

I think the last bit is applicable now, but hard to guarantee/enforce.

calvn (approved):

website/pages/docs/secrets/databases/custom.mdx
pcman312: I copy pasted this from the current code to keep consistency. This function has never had any behavior based on it as it is just used as an identifier.

calvn: Fair enough. I guess we do control when and where we call this, so as long as we are mindful of its usage then this may remain to be the case.

website/pages/docs/secrets/databases/index.mdx
Valarissa: Sounds good, I couldn't see that effect on the non-code view, so I wasn't sure. Thanks for clarifying!

website/pages/docs/secrets/databases/index.mdx
```diff
@@ -69,32 +69,34 @@ management tool.
...
+      is no longer accessible. Because of this, it is highly recommended that a
+      user is created specifically for Vault to use to manage database
+      users.
 
-1. Configure a role that maps a name in Vault to an SQL statement to execute to
-   create the database credential:
+1. Configure a role that maps a name in Vault to a set of creation statements to
+   execute to create the database credential:
```
tvoran, line 99:
```suggestion
1. Configure a role that maps a name in Vault to a set of creation statements to
   create the database credential:
```

website/pages/docs/secrets/databases/index.mdx
```diff
@@ -152,12 +156,39 @@ for more information.
...
-function is also used on the output of `GetPassword().`
-
-Please see the [DB plugin credentials source code](https://github.com/hashicorp/vault/blob/master/sdk/database/dbplugin/database.pb.go)
-for more information.
+Passwords are generated via [Password Policies](/docs/concepts/password-policies).
+Databases can optionally set a password policy for use across all roles for that database.
+In other words, each time you call `vault write database/config/my-database` you can specify
+a password policy for all roles using `my-database`. Each database has a default password
+policy defined as: 20 characters with at least 1 uppercase character, at least 1 lowercase
+character, at least one number, and at least one dash character.
```
tvoran, line 164: The first part of the statement uses the number `1`:
```suggestion
character, at least 1 number, and at least 1 dash character.
```

website/pages/docs/secrets/databases/mysql-maria.mdx
```diff
@@ -84,8 +84,8 @@ the proper permission, it can generate credentials.
     lease_id           database/creds/my-role/2f6a614c-4aa2-7b19-24b9-ad944a8d4de6
     lease_duration     1h
     lease_renewable    true
-    password           8cab931c-d62e-a73d-60d3-5ee85139cd66
-    username           v-vaultuser-e2978cd0-
+    password           yY-57n3X5UQhxnmFRP3f
+    username           v_vaultuser_my-role_crBWVqVh2Hc1
```
tvoran, line 88: Should it only be underscores?
```suggestion
    username           v_vaultuser_my_role_crBWVqVh2Hc1
```

website/pages/docs/secrets/databases/hanadb.mdx
```diff
@@ -73,8 +73,8 @@ the proper permission, it can generate credentials.
     lease_id           database/creds/my-role/2f6a614c-4aa2-7b19-24b9-ad944a8d4de6
     lease_duration     1h
     lease_renewable    true
-    password           8cab931c-d62e-a73d-60d3-5ee85139cd66
-    username           v-vaultuser-e2978cd0-
+    password           sU4zWnITaABEP-Zwy4sF
+    username           v-vaultuser-my-role-jQbCLE6P2VtgsqPBXK0m-1602541873
```
tvoran, line 77: I think hanadb is one that doesn't allow hyphens in usernames?
```suggestion
    username           v_vaultuser_my_role_jQbCLE6P2VtgsqPBXK0m_1602541873
```

tvoran (approved):
Just some small suggestions, non-blocking.

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -12,81 +12,166 @@ description: |-
...
+	// Close attempts to close the underlying database connection that was
+	// established by the backend.
 	Close() error
 }
 ```
 
-You'll notice the first parameter to a number of those functions is a
-`Statements` struct. This struct is used to pass the Role's configured
-statements to the plugin on function call. The struct is defined as:
+Each of the request and response objects can also be found in `sdk/database/dbplugin/v5/database.go`.
```
tomhjp, line 69: Is there existing precedent for representing paths to code this way? 2 considerations:

1. Unrooted paths are sometimes confusing
2. This path makes more sense in the context of reading docs in the repo than the website. For someone reading the docs on the website, this path may be doubly confusing because it's both unrooted and it is in no way relative to the website.

OTOH I can see why you've not used absolute paths, because that would require using a branch or blob name, which would either link rot or quickly become out of date. Overall I think what you've done is probably my preferred option, but is there anything we can do to help clarify? Previously I've seen people use e.g. `vault.git/sdk/...`, which has its own flaws but another option.

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -12,81 +12,166 @@ description: |-
...
 
-```go
-type Statements struct {
-	Creation   []string
-	Revocation []string
-	Rollback   []string
-	Renewal    []string
-}
-```
+In each of the requests, you will see a `Statements` object. This object represents the set of
```
tomhjp, line 71:
```suggestion
In each of the requests, you will see at least 1 `Statements` object. This object represents the set of
```

nit: `UpdateUserRequest` has multiple, and the fact they weren't in the top level was slightly confusing for me initially.

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -12,81 +12,166 @@ description: |-
...
-It is up to your plugin to replace the `{{name}}`, `{{password}}`, and
-`{{expiration}}` in these statements with the proper values.
+| API Argument               | Request Object                                     |
+| -------------------------- | -------------------------------------------------- |
+| `creation_statements`      | `NewUserRequest.Statements.Commands`               |
+| `revocation_statements`    | `DeleteUserRequest.Statements.Commands`            |
+| `rollback_statements`      | `NewUserRequest.RollbackStatements.Commands`       |
+| `renew_statements`         | `UpdateUserRequest.Expiration.Statements.Commands` |
+| `rotation_statements`      | `UpdateUserRequest.Password.Statements.Commands`   |
+| `root_rotation_statements` | `UpdateUserRequest.Password.Statements.Commands`   |
```
tomhjp, line 83: Do we state anywhere what happens if you set both `rotation_statements` and `root_rotation_statements`?

Separately to this PR, perhaps we should mark `root_rotation_statements` as deprecated?

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -101,17 +186,96 @@ You should now be able to register your plugin into the vault catalog. To do
...
+
+The new interface is roughly modeled after a [gRPC](https://grpc.io/) interface. It has improved
+future compatibility by not requiring changes to the interface definition to add additional data
+in the requests or responses. It also simplifies the interface by merging several into a single
+function call.
+
+### Upgrading your custom database
+
+Vault 1.6 supports both version 4 and version 5 database plugins. The support for version 4
+plugins will be removed in a future release. If you are using a version 4 custom database
```
tomhjp, line 242: Although it's implied in the instructions below, I think it might be helpful for users if we explicitly tell them how to determine whether their plugin is version 4 or 5 (i.e. check the `sdk` import paths used in its source, A or B)

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -101,17 +186,96 @@ You should now be able to register your plugin into the vault catalog. To do
...
+   (such as the `Serve` function).
+2. An easy way to see what functions that need to be implemented is to put the following as a
+   global variable within your package: `var _ dbplugin.Database = (*MyDatabase)(nil)`. This
+   will fail to compile if the `MyDatabase` type does not adhere to the `dbplugin.Database` interface.
+3. Update the initialization function. This step can be a little confusing since there are
+   three functions to keep in mind: In the version 4 interface, there was `Initialize` and
+   `Init`. The `Initialize` function was replaced with the `Init` function and was not used
+   some time before Vault 1.6. It was kept in the interface originally for backwards
+   compatibility. The version 5 interface has an `Initialize` function that will need to replace
+   the `Init` function. The two arguments passed into `Init` (`config map[string]interface{}`
```
tomhjp, line 262: This block is quite hard to follow. I think some of the historical context could be pulled out into the background section above to leave this bullet as just the imperative instructions required to migrate.

website/pages/docs/secrets/databases/custom.mdx
tomhjp: Or alternatively, augment some of the references to the fairly abstract "version 4" and "version 5" with concrete references to the interfaces implemented.

tomhjp (approved):
This looks great, and the instructions are very helpful for the migration I'm in the middle of, thanks! Just a few comments attempting to improve clarity, but nothing blocking.

website/pages/api-docs/secret/databases/index.mdx
```diff
@@ -48,16 +48,23 @@ list of additional parameters.
...
-   If the plugin supports [rotating the root credentials](#rotate-root-credentials),
-   it is highly recommended you perform that action after configuring the plugin.
+- `password_policy` `(string: "")` - The name of the
+  [password policy](/docs/concepts/password-policies) to use when generating passwords
+  for this database. If not specified, this will use a default policy defined as:
+  20 characters with at least 1 uppercase, 1 lowercase, 1 number, and 1 dash character.
+
+~> We highly recommended that you use a Vault-specific user rather than the admin user
+   in your database when configuring the plugin. This user will be used to create/
+   update/delete users within the database so it will need to have the appropriate
```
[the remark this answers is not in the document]
pcman312: Stupid markdown...

website/pages/docs/secrets/databases/custom.mdx
```diff
@@ -101,17 +186,96 @@ You should now be able to register your plugin into the vault catalog. To do
...
+-> In version 4, password generation was the responsibility of the plugin via the
+   `GenerateCredentials` function. This is no longer the case with version 5. Vault is
+   responsible for generating passwords and passing them to the plugin via
+   `NewUserRequest.Password` and `UpdateUserRequest.Password.NewPassword`.
+
+1. Change the import path from `github.com/hashicorp/vault/sdk/database/dbplugin` to
+   `github.com/hashicorp/vault/sdk/database/dbplugin/v5`. The package name is the same, so any
+   references to `dbplugin` can remain as long as those symbols exist within the new package
+   (such as the `Serve` function).
+2. An easy way to see what functions that need to be implemented is to put the following as a
```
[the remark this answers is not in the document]
pcman312: :+1:

website/pages/docs/secrets/databases/index.mdx
pcman312: :+1:

website/pages/docs/secrets/databases/index.mdx
pcman312: :+1:

website/pages/docs/secrets/databases/mysql-maria.mdx
pcman312: No because the role name is `my-role` and there isn't an explicit change from `-` to `_`

website/pages/docs/secrets/databases/hanadb.mdx
pcman312: :+1:

website/pages/docs/secrets/databases/custom.mdx
pcman312: I don't think there is precedence for this.

1. Not sure what we can do for this. This page is explicitly for people who want to write a custom plugin and need to have knowledge about writing Go.
2. I disagree. The path is to a `.go` file and is in reference to writing a new plugin. I feel the context makes that clearer.

Yea, I agree there are alternatives to this but none of them are particularly good. Since the `sdk` package isn't going anywhere and we try to remain backwards compatible in that package, I feel pretty comfortable with this.

website/pages/docs/secrets/databases/custom.mdx
pcman312: Good call. I also added a note here that `UpdateUserRequest` has them in sub-fields.

website/pages/docs/secrets/databases/custom.mdx
pcman312: `root_rotation_statements` is not deprecated. The DB engine uses it when the `/database/rotate-root/:name` endpoint is called, whereas `rotation_statements` is used during a static credential rotation. The DB engine selects which one based on what action it is performing.

tomhjp: Ah I read too much into how the config was actually stored from this table. Makes sense, thanks for clearing that up for me.

website/pages/docs/secrets/databases/custom.mdx
pcman312: Good call. I've added a small section describing some ways that a user can tell if they are on v4 or v5 as well as references to the files where the interface definitions are.

website/pages/docs/secrets/databases/custom.mdx
pcman312: Updated. Let me know how it reads.

tvoran (approved):

website/pages/docs/secrets/databases/custom.mdx
tomhjp: Way clearer now 💯  nice job

tomhjp (approved):
LGTM!

PR #10142, opened 2020-10-14
Update HanaDb to v5 Database interface

Valarissa:
Updates the HANA Database plugin to adhere to the v5 Database interface.

```
/usr/local/go/bin/go tool test2json -t /private/var/folders/4v/k2rn39v578d238ttqpwdkkkh0000gp/T/___hana_test_go -test.v -test.run ^\QTestHANA_Initialize\E|\QTestHANA_NewUser\E|\QTestHANA_UpdateUser\E|\QTestHANA_DeleteUser\E$
=== RUN   TestHANA_Initialize
--- PASS: TestHANA_Initialize (0.01s)
=== RUN   TestHANA_NewUser
--- PASS: TestHANA_NewUser (0.08s)
=== RUN   TestHANA_UpdateUser
--- PASS: TestHANA_UpdateUser (0.24s)
=== RUN   TestHANA_DeleteUser
--- PASS: TestHANA_DeleteUser (1.09s)
PASS

Process finished with exit code 0
```


plugins/database/hana/hana.go
```diff
@@ -84,59 +93,55 @@ func (h *HANA) getConnection(ctx context.Context) (*sql.DB, error) {
...
+		return newdbplugin.NewUserResponse{}, err
 	}
 
-	if len(statements.Creation) == 0 {
-		return "", "", dbutil.ErrEmptyCreationStatement
+	if len(req.Statements.Commands) == 0 {
+		return newdbplugin.NewUserResponse{}, dbutil.ErrEmptyCreationStatement
 	}
 
+	dispName := credsutil.DisplayName(req.UsernameConfig.DisplayName, maxIdentifierLength)
```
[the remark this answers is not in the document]
Valarissa: Yep

plugins/database/hana/hana.go
```diff
@@ -43,27 +43,36 @@ func new() *HANA {
...
+func (h *HANA) secretValues() map[string]string {
+	return map[string]string{
+		h.Password: "[password]",
+	}
+}
+
+func (h *HANA) Initialize(ctx context.Context, req newdbplugin.InitializeRequest) (newdbplugin.InitializeResponse, error) {
+	conf, err := h.Init(ctx, req.Config, req.VerifyConnection)
+	if err != nil {
+		return newdbplugin.InitializeResponse{}, errwrap.Wrapf("error initializing db: %s", err)
```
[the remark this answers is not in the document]
Valarissa: Ack'd

plugins/database/hana/hana.go
```diff
@@ -3,38 +3,38 @@ package hana
...
 	"github.com/hashicorp/vault/sdk/helper/dbtxn"
 	"github.com/hashicorp/vault/sdk/helper/strutil"
+
+	_ "github.com/SAP/go-hdb/driver"
 )
 
 const (
-	hanaTypeName = "hdb"
+	hanaTypeName        = "hdb"
+	maxIdentifierLength = 127
```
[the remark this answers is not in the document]
Valarissa: `Maximum length for the identifiers is 127 characters.` from http://sap.optimieren.de/hana/hana/html/_bsql_introduction.html

plugins/database/hana/hana.go
```diff
@@ -148,148 +153,146 @@ func (h *HANA) CreateUser(ctx context.Context, statements dbplugin.Statements, u
...
 	defer tx.Rollback()
 
 	// If expiration is in the role SQL, HANA will deactivate the user when time is up,
 	// regardless of whether vault is alive to revoke lease
-	expirationStr, err := h.GenerateExpiration(expiration)
+	expirationStr := req.Expiration.NewExpiration.String()
 	if err != nil {
-		return err
+		return newdbplugin.UpdateUserResponse{}, err
 	}
```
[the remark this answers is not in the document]
Valarissa: It's usually handled, but was not covered before, so I was just doing the direct migration rather than adding things.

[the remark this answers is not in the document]
Valarissa: Should be fixed now.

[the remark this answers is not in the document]
Valarissa: Added password change support

plugins/database/hana/hana.go
```diff
@@ -148,148 +152,222 @@ func (h *HANA) CreateUser(ctx context.Context, statements dbplugin.Statements, u
...
 
 	return nil
 }
 
 // Revoking hana user will deactivate user and try to perform a soft drop
-func (h *HANA) RevokeUser(ctx context.Context, statements dbplugin.Statements, username string) error {
-	statements = dbutil.StatementCompatibilityHelper(statements)
+func (h *HANA) DeleteUser(ctx context.Context, req newdbplugin.DeleteUserRequest) (newdbplugin.DeleteUserResponse, error) {
+	h.Lock()
+	h.Unlock()
```
tvoran, line 292:
```suggestion
	defer h.Unlock()
```

Valarissa: :( I'm sad I missed that...

plugins/database/hana/hana.go
```diff
@@ -3,38 +3,37 @@ package hana
...
 )
 
 // HANA is an implementation of Database interface
 type HANA struct {
 	*connutil.SQLConnectionProducer
-	credsutil.CredentialsProducer
 }
 
-var _ dbplugin.Database = &HANA{}
+var _ newdbplugin.Database = &HANA{}
```
tvoran, line 30: This is now getting changed slightly in the other db plugins, so there's no unnecessary allocation:
```suggestion
var _ newdbplugin.Database = (*HANA)(nil)
```

tvoran (commented):
Sorry, this is getting caught in the changeover from newdbplugin --> dbplugin/v5.

The other thing I noticed is that updating a user's expiration isn't being tested, though I'm not sure how hard that would be to add.

plugins/database/hana/hana.go
```diff
@@ -148,148 +152,222 @@ func (h *HANA) CreateUser(ctx context.Context, statements dbplugin.Statements, u
...
 	}
 	defer tx.Rollback()
 
-	// If expiration is in the role SQL, HANA will deactivate the user when time is up,
-	// regardless of whether vault is alive to revoke lease
-	expirationStr, err := h.GenerateExpiration(expiration)
-	if err != nil {
-		return err
+	if req.Password != nil {
+		err = h.updateUserPassword(ctx, tx, req.Username, req.Password)
```
pcman312, line 198: Changing passwords and expirations will share a transaction. Is that intentional? I'm okay with it, I just want to make sure everyone else is good with it.

plugins/database/hana/hana.go
```diff
@@ -84,59 +92,55 @@ func (h *HANA) getConnection(ctx context.Context) (*sql.DB, error) {
...
 
-	// Generate password
-	password, err = h.GeneratePassword()
-	if err != nil {
-		return "", "", err
-	}
 	// Most HANA configurations have password constraints
 	// Prefix with A1a to satisfy these constraints. User will be forced to change upon login
+	password := req.Password
 	password = strings.Replace(password, "-", "_", -1)
```
pcman312, line 129: This needs to be removed. Manipulating the password in the plugin is not allowed because Vault won't know about those changes and will report the wrong password to the user.

I notice in the update password function, the default statement includes the password in double quotes. If the creation statement includes double quotes around the password, does this conversion become unnecessary?

plugins/database/hana/hana.go
```diff
@@ -148,148 +152,222 @@ func (h *HANA) CreateUser(ctx context.Context, statements dbplugin.Statements, u
...
+func (h *HANA) updateUserPassword(ctx context.Context, tx *sql.Tx, username string, req *newdbplugin.ChangePassword) error {
+	password := req.NewPassword
+
+	if username == "" || password == "" {
+		return fmt.Errorf("must provide both username and password")
+	}
+
+	stmts := req.Statements.Commands
+	if len(stmts) == 0 {
+		stmts = []string{"ALTER USER {{username}} PASSWORD \"{{password}}\""}
```
pcman312, line 228: nit if you want it:
```suggestion
		stmts = []string{`ALTER USER {{username}} PASSWORD "{{password}}"`}
```

plugins/database/hana/hana_test.go
```diff
@@ -22,8 +23,13 @@ func TestHANA_Initialize(t *testing.T) {
...
 	}
 
+	initReq := newdbplugin.InitializeRequest{
+		Config:           connectionDetails,
+		VerifyConnection: true,
+	}
+
 	db := new()
-	_, err := db.Init(context.Background(), connectionDetails, true)
+	_, err := db.Initialize(context.Background(), initReq)
```
pcman312, line 32: This can be improved by using `dbtesting.AssertInitialize` (from `sdk/database/dbplugin/v5/dbtesting`). That package has a number of functions that run the underlying DB functions with common behavior for the context (with a timeout rather than running indefinitely), and asserts the call didn't return an error.

pcman312, line 32: This should also make assertions on the response object to ensure that the config comes back to Vault properly.

plugins/database/hana/hana_test.go
```diff
@@ -49,38 +55,145 @@ func TestHANA_CreateUser(t *testing.T) {
...
 	if err == nil {
 		t.Fatal("Expected error when no creation statement is provided")
 	}
 
-	statements := dbplugin.Statements{
-		Creation: []string{testHANARole},
+	// Add a statement command
+	req.Statements.Commands = []string{testHANARole}
+
+	userResp, err = db.NewUser(context.Background(), req)
```
pcman312, line 96: Similarly, this can use `dbtesting.AssertNewUser` (and below in the other tests). Note that if you are expecting the error to be not nil then you can't use those assertion functions since they assert the call succeeded.

plugins/database/hana/hana_test.go
```diff
@@ -49,38 +55,145 @@ func TestHANA_CreateUser(t *testing.T) {
...
-	_, err := db.Init(context.Background(), connectionDetails, true)
+	_, err := db.Initialize(context.Background(), initReq)
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
 
-	usernameConfig := dbplugin.UsernameConfig{
-		DisplayName: "test-test",
-		RoleName:    "test-test",
+	password, err := credsutil.RandomAlphaNumeric(32, true)
```
pcman312, line 69: nit: The passwords in these tests don't need to be randomly sourced - they can be hard coded instead

plugins/database/hana/hana_test.go
```diff
@@ -49,38 +55,145 @@ func TestHANA_CreateUser(t *testing.T) {
...
+	// Test custom update statement
+	userResp, err = db.NewUser(context.Background(), newReq)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	if err = testCredsExist(t, connURL, userResp.Username, password); err != nil {
+		t.Fatalf("Could not connect with new credentials: %s", err)
+	}
+
+	updateReq.Password.Statements.Commands = []string{testHANAUpdate}
```
pcman312, line 185: There's a lot of shared state in these tests. Can these be made into table tests instead? MSSQL has some decent examples: https://github.com/hashicorp/vault/blob/master/plugins/database/mssql/mssql_test.go

plugins/database/hana/hana.go
```diff
@@ -84,59 +92,55 @@ func (h *HANA) getConnection(ctx context.Context) (*sql.DB, error) {
...
 	password = strings.Replace(password, "-", "_", -1)
-	password = "A1a" + password
 
 	// If expiration is in the role SQL, HANA will deactivate the user when time is up,
 	// regardless of whether vault is alive to revoke lease
-	expirationStr, err := h.GenerateExpiration(expiration)
-	if err != nil {
-		return "", "", err
-	}
+	expirationStr := req.Expiration.UTC().Format("2006-01-02 15:04:05")
```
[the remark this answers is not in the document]
Valarissa: I was running into issues with testing where it was bouncing back saying that the tests were failing because the expiration date had already passed.

plugins/database/hana/hana.go
Valarissa: I did, in fact, do that intentionally. The way I see it, if a user attempts to do two different actions and one fails, knowing the same request can be used is helpful. I'm not sure how others feel, but that's my preference with most interactions with an API.

tomhjp: +1, if I update two pieces of configuration in the same API call to Vault, I would find it least surprising as a user if they either both succeed or both fail.

plugins/database/hana/hana.go
```diff
@@ -148,148 +152,222 @@ func (h *HANA) CreateUser(ctx context.Context, statements dbplugin.Statements, u
...
 
-	return username, password, nil
+	return resp, nil
 }
 
 // Renewing hana user just means altering user's valid until property
-func (h *HANA) RenewUser(ctx context.Context, statements dbplugin.Statements, username string, expiration time.Time) error {
-	statements = dbutil.StatementCompatibilityHelper(statements)
+func (h *HANA) UpdateUser(ctx context.Context, req newdbplugin.UpdateUserRequest) (newdbplugin.UpdateUserResponse, error) {
+	h.Lock()
```
tomhjp, line 176: Asking because I have the same question for the Redshift DB migration: why do we need to hold a mutex lock inside the plugin? I can see that Vault needs to serialise password updates so that it knows which is the current password after concurrent rotate requests. However, the state is stored outside of this plugin, so the synchronisation would also need to be outside of the plugin i.e. around the state update in the caller. Am I missing another synchronisation requirement? cc @pcman312

pcman312: This was protecting against races when multiple calls are made to the engine at the same time. Specifically: rotate root credentials + any other call since rotate root was changing the plugin instance's configuration. I'm not 100% sure this is needed any longer, but I'd like to keep it for safety until we can confirm that it isn't needed.

plugins/database/hana/hana.go
pcman312: Awesome! That's the same logic I had in my head, I just wanted to make sure everyone was on the same page.

plugins/database/hana/hana_test.go
```diff
@@ -22,122 +22,231 @@ func TestHANA_Initialize(t *testing.T) {
...
-	if !db.Initialized {
-		t.Fatal("Database should be initialized")
-	}
-
-	err = db.Close()
-	if err != nil {
-		t.Fatalf("err: %s", err)
+	db := new()
+	initResp := dbtesting.AssertInitialize(t, db, initReq)
+	if initResp.Config == nil {
```
tvoran, line 32: FWIW, I've seen this style of checking the initialize response in a few places: https://github.com/hashicorp/vault/blob/a69ee0f65a28527376ab3d1cac89914a96406e30/plugins/database/mongodb/mongodb_test.go#L35-L48

plugins/database/hana/hana_test.go
```diff
@@ -22,122 +22,231 @@ func TestHANA_Initialize(t *testing.T) {
...
-		t.Fatalf("err: %s", err)
+	tests := map[string]testCase{
+		"no update statements": {
+			req: dbplugin.UpdateUserRequest{
+				Password: &dbplugin.ChangePassword{
+					NewPassword: "this_is_ALSO_Thirty_2_characters_",
+				},
+			},
+			startingPassword: "this_is_Thirty_2_characters_wow_",
+			expectErrOnLogin: true,
```
tvoran, line 130: I'm curious why this is expected to fail. I thought there was a default statement used if no update statements are included with a password change?

Valarissa: It's expected to fail because logging in after modifying the user prompts that user to change their password by default. (I lost the check that I was using to specifically check for that error message when I converted to the multi-test style.

plugins/database/hana/hana_test.go
Valarissa: Updated

plugins/database/hana/hana_test.go
Valarissa: Added the ability to check the error message received. We expect an error because the db wants the user to change their password on first log in after being modified.

tvoran: Oh interesting, thanks for the explanation!

tvoran (approved):

plugins/database/hana/hana_test.go
```diff
@@ -22,122 +23,240 @@ func TestHANA_Initialize(t *testing.T) {
...
+	initReq := dbplugin.InitializeRequest{
+		Config:           connectionDetails,
+		VerifyConnection: true,
 	}
 
-	err = db.Close()
-	if err != nil {
-		t.Fatalf("err: %s", err)
+	db := new()
+	initResp := dbtesting.AssertInitialize(t, db, initReq)
```
pcman312, line 34: nit: probably should include a `defer dbtesting.AssertClose(t, db)`

plugins/database/hana/hana_test.go
```diff
@@ -22,122 +23,240 @@ func TestHANA_Initialize(t *testing.T) {
...
+				Statements: dbplugin.Statements{
+					Commands: []string{testHANARole},
+				},
+				Expiration: time.Now().Add(time.Hour),
+			}
+
+			userResp := dbtesting.AssertNewUser(t, db, newReq)
+			assertCredsExist(t, connURL, userResp.Username, password)
+
+			test.req.Username = userResp.Username
```
pcman312, line 255: Instead of manipulating the test case, what do you think about making the test case have a list of deletion statements rather than the full request object? That way you can construct the request here without having it be a part of the test case.

pcman312 (approved):

pcman312 (approved):

PR #10144, opened 2020-10-14
Kubernetes Secret as physical storage backend

bonifaido:
This PR adds support for storing Vault physical data inside Kubernetes Secrets, so no PVC storage is needed on Kubernetes. It also allows for HA coordination with Kubernetes [Lease](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#lease-v1-coordination-k8s-io) type.

I'm pretty much open to comments, and I'm sure it is not 100% ready yet, but it is more or less feature complete. Tested with Kubernetes 1.18.


dustin-decker:
I don't think this is a viable storage backend because of the sheer number of entries that get created in Vault. Every identity, every lease, every static secret, configuration entries, etc become a key in the storage backend. Our Vault instance has hundreds of thousands of KV entries. This could quickly overwhelm the K8s API. K8s typically uses etcd under the hood for its KV storage, and that's already a supported physical storage backend.

aphorise:
Hey @bonifaido this request has been dormant for sometime and there's also been a few changes in K8s since.

What's more there's the response pending to [the earlier request](https://github.com/hashicorp/vault/pull/10144#issuecomment-738219033):
> Have you done some load testing with this?

Please update and provide additional context if you've use-cases where this is a better approach than other provided storage interface or why it may be needed.

If no longer relevant then please consider closing. Thanks.

bonifaido:
No, in the meantime I have abandoned this PR, by that time I needed a lot of very simply Vault instances, which could entirely run in Kubernetes, without the need of external Cloud services (Storage, KMS, etc...) and it worked really fine, but I don't have the bandwidth and focus to revive and load-test this now.

PR #10151, opened 2020-10-14
DBPW - Rename `newdbplugin` package to `dbplugin/v5`

pcman312:
# Package renaming - Part 1/4
This is the first in a 4 part series of changes to rename the `sdk/database/newdbplugin` package to `sdk/database/dbplugin/v5`.

1. Make copy of the `newdbplugin` package to `dbplugin/v5` so external/vendored DB plugins can reference it (and not break the current build)
2. Update external repos to the new package name
3. Vendor external repos into Vault
4. Remove the `newdbplugin` package
5. ???
6. Profit

## Overview
- Copies the `newdbplugin` package under `dbplugin` as `sdk/database/dbplugin/v5`.
- Moves the proto files under `sdk/database/dbplugin/v5/proto` so it doesn't keep multiple copies of the same proto (which causes registry collisions)
- Updates references in the DB engine to the `dbplugin/v5` package.

## Description
The `sdk/database/newdbplugin` package name was always intended to be a temporary name during development with the thought that it would replace the existing `sdk/database/dbplugin` package (with that package being renamed out of the way). That plan changed to moving `newdbplugin` under the `sdk/database/dbplugin` folder.

This keeps a copy of the original `newdbplugin` package (with one exception) in place because the package is being referenced in at least one vendored DB plugin (couchbase) causing build errors. Keeping it in place allows us to migrate the external/vendored DB plugins over to the new package seamlessly. The one exception is the `protos` sub-package. This was moved from `newdbplugin` to `dbplugin/v5` and a copy was not kept in the original `newdbplugin` package. This was done so the protobuf registry didn't see two copies of the same protos. All references to the proto have been updated to the new location.

## Testing
I tested this also against a set of bats scripts I have that are doing black box tests against Vault for: Cassandra, InfluxDB, MongoDB, MSSQL, PostgreSQL, and an external mock v4 database plugin.


plugins/database/cassandra/testoutput.json
```diff
@@ -0,0 +1,35 @@
...wr_1602523911 and/or password are incorrect\n2020/10/12 11:31:53 gocql: unable to dial control conn 127.0.0.1: Provided username v_test_test_co7ceqbuubvcnjdnatwr_1602523911 and/or password are incorrect\n2020/10/12 11:31:53 gocql: unable to dial control conn 127.0.0.1: Provided username v_test_test_co7ceqbuubvcnjdnatwr_1602523911 and/or password are incorrect\n2020/10/12 11:31:56 gocql: unable to dial control conn 127.0.0.1: EOF\n2020/10/12 11:31:57 gocql: unable to dial control conn 127.0.0.1: EOF\n2020/10/12 11:31:57 gocql: unable to dial control conn 127.0.0.1: EOF\n2020/10/12 11:31:58 gocql: unable to dial control conn 127.0.0.1: EOF\n2020/10/12 11:32:00 gocql: unable to dial control conn 127.0.0.1: Provided username cassandra and/or password are incorrect\n2020/10/12 11:32:02 gocql: unable to dial control conn 127.0.0.1: Provided username cassandra and/or password are incorrect\n2020/10/12 11:32:06 gocql: unable to dial control conn 127.0.0.1: Provided username cassandra and/or password are incorrect\n2020/10/12 11:32:12 gocql: unable to dial control conn 127.0.0.1: v_test_test_i1gwfjmtnfkxfaayoo3c_1602523931 is not permitted to log in\n","code":0,"duration":"1m31.286593268s"}
```
tvoran, line 1: This file is not like the others 😉

tvoran (approved):
Looks good, except for one extraneous file?

builtin/logical/database/path_roles.go
```diff
@@ -522,11 +522,11 @@ func (b *databaseBackend) pathStaticRoleCreateUpdate(ctx context.Context, req *l
 }
 
 type roleEntry struct {
-	DBName        string              `json:"db_name"`
-	Statements    dbplugin.Statements `json:"statements"`
-	DefaultTTL    time.Duration       `json:"default_ttl"`
-	MaxTTL        time.Duration       `json:"max_ttl"`
-	StaticAccount *staticAccount      `json:"static_account" mapstructure:"static_account"`
+	DBName        string         `json:"db_name"`
+	Statements    v4.Statements  `json:"statements"`
```
calvn, line 526: Not specific to this PR, but noticed that we're still using a v4 struct for storing the statements even though there are v5 equivalents of these. We'll probably have to handle upgrades if we wanted to fully migrate out of v4.

pcman312: We can certainly look into it, though I suspect we'll end up keeping the same or very similar struct definition for these since the list of statements hasn't changed and still needs to be stored.

calvn (approved):

tvoran (approved):

PR #10161, opened 2020-10-16
Updated missing code encoding around two path references

davidadeleon:
Found missing formatting on a couple path references in the Doc. Updated to include inline code formatting to maintain consistency with remainder of document.


raskchanky (approved):
Thanks!

PR #10171, opened 2020-10-19
UI/OIDC: allow passing namespace in state

tvoran:
Suppport in the UI OIDC callback flow to parse namespace out of the
state parameter instead of a separate query parameter in the
redirect_uri. Includes docs for the option that enables this behavior
in the JWT plugin.

The backend option is implemented in https://github.com/hashicorp/vault-plugin-auth-jwt/pull/140

To test this, I've been doing something like this along with the vault-plugin-auth-jwt PR https://github.com/hashicorp/vault-plugin-auth-jwt/pull/140:

  - `yarn start` in the vault/ui directory to start the UI in live-reloading/dev mode
  - `./scripts/local_dev.sh` in the vault-plugin-auth-jwt repo, with a vault enterprise binary in your path
  - visit the UI at http://localhost:4200/


calvn (approved):

ui/app/components/auth-jwt.js
```diff
@@ -128,6 +128,18 @@ export default Component.extend({
     let { namespace, path, state, code } = JSON.parse(oidcState);
     this.getWindow().localStorage.removeItem('oidcState');
 
+    // The namespace can be either be passed as a query paramter, or be embedded
+    // in the state param in the format `<state_id>,ns=<namespace>`. So if
+    // `namespace` is empty, check for namespace in state as well.
+    if (namespace == '') {
```
Monkeychip, line 134: should be triple ===

tvoran: Thanks, changed in 68afc00

chelshaw (commented):
Code-wise this looks good, but we're waiting until we can confirm that this works with local testing before approval

vishalnayak (approved):

chelshaw (approved):

PR #10178, opened 2020-10-19
Backport 1.5.x: Auth list fix by returning dynamicApiPath from the model (#10122)

Monkeychip:
[Backport PR for 10122](https://github.com/hashicorp/vault/pull/10122)

### DIFF
--- a/ui/app/services/path-help.js
+++ b/ui/app/services/path-help.js
@@ -23,6 +23,7 @@ export function sanitizePath(path) {
 
 export default Service.extend({
   attrs: null,
+  dynamicApiPath: '',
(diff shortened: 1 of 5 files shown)


ui/app/services/path-help.js
```diff
@@ -23,6 +23,7 @@ export function sanitizePath(path) {
 
 export default Service.extend({
   attrs: null,
+  dynamicApiPath: '',
```
chelshaw, line 26: We still don't need this, right?

chelshaw (approved):
Same comment as on the other one, but if fixing it here means they don't match then carry on!

PR #10192, opened 2020-10-20
Auto-Join: Configurable Scheme & Port (and add k8s provider)

alexanderbez:
- Allow operators to define a scheme and/or port in the `retry_join` stanza used for auto-join IP discovery. Vault defaults to HTTPS/8200 when these values are not provided.
- Introduce k8s provider (tested by @jasonodonnell)
- Update docs


vault/raft.go
```diff
@@ -940,6 +940,32 @@ func (c *Core) JoinRaftCluster(ctx context.Context, leaderInfos []*raft.LeaderJo
...
+					}
+
+					if u.Scheme == "" {
+						scheme := leaderInfo.Scheme
+						if scheme == "" {
+							// default to HTTPS when no scheme is provided
+							scheme = "https"
+						}
+
+						addr = fmt.Sprintf("%s://%s", scheme, addr)
```
ncabatoff, line 956: Why not just populate `u`'s members, then call `String()` on it?

alexanderbez: We can also do that :)

ncabatoff (approved):

physical/raft/raft.go
```diff
@@ -130,6 +130,13 @@ type LeaderJoinInfo struct {
 	// via 'leader_api_addr'.
 	AutoJoin string `json:"auto_join"`
 
+	// Scheme defines the optional URI protocol scheme for addresses discovered via
+	// auto-join.
+	Scheme string `json:"scheme"`
```
ncabatoff, line 135: Do we want to include AutoJoin in the name to make explicit what it's for?

alexanderbez: Yeah I was thinking we should.

physical/raft/raft.go
```diff
@@ -130,6 +130,14 @@ type LeaderJoinInfo struct {
 	// via 'leader_api_addr'.
 	AutoJoin string `json:"auto_join"`
 
+	// AutoJoinScheme defines the optional URI protocol scheme for addresses
+	// discovered via auto-join.
+	AutoJoinScheme string `json:"auto_join_scheme"`
```
ncabatoff, line 135: Do we want omitempty on this and the other AutoJoin fields?

alexanderbez: Depends. Is this configuration exposed in any requests or other client-facing calls? I'm suspicion is that it's not. @briankassouf?

briankassouf: Yeah, I don't actually think that we ever json marshal this struct, but you may want to confirm that.

vault/raft.go
```diff
@@ -1158,6 +1185,8 @@ func newDiscover() (*discover.Discover, error) {
 		providers[k] = v
 	}
 
+	providers["k8s"] = &discoverk8s.Provider{}
```
alexanderbez, line 1188: /cc @jasonodonnell 👍

physical/raft/raft.go
ncabatoff: I was thinking about the sanitized config output, though perhaps it doesn't marshall this directly.

ncabatoff (approved):

alexanderbez:
1.6.x backport => https://github.com/hashicorp/vault/pull/10229

PR #10195, opened 2020-10-20
DBPW - Migrate Redshift database plugin to v5 interface

tomhjp:
There's one behaviour change around what I believe is a bug. Previously if you left `rotation_statements` blank, it would attempt to use a default statement but always fail.

Acceptance tests output (post-review changes):

```bash
go test -v
=== RUN   TestRedshift_Initialize
--- PASS: TestRedshift_Initialize (0.57s)
=== RUN   TestRedshift_NewUser
--- PASS: TestRedshift_NewUser (3.73s)
=== RUN   TestRedshift_NewUser_NoCreationStatement_ShouldError
--- PASS: TestRedshift_NewUser_NoCreationStatement_ShouldError (0.55s)
=== RUN   TestRedshift_UpdateUser_Expiration
--- PASS: TestRedshift_UpdateUser_Expiration (12.91s)
=== RUN   TestRedshift_UpdateUser_Password
--- PASS: TestRedshift_UpdateUser_Password (4.56s)
=== RUN   TestRedshift_DeleteUser
--- PASS: TestRedshift_DeleteUser (8.64s)
PASS
ok  	github.com/hashicorp/vault/plugins/database/redshift	31.152s
```


plugins/database/redshift/redshift_test.go
```diff
@@ -331,79 +383,73 @@ func TestPostgresSQL_SetCredentials(t *testing.T) {
...
-		Password: password,
+		Password: &newdbplugin.ChangePassword{
+			NewPassword: password1,
+			Statements:  newdbplugin.Statements{},
+		},
 	}
 
 	// Test with no configured Rotation Statement
-	username, password, err := db.SetCredentials(context.Background(), dbplugin.Statements{}, usernameConfig)
-	if err == nil {
```
tomhjp, line 354: There is a behaviour change around this, but IMO this check was enforcing buggy behaviour. Both previously and now, the plugin attempted to use a default rotation statement when none is provided, but previously it used a template that had `{{username}}`, and tried to substitute in `{{name}}`. We substitute in both now, because there was inconsistency in the templated variable used between `RotateRootCredentials` and `SetCredentials`, and those got combined into one function.

So the upshot is now users can expect credential rotation to work by default when they don't provide a custom rotation statement, and the change from `err == nil` to `err != nil` here reflects that. WDYT?

plugins/database/redshift/redshift_test.go
```diff
@@ -421,15 +467,55 @@ func TestPostgreSQL_RotateRootCredentials(t *testing.T) {
...
+		t.Fatalf("Failed to test new admin user credentials after rotation: %s", err)
+	}
+
+	err = db.Close()
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
+	// Switch it back at the end to make the test somewhat less destructive.
+	// We re-initialize the whole plugin, which is roughly what the real database
```
tomhjp, line 492: Is this comment correct? This was my conclusion from taking a look through the backend, when I was trying to figure out how the plugin handled creating new connections after a root credential change.

plugins/database/redshift/redshift_test.go
```diff
@@ -331,79 +383,73 @@ func TestPostgresSQL_SetCredentials(t *testing.T) {
...
 		t.Fatalf("Could not connect with new credentials: %s", err)
 	}
 }
 
-func TestPostgreSQL_RotateRootCredentials(t *testing.T) {
-	/*
-		   Extra precaution is taken for rotating root creds because it's assumed that this
-		   test will run against a live redshift cluster. This test must run last because
-		   it is destructive.
+func TestRedshift_UpdateUser_RootPassword(t *testing.T) {
```
tomhjp, line 435: This test is arguably not valuable anymore, it's kind of a special case of the test above, `TestRedshift_UpdateUser_Password `.

plugins/database/redshift/redshift.go
```diff
@@ -229,132 +172,180 @@ func (r *RedShift) CreateUser(ctx context.Context, statements dbplugin.Statement
...
 	defer func() {
 		tx.Rollback()
 	}()
 
-	expirationStr, err := r.GenerateExpiration(expiration)
-	if err != nil {
-		return err
-	}
+	if req.Expiration != nil {
+		renewStmts := req.Expiration.Statements
```
pcman312, line 216: Check if the username is not empty?

plugins/database/redshift/redshift.go
```diff
@@ -229,132 +172,180 @@ func (r *RedShift) CreateUser(ctx context.Context, statements dbplugin.Statement
...
-				return err
+	if req.Password != nil {
+		username := req.Username
+		password := req.Password.NewPassword
+		if username == "" || password == "" {
+			return dbplugin.UpdateUserResponse{}, errors.New("must provide both username and password")
+		}
+
+		// Check if the role exists
+		var exists bool
```
pcman312, line 249: This is never referenced. Shouldn't we be checking if this is false and returning an error?

plugins/database/redshift/redshift.go
```diff
@@ -229,132 +172,180 @@ func (r *RedShift) CreateUser(ctx context.Context, statements dbplugin.Statement
...
+		err = db.QueryRowContext(ctx, "SELECT exists (SELECT usename FROM pg_user WHERE usename=$1);", username).Scan(&exists)
+		if err != nil && err != sql.ErrNoRows {
```
pcman312, line 251: Should the `ErrNoRows` check be here? If there are no rows returned, the user doesn't exist and we shouldn't attempt to manipulate it. I suppose we could keep it inside the err check as a way of giving a better error message.

pcman312: Though I suppose if we check the `exists` boolean this would become redundant anyway.

plugins/database/redshift/redshift.go
```diff
@@ -229,132 +172,180 @@ func (r *RedShift) CreateUser(ctx context.Context, statements dbplugin.Statement
...
+				}
+				if err := dbtxn.ExecuteTxQuery(ctx, tx, m, query); err != nil {
+					return dbplugin.UpdateUserResponse{}, err
+				}
 			}
 		}
 	}
 
-	return tx.Commit()
+	return dbplugin.UpdateUserResponse{}, tx.Commit()
```
pcman312, line 281: nit: I'm not a fan of putting calls during a return unless that's the entirety of the return statement. I find it makes it harder to follow.

plugins/database/redshift/redshift.go
```diff
@@ -229,132 +172,180 @@ func (r *RedShift) CreateUser(ctx context.Context, statements dbplugin.Statement
...
+	for _, stmt := range req.Statements.Commands {
 		for _, query := range strutil.ParseArbitraryStringSlice(stmt, ";") {
 			query = strings.TrimSpace(query)
 			if len(query) == 0 {
 				continue
 			}
 
 			m := map[string]string{
-				"name": username,
+				"name": req.Username,
```
pcman312, line 320: Should this also have `"username"` as an option? That way we'd be consistent.

plugins/database/redshift/redshift.go
```diff
@@ -229,132 +172,180 @@ func (r *RedShift) CreateUser(ctx context.Context, statements dbplugin.Statement
...
 	var exists bool
 	err = db.QueryRowContext(ctx, "SELECT exists (SELECT usename FROM pg_user WHERE usename=$1);", username).Scan(&exists)
 	if err != nil && err != sql.ErrNoRows {
-		return err
+		return dbplugin.DeleteUserResponse{}, err
 	}
 
 	if !exists {
-		return nil
+		return dbplugin.DeleteUserResponse{}, nil
```
pcman312, line 348: Can you put a comment on this saying that we don't return an error because redshift may have deleted the user via TTL before we got to it?

plugins/database/redshift/redshift_test.go
```diff
@@ -48,38 +47,40 @@ var (
 	vaultACC = "VAULT_ACC"
 )
 
-func redshiftEnv() (url string, user string, password string, errEmpty error) {
+func interpolateConnectionURL(url, user, password string) string {
+	return fmt.Sprintf("postgres://%s:%s@%s", user, password, url)
+}
+
+func redshiftEnv() (connURL string, url string, user string, password string, errEmpty error) {
 	errEmpty = errors.New("err: empty but required env value")
```
pcman312, line 55: Ew. This doesn't tell the user what env variables they are missing.

plugins/database/redshift/redshift_test.go
```diff
@@ -48,38 +47,40 @@ var (
...
 
 	if password = os.Getenv(keyRedshiftPassword); url == "" {
-		return "", "", "", errEmpty
+		return "", "", "", "", errEmpty
 	}
 
-	url = fmt.Sprintf("postgres://%s:%s@%s", user, password, url)
-
-	return url, user, password, nil
+	return interpolateConnectionURL(url, user, password), url, user, password, nil
```
pcman312, line 69: nit: can `interpolateConnectionURL` be split out into its own call? This line is hard to read

plugins/database/redshift/redshift_test.go
```diff
@@ -100,29 +101,31 @@ func TestPostgreSQL_Initialize(t *testing.T) {
 
 	// Test decoding a string value for max_open_connections
 	connectionDetails = map[string]interface{}{
-		"connection_url":       url,
-		"max_open_connections": "5",
+		"connection_url":       connURL,
+		"max_open_connections": "73",
 	}
 
 	_, err = db.Init(context.Background(), connectionDetails, true)
```
pcman312, line 108: Also check the returned config to make sure Vault saves it

pcman312, line 108: Shouldn't this be calling `Initialize` to test the `Database` interface? If so, you can use the `dbtesting` package under `github.com/hashicorp/vault/sdk/database/dbplugin/v5/dbtesting` to make this call (and other calls to the DB) smoother

plugins/database/redshift/redshift_test.go
```diff
@@ -131,32 +134,46 @@ func TestPostgreSQL_CreateUser(t *testing.T) {
...
 
-	usernameConfig := dbplugin.UsernameConfig{
+	usernameConfig := dbplugin.UsernameMetadata{
 		DisplayName: "test",
 		RoleName:    "test",
 	}
 
 	// Test with no configured Creation Statement
-	_, _, err = db.CreateUser(context.Background(), dbplugin.Statements{}, usernameConfig, time.Now().Add(time.Minute))
+	_, err = db.NewUser(context.Background(), dbplugin.NewUserRequest{})
```
pcman312, line 143: I recommend also checking the returned username against a regular expression to make sure it's formed as expected. Here's an example of doing that in the MSSQL tests: https://github.com/hashicorp/vault/blob/master/plugins/database/mssql/mssql_test.go#L98

plugins/database/redshift/redshift_test.go
```diff
@@ -131,32 +134,46 @@ func TestPostgreSQL_CreateUser(t *testing.T) {
...
 	if err == nil {
 		t.Fatal("Expected error when no creation statement is provided")
 	}
 
-	statements := dbplugin.Statements{
-		Creation: []string{testRedshiftRole},
+	const password = "SuperSecurePa55w0rd!"
+	req := dbplugin.NewUserRequest{
```
pcman312, line 149: This should be split out into its own test. Can this be refactored into a table test?

plugins/database/redshift/redshift_test.go
```diff
@@ -131,32 +134,46 @@ func TestPostgreSQL_CreateUser(t *testing.T) {
...
+		UsernameConfig: usernameConfig,
+		Password:       password,
+		Statements: dbplugin.Statements{
+			Commands: []string{testRedshiftRole},
+		},
+		Expiration: time.Now().Add(5 * time.Minute),
 	}
 
-	username, password, err := db.CreateUser(context.Background(), statements, usernameConfig, time.Now().Add(time.Minute))
+	resp, err := db.NewUser(context.Background(), req)
```
pcman312, line 158: These calls should use the `dbtesting` package when the error is expected to be nil so it uses a common set of behaviors such as creating a context with timeout.

plugins/database/redshift/redshift_test.go
```diff
@@ -189,71 +206,103 @@ func TestPostgreSQL_RenewUser(t *testing.T) {
...
 		RoleName:    "test",
 	}
 
-	username, password, err := db.CreateUser(context.Background(), statements, usernameConfig, time.Now().Add(2*time.Second))
+	const password = "SuperSecurePa55w0rd!"
+	newReq := dbplugin.NewUserRequest{
+		UsernameConfig: usernameConfig,
+		Password:       password,
+		Statements:     dbplugin.Statements{Commands: []string{testRedshiftRole}},
+		Expiration:     time.Now().Add(5 * time.Second),
```
pcman312, line 219: I recommend pulling the expiration duration out to a variable so it can be referenced below.

pcman312, line 219: Can this expiration be smaller since we're doing a sleep with it later?

plugins/database/redshift/redshift_test.go
```diff
@@ -189,71 +206,103 @@ func TestPostgreSQL_RenewUser(t *testing.T) {
...
+	}
+
+	_, err = db.UpdateUser(context.Background(), updateReq)
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
 
 	// Sleep longer than the initial expiration time
-	time.Sleep(2 * time.Second)
+	time.Sleep(5 * time.Second)
```
pcman312, line 246: Shouldn't this be longer than 5s?

plugins/database/redshift/redshift_test.go
```diff
@@ -290,18 +346,25 @@ func TestPostgreSQL_RevokeUser(t *testing.T) {
 		t.Fatal("Credentials were not revoked")
 	}
 
-	username, password, err = db.CreateUser(context.Background(), statements, usernameConfig, time.Now().Add(2*time.Second))
+	newRequest.Expiration = time.Now().Add(2 * time.Second)
```
pcman312, line 349: This is sharing state between test cases. Can this be refactored into a table test instead?

plugins/database/redshift/redshift_test.go
pcman312: :+1:

plugins/database/redshift/redshift_test.go
pcman312: I agree - the DB instance doesn't care about which user its updating and doesn't make any changes to its state as a part of this change. I think we can get rid of this test.

plugins/database/redshift/redshift.go
```diff
@@ -229,132 +172,180 @@ func (r *RedShift) CreateUser(ctx context.Context, statements dbplugin.Statement
...
+
+				m := map[string]string{
+					"name":       req.Username,
+					"expiration": expirationStr,
+				}
+				if err := dbtxn.ExecuteTxQuery(ctx, tx, m, query); err != nil {
+					return dbplugin.UpdateUserResponse{}, err
+				}
 			}
+		}
```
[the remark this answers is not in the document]
Valarissa: +1

plugins/database/redshift/redshift.go
```diff
@@ -32,18 +31,21 @@ const (
...
 
-// lowercaseUsername is the reason we wrote this plugin. Redshift implements (mostly)
+var _ dbplugin.Database = (*RedShift)(nil)
+
+// New implements builtinplugins.BuiltinFactory
+// lowerCaseUsername is the reason we wrote this plugin. Redshift implements (mostly)
 // a postgres 8 interface, and part of that is under the hood, it's lowercasing the
 // usernames.
-func New(lowercaseUsername bool) func() (interface{}, error) {
+func New(lowerCaseUsername bool) func() (interface{}, error) {
```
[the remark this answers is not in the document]
tomhjp: Good call, removed.

plugins/database/redshift/redshift.go
```diff
@@ -99,124 +112,54 @@ func (r *RedShift) getConnection(ctx context.Context) (*sql.DB, error) {
...
-		return "", "", err
-	}
-
-	expirationStr, err := r.GenerateExpiration(expiration)
-	if err != nil {
-		return "", "", err
+		return dbplugin.NewUserResponse{}, err
 	}
+	password := req.Password
+	expirationStr := req.Expiration.UTC().Format("2006-01-02 15:04:05")
```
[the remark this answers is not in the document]
tomhjp: In my experience, postgres has been smart enough to handle different timezones correctly by default, so I didn't think this would be a behaviour change, but it's definitely an unnecessary risk to hope that that holds across all deployments of redshift. I've deleted all occurrences.

plugins/database/redshift/redshift.go
tomhjp: I've updated this to split error handling broadly between server and (most likely) user error. The latter has a specific user-friendly error string which completely encapsulates the context of the error value, so I've taken the unusual step of not wrapping it. Let me know what you think.

plugins/database/redshift/redshift.go
tomhjp: `username` was only used in the default root rotation statement, and wasn't documented anywhere as a supported template. However, I do agree consistency within the `UpdateUser` function at least would be nice. Do you think we should spread the consistency further to include `{{username}}` for `NewUser` and `DeleteUser`?

plugins/database/redshift/redshift_test.go
tomhjp: Fixed.

plugins/database/redshift/redshift_test.go
tomhjp: It was 2 seconds before, but I was worried that that felt quite short and could lead to flakiness. Happy to revert it back to 2s if you think that is a more reasonable balance of risk.

plugins/database/redshift/redshift_test.go
tomhjp: The one place where I have not done this was for the `DeleteUser` test, where I found that 2s was always too short a timeout. I'd prefer not to have to set environment variables to override the default for the whole test suite, so just decided to avoid the helper in that one instance instead.

plugins/database/redshift/redshift_test.go
tomhjp: Oops, I thought I got rid of all this kind of state sharing. Thanks for catching.

plugins/database/redshift/redshift_test.go
tomhjp: Thanks - TIL about the `dbtesting` package.

tomhjp:
Thanks for the thorough review @pcman312 - ready for another round of review now.

plugins/database/redshift/redshift.go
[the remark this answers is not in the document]
pcman312: The problem is 1) this is changing the behavior and 2) the output format doesn't include the timezone

plugins/database/redshift/redshift.go
pcman312: I like it :+1:

plugins/database/redshift/redshift.go
pcman312: At this point I think we should. It's been a mix of `user` and `username` across databases (and even across specific DB calls) so I think just making it available as both everywhere is the right call.

plugins/database/redshift/redshift_test.go
```diff
@@ -48,281 +45,226 @@ var (
...
+	})
 
 	if !db.Initialized {
 		t.Fatal("Database should be initialized")
 	}
-
-	err = db.Close()
-	if err != nil {
-		t.Fatalf("err: %s", err)
+	if resp.Config["max_open_connections"] != 73 {
```
pcman312, line 93: Instead of checking a single field in the returned config, I recommend making an `expectedConfig` object that is a copy/duplicate of the `connectionDetails` map above (a copy/duplicate since the `Initialize` call could manipulate the map without us realizing it) and then doing a `if !reflect.DeepEquals` on it

plugins/database/redshift/redshift_test.go
pcman312: That's completely reasonable.

plugins/database/redshift/redshift_test.go
pcman312: Since this is hitting an actual redshift instance in AWS I think it's reasonable to have this larger. I didn't realize this wasn't hitting a local service previously.

pcman312 (approved):
LGTM with one small test improvement

plugins/database/redshift/redshift.go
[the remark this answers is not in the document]
tomhjp: Oops I misunderstood your comment before. I think it's fixed properly now, thanks for spotting.

plugins/database/redshift/redshift_test.go
tomhjp: To be fair, my increase to the timings wasn't based on any data. Even with the Redshift DB based in `us-east-1`, the tests still reliably pass from my machine as they are now.

PR #10196, opened 2020-10-20
Ui/pricing metrics api hookup

chelshaw:
This work cleans up the UI against the latest changes on the pricing metrics API.

- Wrap date inputs in a form so it queries when "enter" is pressed.
- Failsafe against 204 no data response on activity
- Updating language to reflect that the numbers are the namespace and all its children
- Pass default span and retention period to pricing metrics component from config so the inputs are correct on init


Monkeychip (approved):
This looks great.  Nice work!

PR #10201, opened 2020-10-20
Port: Add metrics to report mount table sizes for auth and logical [Vault 671]

HridoyRoy:
Enterprise PR: https://github.com/hashicorp/vault-enterprise/pull/1500


mgritter (approved):

vault/auth_test.go
```diff
@@ -323,7 +405,7 @@ func TestCore_DisableCredential_Cleanup(t *testing.T) {
 		Login:       []string{"login"},
 		BackendType: logical.TypeCredential,
 	}
-	c, _, _ := TestCoreUnsealed(t)
+	c, _, _, _ := TestCoreUnsealedWithMetrics(t)
```
mgritter, line 408: You don't have to fix it in this backport, but I will note that all these changes to the test method could be undone now.  :)

HridoyRoy: Absolutely! I can make a separate PR changing those to TestCoreUnsealed.

PR #10204, opened 2020-10-21
secrets/database: Fix test for autoMTLS

briankassouf:
With this test change the old SHA with AutoMTLS passes. I think we should re-enable that


pcman312:
This looks good when I test it locally, however this PR isn't updated against master which has removed the `AutoMTLS` option (originally anticipating we wouldn't get it in). There are a number of other things in flight that need to get merged first, but this goes on the top of the list once those are completed (hopefully today/tomorrow).

calvn:
@briankassouf @pcman312 I believe this can be closed now.

pcman312:
@calvn I agree - this change was incorporated into https://github.com/hashicorp/vault/pull/10220

PR #10217, opened 2020-10-22
UI: Update replication secondaries language

chelshaw:
Update language on primary to reflect how many known secondaries there are, and update CTA for viewing all secondaries.

replication-secondaries


Monkeychip (approved):
Love these quick kind of fixes.

PR #10218, opened 2020-10-22
UI: Fix error on OIDC/JWT auth enable or config

chelshaw:
A bug was discovered where attempting to enable or view the config for an OIDC or JWT auth method, the UI froze up and the console gave the error `Non primitive defaultValues are not supported...`. The root cause turned out to be a default value set to an object coming from OpenAPI which is resolved in a different PR for that plugin, but to prevent the UI blowing up in a similar way in the future we will return any object default values as a function that returns that value.


Monkeychip (approved):
Thanks for working on this!

PR #10248, opened 2020-10-26
OCI: Don't store region in the backend struct

vishalnayak:
Follow up of https://github.com/hashicorp/vault/pull/9302


physical/oci/oci.go
```diff
@@ -73,7 +72,6 @@ type Backend struct {
 }
 
 func NewBackend(conf map[string]string, logger log.Logger) (physical.Backend, error) {
-	region     := conf["region"]
 	bucketName := conf["bucket_name"]
```
calvn, line 76:
```suggestion
```

calvn (approved):
Looks good!

PR #10251, opened 2020-10-27
Update radius.mdx

martinhristov90:
I think the `-method=radius` parameter should be included as well


aphorise:
Hey Don @martinhristov90 - I believe this has since [been resolved with a contextual note around the `-path` parameter](https://www.vaultproject.io/docs/auth/radius#via-the-cli). Please confirm & close if no longer applicable

PR #10266, opened 2020-10-29
Updating to jwt plugin@master

tvoran:
Pulls in https://github.com/hashicorp/vault-plugin-auth-jwt/pull/137 and https://github.com/hashicorp/vault-plugin-auth-jwt/pull/140


austingebauer (approved):
LGTM

PR #10276, opened 2020-10-30
add scripts/upgrade-packagespec

samsalisbury:
Please see the documentation comments at the head of the script for info on how this works.


scripts/upgrade-packagespec
```diff
@@ -0,0 +1,110 @@
...
+$COMMAND
+
+git add .circleci/ packages*.lock packagespec*
+git commit -m "$COMMAND"
+
+git push -u "$TARGET_REMOTE" "$NEW_BRANCH"
+
+echo "==> All done: upgrade pushed to branch $NEW_BRANCH on ${REMOTES[$TARGET_REMOTE]}"
+
+echo "==> ACTIONS FOR YOU: Open a PR with base: $BRANCH compare: $NEW_BRANCH"
```
mladlow, line 110: I feel like maybe this should delete the temporary directory, but probably it should `cd ..`?

mladlow: I see the `cd $CLONEDIR` but nothing to put you back after all the git commands have run?

samsalisbury: Ah, I understand the confusion. This script is run by invoking the file `./scripts/upgrade-packagespec <args>` hence it runs in its own shell, and when it exits you are back where you started. The similar merge script also runs in its own shell, but it starts an interactive subshell inside a different directory to hand control to the user there explicitly (but even there, after you exit the subshell you are back where you started). The only way for a script to change the directory of your shell is by `source`ing a file that does that.

mdeggies:
Is this one still needed @samsalisbury? Looks like tests are failing. 
Sorry I'm just seeing the requested review now!

scripts/upgrade-packagespec
```diff
@@ -0,0 +1,110 @@
...
+	echo "usage: $0 <packagespec version> <branch name>"
+	exit 1
+fi
+
+set -euo pipefail
+
+declare -A REMOTES
+
+# url_remotes lists remotes along with their push URLs, where they are
+# filesystem paths (i.e. do not start with / or . ).
```
ncabatoff, line 43:
```suggestion
# not filesystem paths (i.e. do not start with / or . ).
```

ncabatoff (approved):

Issue #10067, opened 2020-10-01
Selecting any algorithm_signer for a ssh engine causes sign error


lhw:
**Describe the bug**

**To Reproduce**
```
1. vault secrets enable ssh
2. ssh-keygen -C ca -t ed25519 -f ca -N '' 
3. vault write ssh/config/ca private_key=@ca public_key=@ca.pub
4. vault write ssh/roles/role allow_user_certificates="true" key_type="ca"
5. ssh-keygen -C user -f user -N ''
6. vault write ssh/sign/role public_key=@user.pub -> SUCCESS
7. vault write ssh/roles/role allow_user_certificates="true" key_type="ca" algorithm_signer="ssh-rsa"
8. vault write ssh/sign/role public_key=@user.pub -> FAILURE

Code: 500. Errors:
* 1 error occurred:
        * failed to generate signed SSH key: sign error
```
(I wonder how often the user ca is quoted in this kind of context)

**Expected behavior**
Selecting an, as per documentation, secure signing algorithm should not cause a signing error.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.5.4
* Vault CLI Version (retrieve with `vault version`): 1.5.4 (1a730771ec70149293efe91e1d283b10d255c6d1)
* Server Operating System/Architecture: Ubuntu 20.04.1

Vault server configuration file(s):
`vault server -dev`

**Additional context**

pippo:
Not sure whether that's expected behaviour or not, however, it looks like this is failing because you're trying to use RSA algo while using ed25519 key for it.
Underlying error, coming from `golang.org/x/crypto/ssh` ([here](https://github.com/golang/crypto/blob/master/ssh/keys.go#L988)) is:
```
ssh: unsupported signature algorithm ssh-rsa
```
The issue does not reproduce with `ed25519` changed to `rsa` at step 2 above.

Perhaps, the error could be a bit more informative, if the underlying error was appended to it. E.g.
```
failed to generate signed SSH key: sign error: ssh: unsupported signature algorithm ssh-rsa
```

lhw:
ed25519 keys can sign rsa keys and the other way around. But the signer needs to select the correct signature algorithm. This probably means the #9824 was not expansive enough.

lhw:
I wanted to bump this issue as clients on our side had to downgrade back to rsa for now after I got everyone to update to ed25519 keys.

candlerb:
[...]

PR #10301, opened 2020-11-01
Make signature error more informative

pippo:
The brief "sign error" error message might cause confusion, which, for example, resulted in #10067 being filed.
Changes in this PR wrap an error coming from `golang.org/x/crypto/ssh` to make the error message a bit more informative, in order to simplify troubleshooting by an end-user.


swayne275 (approved):
Thanks!

PR #10322, opened 2020-11-04
Update query params sent to activity endpoint

chelshaw:
This change fixes an issue where a timestamp sent for the end of the month could be read as the next month and then rounded, creating an offset on results:
image (1)

Now, the query params sent take advantage of the internal logic of the API endpoint, which automatically rounds the start date to the beginning of the month and the end date to the end of the month. 
fixed-end-date-param


Monkeychip (approved):
LGTM 👍

PR #10332, opened 2020-11-05
Add more workdirs to the .gitignore list to keep GoLandIDE from retur…

ncabatoff:
…ning files within as search results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,5 @@ website/components/node_modules
 
 .buildcache/
 .releaser/
+.tmp/
+.packagespec/


samsalisbury (approved):
So, does GoLand only look a the root `.gitignore`? Those directories should always contain a `.gitignore` containing exactly `*`... Nothing against adding them here as well if it helps though.

raskchanky (approved):

kalafut (approved):

ncabatoff:
Actually, I should've tested this better.  It didn't actually solve my problem of keeping the code under .packagespec from showing up in the index.  Instead I went under preferences -> Project Structure and marked those folders as Excluded, which seems to have done the trick.

PR #10341, opened 2020-11-06
Ui/test metrics mirage

chelshaw:
This PR adds the Mirage dependency and uses it for testing the pricing metrics UI. 

Up until now, all of our front end tests take advantage of real calls to vault APIs, because none of the data is expected to just live on the server. However, with this new endpoint it does expect data to accumulate over time, which is unreasonable to wait for testing. 

While writing these tests I did find a legitimate bug where if the data is queried but doesn't have any data, "Invalid date" is shown:
before-no-data-for-period

Fixed:
no-data-for-period

### DIFF
--- a/ui/app/templates/vault/cluster/metrics/index.hbs
+++ b/ui/app/templates/vault/cluster/metrics/index.hbs
@@ -6,7 +6,7 @@
   </p.levelLeft>
 </PageHeader>
 
-<div class="tabs-container box is-bottomless is-marginless is-fullwidth is-paddingless">
+<div class="tabs-container box is-bottomless is-marginless is-fullwidth is-paddingless" data-test-pricing-metrics>
   <nav class="tabs">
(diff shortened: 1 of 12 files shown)


ui/app/templates/vault/cluster/metrics/index.hbs
```diff
@@ -51,6 +51,7 @@
   <div class="box is-sideless is-fullwidth is-marginless is-bottomless">
     {{#if (eq model.config.enabled 'Off')}}
       <AlertBanner
+      data-test-tracking-disabled
```
Monkeychip, line 54: nit: indent here for data-test-tracking-disabled ?

Monkeychip (commented):
Nice work!  There's a failing UI test, but other than that looks great!

Monkeychip (approved):

PR #9819, opened 2020-08-25
agent: support providing certificate information in cert's config map

calvn:
Applies the following changes into master:
- Patches f3c4426168a423dbe815b7ecb481ce0948fa090c back to master.
- Patches 2bed51e3f503274b8d12805500fe083105e3b7b7 back to master.

PR #10348, opened 2020-11-08
agent/auth/cert: fix reauthentication when certs specified in config

sorah:
Vault agent is failing to reauthenticate with cert auth method, when the agent auto_auth method is configured with a client certificate.

It has been broken since https://github.com/hashicorp/vault/pull/9819, Vault v1.5.1.

Adding E2E test to avoid the same bug in future.


sorah:
test failures seem unrelated to my change

sorah:
@calvn could you take a look at this please?

cipherboy:
This was fixed in #11576; closing.

PR #10353, opened 2020-11-09
docs: adds documentation for key management secrets engine

austingebauer:
This PR adds documentation for the Key Management secrets engine.

Deploy Previews:
- [User docs](https://deploy-preview-10353--vault-www.netlify.app/docs/secrets/key-management)
- [API docs](https://deploy-preview-10353--vault-www.netlify.app/api-docs/secret/key-management)


website/pages/api-docs/secret/key-management/index.mdx
```diff
@@ -0,0 +1,540 @@
...
+    --header "X-Vault-Token: ..." \
+    --request GET \
+    http://127.0.0.1:8200/v1/keymgmt/kms/example-kms
+```
+
+### Sample Response
+
+```json
+{
+  "data": {
```
catsby, line 341: Does it make sense to return the `tenant_id` and `environment` here? Assuming it's possible to manage multiple Azure KMS with different values

website/pages/api-docs/secret/key-management/index.mdx
```diff
@@ -0,0 +1,540 @@
...
+    http://127.0.0.1:8200/v1/keymgmt/kms
+```
+
+### Sample Response
+
+```json
+{
+  "data": {
+    "keys": ["example-kms"]
+  }
```
catsby, line 371: IS this the correct output for listing configured KMS providers?

website/pages/api-docs/secret/key-management/index.mdx
```diff
@@ -0,0 +1,540 @@
...
+```
+
+## List Keys in KMS Provider
+
+This endpoint returns a list of all keys that have been distributed to the given KMS
+provider. Many keys can be distributed to a single KMS provider.
+
+| Method | Path                     |
+| :----- | :----------------------- |
+| `LIST` | `/keymgmt/kms/:name/key` |
```
catsby, line 488: nit: I feel like this should be `:name/keys` but that may be a bigger discussion to be had internally

catsby (changes requested):
Overall looks great! I left some nitpick cleanup things that I'd like to see before merging but they are all pretty minor I think

website/pages/docs/secrets/key-management/index.mdx
```diff
@@ -0,0 +1,189 @@
...
+
+# Key Management Secrets Engine
+
+-> **Note**: This secrets engine requires [Vault
+Enterprise](https://www.hashicorp.com/products/vault/) with the Advanced Data
+Protection Module.
+
+The Key Management secrets engine provides a consistent workflow for distribution and lifecycle
+management of cryptographic keys in various key management service (KMS) providers. It allows
+organizations to maintain centralized control of their keys from Vault while still taking advantage
```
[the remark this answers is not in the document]
austingebauer: I intended for it to mean "by using". I'm wondering if there is a better word that would make it less ambiguous.

`from Vault` -> `in Vault`? 🤔

website/pages/docs/secrets/key-management/index.mdx
```diff
@@ -0,0 +1,189 @@
...
+1. Remove a key from a KMS provider:
+
+    ```text
+    $ vault delete keymgmt/kms/example-kms/key/example-key
+    ```
+
+    This operation results in the key being deleted from the KMS provider. The key will still exist
+    in the secrets engine and can be redistributed to a KMS provider at a later time.
+
+    To permanently delete the key from the secrets engine, the [delete key](/api/secret/key-management#delete-key)
```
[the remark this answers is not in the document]
calvn: Since it's a `DELETE` operation, I believe this can be done via `vault delete ...`

website/pages/api-docs/secret/key-management/index.mdx
calvn: I believe this is correct since the names are unique and there could be multiple `kms`'es for each provider (i.e. there could be more than one `azurekeyvault` kms set up).

website/pages/api-docs/secret/key-management/index.mdx
```diff
@@ -0,0 +1,540 @@
...
+        "public_key": "-----BEGIN PUBLIC KEY----- ... -----END PUBLIC KEY-----",
+        "type": "rsa-2048"
+      },
+      "2": {
+        "creation_time": "2020-11-04T16:58:47.591718-08:00",
+        "public_key": "-----BEGIN PUBLIC KEY----- ... -----END PUBLIC KEY-----",
+        "type": "rsa-2048"
+      }
+    },
+    "latest_version": 1,
```
calvn, line 97: Should latest version be 2?

website/pages/docs/secrets/key-management/index.mdx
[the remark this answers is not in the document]
austingebauer: This can be done using the CLI. I didn't add a step to the usage section for this because it's a simple operation that doesn't need much conceptual explaining. I thought it was pretty common knowledge that the CLI is mostly just an API client, so I refer to the API here. Happy to add another step with CLI usage if you think that's better.

website/pages/docs/secrets/key-management/index.mdx
```diff
@@ -0,0 +1,189 @@
...
+
+## Key Types
+
+The Key Management secrets engine currently supports generation of the following key types:
+
+- `rsa-2048` - RSA with bit size of 2048 (asymmetric)
+- `rsa-3072` - RSA with bit size of 3072 (asymmetric)
+- `rsa-4096` - RSA with bit size of 4096 (asymmetric)
+
+## KMS Providers
```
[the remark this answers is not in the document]
austingebauer: Indeed! I was thinking the same. Once the next provider is added, we'll break these out into sub-sections as you've mentioned 👍

website/pages/api-docs/secret/key-management/index.mdx
austingebauer: Good find! Fixed in 05359b6.

website/pages/api-docs/secret/key-management/index.mdx
```diff
@@ -0,0 +1,540 @@
...
+
+## Update Key
+
+This endpoint updates a named key.
+
+| Method | Path                 |
+| :----- | :------------------- |
+| `PUT`  | `/keymgmt/key/:name` |
+
+### Parameters
```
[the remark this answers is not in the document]
austingebauer: These currently can't be used during creation. A key only has a single version at creation time, so there isn't really a use for specifying a `min_enabled_version` at that point.

We'd like to allow specifying `deletion_allowed` at creation time eventually. Transit is similar in that it can only be specified for [key updates](https://www.vaultproject.io/api-docs/secret/transit#deletion_allowed).

website/pages/docs/secrets/key-management/index.mdx
```diff
@@ -0,0 +1,189 @@
...
+    Keys created by the secrets engine are considered general-purpose until
+    they're distributed to a KMS provider.
+
+1. Configure a KMS provider:
+
+    ```text
+    $ vault write keymgmt/kms/example-kms \
+        provider="azurekeyvault" \
+        key_collection="keyvault-name" \
+        credentials=client_id="" \
```
calvn, line 67: Should we provide dummy values for credentials?

website/pages/docs/secrets/key-management/index.mdx
```diff
@@ -0,0 +1,189 @@
...
+
+#### Key Transfer Specification
+
+Keys are securely transferred from the secrets engine to Azure key vault instances in accordance
+with the [Azure Bring Your Own Key](https://docs.microsoft.com/en-us/azure/key-vault/keys/byok-specification)
+specification.
+
+## Learn
+
+Refer to the [Key Management Secrets Engine](https://learn.hashicorp.com/vault/adp/TODO) guide for
```
calvn, line 182: This link seems invalid. @yhyakuna might be able to provide you with the actual link to the learn guide.

website/pages/api-docs/secret/key-management/index.mdx
```diff
@@ -0,0 +1,540 @@
...
+  This is provided as part of the request URL.
+
+- `provider` `(string: <required>)` – Specifies the name of a KMS provider that's external to Vault.
+  Cannot be changed after creation. The following values are supported:
+  - `azurekeyvault` - KMS provider represents a named
+  [Azure Key Vault](https://docs.microsoft.com/en-us/azure/key-vault/) instance.
+
+- `key_collection` `(string: <required>)` – Refers to a location to store keys in the specified
+  provider. Cannot be changed after creation. The expected value for this parameter will differ
+  depending on the specified provider. The following values are expected for each provider:
```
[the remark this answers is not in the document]
calvn: Oh actually its not quite the same.

`- azurekeyvault - KMS provider represents a named [Azure Key Vault]`
vs 
`- azurekeyvault - The name of an existing Azure Key Vault instance.`

Having said this, I think it might be a bit confusing since the first case is implying that `azurekeyvault` is one of the valid values to be provided, but for the second case the value for `key_collection` is some the actual instance _within_ Azure Key Vault, and not the value `azurekeyvault` itself.

website/pages/api-docs/secret/key-management/index.mdx
```diff
@@ -0,0 +1,540 @@
...
+| :----- | :-------------------- |
+| `PUT`  | `/keymgmt/kms/:name`  |
+
+### Parameters
+
+- `name` `(string: <required>)` – Specifies the name of the KMS provider to create or update.
+  This is provided as part of the request URL.
+
+- `provider` `(string: <required>)` – Specifies the name of a KMS provider that's external to Vault.
+  Cannot be changed after creation. The following values are supported:
```
calvn, line 265: Maybe something like this instead?
```
Currently, `azurekeyvault` is supported. For more information about each provider, refer to the [KMS Providers](/docs/secrets/key-management#kms-providers) section.
```

austingebauer: I like that better 👍 Done in ae25611.

website/pages/docs/secrets/key-management/index.mdx
austingebauer: Yep, this is invalid. I've been working to review the [learn content](https://github.com/hashicorp/learn/pull/2555) with @brianshumate. I'm not sure what the correct link is. I'll update this as soon as that's known 👍

website/pages/api-docs/secret/key-management/index.mdx
austingebauer: Yep, this is the correct output.

website/pages/api-docs/secret/key-management/index.mdx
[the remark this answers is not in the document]
austingebauer: Good catch! I've updated this in f4e38e4 and 595d7f6.

website/pages/docs/secrets/key-management/index.mdx
austingebauer: Added fake credentials to the example in 9e2e8a5. Let me know if you'd prefer these be something more like `client_id="example-client-id"` instead.

website/pages/docs/secrets/key-management/index.mdx
```diff
@@ -0,0 +1,189 @@
...
```
yhyakuna, line 182:
```suggestion
Refer to the [Key Management Secrets Engine](https://learn.hashicorp.com/tutorials/vault/key-management-secrets) guide for
```

This link will go live on Nov. 11th.

website/pages/docs/secrets/key-management/index.mdx
calvn: I think what you have is fine (similar to [aws sample values](https://www.vaultproject.io/api/auth/aws#sample-payload) in the docs)! It's mostly to indicate that these should be non-empty values.

calvn (approved):
👍 pending Yoko's suggestion!

website/pages/api-docs/secret/key-management/index.mdx
austingebauer: I think it would make sense to return those items here. Unfortunately, it would require a code change, which we're a little late for at this point. I think this would be a good item to gather feedback on in the tech preview.

austingebauer:
@calvn - I'm wondering if there is precedent for how we've communicated that a feature is in tech preview? I'm thinking that it may be something we'd want to mention in the docs.

calvn:
Yeah, maybe something like this along the top of the docs:

```
~> **NOTE:** The Key Management Secrets Engine is currently a **_Tech Preview_**
feature and not recommended for deployment in production.
```

website/pages/docs/secrets/key-management/index.mdx
brianshumate: 👋 The published URL for the learn tutorial is as follows:

```
https://learn.hashicorp.com/tutorials/vault/key-management-secrets?in=vault/secrets-management
```

yhyakuna: > 👋 The published URL for the learn tutorial is as follows:
> 
> ```
> https://learn.hashicorp.com/tutorials/vault/key-management-secrets?in=vault/secrets-management
> ```

You don't need `?in=vault/secrets-management` in the URL.  (`https://learn.hashicorp.com/tutorials/vault/key-management-secrets` is all you need.)

website/pages/docs/secrets/key-management/index.mdx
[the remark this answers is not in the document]
austingebauer: Changed in 1854c7a.

PR #10344, opened 2020-11-06
Error on root rotation when username is empty

pcman312:
Errors if no username is specified in the config when doing a root credential rotation. This should help clarify the error for DB plugins that don't use username/password for root credentials.

PR #10354, opened 2020-11-09
Backport 1.6: Error on root rotation when username is empty (#10344)

tomhjp:
* Error on root rotation when username is empty

* Don't panic if the field doesn't exist


catsby (approved):
LGTM

PR #10365, opened 2020-11-09
"vault operator usage" CLI for client count reporting

mgritter:
Here's the CLI in action:

```
mgritter@ubuntu:~/vault$ bin/vault operator usage
Period start: 2019-11-01T00:00:00Z
Period end: 2020-10-31T23:59:59Z

Namespace path              Distinct entities   Non-Entity tokens   Active clients
--------------              -----------------   -----------------   --------------
[root]                      448                 698                 1146
aaa/                        732                 794                 1526
aaa/ccc/                    1729                942                 2671
aaa/ddd/                    833                 929                 1762
aaa/ddd/eee/                476                 604                 1080
aaa/ddd/ggg/                1897                1117                3014
aaa/ddd/hhh/                515                 455                 970
bbb/                        911                 877                 1788
deleted namespace "mzK8K"   771                 1156                1927
                                                                     
Total                       8312                7572                15884

mgritter@ubuntu:~/vault$ bin/vault operator usage -ns=aaa/ddd
Period start: 2019-11-01T00:00:00Z
Period end: 2020-10-31T23:59:59Z

Namespace path   Distinct entities   Non-Entity tokens   Active clients
--------------   -----------------   -----------------   --------------
aaa/ddd/         833                 929                 1762
aaa/ddd/eee/     476                 604                 1080
aaa/ddd/ggg/     1897                1117                3014
aaa/ddd/hhh/     515                 455                 970
                                                          
Total            3721                3105                6826

mgritter@ubuntu:~/vault$ bin/vault operator usage -end-time=2019-06
No data is available for the given time range.

mgritter@ubuntu:~/vault$ bin/vault operator usage -start-time=2020-10-15
Period start: 2020-10-01T00:00:00Z
Period end: 2020-10-31T23:59:59Z

Namespace path              Distinct entities   Non-Entity tokens   Active clients
--------------              -----------------   -----------------   --------------
[root]                      342                 577                 919
aaa/                        176                 760                 936
aaa/ccc/                    822                 745                 1567
aaa/ddd/                    791                 180                 971
aaa/ddd/eee/                302                 217                 519
aaa/ddd/ggg/                899                 395                 1294
aaa/ddd/hhh/                184                 208                 392
bbb/                        871                 89                  960
deleted namespace "mzK8K"   87                  371                 458
                                                                     
Total                       4474                3542                8016
```


mgritter:
Help output:

```
mgritter@ubuntu:~/vault$ bin/vault operator usage -help
Usage: vault operator usage

  List the client counts for the default reporting period.

	  $ vault operator usage

  List the client counts for a specific time period.

          $ vault operator usage -start-time=2020-10 -end-time=2020-11

HTTP Options:

...

Output Options:

  -format=<string>
      Print the output in the given format. Valid formats are "table", "json",
      or "yaml". The default is table. This can also be specified via the
      VAULT_FORMAT environment variable.

Command Options:

  -end-time=<time>
      End of report period (defaults to end of last month.)

  -start-time=<time>
      Start of report period (defaults to default_reporting_period before end
      time.)
```

command/operator_usage.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func (c *OperatorUsageCommand) Flags() *FlagSets {
+	set := c.flagSet(FlagSetHTTP | FlagSetOutputFormat)
+
+	f := set.NewFlagSet("Command Options")
+
+	f.TimeVar(&TimeVar{
+		Name:       "start-time",
+		Usage:      "Start of report period (defaults to default_reporting_period before end time.)",
```
swayne275, line 53:
```suggestion
		Usage:      "Start of report period (defaults to default_reporting_period before end time).",
```

?

command/operator_usage.go
```diff
@@ -0,0 +1,286 @@
...
+		Target:     &c.flagStartTime,
+		Completion: complete.PredictNothing,
+		Default:    time.Time{},
+		Formats:    TimeVar_TimeOrDay | TimeVar_Month,
+	})
+	f.TimeVar(&TimeVar{
+		Name:       "end-time",
+		Usage:      "End of report period (defaults to end of last month.)",
```
swayne275, line 61:
```suggestion
		Usage:      "End of report period (defaults to end of last month).",
```

?

command/operator_usage.go
```diff
@@ -0,0 +1,286 @@
...
+
+	out := make([]string, len(nsOut))
+	for i := range nsOut {
+		out[i] = nsOut[i].formattedLine
+	}
+
+	return out
+}
+
+func (c *OperatorUsageCommand) addTotalToOutput(out []string, data map[string]interface{}) []string {
```
swayne275, line 256: could take `out` by reference instead of being passed in and returned? not a big deal though

swayne275 (approved):
Looks really good!

Should any of this stuff be unit tested?

mgritter:
I don't know if I can get testVaultServer() to return anything sensible for unit testing.

command/operator_usage.go
mgritter: It looks like "house style" is not to use parentheses at all, like this:

```
  -seal-wrap
      Enable seal wrapping of critical values in the secrets engine. The
      default is false.
```

command/operator_usage.go
mgritter: Hm, isn't the slice already "by reference"?

command/operator_usage.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	resp, err := client.Logical().ReadWithData("sys/internal/counters/activity", data)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error retrieving client counts: %v", err))
+		return 2
+	}
+
+	if resp == nil || resp.Data == nil {
+		c.UI.Warn("No data is available for the given time range.")
```
ivana-mcconnell, line 108: In cases where a user only provides an end date or a start date (and not both), would it be possible to show the time range itself? E.g. if a user gives an end date of October 31, we say "No data is available for the given time range START to END." It gives some visibility into what's actually being done and may help the user form their search better.

ivana-mcconnell (approved):

command/operator_usage.go
swayne275: Good to know!

command/operator_usage.go
swayne275: So interestingly I discussed this with a friend who was confused learning Go a few days ago. The value part of the slice is passed by reference, the metadata is not (capacity and length). So if you want to modify like `input[3]`, that will be reflected if it's passed normally (`someFunc(input []int)`). If you change the memory address of `input` however (using something like `append()`, the changes made in `someFunc` will not be reflected in the caller. If that's the behavior you want, you should write it like `someFunc(input *[]int)`, and then you're free to append or whatever you'd like

mladlow:
Can you add a changelog file for this please?

command/operator_usage.go
mgritter: I like this suggestion, but I'm having a hard time implementing it, because all the logic is on the server side, and not returned in the current API (which just does a 204 no content response.)

Retrieving the default_reporting_period would require an extra round trip.

mgritter: Instead, I added a check whether any queries were available, using the existing API.  So we are paying that round trip but I still felt it was the wrong direction to reproduce the range-calculation logic in the client.  (But I still agree that would be the clearest; maybe it makes sense to change the API response when no data is available.)

swayne275 (approved):

PR #10371, opened 2020-11-10
Add upgrade note for Transform

sgmiller:


website/pages/docs/upgrading/upgrade-to-1.6.0.mdx
```diff
@@ -0,0 +1,21 @@
...
+sidebar_title: Upgrade to 1.6.0
+description: |-
+  This page contains the list of deprecations and important or breaking changes
+  for Vault 1.6.0. Please read it carefully.
+---
+
+# Overview
+
+This page contains the list of deprecations and important or breaking changes
+for Vault 1.6.0 compared to 1.5.3. Please read it carefully.
```
mladlow, line 13: While 1.5.3 is the last one we have an upgrade guide for, the most recent 1.5 is 1.5.5. We don't really have any rules around this, but since it's a major release, maybe 1.5.x or just 1.5.
```suggestion
for Vault 1.6.0 compared to 1.5. Please read it carefully.
```

website/pages/docs/upgrading/upgrade-to-1.6.0.mdx
```diff
@@ -0,0 +1,21 @@
...
+
+## Transform Secrets Engine Storage Upgrade
+
+The Transform Secrets Engine (Enterprise only) will automatically upgrade the
+storage of its configuration in order to accommodate the tokenization
+feature.  Secondaries will receive the modifications via replication.
```
mladlow, line 19: Is it worth saying that we're adding a type to transformations here?

mladlow (changes requested):
I'm requesting changes just for the version - it shouldn't be 1.5.3 because we're at 1.5.5, and I think the delta should encompass the entirety of the 1.5 series. The upgrade note to 1.5 says 1.4.1, but the upgrade note to 1.4 says 1.3.x, so we're not really consistent.

I can't tell if a little more insight into what information we track about transformations would be help people trying to reason or plan their systems, so I'm fine with that section as-is if you don't think more information would be helpful.

website/data/docs-navigation.js
```diff
@@ -404,6 +404,7 @@ export default [
       'upgrade-to-1.5.1',
       'upgrade-to-1.5.2',
       'upgrade-to-1.5.3',
+      'upgrade-to-1.6.0',
```
calvn, line 407: Outside the scope of this PR, but I wonder if we should list on descending order and have latest upgrade guide at the top.

website/pages/docs/upgrading/upgrade-to-1.6.0.mdx
```diff
@@ -0,0 +1,22 @@
...
+# Overview
+
+This page contains the list of deprecations and important or breaking changes
+for Vault 1.6.0 compared to 1.5. Please read it carefully.
+
+## Transform Secrets Engine Storage Upgrade
+
+The Transform Secrets Engine (Enterprise only) will automatically upgrade the
+storage of its configuration in order to accommodate the new transformation
+type oriented configuration API.  Secondaries will receive the
```
calvn, line 19:
```suggestion
type oriented configuration API. Secondaries will receive the 
```

calvn:
I know that we have this info on several places now, but we should also add the database changes to the upgrade guide, particularly on the compatibility between Vault and plugins on different versions. cc: @pcman312

pcman312:
@calvn I agree we can put a note about the DB upgrade here, but I think it should be a separate PR. I'll put together some verbage for it now.

mladlow (approved):

PR #10374, opened 2020-11-11
Note about adding a bunch of nodes

mladlow:
@briankassouf I added a note about the situation we saw today, which I'm not 100% sure about.

I also shuffled one paragraph around and introduced some headings to break up the rest of the page a little.


aphorise:
Closing request in favour of #16772 - which has most of the same content on the new documentation layout.

PR #10375, opened 2020-11-11
Port: Telemetry For Lease Expiration Times

HridoyRoy:
Port from: https://github.com/hashicorp/vault-enterprise/pull/1509/files


command/server/config_test_helpers.go
```diff
@@ -787,3 +787,93 @@ func testParseSeals(t *testing.T) {
...
+			},
+			DisableClustering: true,
+		},
+
+		ServiceRegistration: &ServiceRegistration{
+			Type: "consul",
+			Config: map[string]string{
+				"foo": "bar",
+			},
+		},
```
mgritter, line 852: I'm a little surprised that all this has to be in this test, why not focus on just the telemetry?

mgritter (approved):

command/server/config_test_helpers.go
HridoyRoy: It doesn't was a copy/paste from another test. I can remove it!

PR #10376, opened 2020-11-11
Plugin Portal: Link to community plugin for Artifactory secrets

jsok:
A secrets plugin which issues Artifactory access tokens (short-lived and revocable).


sebinjohn (commented):
👍

PR #10379, opened 2020-11-11
Changelog: fixed a typo and added lease expiry metrics

HridoyRoy:
Related PR: https://github.com/hashicorp/vault/pull/10375


mgritter (commented):
This list is supposed to be grouped by subsystem, so the two "core" entries belong together.

mgritter (approved):
Looks good, but maybe @mladlow would prefer you use the new changelog automation?

mladlow:
@HridoyRoy the changelog updates are in 1.6.0 - did these changes go into 1.6.0? If they did, we can leave the changelog as-is, if they did not, can you use the wiki link I'll send you to change this PR to using the new changelog format. Thanks for pinging me @mgritter!

HridoyRoy:
Going to close this PR out and open an equivalent PR with the changelog tool.

PR #10384, opened 2020-11-11
InfluxDB - Check for errors in the response too

pcman312:
Fixes an issue where we aren't checking the error in the `Response` object in addition to the `err` returned.

### DIFF
--- a/plugins/database/influxdb/influxdb.go
+++ b/plugins/database/influxdb/influxdb.go
@@ -102,13 +102,16 @@ func (i *Influxdb) NewUser(ctx context.Context, req dbplugin.NewUserRequest) (re
 				"username": username,
 				"password": req.Password,
 			}
-			q := influx.NewQuery(dbutil.QueryHelper(query, m), "", "")
-			response, err := cli.Query(q)
-			if err != nil {
+			qry := influx.NewQuery(dbutil.QueryHelper(query, m), "", "")
+			response, err := cli.Query(qry)
+			// err can be nil with response.Error() being not nil, so both need to be handled
+			merr := multierror.Append(err, response.Error())
+			if merr.ErrorOrNil() != nil {
 				if response != nil && response.Error() != nil {
 					attemptRollback(cli, username, rollbackIFQL)
 				}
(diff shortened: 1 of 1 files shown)


plugins/database/influxdb/influxdb.go
```diff
@@ -102,13 +102,17 @@ func (i *Influxdb) NewUser(ctx context.Context, req dbplugin.NewUserRequest) (re
...
+			// err can be nil with response.Error() being not nil, so both need to be handled
+			merr := multierror.Append(err, response.Error())
+			if merr.ErrorOrNil() != nil {
+				// Attempt rollback only when the response has an error
 				if response != nil && response.Error() != nil {
 					attemptRollback(cli, username, rollbackIFQL)
 				}
-				return dbplugin.NewUserResponse{}, fmt.Errorf("failed to run query in InfluxDB: %w", err)
+
+				return dbplugin.NewUserResponse{}, fmt.Errorf("failed to run query in InfluxDB: %w", merr.ErrorOrNil())
```
calvn, line 115: Since this is already wrapped by a nil check, it should be safe to call `merr` right?

plugins/database/influxdb/influxdb.go
```diff
@@ -133,10 +137,10 @@ func attemptRollback(cli influx.Client, username string, rollbackStatements []st
...
 
 			response, err := cli.Query(q)
-			if err != nil {
-				if response != nil && response.Error() != nil {
-					return err
-				}
+			// err can be nil with response.Error() being not nil, so both need to be handled
+			merr := multierror.Append(err, response.Error())
+			if merr.ErrorOrNil() != nil {
+				return merr.ErrorOrNil()
```
calvn, line 143: Nit: We can `return merr` directly in here since we've nil checked.

plugins/database/influxdb/influxdb.go
pcman312: Technically yes, though either works

plugins/database/influxdb/influxdb.go
pcman312: Done

calvn (approved):

PR #10386, opened 2020-11-11
allow null types in batch encryption [VAULT-849]

HridoyRoy:
This PR is to allow null values in the encryption payload, so as to be backward compatible with 1.4 functionality.


builtin/logical/transit/path_encrypt.go
```diff
@@ -154,39 +155,49 @@ func decodeBatchRequestItems(src interface{}, dst *[]BatchRequestItem) error {
 		}
 
 		if v, has := item["context"]; has {
-			if casted, ok := v.(string); ok {
+			val := reflect.ValueOf(v)
```
vishalnayak, line 158: Nit: We may not need `val`. How about `if !reflect.ValueOf(v).IsValid()`?

vishalnayak (approved):
One small comment. Otherwise this is looking good.

builtin/logical/transit/path_encrypt.go
HridoyRoy: Sounds good! Thank you for the review!

HridoyRoy:
I updated the PR to not allow plaintext to be "null". Everything else can have a "null" string as a value, as they are optional parameters in different situations. I let the ciphertext field be "null" as well, even though it shouldn't be necessary, because this value is not set by a vault user.

PR #10387, opened 2020-11-12
Add automated raft snapshots to sidebar

kalafut:


website/data/docs-navigation.js
```diff
@@ -425,6 +425,7 @@ export default [
         category: 'hsm',
         content: ['behavior', 'security'],
       },
+      { category: 'automated-raft-snapshots' },
       { category: 'lease-count-quotas' },
```
mladlow, line 429: This isn't entirely on you, and I can go in and change this, but we should be using integrated storage here and the heading on the page should also change.

kalafut: Updated

mladlow (approved):

PR #10402, opened 2020-11-16
setup example

Monkeychip:
Example for hack project to another team.

### DIFF
--- a/ui/app/router.js
+++ b/ui/app/router.js
@@ -15,6 +15,7 @@ Router.map(function() {
       this.route('logout');
       this.mount('open-api-explorer', { path: '/api-explorer' });
       this.route('license');
+      this.route('meep'); // NOTE: add query params here, ex: { path: '/:param' }
       this.route('metrics', function() {
         this.route('index', { path: '/' });
         this.route('config');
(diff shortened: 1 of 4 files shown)


ui/app/router.js
```diff
@@ -15,6 +15,7 @@ Router.map(function() {
       this.route('logout');
       this.mount('open-api-explorer', { path: '/api-explorer' });
       this.route('license');
+      this.route('meep'); // NOTE: add query params here, ex: { path: '/:param' }
```
Monkeychip, line 18: the colon indicates it's optional. https://guides.emberjs.com/v3.8.0/routing/query-params/

ui/app/routes/vault/cluster/meep.js
```diff
@@ -0,0 +1,8 @@
+import Route from '@ember/routing/route';
+import ClusterRoute from 'vault/mixins/cluster-route';
+
+export default Route.extend(ClusterRoute, {
+  model() {
+    // return this.store.queryRecord('license', {});
```
Monkeychip, line 6: Here's a link on the significance of a route's model.  https://guides.emberjs.com/v3.8.0/routing/specifying-a-routes-model/

ui/app/router.js
Monkeychip: Another note.  We are using Ember version 3.8.  The most recent docs will always land on version 3.22, which is quite different in structure to how our code looks.  Make sure on the upper-left side of the docs pages to select Ember version 3.8 when looking at the documentation.

PR #10404, opened 2020-11-17
Revert "Disable test that fails on OSS. (#10401)"

swayne275:
Revert skipping failing activity log test on OSS


mgritter:
Should this be a draft?

swayne275:
@mgritter are "green main branch" re-enables supposed to be? if so I must have missed that in the doc

swayne275:
This has already been taken care of

PR #10409, opened 2020-11-17
updates to get tests running

sarahethompson:
Ran `yarn install` locally and pushed the changes to yarn.lock to this branch so that the cache is properly persisted between jobs.
Upped the resource_class to xlarge (from medium+) as this seemed to be needed for the tests to run. 
Angel is checking whether this is definitely needed.


Monkeychip (approved):
I ran a package size analyzer between master and this branch and nothing stands out as a reason for the necessary increase in resource_class.  However, if we need the increase in size, and there's no downside I'm unaware of, I think the move from medium+ to xlarge should be good.

PR #10414, opened 2020-11-18
Clarify docs around autoseals

ncabatoff:
Avoid using "KMS seal" since it causes confusion: some people think that excludes HSMs, and it's not obvious that Transit is a KMS.

Highlight that recovery keys can't be used for recovery when an auto seal is broken (unfortunate name, that.)


vishalnayak (approved):
👍

PR #10417, opened 2020-11-18
Fix delete role issue on transform

chelshaw:
Adding and then removing roles on a transformation when no roles have been created causes an error when removing them, which means they don't actually get removed from the model. 

**GIF of issue:**
transform-bug


Monkeychip (commented):
Small test failure, but I tried to break this and couldn't.  Nice fix.

changelog/10417.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
```
Monkeychip, line 1: nice.  good idea to include changelog in the pr record.  I'll start doing this as well.

Monkeychip (approved):

PR #10424, opened 2020-11-20
Fix license caching issue

michelvocks:
This PR fixes an issue where the license is cached and impossible to pickup a new license by the License-Manager if it got updated by an active node.


briankassouf (approved):

mladlow:
@michelvocks Could you add a changelog file for this before merging it please? That way if you wind up backporting it we'll still get the update in the changelog.

mladlow (approved):
Thanks!

PR #10432, opened 2020-11-23
Added documentation for `vault operator usage`.

mgritter:


website/pages/docs/commands/operator/usage.mdx
```diff
@@ -0,0 +1,73 @@
...
+# operator usage
+
+The `operator usage` command allows an administrator to retrieve a
+[client count](https://www.vaultproject.io/docs/concepts/client-count) report
+for the default reporting period, or for a specific time range of months.
+
+The command output will list distinct entities, non-entity tokens, and total
+active clients for each namespace, and the total for the entire cluster.  It may
+indicate that no data is available for the requested time range, which may be
+because the collection of client count is disabled, or the time range is too far
```
swayne275, line 18:
```suggestion
because the client count reporting is disabled, the time range is too far
```

swayne275 (approved):
Looks good! I think the suggestion I made would make it a little easier to grok

website/pages/docs/commands/operator/usage.mdx
mgritter: Sure. I was trying to make a distinction between collection (which might be disabled) and reporting (which never is.)

swayne275: Oh, that makes sense. Maybe `client count <some word instead of reporting>` then?

PR #10433, opened 2020-11-23
MySQL - Fix username generation length bug

pcman312:
Fixes an issue where the display name in MySQL dynamic users is the incorrect length.

This also update the github.com/ory/dockertest dependency because it was failing to build locally.

Fixes https://github.com/hashicorp/vault/issues/10476


Valarissa (approved):
Looks like a much better way of isolating the differences for username lengths. I heartily approve.

austingebauer (approved):

PR #10434, opened 2020-11-24
Using changelog tool to update changelog [VAULT-1055]

HridoyRoy:
From: https://github.com/hashicorp/vault/pull/10379


HridoyRoy:
One question: not sure if we want to keep our release-notes the same as our current changelog -- improvement, change, bug-fix, and feature? I only saw bug and improvement in the other examples, so I decided to put improvement as the release-note, instead of feature.

raskchanky (approved):
It seems fine to use `feature` to me, if that feels like the most appropriate category for this. I would imagine that we want to use the same headings as we did previously, though @mladlow can correct me if I'm wrong. I'm personally not a huge fan of `changes` as a category, given that the entire changelog is literally a log of changes, so that feels a little superfluous.

mladlow:
I think it's either an improvement or a feature. I kind of lean "improvement" since I think we've mostly used "feature" for introducing new features as part of a major release.

PR #10440, opened 2020-11-24
Backport 1.6.1: "vault operator usage" CLI for client count reporting (#10365)

mgritter:
* Working draft of CLI command.
* Sort order, robustness checking.
* Text edits and check of queries_available.
* Added changelog.

### DIFF
--- a/command/operator_usage.go
+++ b/command/operator_usage.go
@@ -1,0 +1,320 @@
+package command
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/vault/api"
+	"github.com/mitchellh/cli"
+	"github.com/posener/complete"
+	"github.com/ryanuber/columnize"
+)
+
+var _ cli.Command = (*OperatorUsageCommand)(nil)
+var _ cli.CommandAutocomplete = (*OperatorUsageCommand)(nil)
+
+type OperatorUsageCommand struct {
+	*BaseCommand
+	flagStartTime time.Time
+	flagEndTime   time.Time
+}
+
+func (c *OperatorUsageCommand) Synopsis() string {
+	return "Lists historical client counts"
+}
+
+func (c *OperatorUsageCommand) Help() string {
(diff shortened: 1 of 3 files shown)


command/operator_usage.go
```diff
@@ -0,0 +1,320 @@
...
+
+func (c *OperatorUsageCommand) AutocompleteFlags() complete.Flags {
+	return c.Flags().Completions()
+}
+
+func (c *OperatorUsageCommand) Run(args []string) int {
+	f := c.Flags()
+
+	if err := f.Parse(args); err != nil {
+		c.UI.Error(err.Error())
```
swayne275, line 84: i'm assuming that `c.UI.<loglevel>` calls output strings to the user? If so, do we want to expose the internal workings of the system by printing the error given back from function calls?

command/operator_usage.go
```diff
@@ -0,0 +1,320 @@
...
+
+	client, err := c.Client()
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 2
+	}
+
+	resp, err := client.Logical().ReadWithData("sys/internal/counters/activity", data)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error retrieving client counts: %v", err))
```
swayne275, line 104: same as above

swayne275 (approved):
Looks good overall.

I guess I'm not sure of the c.UI.Error/Warn calls. From the code it appears both go to stdout for the CLI. In this code it looks like (at least in general) the Warn messages are formatted with a particular style (caps and punctuation), but the Error messages are not. Do we want those to be in sync with each other and of a specific format for UX? Also, do we want to expose errors from called functions via stderr (rather than some user facing message on std err and the more specific details in a log)?

command/operator_usage.go
mgritter: This is how the client handles most errors from Vault, I think. If there's a better example I'll copy it (but that should be addressed in the main branch, not this backport PR.)

swayne275: oh okay, wasn't sure what the ux was

PR #10461, opened 2020-11-27
Cap username length

whs:
Vault MySQL backend allow the displayname part to be as much as the entire username. Some display name, specifically kubernetes can be very long. In my case, our database user is named `v_kubernetes-sa-rolename-` and the random part in the end truncated.

I believe (but did not investigate) that Vault then treat each leases (that has the same username) as separate user but by revoking one user, it unknowingly revoked them all.

This PR sets the display name cap to a sensible value to make sure that at least a few random parts is always present.

### DIFF
--- a/plugins/database/mysql/mysql.go
+++ b/plugins/database/mysql/mysql.go
@@ -28,11 +28,23 @@ const (
 	mySQLTypeName = "mysql"
 )
 
+// Modern
+// v_  displayname_ metadata_ uuid_time
+// -2- -----14----- ----11--- ----5----
+// ---------------32------------------
+//
+// Legacy
+// v_  displayname_ metadata_ uuid_time
+// -2- -----6------ ----5---- ----3----
+// ---------------16------------------
+
 var (
-	MetadataLen       int = 10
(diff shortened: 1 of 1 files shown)


plugins/database/mysql/mysql.go
```diff
@@ -28,11 +28,23 @@ const (
 	mySQLTypeName = "mysql"
 )
 
+// Modern
```
whs, line 31: Leaving this part in for future people changing these constants to know to leave some spaces for all parts.

calvn:
Thanks for the contribution! We’re currently looking at overhauling our username generation to allow for some customization across plugins.

We'll leave this open to track, but will likely take a more comprehensive and generic approach to customizing usernames.

kalafut:
@whs Thanks again for this. As Calvin mentioned we'll be delivering a flexible username customization solution soon, in our next major release. You can view the WIP docs to see what that will look like: https://github.com/hashicorp/vault/pull/10935/files#diff-971425389a1a6dcbec6c2b14a626ac8ac5187f1c3c5d9d4caaa7be55fd07b160

Issue #10463, opened 2020-11-30
Add ListAccessors function to API client


ptzianos:
**Is your feature request related to a problem? Please describe.**
It would be very useful to have a function to list the token accessors so that we can programmatically find and revoke root tokens or tokens that have a certain policy attached to them. (probably there are other scenarios too, but this is the one I'm trying to fix atm) 

**Describe the solution you'd like**
The API client is missing a function for listing token accessors 

**Describe alternatives you've considered**
-

**Explain any additional use-cases**
-

**Additional context**
-

PR #10464, opened 2020-11-30
Api client list accessors

ptzianos:
Fixes: #10463

Changes:
- Add a function to the API client so that a user can list the token accessors of the system
- Split and re-use function for parsing token accessors
- Add a small helper to decode temp root tokens


ptzianos:
note for reviewer: If the inclusion of the root token decoding function is a problem I have no problem splitting it away in a separate PR.

api/secret.go
```diff
@@ -113,69 +146,19 @@ func (s *Secret) TokenPolicies() ([]string, error) {
 		return s.Auth.Policies, nil
 	}
 
-	if s.Data == nil || s.Data["policies"] == nil {
+	if s.Data == nil {
```
ncabatoff, line 149: Looks like you could get rid of the `if` altogether, since the test is already present in extractListOfStrings.

api/secret.go
```diff
@@ -113,69 +146,19 @@ func (s *Secret) TokenPolicies() ([]string, error) {
...
-		for _, v := range list {
-			p, ok := v.(string)
-			if !ok {
-				return nil, fmt.Errorf("unable to convert policy %v to string", v)
-			}
-			tokenPolicies = append(tokenPolicies, p)
-		}
+	tokenPolicies, err := s.extractListOfStrings("policies")
+	if tokenPolicies == nil || err != nil {
+		return tokenPolicies, err
```
ncabatoff, line 155: There are cases in the current code where we goto TOKEN_DONE despite having an empty tokenPolicies.  I think this `if`'s condition should be just `err != nil`.

api/secret.go
```diff
@@ -189,6 +172,23 @@ DONE:
...
 }
 
+// TokenAccessors returns the standardized list of accessors for the given secret
+func (s *Secret) TokenAccessors() ([]string, error) {
+	if s == nil {
+		return nil, nil
+	}
+
+	if s.Auth != nil && len(s.Auth.Policies) > 0 {
+		return s.Auth.Policies, nil
```
ncabatoff, line 182: This doesn't look correct.  Policies aren't accessors, right?

api/secret.go
```diff
@@ -189,6 +172,23 @@ DONE:
 	return policies, nil
 }
 
+// TokenAccessors returns the standardized list of accessors for the given secret
+func (s *Secret) TokenAccessors() ([]string, error) {
```
ncabatoff, line 176: You name this `TokenAccessors`, but it could just as well be called `Keys` or something, right?  It's the same code that would be used to extract the result of any `list` operation from a `Secret`.

ptzianos: Yes, you are right on that. I removed the if statement that would return the secret's accessor value because that was wrong. Instead the function call basically just invokes the `extractListOfStrings` with the correct parameters.

api/secret.go
ptzianos: Good catch. fixed now

api/secret.go
ptzianos: Done

api/secret.go
ptzianos: Yeah, I didn't notice it. Now the latest version of the code should be equivalent in functionality

ptzianos:
@ncabatoff  it looks like I need some sort of authorization from someone in order to get my PR deployed on Vercel? Other than that I fixed all your suggestions and added tests which are passing

PR #10467, opened 2020-11-30
Add retry policy and fix documentation for Cassandra storage backend

byumov:
Shortly, this PR fixes two problems:
1) Current documentation for Cassandra backend is incorrect
2) There is no option to change the default retry policy for cluster now

More details below...

1. Current [documentation](https://www.vaultproject.io/docs/configuration/storage/cassandra#connection_timeout) says, that default value for `connection_timeout` is `0`, you might think there is no timeout by default, but in fact, if we don't set these options timeout will be 600ms.
see: https://github.com/hashicorp/vault/blob/0e8c6c21718bee15fe01debfd6a990e84e5cf16c/physical/cassandra/cassandra.go#L127-L133
`cluster.Timeout` is not changing somewhere else and get default value:
https://github.com/hashicorp/vault/blob/665d668f058192b3e29fff5337d7ad144fb4f20b/vendor/github.com/gocql/gocql/cluster.go#L49

2. If we have a Cassandra cluster with several nodes, we don't want to get an error if one of nodes has gone. For supporting this behavior with [gocql](https://github.com/gocql/gocql)(which used in Cassandra backend) we must set one of RetryPolicy for the cluster: https://github.com/gocql/gocql/blob/5913df4d474e0b2492a129d17bbb3c04537a15cd/policies.go#L158
By default(current behavior) cluster use "retry on same connection" policy. So, if current active node is down client will get an error: 
https://github.com/gocql/gocql/blob/5913df4d474e0b2492a129d17bbb3c04537a15cd/policies.go#L141
We can easily fix this by using SimpleRetryPolicy (retry for another connection) when creating the cluster.

Also, one new option added to set timeout for the initial connection.

I tried to make changes as little as possible.
This PR doesn't change the behavior of current installations.


byumov:
@ncabatoff, Hi! Can you have a look, please?

aphorise:
Enhancement request along with accompanying documentation improvements do seem relevant. Can this be reviewed for release in the near future?

aphorise:
Likely related to #15899 - and on merger user on that issue should be advised to retest before closing that issue too.

HridoyRoy (approved):
lgtm

HridoyRoy:
The changelog check is ok -- I triggered the CI run off another branch and PR, so the check is looking for the wrong changelog entry.

PR #10469, opened 2020-11-30
fix data races in activity log tests

sbunce:
The race detector pointed out these problems. This is why the circleci test-race-go rule was failing.


swayne275:
This is awesome, thanks! We've actually been tracking this for a bit (and fixed [here](https://github.com/hashicorp/vault/pull/10484)) so I'll close the issue.

PR #10470, opened 2020-12-01
Add details on Consul token ttl

danielehc:
The current integration with Consul does not take into account the fact that, when tokens expire, Consul will start failing unless tokens are manually rotated.

This PR adds two callout to warn users about this behavior.

Was thinking to add them also to [/website/pages/api-docs/secret/consul/index.mdx](https://github.com/hashicorp/vault/blob/dc/consul-secrets-engine-docs/website/pages/api-docs/secret/consul/index.mdx) but I'm not sure if I'd break any automation in that case.


website/pages/docs/secrets/consul/index.mdx
```diff
@@ -118,6 +122,11 @@ accessor           6d5a0348-dffe-e87b-4266-2bec03800abb
 token              bc7a42c0-9c59-23b4-8a09-7173c474dc42
 ```
 
+!> **Expired token rotation:** Currently, once the TTL expires for the tokens and unless you set a workflow in place to manually rotate the tokens,
+Consul operations will start failing because of the tokens being expired. At this time, the recommended approach for operators is to rotate the 
+tokens manually by creating a new token using the `vault read consul/creds/my-role` command and then, once the token is synchronized with Consul 
+apply the token to the agents using the Consul API or CLI.
+
```
picatz, line 129:
```suggestion
!> **Expired token rotation:** Once a token's TTL expires, then Consul operations will no longer be allowed with it. This requires you to have an external process to rotate tokens. At this time, the recommended approach for operators is to rotate the tokens manually by creating a new token using the `vault read consul/creds/my-role` command. Once the token is synchronized with Consul, apply the token to the agents using the Consul API or CLI.
```

picatz (approved):
LGTM, just one small wording suggestion. 🎉

danielehc:
@hashicorp/vault I am not authorized to merge the PR, shall I unassign it and leave it free to take?

Also let me know if you want me to add the same message in the API docs.

vishalnayak:
The diff looks good. I don't think this is needed in the API docs. Merging it!

PR #10473, opened 2020-12-01
Add changelog for #10416

tomhjp:
I forgot to add this along with the original PR and it seems worthy of a changelog entry to me.

I've also added a readme pointing to the go-changelog tool for documentation.


mladlow (approved):
Thanks - agree that a changelog is good. Thanks also for adding the readme.

#10416 doesn't have a milestone - if it needs to be backported to 1.6 the changelog entry will need to ride along (either another backport or combined in the backport PR) so it shows up in the 1.6.x changelog.

PR #10475, opened 2020-12-01
Add 2 fuzzers

AdamKorcz:
This PR adds two fuzzers for `vault.ParseACLPolicy` and `random.ParsePolicy` respectively.

A bit of context: I have previously worked on fuzzing Vault and the ParseACLPolicy fuzzer has previously found a bug that has led to [this fix](https://github.com/hashicorp/hcl/pull/410). 

I am therefore committing these two fuzzers with the suggestion of setting up continuous fuzzing for Vault through OSS-fuzz. I will shortly be setting of a draft integration on the OSS-fuzz side, and in case there is interest in completing that integration, all that is needed is to merge the fuzzers and provide at least one maintainers email address. This will allow Google to run the fuzzers continuously and notify maintainers in case bugs are found. The service is offered free of charge to open source projects with the implied expectation that bugs are fixed so that the resources spent on running Vaults fuzzers are put to good use. 

Fuzzing has proven effective to find bugs in mature software systems, and in essence it is a technique to test programs whereby pseudo-random data is passed to a target with the goal of uncovering bugs and vulnerabilities. In that regard fuzzing continuously has found vulnerabilities in large projects like Kubernetes in the past. 

Naturally a lot more can be done in terms of fuzzing Vault, and these two fuzzers are a way for Vault to get started with fuzz-testing.

The fuzzers in this PR are implemented by way of Dvyukovs go-fuzz engine which is a coverage-guided fuzzer. This means that the fuzzer uses feedback from each run in future input-data creation.


AdamKorcz:
I see that the CircleCI badge also returns "Failed". This does not look related to the files in this commti

AdamKorcz:
The failing build looks unrelated to the files in this PR. Inputs on this are highly appreciated.

xntrik:
This is looking pretty great @AdamKorcz. I'm wondering that given the fuzzed functions end up being processed inside of https://github.com/hashicorp/hcl whether it would make more sense to add the fuzzers there?

Interestingly, while the current `master` branch of hcl is v1 (which is used by these Vault methods), the v2 branch currently includes a few fuzz setups already, for example https://github.com/hashicorp/hcl/tree/hcl2/hclsyntax/fuzz.

AdamKorcz:
> I'm wondering that given the fuzzed functions end up being processed inside of https://github.com/hashicorp/hcl whether it would make more sense to add the fuzzers there?

To me it is all the same. If you (and other maintainers) prefer to have fuzzers for entrypoints in Vault over at https://github.com/hashicorp/hcl, absolutely. Should I move them over there?

xntrik:
I think moving them over there makes more sense, as HCLv1 is used in more places than just Vault.

aphorise:
Hey @AdamKorcz thanks for you great efforts here. Can you confirm how relevant this may still be for you?

@ncabatoff any chance this can be considered for merging soon?

biazmoreira:
Hi Adam,

Thanks for the contribution! We think Fuzzing is important, but before introducing it to the codebase or any other related-engine, we think it's best if our product and security teams review it and add examples on how to approach this as well as automating it to be part of our testing suite. 

I have opened a request to the Security team and will close this PR for now. If you have an issue associated with this PR, could you please link it here? It will help us keep an eye on requests like these in the future. Thanks!

PR #10478, opened 2020-12-01
[1.5.x] Update CL

vishalnayak:


changelog/_1576.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
```
mladlow, line 1: I don't think you would need this file - just the 10285 would suffice.

mladlow (approved):

changelog/_1576.txt
vishalnayak: Fixed.

PR #10484, opened 2020-12-02
Fix Racy Activity Log Tests

swayne275:
Also moved some testing utilities away from the main activity log logic

### DIFF
--- a/vault/activity_log_testing_util.go
+++ b/vault/activity_log_testing_util.go
@@ -129,7 +129,6 @@ func (a *ActivityLog) GetStartTimestamp() int64 {
 }
 
 // SetStartTimestamp sets the start timestamp on an activity log
-// TODO move to ent if only used in those tests
 func (a *ActivityLog) SetStartTimestamp(timestamp int64) {
 	a.l.Lock()
 	defer a.l.Unlock()
--- a/vault/activity_log_testing_util.go
+++ b/vault/activity_log_testing_util.go
@@ -1,0 +1,159 @@
+package vault
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/vault/sdk/logical"
+	"github.com/hashicorp/vault/vault/activity"
+)
+
+// Return the in-memory activeEntities from an activity log
+func (c *Core) GetActiveEntities() map[string]struct{} {
+	out := make(map[string]struct{})
+
+	c.stateLock.RLock()
+	c.activityLog.fragmentLock.RLock()
+	for k, v := range c.activityLog.activeEntities {
+		out[k] = v
+	}
+	c.activityLog.fragmentLock.RUnlock()
+	c.stateLock.RUnlock()
+
+	return out
+}
+
+// GetCurrentEntities returns the current entity activity log
+func (a *ActivityLog) GetCurrentEntities() *activity.EntityActivityLog {
+	a.l.RLock()
+	defer a.l.RUnlock()
+	return a.currentSegment.currentEntities
+}
+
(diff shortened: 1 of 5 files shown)


vault/activity_log_testing_util.go
```diff
@@ -0,0 +1,167 @@
...
+// GetCurrentEntities returns the current entity activity log
+func (a *ActivityLog) GetCurrentEntities() *activity.EntityActivityLog {
+	a.l.RLock()
+	defer a.l.RUnlock()
+	return a.currentSegment.currentEntities
+}
+
+// WriteToStorage is used to put entity data in storage
+// `path` should be the complete path (not relative to the view)
+func WriteToStorage(t *testing.T, c *Core, path string, data []byte) {
```
ncabatoff, line 35: I think this should be moved to the testhelpers package.

vault/core.go
```diff
@@ -1926,7 +1926,8 @@ func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c
 		if err := c.setupAuditedHeadersConfig(ctx); err != nil {
 			return err
 		}
-		if err := c.setupActivityLog(ctx); err != nil {
+		var wg sync.WaitGroup
```
ncabatoff, line 1929: No Wait call?  That's fine if intended, but I'd put a comment here to say that's deliberate, since it looks funny otherwise.

vault/activity_log_testing_util.go
```diff
@@ -0,0 +1,167 @@
...
+
+// GetStartTimestamp returns the start timestamp on an activity log
+func (a *ActivityLog) GetStartTimestamp() int64 {
+	a.l.RLock()
+	defer a.l.RUnlock()
+	return a.currentSegment.startTimestamp
+}
+
+// SetStartTimestamp sets the start timestamp on an activity log
+// TODO move to ent if only used in those tests
```
ncabatoff, line 132: Can this TODO be removed?

vault/activity_log_testing_util.go
```diff
@@ -0,0 +1,167 @@
...
+	if verifyTimeNotZero {
+		if a.currentSegment.startTimestamp == 0 {
+			t.Error("bad start timestamp. expected no reset but timestamp was reset")
+		}
+	} else if a.currentSegment.startTimestamp != expectedStart {
+		t.Errorf("bad start timestamp. expected: %v got: %v", expectedStart, a.currentSegment.startTimestamp)
+	}
+}
+
+// ActiveEntitiesEqual checks that only the set of `test` exists in `active`
```
ncabatoff, line 109: Let's put this in testhelpers too.  The `vault` package is already enormous and there's no reason this needs to be here that I can see.

vault/activity_log_testing_util.go
swayne275: `testhelpers` can't be imported into the `vault` package (creates an import cycle), and `WriteToStorage` is being used in the `vault` package. Is there somewhere else where this could go

vault/core.go
swayne275: I didn't want to change the existing behavior (even though it's probably harmless). Added a comment: 3389c4396

vault/activity_log_testing_util.go
swayne275: good catch: e05aad428

vault/activity_log_testing_util.go
swayne275: same problem as above (import cycle). Is there somewhere else we could put this?

ncabatoff (approved):

PR #10485, opened 2020-12-02
core: add warning when disabling activity

alexanderbez:


vault/logical_system_activity.go
```diff
@@ -216,5 +216,13 @@ func (b *SystemBackend) handleActivityConfigUpdate(ctx context.Context, req *log
 	// Set the new config on the activity log
 	a.SetConfig(ctx, config)
 
+	if config.Enabled == "disable" {
+		return &logical.Response{
```
mgritter, line 220: Could we be more specific about doing this only one enable->disable changes?

I think this would appear even if we changed the retention period while activity log was disabled.

Another thing to think about is reset back to "default" when running on OSS, which is also disabled.

alexanderbez: > Could we be more specific about doing this only on enable->disable changes?

Good catch. Updated.

> Another thing to think about is reset back to "default" when running on OSS, which is also disabled.

Can you elaborate a bit more on this?

swayne275: > > Another thing to think about is reset back to "default" when running on OSS, which is also disabled.

> Can you elaborate a bit more on this?

I imagine he just means the following. Activity log is disabled by default on OSS (enabled by default on ENT). If a user is running OSS with activity log set to enable and then wants to restore to default settings, it will disable activity log (thus deleting data which should trigger a notification)

vault/logical_system_activity.go
```diff
@@ -186,8 +188,13 @@ func (b *SystemBackend) handleActivityConfigUpdate(ctx context.Context, req *log
 	{
 		// Parse the enabled setting
 		if enabledRaw, ok := d.GetOk("enabled"); ok {
+			if config.Enabled == "enable" && enabledRaw.(string) == "disable" {
+				warnings = append(warnings, "the current monthly segment will be deleted")
```
swayne275, line 192: does it make sense to add something like "because activity log was disabled"?

vault/logical_system_activity.go
alexanderbez: I've pushed an update to handle when enabled -> default. Since this is in the OSS repo, I imagine when we port this to ENT, we'll have this removed. Is that the correct line of thinking? Or is there some other way to check?

swayne275: Your line of thinking is correct, but I think the implementation might be different. The pattern I've seen so far when the same function signature needs different behavior is to use build tags (so when we compile for OSS we get one version of the function and when we compile for ENT we get a different version of the function). Since this is an OSS PR I'm not sure that I should comment a specific ent example here, so I'll DM you.

@ncabatoff seems to have quite a bit of experience with this, so it might be good to get his thoughts

mgritter: core_util.go has activityLogEnabledDefault, which is false on OSS builds and true on Enterprise builds; it's probably better to use that value than to have other code that differs between OSS and ENT.

vault/logical_system_activity.go
```diff
@@ -186,8 +188,13 @@ func (b *SystemBackend) handleActivityConfigUpdate(ctx context.Context, req *log
 	{
 		// Parse the enabled setting
 		if enabledRaw, ok := d.GetOk("enabled"); ok {
+			if config.Enabled == "enable" && enabledRaw.(string) == "disable" {
```
mgritter, line 191: enabledRaw could also be "default".

I think it would be better to have SetConfig return the warnings list, so we can handle the 3x3 matrix "enable", "disable", "default" correctly all in one place.  The code there already interprets "default" correctly, so all you have to do is check the original and the final value of a.enabled.

alexanderbez: But then we have to change the method signature and all corresponding unit tests. Also, `SetConfig(ctx, config) (warnings []string)` seems like a strange API, no? What's wrong with just having these checks here?

mgritter: I would like to avoid duplicating the logic to parse the enabled state and apply it correctly.  (On the other hand, the parsing *does* appear three times in the code already...)

It should be fine to check a.enabled before and after SetConfig instead, but we would probably then need a new IsEnabled() method that handles the mutex locking correctly to avoid setting off the race detector.

alexanderbez: Sorry, I'm a bit lost here. Is the following not correct?

```go
if config.Enabled == "enable" && (enabledStr == "disable" || enabledStr == "default") {
  warnings = append(warnings, "...")
}
```

I'm not doing any additional parsing here. I'm just checking the new value against the current value.

mgritter: It's correct on OSS but not on enterprise. You would need to incorporate activityLogEnabledDefault, or move the check into a function that is similarly set up to have oss and ent versions.  Something like:

```
if ( config.Enabled == "enable" || (config.Enabled == "default" && activityLogEnabledDefault)) && 
   (enabledStr == "disable" || (enabledStr == "default" && !activityLogEnabledDefault) { 
   ...
}
```

mladlow:
@alexanderbez I've tentatively milestoned this for 1.6.1, but depending on where the quotas work is, it might go to 1.6.2. Can you update or confirm the milestone on Monday please?

alexanderbez:
I've pushed changes @mladlow and @mgritter

mladlow:
I've changed the milestone to 1.6.2 so we can try to move ahead with staging.

vault/logical_system_activity.go
```diff
@@ -186,8 +188,16 @@ func (b *SystemBackend) handleActivityConfigUpdate(ctx context.Context, req *log
...
 		if enabledRaw, ok := d.GetOk("enabled"); ok {
-			config.Enabled = enabledRaw.(string)
+			enabledStr := enabledRaw.(string)
+
+			// If currently enabled with the intent to disable or intent to revert to
+			// default in a OSS context, then we return a warning to the client.
+			if config.Enabled == "enable" && enabledStr == "disable" ||
+				!activityLogEnabledDefault && config.Enabled == "enable" && enabledStr == "default" {
+				warnings = append(warnings, "the current monthly segment will be deleted because the activity log was disabled")
+			}
```
mgritter, line 198: There are three transitions where we need to warn

enabled => disabled                                                    // part before ||, OK
enabled => default && !activityLogEnabledDefault    // part after ||, OK
default && activityLogEnabledDefault => disabled    // I don't see how this gets handled

alexanderbez: `default && activityLogEnabledDefault => disabled` It's not handling this. I'll fix it.

alexanderbez: `default && activityLogEnabledDefault => disabled` It's not handling this. I'll fix it.

alexanderbez: `default && activityLogEnabledDefault => disabled` It's not handling this. I'll fix it.

mgritter (approved):

PR #10489, opened 2020-12-03
core: Record the time a node became active

briankassouf:


vault/core.go
```diff
@@ -1850,6 +1854,10 @@ func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c
 	c.clearForwardingClients()
 	c.requestForwardingConnectionLock.Unlock()
 
+	// Mark the active time. We do this first so it can be corrlated to the logs
```
ncabatoff, line 1857:
```suggestion
	// Mark the active time. We do this first so it can be correlated to the logs
```

http/sys_leader.go
```diff
@@ -48,13 +52,14 @@ func handleSysLeaderGet(core *vault.Core, w http.ResponseWriter, r *http.Request
...
-	HAEnabled                bool   `json:"ha_enabled"`
-	IsSelf                   bool   `json:"is_self"`
-	LeaderAddress            string `json:"leader_address"`
-	LeaderClusterAddress     string `json:"leader_cluster_address"`
-	PerfStandby              bool   `json:"performance_standby"`
-	PerfStandbyLastRemoteWAL uint64 `json:"performance_standby_last_remote_wal"`
-	LastWAL                  uint64 `json:"last_wal,omitempty"`
+	HAEnabled                bool      `json:"ha_enabled"`
+	IsSelf                   bool      `json:"is_self"`
+	ActiveTime               time.Time `json:"active_time"`
```
ncabatoff, line 57: Should we make this `omitempty`?

calvn (approved):
Looks good, pending Nick's suggestions.

http/sys_leader.go
briankassouf: I'll add it, but i'm not sure it will do anything since i'm not using a pointer.

PR #10491, opened 2020-12-03
No 'v' in version HTML anchor

Kloppi313:
The footer version output links to https://www.github.com/hashicorp/vault/blob/master/CHANGELOG.md#v160 (in Version 1.6.0) but you reach the anchor with https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#160 (without 'v' before the version number)


chelshaw:
Hi @Kloppi313 thanks for catching this and tackling it! 

Would you please fix the related test (`changelog-url-for: it builds an OSS URL`) and add a changelog file (see [this commit](https://github.com/hashicorp/vault/commit/bbe77aa0349c7ed84ea30aa4790af3f390c5462b#diff-29a21451ee6fc21e7d127ff247d3982522824bc5a39f872939fd263828fdb8b7) for reference). Thanks!

Kloppi313:
Hi @chelshaw,

I guess the changelog file has to be named after the pull request number.
Thanks for the tip with the corresponding check name!
Hopefully everything is ok now, as the checks are all green in this moment.

chelshaw (approved):
Looks great! Thanks again

PR #10492, opened 2020-12-03
Adjust executor sizes based on use case

ncabatoff:
Name environments after their use case rather than their size.  We need fewer resources for remote-docker because all the heavy work happens in the remote-docker VMs which we can't configure.  Use a different env for race tests which are more resource hungry.


vishalnayak (approved):
Neat!

PR #10493, opened 2020-12-03
Remove use of go cache from go test CI

ncabatoff:
Saving and restoring the go cache was an experiment to reduce build times, but it worked too inconsistently.


vishalnayak (approved):
Seems sane to me.

PR #10498, opened 2020-12-04
Be consistent with how we report init status.

ncabatoff:
This is the same fix as #9652, together with changing the seal-status api to behave the same as the health and init apis, i.e. use Core.Initialized to govern what we return as our initialized status.

### DIFF
--- a/vault/external_tests/raft/raft_test.go
+++ b/vault/external_tests/raft/raft_test.go
@@ -900,14 +900,10 @@ func TestRaft_Join_InitStatus(t *testing.T) {
 		vault.TestWaitActive(t, leaderCore.Core)
 	}
 
-	joinFunc := func(client *api.Client, addClientCerts bool) {
+	joinFunc := func(client *api.Client) {
 		req := &api.RaftJoinRequest{
 			LeaderAPIAddr: leaderAPI,
 			LeaderCACert:  string(cluster.CACertPEM),
-		}
-		if addClientCerts {
-			req.LeaderClientCert = string(cluster.CACertPEM)
-			req.LeaderClientKey = string(cluster.CAKeyPEM)
 		}
 		resp, err := client.Sys().RaftJoin(req)
 		if err != nil {
@@ -953,7 +949,7 @@ func TestRaft_Join_InitStatus(t *testing.T) {
 		verifyInitStatus(i, i < 1)
 	}
 
-	joinFunc(cluster.Cores[1].Client, false)
+	joinFunc(cluster.Cores[1].Client)
 	for i, core := range cluster.Cores {
 		verifyInitStatus(i, i < 2)
 		if i == 1 {
@@ -962,7 +958,7 @@ func TestRaft_Join_InitStatus(t *testing.T) {
 		}
 	}
 
-	joinFunc(cluster.Cores[2].Client, false)
+	joinFunc(cluster.Cores[2].Client)
 	for i, core := range cluster.Cores {
 		verifyInitStatus(i, true)
 		if i == 2 {
--- a/vault/external_tests/raft/raft_test.go
+++ b/vault/external_tests/raft/raft_test.go
@@ -955,7 +955,7 @@ func TestRaft_Join_InitStatus(t *testing.T) {
 
 	joinFunc(cluster.Cores[1].Client, false)
 	for i, core := range cluster.Cores {
-		verifyInitStatus(i, i < 1)
+		verifyInitStatus(i, i < 2)
 		if i == 1 {
 			cluster.UnsealCore(t, core)
 			time.Sleep(5 * time.Second)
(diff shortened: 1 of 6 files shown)


vault/external_tests/raft/raft_test.go
```diff
@@ -875,3 +875,102 @@ func BenchmarkRaft_SingleNode(b *testing.B) {
...
+	for i := range cluster.Cores {
+		verifyInitStatus(i, i < 1)
+	}
+
+	joinFunc(cluster.Cores[1].Client, false)
+	for i, core := range cluster.Cores {
+		verifyInitStatus(i, i < 2)
+		if i == 1 {
+			cluster.UnsealCore(t, core)
+			time.Sleep(5 * time.Second)
```
vishalnayak, line 961: I am sure you might have thought of not using sleep here. Would it be possible to use `RetryUntil` here?

vault/external_tests/raft/raft_test.go
```diff
@@ -875,3 +875,102 @@ func BenchmarkRaft_SingleNode(b *testing.B) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if health.Initialized != expected {
+			t.Errorf("core %d: expected init=%v, sys/health returned %v", coreIdx, expected, health.Initialized)
+		}
+	}
+
+	for i := range cluster.Cores {
+		verifyInitStatus(i, i < 1)
```
vishalnayak, line 953: The `<` logic is neat!

vault/external_tests/raft/raft_test.go
```diff
@@ -875,3 +875,102 @@ func BenchmarkRaft_SingleNode(b *testing.B) {
...
+			verifyInitStatus(i, true)
+		}
+	}
+
+	joinFunc(cluster.Cores[2].Client, false)
+	for i, core := range cluster.Cores {
+		verifyInitStatus(i, true)
+		if i == 2 {
+			cluster.UnsealCore(t, core)
+			time.Sleep(5 * time.Second)
```
vishalnayak, line 971: Here as well.

vishalnayak (approved):
Left, optional test comments. Otherwise LGTM!

ncabatoff:
Actually this seems to be incomplete.  Brian's comments in the jira:
>  I think it logically makes sense that a node that has joined a raft cluster but has yet to be unsealed is, for all intents and purposes, initialized

Prior to the unseal/sleep the init status appears as false, which doesn't fit with the above.

ncabatoff:
> Prior to the unseal/sleep the init status appears as false, which doesn't fit with the above.

Correction: it appears as true prior to unseal after the join, as well as after join+unseal, but in between it could report init=false.  This is when we've bootstrapped enough to set Core.raftInfo = nil, but haven't yet written the master key and other things Core.Initialized checks.  Fixed by checking for whether the *raft backend* is initialized in that case, which would happen before we set raftInfo=nil.

raskchanky (approved):

ncabatoff:
Fixes #9618.

vault/external_tests/raft/raft_test.go
```diff
@@ -875,3 +875,104 @@ func BenchmarkRaft_SingleNode(b *testing.B) {
...
+		verifyInitStatus(i, i < 2)
+		if i == 1 {
+			cluster.UnsealCore(t, core)
+			verifyInitStatus(i, true)
+		}
+	}
+
+	joinFunc(cluster.Cores[2].Client, false)
+	for i, core := range cluster.Cores {
+		verifyInitStatus(i, true)
```
vishalnayak, line 967: Shouldn't this be `i<3`?

ncabatoff: But i is always less than 3.

vault/external_tests/raft/raft_test.go
```diff
@@ -875,3 +875,104 @@ func BenchmarkRaft_SingleNode(b *testing.B) {
...
+
+	// Seal the leader so we can install an address provider
+	{
+		testhelpers.EnsureCoreSealed(t, leaderCore)
+		leaderCore.UnderlyingRawStorage.(*raft.RaftBackend).SetServerAddressProvider(addressProvider)
+		cluster.UnsealCore(t, leaderCore)
+		vault.TestWaitActive(t, leaderCore.Core)
+	}
+
+	joinFunc := func(client *api.Client, addClientCerts bool) {
```
vishalnayak, line 903: May be the join function can do away with the `addClientCerts` logic as we are only setting it to false.

vault/external_tests/raft/raft_test.go
vishalnayak: Ah, right.

vishalnayak (approved):

vault/external_tests/raft/raft_test.go
ncabatoff: Good call.

raskchanky (approved):

PR #10500, opened 2020-12-04
Add template helper library for use in generating custom usernames

pcman312:
# Description
Creates a new helper library in the SDK that allows users to generate text from go template strings. This was designed for custom username generation, but isn't limited to that scope.

# How does this work? Why not use `text/template` directly?
This wrapper was written because there were a number of common values or abilities that were needed in order to match the current username generation behavior encoded into Vault. That behavior is encoded into this library by using custom functions that are then passed to the `text/template` library. The custom functions can be found in `funcs.go`.

These are the custom functions built into this library:
* random (rand) - generates random base62 characters
* timestamp (now_seconds) - returns current unix timestamp in seconds
* now_nano - returns current unix timestamp in nanoseconds
* truncate (trunc) - truncates the provided string to a specific length
* truncate_sha256 (trunc_sha256) - truncates the provided string to a specific length and has the last 8 characters of a truncated string be the first 8 characters of the SHA256 hash of the removed portion of the string
* uppercase (upper) - uppercases the provided string
* lowercase (lower) - lowercases the provided string
* replace - string find/replaceme
* uuid - generates a random UUID
* sha256 - SHA256 hashes the provided string


Valarissa (approved):
Took me a while to grok this until I realized that it was integrating with text/template. Looks good. Great test coverage.

catsby (approved):
LGTM

austingebauer (approved):
LGTM

sdk/helper/template/funcs.go
```diff
@@ -0,0 +1,68 @@
...
+
+func uppercase(str string) string {
+	return strings.ToUpper(str)
+}
+
+func lowercase(str string) string {
+	return strings.ToLower(str)
+}
+
+func replace(find string, replace string, str string) string {
```
calvn, line 62: Nit: Can we have the signature ordering be the same as `strings.Replace`? That is, `str` (the input) becomes the first argument for this func. 

On a similar note, I'm curious why the other funcs like `truncate` take the string input as its last argument.

sdk/helper/template/template.go
```diff
@@ -0,0 +1,140 @@
...
+//     Example: {{ .DisplayName | replace - _ }}
+// - sha256
+//   - SHA256 hashes the previous value.
+//     Example: {{ .DisplayName | sha256 }}
+// - timestamp (now_seconds)
+//   - Provides the current unix time in seconds.
+//     Example: {{ now_seconds}}
+// - now_nano
+//   - Provides the current unix time in nanoseconds.
+//     Example: {{ now_nano}}
```
calvn, line 67:
```suggestion
//     Example: {{ now_nano }}
```

sdk/helper/template/template.go
```diff
@@ -0,0 +1,140 @@
...
+//     Example: {{ .DisplayName | lowercase }}
+// - replace
+//   - Performs a string find & replace
+//     Example: {{ .DisplayName | replace - _ }}
+// - sha256
+//   - SHA256 hashes the previous value.
+//     Example: {{ .DisplayName | sha256 }}
+// - timestamp (now_seconds)
+//   - Provides the current unix time in seconds.
+//     Example: {{ now_seconds}}
```
calvn, line 64:
```suggestion
//     Example: {{ now_seconds }}
```

sdk/helper/template/template.go
```diff
@@ -0,0 +1,140 @@
...
+func NewTemplate(opts ...Opt) (up StringTemplate, err error) {
+	up = StringTemplate{
+		funcMap: map[string]interface{}{
+			"random":          base62.Random,
+			"rand":            base62.Random,
+			"truncate":        truncate,
+			"trunc":           truncate,
+			"truncate_sha256": truncateSHA256,
+			"trunc_sha256":    truncateSHA256,
+			"uppercase":       uppercase,
```
calvn, line 89: I wonder whether it's necessary to expose two versions of the same func. There's extensive documentation on the godoc, and there will likely be just as much documentation on our site on their usage as well.

sdk/helper/template/template.go
```diff
@@ -0,0 +1,140 @@
...
+			"upper":           uppercase,
+			"lowercase":       lowercase,
+			"lower":           lowercase,
+			"replace":         replace,
+			"sha256":          hashSHA256,
+
+			"timestamp":   nowSeconds,
+			"now_seconds": nowSeconds,
```
calvn, line 97: `now_seconds` tells us about the precision of this value, but not the formatting. Maybe `unix_seconds`?

sdk/helper/template/template.go
```diff
@@ -0,0 +1,140 @@
...
+			"truncate_sha256": truncateSHA256,
+			"trunc_sha256":    truncateSHA256,
+			"uppercase":       uppercase,
+			"upper":           uppercase,
+			"lowercase":       lowercase,
+			"lower":           lowercase,
+			"replace":         replace,
+			"sha256":          hashSHA256,
+
+			"timestamp":   nowSeconds,
```
calvn, line 96: When I see timestamp, I think of a format like this `1970-01-01T00:00:00Z`. Having said that, the dashes and colons could be a problem for some username values.

calvn (commented):
Looks great overall! Left a few comments/suggestions.

sdk/helper/template/funcs.go
pcman312: Unfortunately no. The way that the template library passes arguments using `|` seems to require the value being passed to it to be at the end of the argument list.

sdk/helper/template/template.go
pcman312: If I were to choose one I'd rather keep the more explicit one. I originally did this so "power users" could use shortened templates but I'm open to not having duplicates as well.

sdk/helper/template/template_test.go
```diff
@@ -0,0 +1,226 @@
...
+			require.NoError(t, err)
+
+			require.Regexp(t, fmt.Sprintf("^[a-zA-Z0-9]{%d}$", i), actual)
+		}
+	})
+
+	t.Run("rand", func(t *testing.T) {
+		for i := 1; i < 100; i++ {
+			st, err := NewTemplate(
+				Template(fmt.Sprintf("{{random %d}}", i)),
```
calvn, line 118:
```suggestion
				Template(fmt.Sprintf("{{rand %d}}", i)),
```

sdk/helper/template/funcs.go
calvn: Ah, I recall this from past experience now that you mention it.

sdk/helper/template/template.go
calvn: I do see your point of making things easier for power users, but I also like the idea of sticking with one way to do the same thing even if it ends up being a bit longer.

sdk/helper/template/template.go
calvn: Now that I think about it, could we remove this since `now_seconds` is essentially doing the same thing?

Issue #10504, opened 2020-12-05
Split away code for decoding root tokens into a function


ptzianos:
**Is your feature request related to a problem? Please describe.**
When using the API client to generate a temporary root token using the unseal keys, there is no way of easily decoding the root token that is returned. The piece of code that performs this function is inside the `command/operator_generate_root.go` file and can't be re-used.

**Describe the solution you'd like**
Split away the code into a function that can be re-used by anyone who wants to create a temp root token using the API client

PayalSasmal10:
can we decode the root token from this code using API?

ptzianos:
@PayalSasmal10 the workflow that I had in mind (and that I implemented based on this) is that you get a new token via the API and then decode it using the helper function. so in pseudo-Go it looks like this:
```
var tempRootToken string
for i, unsealKey := range unsealKeys {
	resp, err := client.Sys().GenerateRootUpdate(string(unsealKey[:]), rootInitResp.OTP)
	if resp.Complete {
		tempRootToken, err = token.DecodeRootToken(resp.EncodedRootToken, rootInitResp.OTP, rootInitResp.OTPLength)
		break
	}
}
```

PayalSasmal10:
Thanks @ptzianos . Waiting for merging this commit.

PR #10505, opened 2020-12-05
Add helper for encoding/decoding root tokens and OTP generation in SDK module (#10504)

ptzianos:
Fixes: #10504

Changes:
- Splits away code that encodes/decodes root tokens into a helper function in the SDK module
- Splits away code that generates the OTP for the root token encoding into a helper function in the SDK module
- Moves pgpkeys packages from vault/helpers to sdk/helpers
Adds:
- Unit tests for root token encoding/decoding and OTP generation


command/operator_generate_root.go
```diff
@@ -341,36 +340,9 @@ func (c *OperatorGenerateRootCommand) decode(client *api.Client, encoded, otp st
...
-
-		tokenBytes, err = xor.XORBytes(tokenBytes, []byte(otp))
-		if err != nil {
-			c.UI.Error(errwrap.Wrapf("Error xoring token: {{err}}", err).Error())
-			return 1
-		}
-		token = string(tokenBytes)
+	token, err := token.DecodeRootToken(encoded, otp, status.OTPLength)
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error decoding root token: %s", err))
```
ncabatoff, line 345: Looks like you're missing the `return 1` here.

helper/token/decode.go
```diff
@@ -0,0 +1,39 @@
+package token
```
ncabatoff, line 1: I suggest making this helper part of the sdk instead, since I imagine you want to make use of this outside of Vault.

ptzianos: Yeah that's a good point. I'll rework the code

ptzianos:
@ncabatoff hi, I'm back at this, I added all the fixes you suggested. Unfortunately I'm getting some test errors from the rabbitmq part of the code. I will try to debug them today. Is there anything else that this PR needs?

ncabatoff:
> @ncabatoff hi, I'm back at this, I added all the fixes you suggested. Unfortunately I'm getting some test errors from the rabbitmq part of the code. I will try to debug them today. 

Merging in the latest main should fix those errors.

> Is there anything else that this PR needs?

A unit test would be nice, but not essential since this is tested indirectly by operator_generate_root_test.go.

ptzianos:
@ncabatoff I have some good news and some not so good news. The good news is that after a lot of different things I got a proper test suite going. The bad news is that I had to move a lot more stuff around and the PR has increased in size. I moved the code that does the root token encoding + the code that generates the OTP into the same package as the decoding algorithm. And I also had to move the PGP key helper into the sdk module to ensure that there are no dependencies from the SDK module to the Vault module. But things are unit tested now, so yay?

sdk/helper/tokenutil/root/encode.go
```diff
@@ -0,0 +1,48 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/hashicorp/vault/sdk/helper/pgpkeys"
+	"github.com/hashicorp/vault/sdk/helper/xor"
+)
+
+var (
+	// ErrNoTokenProvided is returned when an empty string is provided as an error
+	ErrNoTokenProvided = errors.New("no token provided")
```
ncabatoff, line 14: We don't usually define error types unless we have to, i.e. because some code elsewhere needs to match the error and do something specific when that error is returned.

sdk/helper/tokenutil/root/encode.go
```diff
@@ -0,0 +1,48 @@
...
+// EncodeToken gets a token and an OTP or a PGP Key and encodes it. If no OTP or PGP key is
+// provided, an error will be returned. If an OTP is provided, it must have the same length as the token
+func EncodeToken(token, otp, pgpKey string, cleanupFunc func()) (string, error) {
+	if len(token) == 0 {
+		cleanupFunc()
+		return "", ErrNoTokenProvided
+	}
+	// Get the encoded value first so that if there is an error we don't create
+	// the root token.
+	if len(otp) > 0 {
```
ncabatoff, line 28: Not sure why you replaced the original switch statement with this if-else statement, but we tend to prefer the former; many Go coders eschew `else` altogether in favour of switches and early returns.

sdk/helper/tokenutil/root/otp.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+
+	"github.com/hashicorp/go-secure-stdlib/base62"
+)
+
+// DefaultBase64EncodedOTPLength is the number of characters that will be randonmly generated
```
ncabatoff, line 11:
```suggestion
// DefaultBase64EncodedOTPLength is the number of characters that will be randomly generated
```

sdk/helper/tokenutil/root/otp.go
```diff
@@ -0,0 +1,39 @@
...
+// before the Base64 encoding process takes place.
+const DefaultBase64EncodedOTPLength = 16
```
ncabatoff, line 13: Since this is only used internally within this package, maybe make it private?

sdk/helper/tokenutil/root/encode.go
```diff
@@ -0,0 +1,48 @@
...
+var (
+	// ErrNoTokenProvided is returned when an empty string is provided as an error
+	ErrNoTokenProvided = errors.New("no token provided")
+	// ErrNoOTPorPGPKeyProvided is returned when a non-empty token is provided but not OTP or PGP key
+	ErrNoOTPorPGPKeyProvided = errors.New("no otp or pgp key provided")
+)
+
+// EncodeToken gets a token and an OTP or a PGP Key and encodes it. If no OTP or PGP key is
+// provided, an error will be returned. If an OTP is provided, it must have the same length as the token
+func EncodeToken(token, otp, pgpKey string, cleanupFunc func()) (string, error) {
```
ncabatoff, line 21: Everything else in this patch is really about the otp, it's only this func that knows a tiny bit about encrypting the token using PGP.  I think we should narrow the focus to just the otp part for now.

ncabatoff (commented):
Getting closer!  I haven't looked at the tests because I expect they'll likely change as a result of the other changes I've requested.

api/go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/frankban/quicktest v1.13.0 // indirect
 	github.com/go-test/deep v1.0.2
 	github.com/hashicorp/errwrap v1.1.0
-	github.com/hashicorp/go-cleanhttp v0.5.1
+	github.com/hashicorp/go-cleanhttp v0.5.2
```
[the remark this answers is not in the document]
ptzianos: I'm not sure why but this was causing one of the preflight checks to fail because the go mod download automatically upgraded to the latest patch version (0.5.2) and then the next check was finding that the go.mod file was modified and failed. So I just updated it and tests started passing!?

vault/generate_root.go
```diff
@@ -318,34 +317,9 @@ func (c *Core) GenerateRootUpdate(ctx context.Context, key []byte, nonce string,
...
-			cleanupFunc()
-			c.logger.Error("error encrypting new root token", "error", err)
-			return nil, err
-		}
-		token = base64.StdEncoding.EncodeToString(tokenBytesArr[0])
-
-	default:
-		cleanupFunc()
-		return nil, fmt.Errorf("unreachable condition")
+	token, err = root.EncodeToken(token, c.generateRootConfig.OTP, c.generateRootConfig.PGPKey, cleanupFunc)
```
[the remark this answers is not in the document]
ptzianos: I'm guessing that it was called because the original switch returned immediately. will remove now from EncodeToken

sdk/helper/tokenutil/root/encode.go
ptzianos: I defined here the errors so that I can compare them more accurately in the tests, but I don't have a strong opinion on this. Will change

sdk/helper/tokenutil/root/encode.go
ptzianos: I'm not a huge fan of switch statements tbh. Out of curiosity I checked my company's codebase which is 90% Go and there isn't much love for switches in it either. OTOH I don't have a strong opinion on this either so I will switch to switch (pun intended)

sdk/helper/tokenutil/root/otp.go
ptzianos: Good catch

sdk/helper/tokenutil/root/encode.go
ptzianos: You mean split away the PGP encryption part back into the generate root portion of the Vault module?

ptzianos:
@ncabatoff all comments have been fixed except one. Ready for review round 2

sdk/helper/tokenutil/root/encode.go
ncabatoff: Fair, I wouldn't normally read someone's PR and say "could you turn this if/else into a switch", the only reason I commented here was because you were turning an existing switch into an if/else.  No big deal either way.

sdk/helper/tokenutil/root/encode.go
ncabatoff: Right.  Just handle one aspect of token encoding here.

ptzianos: Fixed. I suppose that I should move the pgpkeys package where it was originally, right?

ncabatoff: Yes please.  It probably makes sense to move it as you've done, but I'd rather that be in its own PR if we do it.

api/go.mod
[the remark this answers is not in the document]
ncabatoff: Looks like we're on 0.5.2 on main already, so I guess you weren't alone.

changelog/10505.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+helper: Add helper for decoding root tokens
```
ncabatoff, line 2:
```suggestion
sdk: Add helper for decoding root tokens
```

ncabatoff:
> @ncabatoff all comments have been fixed except one. Ready for review round 2

I've re-reviewed and I'm ready to approve and merge after you revert the pgp thing.  Also apply the little tweak to the changelog I suggested please.

changelog/10505.txt
ptzianos: fixed

ptzianos:
> > @ncabatoff all comments have been fixed except one. Ready for review round 2
> 
> I've re-reviewed and I'm ready to approve and merge after you revert the pgp thing. Also apply the little tweak to the changelog I suggested please.

Awesome. I just fixed the changelog and moved back the PGP keys. Waiting for tests to pass :crossed_fingers: . Would you like me to squash the commits?

ncabatoff:
> Would you like me to squash the commits?

No need, when I merge it to main I'll be squashing them anyway.

ptzianos:
@ncabatoff thank you very much for your help getting this PR ready for merging. Can I interest you now in a second PR? https://github.com/hashicorp/vault/pull/10464

ptzianos:
@ncabatoff can you :heavy_check_mark: this PR now that we are done with the changes or do we need to get an approval from someone else?

ncabatoff:
Thanks for all your hard work @ptzianos !

PR #10511, opened 2020-12-08
Fix UI custom header values

michelvocks:
`sys/config/ui/headers` (https://www.vaultproject.io/api-docs/system/config-ui#configure-ui-headers) allows you to set multiple custom header values per custom header. 

However, all values are currently ignored except the first since `header.Get` returns only the first value:
```go
// Get gets the first value associated with the given key. If
// there are no values associated with the key, Get returns "".
// It is case insensitive; textproto.CanonicalMIMEHeaderKey is
// used to canonicalize the provided key. To use non-canonical keys,
// access the map directly.
func (h Header) Get(key string) string {
	return textproto.MIMEHeader(h).Get(key)
}
```

This PR fixes the issue and also changes the read API endpoint to return a slice of values instead of one single value.


vault/logical_system_paths.go
```diff
@@ -60,7 +59,7 @@ func (b *SystemBackend) configPaths() []*framework.Path {
 		},
 
 		{
-			Pattern: "config/ui/headers/" + framework.GenericNameRegex("header"),
+			Pattern: "config/ui/headers/" + framework.GenericNameRegex("header") + framework.OptionalParamRegex("multivalue"),
```
ncabatoff, line 62: Wouldn't it be more natural to make this an HTTP param (i.e. `?multivalue=true`) rather than making it part of the path?

michelvocks: Sounds good. How do I do that? Couldn't find an example in the code.

ncabatoff: I believe you simply leave it out of the Pattern field, and define it as a field such as you have below.

ncabatoff: Found an example: https://www.vaultproject.io/api-docs/system/pprof#profile

michelvocks: Thanks! That worked.

vault/logical_system_paths.go
```diff
@@ -72,7 +72,7 @@ func (b *SystemBackend) configPaths() []*framework.Path {
 				},
 				"multivalue": &framework.FieldSchema{
 					Type:        framework.TypeString,
```
ncabatoff, line 74: Any reason not to use TypeBool?

michelvocks: No! It was late 😄  Thanks!

vault/logical_system.go
```diff
@@ -2432,18 +2432,28 @@ func (b *SystemBackend) handleDisableAudit(ctx context.Context, req *logical.Req
...
 	header := data.Get("header").(string)
+	multivalue := data.Get("multivalue").(bool)
 
-	value, err := b.Core.uiConfig.GetHeader(ctx, header)
+	values, err := b.Core.uiConfig.GetHeader(ctx, header)
 	if err != nil {
 		return nil, err
 	}
-	if value == "" {
+	if len(values) == 0 {
```
ncabatoff, line 2441: Is this a behaviour change?  Do we return something different in the case where a single empty header has been set?

michelvocks: No. The conversation library makes sure that `values` from type `TypeStringSlice` will always return an empty slice if the value is "".

ncabatoff (approved):

PR #10514, opened 2020-12-08
metrics: activity log

alexanderbez:


alexanderbez:
@mgritter I've added `vault.identity.entity.active.monthly` and `vault.identity.entity.active.reporting_period`. I haven't dug too much into `vault.identity.entity.active.partial_month`, but I recall you stating that one might not be possible? I can't recall.

I'm marking this ready-for-review just to get some feedback if these are what you expected.

vault/activity_log.go
```diff
@@ -1627,12 +1629,23 @@ func (a *ActivityLog) precomputedQueryWorker() error {
...
 		}
+
 		for nsID, counts := range byNamespace {
 			pq.Namespaces = append(pq.Namespaces, &activity.NamespaceRecord{
 				NamespaceID:     nsID,
 				Entities:        uint64(len(counts.Entities)),
 				NonEntityTokens: counts.Tokens,
 			})
+
+			if startTime.After(activePeriodStart) {
```
mgritter, line 1640: This could fire for every previous month in our retention period, not just once at defaultReportMonths.

If you use the utility function suggested above, that should allow an equality check instead... except for the potential partial month (if the feature was enabled on Jan 15, then that first startTime will be Jan 15, not Jan 1.)

So, I would suggest using timeutil.InRange( startTime, activePeriodStart, EndOfMonth( activePeriodStart )) or writing a new utility in timeutil that does that.

vault/activity_log.go
```diff
@@ -1641,9 +1654,20 @@ func (a *ActivityLog) precomputedQueryWorker() error {
...
 	}
 
-	// Delete the intent log
+	// delete the intent log
 	a.view.Delete(ctx, activityIntentLogKey)
 
+	// emit active entity metrics for the interval (times[0], endTime)
+	for nsID, counts := range byNamespace {
+		a.metrics.SetGaugeWithLabels(
+			[]string{"identity", "entity", "active", "monthly"},
```
mgritter, line 1663: The comment is correct but I don't believe that is what this code does

Because we've finished the loop, byNamespace contains all the data from the interval (times[len(times)-1], endTime).  times is in reverse order with 0 being the most recent month.

A check at the same place you're doing the defaultReportingPeriod check would be sufficient.

mgritter (changes requested):

mgritter:
> @mgritter I've added `vault.identity.entity.active.monthly` and `vault.identity.entity.active.reporting_period`. I haven't dug too much into `vault.identity.entity.active.partial_month`, but I recall you stating that one might not be possible? I can't recall.

`partial_month` is easy without a namespace breakdown, it is just the size of activeEntities (once fully reloaded, might be multiple storage entries.)  To do a breakdown by namespace like you did with these two means we would have to walk the segments on storage where that information lives (whenever we emit the metrics) or keep the per-namespace breakdown in memory all the time.  Both are feasible, but maybe we don't want to touch storage every 10 minutes so a longer interval would be good.

The storage size metric is the one I thought might be infeasible.

website/pages/docs/internals/telemetry.mdx
```diff
@@ -162,6 +162,9 @@ These metrics cover measurement of token, identity, and lease operations, and co
 | `vault.expire.register`                   | Time taken for register operations                                          | ms     | summary |
 | `vault.expire.register-auth`              | Time taken for register authentication operations which create lease entries without lease ID           | ms       | summary |
 | `vault.identity.num_entities`             | Number of identity entities stored in Vault                                                             | entities | gauge   |
+| `vault.identity.entity.active.monthly` (cluster, namespace) | Number of distinct entities that created a token during the past month, per namespace. Only available if client count is enabled. Reported at the start of each month. | entities | gauge   |
```
swayne275, line 165:
```suggestion
| `vault.identity.entity.active.monthly` (cluster, namespace) | Number of distinct entities that created a token during the past month, per namespace. Only available if client count is enabled. Reported at the start of each month. | entities | gauge |
```

website/pages/docs/internals/telemetry.mdx
```diff
@@ -162,6 +162,9 @@ These metrics cover measurement of token, identity, and lease operations, and co
...
+| `vault.identity.entity.active.partial_month` (cluster) | Total number of distinct entities that created a token during the current month. Only available if client count is enabled. Reported periodically within each month. | entities | gauge |
+| `vault.identity.entity.active.reporting_period` (cluster, namespace) | Number of distinct entities that created a token in the past N months, as defined by the client count default reporting period. Only available if client count is enabled. Reported at the start of each month. | entities | gauge   |
```
swayne275, line 167:
```suggestion
| `vault.identity.entity.active.reporting_period` (cluster, namespace) | Number of distinct entities that created a token in the past N months, as defined by the client count default reporting period. Only available if client count is enabled. Reported at the start of each month. | entities | gauge |
```

swayne275 (approved):
Approved but those two suggestions look in line with the other documentation

website/pages/docs/internals/telemetry.mdx
mgritter: This file got moved so I can't apply the suggestion as-is, I tried to make the spacing consistent in the new file.

mgritter (approved):

PR #10516, opened 2020-12-09
Disclosure of internal IP addresses from unauthenticated request

bruj0:
Removing extra information from the returned error, to avoid leaking it to unauthenticated requests


sgmiller (approved):

mladlow:
@bruj0 Should this have a changelog entry?

bruj0:
> @bruj0 Should this have a changelog entry?

I dont think its a visible change, its only triggered by sending the wrong value in "Content-Lenght" which is usually calculated automatically by the HTTP client.
 So to trigger it you have to do it with a reason (we got the report from a 
 customer running an automated vulnerability scanner) or its in bug in the client.

PR #10517, opened 2020-12-09
Fixing the Curl example for restoring Raft snapshot.mdx

martinhristov90:
Fixing the Curl example for restoring Raft snapshot


vishalnayak:
After pulling in changes from main branch, the diff looked quite different from what was intended since the files had moved around. I figured it was easier to do a new PR to cover this. Closing in favor of https://github.com/hashicorp/vault/pull/11489.

Thanks!

PR #10520, opened 2020-12-09
Send a test message before committing a new audit device.

mgritter:
Also, lower timeout on connection attempts in socket device.

Some history here:
1. I first tried just sending a normal message.  This doesn't work because the Salt is not yet initialized, and it can't be initialized until the storage view is readable, and it's set to read-only during construction. (Probably for replication-related reasons, although it's not clear whether that is just copying what secret mounts, or is actually necessary here.)
2. Then I tried doing the check in the backend Factory methods, like the file backend does.  I thought this would be a bad idea, because those checks would then get re-run every time Vault unsealed and recreated the audit devices. Worse, if they failed, then that audit device would be temporarily removed until Vault went through another unseal-- even if it was weeks.
3. So, I added a new method specifically for testing, which creates a Salt that is used once and discarded.

The test message looks like this:

```
{"time":"2020-12-09T20:14:22.451152342Z","type":"request","auth":{"token_type":"default"},"request":{"id":"3ca7b3cc-7581-77c7-8610-b80b6284e43f","operation":"update","namespace":{"id":"root"},"path":"sys/audit/test"}}
```


audit/format.go
```diff
@@ -434,3 +434,25 @@ func parseVaultTokenFromJWT(token string) *string {
 
 	return &claims.ID
 }
+
+// Create a formatter not backed by a persistent salt.
+func NewTemporaryFormatter(format string, prefix string) *AuditFormatter {
```
swayne275, line 439:
```suggestion
func NewTemporaryFormatter(format, prefix string) *AuditFormatter {
```

total nit, but the go-y thing to do

swayne275, line 439:
```suggestion
func NewTemporaryFormatter(format, prefix string) *AuditFormatter {
```

total nit, but the "go-y" thing to do

builtin/audit/file/backend.go
```diff
@@ -258,6 +258,24 @@ func (b *Backend) LogResponse(ctx context.Context, in *logical.LogInput) error {
...
 }
 
+func (b *Backend) LogTestMessage(ctx context.Context, in *logical.LogInput, config map[string]string) error {
+	var writer io.Writer
+	switch b.path {
+	case "stdout":
+		writer = os.Stdout
+	case "discard":
+		return nil
+	}
```
swayne275, line 268: should there be a default case? otherwise i'd think `writer` could be nil?

builtin/audit/socket/backend.go
```diff
@@ -177,6 +177,30 @@ func (b *Backend) LogResponse(ctx context.Context, in *logical.LogInput) error {
...
+	defer b.Unlock()
+
+	err := b.write(ctx, buf.Bytes())
+	if err != nil {
+		rErr := b.reconnect(ctx)
+		if rErr != nil {
+			err = multierror.Append(err, rErr)
+		} else {
+			// Try once more after reconnecting
+			err = b.write(ctx, buf.Bytes())
```
swayne275, line 197: is it possible for the two write errors to be different? if so, maybe we also want to `multierror.Append()`?

sdk/helper/salt/salt.go
```diff
@@ -115,6 +115,23 @@ func NewSalt(ctx context.Context, view logical.Storage, config *Config) (*Salt,
...
+	config := &Config{}
+	config.Location = ""
+	config.HashFunc = SHA256Hash
+	config.HMAC = sha256.New
+	config.HMACType = "hmac-sha256"
+
+	s := &Salt{
+		config: config,
+	}
+	s.salt, _ = uuid.GenerateUUID()
```
swayne275, line 130: is there a scenario where we care about the `GenerateUUID()` err?

vendor/github.com/hashicorp/vault/api/sys_config_cors.go
```diff
@@ -35,63 +35,31 @@ func (c *Sys) CORSStatus() (*CORSResponse, error) {
 	return &result, err
 }
 
-func (c *Sys) ConfigureCORS(req *CORSRequest) (*CORSResponse, error) {
+func (c *Sys) ConfigureCORS(req *CORSRequest) error {
```
swayne275, line 38: I'm confused about two things in this file
1) why are we editing in the vendor folder? i'm not totally sure how that's used here, but in my previous experiences the vendor folder was brought in by imports, so do we want to maybe change the source rather than the imported code (if this is true here)? Either way, I'd love to know the rationale here
2) what was going on with all the secrets logic before? I guess why was that there then, and why can it be removed now?

vendor/github.com/hashicorp/vault/sdk/helper/salt/salt.go
```diff
@@ -115,6 +115,23 @@ func NewSalt(ctx context.Context, view logical.Storage, config *Config) (*Salt,
...
+	config := &Config{}
+	config.Location = ""
+	config.HashFunc = SHA256Hash
+	config.HMAC = sha256.New
+	config.HMACType = "hmac-sha256"
+
+	s := &Salt{
+		config: config,
+	}
+	s.salt, _ = uuid.GenerateUUID()
```
swayne275, line 130: same couple things here (vendor folder, and do we care about the `GenerateUUID()` return error?

swayne275 (commented):
Looks good overall! Just a few comments and things I'd like to know.

builtin/audit/file/backend.go
mgritter: writer can be nil, but confusingly for this backend the write function uses open() when it is

sdk/helper/salt/salt.go
mgritter: I thought about that but we really don't care what the salt is.  The worst case is we end up with "" and so the hash is predictable, but we don't actually hash anything in the test message? For now, anyway.

vendor/github.com/hashicorp/vault/api/sys_config_cors.go
mgritter: This was a change previously made to api that was pulled in when I did "go mod vendor", because the previous commit did not do "go mod vendor" as it should have.  These are not my changes.

For the api and sdk packages, "go mod vendor" just does a copy from api to vendor/github.com/hashicorp/vault/api etc.

builtin/audit/socket/backend.go
mgritter: Good question. I think the last error is really the only one that really needs to be reported, so by preference I wouldn't use multierror at all.  (The existing code does and so we see "connection timed out" or similar errors repeated twice; I just copied the logic from the normal LogRequest/LogResponse functions.)

I think if we successfully reconnected then the original error is not interesting to report, while *maybe* a case could be made that if the write failed and the reconnect failed then they might usefully report different information.

swayne275 (approved):

PR #10530, opened 2020-12-10
minor formatting

assareh:


website/pages/docs/concepts/policies.mdx
```diff
@@ -144,7 +144,7 @@ path "secret/bar/*" {
 }
 
 # Permit reading everything prefixed with "zip-". An attached token could read
-# "secret/zip-zap" or "secret/zip-zap/zong", but not "secret/zip/zap
+# "secret/zip-zap" or "secret/zip-zap/zong", but not "secret/zip/zap".
```
kitography, line 147: Nice catch!

website/pages/docs/concepts/policies.mdx
```diff
@@ -154,7 +154,7 @@ In addition, a `+` can be used to denote any number of characters bounded
 within a single path segment (this appeared in Vault 1.1):
 
 ```ruby
-# Permit reading the "teamb" path under any top-level path under secret/
+# Permit reading the "teamb" path under any top-level path under secret/.
```
kitography, line 157: I think this is more confusing, and would not merge it.  "." in the context of a path means a reference to the folder itself, so I would read this (with the path listed below) as "secret/secret/+/teamb/" if the "." is there.

kitography, line 157: I think this is more confusing, and would not merge it.  "." in the context of a path means a reference to the folder itself, so I would read this (with the path listed below) as "secret/secret/+/teamb/" if the "." is there.

kitography (changes requested):
I think the first of these changes should definitely be merged, and the second really shouldn't be.  Could we limit this to the change on line 147?

PR #10533, opened 2020-12-11
Bug Fix: Unable to click on icon label in radio-box component to select

Monkeychip:
Because we're now in Glimmer land and this is not a crucial fix that needs to be backported, I went ahead and transformed the box-radio component to a glimmer component, and in set us up for decorators later on.  The package json was updated for decorators, a higher babel and lint version as well as sinon.  

This fixes the issue on auth and secrets lists.

This was a bit of a strange approach, made more complicated by the fact I did this as a glimmer component.  Putting as draft PR for feedback. 

TODO:

- [ ] fix the broken test.  The functionality is working but after some attempts at getting the sinon spy to work, I couldn't.  Will revisit as this will be a problem for all future glimmer components tested with sinon spys.
- [ ] test disabled box-radios.
- [ ] test anywhere else box-radio is being used or the mount-backend component is being used.
- [ ] fix styling now that removed the label.

With fix.
box-radio-fix


ui/lib/core/addon/templates/components/box-radio.hbs
```diff
@@ -1,66 +1,68 @@
...
     as |T|>
     <T.trigger @tabindex="-1">
       <label
-        for={{type}}
+        for={{@type}}
         class="box-radio is-disabled is-marginless"
         data-test-mount-type-radio
-        data-test-mount-type={{type}}
+        data-test-mount-type={{@type}}
+        {{on 'click' (fn this.handleClick @onRadioChange @type)}}
```
chelshaw, line 13: 👏 So excited we can use this now

chelshaw: I think we could clean this up (and make the component file more obvious for how this works) by passing only `@type` into this.handleClick (and I'll comment the corresponding changes on the js file)

ui/lib/core/addon/components/box-radio.js
```diff
@@ -16,11 +16,18 @@
...
-  layout,
-  disabled: false,
-  tooltipMessage: 'This option is not available to you at this time.',
-});
+class BoxRadio extends Component {
+  disabled = false;
+  tooltipMessage = 'This option is not available to you at this time.';
+  onRadioChange = '';
+
+  handleClick(method, value) {
```
chelshaw, line 28: Continuing my comment from the hbs file, I was initially very confused what onRadioChange was for, and I think simplifying `handleClick` to only take value, and then call `this.onRadioChange` may help. Or, we could skip `handleClick` alltogether and do:

```js
{{on "click" (fn @onRadioChange @type)}}
```
on the template, which eliminates the need for handleClick altogether, especially since there are no side effects we need to handle within the component. This would potentially require onRadioChange being set to a sane default...or maybe it's ok to fail if it's not passed in because not passing it kind of defeats the purpose of this component. 

Sorry for the long rambling thoughts, let me know if any of this doesn't make sense.

ui/tests/integration/components/box-radio-test.js
```diff
@@ -1,23 +1,25 @@
...
-    this.set('displayName', 'An Option');
-    this.set('mountType', '');
-    this.set('disabled', false);
+    this.type = 'aws';
+    this.displayName = 'An Option';
+    this.mountType = '';
+    this.disabled = false;
   });
 
+  const spy = sinon.spy();
```
chelshaw, line 17: Pretty sure this needs to live within the `test` fn. But i know you're still working on this :)

chelshaw (changes requested):
Wooohoooo glimmer!! ✨

Monkeychip:
Issues with package json and testing, easier to just start with new barnch.  Locally couldn't get the test to past.

PR #10536, opened 2020-12-11
rate limit: fix initialize defaults

alexanderbez:
When coming from 1.5.x to 1.6.x, rate limit resource quotas are incompletely initialized. In partciular, the underling rate limiter was swapped between these two versions and the current one (1.6.x) requires fields that when coming from 1.5.x, do not exist. Given these fields are never mutated or provided by an operator or client, the solution is to just set them to the default(s).


alexanderbez:
Note @vishalnayak, I do not think you can get this to panic unless you go from 1.6.x -> 1.5.x -> 1.6.x. This is because `block_interval` needs to be defined in order to trigger the panic and that only happens in 1.6.x.

vishalnayak (approved):

vishalnayak:
Tested using the following script: https://github.com/hashicorp/vault-tools/blob/master/users/vishal/quotas/rateLimitPanic.sh

vishalnayak:
Note that we didn't have to restart Vault to see the panic. If we just updated a rate limit quota that existed in 1.5.5, that was enough to trigger it. This is because, even the update operation will walk over the `initialize` function.

vishalnayak:
@alexanderbez You might want to add changelog file to this PR.

alexanderbez:
Added a CL entry @vishalnayak

mladlow (changes requested):
This should have a changelog file, not a direct edit to changelog.md

mladlow (approved):
Approving - my approval only covers the changelog though

alexanderbez:
Backport PR: https://github.com/hashicorp/vault/pull/10565

PR #10557, opened 2020-12-14
mod: update auth-jwt to v0.8.1

calvn:
Updates vault-plugin-auth-jwt to v0.8.1

Steps taken:
- `get github.com/hashicorp/vault-plugin-auth-jwt@release/vault-1.6.x`
- `go mod vendor && go mod tidy`

PR #10561, opened 2020-12-14
Adds changelog file for auth-jwt bug fix

austingebauer:
This PR adds a changelog file into `release/1.6.x` for the bug fix introduced in https://github.com/hashicorp/vault/pull/10546. The source for the fix is in an external plugin repository, so the changelog file isn't a part of the 1.6.x backport PR https://github.com/hashicorp/vault/pull/10557.


calvn:
Since this is a fix from an external PR, you can prefix the filename with `_`, and use the original PR's number. In this case, the filename should be `_145.txt`.

calvn (approved):

PR #10572, opened 2020-12-14
Bug: switching between namespaces using the namespace-link caused the model not to refresh

Monkeychip:
*Should not be backported because it's a bug due to Ember Upgrade released in 1.7*

Before the ember upgrade, we used a method in the `cluster.js` route file that got called on the `beforeModel` hook, which searched for existing models and then removed those that should no longer be there.  This was most relevant for when we used the `namespace-link` to switch between namespaces.  [See code here](https://github.com/hashicorp/vault/pull/9972/files#diff-9ee5153ee66d63eb6fb0f6724622c5f0ee4430dfcbb76a9ffc16eb675ddb8a77L39).  

However the functionality used, specifically the get `.getModelTypes()` method is deprecated.  So in order to refresh the model we have to refresh the page route.  This will change the functionality / UX some in that when the user uses the `namespace-link` they will be redirected to the secrets page of the namespace instead of staying on the same route. 

Additionally, this PR replaces the `LinkTo` functionality in the Namespace switcher under manage namespaces.  It calls the same component so it also redirects to the secrets list.


ui/app/components/namespace-link.js
```diff
@@ -30,4 +31,23 @@ export default Component.extend({
...
+
+    let origin =
+      window.location.protocol +
+      '//' +
+      window.location.hostname +
+      (window.location.port ? ':' + window.location.port : '');
+
+    if (!this.normalizedNamespace) return `${origin}/ui/vault/secrets`;
+
+    return `${origin}/ui/vault/secrets?namespace=${encodeURIComponent(this.normalizedNamespace)}`;
```
chelshaw, line 51: Maybe add a note here that the full URL/origin is required so that the page is reloaded?

ui/app/templates/components/namespace-link.hbs
```diff
@@ -1,5 +1,5 @@
 <a
-  href="{{href-to 'vault.cluster.secrets' (query-params namespace=this.normalizedNamespace)}}"
```
chelshaw, line 2: We can remove the `href-to` dep, since this is the only place it was used

chelshaw (changes requested):
A couple small things, looks great though!

ui/app/templates/components/namespace-link.hbs
Monkeychip: good catch

PR #10580, opened 2020-12-15
Update to add the newly released Couchbase DB SE

acahn:


website/pages/docs/plugin-portal/index.mdx
```diff
@@ -54,6 +54,7 @@ exists within the Vault repository, the plugin can be built as instructed in
 <Columns count={2}>
 
 - [Cassandra](/api-docs/secret/databases/cassandra)
+- [Couchbase](/api-docs/secret/databases/couchbase)
```
calvn, line 57:
```suggestion
- [Couchbase](https://github.com/hashicorp/vault-plugin-database-couchbase)
```
We should use the plugin repo link since that 's available.

acahn: Sounds good.  Will make note of it for the next time.

PR #10582, opened 2020-12-16
Allow updating group by name.

luke-clifton:
Specifically, when using the identity/group endpoint with a name
parameter. Currently this is broken. If a name already exists it does
not bail out, and it fails to properly update the entities `group_ids`
fields, resulting in entities that continue to effectively be in the
group without actually being in the group (they continue to be able to
access group policies).

The alternate to this is to actually disallow this code path completely.

**Any Vault instances that have used this endpoint to update by name will find that entities that were removed using this endpoint continue to have access to policies associated with the group they were removed from.**

### DIFF
--- a/vault/identity_store_groups.go
+++ b/vault/identity_store_groups.go
@@ -220,7 +220,11 @@ func (i *IdentityStore) handleGroupUpdateCommon(ctx context.Context, req *logica
 		case groupByName == nil:
 			// Allowed
 		case group.ID == "":
-			group = groupByName
+			groupCloned, err := groupByName.Clone()
+			if err != nil {
+				return nil, err
+			}
+			group = groupCloned
 		case group.ID != "" && groupByName.ID != group.ID:
 			return logical.ErrorResponse("group name is already in use"), nil
 		}


luke-clifton:
The following script will demonstrate the issue.

```
set -e
set -u

VAULT_LISTEN="127.0.0.1:8300"
export VAULT_ADDR="http://$VAULT_LISTEN"
ROOT_TOKEN=abc123
vault server --dev --dev-root-token-id="$ROOT_TOKEN" --dev-listen-address="$VAULT_LISTEN" > /dev/null 2>/dev/null&
SERVER_PID=$!

trap "kill $SERVER_PID; wait $SERVER_PID" EXIT

sleep 1

c() {
ep="$1"
shift
curl -gfsSL -H "X-Vault-Token: $ROOT_TOKEN" "$VAULT_ADDR/v1/$ep" "$@"
}

# Create an entity.
me_id=$(c identity/entity -X POST | jq -r .data.id)

echo "Created entity: $me_id"

# Create a group in the normal fashion with the above entity as the sole member
gr_id=$(c identity/group --data-binary "{\"member_entity_ids\":[\"$me_id\"],\"name\":\"secret-my/super/secret-read\",\"policies\":[\"secret-my/super/secret-read\"]}" | jq -r '.data.id')

# Using the "Create" endpoint with a name when that name already exists.
# This does update the group which had the corresponding name....
# Here we have remove the entity from the member_entity_ids list
c identity/group --data-binary "{\"member_entity_ids\":[],\"name\":\"secret-my/super/secret-read\"}"

# However, this entity still thinks it is part of that group.
echo "Entity .data.group_ids:"
c identity/entity/id/$me_id | jq -c '.data.group_ids'
# Even though the group does not think that is the case.
echo "Group .data.member_entity_ids:"
c identity/group/name/secret-my/super/secret-read | jq '.data.member_entity_ids'

# Using the correct API to try and fix this.
c identity/group/id/$gr_id --data-binary "{\"member_entity_ids\":[]}"

# Does not work.
echo "Entity .data.group_ids:"
c identity/entity/id/$me_id | jq -c '.data.group_ids'
echo "Group .data.member_entity_ids:"
c identity/group/name/secret-my/super/secret-read | jq '.data.member_entity_ids'
```

luke-clifton:
The alternate is implemented in https://github.com/hashicorp/vault/pull/10585

luke-clifton:
Superseded by https://github.com/hashicorp/vault/pull/10812

PR #10586, opened 2020-12-16
Bug Fix: allow click of label and box for box-radio component

Monkeychip:
Because we're now in Glimmer land and this is not a crucial fix that needs to be backported, I went ahead and transformed the `box-radio` component to a glimmer component.

The main fix was removing the extra label element and adjusting the style.

With fix (before you could not click on the round radio button to select)

box-radio

### DIFF
--- a/ui/lib/core/addon/components/box-radio.js
+++ b/ui/lib/core/addon/components/box-radio.js
@@ -16,11 +16,20 @@
  * @param {string} [tooltipMessage=default] - The message that shows in the tooltip if the radio option is disabled
  */
 
-import Component from '@ember/component';
+// import Component from '@ember/component';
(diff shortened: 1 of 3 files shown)


ui/lib/core/addon/components/box-radio.js
```diff
@@ -16,11 +16,13 @@
  * @param {string} [tooltipMessage=default] - The message that shows in the tooltip if the radio option is disabled
  */
 
-import Component from '@ember/component';
+import Component from '@glimmer/component';
```
chelshaw, line 19: 👏

chelshaw (approved):
It's so amazing how much simpler glimmer-ifying made this fix! 🙌 Good job!

PR #10588, opened 2020-12-16
UI/managed namespace changes

chelshaw:
These changes include: 

* un-editable namespace prefix on auth route, if managed
* redirect to base namespace (for logged in user) or managed namespace root if no namespace param is present

- [x] Tests

**When a root namespace is retrieved from endpoint and no namespace param is present, redirect:**
rootNamespaceRedirect


ui/app/controllers/vault/cluster/auth.js
```diff
@@ -7,11 +7,32 @@ export default Controller.extend({
...
   authMethod: '',
   redirectTo: alias('vaultController.redirectTo'),
+  managedNamespaceRoot: alias('configService.managedNamespaceRoot'),
+
+  get managedNamespaceChild() {
+    let fullParam = this.namespaceQueryParam;
+    let split = fullParam.split('/');
+    if (split.length > 1) {
+      split.shift();
+      console.log(split);
```
Monkeychip, line 23: just heads up that there's a console log here

ui/app/controllers/vault/cluster/auth.js
```diff
@@ -7,11 +7,32 @@ export default Controller.extend({
...
+      return `/${split.join('/')}`;
+    }
+    return '';
+  },
+
+  updateManagedNamespace: task(function*(value) {
+    yield timeout(500);
```
Monkeychip, line 30: curious about the need for a timeout here?  maybe add a comment

ui/app/services/config.js
```diff
@@ -0,0 +1,9 @@
+import Service from '@ember/service';
+
+export default Service.extend({
```
Monkeychip, line 3: With octane you should be able to start writing these as Native classes (however, ignore me if we have to backport to prior to 1.7). 
ex: 
```
import Service from '@ember/service';

export default class ShoppingCartService extends Service {
}
```

Monkeychip (commented):
nice work so far!

ui/app/services/config.js
chelshaw: Great catch, however we do indeed have to backport this to prior to 1.7 😭

ui/app/controllers/vault/cluster/auth.js
chelshaw: Yes, same reason for the timeout on the manageNamespace fn but I'll copy it over. 

I could also make this DRYer by refactoring the original method but I wanted to not touch as much of the original functioning code as possible. Thoughts?

Monkeychip: Maybe just add a comment. I agree, that maybe not touch the original function.

Monkeychip (approved):
Nice work!

ui/mirage/config.js
```diff
@@ -2,6 +2,7 @@ export default function() {
   this.namespace = 'v1';
 
   this.get('sys/internal/counters/activity', function(db) {
+    console.log('getting sys/internal/counters/activity');
```
chelshaw, line 5: Remove logs

PR #10590, opened 2020-12-16
docs: update "Policy Syntax" section

alexanderbez:
Explain path matching criteria in more detail, followed by an example. (credit: @mgritter)


mgritter:
I think we need to do something with the new paragraphs to incorporate them into the yellow box?  The preview seems disjointed:

image

website/pages/docs/concepts/policies.mdx
```diff
@@ -173,8 +173,22 @@ capabilities, which controls a token's access to credentials in Vault.
...
+1. If the first wildcard (`+`) or glob (`*`) occurs earlier in `P1`, `P1` is lower priority
+2. If `P1` ends in `*` and `P2` doesn't, `P1` is lower priority
+3. If `P1` has more `+` (wildcard) segments, `P1` is lower priority
+4. If `P1` is shorter, it is lower priority
+5. If `P1` is smaller lexicographically, it is lower priority
+
+For example, given the two paths, `"secret/*"` and `"secret/+/+/foo/*"`, the first
+wildcard appears in the same place, both end in `*` and the latter has two wildcard
+segments while the former has zero. So we end at rule (3), and give `"secret/+/+/foo/*"`
+_lower_ priority.
```
mgritter, line 188: Maybe each paragraph should be prefixed by ~>?  I don't know the details of the markup language well enough to say whether that would create a single note or multiple notes.

mgritter: Or, I suppose, break off everything past "Specifically, when there are multiple...." into the next paragraph so that only the existing sentence is called out with a highlighted box?

mladlow: You can run `make` in `website/` to get a local preview. This is really useful for checking out formatting and link validity.

alexanderbez: I don't know either. That's why I simply just put them beneath the `->` section. I've seen this done in other docs as well.

website/pages/docs/internals/security.mdx
```diff
@@ -118,7 +118,8 @@ client), the highest access level permitted is used. For example, if the
 "engineering" policy permits read/update access to the "eng/" path, and the
 "ops" policy permits read access to the "ops/" path, then the user gets the
 union of those. Policy is matched using the most specific defined policy, which
-may be an exact match or the longest-prefix match glob pattern.
+may be an exact match or the longest-prefix match glob pattern. See
```
mladlow, line 121: It's not always the longest-prefix match right? If it's not, we shouldn't say that part here.

alexanderbez: I do not know. I'll have to refer to @mgritter

website/pages/docs/concepts/policies.mdx
```diff
@@ -173,8 +173,22 @@ capabilities, which controls a token's access to credentials in Vault.
 ~> Policy paths are matched using the **most specific path match**. This may be
 an exact match or the longest-prefix match of a glob. This means if you define a
```
mladlow, line 174: Like my other comment, once we get consensus on a replacement for "longest-prefix match" we should put it here as well.

website/pages/docs/internals/security.mdx
mladlow: @mgritter do you think it would be more correct to say something like:

"which may be an exact match or will otherwise follow the rules described in the [Policy Syntax]...

?

mgritter: Something like that would be better, although it's unfortunate that we can't emphasize the common case that longest-prefix is how it works as long as you're only doing globs.

I don't like this whole sentence, though, "policy is matched using the most specific defined policy" means what?  I think we need to clarify somehow that:
* If the same pattern appears in multiple policies, we take the union of the capabilities
* If different patterns appear in the applicable policies, we take only the highest-priority match from those policies

Something like "The policy rules that Vault applies are determined by the most-specific match available, using the priority rules described [blah]."

website/content/docs/concepts/policies.mdx
```diff
@@ -170,11 +170,30 @@ corresponding path and capability - even Vault's internal core configuration
 endpoints live under the `"sys/"` path. Policies define access to these paths and
 capabilities, which controls a token's access to credentials in Vault.
```
mgritter, line 171: I noticed the preview showed a line break after `"sys/"`, but I don't understand why.

website/content/docs/concepts/policies.mdx
```diff
@@ -170,11 +170,30 @@ corresponding path and capability - even Vault's internal core configuration
...
+1. If the first wildcard (`+`) or glob (`*`) occurs earlier in `P1`, `P1` is lower priority
+2. If `P1` ends in `*` and `P2` doesn't, `P1` is lower priority
+3. If `P1` has more `+` (wildcard) segments, `P1` is lower priority
+4. If `P1` is shorter, it is lower priority
+5. If `P1` is smaller lexicographically, it is lower priority
+
+For example, given the two paths, `"secret/*"` and `"secret/+/+/foo/*"`, the first
+wildcard appears in the same place, both end in `*` and the latter has two wildcard
+segments while the former has zero. So we end at rule (3), and give `"secret/+/+/foo/*"`
+_lower_ priority.
```
mgritter, line 193: Same problem here, I don't think there's anything we can do about it?

image

Seems to be a problem with how Chrome is flowing the text, changing the window width gives different results.

mgritter (approved):
Looks good; preview had some formatting problems that are not the fault of this PR.

PR #10594, opened 2020-12-16
Fix setting Activity Log enable flag through the API

swayne275:
This is a follow-up to https://github.com/hashicorp/vault/pull/10485

- somewhere along the way we lost the actual line that sets enabled from the config update
- updated tests to log warning(s) rather than fail (@mgritter is this in line with the spirit of these tests?)
- cleaned up a little bit

Note: I didn't add a test for `handleActivityConfigUpdate` in `logical_system_test.go` because:
- Currently we don't test this function in that file
- the code is covered by stuff in at least `activity_log_test.go`
- the `activity_log_test.go` tests caught this bug

`activity_log_test.go` tests are passing locally now


swayne275:
Found typo, just a sec

vault/activity_log_test.go
```diff
@@ -643,8 +643,8 @@ func TestActivityLog_API_ConfigCRUD(t *testing.T) {
 		if err != nil {
 			t.Fatalf("err: %v", err)
 		}
-		if resp != nil {
-			t.Fatalf("bad: %#v", resp)
+		if resp != nil && len(resp.Warnings) > 0 {
+			t.Logf("got warning(s): %#v", resp.Warnings)
 		}
```
ncabatoff, line 648: If the warning is expected, why log it?  If anything, since it's expected, shouldn't the test error out if it's absent?

swayne275: good point, in some cases we do want that. i'll audit

swayne275: it's only expected in some cases, and these helpers don't currently track the appropriate state. i can make them

swayne275: Actually, does it make sense to enforce API specifics when we are unit testing the things that are being used by the API? I would think a test like that makes more sense to verify that the API responds appropriately in some API tests.

@mgritter what are your thoughts? what was the original reason for failing if there were warnings?

mgritter: The warnings are part of the expected behavior now, and I think we should check their presence. I should have asked Bez to do this on his original PR.

swayne275: please unresolve if this doesn't fix it: fe38ff91e

vault/activity_log_test.go
```diff
@@ -1640,6 +1642,25 @@ func TestActivityLog_DeleteWorker(t *testing.T) {
...
 }
 
+// checkAPIWarnings ensures there is a warning if switching from enabled -> disabled,
+// and no response otherwise
+func checkAPIWarnings(t *testing.T, originalEnabled, newEnabled bool, resp *logical.Response) {
+	t.Helper()
+
+	expectWarning := originalEnabled == true && newEnabled == false
+
+	if !expectWarning && resp != nil {
```
ncabatoff, line 1652: I find switch statements more readable than nested if/elses like this, but YMMV.

ncabatoff (approved):

vault/activity_log_test.go
swayne275: hmm, how would you set up this particular one as a switch?

ncabatoff:
```
switch {
case !expectWarning && resp != nil: 
	t.Fatalf("got unexpected response: %#v", resp)
case expectWarning && resp == nil:
	t.Fatal("expected response (containing warning) when switching from enabled to disabled")
case expectWarning && len(resp.Warnings) == 0:
	t.Fatal("expected warning when switching from enabled to disabled")
}
```

vault/activity_log_testing_util.go
```diff
@@ -164,3 +164,10 @@ func (a *ActivityLog) SetEnable(enabled bool) {
 	defer a.fragmentLock.Unlock()
 	a.enabled = enabled
 }
+
+// GetEnabled returns the enabled flag on an activity log
+func (a *ActivityLog) GetEnabled() bool {
+	a.fragmentLock.RLock()
+	defer a.fragmentLock.RUnlock()
+	return a.enabled
+}
```
mgritter, line 173: Heh... I told Bez we should just add this so we could check GetEnabled() before and after SetConfig.

mgritter (approved):

vault/activity_log_test.go
swayne275: you can do logical ops in a go switch?? awesome! I'd googled quickly and didn't see examples

vault/activity_log_testing_util.go
swayne275: yeah so this exists in an _ent file but not OSS so I just moved it over

PR #10595, opened 2020-12-17
line break oddity fix

mikegreen:


taoism4504:
The issue appears to be fixed in the latest version of the doc. Please feel free to open another PR if you don't find that to be the case.

Issue #10418, opened 2020-11-18
Cannot View Version History of Secret in Web UI When There is Whitespace in the Path


andrew0wells:
**Describe the bug**
Version history web page displays error message "Unable to find secret at /. Try going back to the root and navigating from there." when the path for this secret contains a whitespace character.

**To Reproduce**
1. Create a secret with white in the path to the secret. via the web interface (IE "Test Problem")
2. Click the History button
3. Click "View version history" from the drop-down menu
4. Error message "Unable to find secret at /. Try going back to the root and navigating from there." is displayed.

**Expected behavior**
Version history should display in the web UI regardless of whether there is a whitespace character in the path. 

**Environment:**
* Vault Server Version (retrieve with `vault status`): 

```hcl
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    5
Threshold       5
Version         1.6.0
Storage Type    file
Cluster Name    vault-cluster-1593d935
Cluster ID      66d79008-fb4f-0ee7-5ac6-4a0187233b6f
HA Enabled      false
```

* Vault CLI Version (retrieve with `vault version`): Vault v1.6.0 (7ce0bd9691998e0443bc77e98b1e2a4ab1e965d4)

* Server Operating System/Architecture:  
```hcl
Get-ComputerInfo | Select-Object OsName, OsVersion

OsName                                 OsVersion 
------                                 --------- 
Microsoft Windows Server 2019 Standard 10.0.17763
```

Vault server configuration file(s):

```hcl
storage "file" {
  path    = "<storage location>"
}


listener "tcp" {
address = "<IP>:8200"
tls_cert_file = "<cert file location>"
tls_key_file  = "<key file location>"
tls_disable_client_certs = "false"
}

api_addr = "https://<IP>:8200"
cluster_addr = "https://<IP>:8201"
ui = true
```

**Additional context**
Google Chrome displays the following error message in the Chrome console when then error message "Unable to find secret at /. Try going back to the root and navigating from there." displays:

```hcl
vendor-b965e34fb26c54d097fa32549a3d857a.js:2121 Error while processing route: vault.cluster.secrets.backend.versions Ember Data Request GET /v1/andrew.wells/metadata/Test%2520Problem returned a 404
Payload (application/json)
[object Object] Error: Ember Data Request GET /v1/andrew.wells/metadata/Test%2520Problem returned a 404
Payload (application/json)
[object Object]
[...]

PR #10596, opened 2020-12-17
Bug: Fix issue with double encoding on space in secret history route

Monkeychip:
This fix address [PR issue 10418](https://github.com/hashicorp/vault/issues/10418)

We call a method `encodePath` which is apart of a third-party library. By default, this method encodes spaces from ` ` to `%20`.  The problem is that we have already encoded white spaces before it gets sent to this method, such that the white spaces essential get encoded twice resulting in ` %2520`  instead of `%20`.  To fix this, I replaced the already encoded space using regex back to whitespace before it gets sent to this method.

Because the URL is getting sent to `encodePath` no matter what, I don't see any issues here, but it's worth some consideration.

[Here is the library](https://github.com/tildeio/route-recognizer) for the encodePath and here is a [PR](https://github.com/tildeio/route-recognizer/pull/109) that talks about the special characters it does not encode (white space is not included).

**With fix:** The version page now shows where before it didn't.
image

**Without fix:** Double encoding on the spaces shows up at %2520 
image

[From a helpful Stackoverflow article](https://stackoverflow.com/questions/16084935/a-html-space-is-showing-as-2520-instead-of-20#:~:text=A%20bit%20of%20explaining%20as,the%20%20%20to%20%2520%20)
>The common space character is encoded as %20 as you noted yourself. The % character is encoded as %25.
>The way you get %2520 is when your url already has a %20 in it, and gets urlencoded again, which transforms the %20 to %2520

### DIFF
--- a/ui/app/adapters/secret-v2.js
+++ b/ui/app/adapters/secret-v2.js
@@ -8,9 +8,10 @@ export default ApplicationAdapter.extend({
   _url(backend, id) {
     let url = `${this.buildURL()}/${encodePath(backend)}/metadata/`;
     if (!isEmpty(id)) {
-      let test = id.replace(/%20/g, ' ');
-      console.log(id, 'id');
-      url = url + encodePath(test);
+      // if path has a whitespace already encoded, return to whitespace because method encodePath will encode it.
+      // if we don't return to a whitespace it will get encoded as %20, resulting in a %2520 (% => 25).
+      let path = id.replace(/%20/g, ' ');
+      url = url + encodePath(path);
     }
     return url;
   },
--- a/ui/app/adapters/secret-v2.js
+++ b/ui/app/adapters/secret-v2.js
@@ -8,10 +8,7 @@ export default ApplicationAdapter.extend({
   _url(backend, id) {
     let url = `${this.buildURL()}/${encodePath(backend)}/metadata/`;
     if (!isEmpty(id)) {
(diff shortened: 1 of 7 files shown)


ui/app/adapters/secret-v2.js
```diff
@@ -8,7 +8,10 @@ export default ApplicationAdapter.extend({
   _url(backend, id) {
     let url = `${this.buildURL()}/${encodePath(backend)}/metadata/`;
     if (!isEmpty(id)) {
-      url = url + encodePath(id);
+      // if path has a whitespace already encoded, return to whitespace because method encodePath will encode it.
+      // if we don't return to a whitespace it will get encoded as %20, resulting in a %2520 (% => 25).
+      let path = id.replace(/%20/g, ' ');
```
chelshaw, line 13: Should we instead decode and then encode? or find where we're encoding upstream and remove it? My concern is that any other encoded values will get double-encoded as well, and then we'll end up in a completely wonky state where it's hard to get back to the original value. Also, I'd like to see some tests (and a changelog) around this before we merge

Monkeychip: @chelshaw let's chat about this during our sync today.  From what I can tell the URL comes into queryRecord (which then goes to the _url method) already encoded.  And ✅  on tests and changelog.

chelshaw (approved):
🙌

PR #10605, opened 2020-12-19
Add suport for Consul tokens with roles and instances

kula:
This adds support to the Consul secret engine to support generating tokens with Consul roles (available since 1.5), Consul Service Identities (available since 1.5) and Consul Node Identities (available since 1.8.1).

Includes updated test suite and updated documentation.

PR #10606, opened 2020-12-20
Add support for Consul tokens with roles and identities, round 2

kula:
This adds support to the Consul secret engine to support generating tokens with Consul roles (available since 1.5), Consul Service Identities (available since 1.5) and Consul Node Identities (available since 1.8.1).

Includes updated test suite and updated documentation.

Note that initial try was in https://github.com/hashicorp/vault/pull/10605, but I messed up something there which was causing test failures off in other areas. I believe I've fixed that here. Note also that this is based on the v1.6.1 tag, not master, because basing it off of there (commit 179df992c0720fabf36794793a729918b1ea0b5d) was causing other test failures which made no sense at all. At least one of those failures passes now, so trying this to let CircleCI make its pass.


kula:
That's not going to work, as the PR is going to try to add this on top of master and then run CircleCI. Closing.

PR #10612, opened 2020-12-21
Add NIST guidance on rotating keys used for AES-GCM encryption

sgmiller:


HridoyRoy (approved):
lgtm!

website/content/docs/internals/rotation.mdx
```diff
@@ -54,3 +54,16 @@ provides the `N+1` encryption key protected by the `N` key. This upgrade key is
...
 This allows standby instances to update their keys and stay in-sync with the active Vault
 without requiring operators to perform another unseal.
+
+## NIST Rotation Guidance
+
+Period rotation of the encryption keys is recommended, even in the absence of 
+compromise.  Due to the nature of the AES-256-GCM encryption used, keys should be 
+rotated before approximately 2<sup>32</sup> encryptions have been performed, following
+the guidelines of NIST publication 800-38D.  It is recommended that operators
+monitor the sum of `vault.barrier.put` and `vault.token.creation` (with a label
```
briankassouf, line 64: Can we add a `vault.barrier.encryption` metric which will capture storage operations, token creations, and anything else we use the barrier for today? We'd have to mention that versions < X should look at these existing metrics though.

briankassouf: We can also add a label with the key version on this new metric to help track encryptions per key

sgmiller: Added that metric sampled probabilistically to maintain performance.  For now I'll leave that out of the docs to be added when 1.7 is released.

PR #10616, opened 2020-12-22
Don't list certs if we were told which cert to use.

ncabatoff:


builtin/credential/cert/path_login.go
```diff
@@ -414,21 +414,33 @@ func (b *backend) loadTrustedCerts(ctx context.Context, storage logical.Storage,
...
+
 	for _, name := range names {
-		// If we are trying to select a single CertEntry and this isn't it
-		if certName != "" && name != certName {
-			continue
-		}
 		entry, err := b.Cert(ctx, storage, strings.TrimPrefix(name, "cert/"))
 		if err != nil {
-			b.Logger().Error("failed to load trusted cert", "name", name, "error", err)
+			if certName == "" {
```
vishalnayak, line 433: Why do we swallow the error when certName is given and we fail to load the cert?

ncabatoff: If I had a good reason, I've forgotten it.  Good catch.

vishalnayak (approved):

raskchanky (approved):

PR #10618, opened 2020-12-23
Fix typo in usage description of operator init -status flag

jace-ys:
This commit fixes a typo in the operator init -status flag's usage
description and aligns it with that found on https://www.vaultproject.io/docs/commands/operator/init#status.


swayne275 (approved):
Thanks!

PR #10643, opened 2021-01-04
UI/namespace error page

Monkeychip:
This PR address a potential issue in HCP where a user would be authenticated to vault via HCP and then click a link that takes them to the vault/ui ; however the link could not append the url with the namespace query params.  

We addressed this issue in this PR, where we grab the namespace from local storage.  However, as a backup, we now have an empty state.

Here is a gif:
namespace


Monkeychip:
Not necessary because we solve this problem in localStorage, nobody should get to this state.

PR #10644, opened 2021-01-04
feat(agent): add retry configuration for vault agent

RXC001:
Adds retry configuration for Vault Agent. Structure is similar to the `retry` blocks found in [consul-template](https://github.com/hashicorp/consul-template#configuration-file-format)


HridoyRoy (commented):
👍 LGTM, but I'd like another pair of eyes on this in case I missed something. @calvn could you take a look?

command/agent/config/config.go
```diff
@@ -40,6 +40,7 @@ type Vault struct {
 	ClientCert       string      `hcl:"client_cert"`
 	ClientKey        string      `hcl:"client_key"`
 	TLSServerName    string      `hcl:"tls_server_name"`
+	Retry            *Retry      `hcl:"retry"`
```
calvn, line 43: The retry behavior in here is specifically for the template server (at least for now) so I'm not entirely convinced that it should live in the `vault` stanza since this doesn't control retry logic when agent communicates with the server for auto-auth and/or caching.

@HridoyRoy thoughts on making this be a top-level `template_retry` stanza instead?

ibspoof: We did think about that, but the existing retry context is under `ctv.Vault.Retry = &ctconfig.RetryConfig{Attempts: &ts.testingLimitRetry}` and consul-template puts it under the vault section so we kept the same structure location.

As for the auto-auth retry that is a feature we would like to add as well as our use case of vault needs to be fault tolerant of network disconnects both for getting new tokens and template rendering, this MR was the first step.

HridoyRoy: Hm, I do agree that "Retry" in the Vault section shouldn't be a configuration parameter that is limited to the consul template use-case. Let me dig into this a bit -- I'll circle back soon.

HridoyRoy: Hi @ibspoof @RXC001 , sorry for the delay in circling back! 

We'd like to move the Retry stanza into as a top level config named `template_retry`. `ctv` is the consul template, so `ctv.Vault.Retry ` can be set to values that exist in a structurally different location in the Vault config. Since this retry functionality is currently limited in scope to the consul-template, we'd want to make that clear with the name and location of the stanza, instead of keeping structural similarity between the consul-template config and the vault config. 

Thanks so much!

HridoyRoy (changes requested):

command/agent/config/config.go
ibspoof: Makes sense we had the same discussion in our team. Will get that changed and pushed.

RXC001:
Moving the Retry stanza to the top level makes sense. I can make that happen.

HridoyRoy (approved):
lgtm!

HridoyRoy:
Hi @RXC001 , I think this needs to pull from HashiCorp/master for the tests to pass (seems to be missing https://github.com/hashicorp/vault/pull/10749 ). 
Thanks, and thank you for your patience.

other-ryan:
Thanks for this feature. This is awesome!

It appears this didn't make it into the 1.6.2 release. Any timeframe of when we can expect this to be part of the GA build?

ncabatoff:
> Thanks for this feature. This is awesome!
> 
> It appears this didn't make it into the 1.6.2 release. Any timeframe of when we can expect this to be part of the GA build?

It should be included in the upcoming 1.7 release.  We don't have an official timeframe that I know of, but if you look at the past cadence of Vault major releases you'll see that we're pretty regular and that 1.7 is likely not far off.

other-ryan:
@ncabatoff thanks for the info. Just a heads up that the documentation for these configs are live at https://www.vaultproject.io/docs/agent/template and may cause some confusion for folks trying to figure out why the configs aren't working.

ncabatoff:
Unfortunately this has been an issue for the longest time.  Our docs reflect not the latest available version, but what's in the master branch.  There is work ongoing to provide a versioned view of documentation which should fix this, but I don't know when that will be released.  Sorry about the confusion in the meantime.

kalafut:
@ncabatoff The docs for this are showing up because the `docs-cherrypick` label was added in https://github.com/hashicorp/vault/pull/10911. As far as I can tell we're rendering `stable-website` as expected.

ncabatoff:
Hi folks,

I'm afraid I have some potentially bad news: I'm reverting this change.  In its place, #11113 is adding similar functionality that will apply not just to templating, but also to requests that go through the proxy/cache.  I say this is potentially bad news because the new configuration *only* allows you to specify the number of retries, there's no provision like in template_retry for specifying backoff or max_backoff.

Some context for this decision:
* we needed the ability to configure retries for proxy requests for another 1.7 feature
* yet another 1.7 feature allows template requests to go through the cache, and having independent configuration for proxy requests and template requests would be problematic in that scenario
* templating uses a different backoff algorithm than proxy requests, so providing backoff configuration that has very different meanings for the two systems would've created confusion
* we're not persuaded that being able to configure backoff intervals is strictly necessary

In the future we plan to do further work to better unify templating with the rest of Agent, such that we won't have multiple backoff algorithms in use, and we can revisit the decision as to whether to have configuration for backoff intervals.

PR #10649, opened 2021-01-05
Fix ip disclosure

sgmiller:
Remove specific information from unauthenticated errors that would have included IP addresses.  At the same time, maintain the HTTP status code.


ncabatoff (approved):

mladlow:
@sgmiller I think this still needs to be backported to 1.6? If not could you hunt down the backport PR?

PR #10650, opened 2021-01-05
Make the error response to the sys/internal/ui/mounts with no client token consistent

sgmiller:
If an authenticated call is made to this endpoint, check the token anyway even if the mount doesn't exist (e.g. don't early out) in order to return the same error whether the mount exists or not.

Addresses VAULT-872.


vault/logical_system.go
```diff
@@ -3363,6 +3363,12 @@ func (b *SystemBackend) pathInternalUIMountRead(ctx context.Context, req *logica
 
 	me := b.Core.router.MatchingMountEntry(ctx, path)
 	if me == nil {
+		// To be consistent with the case no client token was supplied, go through the motions of verifying authorization
```
ncabatoff, line 3366: Do we also need to handle the equivalent case where the mount exists but replication filtering precludes using it?  (see checkReplicatedFiltering below)

sgmiller: Yep, I think you're right, added.

ncabatoff: Why not call b.verifyAuthorizedMountAccess at the outset?  If we're going to return an error when it returns an error, why even bother looking up the mount entry or filtering?

sgmiller: So that we return the error each code path meant to do if the verifyA..M..A.. succeeds.  In other words, only return a consistent error in the absence of a proper authenticated session, otherwise being more specific is okay.

sgmiller: Hmm, yeah, that may be equivalent and less subject to timing attacks.  I'll revisit in the morning but that may be a good point.

vault/logical_system.go
```diff
@@ -3386,25 +3398,34 @@ func (b *SystemBackend) pathInternalUIMountRead(ctx context.Context, req *logica
...
+	}
+
+	if !hasMountAccess(ctx, acl, fullMountPath) {
+		return errResp, logical.ErrPermissionDenied
+	}
+
+	return resp, nil
+}
+
+func (b *SystemBackend) verifyAuthorizedMountAccess(ctx context.Context, req *logical.Request) (*ACL, error) {
```
ncabatoff, line 3413: The name of this method is a bit misleading since I don't think it has anything to do with mounts.

vault/logical_system.go
sgmiller: Your right I think, no harm in doing the token/entity check first.  Changed.

vault/logical_system.go
```diff
@@ -3354,6 +3354,20 @@ func (b *SystemBackend) pathInternalUIMountRead(ctx context.Context, req *logica
...
 	path = sanitizePath(path)
 
+	// Load the ACL policies so we can walk the prefix for this mount
+	acl, te, entity, _, err := b.Core.fetchACLTokenEntryAndEntity(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	if entity != nil && entity.Disabled {
+		b.logger.Warn("permission denied as the entity on the token is disabled")
+		return nil, logical.ErrPermissionDenied
```
ncabatoff, line 3364: Is the change to not return errResp here deliberate?

sgmiller: I couldn't see why unlike everywhere else that ignored the errResp.

ncabatoff (approved):

PR #10656, opened 2021-01-05
docs: update Vault Helm to 0.9.0

jasonodonnell:
Vault Helm 0.9.0 was just released so I added the new configurables and updated relevant documentation.

### DIFF
--- a/website/content/docs/platform/k8s/helm/configuration.mdx
+++ b/website/content/docs/platform/k8s/helm/configuration.mdx
@@ -68,7 +68,7 @@ and consider if they're appropriate for your deployment.
 
       - `tag` (`string: "0.4"`) - The tag of the Docker image for the Leader Election. **This should be pinned to a specific version when running in production.**
     
-    - `ttl` - The frequency in which leader elections are performed in the Vault Agent Injector cluster.
+    - `ttl` (`string: "60s"`) - The frequency in which leader elections are performed in the Vault Agent Injector cluster.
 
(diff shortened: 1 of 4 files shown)


tvoran (commented):
Added some suggestions.

I also see that the injector affinity default is no longer `null`, so we should probably update that in the docs as well.

jasonodonnell:
@tvoran Good call, updated the affinity default for the injector. Thanks for the suggestions 👍

website/content/docs/platform/k8s/helm/configuration.mdx
```diff
@@ -134,16 +169,32 @@ and consider if they're appropriate for your deployment.
       "sample/annotation2": "bar"
     ```
 
-  - `failurePolicy` (`string : ""`) - When set to `Fail`, an error calling the webhook causes the admission to fail and the API request to be rejected.
+  - `failurePolicy` (`string: "Ignore"`) - When set to `Fail`, an error calling the webhook causes the admission to fail and the API request to be rejected. When set to `Ignore`, an error calling the webhook is ignored and the API request is allowed to continue. Allowed values: `Fail`, `Ignore`.
```
tvoran, line 172: Oops, the last sentence already exists on the next line:

```suggestion
  - `failurePolicy` (`string: "Ignore"`) - When set to `Fail`, an error calling the webhook causes the admission to fail and the API request to be rejected.
```

tvoran (approved):
One small oops.

PR #10659, opened 2021-01-06
Adding documentation for multiple vault-k8s replicas

tvoran:
Describes the setup and config for using multiple injector replicas
with auto and manual TLS.

Related to https://github.com/hashicorp/vault-k8s/pull/198 and https://github.com/hashicorp/vault-helm/pull/436


website/content/docs/platform/k8s/injector/installation.mdx
```diff
@@ -65,6 +65,32 @@ The following is required to configure TLS manually:
...
 
+## Multiple Replicas and TLS
+
+The Vault Agent Injector can be run with multiple replicas if using [Manual
+TLS](#manual-tls), and as of v0.7.0 multiple replicas are also supported with
+[Auto TLS](#auto-tls). The number of replicas is controlled in the Vault Helm
+chart by the [injector.replicas
+value](/docs/platform/k8s/helm/configuration#replicas).
+
+With Auto TLS, a leader-elector sidecar container is deployed with each replica.
```
tomhjp, line 76: Is it worth explicitly stating that the leader elector sidecar is only deployed if replicas > 1, and the default is 1?

website/content/docs/platform/k8s/injector/installation.mdx
```diff
@@ -65,6 +65,32 @@ The following is required to configure TLS manually:
...
+generating and distributing the certificate and key to the "followers". The
+followers read the certificate and key needed for the webhook service listener
+from a Kubernetes Secret, which is updated by the leader when a certificate is
+near expiration.
+
+The leader-elector sidecar in use is described in detail [here][k8s-blog]. For
+more information on configuring leader election, see the [Vault Helm
+leaderElector values](/docs/platform/k8s/helm/configuration#leaderelector).
+
+With Manual TLS,
```
tomhjp, line 88: Similar to the above comment:

```suggestion
With Manual TLS and multiple replicas,
```

tomhjp (approved):
Thanks, LGTM! You've done a really nice job distilling lots of options down into a digestible explanation. Reading through it just makes me wonder whether perhaps I should have made multiple replicas the default 🤔

website/content/docs/platform/k8s/injector/installation.mdx
tvoran: Yeah, that's a good point. I think folks should be referring to the injector.replicas link in the previous paragraph for the current default, and I'll add something similar to your other suggestion about multiple replicas here.

tvoran:
@tomhjp Thanks for the notes! I think leaving the default replicas at 1 is fine for now, since that'll mean fewer surprises for folks when they upgrade their charts.

tomhjp (approved):
Good point about upgrades 👍

PR #10662, opened 2021-01-06
Bug: Fix namespace test 

Monkeychip:
_Should not be backported because was an issue caused by Ember Upgrade_

This test was broken in this [PR](https://github.com/hashicorp/vault/pull/10572/files)

The problem was with the new functionality, the page refreshed when switching namespaces.  Qunit doesn't like this.  So instead of using the toggle (which I could stop from refreshing), I instead just manually change to URL to get into the nested namespaces and then confirm they exist –which is the purpose of the test.

I've run the full test suite, and everything passes (CircleCI won't run the enterprise test).


chelshaw (approved):
👏

PR #10686, opened 2021-01-11
Template library: Remove duplicate funcs, add timestamp with format

pcman312:
Follow on from @calvn 's comments on https://github.com/hashicorp/vault/pull/10500 which came after I merged it.

- Removed duplicate functions (`trunc`, `upper`, etc.)
- Changed `now_seconds` to `unix_time`
- Changed `now_nano` to `unix_time_millis` (also changed from nanoseconds to milliseconds)
- Added `timestamp` that allows a custom timestamp format based on Go's time formatting

### DIFF
--- a/sdk/helper/template/funcs.go
+++ b/sdk/helper/template/funcs.go
@@ -10,12 +10,16 @@ import (
 	UUID "github.com/hashicorp/go-uuid"
 )
 
-func nowSeconds() string {
+func unixTime() string {
 	return strconv.FormatInt(time.Now().Unix(), 10)
 }
 
-func nowNano() string {
-	return strconv.FormatInt(time.Now().UnixNano(), 10)
+func unixTimeMillis() string {
+	return strconv.FormatInt(time.Now().UnixNano()/int64(time.Millisecond), 10)
+}
+
+func timestamp(format string) string {
+	return time.Now().Format(format)
 }
 
 func truncate(maxLen int, str string) (string, error) {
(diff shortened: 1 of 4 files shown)


sdk/helper/template/funcs.go
```diff
@@ -10,12 +10,16 @@ import (
...
 
-func nowSeconds() string {
+func unixTime() string {
 	return strconv.FormatInt(time.Now().Unix(), 10)
 }
 
-func nowNano() string {
-	return strconv.FormatInt(time.Now().UnixNano(), 10)
+func unixTimeMillis() string {
+	return strconv.FormatInt(time.Now().UnixNano()/int64(time.Millisecond), 10)
```
calvn, line 18: Is there any reason to not go with nanoseconds? We should provide a way to either go with somewhat precise time (second) or very precise time (nanoseconds), similar to `Unix` and `UnixNano`.

calvn (commented):
Changes look good aside from a question/comment around the unix time funcs.

sdk/helper/template/funcs.go
pcman312: There isn't a need to provide that level of precision, especially if that level of precision adds additional characters to a username. The majority of systems out there that I've personally seen do not go down to the nanosecond level. The main reason I am even including milliseconds is to improve the ability to differentiate between usernames in systems that receive a lot of dynamic user requests. If nanoseconds is something that is desired we can add it at a later time.

calvn (approved):

PR #10687, opened 2021-01-12
Add token as parameter for postgres queries

UXabre:
An initial effort to provide variables in database queries.
I stuck with what I know and limited this to postgres.

Initially I wanted to re-use the code used to interpolate variables in the policies but they both have very different implementations. I kinda like the simplicity of what was implemented here though and am willing to expand the scope of both the database that are supported as well as to rework this code to be on par with the variable interpolations available for policies.

This adds the possibility to use {{token.metadata.<key>}} (next to {{name}}, {{password}} and {{expiration}})

### DIFF
--- a/builtin/logical/postgresql/path_role_create.go
+++ b/builtin/logical/postgresql/path_role_create.go
@@ -113,7 +113,13 @@ func (b *backend) pathRoleCreateRead(ctx context.Context, req *logical.Request, 
 			"expiration": expiration,
 		}
 
-		if err := dbtxn.ExecuteTxQueryDirect(ctx, tx, m, query); err != nil {
+		// inject metadata as parameters
+		metadata := req.TokenEntry().Meta
+		for k, v := range metadata {
+			m[fmt.Sprintf("token.metadata.%s", k)] = v
+		}
+
+		if err := dbtxn.ExecuteTxQuery(ctx, tx, m, query); err != nil {
 			return nil, err
 		}
 	}
--- a/changelog/10682.txt
+++ b/changelog/10682.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+core: support token metadata in template
+```
--- a/sdk/helper/identitytpl/templating.go
+++ b/sdk/helper/identitytpl/templating.go
@@ -14,6 +14,7 @@ import (
 
 var (
 	ErrUnbalancedTemplatingCharacter = errors.New("unbalanced templating characters")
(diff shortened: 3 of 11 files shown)


aphorise:
Hey @UXabre any chance you can resolve the merge conflict here please?

UXabre:
Yes! Will do this straight away!

aphorise:
@UXabre thanks very much for the earlier fix to the merge conflict; would you mind updating it once more?

The only thing I'm wondering is if there should be some examples demonstrations documented around the use of this further / later on within:
 - Vault Documentation: [PostgreSQL Database Secrets Engine](https://developer.hashicorp.com/vault/docs/secrets/databases/postgresql)

 - maybe as an addition the current setup section (renamed to `Basic Setup`) with something like: `Setup with Token Parameters` afterwards - WDYT?

Hey @austingebauer @robmonte @fairclothjm @jasonodonnell - I see you folks have made commits in these areas of the code and was curious to hear any feedback you may have on this PR or if can get consideration for merging with the next release maybe?

UXabre:
Hey, I dug right in to merge this, but it seems a whole lot has changed in the underlying database implementation so would need to carefully examine the new changes that would need to be made before fixing the conflicts.
It'll probably take a bit of time to fix this but in the meantime the others mentioned could give their 2 cents on it :-)

aphorise:
> Hey, I dug right in to merge this, but it seems a whole lot has changed in the underlying database implementation so would need to carefully examine the new changes that would need to be made before fixing the conflicts. It'll probably take a bit of time to fix this but in the meantime the others mentioned could give their 2 cents on it :-)

Hey no worries @UXabre - just to prevent any false pretence or impressions - it seems that I also am missing a bit and this PR is not likely to need any attention for the time being. Lets see what transpires from the other threads that I see you're already discussing on as well :-)

UXabre:
Great! Thanks for your feedback and we'll see how this evolves further. Perhaps it's best that I close this PR and await the future?

aphorise:
> Great! Thanks for your feedback and we'll see how this evolves further. Perhaps it's best that I close this PR and await the future?

I say lets leave it open for now till we get the same or similar functionality as you've already drafted. So this PR is a good reminder or now.

VioletHynes:
Hey there! Given the previous comments, and the fact that this is now quite old with complicated merge conflicts to resolve, I'm going to close this PR. 

Please feel free to open a new PR in the future if you're still interested in this change and find the time to incorporate everything!

PR #10689, opened 2021-01-13
Fix quota enforcing old path issue

vishalnayak:


vault/quotas/quotas_rate_limit.go
```diff
@@ -92,6 +92,20 @@ func NewRateLimitQuota(name, nsPath, mountPath string, rate float64, interval, b
 	}
 }
 
+func (q *RateLimitQuota) Clone() *RateLimitQuota {
```
vishalnayak, line 95: @alexanderbez Should we push the clone() function into the quota interface and potentially call this in setQuotaLocked function before the initialize function? This way even the lease count quota can implement clone and not run into similar issues.

ncabatoff (approved):
Smart!

mgritter (approved):
This looks good, but don't we need to do the same thing for lease count quotas, to catch changes to namespace and path?

```
                switch {
                case quota == nil:
                        quota = quotas.NewLeaseCountQuota(name, ns.Path, mountPath, int64(maxLeases))
                default:
                        lcq := quota.(*quotas.LeaseCountQuota)
                        lcq.NamespacePath = ns.Path
                        lcq.MountPath = mountPath
                        lcq.MaxLeases = int64(maxLeases)
                }
```

vishalnayak:
Yep. We need to. That PR will follow.

PR #10695, opened 2021-01-14
Add changelog for ent PR1668.

ncabatoff:


ncabatoff:
This should be on 1.6 instead.

PR #10705, opened 2021-01-15
agent: add an inflight cache better concurrent request handling

calvn:
This PR refactors `LeaseCache.Send` to better handle identical concurrent requests that are processed at this layer.

The original code relied solely on an `idLock`, which would get updated to a write lock if the request resulted in a cache miss, meaning that non-cacheable identical concurrent requests (e.g. Vault KV reads) would be processed serially. It also was not properly protecting identical concurrent cacheable requests on a clean cache from being proxied more than once.

This PR introduces an inflight cache to allow for one of these requests to win and proceed with processing the request before the other ones are processed. This implementation takes a fan-out approach to ensure that a request is processed at least once (and that only one of them does) so that it has a chance to for `LeaseCache` to cache its response. 

The changes in this PR also ensures that identical concurrent cacheable requests that does not have it response cached yet will only ever have the request proxied once since the inflight cache will only allow one winner to fully process the request before the others can proceed.


command/agent/cache/lease_cache.go
```diff
@@ -453,6 +494,8 @@ func computeIndexID(req *SendRequest) (string, error) {
 	// requests sets the token directly into SendRequest.Token
 	b.Write([]byte(req.Token))
 
+	// b.Write([]byte(time.Now().String()))
```
kalafut, line 497: Stray commented-out code

command/agent/cache/lease_cache.go
```diff
@@ -170,40 +193,59 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
 		return nil, err
 	}
 
-	// Grab a read lock for this particular request
-	idLock := locksutil.LockForKey(c.idLocks, id)
+	// Check the inflight cache to see if there are other inflight requests
+	// of the same kind, based on the computer ID. If so, we increment a counter
```
ncabatoff, line 197:
```suggestion
	// of the same kind, based on the computed ID. If so, we increment a counter
```

command/agent/cache/lease_cache.go
```diff
@@ -89,6 +95,22 @@ type LeaseCacheConfig struct {
...
 }
 
+type inflightRequest struct {
+	// ch is closed by the request that ends up processing the set of
+	// parallel request
+	ch chan struct{}
+
+	// remaining is the number of remaining inflight request that needs to
+	// be processed before this object can be cleaned up
+	remaining atomic.Uint64
```
ncabatoff, line 105: Seems more natural to make this a sync.WaitGroup, what do you think?

ncabatoff: Oh nvm, I read too fast - you don't actually block on this.

command/agent/cache/lease_cache.go
```diff
@@ -170,40 +193,59 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
...
+	if found {
+		idLock.Unlock()
+		inflight = inflightRaw.(*inflightRequest)
+		inflight.remaining.Inc()
+		defer inflight.remaining.Dec()
+
+		// If found it means that there's an inflight request being processed.
+		// We wait until that's finished before proceeding further.
+		select {
+		case <-ctx.Done():
```
ncabatoff, line 226: Shouldn't we return early if the ctx is done?

command/agent/cache/lease_cache.go
calvn: An earlier iteration of this used `sync.WaitGroup`, but that was harder to wrap my head around since the usage of remaining in here is for a non-blocking manner such that we know when we can clean up the `inflightCache` entry on L204. `Wait()` would block the caller, and we wouldn't know which request should call it to do the cleanup.

command/agent/cache/lease_cache.go
calvn: Good catch!

command/agent/cache/lease_cache_test.go
```diff
@@ -509,3 +532,110 @@ func TestCache_DeriveNamespaceAndRevocationPath(t *testing.T) {
...
+
+func TestLeaseCache_Concurrent_NonCacheable(t *testing.T) {
+	lc := testNewLeaseCacheWithDelay(t, false, 50)
+
+	// We are going to send 100 requests, each taking 50ms to process. If these
+	// requests are processed serially, it will take ~5seconds to finish. we
+	// use a ContextWithTimeout to tell us if this is the case by giving ample
+	// time for it process them concurrently but time out if they get processed
+	// serially.
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
ncabatoff, line 544: Might want to bump timeout a little higher for the sake of CI race test slowness.

ncabatoff:
Just to confirm my understanding: let's assume we have a steady stream of non-cacheable requests coming in for the exact same URL, say 1/ms, and they take 20ms each for Vault to process.

The first request to hit Agent will get added to the cache.  The next 19 requests will queue up behind it.  After 20ms the first request is done and Agent will then immediately forward all 19 other requests that came in while we were waiting.

As new requests come in every 1ms, each of them will be forwarded immediately since the inflight req ch is closed.

If there were an interruption in the flow of incoming requests long enough for the cache entry to be removed, the whole sequence would restart, with an initial delay, then all pending requests released at once, then no more delays.

This approach is motivated by the fact that we don't cache some requests, and we don't know at the outset whether a request will be cacheable.

I don't think it's bad, it's certainly an improvement over what we have now, and I don't have a better idea right now, but it does lead to some burstiness as well as somewhat unpredictable latency.

ncabatoff (approved):

calvn:
> As new requests come in every 1ms, each of them will be forwarded immediately since the inflight req ch is closed.

Yes, that's correct.

> If there were an interruption in the flow of incoming requests long enough for the cache entry to be removed, the whole sequence would restart, with an initial delay, then all pending requests released at once, then no more delays.

Yes, that's the case as well. Although, if there is no interrupt and this is a continuous steady stream of requests, the inflight cache entry might never get cleaned up. On the bright side, the delta here is just on the `remaining` counter and no an increase in the case size. In theory it means that it prevents the inflight cache from shrinking back down, but in practice I don't see this scenario to be large enough (numerous steady concurrent requests against a single agent instance) to overwhelm the cache/memory. You'd need to have a lot of these steady stream requests happening at the same time and on unique requests to populate and take up a large enough cache size for it to become an issue.

Good point on the burstiness aspect of this logic. I don't have a straight answer to this other than introducing a jitter, but this would also add additional latency. Maybe that's a fine trade-off to avoid overwhelming the Vault server.

ncabatoff:
> Good point on the burstiness aspect of this logic. I don't have a straight answer to this other than introducing a jitter, but this would also add additional latency. Maybe that's a fine trade-off to avoid overwhelming the Vault server.

I wasn't asking for any changes, just making sure we're clear on the new behaviour.  Like I say, I don't think is particularly bad, it's just potentially a bit surprising.  I vote against adding more complexity to the solution unless it improves the user experience.  I don't think there's a straightforward way to stagger the backlogged requests, at least I don't see a way to improve things in general - it feels like any attempt we make could help some in some cases and hurt in others.

command/agent/cache/lease_cache.go
```diff
@@ -170,40 +192,60 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
...
-		return nil, err
-	}
-	if sendResp != nil {
-		c.logger.Debug("returning cached response", "path", req.Request.URL.Path)
-		return sendResp, nil
-	}
+	defer func() {
+		// Cleanup on the cache if there are no remaining inflight requests.
+		// This is the last step, so we defer the call first
+		if inflight != nil && inflight.remaining.Load() == 0 {
```
kalafut, line 203: When would `inflight` be `nil` at this point?

command/agent/cache/lease_cache.go
```diff
@@ -170,40 +192,60 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
...
+			c.inflightCache.Delete(id)
```
kalafut, line 204: This felt a little racy since we can be here multiple times for the same id if a request comes in right after the condition `Load() == 0` is checked. But I've gone through it and don't think there is any harmful behavior. The `inflight` object still exists even if it's deleted from the cache so the final request can complete, and calling `Delete()` on an id not present is a no-op.

command/agent/cache/lease_cache.go
calvn: It _shouldn't_ be nil because the conditional further down always assigns this values to something, but I'm nil-checking just in case since this within a defer that's called right after the variable is declared but before the value is assigned.

calvn:
@ncabatoff @kalafut @briankassouf thoughts on having this backported to 1.6.x? It's technically a bug fix, but at the same time it's not a common case and not noticeable unless it's under a specific scenario. We're also tight on time for the upcoming patch release.

command/agent/cache/lease_cache.go
```diff
@@ -170,40 +192,60 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
...
+		inflight = inflightRaw.(*inflightRequest)
+		inflight.remaining.Inc()
+		defer inflight.remaining.Dec()
+
+		// If found it means that there's an inflight request being processed.
+		// We wait until that's finished before proceeding further.
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-inflight.ch:
```
briankassouf, line 227: So right now if we detect here that the thread that is doing the request has completed (channel has been closed) then we simply continue. But once we get down to:

```
cachedResp, err := c.checkCacheForRequest(id)
```
We'd see an nil cachedResp since that is still going to only cache leased values. Then, i think, we'd simply re-send the request to the Vault server. I think this fix is missing a step where we store the resulting request in the `inflightRequest` object and access it here when the channel is closed. Thoughts?

calvn: The winner who had closed the channel will have cached the response before this thread gets to call `c.checkCacheForRequest(id)` so it will result in a cache hit. In the case that the request resulted in a  non-cacheable response, it would proxy to Vault as it should. 

The changes in the PR don't actually prevent identical non-cacheable requests from being proxied to Vault; it simply allows one of the requests to be processed first (since we don't know if it's cacheable) before opening the floodgate to let other identical request to be processed concurrently. I don't think there's a need to store the actual request/response object in the `inflightRequest` .

PR #10709, opened 2021-01-15
Updating snowflake plugin to 0.1.1

Valarissa:
This PR seeks to update the Snowflake Secrets Engine plugin to v0.1.1, which provides support for hyphens in role names. Previously, having a hyphen in the role name would cause the hyphen to be passed to Snowflake as part of the username, which does not support that character in usernames.


calvn (approved):
Looks good! It's worth noting that hyphens in role names are valid, but will be replaced by underscores when the username is generated.

Related to https://github.com/hashicorp/vault-plugin-database-snowflake/pull/2

PR #10710, opened 2021-01-15
Clarify slash is needed on gcs and azure

mikegreen:
Clarify user question, unexpected behavior with no slash on gcs.


mikegreen:
This is a stop gap to 1199590306654091

ncabatoff (approved):

Issue #10701, opened 2021-01-14
Unable to create transit key with "create" capability in policy


jza201:
**Describe the bug**
We want to create a new policy for team members who can only create keys in transit engine, but cannot update existing keys which are created by others. The policy we have is:
```
path "transit/keys/*" {
    capabilities = [ "create" ]
}
```
However we got the permission error when running `vault write -f transit/keys/my-key`
```
Error writing data to transit/keys/my-key: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/transit/keys/my-key
Code: 403. Errors:

* 1 error occurred:
	* permission denied
```

**To Reproduce**

1. Login as root
2. Run `vault secrets enable transit`
3. Create a new policy 
```
path "transit/keys/*" {
    capabilities = [ "create" ]
}
```
4. Login with this new policy
5. Run `vault write -f transit/keys/my-key`

**Expected behavior**
Transit key is created. 

**Environment:**
* Vault Server Version: 1.4.2
* Vault CLI Version: 1.4.2
* Server Operating System/Architecture: Ubuntu 18.04.5 LTS (GNU/Linux 4.15.0-123-generic x86_64)

HridoyRoy:
Hi @jza201 , because the same API is used to create and update transit keys, you would need the associated policy to have the "update" capability as well. 

We could potentially separate out the create endpoint from the update endpoint and restrict access to those endpoints separately, but so far we haven't come across a use case where a party is allowed to create keys but not update them, so we haven't introduced that complexity. 

Please let me know if there's anything else we can help with, or if we can close out this issue. 
Thanks so much!

HridoyRoy:
Hi @jza201 , I'm reopening this issue for discussion, as making this change would, to the best of my knowledge, require an [existence check](https://github.com/hashicorp/vault/pull/10717) to be added on the create operation for transit keys. As such, this would require existing vault customers to change all policies that allow for the creation of transit keys with the update operation to have both create and update permissions.

jza201:
Thanks @HridoyRoy. The reason we would like to just allow users "create" capability is that we don't want users to change transit keys created by others. Let's say that user1 creates a transit key and use it to encrypt some text, then user2 updates the transit key and then user1 cannot decrypt with this transit key. I did some testing and found out that updating an existing transit key actually does cause the problem I mentioned above. So for now the need for the capabilities of both "create" and "update" work for us.

HridoyRoy:
[...]

PR #10717, opened 2021-01-18
Existence Check for transit key creation endpoint [VAULT-1352]

HridoyRoy:
This goes hand in hand with: 
https://github.com/hashicorp/vault/pull/10706

and:
https://github.com/hashicorp/vault/issues/10701

and has been tested. 

However, the issue is that with this change, the "create" permission is now necessary to create keys, whereas before only the "update" permission was necessary. This means the change will be breaking for existing customers that have "update" permissions for key creation. 

Currently, I'm not sure how to make this backward compatible so that "update" would function as both create and update simultaneously, or if that is even possible. So, we may want to defer this PR to a later release version. If we do decide to defer, I will roll back https://github.com/hashicorp/vault/pull/10706 as well.

### DIFF
--- a/builtin/logical/transit/path_keys.go
+++ b/builtin/logical/transit/path_keys.go
@@ -99,6 +99,7 @@ return the public key for the given context.`,
 
 		Callbacks: map[logical.Operation]framework.OperationFunc{
 			logical.UpdateOperation: b.pathPolicyWrite,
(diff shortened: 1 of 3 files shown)


HridoyRoy:
We've decided that this is a won't fix, since it would require existing customers to change their policies and would be a majorly breaking change. Going to go ahead and close this PR for now.

PR #10719, opened 2021-01-19
Benchmarking CIDRs bound configurations

bruj0:
Benchmarking how the number of cidrs bound to secret id generation affects the login operation in AppRole.


aphorise:
I do recall @bruj0 performing these tests to verify memory consumption and behaviour with varying sizes of CIDR requests during some related pursuits that we were engaged in then. If the tests are not useful or likely to be used then it would be best to close this PR.

PR #10743, opened 2021-01-21
make token create case insensitive [VAULT-1021]

HridoyRoy:
https://hashicorp.atlassian.net/browse/VAULT-1021

This allows for tokens creation to be case insensitive against allowed aliases. Previously there was a mix of case sensitivity and case insensitivity, so mixed case aliases were disallowed. Note that entity aliases will still retain their case sensitivity internally, but for every verification step, they will be checked in a case-insensitive way.


website/content/api-docs/auth/token.mdx
```diff
@@ -107,7 +107,11 @@ during this call.
 - `entity_alias` `(string: "")` - Name of the entity alias to associate with
   during token creation. Only works in combination with `role_name` argument
   and used entity alias must be listed in `allowed_entity_aliases`. If this has
-  been specified, the entity will not be inherited from the parent.
+  been specified, the entity will not be inherited from the parent. Note that 
+  `allowed_entity_aliases` is not case sensitive. For example, creating tokens 
+  with alias `aLiAs` and alias `alias` will both succeed even if the
+  `allowed_entity_aliases` list only allows for token creation for entity alias 
+  `alias`.
```
mgritter, line 114: I think this is a bit wordy, I would maybe note in allowed_entity_aliases that the matching is not case-sensitive.

mgritter (approved):
Looks good.

mgritter (approved):

website/content/api-docs/auth/token.mdx
HridoyRoy: Done!

PR #10745, opened 2021-01-22
Add documentation about the horizontal cluster scalability of PKI secret engine operations

sgmiller:


website/content/api-docs/secret/pki.mdx
```diff
@@ -1602,3 +1602,30 @@ $ curl \
     --data @payload.json \
     http://127.0.0.1:8200/v1/pki/tidy
 ```
+
+# Cluster Scalability
```
ncabatoff, line 1606: Is it worth also talking about generate_lease?

website/content/api-docs/secret/pki.mdx
```diff
@@ -1602,3 +1602,30 @@ $ curl \
...
+
+Most non-introspection operations in the PKI secrets engine require a write to 
+storage, and so are forwarded to the cluster's primary for execution.  This table
+outlines which operations can be executed on secondaries and thus scale
```
ncabatoff, line 1610: s/primary/active node/
s/secondaries/performance standbys/

website/content/api-docs/secret/pki.mdx
sgmiller: Ah, yes, good call.

website/content/api-docs/secret/pki.mdx
sgmiller: Still having trouble with that...

website/content/api-docs/secret/pki.mdx
ncabatoff: Actually, if generate_lease is set true, the perf standby will handle it.  It will do an RPC at the end to the active node to register the lease.  So it's kind of a halfway point between fully handled on the perf standby and not handled at all.

sgmiller: Noted that.

website/content/api-docs/secret/pki.mdx
```diff
@@ -1602,3 +1602,32 @@ $ curl \
...
+| ca_chain                    | Read                |
+| crl[/pem]                   | Read                |
+| crl/pem                     | Read                | 
+| issue                       | Update <sup>*</sup> |
+| revoked/*                   | Read, List          |
+| sign                        | Update <sup>*</sup> |
+| sign-verbatim               | Update <sup>*</sup> |
+
+\* Only if the corresponding role has `no_store` set to true and `generate_lease`
+set to false.  If `generate_lease` and `no_store` are true, the lease is forwarded
```
ncabatoff, line 1629: > If `generate_lease` and `no_store` are true

I don't think that's quite right.  How about "If `generate_lease` is true the lease creation will be forwarded to the active node; if `no_store` is false the entire request will be forwarded to the active node"?

sgmiller: Sounds good.

ncabatoff (approved):

Issue #10602, opened 2020-12-18
Vault agent not setting namespace correctly for requests


csquire:
**Describe the bug**
The namespace set in vault agent's auto-auth isn't being respected for requests. This was potentially caused by this PR put into 1.6.1.  https://github.com/hashicorp/vault/pull/10556

**To Reproduce**
1. Configure vault agent to auto-auth against a namespace other than root (I have tested approle and kubernetes auth, I was also rendering a template, if that makes a difference)
2. Do not set VAULT_NAMESPACE env var
3. Run the agent using Vault 1.6.1 (1.6.0 works fine)

**Expected behavior**
Expect that the agent gets my secret and renders the template, but instead I get 403s because the request is going to the root namespace and not the namespace specified in the auto-auth config.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.5.3
* Vault Agent Version (retrieve with `vault version`): 1.6.1
* Server Operating System/Architecture: Linux

Vault server configuration file(s):

```hcl
exit_after_auth = false
pid_file = "./pidfile"

vault {
  address = "https://vault.example.com"
}

cache {
  use_auto_auth_token = true
}

listener "tcp" {
  address = "0.0.0.0:8200"
  tls_disable = "true"
}

auto_auth {
  method "approle" {
    namespace = "mynamespace"
    config = {
      role_id_file_path = "roleid.txt"
      secret_id_file_path = "secretid.txt"
      remove_secret_id_file_after_reading = "false"
    }
  }
}

sink "file" {
  path = "token.txt"
}

template {
  source = "kv.tpl"
  destination = "kv.out"
}
```

**Additional context**
# Vault agent 1.6.0 works without VAULT_NAMESPACE #
```
~/testing/vault/agent ./vault160 agent --config config-test.hcl 
==> Vault agent started! Log data will stream in below:

==> Vault agent configuration:

          Api Address 1: http://0.0.0.0:8200
                    Cgo: disabled
              Log Level: info
                Version: Vault v1.6.0
            Version Sha: 7ce0bd9691998e0443bc77e98b1e2a4ab1e965d4

2020-12-18T15:06:38.524-0600 [INFO]  template.server: starting template server
2020-12-18T15:06:38.524-0600 [INFO]  auth.handler: starting auth handler
2020-12-18T15:06:38.525-0600 [INFO]  auth.handler: authenticating
2020-12-18T15:06:38.524-0600 [INFO]  sink.server: starting sink server
[...]

Issue #8534, opened 2020-03-11
Agent Auth Failure When Namespace in ENV and Config file


acilate:
**Describe the bug**
When starting vault in agent mode with a config file and directing it to auto-auth an approle, specifying the namespace in the configuration file as well as the environment causes an HTTP400/HTTP403 error (depending on if you're in windows or linux)

Specifying the namespace in only one place, ENV or Config file works as expected.

**To Reproduce**
1. export VAULT_NAMESPACE='my_namespace'
2. vault agent -config ./agent-config.hcl

```
==> Vault server started! Log data will stream in below:

==> Vault agent configuration:

                     Cgo: disabled
               Log Level: trace
                 Version: Vault v1.3.2

2020-03-11T11:07:44.214-0700 [INFO]  sink.file: creating file sink
2020-03-11T11:07:44.214-0700 [TRACE] sink.file: enter write_token: path=./token
2020-03-11T11:07:44.214-0700 [TRACE] sink.file: exit write_token: path=./token
2020-03-11T11:07:44.214-0700 [INFO]  sink.file: file sink configured: path=./token mode=-rw-r-----
2020-03-11T11:07:44.215-0700 [INFO]  auth.handler: starting auth handler
2020-03-11T11:07:44.215-0700 [INFO]  auth.handler: authenticating
2020-03-11T11:07:44.215-0700 [INFO]  template.server: starting template server
2020-03-11T11:07:44.215-0700 [INFO]  sink.server: starting sink server
2020-03-11T11:07:44.215-0700 [INFO]  template.server: no templates found
2020-03-11T11:07:44.215-0700 [INFO]  template.server: template server stopped
2020-03-11T11:07:45.256-0700 [ERROR] auth.handler: error authenticating: error="Error making API request.

URL: PUT https://vault.addr/v1/my_namespace/auth/approle/login
Code: 403. Errors:

* 1 error occurred:
	* permission denied

" backoff=1.158592927
```

**Expected behavior**
Specifying the namespace in two places should trigger precedence logic where one negates the other.  It appears that something is making this additive and producing an invalid request.

**Environment:**
* Vault Server Version (retrieve with `vault status`): Vault v1.3.2 (I don't operate the server, but this is what I'm told)
* Vault CLI Version (retrieve with `vault version`): Vault v1.3.2
* Server Operating System/Architecture: Linux x64

**Vault AGENT** configuration file(s):

```hcl
pid_file = "./vault-agent.pid"

vault {
[...]

PR #10757, opened 2021-01-25
Set namespace for template server in agent

vishalnayak:
Fixes https://github.com/hashicorp/vault/issues/10602

### DIFF
--- a/command/agent.go
+++ b/command/agent.go
@@ -371,6 +371,7 @@ func (c *AgentCommand) Run(args []string) int {
 		// Check if a default namespace has been set
 		mountPath := config.AutoAuth.Method.MountPath
 		if cns := config.AutoAuth.Method.Namespace; cns != "" {
+			namespace = cns
 			// Only set this value if the env var is empty, otherwise we end up with a nested namespace
 			if ens := os.Getenv(api.EnvVaultNamespace); ens == "" {
 				mountPath = path.Join(cns, mountPath)
--- a/changelog/10757.txt
+++ b/changelog/10757.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+agent: Set namespace for template server in agent.
+```


calvn (approved):

calvn:
You might also want to include the CL entry in this PR :)

raskchanky (approved):

command/agent.go
```diff
@@ -371,6 +371,7 @@ func (c *AgentCommand) Run(args []string) int {
 		// Check if a default namespace has been set
 		mountPath := config.AutoAuth.Method.MountPath
 		if cns := config.AutoAuth.Method.Namespace; cns != "" {
+			namespace = cns
```
ncabatoff, line 374: I wonder if we should incorporate the logic for EnvVaultNamespace here?  It's confusing to rely on the code in consult-template that consults the env when we don't specify a namespace explicitly.  Especially since we *do* look at the env to set mountPath.

ncabatoff:
Did you test to validate that #8534 is still fixed with this change?

ncabatoff:
We could've maybe avoided this regression if we'd adopted #9777 and used it to write automated tests against this part of the code.  It's not too late (well, it is for 1.6.2, but going forward...)

ncabatoff:
> We could've maybe avoided this regression if we'd adopted #9777 and used it to write automated tests against this part of the code. It's not too late (well, it is for 1.6.2, but going forward...)

Actually maybe not, it looks like the problematic area is outside the scope of what I tackled in that PR.  Nonetheless, it'd be nice to see more of agent's Run method moved into more testable places.

command/agent.go
vishalnayak: The EnvVaultNamespace logic here seems like it is only applicable to the auth subsystem of agent, and it seems like a sane thing to do. It is possible that we can refactor this to better manage the code, but I am feeling reluctant to touch it at this point in the release cycle.

vishalnayak:
Yes, I have verified that #8534 fix isn't broken. Used the script here to do it: https://github.com/hashicorp/vault-tools/blob/master/users/vishal/agent/agent_auto_auth_namespace.sh

command/agent.go
ncabatoff: It looks like by setting `namespace = cns`, if there is an env var set, the templating code will ignore it.

calvn:
@ncabatoff I plan on re-visiting #9777 shortly.

PR #10760, opened 2021-01-25
Change netlify to vercel in UI & storybook docs

jescalan:
As we have moved both of these apps over to vercel, this ensures the URLs in the readme are correct!


chelshaw (approved):
Thanks for tackling this!

PR #10766, opened 2021-01-25
PostgreSQL - Add username customization

pcman312:
Adds the ability to customize username generation for dynamic users in PostgreSQL.

Docs PR will come after all other updates for username customization are complete for this release.


plugins/database/postgresql/postgresql.go
```diff
@@ -427,10 +451,10 @@ func (p *PostgreSQL) defaultDeleteUser(ctx context.Context, username string) err
 
 	// can't drop if not all privileges are revoked
 	if rows.Err() != nil {
-		return errwrap.Wrapf("could not generate revocation statements for all rows: {{err}}", rows.Err())
+		return fmt.Errorf("could not generate revocation statements for all rows: %w", rows.Err())
```
calvn, line 454: Looks like we're [using](https://github.com/hashicorp/vault/blob/master/go.mod#L56) `errwrap` v1.0.0 which doesn't support upwrapping errors wrapped by `%w`. We should update go.mod first (probably as a separate PR) before merging this in.

pcman312: Updated

plugins/database/postgresql/postgresql.go
```diff
@@ -68,19 +72,46 @@ func new() *PostgreSQL {
...
+
+	up, err := template.NewTemplate(template.Template(usernameTemplate))
+	if err != nil {
+		return dbplugin.InitializeResponse{}, fmt.Errorf("unable to initialize username template: %w", err)
+	}
+	p.usernameProducer = up
+
+	_, err = p.usernameProducer.Generate(dbplugin.UsernameMetadata{})
+	if err != nil {
+		return dbplugin.InitializeResponse{}, fmt.Errorf("invalid username template - did you reference a field that isn't available? : %w", err)
```
calvn, line 98: Nit: does the error returned here tell you which field evaluated was invalid? I feel that `"invalid username template: %w"` might be sufficient here if so.

plugins/database/postgresql/postgresql.go
```diff
@@ -68,19 +72,46 @@ func new() *PostgreSQL {
...
+
 	resp := dbplugin.InitializeResponse{
 		Config: newConf,
 	}
 	return resp, nil
 }
 
+func getString(m map[string]interface{}, key string) string {
+	var result string
+	if err := mapstructure.Decode(m[key], &result); err != nil {
```
calvn, line 109: Can we access the map and do an ok check + type assertion here? `mapstructure.Decode` ends up calling `reflect` funcs which might be a bit heavy-handed for this.

plugins/database/postgresql/postgresql.go
```diff
@@ -8,14 +8,16 @@ import (
 	"strings"
 
 	"github.com/hashicorp/errwrap"
+
```
calvn, line 11:
```suggestion
```

plugins/database/postgresql/postgresql.go
pcman312: Example errors:

Input: `foo-{{random 5}}-{{.FooBar}}`
`unable to apply template: template: template:1:19: executing "template" at <.FooBar>: can't evaluate field FooBar in type dbplugin.UsernameMetadata`

Input: `foo-{{random 5}}-{{FooBar}}`
`unable to parse template: template: template:1: function "FooBar" not defined`

I'll look to see if the error that's returned is structured in any meaningful way so we can make this a bit nicer to users who aren't familiar with these errors.

pcman312: Unfortunately the errors coming back from the template library (the Go one, not the one in Vault) are not structured in a useful way here. I will remove this message and put a note in the docs about common error messages.

plugins/database/postgresql/postgresql.go
pcman312: Updated the `GetString` function to not use mapstructure. I also moved it into the `strutil` package as I find I have needed to use it in a bunch of places.

plugins/database/postgresql/postgresql.go
pcman312: Done

catsby (approved):
LGTM

calvn (approved):

PR #10772, opened 2021-01-26
Clarify agent lease renewal docs

kalafut:
Alternatively this could be changed to "1/3 remaining", but I think this is clearer?

### DIFF
--- a/website/content/docs/agent/template.mdx
+++ b/website/content/docs/agent/template.mdx
@@ -135,8 +135,8 @@ high level overview of different behaviors.
 
 ### Renewable Secrets
 
-If a secret or token is renewable, Vault Agent will renew the secret at 2/3
-of the secret's lease duration.
+If a secret or token is renewable, Vault Agent will renew the secret after 2/3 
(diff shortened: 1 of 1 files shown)


website/content/docs/agent/template.mdx
```diff
@@ -135,8 +135,8 @@ high level overview of different behaviors.
 
 ### Renewable Secrets
 
-If a secret or token is renewable, Vault Agent will renew the secret at 1/3
-of the secret's lease.
+If a secret or token is renewable, Vault Agent will renew the secret at 2/3
```
calvn, line 138: Maybe something like this?

> ... Vault Agent will renew the secret after 2/3 of the secret's lease duration has elapsed.

calvn (approved):
Small suggestion to make it a bit more explicit but the way it is now is fine as well.

website/content/docs/agent/template.mdx
kalafut: Updated

PR #10774, opened 2021-01-26
Track barrier encryptions and rotate when an operation threshold or time limit is reached.

sgmiller:
One note on the design here: Checking when to rotate is done as encryptions
happen, to avoid needing a goroutine whose lifecycle we'd need to manage. 
But as a result, this requires the AESGCMBarrier have access to the random
generator.  I did this via a factory pattern, but it's in contrast to the
existing SecurityBarrier interface which passes in the random reader to all
calls that need it.


vault/barrier_aes_gcm.go
```diff
@@ -75,18 +86,40 @@ type AESGCMBarrier struct {
...
+
+	randomReaderAccessor func() io.Reader
+	RotationConfig       *configutil.BarrierConfig
 }
 
 // NewAESGCMBarrier is used to construct a new barrier that uses
 // the provided physical backend for storage.
-func NewAESGCMBarrier(physical physical.Backend) (*AESGCMBarrier, error) {
+func NewAESGCMBarrier(physical physical.Backend, rotConfig *configutil.BarrierConfig, randomReaderAccessor func() io.Reader, l log.Logger) (*AESGCMBarrier, error) {
+	// Should we have minimums?
```
HridoyRoy, line 98: I don't think we need minimums, do we? We could probably give guidance on what to set these values to.

vault/barrier_aes_gcm.go
```diff
@@ -1049,3 +1106,40 @@ func (b *AESGCMBarrier) Keyring() (*Keyring, error) {
...
+	}
+	// Increment the local encryption count, and track metrics
+	ops := atomic.AddInt64(&keyring.LocalEncryptions, 1)
+	metrics.IncrCounterWithLabels(barrierEncryptsMetric, 1, termLabel(term))
+
+	// time.Now is expensive, don't check time based rotation on every operation
+	if b.shouldRotate(ops, keyring) {
+		b.l.Lock()
+		defer b.l.Unlock()
+		// Retest the precondition
```
HridoyRoy, line 1123: Why does the precondition need to be retested?

vault/barrier_aes_gcm.go
```diff
@@ -1049,3 +1106,40 @@ func (b *AESGCMBarrier) Keyring() (*Keyring, error) {
...
+			go b.autorotateBarrierKey(ctx)
+		}
+	}
+	return ct, nil
+
+}
+
+func (b *AESGCMBarrier) shouldRotate(ops int64, keyring *Keyring) bool {
+	return ops > b.RotationConfig.KeyRotationMaxOperations ||
+		(b.RotationConfig.KeyRotationInterval > 0 && ops%keyRotationTimeSampleRate == 0 &&
```
HridoyRoy, line 1134: Should keyRotationTimeSampleRate be configurable via the rotation config?

HridoyRoy (changes requested):

internalshared/configutil/barrier.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// Get our one item
+	item := list.Items[0]
+
+	if result.Barrier == nil {
+		result.Barrier = &BarrierConfig{}
+	}
+
+	if err := hcl.DecodeObject(&result.Barrier, item.Val); err != nil {
+		return multierror.Prefix(err, "telemetry:")
```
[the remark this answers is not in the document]
sgmiller: This will go away after merging the other PR, but the answer is copy paste fail.

internalshared/configutil/barrier.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	if result.Barrier.KeyRotationIntervalRaw != nil {
+		var err error
+		if result.Barrier.KeyRotationInterval, err = parseutil.ParseDurationSecond(result.Barrier.KeyRotationIntervalRaw); err != nil {
+			return err
+		}
+		result.Barrier.KeyRotationIntervalRaw = nil
```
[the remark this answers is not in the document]
sgmiller: Should disappear as well, but HCL doesn't support durations, so there is this "raw" field that gets populated from the config, then you parse into the typed field for use in code and clear the raw field.

vault/barrier_aes_gcm.go
sgmiller: The config param disappears with the api based config, but yeah, I'll add a comment on what the reader can/should be.

vault/barrier_aes_gcm.go
sgmiller: There is a potential race here where two encryptions happen close enough together that they both want to trigger rotation.  If you don't retest you can rotate twice.

vault/barrier_aes_gcm.go
sgmiller: No, that's just an internal configuration.  Trying to pick a value that would avoid calling time.Now() (which requires a system call) on each encryption.  Constant is there because the unit tests use it to do enough writes to trigger the rotation.

vault/barrier_aes_gcm.go
sgmiller: This got cleaned up as I brought over the changes made in ENT to here.

sgmiller:
Just a note to review this on the ENT branch.  This will be updated with the OSS portions after review feedback/change iteration.

PR #10776, opened 2021-01-26
CHANGELOG update: GCP secrets WAL issue fix

Valarissa:


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ initialized=true when a Raft join is in progress. [[GH-10498](https://github.com
 * secrets/database/influxdb: Fix issue where not all errors from InfluxDB were being handled [[GH-10384](https://github.com/hashicorp/vault/pull/10384)]
 * secrets/database/mysql: Fixes issue where the DisplayName within generated usernames was the incorrect length [[GH-10433](https://github.com/hashicorp/vault/pull/10433)]
 * secrets/database: Sanitize `private_key` field when reading database plugin config [[GH-10416](https://github.com/hashicorp/vault/pull/10416)]
+* secrets/gcp: Fix issue with account and iam_policy roleset WALs not being removed after attempts when GCP project no longer exists [[GH-10759](https://github.com/hashicorp/vault/pull/10759)]
```
pcman312, line 31: Should this be in https://github.com/hashicorp/vault/tree/master/changelog ?

CHANGELOG.md
```diff
@@ -43,7 +44,7 @@ initialized=true when a Raft join is in progress. [[GH-10498](https://github.com
 SECURITY:
 
 * LDAP Auth Method: We addressed an issue where error messages returned by the
-  LDAP auth methold allowed user enumeration [[GH-10537](https://github.com/hashicorp/vault/pull/10537)]. This vulnerability affects Vault OSS and Vault 
+  LDAP auth method allowed user enumeration [[GH-10537](https://github.com/hashicorp/vault/pull/10537)]. This vulnerability affects Vault OSS and Vault
```
pcman312, line 47: Same question re: https://github.com/hashicorp/vault/blob/release/1.6.x/changelog/10537.txt

Though for this one 1.6.1 is already released so I'm unsure of the correct approach here.

Valarissa: I didn't create that one, but, sure, I can add it. I didn't even know this changelog folder was a thing...

calvn:
The entry _might_ be overwritten by go-changelog when that's auto-gen'ed. @mladlow is it fine to append entries like this directly or should it be always though a CL file?

mladlow:
The note in 1.6.1 is good - will not be overridden, and is how we need to update old changelogs. The one for 1.7 will need a changelog file, it will be overridden when I do the changelog for 1.7. @Valarissa there's a wiki page describing the changelog process, and you probably have a link to the RFC in your email.

Valarissa:
Just updated this one.

CHANGELOG.md
calvn: We can remove this entry now that it's in `changelog/10759.txt`.

calvn (approved):
Small comment, otherwise looks good!

calvn (approved):

PR #10792, opened 2021-01-27
More CL notes for 1.6.2

mladlow:


changelog/_2021Jan26.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:security
+Limited Unauthenticated Remove Peer: As of Vault 1.6, the remove-peer command
+on DR secondaries did not require authentication. This issue impacts the
+stability of HA architecture, as a bad actor could remove all peers from a DR
```
vishalnayak, line 4: The term "all peers" might give a notion of bringing down all the nodes. We may want to mention something like "all standby nodes".

vishalnayak (approved):

PR #10797, opened 2021-01-27
Add base64 function to string template library

pcman312:
Adds `base64` function to string template library. This function base64 encodes the provided string similar to the existing `sha256` function.


Valarissa (approved):
LGTM

PR #10805, opened 2021-01-28
Document identity behaviour on local auth mounts.

ncabatoff:


vishalnayak (approved):

website/content/docs/secrets/identity.mdx
```diff
@@ -81,6 +81,12 @@ these identity providers. Hence, the alias name in combination with the
 authentication backend mount's accessor, serve as the unique identifier of an
 alias.
 
+### Local Auth Methods
+
+If a login is performed using an auth method that was mounted locally on a
+performance secondary (i.e. using `local=true`), no entities or groups will
```
mgritter, line 87: I believe this is true for local mounts on the primary as well.

ncabatoff: Whoops, good catch.  https://github.com/hashicorp/vault/pull/10807

PR #10807, opened 2021-01-29
Note that all local mounts are impacted, including on primary.

ncabatoff:


website/content/docs/secrets/identity.mdx
```diff
@@ -84,8 +84,8 @@ alias.
 ### Local Auth Methods
 
 If a login is performed using an auth method that was mounted locally on a
```
mgritter, line 86:
```suggestion
If a login is performed using an auth method that was mounted locally
```

mgritter (approved):

PR #10809, opened 2021-01-29
Use the exponential backoff lib for auth retries

sgmiller:


command/agent/auth/auth.go
```diff
@@ -126,7 +128,7 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
 		}
 
 		// Create a fresh backoff value
```
ncabatoff, line 130: Add a Reset once we reach LifetimeWatcherLoop?

command/agent/auth/auth.go
```diff
@@ -3,6 +3,7 @@ package auth
 import (
 	"context"
 	"errors"
+	"github.com/cenkalti/backoff/v3"
```
calvn, line 6: Looks like there's a v4 out, is there a reason we're pinning to v3?

command/agent/auth/auth.go
```diff
@@ -126,7 +128,7 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
 		}
 
 		// Create a fresh backoff value
-		backoff := 2*time.Second + time.Duration(ah.random.Int63()%int64(time.Second*2)-int64(time.Second))
+		backoff := eb.NextBackOff()
```
kalafut, line 131: Looks like this call can return `backoff.Stop` (-1), though maybe only if we don't configure `DefaultMaxElapsedTime`?

calvn: I'd assume that `NewExponentialBackOff` uses the `DefaultMaxElapsedTime` value unless otherwise specified, which means that this would return -1 after 15 minutes.

calvn:
Are we planning on going with the library's default set of values or do we want to expose them via the config?

sgmiller:
Closing this for now, since we didn't need it yet.  We should take it on properly with config in the future.

PR #10817, opened 2021-02-01
Add docs for OpenLDAP dynamic secrets

pcman312:
Adds documentation for OpenLDAP dynamic secrets (https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/21)

### DIFF
--- a/website/content/api-docs/secret/openldap.mdx
+++ b/website/content/api-docs/secret/openldap.mdx
@@ -266,8 +266,8 @@ Active Directory domain user account.
 
 | Method   | Path                       |
 | :------- | :------------------------- |
-| `POST`   | `openldap/role/:role_name` |
-| `DELETE` | `openldap/role/:role_name` |
+| `POST`   | `/openldap/role/:role_name` |
+| `DELETE` | `/openldap/role/:role_name` |
 
 The `POST` endpoint allows for partial updates of existing roles. If a role exists and a `POST` request is made
 against it, only the keys specified in the request will be updated. To delete a value, specify the key with an
@@ -355,7 +355,7 @@ $ curl \
 
 | Method | Path                       |
 | ------ | -------------------------- |
-| `GET`  | `openldap/role/:role_name` |
+| `GET`  | `/openldap/role/:role_name` |
 
 Retrieves a dynamic role's configuration.
 
@@ -386,8 +386,7 @@ and password within an LDIF statement) are not known at configuration time. Addi
 lot of flexibility to the operator to construct the LDIF and username.
 
 Templated fields are delimited by `{{` and `}}`. To reference a field (such as the generated `Username`), a period is
-placed in front of the field name. Example: `{{.Username}}`. This operates very similarly to
-[`jq`](https://stedolan.github.io/jq/). Spaces can be included between `{{` and `}}`. For instance:
+placed in front of the field name. Example: `{{.Username}}`. Spaces can be included between `{{` and `}}`. For instance:
 `{{.Username|lowercase}}` is the same as `{{ .Username | lowercase }}`.
 
 If a field needs to be modified (such as SHA256 hashing, base64 encoding, etc.) the value can be sent to one of the
@@ -445,7 +444,11 @@ the dashes with underscores. See [Template Functions](#template-functions) for m
 
 #### Template Functions
 
-Both the LDIF templates as well as the username template support using the following functions:
+Both the LDIF templates and the username template use the [Go template language](https://golang.org/pkg/text/template)
+so all [functions](https://golang.org/pkg/text/template/#hdr-Functions) and capabilities from that language are
+supported including functions such as `printf`.
+
+In addition to the functions available through the template language, the following custom functions are also available:
 
 `random` - Generates a random string from lowercase letters, uppercase letters, and numbers. Must include a
 number indicating how many characters to generate.<br />
@@ -455,13 +458,25 @@ number indicating how many characters to generate.<br />
 **Example:** `{{.FieldName | truncate 10}}`
 
 `truncate_sha256` - Truncates the previous value to the provided number of characters. The last 8 characters of the
-truncated value will be the first 8 characters of the SHA256 hash of the truncated characters.<br />
-**Example:** `{{.FieldName | truncate_sha256 20}}`. If `FieldName` is `abcdefghijklmnopqrstuvwxyz`, all characters after
-the 12th (`l`) are removed and SHA256 hashed (`872808ffbf0158b200d397a9884342497f23b153b1630229014cea1886ca6f20`.
-The first 8 characters of the hash (`872808ff`) are then appended to the end of the first 12 characters from the
-original value: `abcdefghijkl872808ff`.
+truncated value will be the first 8 characters of the SHA256 hash of the truncated characters.
 
-`uppercase` - Uppercases the provided value.
+**Example:** `{{.FieldName | truncate_sha256 20}}`.<br />
+If `FieldName` is `abcdefghijklmnopqrstuvwxyz`, all characters after the 12th (`l`) are removed and SHA256 hashed
+(`872808ffbf0158b200d397a9884342497f23b153b1630229014cea1886ca6f20`. The first 8 characters of the hash (`872808ff`)
+are then appended to the end of the first 12 characters from the original value: `abcdefghijkl872808ff`.
+
+This can aid in identifying values (such as the role & display names) if they need to be truncated to a particular
+length, particularly if the value has common prefixes. An example of this is having many roles with a common prefix in
+the name of the role, but the role name is truncated such that only the prefix is shown. This function will keep the
+non-prefix part of the role name in the username which will aid in identification while also keeping uniqueness.
+
+**Example Scenario:**<br />
+Two roles exist: `mylongprefix-foobar` and `mylongprefix-bazqux`. If the template is
+`v_{{.RoleName | truncate 12}}_{{unix_time}}` the username would be `v_mylongprefix_1234567890` for both of the roles.
+However, if the template is `v_{{.RoleName | truncate_sha256 12}}_{{unix_time}}` the username for the first role
+would be: `v_mylo998a317d_1234567890` and the username for the second role would be: `v_mylo224e847c_1234567890`.
+
+`uppercase` - Uppercases the provided value.<br />
 **Example:** `{{.FieldName | uppercase}}`
 
 `lowercase` - Lowercases the provided value.<br />
@@ -479,7 +494,7 @@ original value: `abcdefghijkl872808ff`.
 `unix_time` - The current unix timestamp (number of seconds since Jan 1 1970).<br />
 **Example:** `{{unix_time}}`
 
-`unix_time_millis` - The current unix timestamp in milliseconds.
+`unix_time_millis` - The current unix timestamp in milliseconds.<br />
 **Example:** `{{unix_time_millis}}`
 
 `timestamp` - The current time. Must provide a formatting string based on Go’s time package.<br />
--- a/website/content/api-docs/secret/openldap.mdx
+++ b/website/content/api-docs/secret/openldap.mdx
@@ -8,8 +8,8 @@ description: This is the API documentation for the Vault OpenLDAP secrets engine
 # OpenLDAP Secrets Engine (API)
 
 This is the API documentation for the Vault OpenLDAP secrets engine. For general
-information about the usage and operation of
-the OpenLDAP secrets engine, please see [these docs](/docs/secrets/openldap).
+information about the usage and operation of the OpenLDAP secrets engine,
+please see [these docs](/docs/secrets/openldap).
 
 This documentation assumes the OpenLDAP secrets engine is enabled at the `/openldap` path
 in Vault. Since it is possible to mount secrets engines at any path, please
@@ -149,7 +149,7 @@ The `static-role` endpoint configures Vault to manage the passwords of existing 
   The minimum rotation period is 5 seconds.<br />
   **Example:** "5s", "1h".
 - `username` `(string: <required>)` - The name of the user to be used when logging in. This is useful when `dn`
-  isn't used for login purposes (such as SSH).
+  isn't used for login purposes (such as SSH).<br />
   **Example:** "bob"
 
 | Method   | Path                               |
@@ -240,7 +240,7 @@ $ curl \
 
 ### Manually Rotate Static Role Password
 
-The `rotate-role` endpoint offers manual rotation of static roles.
+The `rotate-role` endpoint rotates the password of an existing static role.
 
 | Method | Path                               |
 | :----- | :--------------------------------- |
--- a/website/content/api-docs/secret/openldap.mdx
+++ b/website/content/api-docs/secret/openldap.mdx
@@ -30,12 +30,13 @@ to search and change entry passwords in OpenLDAP.
 
 ### Parameters
 
-- `binddn` `(string: <required>)` - Distinguished name (DN) of object to bind for managing user entries.
-  Example: `cn=vault,ou=Users,dc=hashicorp,dc=com`
+- `binddn` `(string: <required>)` - Distinguished name (DN) of object to bind for managing user entries.<br />
+  **Example:** `cn=vault,ou=Users,dc=hashicorp,dc=com`
 - `bindpass` `(string: <required>)` - Password to use along with `binddn` for managing user entries.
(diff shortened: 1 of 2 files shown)


website/content/api-docs/secret/openldap.mdx
```diff
@@ -119,17 +120,37 @@ $ curl \
...
+### Parameters
 
-Distinguished name (DN) of entry Vault should manage. Example: `cn=bob,ou=Users,dc=hashicorp,dc=com`
+- `dn` `(string: <required>)` - Distinguished name (DN) of entry Vault should manage.<br />
+  **Example:** `cn=bob,ou=Users,dc=hashicorp,dc=com`
+- `rotation_period` `(string: <required>)` - How often Vault should rotate the password of the user entry, in seconds.
+  The minimum rotation period is 5 seconds.<br />
+  **Example:** "5s", "1h".
+- `username` `(string: <required>)` - The name of the user to be used when logging in. This is useful when `dn`
+  isn't used for login purposes (such as SSH).
```
calvn, line 152: FWIW you can also force a newline by adding an empty newline instead of using <br /> (though that is less clear and may be accidentally cleaned up later)

```suggestion
  isn't used for login purposes (such as SSH).<br />
```

website/content/api-docs/secret/openldap.mdx
```diff
@@ -216,36 +238,257 @@ $ curl \
 }
 ```
 
-## Rotate Root Password
+### Manually Rotate Static Role Password
 
-The `rotate-root` endpoint offers password rotation for the `binddn` entry used to manage OpenLDAP. This generated password will only be known to Vault and will not be retrievable once rotated.
+The `rotate-role` endpoint offers manual rotation of static roles.
```
calvn, line 243:
```suggestion
The `rotate-role` endpoint offers manual password rotation of static roles.
```

website/content/api-docs/secret/openldap.mdx
```diff
@@ -119,17 +120,37 @@ $ curl \
...
+    --request POST \
+    http://127.0.0.1:8200/v1/openldap/rotate-root
+```
+
+## Static Roles
 
-- `dn` `(string: <required>)` - Distinguished name (DN) of entry Vault should manage. Example: `cn=bob,ou=Users,dc=hashicorp,dc=com`
-- `rotation_period` `(string: <required>)` - How often Vault should rotate the password of the user entry, in seconds. The minimum rotation period is 5 seconds. Example: "5s", "1h".
-- `username` `(string: <required>)` - The name of the user to be used when logging in. This is useful when `dn` isn't used for login purposes (such as SSH). Example: "bob".
+The `static-role` endpoint configures Vault to manage the passwords of existing individual OpenLDAP entries.
```
calvn, line 142:
```suggestion
The `static-role` endpoint configures Vault to manage the passwords of existing individual OpenLDAP or Active Directory domain user accounts.
```

website/content/docs/secrets/openldap.mdx
```diff
@@ -172,11 +175,125 @@ only Vault will know.
...
+objectClass: person
+objectClass: organizationalPerson
+objectClass: user
+userPrincipalName: {{.Username}}@adtesting.lab
+sAMAccountName: {{.Username}}
+
+dn: CN={{.Username}},OU=HashiVault,DC=adtesting,DC=lab
+changetype: modify
+replace: unicodePwd
+unicodePwd::{{ printf "\"%s\"" .Password | utf16le | base64 }}
```
calvn, line 284: They can use `%q` to quote the string right?

```suggestion
unicodePwd::{{ printf "%q" .Password | utf16le | base64 }}
```

pcman312: `%q` returns single quoted values, not double quotes ones so that would be different =/

calvn: > %q	a double-quoted string safely escaped with Go syntax

`%q` should return double-quoted strings: https://play.golang.org/p/3ZObvmbBMAM

pcman312: I stand corrected. The docs state `a single-quoted character literal safely escaped with Go syntax.` which looks like it's wrong.

calvn: It looks like that's the case (single quoted) if the value is an integer, but it's double quotes for strings and byte slices.

website/content/api-docs/secret/openldap.mdx
pcman312: Adding an extra newline here doesn't add a `<br />`, it creates a new `<p>` tag which will have significantly more spacing:
Screen Shot 2021-02-03 at 1 25 15 PM
(the bottom one has a newline, while the top one has `<br />`)

website/content/api-docs/secret/openldap.mdx
pcman312: Updated, though with different phrasing

website/content/api-docs/secret/openldap.mdx
pcman312: I'm hesitant to add this here as it really should be noted in the main docs that Active Directory is compatible. Otherwise we can pretty easily end up with a situation where we have to say `OpenLDAP (or Active Directory)` all over the API docs.

website/content/docs/secrets/openldap.mdx
pcman312: Integers don't appear to come out as quoted string representations of integers: https://play.golang.com/p/dhRW2zYyb1l
```
Raw value: [](string)
  => [""]
Raw value: [string](string)
  => ["string"]
Raw value: [77](int)
  => ['M']
Raw value: [77](int)
  => ['M']
Raw value: [3892](int)
  => ['༴']
Raw value: [77](float64)
  => [%!q(float64=77)]
Raw value: [true](bool)
  => [%!q(bool=true)]
```

website/content/api-docs/secret/openldap.mdx
calvn: Good point, `<br />` makes more sense then!

website/content/api-docs/secret/openldap.mdx
calvn: Depending on the config's schema, static roles also support AD right? Only suggesting this because the dynamic role section mentions AD, so this makes it sound like static role only supports OpenLDAP.

> This provides instructions to Vault on how to create an OpenLDAP or Active Directory domain user account.

website/content/docs/secrets/openldap.mdx
calvn: Isn't `.Password | utf16le | base64` always going to result in a string though? In this example `%q` should work fine.

pcman312: Yes it will return a string. I'm not disagreeing with that point. I was just pointing out that `%q` is not a "string quoting" verb.

calvn: Gotcha. That's right, `%q` is not specific to string quoting and behaves differently depending on the value's type.

website/content/api-docs/secret/openldap.mdx
```diff
@@ -119,17 +120,37 @@ $ curl \
...
+
+### Parameters
 
-Distinguished name (DN) of entry Vault should manage. Example: `cn=bob,ou=Users,dc=hashicorp,dc=com`
+- `dn` `(string: <required>)` - Distinguished name (DN) of entry Vault should manage.<br />
+  **Example:** `cn=bob,ou=Users,dc=hashicorp,dc=com`
+- `rotation_period` `(string: <required>)` - How often Vault should rotate the password of the user entry, in seconds.
```
kalafut, line 148: Maybe revise or remove `in seconds`, especially since one of the examples is "1h".

website/content/api-docs/secret/openldap.mdx
```diff
@@ -216,36 +238,257 @@ $ curl \
...
-| :----- | :--------------------------------- |
-| `POST` | `/openldap/rotate-role/:role_name` |
+### Create/Delete Dynamic Role Configuration
 
-### Sample Request
+#### Parameters
+
+| Method   | Path                       |
+| :------- | :------------------------- |
+| `POST`   | `openldap/role/:role_name` |
```
kalafut, line 269:
```suggestion
| `POST`   | `/openldap/role/:role_name` |
```

website/content/api-docs/secret/openldap.mdx
```diff
@@ -216,36 +238,257 @@ $ curl \
...
+| `DELETE` | `openldap/role/:role_name` |
```
kalafut, line 270:
```suggestion
| `DELETE` | `/openldap/role/:role_name` |
```

website/content/api-docs/secret/openldap.mdx
```diff
@@ -216,36 +238,257 @@ $ curl \
...
+### Templates
+
+LDIF and username templates use the [Go template syntax](https://golang.org/pkg/text/template/) to construct the
+LDIF/username that will be executed against the server. This is done because certain values (such as the username
+and password within an LDIF statement) are not known at configuration time. Additionally, the template gives a
+lot of flexibility to the operator to construct the LDIF and username.
+
+Templated fields are delimited by `{{` and `}}`. To reference a field (such as the generated `Username`), a period is
+placed in front of the field name. Example: `{{.Username}}`. This operates very similarly to
+[`jq`](https://stedolan.github.io/jq/). Spaces can be included between `{{` and `}}`. For instance:
```
kalafut, line 390: Not sure about the `jq` reference, which might either: a) be unfamiliar  b) lead readers to think at other jq-ism will work here.

website/content/api-docs/secret/openldap.mdx
```diff
@@ -216,36 +238,257 @@ $ curl \
...
+Both the LDIF templates as well as the username template support using the following functions:
+
+`random` - Generates a random string from lowercase letters, uppercase letters, and numbers. Must include a
+number indicating how many characters to generate.<br />
+**Example:** `{{random 20}}` generates 20 random characters
+
+`truncate` - truncates the previous value to the provided number of characters.<br />
+**Example:** `{{.FieldName | truncate 10}}`
+
+`truncate_sha256` - Truncates the previous value to the provided number of characters. The last 8 characters of the
```
kalafut, line 457: It might be helpful to describe when this function is useful.

website/content/api-docs/secret/openldap.mdx
```diff
@@ -216,36 +238,257 @@ $ curl \
...
+truncated value will be the first 8 characters of the SHA256 hash of the truncated characters.<br />
+**Example:** `{{.FieldName | truncate_sha256 20}}`. If `FieldName` is `abcdefghijklmnopqrstuvwxyz`, all characters after
+the 12th (`l`) are removed and SHA256 hashed (`872808ffbf0158b200d397a9884342497f23b153b1630229014cea1886ca6f20`.
+The first 8 characters of the hash (`872808ff`) are then appended to the end of the first 12 characters from the
+original value: `abcdefghijkl872808ff`.
+
+`uppercase` - Uppercases the provided value.
```
kalafut, line 464: Some of these end in `<br />`, but some don't?

website/content/docs/secrets/openldap.mdx
```diff
@@ -172,11 +176,125 @@ only Vault will know.
...
+objectClass: person
+objectClass: organizationalPerson
+objectClass: user
+userPrincipalName: {{.Username}}@adtesting.lab
+sAMAccountName: {{.Username}}
+
+dn: CN={{.Username}},OU=HashiVault,DC=adtesting,DC=lab
+changetype: modify
+replace: unicodePwd
+unicodePwd::{{ printf "%q" .Password | utf16le | base64 }}
```
kalafut, line 285: I didn't see `printf` in the list of functions above. If it is the case that any of the Go template functions are also available, it is probably good to highlight that fact at the start of the template function list and link to the list of functions in Go docs.

kalafut (commented):
Looking really good! Very comprehensive. Just a few minor comments.

website/content/api-docs/secret/openldap.mdx
pcman312: :+1:

website/content/api-docs/secret/openldap.mdx
pcman312: :+1:

website/content/api-docs/secret/openldap.mdx
pcman312: That's a good point. I've removed the note about `jq`

website/content/api-docs/secret/openldap.mdx
pcman312: Updated to have more details. I also included a more concrete scenario example. Let me know how this reads.

website/content/api-docs/secret/openldap.mdx
pcman312: They all should to put the `Example` on the next line. Looks like I missed a couple =/

website/content/docs/secrets/openldap.mdx
pcman312: It is available through the template language (not a custom one from us). I'll add a note about this.

website/content/api-docs/secret/openldap.mdx
pcman312: Wordsmithed to be similar to the `default_ttl` and `max_ttl` variables.

website/content/api-docs/secret/openldap.mdx
```diff
@@ -216,36 +240,274 @@ $ curl \
...
+This can aid in identifying values (such as the role & display names) if they need to be truncated to a particular
+length, particularly if the value has common prefixes. An example of this is having many roles with a common prefix in
+the name of the role, but the role name is truncated such that only the prefix is shown. This function will keep the
+non-prefix part of the role name in the username which will aid in identification while also keeping uniqueness.
+
+**Example Scenario:**<br />
+Two roles exist: `mylongprefix-foobar` and `mylongprefix-bazqux`. If the template is
+`v_{{.RoleName | truncate 12}}_{{unix_time}}` the username would be `v_mylongprefix_1234567890` for both of the roles.
+However, if the template is `v_{{.RoleName | truncate_sha256 12}}_{{unix_time}}` the username for the first role
+would be: `v_mylo998a317d_1234567890` and the username for the second role would be: `v_mylo224e847c_1234567890`.
```
calvn, line 481: The two examples can be combined into one to convey the reasoning.

```suggestion
This can aid in identifying values (such as the role & display names) if they need to be truncated to a particular
length, particularly if the value has common prefixes. An example of this is having many roles with a common prefix in
the name of the role, but the role name is truncated such that only the prefix is shown. This function will keep the
non-prefix part of the role name in the username which will aid in identification while also keeping uniqueness.

**Example:**<br />
`v_{{.RoleName | truncate_sha256 12}}_{{unix_time}}`. <br />
If two roles exist for this template, `mylongprefix-foobar` and `mylongprefix-bazqux`, the username for the first role
would be `v_mylo998a317d_1234567890` and the username for the second role would be `v_mylo224e847c_1234567890`.
```

pcman312: :+1: with some tiny formatting changes

website/content/api-docs/secret/openldap.mdx
pcman312: Updated to hopefully clarify this. I've removed the "or Active Directory" part of this line and tried to make the static & dynamic sections a bit more consistent in showing their support for AD.

website/content/docs/secrets/openldap.mdx
```diff
@@ -172,11 +176,125 @@ only Vault will know.
...
+  - You will likely also want to disable AD's password expiration for this dynamic user account. The
+`userAccountControl` value for this is: `65536`
+  - `userAccountControl` flags are cumulative, so to set both of the above two flags, add up the two values
+(`512 + 65536 = 66048`): set `userAccountControl` to `66048`
+  - See [here](https://docs.microsoft.com/en-us/troubleshoot/windows-server/identity/useraccountcontrol-manipulate-account-properties#property-flag-descriptions)
+for details on `userAccountControl` flags
+
+`sAMAccountName` is a common field when working with AD users. It is used to provide compatibility with legacy
+Windows NT systems and has a limit of 20 characters. Keep this in mind when defining your `username_template`.
+See [here](fhttps://docs.microsoft.com/en-us/windows/win32/adschema/a-samaccountname) for additional details.
```
calvn, line 258:
```suggestion
See [here](https://docs.microsoft.com/en-us/windows/win32/adschema/a-samaccountname) for additional details.
```

pcman312: :+1:

calvn (approved):

kalafut (approved):

PR #10824, opened 2021-02-02
Add environment variable for bearer auth token

hixichen:
Our vault instances are behind company internal network with edge gateway that requires a valid bearer token. 
Right now, we have to ssh into jump box to run the script with Vault.
It would be great that if Vault supports adding auth bearer token environment into http request header.


VinnyHC:
In Vault 1.9 there are capabilities to define custom headers in the configuration, which I think will get you what you need! Depending on how you start Vault you can feed in an environment variable.
Quick snippet:
``` hcl
...

listener "tcp" {
  address     = "127.0.0.1:8200"
  custom_response_headers {
    "default" = {
      "Authorization" = ["Bearer ${BEARER_TOKEN}"],
    }
  }
}

...
```
Check out this tutorial [Create Customized HTTP Headers](https://learn.hashicorp.com/tutorials/vault/customize-http-headers) for a more complete solution.
I'm going to close this PR but feel free to re-open if I misunderstood your request.

PR #10826, opened 2021-02-03
approle: add ttl to the secret ID generation response

calvn:
This PR adds `ttl` to the response on the [secret ID generation endpoint](https://www.vaultproject.io/api-docs/auth/approle#generate-new-secret-id). This allows external parties, such as consul-template, to parse and perform renew logic based on this value.

### DIFF
--- a/builtin/credential/approle/path_role.go
+++ b/builtin/credential/approle/path_role.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math"
 	"strings"
 	"time"
 
@@ -2363,20 +2362,8 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
 		Data: map[string]interface{}{
 			"secret_id":          secretID,
 			"secret_id_accessor": secretIDStorage.SecretIDAccessor,
+			"secret_id_ttl":      int64(b.deriveSecretIDTTL(secretIDStorage.SecretIDTTL).Seconds()),
 		},
-	}
-
-	// Follow the behavior from registerSecretIDEntry and set the response TTL
-	// to the entry's TTL unless it's negative or greater than the system's max
-	// lease TTL.
-	switch {
-	case secretIDStorage.SecretIDTTL < time.Duration(0):
-		// This case should not be possible since TypeDurationSecond does not
-		// support negative values, but we follow the same convention as
-		// registerSecretIDEntry
-		resp.Data["ttl"] = int(b.System().MaxLeaseTTL().Seconds())
-	case secretIDStorage.SecretIDTTL > time.Duration(0):
-		resp.Data["ttl"] = int(math.Min(secretIDStorage.SecretIDTTL.Seconds(), b.System().MaxLeaseTTL().Seconds()))
 	}
 
 	return resp, nil
--- a/builtin/credential/approle/path_role.go
+++ b/builtin/credential/approle/path_role.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


builtin/credential/approle/path_role.go
```diff
@@ -2358,12 +2359,27 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
...
+	// Follow the behavior from registerSecretIDEntry and set the response TTL
+	// to the entry's TTL unless it's negative or greater than the system's max
+	// lease TTL.
+	switch {
+	case secretIDStorage.SecretIDTTL < time.Duration(0):
+		// This case should not be possible since TypeDurationSecond does not
+		// support negative values, but we follow the same convention as
+		// registerSecretIDEntry
+		resp.Data["ttl"] = int(b.System().MaxLeaseTTL().Seconds())
+	case secretIDStorage.SecretIDTTL > time.Duration(0):
```
calvn, line 2378: I wonder if we should always be returning a value, i.e. return `"ttl": 0` on non-expiring secret IDs.

calvn: Also, should this be `ttl` (conforms to other TTL values on other backends) or `secret_id_ttl` (conforms to the other values in this particular response)?

builtin/credential/approle/path_role.go
```diff
@@ -2358,12 +2359,27 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
...
 		Data: map[string]interface{}{
 			"secret_id":          secretID,
 			"secret_id_accessor": secretIDStorage.SecretIDAccessor,
 		},
-	}, nil
+	}
+
+	// Follow the behavior from registerSecretIDEntry and set the response TTL
+	// to the entry's TTL unless it's negative or greater than the system's max
+	// lease TTL.
```
briankassouf, line 2371: If `registerSecretIDEntry` is setting the TTL on the entry do we need to duplicate the logic here or will secretIDStorage.SecretIDTTL already reflect the calculated TTL?

calvn: `registerSecretIDEntry` calculates the expiration time (no duration) based off the `SecretIDTTL`, but it doesn't modify it which is why we need to check and cap it outside of the function.

calvn: I'll extract the logic to a helper func that can be called in both places.

builtin/credential/approle/path_role.go
```diff
@@ -2358,12 +2359,27 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
...
+	switch {
+	case secretIDStorage.SecretIDTTL < time.Duration(0):
+		// This case should not be possible since TypeDurationSecond does not
+		// support negative values, but we follow the same convention as
+		// registerSecretIDEntry
+		resp.Data["ttl"] = int(b.System().MaxLeaseTTL().Seconds())
```
kalafut, line 2377: Do we want to use `secret_id_ttl` instead? My thoughts are that since this isn't a standard lease, and we're just relaying configuration details, it might be better to be different. Also, since we're going to look for these details in consul-template, it could reduce the chance of unintentionally picking up some other secret as AppRole.

calvn: Done in 8b90861.

builtin/credential/approle/path_role.go
calvn: Done in 8b90861.

builtin/credential/approle/path_role.go
calvn: Both comments addressed in 8b90861. @kalafut I've changed it to always return `secret_id_ttl` such that the zero-value indicates non-expiry. This is the same as how other zero-value TTL responses behave throughout Vault.

kalafut (approved):

PR #10828, opened 2021-02-03
Add user search filter to ldap auth engine

ixe013:
Allow the user to specify an search filter for users, similar to the one that exists for groups. Enables a number of use cases that are not possible today (as of Vault 1.6.2), such as searching for:

- Active users only
- Users by either `uid` or `email`
- Users with a manager
- Users with a custom object class

Code is backward compatible is the proposed `searchfilter` parameter is not set.


ixe013:
Closing because 

- I don't have the rights to restart the build
- Found a typo in the documentation https://vault-q4zey2s0l.vercel.app/docs/auth/ldap#searchfilter
- Will add the new field to the UI

PR #10833, opened 2021-02-03
Set TokenParent in the Index to be cached

tvoran:
It seems like TokenParent should be set on the index about to be stored, instead of the entry that was looked up, so that when a token is revoked, it's possible to lookup all the child tokens when the parent is revoked:

https://github.com/hashicorp/vault/blob/464f2138302dc52585e7b3b97fb4d740be5385cc/command/agent/cache/lease_cache.go#L768-L781

### DIFF
--- a/command/agent/cache/lease_cache_test.go
+++ b/command/agent/cache/lease_cache_test.go
@@ -162,6 +162,18 @@ func TestLeaseCache_SendCacheable(t *testing.T) {
 		t.Fatalf("expected getting proxied response: got %v", diff)
 	}
 
+	// Check TokenParent
+	cachedItem, err := lc.db.Get(cachememdb.IndexNameToken, "testtoken")
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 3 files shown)


command/agent/cache/lease_cache_test.go
```diff
@@ -162,6 +162,18 @@ func TestLeaseCache_SendCacheable(t *testing.T) {
...
 
+	// Check TokenParent
+	cachedItem, err := lc.db.Get(cachememdb.IndexNameToken, "testtoken")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if cachedItem == nil {
+		t.Fatalf("expected token entry from cache")
+	}
+	if "autoauthtoken" != cachedItem.TokenParent {
```
calvn, line 173: Nit:
```suggestion
	if cachedItem.TokenParent != "autoauthtoken" {
```

calvn (approved):
Nice find!

PR #8040, opened 2019-12-17
Fix MySQL Plugin password special character escape bug

michelvocks:
Fixes #7834

A few months ago, we merged https://github.com/hashicorp/vault/pull/7089 which fixed a bug for PostgreSQL but introduced a new bug for the MySQL database plugin.

PR #10834, opened 2021-02-03
MySQL - Add username customization

pcman312:
Adds the ability to customize username generation for dynamic users in MySQL.


plugins/database/mysql/mysql.go
```diff
@@ -85,6 +81,40 @@ func (m *MySQL) Initialize(ctx context.Context, req dbplugin.InitializeRequest)
 	if err != nil {
 		return dbplugin.InitializeResponse{}, err
 	}
+
+	usernameTemplate, err := getString(req.Config, "username_template")
```
calvn, line 85: You can probably switch to use `strutil.GetString` from your other PR now that exists.

helper/testhelpers/mysql/mysqlhelper.go
```diff
@@ -40,7 +40,7 @@ func PrepareTestContainer(t *testing.T, legacy bool, pw string) (func(), string)
 
 	svc, err := runner.StartService(context.Background(), func(ctx context.Context, host string, port int) (docker.ServiceConfig, error) {
 		hostIP := docker.NewServiceHostPort(host, port)
-		connString := fmt.Sprintf("root:%s@(%s)/mysql?parseTime=true", pw, hostIP.Address())
+		connString := fmt.Sprintf("root:%s@tcp(%s)/mysql?parseTime=true", pw, hostIP.Address())
```
calvn, line 43: Curious, was this broken before?

plugins/database/mysql/connection_producer.go
```diff
@@ -108,20 +107,6 @@ func (c *mySQLConnectionProducer) Init(ctx context.Context, conf map[string]inte
 		mysql.RegisterTLSConfig(c.tlsConfigName, tlsConfig)
 	}
 
-	// Set initialized to true at this point since all fields are set,
```
calvn, line 111: Why are we remove this code block from `Init`? Since this func deals with the connection logic, it makes sense to keep this verification within `mySQLConnectionProducer.Init` instead of moving it into `MySQL.Initialize`.

plugins/database/mysql/mysql.go
```diff
@@ -26,44 +26,40 @@ const (
...
 	mySQLTypeName = "mysql"
-)
 
-var (
-	MetadataLen       int = 10
-	LegacyMetadataLen int = 4
-	UsernameLen       int = 32
-	LegacyUsernameLen int = 16
+	DefaultUserNameTemplate       = `{{ printf "v-%s-%s-%s-%s" (.DisplayName | truncate 10) (.RoleName | truncate 10) (random 20) (unix_time) | truncate 32 }}`
+	DefaultLegacyUserNameTemplate = `{{ printf "v-%s-%s-%s" (.RoleName | truncate 4) (random 20) (unix_time) | truncate 16 }}`
```
calvn, line 31: The behavior here looks to be the same as what we had, but now I'm wondering if the `(unix_time)` part ever got printed since `(random 20)` already takes 20 chars, which is well over the 16 char limit on this legacy template.

plugins/database/mysql/mysql.go
pcman312: 👍 that was the plan!

helper/testhelpers/mysql/mysqlhelper.go
pcman312: No it wasn't broken. I only did this to match what comes out from `config.FormatDSN()` in the MySQL library itself.

plugins/database/mysql/connection_producer.go
pcman312: Because it sets the `Initialized` variable when it hasn't finished initializing. In working on the next DB I realized I could move this back and shuffle some other logic around to make that happy without changing this.

plugins/database/mysql/mysql.go
pcman312: Nope. Never did. Thanks MySQL for having ridiculous username restrictions. I just made it match the existing behavior rather than try to simplify it.

plugins/database/mysql/connection_producer.go
calvn: Can we restore this deleted godoc line?

plugins/database/mysql/connection_producer.go
```diff
@@ -10,12 +10,13 @@ import (
 	"sync"
 	"time"
 
-	"github.com/go-sql-driver/mysql"
-	"github.com/hashicorp/errwrap"
-	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/vault/sdk/database/helper/connutil"
 	"github.com/hashicorp/vault/sdk/database/helper/dbutil"
 	"github.com/hashicorp/vault/sdk/helper/parseutil"
+
```
calvn, line 16: Revert import ordering?

plugins/database/mysql/mysql.go
calvn: I'd be okay to remove the `(unix_time)` bit since someone glancing at this might think that there would be a timestamp in there even though it's never the case. The context of leaving this because we had it like so will be lost over time.

plugins/database/mysql/mysql_test.go
```diff
@@ -483,35 +718,6 @@ func TestMySQL_UpdateUser(t *testing.T) {
 	}
 }
 
-func TestMySQL_Initialize_ReservedChars(t *testing.T) {
```
calvn, line 486: This test was added for [coverage against a bug](https://github.com/hashicorp/vault/pull/8040) that we ended up fixing and is not specific to templating, but rather a check against [this codeblock](https://github.com/hashicorp/vault/blob/master/sdk/database/helper/connutil/sql.go#L66-L69). We might want to keep it around.

calvn (approved):
Few suggestions, otherwise 👍 !

plugins/database/mysql/connection_producer.go
pcman312: Fixed

plugins/database/mysql/connection_producer.go
pcman312: Done. I missed adding it back in previously =/

plugins/database/mysql/mysql.go
pcman312: Removed

plugins/database/mysql/mysql_test.go
pcman312: This is being checked in the `TestMySQL_Initialize` table tests as the password being used is the same as in this tests. I'll see about making this check more explicit though. I'd like to keep it in the table tests, but TBD on whether that's reasonable.

calvn: Looks like you've addressed this on 716ddc4, thanks!

calvn (approved):

PR #10848, opened 2021-02-05
Updating date-fns library from 1.x to 2.x

Monkeychip:
We need to update this library to take advantage of the new methods available in version 2.^ for the Database Secret Engine project.

Worked through the breaking changes according to the date-fns [changelog](https://github.com/date-fns/date-fns/blob/master/CHANGELOG.md#changed-5), but also issues that came up in the failed test.  For example, AM/PM from `A` to `a`.

### DIFF
--- a/ui/app/routes/vault/cluster/metrics/index.js
+++ b/ui/app/routes/vault/cluster/metrics/index.js
@@ -1,25 +1,26 @@
 import Route from '@ember/routing/route';
 import ClusterRoute from 'vault/mixins/cluster-route';
 import { hash } from 'rsvp';
-import { format, addDays } from 'date-fns';
+import { getTime } from 'date-fns';
 import { parseDateString } from 'vault/helpers/parse-date-string';
 
 const getActivityParams = ({ start, end }) => {
-  // Expects MM-YYYY format
+  // Expects MM-yyyy format
   // TODO: minStart, maxEnd
   let params = {};
   if (start) {
     let startDate = parseDateString(start);
     if (startDate) {
       // TODO: Replace with formatRFC3339 when date-fns is updated
-      params.start_time = format(addDays(startDate, 1), 'X');
+      // converts to milliseconds, divide by 1000 to get epoch
+      params.start_time = getTime(startDate) / 1000;
     }
   }
   if (end) {
     let endDate = parseDateString(end);
     if (endDate) {
       // TODO: Replace with formatRFC3339 when date-fns is updated
-      params.end_time = format(addDays(endDate, 10), 'X');
+      params.end_time = getTime(endDate) / 1000;
     }
   }
   return params;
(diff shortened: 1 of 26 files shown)


ui/app/routes/vault/cluster/metrics/index.js
```diff
@@ -1,25 +1,26 @@
...
+  // Expects MM-yyyy format
   // TODO: minStart, maxEnd
   let params = {};
   if (start) {
     let startDate = parseDateString(start);
     if (startDate) {
       // TODO: Replace with formatRFC3339 when date-fns is updated
-      params.start_time = format(addDays(startDate, 1), 'X');
+      // converts to milliseconds, divide by 1000 to get epoch
+      params.start_time = getTime(startDate) / 1000;
```
chelshaw, line 16: Pretty sure we added days because there was some weirdness with UTC, but if the tests pass with this update we should be ok

ui/app/templates/components/license-info.hbs
```diff
@@ -54,7 +54,7 @@
       <div class="field box is-fullwidth is-shadowless is-paddingless is-marginless">
         <InfoTableRow @label="License ID" @value={{licenseId}} />
         <InfoTableRow @label="Valid from" @value={{startTime}}>
-          {{date-format model.startTime 'MMM DD, YYYY hh:mm:ss A'}} to {{date-format expirationTime 'MMM DD, YYYY hh:mm:ss A'}}
+          {{date-format startTime 'MMM dd, yyyy hh:mm:ss a'}} to {{date-format expirationTime 'MMM dd, yyyy hh:mm:ss a'}}
```
chelshaw, line 57: Just double checking that removing `model` intentionally here

ui/tests/integration/helpers/date-from-now-test.js
```diff
@@ -12,6 +12,6 @@ module('Integration | Helper | date-from-now', function(hooks) {
 
   test('you can include a suffix', function(assert) {
     let result = dateFromNow([1481022124443], { addSuffix: true });
-    assert.ok(result.includes(' ago'));
+    assert.ok(result.includes('about'));
```
chelshaw, line 15: This is a bit of a red flag -- maybe let's double check that the `addSuffix` option is being correctly passed to the update underlying date-fns function

chelshaw (changes requested):
A couple things to double-check, but thanks for making all these updates!

ui/app/templates/components/license-info.hbs
Monkeychip: yes, this was strange.  It threw an error, and I double-checked and the param is passed into the component as @startTime = {{model.startTime}}.

ui/tests/integration/helpers/date-from-now-test.js
Monkeychip: It's because the method underneath changed so it changes its wording.  I double-checked on that one.  The old method was deprecated and we moved to formatDistanceToNow, which has is a little smoother with the returns.

ui/tests/integration/helpers/date-from-now-test.js
```diff
@@ -14,4 +17,31 @@ module('Integration | Helper | date-from-now', function(hooks) {
     let result = dateFromNow([1481022124443], { addSuffix: true });
     assert.ok(result.includes(' ago'));
   });
+
+  test('you can pass in UTC timestamp', function(assert) {
```
chelshaw, line 21: 👏

chelshaw (approved):
Great work! Thanks for tackling this

PR #10850, opened 2021-02-05
agent: allow auto-auth to use an existing token

jasonodonnell:
This PR changes auto-auth to accept a token during the creation of the auth handler. This feature will be used with persistent caching where a token is decrypted from the cache file and passed into auto-auth. Reusing an existing token is critical to maintaining leases previously owned by the token loaded from the cache.

Currently this feature only attempts to load the token on the first run and falls back to auto-auth if that's unsuccessful. Retry logic could be added if valuable, but likely if there's an issue loading retrying won't help.

Once the token is no longer renewable, the renewer will fall back to auto-auth and reauthenticate.

To test this feature you can run the unit tests:
```bash
$ cd ./command/agent
$ go test -run 'TestTokenPreload*'
```


command/agent/auth/auth.go
```diff
@@ -169,12 +177,58 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
...
+				ah.logger.Debug("attempting to read preloaded token from path", "path", tokenPath)
+				tokenRaw, err := ioutil.ReadFile(tokenPath)
+				if err != nil {
+					return fmt.Errorf("error opening token file: %s", err)
+				}
+				token = string(tokenRaw)
+			} else {
+				ah.logger.Debug("attempting to read preloaded token from env VAULT_TOKEN")
+				token = os.Getenv("VAULT_TOKEN")
+			}
```
tvoran, line 199: I curious about the order here, since preferring ~/.vault-token over VAULT_TOKEN seems reversed from how the vault cli behaves. i.e. VAULT_TOKEN seems to always take precedence over a token helper like ~/.vault-token.

tvoran (commented):
The tests and example bootstrap script work for me. Just a couple questions.

command/agent/auth/auth.go
```diff
@@ -169,12 +177,58 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
...
+			tokenPath := fmt.Sprintf("%s/.vault-token", home)
+
+			if fileExists(tokenPath) {
+				ah.logger.Debug("attempting to read preloaded token from path", "path", tokenPath)
+				tokenRaw, err := ioutil.ReadFile(tokenPath)
+				if err != nil {
+					return fmt.Errorf("error opening token file: %s", err)
+				}
+				token = string(tokenRaw)
```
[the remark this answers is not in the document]
jasonodonnell: No longer applicable since we now pass in the token with `NewAuthHandler`.

command/agent/auth/auth.go
jasonodonnell: No longer applicable since we now pass in the token.

command/agent/auth/auth.go
```diff
@@ -169,12 +174,38 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
...
+		if first && ah.token != "" {
+			ah.logger.Debug("using preloaded token")
+
+			first = false
+			ah.logger.Debug("lookup-self with preloaded token")
+			clientToUse.SetToken(ah.token)
+
+			secret, err = clientToUse.Logical().Read("auth/token/lookup-self")
+			if err != nil {
+				ah.logger.Debug("could not look up token", "err", err, "backoff", backoff.Seconds())
```
calvn, line 187: This debug seems redundant since there's an error log right after.

command/agent/auth/auth.go
```diff
@@ -169,12 +174,38 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
...
 		}
 
-		secret, err := clientToUse.Logical().Write(path, data)
-		// Check errors/sanity
-		if err != nil {
-			ah.logger.Error("error authenticating", "error", err, "backoff", backoff.Seconds())
-			backoffOrQuit(ctx, backoff)
-			continue
+		var secret *api.Secret = new(api.Secret)
+		if first && ah.token != "" {
```
calvn, line 178: Do we want to do this before `am.Authenticate(ctx, ah.client)`?

This change also makes me wonder whether we should disable sink writing back the token if it's a token that is being preloaded.

command/agent/auth/auth.go
jasonodonnell: The debug actually shows the error, but the other does not. If you don't suppress the error you get a multi-line error in the logs.

command/agent/auth/auth.go
jasonodonnell: > Do we want to do this before am.Authenticate(ctx, ah.client)?

I don't think so, actually. `am.Authenticate` will catch misconfigurations of auto-auth in which case we'll want to bail. Thoughts?

> This change also makes me wonder whether we should disable sink writing back the token if it's a token that is being preloaded.

The sink should still be shared, even if its a preloaded token, because it will need to be rewrapped. Additionally the sink is used in acceptance testing and is a good validation that the token flowed through auto-auth. Thoughts?

command/agent/auth/auth.go
calvn: I think that's fine. IIRC there are already places where we log the error as-is, even if they are multi-line errors. Knowing why it failed at the error level is useful anyways so there's no reason to hide it under debug (and you're already outputting the error directly on the `Logical().Write` call down below).

jasonodonnell: Removed!

command/agent/auth/auth.go
calvn: For the case of approle auto-auth, if the token is being pre-loaded why would we want to check whether the role ID file exists first (within `Authenticate`)? If we're preloading the token, we can probably use that and then fallback to the auto-auth's method of getting the token if that fails.

General config parsing and validation should already be done by the `config` package.

> The sink should still be shared, even if its a preloaded token, because it will need to be rewrapped.

IIRC the token that the sink drops does not have to necessarily be wrapped (i.e. the raw token is written), and it's [off by default](https://www.vaultproject.io/docs/agent/autoauth#wrap_ttl) unless explicitly set to a non-empty value.

jasonodonnell: > For the case of approle auto-auth, if the token is being pre-loaded why would we want to check whether the role ID file exists first (within Authenticate)? If we're preloading the token, we can probably use that and then fallback to the auto-auth's method of getting the token if that fails.

We want to verify its OK because we expect at some point the preloaded token is no longer usable (max ttl) and we will then need to switch to auto-auth. We should ensure that the everything is configured correctly before that happens to avoid errors that may not appear for a long time.

> IIRC the token that the sink drops does not have to necessarily be wrapped (i.e. the raw token is written), and it's off by default unless explicitly set to a non-empty value.

That's true but I still don't see the advantage of disabling it. Less code/special cases to handle and the existing behavior doesn't change, even though where the token came from did.

calvn: > We should ensure that the everything is configured correctly before that happens to avoid errors that may not appear for a long time.

This is true, though this doesn't mitigate the case where the data returned by `Authenticate` will be out-of-date for the `Logical().Write` call below, when that gets consumed. I'm just slightly concerned by the gap between `Authenticate` and the actual call to auth since this sit in between (same point as you brought up, but from a different angle).

> That's true but I still don't see the advantage of disabling it.

I was thinking about this from a security perspective, since you're now threading/dropping a token that was not necessarily auth'ed by this agent instance. You're right though -- sink file access remains unchanged and that's transparent to apps that consume it, so probably not a big deal if we write the token to sinks.

command/agent/auth/auth.go
```diff
@@ -169,12 +201,14 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
 			}
 		}
 
-		secret, err := clientToUse.Logical().Write(path, data)
-		// Check errors/sanity
-		if err != nil {
-			ah.logger.Error("error authenticating", "error", err, "backoff", backoff.Seconds())
-			backoffOrQuit(ctx, backoff)
-			continue
+		if secret.Auth == nil {
```
calvn, line 204: Can we include a comment here on when this case gets hit?

calvn (approved):
One small note, otherwise looks good!

command/agent/auth/auth.go
jasonodonnell: Done!

command/agent/auth/auth.go
jasonodonnell: Made some changes and works as expected. 👍

tvoran (approved):

command/agent/auth/auth.go
```diff
@@ -151,6 +151,38 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
...
+			ah.logger.Debug("using preloaded token")
+
+			first = false
+			ah.logger.Debug("lookup-self with preloaded token")
+			clientToUse.SetToken(ah.token)
+
+			secret, err = clientToUse.Logical().Read("auth/token/lookup-self")
+			if err != nil {
+				ah.logger.Error("could not look up token", "err", err, "backoff", backoff.Seconds())
+				backoffOrQuit(ctx, backoff)
```
tomhjp, line 165: I appreciate it's an inherited pattern, but is it worth considering putting `backoffOrQuit()` calls at the beginning of the loop behind a `if !first` check? It would simplify `continue` blocks, and reduce both duplication and the potential for bugs where future developers forget to call `backoffOrQuit`.

tomhjp (approved):
LGTM, just one suggestion to simplify control logic around backoffs.

command/agent/auth/auth.go
jasonodonnell: Honestly this code doesn't change much and is likely going to be refactored when agent code is decoupled from Vault, so I think we'll defer this for now.

PR #10853, opened 2021-02-08
docs: Update Azure AD OIDC documentation

### LINKED ISSUE #8191: Clearer AzureAD Auth/External Group documentation?
I just set up AzureAD authentication via OIDC, with External Groups, and found the docs somewhat confusing/multiple places to look and pieces of information to pull together.

Would it be worth having a central place for that - maybe adding some additional details to the jwt_oidc_providers page?

I also got the AzureAD piece working with UPN, which is more "standard" for us anyway than email, and requires a few additional steps that I'm happy to document if people think it'd be helpful.

cc @jescalan

joatmon08:
Resolves #8191 with clearer external groups configuration.


austingebauer (approved):
Looks great! Thanks for doing this. I had a todo for updating the Azure-specific handling section to use the Microsoft Graph API instead of AAD Graph API, but you beat me to it 👍

PR #10856, opened 2021-02-08
Autopilot: Server Stabilization, State and Dead Server Cleanup

vishalnayak:


vault/logical_system_raft.go
```diff
@@ -335,6 +382,112 @@ func (b *SystemBackend) handleStorageRaftSnapshotRead() framework.OperationFunc
...
+			configClone.ServerStabilizationTime = time.Duration(serverStabilizationTime.(int)) * time.Second
+			persist = true
+		}
+
+		if configClone.CleanupDeadServers && configClone.MinQuorum < 3 {
+			return logical.ErrorResponse("min_quorum must be set when cleanup_dead_servers is set and it should at least be 3"), logical.ErrInvalidRequest
+		}
+
+		if persist {
+			entry, err := logical.StorageEntryJSON("core/raft/autopilot/configuration", configClone)
```
ncabatoff, line 480: When you do the ent PR, don't forget to add this to ignoredPrefixes so a primary's config doesn't overwrite a secondary's.

vault/logical_system_raft.go
```diff
@@ -335,6 +382,116 @@ func (b *SystemBackend) handleStorageRaftSnapshotRead() framework.OperationFunc
...
+	}
+}
+
+func (b *SystemBackend) handleStorageRaftAutopilotConfigRead() framework.OperationFunc {
+	return func(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
+		raftStorage, ok := b.Core.underlyingPhysical.(*raft.RaftBackend)
+		if !ok {
+			return logical.ErrorResponse("raft storage is not in use"), logical.ErrInvalidRequest
+		}
+		config := raftStorage.AutopilotConfig()
```
ncabatoff, line 421: Can this be nil?  Do we want to give a different response if disabled?

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,551 @@
...
+func (b *RaftBackend) GetAutopilotServerHealth(ctx context.Context) (*AutopilotHealth, error) {
+	b.l.RLock()
+	defer b.l.RUnlock()
+
+	if b.raft == nil {
+		return nil, errors.New("raft storage is not initialized")
+	}
+
+	if b.autopilot == nil {
+		return nil, nil
```
ncabatoff, line 509: I wonder if we should treat this as an error?

ncabatoff (commented):
Lots of minor comments, looking good overall!

api/sys_raft.go
```diff
@@ -27,6 +30,35 @@ type RaftJoinRequest struct {
...
 
+// AutopilotHealth represents the response of the raft autopilot health API
+type AutopilotHealth struct {
+	Healthy                    bool                       `json:"healthy" mapstructure:"healthy"`
+	FailureTolerance           int                        `json:"failure_tolerance" mapstructure:"failure_tolerance"`
+	OptimisticFailureTolerance int                        `json:"optimistic_failure_tolerance" mapstructure:"optimistic_failure_tolerance"`
+	Servers                    map[string]AutopilotServer `json:"servers" mapstructure:"servers"`
+	Leader                     string                     `json:"leader" mapstructure:"leader"`
+	Voters                     []string                   `json:"voters" mapstructure:"voters"`
+	ReadReplicas               []string                   `json:"read_replicas,omitempty" mapstructure:"read_replicas"`
```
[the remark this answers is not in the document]
vishalnayak: Fixed.

command/commands.go
```diff
@@ -359,6 +359,21 @@ func initCommands(ui, serverCmdUi cli.Ui, runOpts *RunOptions) {
...
+			return &OperatorRaftAutopilotGetConfigCommand{
+				BaseCommand: getBaseCommand(),
+			}, nil
+		},
+		"operator raft autopilot set-config": func() (cli.Command, error) {
+			return &OperatorRaftAutopilotSetConfigCommand{
+				BaseCommand: getBaseCommand(),
+			}, nil
+		},
+		"operator raft autopilot health": func() (cli.Command, error) {
```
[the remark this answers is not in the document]
vishalnayak: Fixed.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,551 @@
...
+		MaxTrailingLogs:         ac.MaxTrailingLogs,
+		MinQuorum:               ac.MinQuorum,
+		ServerStabilizationTime: ac.ServerStabilizationTime,
+	}
+
+}
+
+// MarshalJSON makes the autopilot config fields JSON compatible
+func (ac *AutopilotConfig) MarshalJSON() ([]byte, error) {
+	return json.Marshal(map[string]interface{}{
```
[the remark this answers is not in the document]
vishalnayak: You are right. We don't need the JSON tags. It was needed before, but then I had to add the custom marshaller and the tags were not removed.

vault/logical_system_raft.go
vishalnayak: Done

physical/raft/raft_autopilot.go
vishalnayak: It need not be. For nodes where autopilot is not yet enabled, I wanted the API be a no-op.

vault/logical_system_raft.go
vishalnayak: It can be nil. Fixed. API will silently return nothing.

api/sys_raft.go
```diff
@@ -27,6 +30,35 @@ type RaftJoinRequest struct {
...
+	ID          string            `json:"id" mapstructure:"id"`
+	Name        string            `json:"name" mapstructure:"name"`
+	Address     string            `json:"address" mapstructure:"address"`
+	NodeStatus  string            `json:"node_status" mapstructure:"node_status"`
+	LastContact string            `json:"last_contact" mapstructure:"last_contact"`
+	LastTerm    uint64            `json:"last_term" mapstructure:"last_term"`
+	LastIndex   uint64            `json:"last_index" mapstructure:"last_index"`
+	Healthy     bool              `json:"healthy" mapstructure:"healthy"`
+	StableSince string            `json:"stable_since" mapstructure:"stable_since"`
+	ReadReplica bool              `json:"read_replica" mapstructure:"read_replica"`
```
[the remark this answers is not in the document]
vishalnayak: Fixed.

briankassouf (commented):
I think we should switch the config logic around so that we have a default config that turns server stabilization on, dead server cleanup off, and allows the Config file or API to override the defaults.

Also we'll need to figure out the auditing issue

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,556 @@
...
+		return nil, fmt.Errorf("expected a single block of autopilot config")
+	}
+
+	return configs[0], nil
+}
+
+// StartAutopilot puts autopilot subsystem to work. This should only be called
+// on the active node.
+func (b *RaftBackend) StartAutopilot(ctx context.Context) {
+	if b.autopilot == nil {
```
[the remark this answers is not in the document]
vishalnayak: Yep. Fixed.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,556 @@
...
+		}
+	}
+}
+
+// FetchServerStats is called by the autopilot library to retrieve information
+// about all the nodes in the raft cluster.
+func (d *Delegate) FetchServerStats(ctx context.Context, servers map[raft.ServerID]*autopilot.Server) map[raft.ServerID]*autopilot.ServerStats {
+	ret := make(map[raft.ServerID]*autopilot.ServerStats)
+
+	followerStates := d.RaftBackend.followerStates
```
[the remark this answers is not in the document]
vishalnayak: It isn't safe to access this without holding the lock. Fixed it.
Regarding the nil check for followerStates, as it will never be nil.

vishalnayak:
> I think we should switch the config logic around so that we have a default config that turns server stabilization on, dead server cleanup off, and allows the Config file or API to override the defaults.

Addressed ☝️

command/operator_raft_autopilot_set_config.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (c *OperatorRaftAutopilotSetConfigCommand) Flags() *FlagSets {
+	set := c.flagSet(FlagSetHTTP | FlagSetOutputFormat)
+
+	f := set.NewFlagSet("Common Options")
+
+	f.BoolVar(&BoolVar{
+		Name:    "cleanup-dead-servers",
+		Target:  &c.flagCleanupDeadServers,
+		Default: false,
```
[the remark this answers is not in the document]
vishalnayak: I do share your thoughts. The only wrinkle I have here is that we don't have a way to know if a flag was set or not. This particularly affects if the flag is a bool like the `flagCleanupDeadServers`. Say the value was set to true previously; now if CLI tries to set false, how would the user do it? Not supplying the flag will be considered an explicit intent to not touch the value and we end up not flipping the value. Whereas by having default values, each time CLI command is invoked, all the values are set.

I am not sure if there is a way to know if the flag was set or not, and if I am missing it.

command/operator_raft_autopilot_get_config.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (c *OperatorRaftAutopilotGetConfigCommand) Synopsis() string {
+	return "Returns the configuration of the autopilot subsystem under integrated storage"
+}
+
+func (c *OperatorRaftAutopilotGetConfigCommand) Help() string {
+	helpText := `
+Usage: vault operator raft autopilot get-config
+
+  Returns the configuration of the autopilot subsystem under integrated storage.
```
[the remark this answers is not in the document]
vishalnayak: Fixed.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+}
+
+// NotifyState is called by the autopilot library whenever there is a state
+// change. We update a few metrics when this happens.
+func (d *Delegate) NotifyState(state *autopilot.State) {
+	d.logger.Info("notify state called")
+	if d.raft.State() == raft.Leader {
+		metrics.SetGauge([]string{"autopilot", "failure_tolerance"}, float32(state.FailureTolerance))
+		if state.Healthy {
```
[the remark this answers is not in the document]
vishalnayak: Added a health metric per node.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,556 @@
...
+}
+
+// AutopilotServer represents the health information of individual server node
+// retrieved from autopilot.
+type AutopilotServer struct {
+	ID          string            `json:"id"`
+	Name        string            `json:"name"`
+	Address     string            `json:"address"`
+	NodeStatus  string            `json:"node_status"`
+	LastContact *ReadableDuration `json:"last_contact"`
```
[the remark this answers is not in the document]
vishalnayak: This wasn't the issue. Fixed the audit log problem. Using a `*AutopilotServer` instead of `AutopilotServer` in the Servers map under AutopilotState resolved the problem.

command/operator_raft_autopilot_set_config.go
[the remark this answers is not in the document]
ncabatoff: I don't think there is a way currently.  Consul has their own flags package, e.g.:

```
// BoolValue provides a flag value that's aware if it has been set.
type BoolValue struct {
	v *bool
}
```

Maybe this is something we can tackle during feature freeze?

helper/testhelpers/teststorage/teststorage.go
```diff
@@ -199,6 +206,18 @@ func RaftBackendSetup(conf *vault.CoreConfig, opts *vault.TestClusterOptions) {
 	}
 }
 
+func RaftBackendWithAutopilotSetup(conf *vault.CoreConfig, opts *vault.TestClusterOptions) {
+	conf.DisablePerformanceStandby = true
```
ncabatoff, line 210: I think it's time to stop setting DisablePerformanceStandby=true by default.  I've removed that line in my PR and the tests are still passing on ent.

vishalnayak: 👍 done

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,583 @@
...
+		metrics.SetGauge([]string{"autopilot", "failure_tolerance"}, float32(state.FailureTolerance))
+		if state.Healthy {
+			metrics.SetGauge([]string{"autopilot", "healthy"}, 1)
+		} else {
+			metrics.SetGauge([]string{"autopilot", "healthy"}, 0)
+		}
+
+		for id, state := range state.Servers {
+			if state.Health.Healthy {
+				metrics.SetGauge([]string{"autopilot", string(id), "healthy"}, 1)
```
ncabatoff, line 240: Would be more standard to make the node id be a label instead of part of the metric name.  (and would permit doing rollup queries with more systems)

vault/logical_system_raft.go
```diff
@@ -335,6 +382,121 @@ func (b *SystemBackend) handleStorageRaftSnapshotRead() framework.OperationFunc
...
+			entry, err := logical.StorageEntryJSON(raftAutopilotConfigurationStoragePath, configClone)
+			if err != nil {
+				return nil, err
+			}
+			if err := b.Core.barrier.Put(ctx, entry); err != nil {
+				return nil, err
+			}
+		}
+
+		raftStorage.SetAutopilotConfig(configClone)
```
ncabatoff, line 494: Worth logging the new config (since we log it when autopilot starts)?

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,583 @@
...
+func (b *RaftBackend) SetAutopilotConfig(config *AutopilotConfig) {
+	b.l.Lock()
+	b.autopilotConfig = config
+	b.l.Unlock()
+}
+
+// AutopilotConfig returns the autopilot configuration in the backend.
+func (b *RaftBackend) AutopilotConfig() *AutopilotConfig {
+	b.l.RLock()
+	conf := b.autopilotConfig
```
ncabatoff, line 359: Maybe return a copy instead, so that callers don't inadvertently modify ours?

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,583 @@
...
+	// Start with a default config
+	b.autopilotConfig = b.defaultAutopilotConfig()
+
+	// Check if the config was present in storage
+	switch opts.AutopilotConfig {
+	case nil:
+		// Autopilot config wasn't found in storage. Check if autopilot settings were part of config file.
+		conf, err := b.autopilotConf()
+		if err != nil {
+			return err
```
ncabatoff, line 567: Maybe if we have problems parsing the autopilot config we should log and fall back to the default config?  It goes against my own instincts but seems more consistent with how Vault behaves generally.

vault/raft.go
```diff
@@ -198,11 +165,47 @@ func (c *Core) startRaftBackend(ctx context.Context) (retErr error) {
...
+	if raftBackend == nil {
+		return nil
+	}
+
+	c.logger.Info("starting raft active node")
+
+	// Read autopilot configuration from storage
+	autopilotConfig, err := c.autopilotConfiguration(ctx)
+	if err != nil {
+		return err
```
ncabatoff, line 178: Might want to log and fall back to default config here instead of failing to come up.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,583 @@
...
+		server := &autopilot.Server{
+			ID:          raft.ServerID(id),
+			Name:        id,
+			RaftVersion: raft.ProtocolVersionMax,
+			Ext:         d.autopilotServerExt(state.NonVoter),
+		}
+
+		switch state.IsDead {
+		case true:
+			d.logger.Info("informing autopilot that the node left", "id", id)
```
ncabatoff, line 316: This is probably too spammy for Info.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,583 @@
...
+}
+
+// KnownServers is called by the autopilot library to know the status of each
+// node in the raft cluster. If the application thinks that certain nodes left,
+// it is here that we let the autopilot library know of the same.
+func (d *Delegate) KnownServers() map[raft.ServerID]*autopilot.Server {
+	d.l.RLock()
+	future := d.raft.GetConfiguration()
+	if err := future.Error(); err != nil {
+		d.logger.Error("failed to get raft configuration when computing known servers", "error", err)
```
ncabatoff, line 285: Might be best to unlock and return nil in this case.  Though looking at callers it seems they too call GetConfiguration and return before calling us on failure, so it's probably moot.

physical/raft/raft_autopilot.go
vishalnayak: Good point. Done.

vault/logical_system_raft.go
vishalnayak: Yep. Updated.

physical/raft/raft_autopilot.go
vishalnayak: Very good point. Thanks for this comment. This can avoid unnecessary pain if we do encounter this case. Fixed.

vault/raft.go
vishalnayak: Done.

physical/raft/raft_autopilot.go
vishalnayak: Done.

physical/raft/raft_autopilot.go
vishalnayak: This happens only when the node is considered dead by Vault. If cleaning dead servers is enabled, autopilot will knock the node out pretty soon (seconds maybe) and this log will not show up after that. If dead server cleanup is not enabled, this can become spammy, but maybe for good? We will need to inform operators that the node is dead, requiring manual operator intervention to invoke a remove-peer.

physical/raft/raft_autopilot.go
vishalnayak: I agree. Done.

command/operator_raft_autopilot_set_config.go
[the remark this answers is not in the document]
vishalnayak: VAULT-1615 is tracking this. I'm interested to tackle this, but not in this PR though. I'll try to get to it before code freeze. Resolving for now.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,590 @@
...
+
+		for id, state := range state.Servers {
+			labels := []metrics.Label{
+				{
+					Name:  "node_id",
+					Value: string(id),
+				},
+			}
+			if state.Health.Healthy {
+				metrics.SetGaugeWithLabels([]string{"autopilot", "healthy"}, 1, labels)
```
ncabatoff, line 246: Sorry to keep coming back to this.  Now there's a problem in that we have autopilot_healthy both with and without labels, which is problematic for Prometheus (maybe others too).  Can we make this one autopilot_healthy_nodes?

physical/raft/raft_autopilot.go
ncabatoff: I believe this will log every 2s by default (autopilot update interval) but we won't do anything about it for a while, since I think the dead server cleanup happens at the reconcile interval (every 10s by default).  Moreover, I believe when autopilot removes the dead server it will log then, making this redundant.

Also consider the case of a user who isn't using autopilot and isn't watching the logs who now risks filling up their log device.  In general I don't like having log messages that can occur at such a high frequency going on indefinitely.  I don't know of any at INFO threshold currently, and I don't think we should start.

physical/raft/raft_autopilot.go
vishalnayak: No problem! Using `autopilot_node_healthy` now.

physical/raft/raft_autopilot.go
vishalnayak: Sounds good. Reduced to `Debug`.

vishalnayak: Let me know if you think `Trace` is the right level.

ncabatoff: So long as it's lower than Info I'm happy.  Thanks!

physical/raft/raft.go
```diff
@@ -127,13 +127,6 @@ type RaftBackend struct {
 	// However, only active node will have a "running" autopilot.
 	autopilot *autopilot.Autopilot
 
-	// disableAutopilot if set will not put autopilot implementation to use. The
-	// fallback will be to interact with the raft instance directly. This is useful
-	// to retain state of operations on existing clusters which do not have
-	// autopilot enabled yet. For new clusters however, the default will be to have
-	// autopilot enabled from get-go.
-	disableAutopilot bool
```
briankassouf: Is there still a way to disable?

vault/raft.go
```diff
@@ -171,6 +171,26 @@ func (c *Core) setupRaftActiveNode(ctx context.Context) error {
...
 	c.logger.Info("starting raft active node")
+
+	// Read autopilot configuration from storage
+	autopilotConfig, err := c.autopilotConfiguration(ctx)
+	if err != nil {
+		return err
+	}
+	// If the config wasn't found in storage, then either the default config, or
+	// the HCL config would have been used to instantiate autopilot. The
+	// resulting configuration should be persisted in storage.
```
briankassouf, line 182: Why do we need to persist the default config? Can we only persist if the config has been changed from default? Doing so would give us the flexibility to update the default config overtime and have it apply to existing clusters

physical/raft/raft.go
vishalnayak: There is now. Using `VAULT_RAFT_AUTOPILOT_DISABLE` env var. It requires a restart to take effect though.

vault/raft.go
vishalnayak: Good idea. Done.

vault/external_tests/raft/raft_test.go
```diff
@@ -32,30 +33,50 @@ import (
...
 	}
+
 	teststorage.RaftBackendSetup(conf, &opts)
-	if !joinNodes {
+	if ropts.DisableFollowerJoins {
 		opts.SetupFunc = nil
 	}
+
+	if !ropts.EnableAutopilot {
+		envVar := "VAULT_RAFT_AUTOPILOT_DISABLE"
```
briankassouf, line 75: Could we use a config value instead of an ENV var for this? This will apply to all tests in the same test binary

command/server/config_test_helpers.go
```diff
@@ -895,3 +895,36 @@ func testLoadConfigFileLeaseMetrics(t *testing.T) {
 		t.Fatal(diff)
 	}
 }
+
+func testConfigRaftAutopilot(t *testing.T) {
```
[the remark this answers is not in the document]
vishalnayak: Did this as part of external test.

vault/external_tests/raft/raft_test.go
vishalnayak: Despite being unset?

The main reason I was hesitant to add a config option is that it might be hard to deprecate it going forward, and we'll end up supporting this forever.

If we really want to go the config route, can we make it an undocumented thing, which we can backoff from (or make public) later?

vishalnayak: The tests seem to be doing the right thing with env var, but may be I am missing your point.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,710 @@
...
+	conf := b.autopilotConfig
+	b.l.RUnlock()
+	return conf.Clone()
+}
+
+func (b *RaftBackend) defaultAutopilotConfig() *AutopilotConfig {
+	return &AutopilotConfig{
+		CleanupDeadServers:          false,
+		LastContactThreshold:        10 * time.Second,
+		LastContactFailureThreshold: 24 * time.Hour,
```
vishalnayak, line 397: We might need consensus on the name and value of this.

command/format.go
```diff
@@ -115,6 +117,102 @@ func (y YamlFormatter) Output(ui cli.Ui, secret *api.Secret, data interface{}) e
...
+
+		for _, output := range outputs {
+			buffer.WriteString(output.value)
+		}
+	}
+
+	return buffer.String()
+}
+
+func (p PrettyFormatter) OutputAutopilotHealth(ui cli.Ui, data interface{}) {
```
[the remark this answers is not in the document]
vishalnayak: State API fetching is abstracted via `client.Sys().RaftAutopilotState()` which is now being tested.

vault/external_tests/raft/raft_test.go
ncabatoff: The concern with the env var is that if you have two tests in the same package and test1 wants to disable autopilot, if the tests run in parallel then test2 might get it disabled too.

You avoid the concern about deprecation by keeping the env var for end users, and adding a config setting like Brian was saying - not something that goes in a config file, something that's only available to tests, e.g. a CoreConfig setting.

helper/testhelpers/teststorage/teststorage.go
```diff
@@ -83,6 +83,14 @@ func MakeRaftBackend(t testing.T, coreIdx int, logger hclog.Logger) *vault.Physi
 	return MakeRaftBackendWithConf(t, coreIdx, logger, nil)
 }
 
+func MakeRaftWithHCLAutopilotBackend(t testing.T, coreIdx int, logger hclog.Logger) *vault.PhysicalBackendBundle {
+	extraConfAutopilotValue := os.Getenv("TEST_AUTOPILOT_HCL_VALUE")
```
ncabatoff, line 87: It feels like this env var could go into a new TestClusterOptions field, say PhysicalFactoryConfig map[string]interface{}.  Then newCore could pass this as a new 4th argument to the PhysicalFactory invocation.

physical/raft/raft_autopilot.go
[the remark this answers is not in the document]
ncabatoff: In hindsight I regret suggesting this, now that we're passing AutopilotConfig to log statements.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,710 @@
...
+			ID:          raft.ServerID(id),
+			Name:        id,
+			RaftVersion: raft.ProtocolVersionMax,
+			Ext:         d.autopilotServerExt(state.NonVoter),
+		}
+
+		switch state.IsDead {
+		case true:
+			d.logger.Debug("informing autopilot that the node left", "id", id)
+			server.NodeStatus = autopilot.NodeLeft
```
ncabatoff, line 334: MarkFollowerAsDead is called by startFollowerHeartbeatTracker when it sees the LastContactFailureThreshold is exceeded.  But doesn't that mean Autopilot will never get to see the threshold exceeded?

I don't think this really breaks anything, except maybe we won't get to see all the server stats in the autopilot state endpoint.  But does it buy us anything over not setting NodeStatus and leaving all the health logic in Autopilot?

ncabatoff: Oh, I misread, you're using LastContactFailureThreshold, not LastContactThreshold.

physical/raft/raft_autopilot.go
ncabatoff: LeftServerLastContactThreshold?

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,710 @@
...
+			}
+			b.l.RUnlock()
+		}
+	}
+}
+
+// StopAutopilot stops a running autopilot instance. This should only be called
+// on the active node.
+func (b *RaftBackend) StopAutopilot() {
+	b.l.RLock()
```
ncabatoff, line 515: s/RLock/Lock/?

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,710 @@
...
+			return
+		case <-ticker.C:
+			b.l.RLock()
+			if b.autopilotConfig.CleanupDeadServers && b.autopilotConfig.LastContactFailureThreshold != 0 {
+				b.followerStates.l.RLock()
+				followers := b.followerStates.followers
+				b.followerStates.l.RUnlock()
+				for id, state := range followers {
+					now := time.Now()
+					threshold := state.LastHeartbeat.Add(b.autopilotConfig.LastContactFailureThreshold)
```
ncabatoff, line 500: We may need extra logic here so that if the cluster was down for a day, we don't automatically mark everyone as failed immediately when we come back up.

vault/external_tests/raft/raft_autopilot_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hashicorp/vault/helper/namespace"
+	"github.com/hashicorp/vault/helper/testhelpers"
+	"github.com/hashicorp/vault/physical/raft"
+	"github.com/hashicorp/vault/vault"
+)
+
+func TestRaft_Autopilot_ServerStabilization(t *testing.T) {
```
ncabatoff, line 19: How about a test that demonstrates unstable servers not becoming voters?

helper/testhelpers/teststorage/teststorage.go
vishalnayak: Good suggestion. Done.

physical/raft/raft_autopilot.go
vishalnayak: Autopilot library doesn't support having a threshold for computing dead servers. It only supports threshold for computing of a node is healthy or not. The way autopilot realizes that a node is dead is by application (in this case) Vault letting it know that the node left. In Consul, Serf informs Consul that the node left. In Vault, we rely on LastContactFailureThreshold to compute "when" to consider a node to be dead.

Vault must set the NodeStatus as NodeLeft. This will turn the node to be unhealthy in autopilot. If dead server cleanup is not enabled, it will continue to show up in the state API response as an unhealthy node. If dead server cleanup is enabled, it will be pruned.

Does this address your question?

ncabatoff: Yeah, I misread LastContactFailureThreshold as LastContactThreshold, that's all.

physical/raft/raft_autopilot.go
vishalnayak: Yep. Fixed.

physical/raft/raft_autopilot.go
vishalnayak: 1) The entire cluster was down for a period.
When the active node comes up, it will reset the follower states. All computations will happen fresh and hence cluster going down shouldn't be an issue.

2) One or more followers could go down and come back up after the threshold.
2.a) Dead server cleanup is disabled.
No-op. No issues.

2.b) Dead server cleanup is enabled.
Cleanup would have already happened and the follower entry would both be already gone from raft config and followerStates in Vault. That node will need to rejoin the cluster. Everything will be fresh. No issues.

Did I miss the use-case you were thinking?

ncabatoff: Nope, I forgot about the `!state.LastHeartbeat.IsZero()`, so my concern wasn't warranted.  Though that does bring up another scenario to consider: suppose the active node crashes, and a new node takes over.  That new active node will have no follower state info initially, so the LastHeartbeat value it has for the former active node will be zero.  If the former active node fails to come back up, we'll never mark it dead will we?

vishalnayak: In that case, the new active node will never have the old active node as the follower. Vault will skip informing about that server to autopilot. Autopilot in this case considers the old active node as a stale server (not a dead server), in which case that node will be unilaterally removed by autopilot from the raft config.

It's good to have test cases for all these. Working on the same. I'm slowly adding tests. I'll get to these eventually.

vishalnayak: Basically, servers known to Vault but not responding can be unhealthy/dead servers. Servers unknown to Vault are stale servers.

ncabatoff: It looks like we populate followerStates in raftTLSRotatePhased based on the servers known to raft, so I'm not sure it's true that the new active node won't have the old active node as the follower.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,710 @@
...
+	for {
+		select {
+		case <-b.followerHeartbeatTrackerStopCh:
+			return
+		case <-ticker.C:
+			b.l.RLock()
+			if b.autopilotConfig.CleanupDeadServers && b.autopilotConfig.LastContactFailureThreshold != 0 {
+				b.followerStates.l.RLock()
+				followers := b.followerStates.followers
+				b.followerStates.l.RUnlock()
```
ncabatoff, line 497: I don't think it's safe to release this lock before iterating over followers - what if someone calls Delete to remove one?

physical/raft/raft_autopilot.go
vishalnayak: I see. I'll look into this case.

physical/raft/raft_autopilot.go
vishalnayak: I like `LeftServerLastContactThreshold` better than `LastContactFailureThreshold`. I'll make this change.

physical/raft/raft_autopilot.go
vishalnayak: Adjusted the locking. Let me know what you think.

ncabatoff: I meant the followerStates lock.

physical/raft/raft_autopilot.go
briankassouf: Is this an existing Consul setting? If it's only used for dead server cleanup should we call it `DeadServerLastContactThreshold`?

vishalnayak: Done.

vishalnayak: Nope. This isn't an existing Consul setting. Consul doesn't need it since Serf does the dead server detection. For Vault we need to have our own way to flag a node as dead, and this is that. I like `DeadServerLastContactThreshold` over `LeftServerLastContactThreshold`. I'll do that.

physical/raft/raft_autopilot.go
vishalnayak: Yep. That is the lock I adjusted. The lock is now released after iterating.

vishalnayak: What I also did was to grab the write lock only if there is a criteria match for marking the node as dead. Coz otherwise, we would hold the write lock far too often.

command/operator_raft_autopilot_set_config.go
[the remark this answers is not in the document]
vishalnayak: Getting https://github.com/hashicorp/vault/pull/10988 in should help in addressing this.

vault/external_tests/raft/raft_test.go
vishalnayak: Got it. Thanks for the suggestion Bk and for the explanation Nick. Done.

physical/raft/raft_autopilot.go
vishalnayak: Done.

physical/raft/raft_autopilot.go
vishalnayak: I came to a thought that setting the LastHeartbeat to time.Now() when a node assumes active duty is enough to not encounter this issue. I've pushed up the diff with it. I'll test it out tomorrow morning.

physical/raft/raft_autopilot.go
ncabatoff:
```
                                b.followerStates.l.RLock()
				followers := b.followerStates.followers
				b.followerStates.l.RUnlock()
                                for id, state := range followers {
```

I worry that `range followers` is unsafe while we don't have `b.followerStates.l` locked, since followers is just a pointer to the same map, and FollowerStates.Delete could delete it from it at any time.

vishalnayak: Right. That's what I thought you suggested and have the lock held for the entire duration of range now. The diff here is pointing to the old code. Can you take a look at the latest changes and let me know if that addresses the concern?

ncabatoff: It's still not showing in github for some reason, but I've pulled the changes locally and see what you mean.  Thanks!

command/operator_raft_autopilot_set_config.go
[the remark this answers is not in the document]
vishalnayak: Addressed in bdb7b86f59586943cfd6e73587d13542794b3729.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,722 @@
...
+					threshold := state.LastHeartbeat.Add(b.autopilotConfig.LeftServerLastContactThreshold)
+					if !state.LastHeartbeat.IsZero() && now.After(threshold) && !state.IsDead {
+						markingCandidatePresent = true
+					}
+				}
+				b.followerStates.l.RUnlock()
+
+				if markingCandidatePresent {
+					// Switch locks and check again
+					b.followerStates.l.Lock()
```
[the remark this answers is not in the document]
vishalnayak: Talked about this offline. This won't hurt, even if it doesn't help a lot. Retaining as-is.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,684 @@
...
+		case <-b.followerHeartbeatTrackerStopCh:
+			return
+		case <-ticker.C:
+			markingCandidatePresent := false
+			b.l.RLock()
+			if b.autopilotConfig.CleanupDeadServers && b.autopilotConfig.DeadServerLastContactThreshold != 0 {
+				b.followerStates.l.RLock()
+				for _, state := range b.followerStates.followers {
+					if state.LastHeartbeat.IsZero() || state.IsDead {
+						continue
```
raskchanky, line 452: I'm not 100% sure of Go's lock semantics, but do we need to `b.l.RUnlock()` and `b.followerStates.l.RUnlock()` here? If we don't unlock those 2 locks first, and we `continue` to the top of the loop, will we recursively acquire both of those read locks again?

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,684 @@
...
+				}
+				b.followerStates.l.RUnlock()
+
+				// If marking a node as dead, switch read with a write lock and
+				// check the conditions again.
+				if markingCandidatePresent {
+					b.followerStates.l.Lock()
+					for id, state := range b.followerStates.followers {
+						if state.LastHeartbeat.IsZero() || state.IsDead {
+							continue
```
raskchanky, line 467: Same comment here as with the above `continue` re: unlocking

vault/raft.go
```diff
@@ -198,11 +169,56 @@ func (c *Core) startRaftBackend(ctx context.Context) (retErr error) {
...
+
+		hclConfig, err := raftBackend.AutopilotHCLConfig()
+		if err != nil {
+			c.logger.Error("failed to load autopilot config supplied via config file; falling back to default config", "error", err)
+		}
+
+		// If the autopilot config wasn't found in storage, and if it was found
+		// in HCL config file, then HCL config would have been put to use.
+		// Persist that to storage.
+		if storageConfig == nil && hclConfig != nil {
```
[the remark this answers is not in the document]
vishalnayak: Not persisting anymore.

physical/raft/raft_autopilot.go
vishalnayak: `continue` is happening _within_ the critical section. So, it should be good, unless I am missing something here.

physical/raft/raft_autopilot.go
vishalnayak: Same response as above :-)

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,684 @@
...
+func (s *FollowerStates) Delete(nodeID string) {
+	s.l.Lock()
+	delete(s.followers, nodeID)
+	s.l.Unlock()
+}
+
+// Get retrieves information about a peer from the follower states.
+func (s *FollowerStates) Get(nodeID string) FollowerState {
+	s.l.RLock()
+	state := s.followers[nodeID]
```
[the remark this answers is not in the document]
vishalnayak: Good catch. 1) This wasn't used anywhere; so I ended up deleting the whole function. 2) If we want Get in future, it will be nice if we work with pointer to FollowerState instead. So updated the map to hold `*FollowerState`.

physical/raft/raft.go
```diff
@@ -1212,6 +1280,22 @@ func (b *RaftBackend) LockWith(key, value string) (physical.Lock, error) {
 	}, nil
 }
 
+// SetNonVoter sets a field in the backend indicating that this node is intended
+// to be a permanent non-voter.
+func (b *RaftBackend) SetNonVoter(nonVoter bool) {
```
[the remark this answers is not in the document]
vishalnayak: Fetching suffrage from configbucket is now being done.

physical/raft/raft_autopilot.go
vishalnayak: I don't think locks are getting recursively acquired.

raskchanky: Yeah this was a bit of a 🤦 for me. I didn't notice that there were nested `for` loops on my first read through of this. You're right!

physical/raft/fsm.go
```diff
@@ -243,6 +251,90 @@ func writeSnapshotMetaToDB(metadata *raft.SnapshotMeta, db *bolt.DB) error {
...
+}
+
+func (f *FSM) persistConfig(config *ConfigurationValue) error {
+	configBytes, err := proto.Marshal(config)
+	if err != nil {
+		return err
+	}
+
+	return f.db.Update(func(tx *bolt.Tx) error {
+		b, err := tx.CreateBucketIfNotExists(configBucketName)
```
ncabatoff, line 330: You should be able to assume the bucket already exists, since openDBFile is called by NewFSM and it already did this call.

physical/raft/fsm.go
```diff
@@ -243,6 +251,90 @@ func writeSnapshotMetaToDB(metadata *raft.SnapshotMeta, db *bolt.DB) error {
...
+		if err != nil {
+			return err
+		}
+		return b.Put(latestConfigKey, configBytes)
```
ncabatoff, line 334: Are you sure you want to be writing to the same key as the existing raft config storage?  I'd assumed the idea was to create an entirely new record in the config bucket, rather than augmenting what was already there.

physical/raft/fsm.go
vishalnayak: Yep. Thanks! done.

physical/raft/fsm.go
vishalnayak: I thought, this was the idea. But now that you ask, it maybe is a good idea to have it separate since we don't have to wait on any other log messages to arrive. I'll think a bit more on this.

helper/testhelpers/teststorage/teststorage.go
```diff
@@ -102,7 +98,10 @@ func MakeRaftBackendWithConf(t testing.T, coreIdx int, logger hclog.Logger, extr
 		"performance_multiplier": "8",
 	}
 	for k, v := range extraConf {
-		conf[k] = v
+		val, ok := v.(string)
```
ncabatoff, line 101: What motivated this change?

physical/raft/fsm.go
```diff
@@ -38,10 +38,11 @@ const (
...
 	// dataBucketName is the value we use for the bucket
-	dataBucketName   = []byte("data")
-	configBucketName = []byte("config")
-	latestIndexKey   = []byte("latest_indexes")
-	latestConfigKey  = []byte("latest_config")
+	dataBucketName     = []byte("data")
+	configBucketName   = []byte("config")
+	latestIndexKey     = []byte("latest_indexes")
+	latestConfigKey    = []byte("latest_config")
+	desiredSuffrageKey = []byte("desired_suffrage")
```
ncabatoff, line 45: I know I suggested using a dedicated key to store desired suffrage, but maybe we should instead have a `local_node_config` key and store a JSON blob?  Gives us a little more flexibility going forward.

ncabatoff: Oh, looks like it's already a PB.  Then all the more reason to use a more general key name.

helper/testhelpers/teststorage/teststorage.go
vishalnayak: Param changing from `map[string]string` into `map[string]interface{}`. This was part of adding `PhysicalFactoryConfig` for backends. We are not using it anymore since we stopped supporting autopilot HCL configs, but kept the changes in as its might turn out to be useful in future.

physical/raft/fsm.go
```diff
@@ -243,6 +252,87 @@ func writeSnapshotMetaToDB(metadata *raft.SnapshotMeta, db *bolt.DB) error {
...
+	}
+
+	return f.persistDesiredSuffrage(&ds)
+}
+
+// witnessSuffrage is called when a node successfully joins the cluster. This
+// intent should land in the stored configuration. If the config isn't available
+// yet, we still go ahead and store the intent in the fsm. During the next
+// update to the configuration, this intent will be persisted.
+func (f *FSM) witnessSuffrage(desiredSuffrage string) error {
```
ncabatoff, line 318: "witness" doesn't evoke the right thing to me for what this does - how about "recordSuffrage"?

physical/raft/fsm.go
vishalnayak: Great suggestion. Renamed the key name to `local_node_config`. The entry is a proto message. It should accommodate future values easily.

physical/raft/fsm.go
vishalnayak: Yep. Sounds good. Done.

physical/raft/fsm.go
vishalnayak: Addressed now.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,721 @@
...
+	b.logger.Info("updated autopilot configuration", "config", b.autopilotConfig)
+	b.l.Unlock()
+}
+
+// AutopilotConfig returns the autopilot configuration in the backend.
+func (b *RaftBackend) AutopilotConfig() *AutopilotConfig {
+	b.l.RLock()
+	conf := b.autopilotConfig
+	b.l.RUnlock()
+	return conf.Clone()
```
ncabatoff, line 419: Is it safe to call Clone without the lock?

vishalnayak: Yep. We were anyways returning the same values. We are now just returning it as a different reference. The only main goal here was that the caller doesn't inadvertently modify the config in the backend.

physical/raft/fsm.go
```diff
@@ -243,6 +252,86 @@ func writeSnapshotMetaToDB(metadata *raft.SnapshotMeta, db *bolt.DB) error {
...
+}
+
+// recordSuffrage is called when a node successfully joins the cluster. This
+// intent should land in the stored configuration. If the config isn't available
+// yet, we still go ahead and store the intent in the fsm. During the next
+// update to the configuration, this intent will be persisted.
+func (f *FSM) recordSuffrage(desiredSuffrage string) error {
+	var lnConfig LocalNodeConfigValue
+	lnConfig.DesiredSuffrage = desiredSuffrage
+	f.desiredSuffrage = desiredSuffrage
```
ncabatoff, line 320: Probably better to try to persist first, that way if an error occurs we haven't updated f.desiredSuffrage.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,721 @@
...
+}
+
+func (b *RaftBackend) defaultAutopilotConfig() *AutopilotConfig {
+	return &AutopilotConfig{
+		CleanupDeadServers:             false,
+		LastContactThreshold:           10 * time.Second,
+		DeadServerLastContactThreshold: 24 * time.Hour,
+		MaxTrailingLogs:                1000,
+		MinQuorum:                      3,
```
briankassouf, line 428: Is it safe to set 3 as the default value? I thought we were unable to know the default here

vault/raft.go
```diff
@@ -160,11 +122,24 @@ func (c *Core) startRaftBackend(ctx context.Context) (retErr error) {
...
+
+	if c.disableAutopilot {
+		raftBackend.DisableAutopilot()
+	}
+
 	if err := raftBackend.SetupCluster(ctx, raft.SetupOpts{
 		TLSKeyring:      raftTLS,
 		ClusterListener: c.getClusterListener(),
 		StartAsLeader:   creating,
+		AutopilotConfig: autopilotConfig,
```
briankassouf, line 142: I think there is some kind of bug here:

Read Config on leader node (default config):
```
% vault2 read sys/storage/raft/autopilot/configuration
Key                                   Value
---                                   -----
cleanup_dead_servers                  false
dead_server_last_contact_threshold    24h0m0s
last_contact_threshold                10s
max_trailing_logs                     1000
min_quorum                            3
server_stabilization_time             10s
```

Update Config:
```
% vault2 write sys/storage/raft/autopilot/configuration cleanup_dead_servers=true min_quorum=3
Success! Data written to: sys/storage/raft/autopilot/configuration
```

Read config again (works fine):
```
% vault2 read sys/storage/raft/autopilot/configuration
Key                                   Value
---                                   -----
cleanup_dead_servers                  true
dead_server_last_contact_threshold    24h0m0s
last_contact_threshold                10s
max_trailing_logs                     1000
min_quorum                            3
server_stabilization_time             10s
```

Stepdown as leader:
```
% vault2 operator step-down
Success! Stepped down: http://127.0.0.2:8200
```

Read config on new leader (reverted back to default config):
```
% vault4 read sys/storage/raft/autopilot/configuration
Key                                   Value
---                                   -----
cleanup_dead_servers                  false
dead_server_last_contact_threshold    24h0m0s
last_contact_threshold                10s
max_trailing_logs                     1000
min_quorum                            3
server_stabilization_time             10s
```

briankassouf: It might be that we need to re-load the config in `setupRaftActiveNode` when the node becomes active?

physical/raft/raft.go
```diff
@@ -780,6 +807,13 @@ func (b *RaftBackend) SetupCluster(ctx context.Context, opts SetupOpts) error {
 
 	b.raft = raftObj
 	b.raftNotifyCh = raftNotifyCh
+	b.setupAutopilot(opts)
```
briankassouf, line 810: Should AutoPilot be setup on follower nodes? `SetupCluster` will be called on every node.

physical/raft/fsm.go
```diff
@@ -243,6 +252,105 @@ func writeSnapshotMetaToDB(metadata *raft.SnapshotMeta, db *bolt.DB) error {
...
+	}
+
+	return f.persistDesiredSuffrage(lnConfig)
+}
+
+// recordSuffrage is called when a node successfully joins the cluster. This
+// intent should land in the stored configuration. If the config isn't available
+// yet, we still go ahead and store the intent in the fsm. During the next
+// update to the configuration, this intent will be persisted.
+func (f *FSM) recordSuffrage(desiredSuffrage string) error {
```
briankassouf, line 336: Do we need to hold the fsm's Lock in this function?

physical/raft/fsm.go
vishalnayak: Yep. Fixed.

physical/raft/fsm.go
```diff
@@ -671,6 +785,24 @@ func (f *FSM) Restore(r io.ReadCloser) error {
 		retErr = multierror.Append(retErr, errwrap.Wrapf("failed to open new bolt file: {{err}}", err))
 	}
 
+	// Handle local node config restore
+	switch lnConfig {
+	case nil:
+		// If it wasn't there. It is possible that the snapshot was taken before
+		// 1.7, where we did not have local node config. Perform the upgrade so we will have one now.
```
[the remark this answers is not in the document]
vishalnayak: You are totally right. Fixed.

physical/raft/raft_autopilot.go
vishalnayak: We had a running thought of defaulting to 3, AFAIR. While having the check that it cannot be less than 3, we could force users to provide a value here. Do we want to do that instead?

Also, kindly give it a once over for all the default values. We have mainly relied on Consul values, but based on RFC feedback of being too aggressive on the defaults, I have pretty much randomly selected the default values.

vault/raft.go
vishalnayak: I had thought of doing that. It must have slipped my mind. Done. Fixed.

physical/raft/raft.go
vishalnayak: Followed guidance from here: https://github.com/hashicorp/raft-autopilot/blob/master/autopilot.go#L105

vishalnayak: Another reason was that I was thinking that when perf standby replications is enabled, follower nodes should respond to read requests. But now I wonder how can the follower nodes (who do not have a running autopilot instance), respond to state and get-config APIs.

physical/raft/fsm.go
vishalnayak: Its not particularly contentious, in the sense that we should only touch the `desiredSuffrage` field once during startup, once when a node joins the cluster, and potentially in future whenever the suffrage of the node change (if and when we support it). Having said that it shouldn't hurt if we grab the lock either. Fixed.

vault/logical_system_raft.go
```diff
@@ -296,8 +353,16 @@ func (b *SystemBackend) handleRaftBootstrapAnswerWrite() framework.OperationFunc
...
+		switch nonVoter {
+		case true:
+			suffrage = "voter"
+		default:
+			suffrage = "non-voter"
+		}
+
 		if b.Core.raftFollowerStates != nil {
-			b.Core.raftFollowerStates.update(serverID, 0)
+			b.Core.raftFollowerStates.Update(serverID, 0, 0, suffrage)
```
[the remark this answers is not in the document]
vishalnayak: Yeah. I should rename the field I suppose.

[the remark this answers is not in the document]
vishalnayak: Done.

physical/raft/raft.go
vishalnayak: Can we force these APIs to be serviced by the leader?

vishalnayak: I was meaning to ask if its okay to add the state and config get calls to `perfStandbyAlwaysForwardPaths`.

vishalnayak: If yes, then probably we don't need to set up autopilot here. We can do it all in `setupRaftActiveNode`. Its just about creating autopilot instance really.

vishalnayak: Oh, all raft operations are already getting forwarded. So, we are good there. I'll move this to `setupRaftActiveNode`.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,718 @@
...
+		return nil
+	}
+
+	servers := future.Configuration().Servers
+	serverIDs := make([]string, 0, len(servers))
+	for _, server := range servers {
+		serverIDs = append(serverIDs, string(server.ID))
+	}
+	followerStates := d.followerStates
+	d.l.RUnlock()
```
ncabatoff, line 348: Is there any reason we can't defer this unlock right after locking?  It doesn't look like there's anything hitting storage or the network later in the func, so it's not like we'd be holding the lock for a long time.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,718 @@
...
+	return ret
+}
+
+// RemoveFailedServer is called by the autopilot library when it desires a node
+// to be removed from the raft configuration. This function removes the node
+// from the raft cluster and stops tracking its information in follower states.
+// This function needs to return quickly. Hence removal is performed in a
+// goroutine.
+func (d *Delegate) RemoveFailedServer(server *autopilot.Server) {
+	go func() {
```
ncabatoff, line 397: The possibility of accumulating goroutines trying to do this when there's a failure makes me a little uneasy.  I note on the Consul side they're just sending a gossip message in the equivalent method.  I wonder if we should do something more, like have a map keyed by nodeid to keep track of ongoing removals?

physical/raft/raft_autopilot.go
ncabatoff: It would be a tiny bit safer if you deferred the unlock instead, that way if someday code starts modifying the map in place, there won't be a race.

physical/raft/raft_autopilot.go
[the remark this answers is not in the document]
ncabatoff: I missed the offline convo, but it seems to me this method would be clearer if you just deleted the preceding loop and the markingCandidatePresent thing.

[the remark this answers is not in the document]
ncabatoff: If you don't like the write lock, now that the map values are pointers, you could just use a read lock and make isDead be an atomic.  Assuming you modify the pointed-to structure in-place.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,718 @@
...
+
+	if b.autopilot == nil {
+		return
+	}
+	b.autopilot.Stop()
+
+	if b.followerHeartbeatTrackerStopCh != nil {
+		close(b.followerHeartbeatTrackerStopCh)
+	}
+	b.followerHeartbeatTrackerStopCh = nil
```
ncabatoff, line 529: Is it wise to set this nil?  For one thing, it seems racy.  For another, reads from nil channels block forever, so it could prevent the startFollowerHeartbeatTracker goroutine from exiting.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,718 @@
...
+
+func (b *RaftBackend) DisableAutopilot() {
+	b.l.Lock()
+	b.disableAutopilot = true
+	b.l.Unlock()
+}
+
+// SetupAutopilot instantiates autopilot and tells if its okay to start it.
+func (b *RaftBackend) SetupAutopilot(ctx context.Context, storageConfig *AutopilotConfig) {
+	// TODO: Adding this lock is creating a dead lock. Figure it out.
```
ncabatoff, line 691: You might want to try using https://github.com/sasha-s/go-deadlock.

vault/external_tests/raft/raft_autopilot_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+		state, err = client.Sys().RaftAutopilotState()
+		require.NoError(t, err)
+		require.Equal(t, false, state.Healthy)
+		require.Len(t, state.Servers, numServers)
+		require.Equal(t, false, state.Servers[nodeID].Healthy)
+		require.Equal(t, "alive", state.Servers[nodeID].NodeStatus)
+		require.Equal(t, "non-voter", state.Servers[nodeID].Status)
+
+		// Wait till the stabilization period is over
```
ncabatoff, line 86: Should we be testing to make sure the server doesn't become a voter before the configured stabilization time?

physical/raft/raft.go
vishalnayak: Done.

physical/raft/fsm.go
```diff
@@ -243,6 +252,112 @@ func writeSnapshotMetaToDB(metadata *raft.SnapshotMeta, db *bolt.DB) error {
...
+		if srv.Id == f.localID {
+			switch srv.Suffrage {
+			case int32(raft.Nonvoter):
+				lnConfig.DesiredSuffrage = "non-voter"
+			default:
+				lnConfig.DesiredSuffrage = "voter"
+			}
+		}
+		// Bring the intent to the fsm instance.
+		f.desiredSuffrage = lnConfig.DesiredSuffrage
```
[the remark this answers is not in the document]
vishalnayak: Done.

physical/raft/raft.go
```diff
@@ -781,6 +805,12 @@ func (b *RaftBackend) SetupCluster(ctx context.Context, opts SetupOpts) error {
 	b.raft = raftObj
 	b.raftNotifyCh = raftNotifyCh
 
+	if err := b.fsm.upgradeLocalNodeConfig(); err != nil {
+		// TODO: Should we swallow the error?
```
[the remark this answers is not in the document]
vishalnayak: Agreed. Thanks for voicing your opinion. Removed the TODO.

vault/raft.go
```diff
@@ -345,10 +338,9 @@ func (c *Core) raftTLSRotatePhased(ctx context.Context, logger hclog.Logger, raf
 	}
 	for _, server := range raftConfig.Servers {
 		if server.NodeID != raftBackend.NodeID() {
-			followerStates.update(server.NodeID, 0)
+			followerStates.Update(server.NodeID, 0, 0, raftBackend.DesiredSuffrage())
```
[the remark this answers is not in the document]
vishalnayak: Right. Fixed.

vault/raft.go
```diff
@@ -1105,17 +1107,36 @@ func (c *Core) joinRaftSendAnswer(ctx context.Context, sealAccess *seal.Access,
...
+	}
+	err = raftBackend.SetupCluster(ctx, opts)
 	if err != nil {
 		return errwrap.Wrapf("failed to setup raft cluster: {{err}}", err)
 	}
 
 	return nil
 }
 
+func (c *Core) autopilotConfiguration(ctx context.Context) (*raft.AutopilotConfig, error) {
```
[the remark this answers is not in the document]
vishalnayak: Done.

vault/logical_system_raft.go
```diff
@@ -145,6 +148,60 @@ func (b *SystemBackend) raftStoragePaths() []*framework.Path {
...
+					Type:        framework.TypeInt,
+					Description: "Amount of entries in the Raft Log that a server can be behind before being considered unhealthy.",
+				},
+				"min_quorum": {
+					Type:        framework.TypeInt,
+					Description: "Minimum number of servers allowed in a cluster before autopilot can prune dead servers. This should at least be 3.",
+				},
+				"server_stabilization_time": {
+					Type:        framework.TypeDurationSecond,
+					Description: "Minimum amount of time a server must be in a stable, healthy state before it can be added to the cluster as a voter.",
```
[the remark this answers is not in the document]
vishalnayak: Done.

physical/raft/raft_autopilot.go
```diff
@@ -0,0 +1,718 @@
...
+// markFollowerAsDead marks the node represented by the nodeID as dead. This
+// implies that Vault will indicate that the node "left" the cluster, the next
+// time autopilot asks for known servers. This function should be called with
+// the FollowerStates lock held.
+func (s *FollowerStates) markFollowerAsDead(nodeID string) {
+	state, ok := s.followers[nodeID]
+	if !ok {
+		return
+	}
+	s.followers[nodeID] = &FollowerState{
```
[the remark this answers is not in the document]
vishalnayak: Done.

physical/raft/raft_autopilot.go
vishalnayak: This should be okay. We take this approach with `raftTLSRotationStopCh` as well. The main reason to do this is that, for reasons unknown to me yet, `StopAutopilot` gets called more than once which leads to a "closed channel getting closed agiain"-ish error. I figured the `raftTLSRotationStopCh` also had the same issue and it was circumvented by this approach. IMO, we shouldn't land in a scenario where we are reading from nil channel.

physical/raft/raft_autopilot.go
vishalnayak: There is no reason why we can't. Fixed.

physical/raft/raft.go
```diff
@@ -905,14 +954,23 @@ func (b *RaftBackend) AddPeer(ctx context.Context, peerID, clusterAddr string) e
 	b.l.RLock()
 	defer b.l.RUnlock()
 
-	if b.raft == nil {
-		return errors.New("raft storage is not initialized")
+	if b.autopilot == nil {
```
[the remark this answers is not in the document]
vishalnayak: This was mainly done to accommodate for tests that use `getRaft` which basically creates an instance of raft backend and sets up a cluster and begins operating on them. But in the current code we have made such that we only instantiate autopilot if we are the active node, which isn't a thing when using `getRaft`. All those tests would fail telling that autopilot wasn't initialized.

This was one way to go about it.

An alternative is to have `SetupAutopilot` config in `raft.SetupOpts` and have SetupCluster do the initialization of autopilot if `SetupAutopilot` is true.

Another alternative is to disable autopilot in those tests explicitly.

vault/raft.go
```diff
@@ -972,6 +964,11 @@ func (c *Core) JoinRaftCluster(ctx context.Context, leaderInfos []*raft.LeaderJo
 						c.logger.Warn("join attempt failed", "error", err)
 					} else {
 						// successfully joined leader
+						if err := raftBackend.SetDesiredSuffrage(nonVoter); err != nil {
```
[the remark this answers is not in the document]
vishalnayak: No reason not to. Done.

physical/raft/raft_autopilot.go
ncabatoff: I believe you're seeing StopAutopilot called multiple times because preSeal gets called when we stop being the leader, as well as when we're about to seal.

I'm not convinced that this is ok: right after we close we set to nil, so the select in startFollowerHeartbeatTracker is unlikely to ever return. 

As an alternative, you could consider replacing the backend field followerHeartbeatTrackerStopCh with the ticker that the goroutine ranges over, and calling Stop on the ticker in StopAutopilot.  I think it's safe to call that Stop twice: https://play.golang.org/p/LWF1C7o-_kP

ncabatoff: Re the racy assignment: https://app.circleci.com/pipelines/github/hashicorp/vault/15693/workflows/88f6b46e-c356-4605-9823-04821b0f60e1/jobs/160346

physical/raft/raft_autopilot.go
```diff
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strconv"
 	"sync"
+	"sync/atomic"
```
ncabatoff, line 12: I suggest using uber's atomic instead, so you don't have to worry about alignment.

physical/raft/raft_autopilot.go
ncabatoff: I was curious so I gave it a look.  If you ever want to do this btw, just change the package of the mutex type from `sync` to `deadlock` (as with Core.stateLock) and run your test with the build tag `deadlock`.  That yields

```
POTENTIAL DEADLOCK: Duplicate locking, saw callers this locks in one goroutine:
current goroutine 282 lock &{{{1 0} 0 0 -1073741824 0}} 
all callers to this lock in the goroutine
2021-02-26T12:26:57.049-0500 [TRACE] core0.activity: scanned existing logs: out=[]
../../../physical/raft/raft_autopilot.go:658 raft.(*RaftBackend).SetupAutopilot { b.l.Lock() } <<<<<
../../raft.go:182 vault.(*Core).setupRaftActiveNode { raftBackend.SetupAutopilot(c.activeContext, autopilotConfig) }
../../core.go:2002 vault.standardUnsealStrategy.unseal { if err := c.setupRaftActiveNode(ctx); err != nil { }
../../core.go:2056 vault.(*Core).postUnseal { if err := unsealer.unseal(ctx, c.logger, c); err != nil { }
../../ha.go:565 vault.(*Core).waitForLeadership { err = c.postUnseal(activeCtx, activeCtxCancel, standardUnsealStrategy{}) }
../../ha.go:387 vault.(*Core).runStandby.func7 { c.waitForLeadership(newLeaderCh, manualStepDownCh, leaderStopCh) }
~/go/pkg/mod/github.com/oklog/run@v1.0.0/group.go:38 run.(*Group).Run.func1 { errors <- a.execute() }

../../../physical/raft/raft_autopilot.go:249 raft.(*Delegate).AutopilotConfig { d.l.RLock() } <<<<<
~/go/pkg/mod/github.com/hashicorp/raft-autopilot@v0.1.2/state.go:61 raft-autopilot.(*Autopilot).gatherNextStateInputs { config := a.delegate.AutopilotConfig() }
~/go/pkg/mod/github.com/hashicorp/raft-autopilot@v0.1.2/state.go:133 raft-autopilot.(*Autopilot).nextState { inputs, err := a.gatherNextStateInputs(ctx) }
~/go/pkg/mod/github.com/hashicorp/raft-autopilot@v0.1.2/state.go:333 raft-autopilot.(*Autopilot).updateState { newState, err := a.nextState(ctx) }
~/go/pkg/mod/github.com/hashicorp/raft-autopilot@v0.1.2/run.go:38 raft-autopilot.(*Autopilot).Start { a.updateState(ctx) }
../../../physical/raft/raft_autopilot.go:681 raft.(*RaftBackend).SetupAutopilot { b.autopilot.Start(ctx) }
../../raft.go:182 vault.(*Core).setupRaftActiveNode { raftBackend.SetupAutopilot(c.activeContext, autopilotConfig) }
../../core.go:2002 vault.standardUnsealStrategy.unseal { if err := c.setupRaftActiveNode(ctx); err != nil { }
../../core.go:2056 vault.(*Core).postUnseal { if err := unsealer.unseal(ctx, c.logger, c); err != nil { }
../../ha.go:565 vault.(*Core).waitForLeadership { err = c.postUnseal(activeCtx, activeCtxCancel, standardUnsealStrategy{}) }
../../ha.go:387 vault.(*Core).runStandby.func7 { c.waitForLeadership(newLeaderCh, manualStepDownCh, leaderStopCh) }
~/go/pkg/mod/github.com/oklog/run@v1.0.0/group.go:38 run.(*Group).Run.func1 { errors <- a.execute() }
```

So I think all you need to do is release the lock before you call b.autopilot.Start().

physical/raft/raft_autopilot.go
vishalnayak: Loved the suggestion. Done.

physical/raft/raft_autopilot.go
vishalnayak: Done.

physical/raft/raft_autopilot.go
vishalnayak: Noted the library suggestion. The deadlock stopped showing up after the refactor and simplification.

physical/raft/raft.go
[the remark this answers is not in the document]
vishalnayak: I actually came to see why having multiple causes to have autopilot disabled can cause reasoning troubles. For tests that operates on raft at the backend level only, having autopilot explicitly disabled seemed like the right way to go. Fixed. Thanks for the suggestion.

physical/raft/raft_autopilot.go
vishalnayak: Something like this? https://github.com/hashicorp/vault/pull/11019
The problem I see there is I am not sure how to clean the entries in the map.

ncabatoff: Yeah, that's the sort of thing I had in mind.  I don't think a sync.Map is adequate to manage concurrency here - I mean in practice it'll probably be, because we don't expect RemoveFailedServer to be invoked twice in very quick succession, but even so I'd rather we not commit inappropriate patterns that others might copy.  Using sync.Map will prevent the race detector from complaining, but suppose that we had two more or less simultaneous calls.  Each would call Load, see that there's no entry for their serverID, and proceed in parallel instead of one returning.  This would be a case where you'd need a mutex to get sensible behaviour.

As to cleaning map entries, I would just remove the map entry whenever the goroutine returns, e.g. in a defer.  Though you'll need to grab the lock for that removal too.

Please note that this is a pretty low-priority concern of mine.  I think it makes sense to address it, but I'd be totally fine with you ignoring it for now in favour of testing or whatever.  I don't expect this'll be an issue except possibly in extreme scenarios.

vishalnayak: Resolved in https://github.com/hashicorp/vault/pull/11019

physical/raft/raft_autopilot.go
vishalnayak: Fixed.

briankassouf (approved):
There is a failing test from your most recent commit, but otherwise looks good!

ncabatoff (approved):

raskchanky (approved):

PR #10857, opened 2021-02-08
Autopilot: Dead server cleanup

vishalnayak:


physical/raft/raft.go
```diff
@@ -484,8 +488,19 @@ func (d *Delegate) KnownServers() map[raft.ServerID]*autopilot.Server {
...
+// This function needs to return quickly. Hence removal is performed in a
+// goroutine.
 func (d *Delegate) RemoveFailedServer(server *autopilot.Server) {
-	// TODO: implement me
+	go func() {
+		d.logger.Info("removing dead server from raft configuration", "id", server.ID)
+		if future := d.raft.RemoveServer(server.ID, 0, 0); future.Error() != nil {
+			d.logger.Error("failed to remove server", "server_id", server.ID, "server_address", server.Address, "server_name", server.Name, "error", future.Error())
+		}
+		d.followerStates.Delete(string(server.ID))
```
ncabatoff, line 502: I wonder if simply deleting from followerStates is worth doing.  A subsequent echo would just add it back, so I kind of feel like we either shouldn't bother (so we don't have the unjustified expectation that the removed server won't be there any more) or we should take additional steps to keep the server from getting back into followerStates.  Either way I think we need something more to avoid the removed server from continuing to show up in e.g. the autopilot health endpoint.

vishalnayak: Autopilot state endpoint will get the state from autopilot lib to return over the API. Library gets to know about the servers using KnownServers call which relies on follower states. KnownServers is now made to return a follower state only if the same follower is in the raft config.

Thanks for this comment!

physical/raft/raft.go
```diff
@@ -1185,13 +1201,33 @@ func (b *RaftBackend) SetupCluster(ctx context.Context, opts SetupOpts) error {
...
 
+func (b *RaftBackend) startFollowerHeartbeatTracker() {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-b.followerHeartbeatTrackerStopCh:
+			return
+		case <-ticker.C:
+			for id, state := range b.followerStates.followers {
```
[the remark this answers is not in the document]
vishalnayak: Yep. Done. Thanks!

vishalnayak:
Merging this into feature branch to make both testing and review easier.

PR #10858, opened 2021-02-08
MongoDB - add username customization

pcman312:
Adds the ability to customize username generation for dynamic users in MongoDB.


Valarissa (approved):
*weeps in vendor diff*

Looks good!

calvn:
Does the dep update have to go on the same PR, or can it be addressed separately?

pcman312:
@calvn I'm using this PR to make sure it works. If it does, I'll split it out into a separate PR.

PR #10860, opened 2021-02-08
Wait for initial retention run to finish before adding segments.

mgritter:
Fixes a unit test race that makes this test flaky.

### DIFF
--- a/vault/activity_log_test.go
+++ b/vault/activity_log_test.go
@@ -2307,17 +2307,26 @@ func TestActivityLog_NextMonthStart(t *testing.T) {
 	}
 }
 
+// The retention worker is called on unseal; wait for it to finish before
+// proceeding with the test.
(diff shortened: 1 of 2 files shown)


vault/activity_log_test.go
```diff
@@ -2313,6 +2313,12 @@ func TestActivityLog_Deletion(t *testing.T) {
 	core, _, _ := TestCoreUnsealed(t)
 	a := core.activityLog
 
+	// Wait for retention worker to finish its first pass.
+	select {
+	case <-a.retentionDone:
```
swayne275, line 2318: do you want to have a timeout in this `select` so that it'll fail gracefully if `retentionDone` isn't closed for some reason?

mgritter: Yeah, that's probably best. I thought about it, figured it was unlikely, but I should have gone with my first reaction.

swayne275 (approved):

PR #10871, opened 2021-02-09
Update kubernetes.mdx

raygj:
Add suggestion to validate K8S cluster address and TCP port using cluster-info command and update sample config (line 77) to prompt for validation. Feedback from prospect was that doc is misleading since our sample uses TCP 8443.


jescalan:
This looks fine to me, but I work on the website and not the content so it's probably best to get a review from someone else!

jasonodonnell (approved):
LGTM

PR #10877, opened 2021-02-09
Correct lock acquisition order in the `pathEntityMergeID` identity to fix deadlock condition

### LINKED ISSUE #10876: Deadlock in Identity Backend's Merge and Create paths
**Describe the bug**

Requests to create an entity (POST `identity/entity`) and to merge 2 entities (POST `identity/entity/merge`) both rely on two Mutex's: `IdentityStore`'s lock and a write lock for the [go-memdb](https://github.com/hashicorp/go-memdb) instance used to store identity information in memory.

The two paths acquire the locks in a different order from each other, making it possible to deadlock all write calls to the vault identity backend.

The below examples all use stack traces and links from Vault v1.6.2:

The entity creation request path locks the IdentityStore lock (1) and then memdb lock (2):
```
((2) THIS FUNCTION REQUIRES MEM DB LOCK) github.com/hashicorp/go-memdb.(*MemDB).Txn(memdb.go:57) 
github.com/hashicorp/vault/vault.(*IdentityStore).upsertEntity(identity_store_util.go:501)
((1) THIS FUNCTION LOCKS IDENTITY LOCK) github.com/hashicorp/vault/vault.(*IdentityStore).handleEntityUpdateCommon.func1(identity_store_entities.go:270)   
github.com/hashicorp/vault/sdk/framework. (*Backend).HandleRequest(backend.go:272)
```
1. https://github.com/hashicorp/vault/blob/c7f674ea72f2fbfdf9ec05c387aa08cb7f9e5f44/vault/identity_store_entities.go#L194
2. https://github.com/hashicorp/vault/blob/72752ca7021e0df6e0df96ca131597502e5e069e/vault/identity_store_util.go#L494

But the entity merge request path locks the memdb lock (1) and then the Identity store lock (2)
```
((2) THIS FUNCTION REQUIRES IDENTITY LOCK) github.com/hashicorp/vault/vault.(*IdentityStore).mergeEntity(identity_store_entities.go:716) 
((1) THIS FUNCTION LOCKS MEM DB LOCK) github.com/hashicorp/vault/vault.(*IdentityStore).pathEntityMergeID.func1(identity_store_entities.go:175) 
github.com/hashicorp/vault/sdk/framework.(*Backend).HandleRequest(backend.go:272)
```
1. https://github.com/hashicorp/vault/blob/c7f674ea72f2fbfdf9ec05c387aa08cb7f9e5f44/vault/identity_store_entities.go#L717
2. https://github.com/hashicorp/vault/blob/c7f674ea72f2fbfdf9ec05c387aa08cb7f9e5f44/vault/identity_store_entities.go#L167

This means that a concurrent pair of one create entity request and one merge request can come in, with the create entity request getting the identity lock and the merge entity request getting the memdb lock, and neither request being able to proceed any further.

At that point, all requests that involve a modification to the identity store will hang and timeout, until the Vault process is restarted or terminated and replaced.

Looking at the rest of the identity code, it appears that the Identity lock is always locked prior to a memdb write transaction being created. Based on that, I believe that the lock ordering in `pathEntityMergeID` is incorrect, and the IdentityStore lock call should be moved to happen prior to the memdb write transaction being created. 

I've created a pull request that I believe resolves this: https://github.com/hashicorp/vault/pull/10877

**To Reproduce**
Requires `vault`, `jq` and something like `uuidgen` for generating ids

1. Run `vault server -dev -log-level=trace -dev-root-token-id=dev-root-token`
2. In one process run a loop of entity creations:
```bash
export VAULT_ADDR=http://127.0.0.1:8200 
export VAULT_TOKEN=dev-root-token 
while true; do
  vault write -format=json identity/entity name=$(uuidgen) |  | jq -r '.data.id'
done;
```
`uuidgen` can be replaced with anything that gives you unique/random strings for names

3. In another process at the same time run a loop that creates 2 entities and then merges them:
```bash
export VAULT_ADDR=http://127.0.0.1:8200 
export VAULT_TOKEN=dev-root-token 
while true; do
  a=$(vault write -format=json identity/entity name=$(uuidgen) |  | jq -r '.data.id')
  b=$(vault write -format=json identity/entity name=$(uuidgen) |  | jq -r '.data.id')
  vault write identity/entity/merge from_entity_ids=$a to_entity_id=$b
done;
```

On my laptop, usually within 300-400 requests, if not sooner, both processes will lock up at the same time, and then begin logging context deadline exceeded exceptions every 60 seconds. Running multiple instances of each loop may accelerate this, but I usually could reproduce it with just one instance of each loop in under a minute.

**Expected behavior**
Create and Merge calls should be able to work in parallel without issue

**Environment:**
* Vault Server Version (retrieve with `vault status`): Reproduced on both v1.5.5 and v.1.6.2
* Vault CLI Version (retrieve with `vault version`): v1.5.5, v1.6.2
* Server Operating System/Architecture: Ubuntu , OS X

We initially observed this on one of our Vault clusters under normal traffic from a variety of clients/services, and I was then able to reproduce on both Vault v1.5.5 and Vault v1.6.2 locally on my laptop

**Additional Information**
I've created a pull request that I believe resolves this: https://github.com/hashicorp/vault/pull/10877

ianferguson:
fixes #10876

This fixes lock acquisition ordering in the  `pathEntityMergeID` function so that it matches other code in the identity backend's locking order by locking the `IdentityStore` lock before creating a memdb write transaction that requires a memdb lock.  This avoids the dead lock scenario encountered in #10876.

In order to confirm this works, I ran the same reproduction harness described in #10876 against a locally built version of Vault with this patch and both loops got through over 20,000 requests without issue -- without the patch I had never seen the loops get past 400-500 requests.

### DIFF
--- a/changelog/10877.txt
+++ b/changelog/10877.txt
@@ -1,1 +1,3 @@
+```release-note:bug
 core/identity: Fix deadlock in entity merge endpoint.
+```
--- a/changelog/10877.txt
+++ b/changelog/10877.txt
@@ -1,0 +1,1 @@
+core/identity: Fix deadlock in entity merge endpoint.
--- a/vault/identity_store_entities.go
+++ b/vault/identity_store_entities.go
@@ -164,6 +164,9 @@ func (i *IdentityStore) pathEntityMergeID() framework.OperationFunc {
 		force := d.Get("force").(bool)
 
 		// Create a MemDB transaction to merge entities
+		i.lock.Lock()
+		defer i.lock.Unlock()
+
 		txn := i.db.Txn(true)
 		defer txn.Abort()
 
@@ -172,7 +175,7 @@ func (i *IdentityStore) pathEntityMergeID() framework.OperationFunc {
 			return nil, err
 		}
 
-		userErr, intErr := i.mergeEntity(ctx, txn, toEntity, fromEntityIDs, force, true, false, true)
+		userErr, intErr := i.mergeEntity(ctx, txn, toEntity, fromEntityIDs, force, false, false, true)
 		if userErr != nil {
 			return logical.ErrorResponse(userErr.Error()), nil
 		}


ncabatoff (approved):

ncabatoff:
Hi @ianferguson,

Thanks for the PR!  This looks good to me, but since I'm not that familiar with the identity subsystem I'm going to ask for another reviewer to check it out.  As a follow-up someone (doesn't need to be you) should remove the grabLock option in mergeEntity now that no one will be using it.  But that can wait.

vishalnayak (approved):

vishalnayak:
Maybe the `grabLock` parameter of `mergeEntity` is useless anymore.

ianferguson:
@vishalnayak @ncabatoff thank you for the fast review! 

I added the changelog in 521d329bf

If you'd like I can open a second PR once this one is in that refactors the `grabLock` parameter out of the `mergeEntity` function -- let me know.

Also, while I've got y'alls eyes: while doing some profiling and performance troubleshooting on Vault adjacent to this issue, I ran into challenges observing the usage of the internal semaphores that gate many of the physical backend calls. I opened this PR to add metrics, and am happy to make any change to it if y'all are open to accepting the additional telemetry it adds: https://github.com/hashicorp/vault/pull/10773

ncabatoff:
Sorry, the changelog isn't quite right, it should include backticks.  Can you look at one of the other files in that directory for an example?  It's hard for me to show what it should look like here because the markup is interpreted by github too.  Vishal pointed out that we should have contributor docs to explain this, which we'll try to take care of soon.

> If you'd like I can open a second PR once this one is in that refactors the grabLock parameter out of the mergeEntity function -- let me know.

Up to you!

Re #10773: I saw it, I want to merge it, but we're trying to complete our 1.7 work before the feature freeze so I can't take the time to review it right now.  I made an exception for this PR because it's a significant bugfix, and I wanted to get it into 1.6.3.

ianferguson:
@ncabatoff sorry about that, fixed the formatting in 9fb3ea41a, hopefully got it right this time!

> Up to you!

I'll probably get it open later today, and y'all can review/include or not at your leisure

>  we're trying to complete our 1.7 work before the feature freeze so I can't take the time to review it right now. I made an exception for this PR because it's a significant bugfix, and I wanted to get it into 1.6.3.

that makes perfect sense to me and sounds great!

fornever847 (approved):

fornever847:
Koo

PR #10883, opened 2021-02-10
Respect VAULT_MAX_RETRIES from environment in DefaultConfig()

andrejvanderzee:
The value api.Config.MaxRetries from environment variable VAULT_MAX_RETRIES is always overwritten by 2. This forces default clients to call config.ReadEnvironment() again, just to set this value correctly.

This PR fixes the above.


lueranewera (approved):

briankassouf (approved):

briankassouf:
Thanks for the PR!

Issue #10388, opened 2020-11-12
SSH Secrets engine: allowed_users templating strict regex


adongy:
**Is your feature request related to a problem? Please describe.**

Hi, reading the documentation on the SSH secrets engine, we can leverage identity templating in the allowed_users part : https://www.vaultproject.io/api-docs/secret/ssh#allowed_users_template

After testing it, it turns out the templating detection uses a pretty strict regexp https://github.com/hashicorp/vault/blob/3abc7ee582ffcc92b0a54c99b6eca2c7f209872c/builtin/logical/ssh/path_sign.go#L224

The surrounding `^$` makes it so that I can't use templating in the middle of the principal.

Concrete examples:

This is valid and will perform as expected:

```
allowed_users = "{{ identity.entity.name }}"
```

This is valid, but will not apply templating (and Vault will expect the exact value to be a valid principal)

```
allowed_users = "username-{{ identity.entity.name }}"
```

**Describe the solution you'd like**

I believe the strict regexp was useful when the `allowed_users_template` did not exist in the initial PR https://github.com/hashicorp/vault/pull/7548 but operators have to explicitly opt in to the templating now.

I would like to relax the regexp so that I can use the feature more flexibly.

This would fit my expectations of how templating works in other parts of Vault, for example paths in policies where the templating can be anywhere https://www.vaultproject.io/docs/concepts/policies#templated-policies

**Describe alternatives you've considered**

Documentation on the values where the templating actually happens could be a compromise if we won't relax the regexp.

**Explain any additional use-cases**

My current system uses more information than the username as a principal to identify connecting users (for example, you could differentiate users from service accounts). In this case, having surrounding info on the principal can be useful. 

**Additional context**

Original PR https://github.com/hashicorp/vault/pull/7548

Thanks,

phihos:
A (rather stupid) workaround if you can choose the username freely:

```
{{ identity.entity.name }}-username-{{ identity.entity.name }}
```

It works because the regex checks for `{{` at the front and `}}` at the back. Having multiple templates sections works just fine.
So this is an indicator, that just removing `^` and `$` should fix this.

PR #10886, opened 2021-02-10
Let allowed_users template mix templated and non-templated parts 

phihos:
As outlined in #10388 the regex checking for templates tags in _allowed_users_ when signing SSH certs could be more liberal. By removing `^` and `$` templates like `username-{{ identity.entity.name }}` are also allowed.


phihos:
I also would like to provide a test, but I could not find out how to create a new entity and acquire a token for it with the `logicaltest.TestCase`. I you could point me in the right direction I will gladly write a test for that.

candlerb:
A quick scan suggested these may be helpful:
* https://github.com/hashicorp/vault/blob/master/vault/token_store_test.go#L2623-L2635
* helper functions `testMakeTokenXXX`

phihos:
@candlerb Thanks for the pointer. I implemented a test now, but I must admit that I am not fully satisfied with the result. `logicaltest.TestCase` would have been perfect because it is structured nicely, but since it does not allow to declare both a `LogicalFactory` and a `CredentialFactory` I was forced to create a Vault instance with `NewTestCluster`. The test turned out way more verbose than I intended. Maybe you or somebody else knows how to improve this without writing a new vault testing framework.

builtin/logical/ssh/backend_test.go
```diff
@@ -319,6 +320,118 @@ func TestBackend_allowed_users(t *testing.T) {
...
+	testEntityID := "c65c8cff-5142-3d41-93d0-9d06286e7203"
+	testSshUsername := "testsshusername"
+	testEntityMetadata := map[string]string{
+		"ssh_username": testSshUsername,
+	}
+	testAllowedUsersTemplate := "ssh-{{ identity.entity.metadata.ssh_username }}"
+	expectedValidPrincipal := "ssh-" + testSshUsername
+
+	// Setup Vault cluster with a preexisting entity.
+	coreConfig := &vault.CoreConfig{
```
pmmukh, line 333: Would it be possible to rewrite the setup code in this function to reuse `getSshCaTestCluster` ?

pmmukh:
Another thing, I think it might be nice to add a test that checks just the currently supported `{{..}}` format. This is really only cause I noticed the prior PR did not have tests in them, so a test similar to what you currently have but for the existing format might be nice, to show there's no regression with this change. But also, I totally understand if that feels a bit out of scope for this PR and am fine skipping this too!

phihos:
Hi @pmmukh. Sorry for the late reply, I was pretty caught up in work lately. I will try to implement your comments today or this weekend should that fail. I am happy to hear about your interest to merge this PR.

phihos:
I added the doc line you suggested. I am not sure if I did in the correct location though. @pmmukh Can you confirm that I did it correctly?

PR #10897, opened 2021-02-10
Diagnose startup coverage

mgritter:
This PR instruments the server command's Run() method to programatically report errors.

It would be somewhat better to refactor some of this code into separate functions.  But, the problem for Diagnose is that we would either have to run them one-by-one in the same order, duplicating what Run() would be doing anyway. I think the observer pattern here is a little closer to what we'd actually like to do in the long run with trace spans.

However, there are downsides as illustrated by the need to pass the config object back to the observer.  If Diagnose called parseConfig() itself then it would already have that object.

Sample outputs from hand testing:
```
mgritter@ubuntu:~/vault$ VAULT_DIAGNOSE=1 bin/vault operator diagnose -config ../vault-enterprise/splunk-testing.hcl
Vault v1.6.0-dev ('b421c3348d2a299563ac4252392b57838e7796f0+CHANGES')
[  ok  ] Parsed configuration
[  ok  ] Started raft storage
[  ok  ] Started listeners
[      ] Shamir unseal requires user input
Vault diagnose could not detect any problems.
```

```
[  ok  ] Parsed configuration
[  ok  ] Started raft storage
[failed] Error initializing listeners
listen tcp 127.0.0.1:8200: bind: address already in use
```

```
[  ok  ] Parsed configuration
[  ok  ] Started file storage
[failed] Error during transit unseal
Put "https://test-vault.example.com:8300/v1/transit/encrypt/autounseal": dial tcp 127.0.0.1:8300: connect: connection refused
```

```
[  ok  ] Parsed configuration
[failed] Storage error initializing raft
Error initializing storage of type raft: failed to create fsm: failed to open bolt file: open /var/doesnotexist/raft/vault.db: no such file or directory
```

```
mgritter@ubuntu:~/vault$ VAULT_DIAGNOSE=1 bin/vault operator diagnose -config foo.hcl
Vault v1.6.0-dev ('b421c3348d2a299563ac4252392b57838e7796f0+CHANGES')
[failed] Configuration error
No configuration files found.
```

### DIFF
--- a/command/server.go
+++ b/command/server.go
@@ -1096,8 +1096,9 @@ func (c *ServerCommand) RunWithObserver(observer DiagnoseObserver) (status int) 
 	if config.ServiceRegistration != nil {
 		sdFactory, ok := c.ServiceRegistrations[config.ServiceRegistration.Type]
 		if !ok {
-			observer.Error("config-service-registration", nil)
-			c.UI.Error(fmt.Sprintf("Unknown service_registration type %s", config.ServiceRegistration.Type))
+			err = fmt.Errorf("Unknown service_registration type %s", config.ServiceRegistration.Type)
+			observer.Error("config-service-registration", err)
+			c.UI.Error(err.Error())
 			return 1
 		}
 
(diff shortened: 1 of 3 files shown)


command/server.go
```diff
@@ -965,8 +979,11 @@ func (c *ServerCommand) Run(args []string) int {
 		return 1
 	}
 
+	observer.ConfigCreated(config)
```
swayne275, line 982: I think `SetConfig()` might be more clear for this method?

command/server.go
```diff
@@ -1025,6 +1043,7 @@ func (c *ServerCommand) Run(args []string) int {
 		var err error
```
swayne275, line 1043: is this declaration necessary? could you not just have `:=` on the next line? (i know this isn't part of your changes in this PR)

command/server.go
```diff
@@ -1259,47 +1291,59 @@ func (c *ServerCommand) Run(args []string) int {
...
+			err = fmt.Errorf("HA storage cannot be declared when Raft is the storage type")
+			observer.Error("config-ha-storage", err)
+			c.UI.Error(err.Error())
 			return 1
 		}
 
 		factory, exists := c.PhysicalBackends[config.HAStorage.Type]
 		if !exists {
-			c.UI.Error(fmt.Sprintf("Unknown HA storage type %s", config.HAStorage.Type))
+			err = fmt.Errorf(fmt.Sprintf("Unknown HA storage type %s", config.HAStorage.Type))
```
swayne275, line 1309: hmm, why an `fmt.Sprintf()` inside of an `Errorf`, maybe i'm missing something?

command/server.go
```diff
@@ -1259,47 +1291,59 @@ func (c *ServerCommand) Run(args []string) int {
...
-
 		}
 
 		namedHALogger := c.logger.Named("ha." + config.HAStorage.Type)
 		c.allLoggers = append(c.allLoggers, namedHALogger)
 		habackend, err := factory(config.HAStorage.Config, namedHALogger)
 		if err != nil {
-			c.UI.Error(fmt.Sprintf(
-				"Error initializing HA storage of type %s: %s", config.HAStorage.Type, err))
+			err = fmt.Errorf(
```
swayne275, line 1319: if you use the `%w` verb for the error in an `fmt.Errorf()` then it can be unwrapped

command/server.go
```diff
@@ -1504,13 +1559,15 @@ CLUSTER_SYNTHESIS_COMPLETE:
...
 				if err != nil {
+					observer.Error("listener-cluster-address", err)
 					c.UI.Error(fmt.Sprintf("Error resolving cluster_address: %s", err))
 					return 1
 				}
 				clusterAddrs = append(clusterAddrs, tcpAddr)
 			} else {
 				tcpAddr, ok := ln.Addr().(*net.TCPAddr)
 				if !ok {
+					observer.Error("listener-cluster-address", err)
```
swayne275, line 1570: where is this `err` supposed to be coming from? i'm only seeing the err from 1544, and that's already handled?

swayne275 (commented):
looks good overall, just a few comments

PR #10899, opened 2021-02-11
fix uninitialized tlsConfig in influxdb plugin

jleinfors:
fix for panic when configuring influxdb with tls but without pem_bundle or pem_json
crash log: https://gist.github.com/jleinfors/425c105f0146ba89ccb934f69ddce5d0

### DIFF
--- a/plugins/database/influxdb/connection_producer.go
+++ b/plugins/database/influxdb/connection_producer.go
@@ -168,7 +168,7 @@ func (i *influxdbConnectionProducer) createClient() (influx.Client, error) {
 	}
 
 	if i.TLS {
-		var tlsConfig *tls.Config
+		tlsConfig := &tls.Config{}
 		if len(i.certificate) > 0 || len(i.issuingCA) > 0 {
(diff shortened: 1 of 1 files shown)


plugins/database/influxdb/connection_producer.go
```diff
@@ -192,6 +192,8 @@ func (i *influxdbConnectionProducer) createClient() (influx.Client, error) {
 			if err != nil || tlsConfig == nil {
 				return nil, errwrap.Wrapf(fmt.Sprintf("failed to get TLS configuration: tlsConfig:%#v err:{{err}}", tlsConfig), err)
 			}
+		} else {
+			tlsConfig = &tls.Config{}
```
kalafut, line 196: Instead of an else block what about changing line 171 to: `tlsConfig := &tls.Config{}` ?

jleinfors: thanks, I have updated the pr

kalafut (approved):
Thanks!

PR #10900, opened 2021-02-11
auth/kubernetes docs: Correct default issuer

tomhjp:
As per https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/207d1b4c1cdc4babbde00aeeb76f0a4a021a9954/path_login.go#L24, the default issuer when none is set is `kubernetes/serviceaccount`.


jasonodonnell (approved):
LGTM

PR #10903, opened 2021-02-11
Vault Dependency Upgrades [VAULT-871]

HridoyRoy:
This set of dependency upgrades fixes the known CVEs in our libraries: https://hashicorp.atlassian.net/browse/VAULT-871 . The only dependency not upgraded is `github.com/opencontainers/runc` which I believe is only used in our docker tests. The reason I couldn't upgrade runc is in a comment below.

This change is large, but should not be too risky, as the major changes are mostly changes to our testing. There seem to be a few rabbitmq changes as well, that should be tested manually, and vault on windows should also be tested manually due to the winio upgrade.


HridoyRoy:
So far this upgrades the following top level dependencies, and all nested dependencies:

`github.com/hashicorp/vault-plugin-auth-cf
github.com/hashicorp/vault-plugin-auth-centrify
github.com/hashicorp/vault-plugin-auth-kerberos
github.com/hashicorp/vault-plugin-database-elasticsearch
github.com/hashicorp/vault-plugin-secrets-alicloud
github.com/hashicorp/vault-plugin-secrets-kv
github.com/hashicorp/vault-plugin-auth-azure
github.com/hashicorp/vault-plugin-auth-alicloud
github.com/hashicorp/vault-plugin-database-mongodbatlas
github.com/hashicorp/vault-plugin-secrets-ad
github.com/hashicorp/vault-plugin-secrets-openldap
github.com/hashicorp/vault-plugin-secrets-gcp
github.com/hashicorp/vault-plugin-auth-jwt
github.com/hashicorp/vault-plugin-secrets-mongodbatlas`

However, this manually fixes the version for:

`github.com/coreos/etcd
go.etcd.io/bbolt
go.etcd.io/etcd
google.golang.org/grpc`

due to: https://github.com/etcd-io/etcd/issues/12569 .

HridoyRoy:
github.com/opencontainers/runc can’t be upgraded right now due to the following error:

`go: found github.com/opencontainers/runc/libcontainer/nsenter in github.com/opencontainers/runc v1.0.0-rc9
package github.com/opencontainers/runc/libcontainer/nsenter: C source files not allowed when not using cgo or SWIG: cloned_binary.c nsexec.c`

go.mod
```diff
@@ -90,76 +130,108 @@ require (
...
-	golang.org/x/tools v0.0.0-20200521155704-91d71f6c2f04
-	google.golang.org/api v0.29.0
-	google.golang.org/grpc v1.29.1
+	golang.org/x/term v0.0.0-20201210144234-2321bbc49cbf // indirect
+	golang.org/x/text v0.3.5 // indirect
+	golang.org/x/time v0.0.0-20201208040808-7e3f01d25324 // indirect
+	golang.org/x/tools v0.1.0
+	google.golang.org/api v0.39.0
+	google.golang.org/genproto v0.0.0-20210211221406-4ccc9a5e4183 // indirect
+	google.golang.org/grpc v1.35.0
```
ncabatoff, line 224: This seems misleading.  We're replacing our existing dependency on 1.29.1 with 1.35.0, but then in the replace directive above we're forcing it to 1.27.1?  Can you see if it still works if you change both here and the replace directive to be 1.29.1?

go.mod
```diff
@@ -90,76 +130,108 @@ require (
...
+	github.com/vishvananda/netlink v1.1.0 // indirect
+	github.com/vishvananda/netns v0.0.0-20210104183010-2eb08e3e575f // indirect
+	github.com/vmware/govmomi v0.18.0 // indirect
 	github.com/xdg/stringprep v1.0.0 // indirect
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 	github.com/yuin/gopher-lua v0.0.0-20200816102855-ee81675732da // indirect
 	go.etcd.io/bbolt v1.3.5
-	go.etcd.io/etcd v0.5.0-alpha.5.0.20200425165423-262c93980547
-	go.mongodb.org/mongo-driver v1.4.2
+	go.etcd.io/etcd v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
```
ncabatoff, line 209: Do we maybe skip etcd for now?  Waiting might allow us to avoid the whole replace nonsense which feels like a bit of a minefield.

go.mod
HridoyRoy: Yeah, I definitely see that. I think setting that to 1.29.1 should work.

go.mod
HridoyRoy: I think the replace directive mitigates https://github.com/etcd-io/etcd/security/advisories/GHSA-chh6-ppwq-jh92 , which is a moderate vuln but is marked as high on the spreadsheet. I'll ask if we can skip this one, but I do agree that it's kind of a minefield.

vault/testing.go
```diff
@@ -390,6 +392,10 @@ func testCoreAddSecretMount(t testing.T, core *Core, token string) {
 
 }
 
+func TestCoreUnsealedBackendBenchmarking(t gotesting.TB, backend physical.Backend) (*Core, [][]byte, string) {
```
ncabatoff, line 395: I'm not seeing why these are needed?

HridoyRoy: Ah, so github.com/mitchellh/go-testing-interface v1.14.1 has updated the T interface to include a Parallel() method. This method does not exist in the TB interface in the native testing library, and so methods that were originally using github.com/mitchellh/go-testing-interface v1.14.1's T interface and being called by functions that needed to use the TB interface because of benchmarking tests no longer worked. 

The solution I thought of was just to wrap the methods that require the github.com/mitchellh/go-testing-interface v1.14.1 T interface with a "Benchmarking" function that takes in a TB interface parameter and casts it to a github.com/mitchellh/go-testing-interface v1.14.1 T interface. In the case where a native Golang testing T struct is used, everything will work fine. In the case where a B struct is used, B will get cast to TB, which can be cast to T because T has all the methods in TB, but not vice versa.

ncabatoff: Ok, thanks for explaining.  I'm not thrilled with this solution, but it'll do and I don't have a better one off the top of my head.  I mean we could PR a TB type into go-testing-interface and use that in all the test helper signatures, but I don't want to block this PR just for that.

go.mod
ncabatoff: I don't see the endorsement (I don't think the person who suggested it is part of the etcd dev team), but either way, my main concern is, might this prevent us from upgrading boltdb or grpc if we need to for a more serious reason?  These are way more essential components of Vault than etcd.

builtin/logical/transit/path_encrypt.go
```diff
@@ -159,7 +159,7 @@ func decodeBatchRequestItems(src interface{}, dst *[]BatchRequestItem) error {
 			} else if casted, ok := v.(string); ok {
 				(*dst)[i].Context = casted
 			} else {
-				errs.Errors = append(errs.Errors, fmt.Sprintf("'[%d].context' expected type 'string', got unconvertible type '%T'", i, item["context"]))
+				errs.Errors = append(errs.Errors, fmt.Sprintf("'[%d].context' expected type 'string', got unconvertible type '%T', value: '%v'", i, item["context"], item["context"]))
```
ncabatoff, line 162: How was this necessary for bumping dependencies?

HridoyRoy: TestTransit_decodeBatchRequestItems has the following check: 

`			expectedErr := mapstructure.Decode(tt.src, &expectedDest)

			gotErr := decodeBatchRequestItems(tt.src, &tt.dest)
			gotDest := tt.dest

			if !reflect.DeepEqual(expectedErr, gotErr) {
				t.Errorf("decodeBatchRequestItems unexpected error value, want: '%v', got: '%v'", expectedErr, gotErr)
			}`

which fails without these changes like so: https://app.circleci.com/pipelines/github/hashicorp/vault/15230/workflows/0b4e1411-9665-4b36-9991-e6cac823a4ac/jobs/149962

Since the comment above decodeBatchRequestItems says: 

`
// It aims to behave as closely possible to the original mapstructure.Decode 
`

I figured I should make the two error messages match instead of changing the test.

ncabatoff: Fair enough.

ncabatoff (approved):
One outstanding comment I don't feel too strongly about.  Probably best to have another approver before merging though.

go.mod
HridoyRoy: I tried to change this to 1.29.1, but doing a go mod vendor after changing this value changes it back to 1.35.0.

ncabatoff: Ok, don't worry about it then.

PR #10906, opened 2021-02-11
Cassandra - Add username customization

pcman312:
Adds the ability to customize username generation for dynamic users in Cassandra.

### DIFF
--- a/plugins/database/cassandra/connection_producer.go
+++ b/plugins/database/cassandra/connection_producer.go
@@ -8,14 +8,14 @@ import (
 	"sync"
 	"time"
 
-	"github.com/gocql/gocql"
-	"github.com/hashicorp/errwrap"
-	dbplugin "github.com/hashicorp/vault/sdk/database/dbplugin/v5"
 	"github.com/hashicorp/vault/sdk/database/helper/connutil"
 	"github.com/hashicorp/vault/sdk/database/helper/dbutil"
 	"github.com/hashicorp/vault/sdk/helper/certutil"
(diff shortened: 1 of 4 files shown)


plugins/database/cassandra/connection_producer.go
```diff
@@ -125,15 +125,11 @@ func (c *cassandraConnectionProducer) Initialize(ctx context.Context, req dbplug
 
 	if req.VerifyConnection {
 		if _, err := c.Connection(ctx); err != nil {
-			return dbplugin.InitializeResponse{}, errwrap.Wrapf("error verifying connection: {{err}}", err)
+			return errwrap.Wrapf("error verifying connection: {{err}}", err)
```
catsby, line 128: nit: could use this opportunity to change these to `fmt.Errorf("blah %w", err)` while we're here

catsby (approved):
LGTM 👍 minor nit/suggestion about changing from `errwrap` -> `fmt.Errorf` while we're here, but don't feel blocked by it

plugins/database/cassandra/connection_producer.go
pcman312: I had held off on this because we hadn't updated `errwrap` to 1.1 in a previous PR, but it looks like it was updated shortly after that PR was merged. I'll update accordingly and as I see them.

arengoh (approved):

arengoh (approved):

arengoh (approved):

PR #10908, opened 2021-02-11
Added a table showing the entity alias name used by each auth method.

mgritter:
Corrected the text about token auth method.


briankassouf (approved):
👍

vishalnayak (approved):

PR #10911, opened 2021-02-12
Add docs for Agent's template_retry option added in #10644, based on …

ncabatoff:
…those from consul-template configuration.  Also fix some existing config docs that weren't adhering to our conventions.

### DIFF
--- a/website/content/docs/agent/template.mdx
+++ b/website/content/docs/agent/template.mdx
@@ -4,7 +4,7 @@ page_title: Vault Agent Template
 sidebar_title: Templates
 description: >-
   Vault Agent's Template functionality allows Vault secrets to be rendered to
-  files 
+  files
 
   using Consul Template markup.
 ---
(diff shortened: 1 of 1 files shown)


website/content/docs/agent/template.mdx
```diff
@@ -120,12 +119,39 @@ template {
...
+}
 ```
 
 If you only want to use the Vault agent to render one or more templates and do
 not need to sink the acquired credentials, you can omit the `sink` stanza from
 the `auto_auth` stanza in the agent configuration.
 
+The `template_retry` stanza controls the retry behavior when an error is returned
+from Vault. Vault Agent is highly fault tolerant, meaning it does not exit in the
+face of failure. Instead, it uses exponential back-off and retry functions
```
HridoyRoy, line 137: Can/should we mention here that this stanza is specifically for the template server?

HridoyRoy (approved):

Issue #10842, opened 2021-02-05
Transit API: Wrong encryption context results in 500 rather than 400


washtubs:
**Describe the bug**

While decrypting a data key, if the client presents an encryption context which differs from the encryption context supplied on data key creation, the server should respond with a 400. At some point this changed to a 500 with a more obscure error: "cipher: message authentication failed". Our code retries on 500 which is inline with the guidance here: https://www.vaultproject.io/api-docs. Obviously retrying is not great in this case.

This appears to have been introduced somewhere between 1.4.0 and 1.6.0. 

**To Reproduce**
Start a dev server...

    vault secrets enable transit
    vault write -f transit/keys/my-key convergent_encryption=true derived=true
    vault write -f transit/datakey/plaintext/my-key context=$(echo 100 | base64)
    # obtain the ciphertext, using it unmodified while modifying the context
    vault write -f transit/decrypt/my-key ciphertext=<ciphertext> context=$(echo 101 | base64)
     
    URL: PUT http://127.0.0.1:8200/v1/transit/decrypt/my-key
    Code: 500. Errors:

    * 1 error occurred:
            * cipher: message authentication failed

**Expected behavior**

It should respond with a 400.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.6.0
* Vault CLI Version (retrieve with `vault version`): 1.6.0
* Server Operating System/Architecture: Darwin Kernel Version 19.6.0 (also tested linux)

Vault server configuration file(s):

None needed, just use the dev server

Freyert:
I believe the error you were seeing before was:

```
Error writing data to transit/decrypt/my-key: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/transit/decrypt/my-key
Code: 400. Errors:

* invalid ciphertext: no prefix
```

washtubs:
It's a 400 so functionally the same but the error I was seeing was:

```
Error writing data to transit/decrypt/my-key: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/transit/decrypt/my-key
Code: 400. Errors:

* invalid ciphertext: unable to decrypt
```

NeilMadden:
I'm also running into this with Vault 1.6.2 - supplying an invalid GCM authentication tag results in a 500 error and we have to scan through the error messages looking for `message authentication failed` to detect this case.

victorr:
I have verified that PR#13111 fixes this issue.

[...]

PR #10914, opened 2021-02-12
Return a UserError if aead.Open() fails to align with documentation

crozzy:
As mentioned in this issue: https://github.com/hashicorp/vault/issues/10842 the API documentation states that:
```
500 - Internal server error. An internal error has occurred, try again later. If the error persists, report a bug.
400 - Invalid request, missing or invalid data.
```

This change means that failures in key decryption at the `aead.Open()` step will result in a 400 status code being returned as opposed to a 500 as the problem probably originates from incorrect user data entry (there are other possibilites I imagine).

### DIFF
--- a/sdk/helper/keysutil/policy.go
+++ b/sdk/helper/keysutil/policy.go
@@ -1697,7 +1697,7 @@ func (p *Policy) SymmetricDecryptRaw(encKey, ciphertext []byte, opts SymmetricOp
 	// Verify and Decrypt
 	plain, err := aead.Open(nil, nonce, trueCT, opts.AdditionalData)
 	if err != nil {
-		return nil, err
+		return nil, errutil.UserError{Err: err.Error()}
 	}
 	return plain, nil
 }


washtubs:
Note the 400 message before was pretty generic : `invalid ciphertext: unable to decrypt`

Currently the 500 message also tells you what algorithm is used which seems like revealing unnecessary (albeit benign) details. Like if it's a poly1305-chacha20 keyring it will say that in the error message.

crozzy:
It's an interesting one, traditionally you'd want neither the 500 nor the 400 revealing anything internal but it's also possible to go too far (IMO) at the expense of user experience/debugging. It's also 100% possible that someone changes some code causing `aead.Open` to always error and instead of 500ing it's masked by a 400.

Probably worth someone working on Vault full-time reviewing it who is aware of those internal decisions.

victorr:
HI @crozzy, even though issue #10842 was closed, your fix is still relevant. If you want to rebase your PR I'll be happy to review it.

crozzy:
> HI @crozzy, even though issue #10842 was closed, your fix is still relevant. If you want to rebase your PR I'll be happy to review it.

Looks like you changed the go.mod set up and the name of `master/main` since this PR, was pretty confused there for a second :laughing:

victorr (approved):
👍

victorr:
Thank you for the fix @crozzy, it is very much appreciated.

PR #10926, opened 2021-02-16
Add `nonce` configuration parameter to agent AWS auto-auth documentation 

djcrabhat:
Was looking how to give the vault agent with AWS auth-auth the same nonce, but saw it wasn't documented.  Dove through the code, found https://github.com/hashicorp/vault/blob/master/command/agent/auth/aws/aws.go#L139 and https://github.com/hashicorp/vault/blob/master/command/agent/auth/aws/aws.go#L215 

(tried to call out the importance and point to docs, know setting `nonce` poorly could be very bad!)


djcrabhat:
Thank for taking a look, @hghaf099 .  Freshened up my branch with latest main.

PR #10927, opened 2021-02-16
agent: route templating server through cache

jasonodonnell:
This PR enhances the Vault Agent Templating server to support using the local proxy cache. The templating server will automatically configure itself to use the cache if caching is enabled and a listener is present. If either of these requirements aren't satisfied, the templating server will use the `vault` configuration as it does today.

The only caveat to this change is we ignore TLS verification if the listener has TLS enabled. We do this because a CA would be required to verify the certificate but the proxy and templating server are spawned by the same Vault Agent process.

### DIFF
--- a/command/agent/template/template.go
+++ b/command/agent/template/template.go
@@ -27,7 +27,8 @@ import (
 type ServerConfig struct {
 	Logger hclog.Logger
 	// Client        *api.Client
-	VaultConf     *config.Vault
+	AgentConfig *config.Config
+
 	ExitAfterAuth bool
 	TemplateRetry *config.TemplateRetry
 	Namespace     string
@@ -239,7 +240,19 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
 	// Always set these to ensure nothing is picked up from the environment
 	conf.Vault.RenewToken = pointerutil.BoolPtr(false)
 	conf.Vault.Token = pointerutil.StringPtr("")
-	conf.Vault.Address = &sc.VaultConf.Address
+	conf.Vault.Address = &sc.AgentConfig.Vault.Address
+
+	if sc.AgentConfig.Cache != nil && len(sc.AgentConfig.Listeners) != 0 {
+		scheme := "unix:/"
+		if sc.AgentConfig.Listeners[0].Type == "tcp" {
+			scheme = "https://"
+			if sc.AgentConfig.Listeners[0].TLSDisable {
+				scheme = "http://"
+			}
+		}
+		address := fmt.Sprintf("%s%s", scheme, sc.AgentConfig.Listeners[0].Address)
+		conf.Vault.Address = &address
+	}
 
 	if sc.Namespace != "" {
 		conf.Vault.Namespace = &sc.Namespace
@@ -255,16 +268,26 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
 		ServerName: pointerutil.StringPtr(""),
 	}
 
-	if strings.HasPrefix(sc.VaultConf.Address, "https") || sc.VaultConf.CACert != "" {
-		skipVerify := sc.VaultConf.TLSSkipVerify
+	if strings.HasPrefix(*conf.Vault.Address, "https") || sc.AgentConfig.Vault.CACert != "" {
+		skipVerify := sc.AgentConfig.Vault.TLSSkipVerify
 		verify := !skipVerify
 		conf.Vault.SSL = &ctconfig.SSLConfig{
 			Enabled: pointerutil.BoolPtr(true),
 			Verify:  &verify,
-			Cert:    &sc.VaultConf.ClientCert,
-			Key:     &sc.VaultConf.ClientKey,
-			CaCert:  &sc.VaultConf.CACert,
-			CaPath:  &sc.VaultConf.CAPath,
+			Cert:    &sc.AgentConfig.Vault.ClientCert,
+			Key:     &sc.AgentConfig.Vault.ClientKey,
+			CaCert:  &sc.AgentConfig.Vault.CACert,
+			CaPath:  &sc.AgentConfig.Vault.CAPath,
+		}
+
+		// Only configure TLS Skip Verify if CT is not going through the cache. We can
+		// skip verification if its using the cache because they're part of the same agent.
+		// Agent listener doesn't support mTLS listeners.
+		if sc.AgentConfig.Cache != nil {
+			conf.Vault.SSL.Enabled = pointerutil.BoolPtr(true)
+			conf.Vault.SSL.Verify = pointerutil.BoolPtr(false)
+			conf.Vault.SSL.Cert = pointerutil.StringPtr("")
+			conf.Vault.SSL.Key = pointerutil.StringPtr("")
 		}
 	}
 
--- a/command/agent/template/template.go
+++ b/command/agent/template/template.go
@@ -30,7 +30,6 @@ type ServerConfig struct {
 	AgentConfig *config.Config
 
 	ExitAfterAuth bool
-	TemplateRetry *config.TemplateRetry
 	Namespace     string
 
 	// LogLevel is needed to set the internal Consul Template Runner's log level
@@ -165,12 +164,12 @@ func (ts *Server) Run(ctx context.Context, incoming chan string, templates []*ct
 					},
 				}
 
-				if ts.config.TemplateRetry != nil && ts.config.TemplateRetry.Enabled {
+				if ts.config.AgentConfig.TemplateRetry != nil && ts.config.AgentConfig.TemplateRetry.Enabled {
 					ctv.Vault.Retry = &ctconfig.RetryConfig{
-						Attempts:   &ts.config.TemplateRetry.Attempts,
-						Backoff:    &ts.config.TemplateRetry.Backoff,
-						MaxBackoff: &ts.config.TemplateRetry.MaxBackoff,
-						Enabled:    &ts.config.TemplateRetry.Enabled,
+						Attempts:   &ts.config.AgentConfig.TemplateRetry.Attempts,
+						Backoff:    &ts.config.AgentConfig.TemplateRetry.Backoff,
+						MaxBackoff: &ts.config.AgentConfig.TemplateRetry.MaxBackoff,
+						Enabled:    &ts.config.AgentConfig.TemplateRetry.Enabled,
 					}
 				} else if ts.testingLimitRetry != 0 {
 					// If we're testing, limit retries to 3 attempts to avoid
@@ -243,7 +242,7 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
 	conf.Vault.Address = &sc.AgentConfig.Vault.Address
 
 	if sc.AgentConfig.Cache != nil && len(sc.AgentConfig.Listeners) != 0 {
-		scheme := "unix:/"
+		scheme := "unix://"
 		if sc.AgentConfig.Listeners[0].Type == "tcp" {
 			scheme = "https://"
 			if sc.AgentConfig.Listeners[0].TLSDisable {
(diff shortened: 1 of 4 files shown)


command/agent/template/template.go
```diff
@@ -27,7 +27,8 @@ import (
 type ServerConfig struct {
 	Logger hclog.Logger
 	// Client        *api.Client
-	VaultConf     *config.Vault
+	AgentConfig *config.Config
+
 	ExitAfterAuth bool
 	TemplateRetry *config.TemplateRetry
```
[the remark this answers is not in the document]
jasonodonnell: Good call, done!

command/agent/template/template.go
```diff
@@ -239,7 +240,19 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
 	// Always set these to ensure nothing is picked up from the environment
 	conf.Vault.RenewToken = pointerutil.BoolPtr(false)
 	conf.Vault.Token = pointerutil.StringPtr("")
-	conf.Vault.Address = &sc.VaultConf.Address
+	conf.Vault.Address = &sc.AgentConfig.Vault.Address
+
+	if sc.AgentConfig.Cache != nil && len(sc.AgentConfig.Listeners) != 0 {
+		scheme := "unix:/"
```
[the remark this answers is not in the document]
jasonodonnell: Fixed in https://github.com/hashicorp/vault/pull/10927/commits/f1eb3dc625db9dcf74ed4bb1eee8464a2450eaf7

command/agent/template/template.go
```diff
@@ -255,16 +267,26 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
...
+			Key:     &sc.AgentConfig.Vault.ClientKey,
+			CaCert:  &sc.AgentConfig.Vault.CACert,
+			CaPath:  &sc.AgentConfig.Vault.CAPath,
+		}
+
+		// Only configure TLS Skip Verify if CT is not going through the cache. We can
+		// skip verification if its using the cache because they're part of the same agent.
+		// Agent listener doesn't support mTLS listeners.
+		if sc.AgentConfig.Cache != nil {
+			conf.Vault.SSL.Enabled = pointerutil.BoolPtr(true)
```
calvn, line 286: Do we need this to be set again? It's already set on L274.

command/agent/template/template.go
```diff
@@ -255,16 +267,26 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
...
-			CaPath:  &sc.VaultConf.CAPath,
+			Cert:    &sc.AgentConfig.Vault.ClientCert,
+			Key:     &sc.AgentConfig.Vault.ClientKey,
+			CaCert:  &sc.AgentConfig.Vault.CACert,
+			CaPath:  &sc.AgentConfig.Vault.CAPath,
+		}
+
+		// Only configure TLS Skip Verify if CT is not going through the cache. We can
+		// skip verification if its using the cache because they're part of the same agent.
+		// Agent listener doesn't support mTLS listeners.
```
calvn, line 284: The listener stanza supports [all standard listener params](https://www.vaultproject.io/docs/configuration/listener/tcp#tls_client_ca_file), including mTLS params.

jasonodonnell: Interesting because the cache documentation doesn't have them listed: https://www.vaultproject.io/docs/agent/caching#configuration-listener.

command/agent/template/template.go
jasonodonnell: Good catch, I removed it.

command/agent/template/template.go
calvn: The docs on that page might be incomplete. `StartNewListener` [uses all of the TLS params](https://github.com/hashicorp/vault/blob/f5be0716dbe41ea5b909d3c8a81ebac37fd124e0/command/agent/cache/listener.go#L60-L66) that it accepts and doesn't seem to do any special handling on the agent side.

The listener stanza [section](https://www.vaultproject.io/docs/agent#listener-stanza) on Agent also mentions that it accepts all of the standard listener configuration.

command/agent/template/template.go
```diff
@@ -255,16 +267,24 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
 		ServerName: pointerutil.StringPtr(""),
 	}
 
-	if strings.HasPrefix(sc.VaultConf.Address, "https") || sc.VaultConf.CACert != "" {
-		skipVerify := sc.VaultConf.TLSSkipVerify
+	if strings.HasPrefix(*conf.Vault.Address, "https") || sc.AgentConfig.Vault.CACert != "" {
```
calvn, line 270: Since we're now pointing the CT's vault config to the agent's listener, shouldn't we be checking and using values from `sc.AgentConfig.Listerners[0]` (e.g. `sc.AgentConfig.Listerners[0].TLSCertFile` instead of `c.AgentConfig.Vault.CACert`) and fall back to using values from `sc.AgentConfig.Vault` if the cache stanza is not present?

command/agent/template/template.go
jasonodonnell: Removed the comment, thanks for clarifying.

command/agent/template/template.go
jasonodonnell: We actually override the address at 253 so this should be correct, however, this logic should be reworked a bit.

jasonodonnell: Reworked, thanks for pointing this out!

command/agent/template/template.go
```diff
@@ -255,16 +255,35 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
...
+		if sc.AgentConfig.Listeners[0].Type == "tcp" {
+			scheme = "https://"
+			if sc.AgentConfig.Listeners[0].TLSDisable {
+				scheme = "http://"
+			}
+		}
+		address := fmt.Sprintf("%s%s", scheme, sc.AgentConfig.Listeners[0].Address)
+		conf.Vault.Address = &address
+
+		// Only configure TLS Skip Verify if CT is not going through the cache. We can
```
calvn, line 270: This comment sounds a bit confusing since the check above is `sc.AgentConfig.Cache != nil`, which make it seem like things are set this way if caching _is_ enabled. May be fine to remove the first sentence.

command/agent/template/template.go
```diff
@@ -255,16 +255,35 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
...
+		// skip verification if its using the cache because they're part of the same agent.
+		if scheme == "https" {
+			conf.Vault.SSL.Verify = pointerutil.BoolPtr(false)
+			conf.Vault.SSL.Cert = pointerutil.StringPtr("")
+			conf.Vault.SSL.Key = pointerutil.StringPtr("")
```
calvn, line 275: Don't we have to provide `conf.Vault.SSL.CaCert` if `sc.AgentConfig.Vault.Listeners[0].TLSCertFile` got provided? 

I haven't tested this manually, but it looks like mTLS [can be enabled/enforced on the listener](https://www.vaultproject.io/docs/configuration/listener/tcp#tls_client_ca_file) if they provide `tls_client_ca_file` and they don't toggle on `tls_require_and_verify_client_cert` so I'm not sure hardcoding `Cert` and `Key` to empty string is proper in those cases.

jasonodonnell: @calvn `conf.Vault.SSL.Verify = pointerutil.BoolPtr(false)` disables verification, which is what the CA is used for.

Updated cert/key.

command/agent/template/template.go
jasonodonnell: Oops, fixed.

command/agent/template/template_test.go
```diff
@@ -27,6 +29,233 @@ func TestNewServer(t *testing.T) {
...
+	if !strings.HasPrefix(*ctConfig.Vault.Address, "unix") {
+		t.Fatalf("expected unix address, got %s", *ctConfig.Vault.Address)
+	}
+
+	expected := "unix://foobar"
+	if *ctConfig.Vault.Address != expected {
+		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
+	}
+}
+func TestCacheConfigHTTP(t *testing.T) {
```
tvoran, line 115: Missing a new line here:

```suggestion

func TestCacheConfigHTTP(t *testing.T) {
```

command/agent/template/template_test.go
```diff
@@ -27,6 +29,233 @@ func TestNewServer(t *testing.T) {
...
+	}
+
+	if !strings.HasPrefix(*ctConfig.Vault.Address, "unix") {
+		t.Fatalf("expected unix address, got %s", *ctConfig.Vault.Address)
+	}
+
+	expected := "unix://foobar"
+	if *ctConfig.Vault.Address != expected {
+		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
+	}
```
tvoran, line 113: Is the first check for the address prefix necessary, since the second compares the entire address?

```suggestion
	expected := "unix://foobar"
	if *ctConfig.Vault.Address != expected {
		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
	}
```

command/agent/template/template_test.go
```diff
@@ -27,6 +29,233 @@ func TestNewServer(t *testing.T) {
...
+
+	if !strings.HasPrefix(*ctConfig.Vault.Address, "http") {
+		t.Fatalf("expected http address, got %s", *ctConfig.Vault.Address)
+	}
+
+	expected := "http://127.0.0.1:8300"
+	if *ctConfig.Vault.Address != expected {
+		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
+	}
```
tvoran, line 153: Same question here.

```suggestion
	expected := "http://127.0.0.1:8300"
	if *ctConfig.Vault.Address != expected {
		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
	}
```

command/agent/template/template_test.go
```diff
@@ -27,6 +29,233 @@ func TestNewServer(t *testing.T) {
...
+
+	if !strings.HasPrefix(*ctConfig.Vault.Address, "https") {
+		t.Fatalf("expected https address, got %s", *ctConfig.Vault.Address)
+	}
+
+	expected := "https://127.0.0.1:8300"
+	if *ctConfig.Vault.Address != expected {
+		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
+	}
```
tvoran, line 194:
```suggestion
	expected := "https://127.0.0.1:8300"
	if *ctConfig.Vault.Address != expected {
		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
	}
```

command/agent/template/template_test.go
```diff
@@ -27,6 +29,233 @@ func TestNewServer(t *testing.T) {
...
+
+	if !strings.HasPrefix(*ctConfig.Vault.Address, "http") {
+		t.Fatalf("expected http address, got %s", *ctConfig.Vault.Address)
+	}
+
+	expected := "http://127.0.0.1:1111"
+	if *ctConfig.Vault.Address != expected {
+		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
+	}
```
tvoran, line 235:
```suggestion
	expected := "http://127.0.0.1:1111"
	if *ctConfig.Vault.Address != expected {
		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
	}
```

command/agent/template/template_test.go
```diff
@@ -27,6 +29,233 @@ func TestNewServer(t *testing.T) {
...
```
tvoran, line 256:
```suggestion
	expected := "http://127.0.0.1:1111"
	if *ctConfig.Vault.Address != expected {
		t.Fatalf("expected %s, got %s", expected, *ctConfig.Vault.Address)
	}
```

tvoran (commented):
Some comments in the tests. It also looks like the TestCacheConfig* tests could be combined into a table test if you wanted, since the only changing parameters are the listeners and the expected address.

tvoran (approved):

calvn (approved):
Looks good, pending the change on removing client cert and keys around [here](https://github.com/hashicorp/vault/pull/10927/files#diff-90d2b6ef725d713c0515d4ae01b17766d425facd05e03283c3a6034ade76e7d7R273) that we discussed offline.

tomhjp (approved):
LGTM 👍

PR #10933, opened 2021-02-17
Update KV Secrets Engine index

onematchfox:
Just a suggestion on how to perhaps improve the language as I found myself re-reading the sentences due to the missing "either ... or ..." having been _told_ that it `can be run in two modes`.


briankassouf (approved):
Thank you!

PR #10934, opened 2021-02-17
Avoid storage operations with pendingLock is held.

mgritter:
Allow renew and lazy revoke methods to write the new lease entry without holding pendingLock. This should reduce contention on pendingLock.

This means that, in the worst case, the entry on disk and in-memory could have different expiration times.

An additional safety measure is the following
  * renew will attempt to stop the pending timer
  * if it succeeds, no concurrent renew can set a different time
  * if it fails, the renew operation fails with an error
 
This mechanism protects against concurrent renew operations with different times, but not against concurrent renew and revoke.  However, it was already the case that a Renew() operation could overwrite a LeaseEntry that has been deleted by a concurrent revokeCommon(), and this remains an open bug.

One complexity in the implementation is that should a subsequent error cause a renew to fail, we want to restore the original timer so that the lease expires as normal.

### DIFF
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -709,7 +709,6 @@ func (m *ExpirationManager) LazyRevoke(ctx context.Context, leaseID string) erro
 	le.ExpireTime = time.Now()
 	// TODO: potential race with concurrent renew.
 	if err := m.persistEntry(ctx, le); err != nil {
-		m.pendingLock.Unlock()
 		return err
 	}
 	m.updatePending(le)
@@ -853,7 +852,6 @@ func (m *ExpirationManager) RevokeByToken(ctx context.Context, te *logical.Token
 
 			// TODO: storage race with concurrent renew
 			if err := m.persistEntry(ctx, le); err != nil {
-				m.pendingLock.Unlock()
 				return err
 			}
 			m.updatePending(le)
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -991,7 +991,7 @@ func (m *ExpirationManager) Renew(ctx context.Context, leaseID string, increment
 	// renewal fails due to a later cause. On an error, re-insert the
 	// original lease saved above.
 	defer func() {
-		if renewErr != nil || renewResp == nil {
+		if renewError != nil || renewResp == nil {
 			m.updatePending(originalLease)
 		} else {
 			m.updatePending(le)
@@ -2192,15 +2192,15 @@ type leaseEntry struct {
 // Returns a shallow clone; data, secret, and auth are shared with original object.
 func (le *leaseEntry) Clone() *leaseEntry {
 	return &leaseEntry{
-		LeaseID:         le.leaseID,
+		LeaseID:         le.LeaseID,
 		ClientToken:     le.ClientToken,
 		ClientTokenType: le.ClientTokenType,
 		Path:            le.Path,
 		Data:            le.Data,
 		Secret:          le.Secret,
 		Auth:            le.Auth,
-		IssueType:       le.IssueTime,
-		ExpireTime:      le.ExireTime,
+		IssueTime:       le.IssueTime,
+		ExpireTime:      le.ExpireTime,
 		LastRenewalTime: le.LastRenewalTime,
 		Version:         le.Version,
 		namespace:       le.namespace,
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -707,16 +707,12 @@ func (m *ExpirationManager) LazyRevoke(ctx context.Context, leaseID string) erro
 	}
 
 	le.ExpireTime = time.Now()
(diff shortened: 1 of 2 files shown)


vault/expiration.go
```diff
@@ -707,16 +707,12 @@ func (m *ExpirationManager) LazyRevoke(ctx context.Context, leaseID string) erro
...
-		m.pendingLock.Lock()
-		if err := m.persistEntry(ctx, le); err != nil {
-			m.pendingLock.Unlock()
-			return err
-		}
-
-		m.updatePendingInternal(le)
+	// TODO: potential race with concurrent renew.
+	if err := m.persistEntry(ctx, le); err != nil {
 		m.pendingLock.Unlock()
```
swayne275, line 712: i'm not seeing where `m.pendingLock` was `Lock()`d in the new code?

vault/expiration.go
```diff
@@ -836,16 +851,12 @@ func (m *ExpirationManager) RevokeByToken(ctx context.Context, te *logical.Token
...
 
-			{
-				m.pendingLock.Lock()
-				if err := m.persistEntry(ctx, le); err != nil {
-					m.pendingLock.Unlock()
-					return err
-				}
-
-				m.updatePendingInternal(le)
+			// TODO: storage race with concurrent renew
```
swayne275, line 854: are these possible races what we discussed on Tues or Weds? What did we decide on for a solution?

vault/expiration.go
```diff
@@ -836,16 +851,12 @@ func (m *ExpirationManager) RevokeByToken(ctx context.Context, te *logical.Token
...
+			if err := m.persistEntry(ctx, le); err != nil {
 				m.pendingLock.Unlock()
```
swayne275, line 856: i'm also not seeing the lock associated with this unlock

vault/expiration.go
```diff
@@ -707,16 +707,12 @@ func (m *ExpirationManager) LazyRevoke(ctx context.Context, leaseID string) erro
...
 	le.ExpireTime = time.Now()
-	{
-		m.pendingLock.Lock()
-		if err := m.persistEntry(ctx, le); err != nil {
-			m.pendingLock.Unlock()
-			return err
-		}
-
-		m.updatePendingInternal(le)
+	// TODO: potential race with concurrent renew.
```
briankassouf, line 710: Should we be stopping the timer in this function to reduce concurrent renews?

vault/expiration.go
```diff
@@ -1024,19 +1044,20 @@ func (m *ExpirationManager) Renew(ctx context.Context, leaseID string, increment
...
-		m.pendingLock.Unlock()
+	// TODO: there is a race condition here with revocation, which may
+	// begin after we have stopped the entry timer, but finish before we get here
+	// and persist the entry.
+	//
+	// The stopEntry() call above protects against concurrent renewals; only one
+	// will be able to write the entry. Extending the lock to cover persistEntry,
+	// as was previously the case, would create more contention, but not solve
+	// the race with revocation.
+	if err := m.persistEntry(ctx, le); err != nil {
```
briankassouf, line 1055: Do we want to clone `le` so we can decide to put the old or new value back into pending depending on if we return an error?

vault/expiration.go
mgritter: You're correct, both these pendingLock calls should be removed.

vault/expiration.go
mgritter: I didn't feel good about that.  The reasons were:
  * we're not doing a lot (no actual revoke call, etc.) so the scope is limited
  * on persistEntry error we have to re-insert
  
Now that I solved the second point for renew() -- sort of -- the same idea could be done here.

vault/expiration.go
mgritter: The solution is "leave them in".

The locking change here means that it's possible the in-memory version (in pending) and the storage version of the leaseEntry could be inconsistent.

But, it was already possible for the renew call to re-create a leaseEntry that was just deleted.

vault/expiration.go
mgritter: If the renewEntry call succeeds, but persistEntry fails, which version do we want to have included in-memory?

My guess is that after renewEntry succeeds then some persistent change might already have been made.  I see in the documentation that the secret is supposed to be the "same", but I'm unclear what a secret backend is actually allowed/supposed to do with a RenewOperation.

vault/expiration.go
mgritter: Latest revision does call stopPending, and clones the old lease entry for re-insertion on storage failure.

vault/expiration.go
mgritter: I did add the clone, but let me know what your opinion is on failures between renewEntry and persistEntry.

PR #10935, opened 2021-02-17
Add Username Templating Concepts page

pcman312:
Adds documentation for username templating for PostgreSQL, MSSQL, MySQL, MongoDB, Cassandra, & Oracle.

Also updates the OpenLDAP docs to reference the new concepts page for username templating.


website/content/api-docs/secret/databases/couchbase.mdx
```diff
@@ -46,6 +46,9 @@ has a number of parameters to further configure a connection.
 - `bucket_name` `(string: "")` - Required for Couchbase versions prior to 6.5.0. This
   is only used to verify vault's connection to the server.
 
+- `username_template` `(string: "")` - [Template](/docs/concepts/username-templating) describing how
+  dynamic usernames are generated. Defaults to the current behavior.
```
catsby, line 50: Could we add an example of the format instead of referencing "the current behavior"? For someone unfamiliar, this doesn't give them much information and will make less sense over time

website/content/api-docs/secret/openldap.mdx
```diff
@@ -307,6 +307,10 @@ strings ("1h") or an integer number of seconds. Defaults to system/mount default
 be less than the mount max TTL (or, if not set, the system max TTL), but it is not allowed to be longer. The
 [available units](https://golang.org/pkg/time/#ParseDuration) are: `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`.
 
+The `creation_ldif`, `deletion_ldif`, `rollback_ldif`, and `username_template` fields are all templated fields. See
+[Username Templating](/docs/concepts/username-templating) for details on how to use templating. Also see
+[#Templates](Templates) for specifics on what data is available for each template.
```
catsby, line 312: `[#Templates](Templates)` dead link?  (from the PR preview, the resulting link is https://vault-mot4pi4um.vercel.app/api-docs/secret/Templates)

website/content/api-docs/secret/databases/couchbase.mdx
catsby: Alternative to describing/showing the current behavior for each instance of `username_template` doc, we could add a small section to the `Username Templating` doc page that explains to the effect of "if the engine you're using doesn't support `username_templating`, this is the basic format you'll get with usernames: <example>"

So we define it in one spot, and people who are otherwise just reading the templating docs understand the basic default

website/content/docs/concepts/username-templating.mdx
```diff
@@ -0,0 +1,143 @@
...
+sidebar_title: Username Templating
+description: >-
+  Username templating are used in some secret engines to allow operators to define
+  how dynamic usernames are generated.
+---
+
+# Username Templating
+
+Some of the secrets engines that generate dynamic users for external systems provide the ability for Vault operators
+to customize how usernames are generated for said external systems. This uses the
```
catsby, line 13:
```suggestion
to customize how usernames are generated for said external systems. This customization feature uses the
```

website/content/docs/concepts/username-templating.mdx
```diff
@@ -0,0 +1,143 @@
...
+`lowercase` - Lowercases the provided value.<br/>
+**Example**: `{{.FieldName | lowercase}}`
+
+`replace` - Find/replace on the provided value.<br/>
+**Example**: `{{.FieldName | replace - _}}`
+
+`truncate` - truncates the previous value to the provided number of characters.<br/>
+**Example**: `{{.FieldName | truncate 10}}`
+
+`truncate_sha256` - Truncates the previous value to the provided number of characters. The last 8 characters of the
```
catsby, line 37: Should we refer to the value as the "input value" and not "previous value"? I feel that may make more sense. Also `lowercase` and `replace` use "provided value" while `truncate` uses "previous value". Think we should be consistent across the functions either way

website/content/docs/concepts/username-templating.mdx
```diff
@@ -0,0 +1,143 @@
...
+truncated value will be the first 8 characters of the SHA256 hash of the truncated characters.<br/>
```
catsby, line 38:
```suggestion
new value will be replaced by the first 8 characters of the SHA256 hash of the truncated characters.<br/>
```

Assuming I understand the operation, I think this reads more clearly. Let me know if I do not understand correctly, which is entirely possibly.

website/content/docs/concepts/username-templating.mdx
```diff
@@ -0,0 +1,143 @@
...
+`uppercase` - Uppercases the provided value.<br/>
+**Example**: `{{.FieldName | uppercase}}`
+
+### Generating Values
+
+`random` - Generates a random string from lowercase letters, uppercase letters, and numbers. Must include a
+number indicating how many characters to generate.<br/>
+**Example**: `{{random 20}}` generates 20 random characters
+
+`timestamp` - The current time. Must provide a formatting string based on Go’s time package.<br/>
```
catsby, line 53: If I'm wanting this timestamp in my timezone, I need to provide the formatting string that uses my desired TZ correct? This part of time formatting always throws me. Does this function accept things like `time.RFC3339` ?

catsby (commented):
Looks great! I have some questions and suggestions, please let me know what you think

website/content/api-docs/secret/databases/couchbase.mdx
pcman312: Re: adding the default templates here: I am hesitant to add them here as 1) they're fairly long, and 2) we never specified them previously and I don't want to imply that those defaults can't change. Maybe that's a pipe dream given Hyrum's law though. I'll look into it again and see if there's something reasonable we can do here.

Re: option 2: there isn't a single default for the databases so I don't think it makes sense to put them in a central location. Each DB is different with little overlap other than the most basic of structure.

website/content/api-docs/secret/openldap.mdx
pcman312: Whoops - inverted link format `[Templates](#Templates)`

website/content/docs/concepts/username-templating.mdx
pcman312: :+1:

website/content/docs/concepts/username-templating.mdx
pcman312: Good catch. Updated. Let me know how this looks. I moved most of them over to "input" and removed "previous".

website/content/docs/concepts/username-templating.mdx
pcman312: That looks good and matches with what it does. This function is a bit complicated to explain so anything that can be done to make this clearer is appreciated.

website/content/docs/concepts/username-templating.mdx
pcman312: It does not accept `time.RFC3339` (like: `{{ timestamp time.RFC3339 }}`). I wish it did, but I don't think there is an easy way of handling that. As for timezones, if you specify `Z` I believe it will include your local timezone. In this case, it would be the local timezone of the Vault server. At the moment I don't anticipate this function being used very often. If we get more usage of it with feedback for improvements we can look into it then.

website/content/api-docs/secret/openldap.mdx
catsby: 👍 unfortunately that still doesn't go anywhere for me (https://vault-1qjyd1t1v-hashicorp.vercel.app/api-docs/secret/openldap#Templates)

website/content/docs/concepts/username-templating.mdx
catsby: LGTM

website/content/docs/concepts/username-templating.mdx
catsby: It reads clearer to me now, thanks!

website/content/docs/concepts/username-templating.mdx
catsby: 👍

website/content/api-docs/secret/openldap.mdx
catsby: Fixed in 3fdec70 and da4182a

catsby (approved):
LGTM 🚀

PR #10938, opened 2021-02-18
agent: persistent caching support

tvoran:
Adds the option of a write-through cache to the agent, backed by boltdb, with the aim of preserving the renewal of leases and tokens across agent restarts.

This functionality is configured through a new `persist` stanza within the `cache` portion of the vault agent config file:

```hcl
cache {
  persist "kubernetes" {
    path = "/path/to/db/"
    exit_on_err = false
    keep_after_import = false
    service_account_token_file = "/path/to/jwt"
}
```

- The persist `type` specifies the key protection scheme to use for the key encrypting/decrypting the persisted data. Currently "kubernetes" is supported, where the encryption key is saved in the boltdb, using the service account jwt as the AAD (extra data) used to encrypt and decrypt.
  - `service_account_token_file` is optional, and defaults to "/var/run/secrets/kubernetes.io/serviceaccount/token"

- The `path` parameter is required, and is the file system location to write the cache db file. The file name for the cache file is `vault-agent-cache.db`. So in the above example the full file and path would be `/path/to/db/vault-agent-cache.db`.
  - If the db file exists at path when the agent starts, anything in the db file will be imported into the vault-agent in-memory cache. Otherwise a new db file is created.
- If `keep_after_import = false`, the db file will then be deleted after importing, and agent tokens and secrets will only be cached in memory.
- If `exit_on_err = true`, errors during import of an existing db cache will cause the agent to exit.

The bolt db is laid out in multiple buckets:
- The "meta" bucket contains a `version: 1` entry for use in the future if migrations are required. It also contains the most recent auto-auth token cached by the agent, and the key material used to decrypt and restore an existing cache. This key material may either be the key, or some other token used to retrieve the key.
- There are three buckets for the different cached item types: "token", "auth-lease", and "secret-lease". They're separated this way due to dependencies during the restore process (e.g. secret-leases depend on token's and auto-lease's).


command/agent.go
```diff
@@ -30,6 +31,8 @@ import (
 	"github.com/hashicorp/vault/command/agent/auth/kerberos"
 	"github.com/hashicorp/vault/command/agent/auth/kubernetes"
 	"github.com/hashicorp/vault/command/agent/cache"
+	"github.com/hashicorp/vault/command/agent/cache/cachememdb"
+	"github.com/hashicorp/vault/command/agent/cache/cachepersist"
```
jasonodonnell, line 35: Wonder if this should be `cacheboltdb` instead?

command/agent/cache/cachepersist/bolt.go
```diff
@@ -0,0 +1,216 @@
...
+	bolt "go.etcd.io/bbolt"
+)
+
+const (
+	// Keep track of schema version for future migrations
+	storageVersionKey = "version"
+	storageVersion    = "1"
+
+	// CacheFileName - filename for the persistent cache file
+	CacheFileName = "vault-agent-cache.db"
```
jasonodonnell, line 18:
```suggestion
	DatabaseFileName = "vault-agent-cache.db"
```

command/agent/cache/cachepersist/bolt.go
```diff
@@ -0,0 +1,216 @@
...
+	RootBucket string
+	Logger     hclog.Logger
+	Encryption Encryption
+}
+
+// NewBoltStorage opens a new bolt db at the specified file path and returns it.
+// If the db already exists the buckets will just be created if they don't
+// exist.
+func NewBoltStorage(config *BoltStorageConfig) (*BoltStorage, error) {
+	cachePath := filepath.Join(config.Path, CacheFileName)
```
jasonodonnell, line 46:
```suggestion
	cachePath := filepath.Join(config.Path, DatabaseFileName)
```

command/agent/cache/cachepersist/bolt.go
```diff
@@ -0,0 +1,216 @@
...
+	db, err := bolt.Open(cachePath, 0600, &bolt.Options{Timeout: 1 * time.Second})
```
jasonodonnell, line 47:
```suggestion
	db, err := bolt.Open(dbPath, 0600, &bolt.Options{Timeout: 1 * time.Second})
```

command/agent.go
tvoran: Sure, I like that name.

command/agent/cache/cachepersist/bolt.go
tvoran: Hmm, both make sense to me, though I do see this sort of thing referred to as dbpath and dbfile in the raft package, so ok.

command/agent/cache/cacheboltdb/bolt.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+// SetKeyMgr sets the encryption for a bolt storage
+func (b *BoltStorage) SetKeyMgr(km crypto.KeyManager) {
+	b.keyMgr = km
+}
+
+// Set an index in bolt storage
+func (b *BoltStorage) Set(ctx context.Context, id string, plainText []byte, indexType string) error {
+	cipherText, err := b.keyMgr.Encrypt(ctx, plainText, []byte(b.aad))
```
calvn, line 140: This feels like key operation is leaking a bit into the `KeyManager` interface. `KeyManager` should just be in charge of the key management such as retrieval, renewal/rewrap, etc.`BoltStorage` would have `wrapping.Wrapper` as a field to handle the encryption. Instead of passing in KeyManager, you pass in the key material and aad.

Upon reading the code a bit further, I feel like the correct layout would be to have an implementation of KeyManager on BoltDB. The `KubeEncryptionKey` implementation has nothing that's specific to k8s. On the agent config side, we can still call it `crypto "kubernetes"`, but this is really being able to get and set the key from and to BoltDB (and in fact we're already exposing methods such as `GetKey` and `SetKey` which is pretty much what KeyManager should be doing).

command/agent/cache/lease_cache.go
```diff
@@ -850,6 +886,202 @@ func (c *LeaseCache) handleRevocationRequest(ctx context.Context, req *SendReque
...
+	if c.ps != nil {
+		b, err := index.Serialize()
+		if err != nil {
+			return err
+		}
+
+		if err := c.ps.Set(ctx, index.ID, b, indexType); err != nil {
+			return err
+		}
+		c.logger.Debug("set entry in persistent storage", "type", indexType, "path", index.RequestPath, "id", index.ID, "lease", index.Lease)
```
calvn, line 905: Should we set this to Trace to make it consistent with Evict?

command/agent/cache/lease_cache.go
```diff
@@ -850,6 +886,202 @@ func (c *LeaseCache) handleRevocationRequest(ctx context.Context, req *SendReque
...
+}
+
+// Flush the cachememdb and persistent cache (if enabled)
+func (c *LeaseCache) Flush() error {
+	if err := c.db.Flush(); err != nil {
+		return err
+	}
+
+	if c.ps != nil {
+		return c.ps.Clear()
```
calvn, line 935: We can add a log message in here as well.

command/agent.go
```diff
@@ -470,6 +476,136 @@ func (c *AgentCommand) Run(args []string) int {
...
+					Logger: cacheLogger.Named("cacheboltdb"),
+					AAD:    aad,
+				})
+				if err != nil {
+					c.UI.Error(fmt.Sprintf("Error opening persistent cache: %v", err))
+					return 1
+				}
+				// Grab the key material from bolt, then setup encryption so
+				// that restore is possible
+				key, err := ps.GetKey()
```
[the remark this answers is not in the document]
jasonodonnell: This code here is during a restore, so we need to get the key from bolt. That key may definitely be remote, so we do need this to be flexible enough to do that, but in this case (Kubernetes), it's stored in the DB.

command/agent/cache/cacheboltdb/bolt.go
jasonodonnell: Originally the KubeEncryptionKey code did have Kube specific code, namely getting the service account JWT which we use for AAD. We can add that back in so the caller doesn't need to know about it and that would make it more Kube specific.

command/agent/config/test-fixtures/config-cache.hcl
```diff
@@ -21,6 +21,15 @@ auto_auth {
 
 cache {
 	use_auto_auth_token = true
+	persist = {
+		path = "/tmp/bolt-file.db"
+		keep_after_import = true
+		exit_on_err = true
+		crypto = {
```
jasonodonnell, line 28: Is this a bug? Should it be `crypto "kubernetes" {}` or are we testing for this?

tvoran: I believe this is the non-embedded way to specify the type, going off of these test fixtures:

embedded:
https://github.com/hashicorp/vault/blob/e0b9cf80a415b109cb87d917a35344118ab8a12b/command/agent/config/test-fixtures/config-embedded-type.hcl#L4-L10

not embedded:
https://github.com/hashicorp/vault/blob/e0b9cf80a415b109cb87d917a35344118ab8a12b/command/agent/config/test-fixtures/config.hcl#L4-L10

command/agent/cache/lease_cache.go
tvoran: Added in add5757bc1

command/agent/cache/lease_cache.go
tvoran: Changed in add5757bc1

command/agent/cache/lease_cache.go
```diff
@@ -83,6 +85,13 @@ type LeaseCache struct {
 
 	// inflightCache keeps track of inflight requests
 	inflightCache *gocache.Cache
+
+	// ps is the persistent storage for tokens and leases
+	ps *cacheboltdb.BoltStorage
+
+	// shuttingDown is used to determine if cache needs to be evicted or not
+	// when the context is cancelled
+	shuttingDown bool
```
calvn, line 94: I'd suggest using an [atomic bool](https://pkg.go.dev/go.uber.org/atomic#Bool) value since this can be modified by rungroups.

command/agent/cache/cacheboltdb/bolt.go
```diff
@@ -0,0 +1,330 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create auth lease sub-bucket: %w", err)
+	}
+	_, err = root.CreateBucketIfNotExists([]byte(SecretLeaseType))
+	if err != nil {
+		return fmt.Errorf("failed to create secret lease sub-bucket: %w", err)
+	}
+
+	// check and set file version in the root bucket
+	version := root.Get([]byte(storageVersionKey))
```
calvn, line 121: Can we do this first and bail early on error?

command/agent/cache/lease_cache.go
tvoran: Changed in https://github.com/hashicorp/vault/pull/10938/commits/0dd3ae4cdddefe9fa567b29be9c25fbd3a51fb93

command/agent/cache/cacheboltdb/bolt.go
tvoran: Yep, changed in https://github.com/hashicorp/vault/pull/10938/commits/0dd3ae4cdddefe9fa567b29be9c25fbd3a51fb93

command/agent/cache/cacheboltdb/bolt.go
```diff
@@ -0,0 +1,344 @@
...
+	rootBucketName = "root"
+
+	// SecretLeaseType - Bucket/type for leases with secret info
+	SecretLeaseType = "secret-lease"
+
+	// AuthLeaseType - Bucket/type for leases with auth info
+	AuthLeaseType = "auth-lease"
+
+	// TokenType - Bucket/type for auto-auth tokens
+	TokenType = "token"
```
calvn, line 37: Looks like this bucket is carved out for the auto-auth token, but can that be stored in the root directly?

tvoran: Perhaps? Though I left it as a separate bucket since this is how the cachememdb behaves, namely old auto-auth tokens are not removed if a new one comes in, and they aren't expired. So I added this bucket out of abundance of caution, in case a lease that needed to be restored relied on a previous auto-auth token.

command/agent/cache/lease_cache.go
```diff
@@ -997,3 +1264,30 @@ func parseCacheClearInput(req *cacheClearRequest) (*cacheClearInput, error) {
...
+		return false, fmt.Errorf("failed to parse response as secret: %w", err)
+	}
+
+	elapsed := currentTime.Sub(index.LastRenewed)
+	var leaseDuration int
+	switch index.Type {
+	case cacheboltdb.AuthLeaseType:
+		leaseDuration = secret.Auth.LeaseDuration
+	default:
+		leaseDuration = secret.LeaseDuration
```
calvn, line 1285: Can we add a case for handling this by checking for `cacheboltdb.SecretLeaseType` and error on the default case? If somehow we're checking against `TokenType` this may behave unexpectedly.

command/agent.go
```diff
@@ -490,6 +496,152 @@ func (c *AgentCommand) Run(args []string) int {
...
+			if config.Cache.Persist.Path == "" {
+				c.UI.Error("must specify persistent cache path")
+				return 1
+			}
+
+			// Set AAD based on key protection type
+			var aad string
+			switch config.Cache.Persist.Type {
+			case "kubernetes":
+				aad, err = cacheboltdb.GetServiceAccountJWT(config.Cache.Persist.ServiceAccountTokenFile)
```
calvn, line 510: The `cacheboltdb` package does not need to be aware of where/how it's getting this aad value so maybe turn this into a private func, `getServiceAccountJWT`, that lives directly within this package.

command/agent/cache/lease_cache.go
```diff
@@ -850,6 +907,213 @@ func (c *LeaseCache) handleRevocationRequest(ctx context.Context, req *SendReque
...
+			return err
+		}
+		c.logger.Trace("restored lease", "id", newIndex.ID, "path", newIndex.RequestPath)
+	}
+	return nil
+}
+
+// ReCreateLeaseRenewCtx re-creates a RenewCtx for an index object and starts
+// the watcher go routine
+func (c *LeaseCache) ReCreateLeaseRenewCtx(index *cachememdb.Index) error {
```
calvn, line 1041: Consider making this private and renaming it to `restoreLeaseRenewCtx` to match the caller's func name

command/agent/cache/lease_cache.go
tvoran: Added in cee5bfb

command/agent.go
tvoran: Done in cee5bfb

command/agent/cache/lease_cache.go
tvoran: Done in cee5bfb

calvn (approved):

jasonodonnell (approved):

PR #10940, opened 2021-02-18
Vault Dependency Upgrades 

HridoyRoy:
This is the same PR as: https://github.com/hashicorp/vault/pull/10903


builtin/logical/transit/path_encrypt.go
```diff
@@ -159,7 +159,7 @@ func decodeBatchRequestItems(src interface{}, dst *[]BatchRequestItem) error {
 			} else if casted, ok := v.(string); ok {
 				(*dst)[i].Context = casted
 			} else {
-				errs.Errors = append(errs.Errors, fmt.Sprintf("'[%d].context' expected type 'string', got unconvertible type '%T'", i, item["context"]))
+				errs.Errors = append(errs.Errors, fmt.Sprintf("'[%d].context' expected type 'string', got unconvertible type '%T', value: '%v'", i, item["context"], item["context"]))
```
briankassouf, line 162: How come these changed in this PR? I'm not sure we want to be returning these as errors since they may get logged. Especially the plaintext one below.

HridoyRoy: These changed due to the upgrade in the map structure here: https://github.com/mitchellh/mapstructure/blob/07998cfd7ab2f7f7e5ddce402c4edf38317451eb/mapstructure_examples_test.go#L65

go.mod
```diff
@@ -100,29 +114,29 @@ require (
...
-	github.com/mitchellh/copystructure v1.0.0
+	github.com/michaelklishin/rabbit-hole v1.5.0
+	github.com/mitchellh/cli v1.1.0
+	github.com/mitchellh/copystructure v1.1.1
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/mitchellh/go-testing-interface v1.14.0
-	github.com/mitchellh/gox v1.0.1
-	github.com/mitchellh/mapstructure v1.3.3
+	github.com/mitchellh/go-testing-interface v1.14.2-0.20210217184823-a52172cd2f64
+	github.com/mitchellh/gox v0.4.0
```
calvn, line 132: How come gox got downgraded?

HridoyRoy: Good catch -- I think it had something to do with the versioning issue with go-testing-interface, but it was an unnecessary straggler dependency. Upgraded it back.

PR #10941, opened 2021-02-18
Adds method for rotating a policy without a persist

austingebauer:
## Overview

This PR adds a `RotateInMemory()` method to `keysutil.Policy`. The method rotates a policy without a call to `Persist()`. This method is useful in situations where rotating a policy is part of a multistep operation that may partially fail.

This PR is primarily moving around existing code. The signature and behavior of `Rotate()` is unchanged.

## Testing

I ran the following tests with the changes in this PR to ensure that changes to `Rotate()` don't introduce regressions.

Test output from keysutil package:
```
$ go test -v
=== RUN   TestEncrytedKeysStorage_BadPolicy
--- PASS: TestEncrytedKeysStorage_BadPolicy (0.00s)
=== RUN   TestEncryptedKeysStorage_List
--- PASS: TestEncryptedKeysStorage_List (0.00s)
=== RUN   TestEncryptedKeysStorage_CRUD
--- PASS: TestEncryptedKeysStorage_CRUD (0.00s)
=== RUN   TestPolicy_KeyEntryMapUpgrade
--- PASS: TestPolicy_KeyEntryMapUpgrade (0.00s)
=== RUN   Test_KeyUpgrade
--- PASS: Test_KeyUpgrade (0.00s)
=== RUN   Test_ArchivingUpgrade
--- PASS: Test_ArchivingUpgrade (0.00s)
=== RUN   Test_Archiving
--- PASS: Test_Archiving (0.00s)
=== RUN   Test_StorageErrorSafety
--- PASS: Test_StorageErrorSafety (0.00s)
=== RUN   Test_BadUpgrade
--- PASS: Test_BadUpgrade (0.00s)
=== RUN   Test_BadArchive
--- PASS: Test_BadArchive (0.00s)
PASS
ok      github.com/hashicorp/vault/sdk/helper/keysutil  0.190s
```

Test output from transit package:
```
$ go test ./...
ok      github.com/hashicorp/vault/builtin/logical/transit      40.500s
?       github.com/hashicorp/vault/builtin/logical/transit/cmd/transit  [no test files]
```

Test output from transform secrets engine:
```
$ go test ./...
ok      github.com/hashicorp/vault-plugin-secrets-transform     79.825s
?       github.com/hashicorp/vault-plugin-secrets-transform/cmd/vault-plugin-secrets-transform  [no test files]
?       github.com/hashicorp/vault-plugin-secrets-transform/helpers/testhelpers/docker  [no test files]
?       github.com/hashicorp/vault-plugin-secrets-transform/helpers/testhelpers/mysql   [no test files]
?       github.com/hashicorp/vault-plugin-secrets-transform/helpers/testhelpers/postgresql      [no test files]
```

### DIFF
--- a/sdk/helper/keysutil/policy.go
+++ b/sdk/helper/keysutil/policy.go
@@ -1452,14 +1452,14 @@ func (p *Policy) RotateInMemory(randReader io.Reader) (retErr error) {
 		}
 	}
 
+	p.LatestVersion += 1
+
 	if p.Keys == nil {
 		// This is an initial key rotation when generating a new policy. We
 		// don't need to call migrate here because if we've called getPolicy to
 		// get the policy in the first place it will have been run.
 		p.Keys = keyEntryMap{}
 	}
-
-	p.LatestVersion += 1
 	p.Keys[strconv.Itoa(p.LatestVersion)] = entry
 
 	// This ensures that with new key creations min decryption version is set
--- a/sdk/helper/keysutil/policy.go
+++ b/sdk/helper/keysutil/policy.go
@@ -1357,7 +1357,7 @@ func (p *Policy) Rotate(ctx context.Context, storage logical.Storage, randReader
 		}
 	}()
 
-	if err := p.RotateInMemory(randReader, false); err != nil {
+	if err := p.RotateInMemory(randReader); err != nil {
 		return err
 	}
 
@@ -1365,38 +1365,7 @@ func (p *Policy) Rotate(ctx context.Context, storage logical.Storage, randReader
 }
 
 // RotateInMemory rotates the policy but does not persist it to storage.
-// If restoreOnError is true and the rotation partially fails, the policy
-// state will be restored.
-func (p *Policy) RotateInMemory(randReader io.Reader, restoreOnError bool) (retErr error) {
-	if restoreOnError {
-		priorLatestVersion := p.LatestVersion
-		priorMinDecryptionVersion := p.MinDecryptionVersion
-		var priorKeys keyEntryMap
-
-		if p.Keys != nil {
-			priorKeys = keyEntryMap{}
-			for k, v := range p.Keys {
-				priorKeys[k] = v
-			}
-		}
-
-		defer func() {
-			if retErr != nil {
-				p.LatestVersion = priorLatestVersion
-				p.MinDecryptionVersion = priorMinDecryptionVersion
-				p.Keys = priorKeys
-			}
-		}()
-	}
-
-	if p.Keys == nil {
-		// This is an initial key rotation when generating a new policy. We
-		// don't need to call migrate here because if we've called getPolicy to
-		// get the policy in the first place it will have been run.
-		p.Keys = keyEntryMap{}
-	}
-
-	p.LatestVersion += 1
+func (p *Policy) RotateInMemory(randReader io.Reader) (retErr error) {
 	now := time.Now()
 	entry := KeyEntry{
 		CreationTime:           now,
@@ -1483,6 +1452,14 @@ func (p *Policy) RotateInMemory(randReader io.Reader, restoreOnError bool) (retE
 		}
 	}
 
+	if p.Keys == nil {
+		// This is an initial key rotation when generating a new policy. We
+		// don't need to call migrate here because if we've called getPolicy to
+		// get the policy in the first place it will have been run.
+		p.Keys = keyEntryMap{}
+	}
+
+	p.LatestVersion += 1
 	p.Keys[strconv.Itoa(p.LatestVersion)] = entry
 
 	// This ensures that with new key creations min decryption version is set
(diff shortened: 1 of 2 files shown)


sdk/helper/keysutil/policy.go
```diff
@@ -1355,6 +1357,38 @@ func (p *Policy) Rotate(ctx context.Context, storage logical.Storage, randReader
...
+	}
+
+	return p.Persist(ctx, storage)
+}
+
+// RotateInMemory rotates the policy but does not persist it to storage.
+// If restoreOnError is true and the rotation partially fails, the policy
+// state will be restored.
+func (p *Policy) RotateInMemory(randReader io.Reader, restoreOnError bool) (retErr error) {
+	if restoreOnError {
```
pcman312, line 1371: This feels overcomplicated and prone to unexpected behavior (particularly due to the defer operating on the returned error). What about changing the function such that it runs all of the errorable operations without modifying `p` first. Then at the end, you modify `p` with the appropriate new data.

austingebauer: I agree with you wrt how restoring the policy state is handled. This isn't new code being introduced. See that it already exists in [policy.go#L1338-L1356](https://github.com/hashicorp/vault/blob/master/sdk/helper/keysutil/policy.go#L1338-L1356). I've just moved existing code around to support being able to rotate the policy without the final persist.

This code is being used in other places (e.g., transit, transform), so I'm hesitant to modify it and risk introducing bugs.

sdk/helper/keysutil/policy.go
```diff
@@ -1335,6 +1335,8 @@ func (p *Policy) VerifySignature(context, input []byte, hashAlgorithm HashType,
 	}
 }
 
+// Rotate rotates the policy and persists it to storage.
+// If the rotation partially fails, the policy state will be restored.
```
calvn, line 1339: This comment is not correct right (and behavior remains unchanged), since we're passing `false` to `restoreOnError`?

austingebauer: I'm fairly confident that the comment is correct. The logic for restoring the policy state exists in both `Rotate` and `RotateInMemory`.

In `Rotate`, we need to restore the policy state for a failure to both `RotateInMemory` and `Persist`. That's why [this snippet](https://github.com/hashicorp/vault/blob/policy-rotate-no-persist/sdk/helper/keysutil/policy.go#L1338-L1356) of code exists in both methods.

calvn: Oh I didn't realize that Rotate had it's own restore defer logic (that was collapsed on GH).

calvn: If we're restoring within `RotateInMemory` do we need to do that again on `Rotate` itself?

austingebauer: Ya, we need to do restore in `Rotate`, since it's possible to fail at the call to `Persist`.

austingebauer: This test case failed without the restore logic in `Rotate` ([policy_test.go#L461-L473](https://github.com/hashicorp/vault/blob/master/sdk/helper/keysutil/policy_test.go#L461-L473)). If the call to persist fails, it's expected that the policy state is restored.

calvn: Ah that's right, we have to roll back on persist failure as well.

sdk/helper/keysutil/policy.go
```diff
@@ -1355,6 +1357,38 @@ func (p *Policy) Rotate(ctx context.Context, storage logical.Storage, randReader
...
+		defer func() {
+			if retErr != nil {
+				p.LatestVersion = priorLatestVersion
+				p.MinDecryptionVersion = priorMinDecryptionVersion
+				p.Keys = priorKeys
+			}
+		}()
+	}
+
 	if p.Keys == nil {
```
calvn, line 1392: I haven't tested this yet, but if you move this to L1485, you might be able to skip on the defer restore (though you'd still have to do it on `Rotate`).

calvn, line 1392: I haven't tested this yet, but if you move this to L1485, you might be able to skip on the defer restore (though you'd still have to do it on `Rotate`).

austingebauer: I think you're right. Let me test this out. If it works, I'll remove the defer restore from `RotateInMemory` and keep it in `Rotate` only.

austingebauer: That worked out, thanks 👍 I've removed the defer restore from `RotateInMemory` in ca335cf. The logic still exists in `Rotate`. I also reran the tests without any failures.

sdk/helper/keysutil/policy.go
```diff
@@ -1449,6 +1452,14 @@ func (p *Policy) Rotate(ctx context.Context, storage logical.Storage, randReader
 		}
 	}
 
+	if p.Keys == nil {
```
calvn, line 1455: Nit: maybe do `p.LatestVersion += 1` first to keep `p.Keys` context closer to the assignment.

calvn (approved):
One comment, otherwise LGTM.

sdk/helper/keysutil/policy.go
austingebauer: Done in 0e7b526.

PR #10942, opened 2021-02-18
Add TOTP support to Okta Auth

kalafut:
Okta Verify Push is currently supported. This PR adds support for the TOTP method, via an optional login parameter.

### DIFF
--- a/builtin/credential/okta/backend.go
+++ b/builtin/credential/okta/backend.go
@@ -12,6 +12,11 @@ import (
 	"github.com/okta/okta-sdk-golang/v2/okta"
 )
 
+const (
+	mfaPushMethod = "push"
+	mfaTOTPMethod = "token:software:totp"
+)
+
 func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend, error) {
 	b := Backend()
 	if err := b.Setup(ctx, conf); err != nil {
@@ -57,7 +62,7 @@ type backend struct {
 	*framework.Backend
 }
 
-func (b *backend) Login(ctx context.Context, req *logical.Request, username string, password string) ([]string, *logical.Response, []string, error) {
+func (b *backend) Login(ctx context.Context, req *logical.Request, username, password, totp string) ([]string, *logical.Response, []string, error) {
 	cfg, err := b.Config(ctx, req.Storage)
 	if err != nil {
 		return nil, nil, nil, err
@@ -160,6 +165,8 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username stri
 		// active factor enrollment). This bypass removes visibility
 		// into the authenticating user's password expiry, but still ensures the
 		// credentials are valid and the user is not locked out.
+		//
+		// API reference: https://developer.okta.com/docs/reference/api/authn/#verify-factor
 		if cfg.BypassOktaMFA {
 			result.Status = "SUCCESS"
 			break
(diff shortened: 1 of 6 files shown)


builtin/credential/okta/cli.go
```diff
@@ -38,13 +38,8 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
 		"password": password,
 	}
 
-	mfa_method, ok := m["method"]
```
calvn, line 41: Were these use copy-pasted values that we're removing?

builtin/credential/okta/path_login.go
```diff
@@ -117,7 +122,8 @@ func (b *backend) pathLoginRenew(ctx context.Context, req *logical.Request, d *f
 		return nil, err
 	}
 
-	loginPolicies, resp, groupNames, err := b.Login(ctx, req, username, password)
+	// No TOTP entry is possible on renew. If push MFA is enabled it will still be triggered, however.
```
calvn, line 125: So what happens if push MFA is enabled and a renewal happens since we are passing in an empty passcode?

builtin/credential/okta/backend.go
```diff
@@ -168,22 +175,39 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username stri
...
+
+		// Okta push and totp are currently supported. If a totp passcode is provided during
+		// login and is supported, that will be the preferred method.
 		for _, v := range result.Embedded.Factors {
-			if v.Type == "push" && v.Provider == "OKTA" {
+			if v.Provider != "OKTA" {
+				continue
+			}
+
+			if v.Type == mfaTOTPMethod && totp != "" {
```
calvn, line 186: Should we just check `v.Type == mfaTOTPMethod` in here and do the `totp` check separately afterwards? The error response below is not entirely accurate if a TOTP factor was set but a totp was not provided in the login request.

builtin/credential/okta/cli.go
kalafut: I thought so but after looking deeper I see they could be used by the legacy MFA system. I've restored them.

builtin/credential/okta/path_login.go
kalafut: You get a push notification.

builtin/credential/okta/backend.go
kalafut: That doesn't quite work since if push is available and they didn't provide `totp`, we should just use push. But I did just redo the handling of these (https://github.com/hashicorp/vault/pull/10942/commits/0514d1378d732a156172580ccb05e08c28679675) to allow separate, more specific error messages.

builtin/credential/okta/backend.go
```diff
@@ -57,7 +62,7 @@ type backend struct {
 	*framework.Backend
 }
 
-func (b *backend) Login(ctx context.Context, req *logical.Request, username string, password string) ([]string, *logical.Response, []string, error) {
+func (b *backend) Login(ctx context.Context, req *logical.Request, username, password, totp string) ([]string, *logical.Response, []string, error) {
```
pcman312, line 65: Returning 4 parameters is a code smell. I realize it's a bit outside the context of this change, but can this be refactored to return fewer values? Barring that, returning them in a response object type? At the minimum, labeling the return values would help.

builtin/credential/okta/path_login.go
pcman312: Can you make note of this behavior in the comments here?

builtin/credential/okta/backend.go
```diff
@@ -160,30 +165,52 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username stri
...
+		var selectedFactor, totpFactor, pushFactor *mfaFactor
 
-		var selectedFactor mfaFactor
-		// only okta push is currently supported
+		// Scan for available factors
 		for _, v := range result.Embedded.Factors {
-			if v.Type == "push" && v.Provider == "OKTA" {
-				factorAvailable = true
-				selectedFactor = v
+			v := v
```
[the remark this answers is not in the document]
kalafut: Without a new copy I'll get the same pointer every time since the loop variable doesn't change. I can add a comment.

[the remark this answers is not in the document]
pcman312: Ah that makes sense. A comment would be good, but I also think having it be a different variable name may also help.

builtin/credential/okta/backend.go
```diff
@@ -172,30 +172,33 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username, pas
...
 			}
 		}
 
-		if !factorAvailable {
+		// Okta push and totp are currently supported. If a totp passcode is provided during
+		// login and is supported, that will be the preferred method.
+		switch {
+		case totpFactor != nil && totp != "":
+			selectedFactor = totpFactor
+		case pushFactor != nil:
```
calvn, line 197: This switch is pretty nice since it allows us to attempt doing push MFA if a totp is not provided and there are multiple factors enabled.

calvn (approved):

builtin/credential/okta/path_login.go
kalafut: Updated

PR #10948, opened 2021-02-19
Adds "raw(/pem)" format to individual cert routes closes #10947

### LINKED ISSUE #10947: Enable a "forward rolling" PKI strategy that does not break AIA or CRLs

As documented in other issues the pki rotation isn't as clear or clean as it could be.

Related to https://github.com/hashicorp/vault/issues/2581

Here are the steps to create a "forward rolling" PKI strategy that does not break AIA or CRLs. Assuming new endpoints are added to enable this process.

Given the endpoints "pki/cert/{serial}/raw(/pem)" are implemented (similar to "/pki/ca(/pem)" endpoints).

Given you have a root or intermidate at "pki-int-1" mount.

Create Intermediate CA at "pki" mount. Sign with "pki-int-1". Set signed intermediate on "pki". The "pki" cert should now exist in both "pki-int-1" and "pki".

Create/sign another mount at "pki1", this is just for bootstrapping CRLs.

Set "issuing_certificates" to "pki-int-1/cert/{serial}/raw" with the serial of the intermediate cert.

Set "crl_distribution_points" to "/pki/crl" & "/pki1/crl".

Any certs created on "pki" mount will have AIA info pointing at the exact cert in "pki-int-1" and CRL pointing to both "/pki/crl" & "/pki1/crl".

Process to Roll Forward"pki"

Add a new "pki-next" create and sign against "pki-int-1" same as "pki".

Set "issuing_certificates" to "pki-int-1/cert/{serial-for-new-pki-next}/raw"

Set "crl_distribution_points" to "/pki/crl" & "/pki1/crl", same as bootstrapping.

Use `/sys/remount` to move "pki1" to "pki2" (or drop "pki1"), "pki" to "pki1", "pki-next" to "pki". This should be close to zero downtime if scripted.

CRLs work because they are always looking in both locations (previous pki and new pki). CRLs for certs on the old "pki1" should have all rolled/off expired by now anyway (assuming we're rotating at a reasonable interval).

AIA works because they're set to at exact paths under "pki-int-1".

Process to Key-Roll "pki-int-1"

Create/sign the new root or intermediate under "pki-int-2". Use "pki-int-2" for future "pki" "issuing_certificates".

Instead of using the "rotate between two pki mounts" suggested here https://github.com/hashicorp/vault/issues/2581#issuecomment-293255409, this allows consumers to always use "pki" mount. Using the "rotate between two pki mounts" method means all consumers need the extra knowledge of which pki mount is the active pki mount. It's easy to store that in consul but adds that complexity to all consumers. This way we keep the complexity on the pki configuration/rolling strategy and the consumers can remain "stupid".

The alternatives are to use the "rotate between two pki mounts" method or don't use rely on AIA/CRL.

abriening:
Similar to "/pki/ca(/pem)" routes to retrieve
certificates in raw or pem formats, this adds
"pki/cert/{serial}/raw(/pem)" routes for any
certificate.

### DIFF
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -71,6 +71,7 @@ func Backend(conf *logical.BackendConfig) *backend {
 			pathFetchCAChain(&b),
 			pathFetchCRL(&b),
 			pathFetchCRLViaCertPath(&b),
+			pathFetchValidRaw(&b),
 			pathFetchValid(&b),
 			pathFetchListCerts(&b),
 			pathRevoke(&b),
--- a/builtin/logical/pki/path_fetch.go
+++ b/builtin/logical/pki/path_fetch.go
@@ -53,6 +53,27 @@ func pathFetchCRL(b *backend) *framework.Path {
 	}
 }
 
+// Returns any valid (non-revoked) cert in raw format.
+func pathFetchValidRaw(b *backend) *framework.Path {
+	return &framework.Path{
+		Pattern: `cert/(?P<serial>[0-9A-Fa-f-:]+)/raw(/pem)?`,
+		Fields: map[string]*framework.FieldSchema{
+			"serial": &framework.FieldSchema{
+				Type: framework.TypeString,
+				Description: `Certificate serial number, in colon- or
+hyphen-separated octal`,
+			},
+		},
+
+		Callbacks: map[logical.Operation]framework.OperationFunc{
+			logical.ReadOperation: b.pathFetchRead,
+		},
+
+		HelpSynopsis:    pathFetchHelpSyn,
+		HelpDescription: pathFetchHelpDesc,
+	}
+}
+
 // Returns any valid (non-revoked) cert. Since "ca" fits the pattern, this path
 // also handles returning the CA cert in a non-raw format.
 func pathFetchValid(b *backend) *framework.Path {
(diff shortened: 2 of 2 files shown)


abriening:
I did not see any existing tests for the `raw(/pem)` routes so I didn't add any. The existing `builtin/logical/pki/backend_test.go` tests are passing. Looks like `build-common-layers` depends on inaccessible SSH Keys so the build is failing.

cipherboy:
\o Greetings @abriening, sorry about the delay. We've discussed this PR and are happy to accept it.

Do you have time to add a changelog entry, tests, and docs? Former would probably be something like:

<pre>```release-note:improvement
secrets/pki: Add ability to fetch individual certificate DER and PEM
```</pre>

Or something similar, in a file named `/changelog/10948.txt`.

For docs, you could add it to `website/content/api-docs/secret/pki.mdx` near the top, after `Read Certificate`. 

For tests, in [`builtin/logical/pki/backend_test.go`](https://github.com/hashicorp/vault/blob/main/builtin/logical/pki/backend_test.go#L2868-L2887), you could add one there since the certificate has a known serial number... It looks like `req := client.NewRequest("GET", "/pki/cert/f00bar/raw/pem")` would do what you need, plus `client.RawRequest(req)`. This'll give you a Response object you can read for the PEM/DER. :-)

Let me know if you're not interested and I'll add these and cherry-pick. Thank you!

abriening:
Will do.

abriening:
@cipherboy Rebased against master since it's been a while. Added docs, changelog entry & backend test. Tested directly against the Backend, since that's how most other tests in that suite were running. 

Not really needed for this PR but testing `content-type == "application/pkix-cert"` does have me wondering if that is correct for PEM. Should it be `application/x-pem-file`? But that's another MR for another day.

cipherboy (approved):

cipherboy (approved):

sgmiller (approved):

cipherboy:
Many thanks for your contribution @abriening! :-)

PR #10949, opened 2021-02-19
UI/Database Secrets Engine cleanup

chelshaw:
A few straggling issues from the main sidebranch.

- Serialization for non-array values specific to MongoDB
- Fix defaultShown on info-table-row to not show default unless there is no value
- Remove root rotation statements from mongodb connection fields
- wrap router in try/catch if in promise's then


Monkeychip (approved):
Nice work.  Thanks for tackling these!

PR #10953, opened 2021-02-19
Add IAM tagging support for iam_user roles in AWS secret engine

Valarissa:
This PR seeks to introduce the ability to add tags to user credentials generated by the AWS secret engine.

Details in  usage are provided in the docs that have been updated with this feature, however, for ease of use, the following is an example of adding two tags to a role:
```
vault write aws/roles/my-role \
credential_type=iam_user \
iam_tags=test_key=test_value \
iam_tags=test_key2=test_value2 \
policy_document=-<<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": "ec2:*",
      "Resource": "*"
    }
  ]
}
EOF
```

### DIFF
--- a/builtin/logical/aws/backend_test.go
+++ b/builtin/logical/aws/backend_test.go
@@ -1444,8 +1444,8 @@ func TestBackend_iamTagsCrud(t *testing.T) {
 		LogicalBackend: getBackend(t),
 		Steps: []logicaltest.TestStep{
 			testAccStepConfig(t),
-			testAccStepWriteIamTags(t, "test", map[string]string{"key1":"value1", "key2":"value2"}),
-			testAccStepReadIamTags(t, "test", map[string]string{"key1":"value1", "key2":"value2"}),
+			testAccStepWriteIamTags(t, "test", map[string]string{"key1": "value1", "key2": "value2"}),
+			testAccStepReadIamTags(t, "test", map[string]string{"key1": "value1", "key2": "value2"}),
 			testAccStepDeletePolicy(t, "test"),
 			testAccStepReadIamTags(t, "test", map[string]string{}),
 		},
@@ -1473,7 +1473,7 @@ func testAccStepReadIamTags(t *testing.T, name string, tags map[string]string) l
 					return nil
 				}
 
-				return fmt.Errorf("bad: %#v", resp)
+				return fmt.Errorf("vault response not received")
 			}
 
 			expected := map[string]interface{}{
--- a/builtin/logical/aws/backend_test.go
+++ b/builtin/logical/aws/backend_test.go
@@ -1437,6 +1437,66 @@ func testAccStepReadIamGroups(t *testing.T, name string, groups []string) logica
 	}
 }
 
+func TestBackend_iamTagsCrud(t *testing.T) {
+	t.Parallel()
+	logicaltest.Test(t, logicaltest.TestCase{
+		AcceptanceTest: true,
+		LogicalBackend: getBackend(t),
+		Steps: []logicaltest.TestStep{
+			testAccStepConfig(t),
+			testAccStepWriteIamTags(t, "test", map[string]string{"key1":"value1", "key2":"value2"}),
+			testAccStepReadIamTags(t, "test", map[string]string{"key1":"value1", "key2":"value2"}),
+			testAccStepDeletePolicy(t, "test"),
+			testAccStepReadIamTags(t, "test", map[string]string{}),
+		},
+	})
+}
+
+func testAccStepWriteIamTags(t *testing.T, name string, tags map[string]string) logicaltest.TestStep {
+	return logicaltest.TestStep{
+		Operation: logical.UpdateOperation,
+		Path:      "roles/" + name,
+		Data: map[string]interface{}{
+			"credential_type": iamUserCred,
+			"iam_tags":        tags,
+		},
+	}
+}
+
+func testAccStepReadIamTags(t *testing.T, name string, tags map[string]string) logicaltest.TestStep {
+	return logicaltest.TestStep{
+		Operation: logical.ReadOperation,
+		Path:      "roles/" + name,
+		Check: func(resp *logical.Response) error {
+			if resp == nil {
+				if len(tags) == 0 {
+					return nil
+				}
+
+				return fmt.Errorf("bad: %#v", resp)
+			}
+
+			expected := map[string]interface{}{
+				"policy_arns":              []string(nil),
+				"role_arns":                []string(nil),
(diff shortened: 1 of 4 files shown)


builtin/logical/aws/backend_test.go
```diff
@@ -1437,6 +1437,66 @@ func testAccStepReadIamGroups(t *testing.T, name string, groups []string) logica
...
+	return logicaltest.TestStep{
+		Operation: logical.ReadOperation,
+		Path:      "roles/" + name,
+		Check: func(resp *logical.Response) error {
+			if resp == nil {
+				if len(tags) == 0 {
+					return nil
+				}
+
+				return fmt.Errorf("bad: %#v", resp)
```
kalafut, line 1476: Can this provide a little more information, e.g. that we expected a non-nil response? Currently it will always just be (I think): `bad: nil`

builtin/logical/aws/path_roles.go
```diff
@@ -301,6 +312,11 @@ func (b *backend) pathRolesWrite(ctx context.Context, req *logical.Request, d *f
 		roleEntry.IAMGroups = iamGroups.([]string)
 	}
 
+	if iamTags, ok := d.GetOk("iam_tags"); ok {
+
```
kalafut, line 316:
```suggestion
```

builtin/logical/aws/secret_access_keys.go
```diff
@@ -3,6 +3,9 @@ package aws
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/vault/sdk/framework"
```
kalafut, line 6: This `goimports` rearrangement looks wrong

builtin/logical/aws/secret_access_keys.go
```diff
@@ -286,6 +287,23 @@ func (b *backend) secretAccessKeysCreate(
 		}
 	}
 
+	if len(role.IAMTags) > 0 {
+		var tags []*iam.Tag
+		for key, value := range role.IAMTags {
+			k, v := key, value
```
kalafut, line 293: Might want to add a comment on why this is necessary, since it's a go-ism that isn't super comment, but is subtle and tricky to debug if you forget to do it.

builtin/logical/aws/backend_test.go
Valarissa: Sorry, it was some copy-pasta from the `IAMGroups` tests above it. I'll see what I can do.

builtin/logical/aws/secret_access_keys.go
Valarissa: Resolved

builtin/logical/aws/secret_access_keys.go
Valarissa: Commented

builtin/logical/aws/path_roles.go
Valarissa: Deleted

builtin/logical/aws/backend_test.go
Valarissa: Informed

builtin/logical/aws/backend_test.go
```diff
@@ -1437,6 +1437,66 @@ func testAccStepReadIamGroups(t *testing.T, name string, groups []string) logica
 	}
 }
 
+func TestBackend_iamTagsCrud(t *testing.T) {
+	t.Parallel()
```
pcman312, line 1441: Generally you should avoid using `t.Parallel` as it doesn't provide much value and often confuses the test output.

builtin/logical/aws/path_roles.go
```diff
@@ -301,6 +312,10 @@ func (b *backend) pathRolesWrite(ctx context.Context, req *logical.Request, d *f
 		roleEntry.IAMGroups = iamGroups.([]string)
 	}
 
+	if iamTags, ok := d.GetOk("iam_tags"); ok {
+		roleEntry.IAMTags = iamTags.(map[string]string)
```
pcman312, line 316: Will this panic if you pass in a JSON blob with non-string values?
```
{
...
  "iam_tags": {
    "tag1": 42
  }
}
```

builtin/logical/aws/secret_access_keys.go
```diff
@@ -286,6 +288,26 @@ func (b *backend) secretAccessKeysCreate(
 		}
 	}
 
+	if len(role.IAMTags) > 0 {
```
pcman312, line 291: nit: this length check is sort of redundant. You could change this to:
```go
var tags []*iam.Tag
for key, value := range role.IAMTags {
	// This assignment needs to be done in order to create unique addresses for
	// these variables. Without doing so, all the tags will be copies of the last
	// tag listed in the role.
	k, v := key, value
	tags = append(tags, &iam.Tag{Key: &k, Value: &v})
}

if len(tags) > 0 {
	_, err = iamClient.TagUser(&iam.TagUserInput{
		Tags:     tags,
		UserName: &username,
	})

	if err != nil {
		return logical.ErrorResponse("Error adding tags to user: %s", err), awsutil.CheckAWSError(err)
	}
}
```

If the `iamClient.TagUser` call can accept nil/empty `Tags` then that length check is also redundant.

website/content/api-docs/secret/aws.mdx
```diff
@@ -329,6 +335,22 @@ Using groups:
 }
 ```
 
+Using tags:
```
pcman312, line 338: nit: this could be improved with tabs: https://github.com/hashicorp/vault/tree/master/website#tabs

builtin/logical/aws/backend_test.go
Valarissa: Damnable copypasta

builtin/logical/aws/path_roles.go
Valarissa: This approach works fine. It's apparent to me now that you can format a curl response in a number of ways, I may update the docs as part of this realization.

builtin/logical/aws/secret_access_keys.go
Valarissa: Addressed

builtin/logical/aws/backend_test.go
Valarissa: Removed

website/content/api-docs/secret/aws.mdx
Valarissa: This should have been addressed

pcman312 (approved):

PR #10955, opened 2021-02-20
expiration: Add a few metrics to measure revoke queue lengths

briankassouf:


swayne275 (approved):
Cool! I've got one coming up for per-queue length (got slowed down a bit trying to figure out how to test the metrics but maybe that isn't needed)

helper/fairshare/jobmanager.go
```diff
@@ -110,6 +112,9 @@ func (j *JobManager) AddJob(job Job, queueID string) {
 	}
 
 	j.queues[queueID].PushBack(job)
+	j.totalJobs++
+	metrics.AddSampleWithLabels([]string{j.name, "job_manager", "queue_length"}, float32(j.queues[queueID].Len()), []metrics.Label{{"queue_id", queueID}})
+	metrics.SetGauge([]string{j.name, "job_manager", "total_jobs"}, float32(j.totalJobs))
```
mgritter, line 117: I think including the name first is undesirable.  job_manager.queue_length.{name} would be better, or use a label. 

Also, sorry to bring this up again, but I'd really like it if new metrics we add have cluster labels, as this activity can bounce around different hosts.  (I think that perf standbys are still going to just remove from cache directly, not use a worker pool?)

briankassouf: You're correct about standbys not using the worker pool, I'll switch this to adding the cluster label. I however think the name being first is preferable so we can keep all the expiration related metrics under the same prefix: `expire.*`.

mgritter: So, each instance would be documented separately in the telemetry guide?

mgritter: (And maybe we need some documentation, now that I think of it.)

swayne275 (approved):

PR #10957, opened 2021-02-20
add per mount queue length metric

swayne275:
First time doing a metric, so hopefully this is right!

Testing to verify the metric works was... more difficult than expected (so I'm leaving that part out)

### DIFF
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -245,7 +245,14 @@ func expireLeaseStrategyFairsharing(ctx context.Context, m *ExpirationManager, l
 		return
 	}
 
-	m.jobManager.AddJob(job, mountAccessor)
+	queueSize := m.jobManager.AddJob(job, mountAccessor)
+	metrics.SetGaugeWithLabels(
+		[]string{"expire", "lease_expiration", "pending"},
+		float32(queueSize),
+		[]metrics.Label{
+			{"mount_point", mountAccessor},
+		},
+	)
 }
 
 func revokeExponentialBackoff(attempt uint8) time.Duration {
--- a/helper/fairshare/jobmanager.go
+++ b/helper/fairshare/jobmanager.go
@@ -94,7 +94,9 @@ func (j *JobManager) Stop() {
 }
 
 // AddJob adds a job to the given queue, creating the queue if it doesn't exist
-func (j *JobManager) AddJob(job Job, queueID string) {
+// it returns the length of the queue associated with `queueID` after the job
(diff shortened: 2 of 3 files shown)


vault/expiration.go
```diff
@@ -245,7 +245,14 @@ func expireLeaseStrategyFairsharing(ctx context.Context, m *ExpirationManager, l
 		return
 	}
 
-	m.jobManager.AddJob(job, mountAccessor)
+	queueSize := m.jobManager.AddJob(job, mountAccessor)
+	metrics.SetGaugeWithLabels(
+		[]string{"expire", "lease_expiration", "pending"},
+		float32(queueSize),
+		[]metrics.Label{
+			{"mount_point", mountAccessor},
```
ncabatoff, line 253: Can we use mountpoint instead of mountAccessor here?  We have a bunch of other metrics that use mountpoint and none that use mountAccessor - plus users aren't likely to know what any mount accessors correspond to.

vault/expiration.go
```diff
@@ -245,7 +245,14 @@ func expireLeaseStrategyFairsharing(ctx context.Context, m *ExpirationManager, l
 		return
 	}
 
-	m.jobManager.AddJob(job, mountAccessor)
+	queueSize := m.jobManager.AddJob(job, mountAccessor)
```
ncabatoff, line 248: Should we also update the gauge when an entry is removed from a queue?  Otherwise if nothing new gets enqueued we'll continue to report stale queue sizes.

vault/expiration.go
swayne275: Can you remind me what the differences are? If i'm remembering right, mount accessor is more in line with where the data physically lives, and mount point can be user defined (and i think can be not 1:1 related to a mount accessor).

The goal of this is to see which backends might be causing issues

vault/expiration.go
swayne275: sure

vault/expiration.go
ncabatoff: Mount accessor is a random UUID that doesn't change when you remount a backend on a new path (I don't recall if it's the same UUID that's used as a storage prefix).  Mountpoint is how the user thinks of the mount, i.e. the HTTP path prefix for requests that should go to that backend.

swayne275: can multiple mount points be associated with the same mount accessor?

ncabatoff: No

Issue #10852, opened 2021-02-08
Lowercase conversion is causing inconsistent behaviour in PKI backend


gnugnug:
**Describe the bug**
Our users have a habit of writing hostnames in uppercase (eg MYSERVER.INT). When issuing certificates through the PKI backend of Vault this has odd outcomes, because Vault converts the subject alternative names to lowercase via `ParseDedupLowercaseAndSortStrings()`, but leaves the common name as it is. After that both the subject alternative names and the common name are sent through `validateNames()`, which performs a case-senstive comparison.

That means the SANs are converted to lowercase before the case-senstive validation takes place and the common name is converted to lowercase after this validation. In the end certificate requests are allowed or denied depending on the order of the DNS names in the certificate, which is confusing to the users.

**To Reproduce**
Preparations:
```
>vault version
Vault v1.6.2
>vault secrets enable pki
Success! Enabled the pki secrets engine at: pki/
>vault write pki/root/generate/internal common_name=my-root
certificate      -----BEGIN CERTIFICATE-----
[...]
>vault write pki/roles/example allowed_domains="example.net, EXAMPLE.COM" allow_bare_domains=true
Success! Data written to: pki/roles/example
```
Request a certificate:
```
>vault write pki/issue/example common_name="EXAMPLE.COM" alt_names="example.net" ttl=1
[Success]
```
Request the same certificate with DNS names in reversed order:
```
>vault write pki/issue/example common_name="example.net" alt_names="EXAMPLE.COM" ttl=1
Error writing data to pki/issue/example: Error making API request.
* subject alternate name example.com not allowed by this role
```
One could argue that allowed_domains should not contain upperspace characters, but this results in other inconsistent behaviour:
```
>vault write pki/roles/example allowed_domains="example.net, example.com" allow_bare_domains=true
Success! Data written to: pki/roles/example
```
Certificate request for two DNS names succeeds:
```
>vault write pki/issue/example common_name="example.com" alt_names="EXAMPLE.NET" ttl=1
[Success]
```
If we remove example.com it suddenly fails:
```
>vault write pki/issue/example common_name="EXAMPLE.NET" ttl=1
Error writing data to pki/issue/example: Error making API request.
* common name EXAMPLE.NET not allowed by this role
```

**Expected behavior**
[...]

PR #10959, opened 2021-02-20
Make cert domain name validation case insensitive

livyfu:
This aims to fix https://github.com/hashicorp/vault/issues/10852


thaniri (approved):

builtin/logical/pki/cert_util.go
```diff
@@ -180,8 +180,8 @@ func fetchCertBySerial(ctx context.Context, req *logical.Request, prefix, serial
 // If one does not pass, it is returned in the string argument.
 func validateNames(b *backend, data *inputBundle, names []string) string {
 	for _, name := range names {
-		sanitizedName := name
-		emailDomain := name
+		sanitizedName := strings.ToLower(name)
```
HridoyRoy, line 183: Hello, 
It seems like ToLower doesn't account for non-latin characters, and IDNs can have these characters. Can we instead use EqualFold (https://golang.org/pkg/strings/#EqualFold) in the string comparison? 
Thanks!

HridoyRoy (changes requested):

HridoyRoy (approved):

PR #10976, opened 2021-02-22
UI: Small tickets/ bug fixes for DBSE 

Monkeychip:
A couple of small fixes to the DBSE engine:

1. Add link to roles from Connection view. 

2. Add link to connection from role view.

3. Handle errors in generate credentials better. *note a 400 error will always occur when navigating to this page because it checks for both static and dynamic role, one will have to return a 400.

**500 error:** happens in this case when the user created a dynamic role without a creation statement.

**403 error:**  when you don't have permission or you are not authorized.  It's hard for us to tell when they don't have list permissions so we insert a more generic error.

**generic error (all other):**

4. add conditional that was causing issue on view role. 
5. remove the word "yet" from empty state message.  Confirmed okay with design.

### DIFF
--- a/ui/app/styles/components/empty-state.scss
+++ b/ui/app/styles/components/empty-state.scss
@@ -13,6 +13,7 @@
 }
 
 .empty-state-title {
+  white-space: nowrap;
   color: $grey;
   font-size: $size-4;
   font-weight: $font-weight-semibold;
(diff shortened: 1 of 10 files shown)


ui/app/styles/components/empty-state.scss
```diff
@@ -13,6 +13,7 @@
 }
 
 .empty-state-title {
+  white-space: nowrap;
```
Monkeychip, line 16: per design's request

chelshaw (approved):
Nice work! ✨

PR #10982, opened 2021-02-23
UI: Add the wizard to the database secret engine

Monkeychip:
Adds the Wizard designs to the DBSE engine.
_*Note: confirmed with designs that the save buttons when creating a database and role should be "Create database" and "Create role" respectively.  I spotted this in the difference between designs for the wizard vs. the ones for creating a role._

See screenshots (the copy is based on designs):

1. Enable: 

2. Connect Database:

3. Create Connection:

4. Create a role:

5. Save role (not in screenshot save is shown, will fix to show Create Role)

6. All set / the end.


ui/app/helpers/wizard-constants.js
```diff
@@ -42,6 +42,7 @@ export const DEFAULTS = {
 
 export const FEATURE_MACHINE_STEPS = {
   secrets: {
+    connection: 4, // ARG TODO unsure just making this up for now
```
chelshaw, line 45: ??

ui/app/templates/components/wizard-progress.hbs
```diff
@@ -1,5 +1,6 @@
 <div class="progress-container">
   {{#each progressBar as |bar|}}
+  {{log bar "bar ARG TODO COME BACK TO ME"}}
```
chelshaw, line 3: remove

Monkeychip: Oish, thank you.  There is a progress bar issue.  I'll work on this.

ui/app/helpers/wizard-constants.js
Monkeychip: Yep, need to come back to this. Had a note, and this missed it. Thanks!

ui/app/templates/components/wizard-progress.hbs
```diff
@@ -1,5 +1,6 @@
+
 <div class="progress-container">
-  {{#each progressBar as |bar|}}
+  {{#each @progressBar as |bar|}}
```
Monkeychip, line 3: the reason the progress bar wasn't showing, bug after ember upgrade

chelshaw (approved):
good catch on the progress bar

PR #10983, opened 2021-02-23
update sdk version to 1.6.3

swayne275:


briankassouf (approved):

raskchanky (approved):

swayne275:
failed race tests look like a result of not having ported VAULT-899 and similar to this release chain (those fix the activity log race tests)

PR #10986, opened 2021-02-23
agent: add caching encryption package

jasonodonnell:
This is a small package that AES encryption/decryption to be used in persistent caching. Currently this only implements a K8s specific use case but will follow up with response wrapping.

The configs to enable this will be added in a different PR relevant to caching persistence.

### DIFF
--- a/command/agent/cache/crypto/k8s.go
+++ b/command/agent/cache/crypto/k8s.go
@@ -58,8 +58,8 @@ func (k *KubeEncryptionKey) GetKey() []byte {
 
 // GetPersistentKey returns the key which should be stored in the persisent
 // cache. In k8s we store the key as is, so just return the key stored.
-func (k *KubeEncryptionKey) GetPersistentKey() []byte {
-	return k.wrapper.GetKeyBytes()
+func (k *KubeEncryptionKey) GetPersistentKey() ([]byte, error) {
+	return k.wrapper.GetKeyBytes(), nil
 }
 
 // Renewable lets the caller know if this encryption key type is
--- a/command/agent/cache/crypto/k8s.go
+++ b/command/agent/cache/crypto/k8s.go
@@ -1,0 +1,91 @@
+package crypto
+
+import (
+	"context"
+	"crypto/rand"
+	"fmt"
+
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+	"github.com/hashicorp/go-kms-wrapping/wrappers/aead"
+)
+
+var _ KeyManager = (*KubeEncryptionKey)(nil)
+
+// KubeEncryptionKey TODO
+type KubeEncryptionKey struct {
+	renewable bool
+	wrapper   *aead.Wrapper
+}
+
+// NewK8s returns a new instance of the Kube encryption key. Kubernetes
+// encryption keys aren't renewable.
+func NewK8s(existingKey []byte) (*KubeEncryptionKey, error) {
+	k := &KubeEncryptionKey{
+		renewable: false,
+		wrapper:   aead.NewWrapper(nil),
+	}
+
+	k.wrapper.SetConfig(map[string]string{"key_id": KeyID})
+
+	var rootKey []byte = nil
+	if len(existingKey) != 0 {
+		if len(existingKey) != 32 {
+			return k, fmt.Errorf("invalid key size, should be 32, got %d", len(existingKey))
+		}
+		rootKey = existingKey
+	}
+
+	if rootKey == nil {
+		newKey := make([]byte, 32)
+		_, err := rand.Read(newKey)
+		if err != nil {
+			return k, err
+		}
+		rootKey = newKey
+	}
+
+	if err := k.wrapper.SetAESGCMKeyBytes(rootKey); err != nil {
+		return k, err
+	}
+
+	return k, nil
+}
+
+// Get returns the encryption key in a format optimized for storage.
+// In k8s we store the key as is, so just return the key stored.
+func (k *KubeEncryptionKey) Get() []byte {
+	return k.wrapper.GetKeyBytes()
+}
+
+// Renewable lets the caller know if this encryption key type is
+// renewable. In Kubernetes the key isn't renewable.
+func (k *KubeEncryptionKey) Renewable() bool {
+	return k.renewable
+}
+
+// Renewer is used when the encryption key type is renewable. Since Kubernetes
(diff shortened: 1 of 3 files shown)


command/agent/cache/crypto/k8s_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	key := k8sKey.GetKey()
+	if key == nil {
+		t.Fatalf(fmt.Sprintf("key is nil, it shouldn't be: %s", key))
+	}
+
+	persistentKey := k8sKey.GetPersistentKey()
+	if persistentKey == nil {
+		t.Fatalf(fmt.Sprintf("key is nil, it shouldn't be: %s", persistentKey))
+	}
```
tvoran: Could check if key == persistentKey here as well, since they should be the same?

command/agent/cache/crypto/k8s_test.go
```diff
@@ -0,0 +1,142 @@
...
+	if string(persistentKey) != string(rootKey) {
+		t.Fatalf(fmt.Sprintf("expected keys to be the same, they weren't: expected: %s, got: %s", rootKey, key))
+	}
```
tvoran, line 73: Could check that key == rootKey here too.

command/agent/cache/crypto/k8s_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+func TestCrypto_KubernetesPassGeneratedKey(t *testing.T) {
+	k8sFirstKey, err := NewK8s([]byte{})
+	if err != nil {
+		t.Fatalf(fmt.Sprintf("unexpected error: %s", err))
+	}
+
+	firstPersistentKey := k8sFirstKey.GetKey()
+	if firstPersistentKey == nil {
+		t.Fatalf(fmt.Sprintf("key is nil, it shouldn't be: %s", firstPersistentKey))
```
tvoran, line 105: firstPersistentKey will be nil here, so it probably doesn't need to be printed.

command/agent/cache/crypto/k8s_test.go
```diff
@@ -0,0 +1,142 @@
...
+	loadedKey, _ := k8sLoadedKey.GetPersistentKey()
+	if loadedKey == nil {
+		t.Fatalf(fmt.Sprintf("key is nil, it shouldn't be: %s", loadedKey))
+	}
+
+	if string(loadedKey) != string(firstPersistentKey) {
+		t.Fatalf(fmt.Sprintf("keys do not match"))
+	}
+
+	plaintext, err := k8sFirstKey.Decrypt(nil, ciphertext, aad)
```
tvoran, line 134: Should k8sLoadedKey be decrypting here?

tvoran (approved):
Just some comments in the tests.

command/agent/cache/crypto/k8s_test.go
jasonodonnell: Done!

command/agent/cache/crypto/k8s_test.go
jasonodonnell: Done!

command/agent/cache/crypto/k8s_test.go
jasonodonnell: Good catch, fixed.

command/agent/cache/crypto/k8s_test.go
jasonodonnell: Good catch! Fixed.

command/agent/cache/crypto/k8s.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if rootKey == nil {
+		newKey := make([]byte, 32)
+		_, err := rand.Read(newKey)
+		if err != nil {
+			return k, err
+		}
+		rootKey = newKey
+	}
```
calvn, line 44: Consider turning this into a switch case:
```
switch len(exisistingKey) {
	case 0:
	// generates key
	case 32:
	// set to root key
	default:
	// error case
}
```

command/agent/cache/crypto/k8s.go
```diff
@@ -0,0 +1,97 @@
...
+	"crypto/rand"
+	"fmt"
+
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+	"github.com/hashicorp/go-kms-wrapping/wrappers/aead"
+)
+
+var _ KeyManager = (*KubeEncryptionKey)(nil)
+
+type KubeEncryptionKey struct {
```
calvn, line 14: Is this really a k8s encryption key, or is there an alternate name for this?

command/agent/cache/crypto/k8s.go
```diff
@@ -0,0 +1,97 @@
...
+	renewable bool
+	wrapper   *aead.Wrapper
+}
+
+// NewK8s returns a new instance of the Kube encryption key. Kubernetes
+// encryption keys aren't renewable.
+func NewK8s(existingKey []byte) (*KubeEncryptionKey, error) {
```
calvn, line 21: Nit: The constructor's naming should match the returned struct name (i.e. `NewKubeEncryptionKey` in this case).

command/agent/cache/crypto/k8s.go
```diff
@@ -0,0 +1,97 @@
...
+	k := &KubeEncryptionKey{
+		renewable: false,
+		wrapper:   aead.NewWrapper(nil),
+	}
+
+	k.wrapper.SetConfig(map[string]string{"key_id": KeyID})
+
+	var rootKey []byte = nil
```
calvn, line 29:
```suggestion
	var rootKey []byte
```

command/agent/cache/crypto/k8s.go
```diff
@@ -0,0 +1,97 @@
...
+	return k.wrapper.GetKeyBytes()
+}
+
+// GetPersistentKey returns the key which should be stored in the persisent
+// cache. In k8s we store the key as is, so just return the key stored.
+func (k *KubeEncryptionKey) GetPersistentKey() ([]byte, error) {
+	return k.wrapper.GetKeyBytes(), nil
+}
+
+// Renewable lets the caller know if this encryption key type is
```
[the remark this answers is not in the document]
jasonodonnell: I'm using `Renewable` in the interface to indicate some implementations will have a Renewer which you'll need to use (response wrapping will rewrap the token based on TTL).

PR #10988, opened 2021-02-24
Bool pointer flag

vishalnayak:
Lets you know if a boolean flag was set or not.

### DIFF
--- a/command/base_flags.go
+++ b/command/base_flags.go
@@ -30,6 +30,81 @@ type FlagBool interface {
 	IsBoolFlag() bool
 }
 
+// BoolPtr is a bool which is aware if it has been set.
+type BoolPtr struct {
+	v *bool
+}
+
+func (b *BoolPtr) Set(v string) error {
+	if b.v == nil {
+		b.v = new(bool)
+	}
+	var err error
+	*(b.v), err = strconv.ParseBool(v)
+	return err
+}
+
+func (b *BoolPtr) IsSet() bool {
+	return b.v != nil
+}
+
+func (b *BoolPtr) Get() bool {
+	if b.v == nil {
+		return false
+	}
+	return *b.v
+}
+
+func (b *BoolPtr) String() string {
+	var current bool
+	if b.v != nil {
+		current = *(b.v)
+	}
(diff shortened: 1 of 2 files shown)


command/base_flags.go
```diff
@@ -30,6 +30,86 @@ type FlagBool interface {
...
+type BoolPtr struct {
+	v *bool
+}
+
+func (b *BoolPtr) Set(v string) error {
+	if b.v == nil {
+		b.v = new(bool)
+	}
+	var err error
+	*(b.v), err = strconv.ParseBool(v)
```
ncabatoff, line 43: When there's a possible error It's usually preferable to use a temporary value rather than assigning directly to the target.  If ParseBool returns an error we'll be overwriting b.v, possibly setting a default-true BoolPtr to be default-false.

command/base_flags.go
```diff
@@ -30,6 +30,86 @@ type FlagBool interface {
...
+	return b.target.Set(s)
+}
+
+func (b *boolPtrValue) String() string { return b.target.String() }
+
+type BoolPtrVar struct {
+	Name       string
+	Aliases    []string
+	Usage      string
+	Hidden     bool
```
ncabatoff, line 97: If you're going to define a Hidden field you may as well implement FlagVisibility.

ncabatoff (approved):
Seems fine, though you might want to add EnvVar for consistency with the others.  Also should there be a new FlagSet method `BoolPtrVar`?

command/base_flags.go
vishalnayak: Hidden and EnvVar is now supported. 

> Also should there be a new FlagSet method BoolPtrVar?
Yep. Its already there.

command/base_flags.go
vishalnayak: Yep. Fixed.

command/base_flags.go
ncabatoff: Haha my early morning reading skills aren't the best it seems.

ncabatoff (approved):
LGTM!

PR #10993, opened 2021-02-24
UI Add test and clean up for DBSE

Monkeychip:
1.  Added an integration test to handle new situations for the `info-table-row` component.  Specifically, around arrays and links.
2. Removed no-wrap as it causes more problems, and after testing doesn't appear to be needed anymore.

### DIFF
--- a/ui/lib/core/addon/templates/components/info-table-row.hbs
+++ b/ui/lib/core/addon/templates/components/info-table-row.hbs
@@ -42,8 +42,6 @@
           @viewAll={{viewAll}}
           @wildcardLabel={{wildcardLabel}}
         />
-      {{else if isLink}}
-        <LinkTo @route="vault.cluster.secrets.backend.show" @model={{value.[0]}} >{{value}}</LinkTo>
       {{else}}
         {{#if tooltipText}}
           <ToolTip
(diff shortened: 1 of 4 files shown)


ui/lib/core/addon/templates/components/info-table-row.hbs
```diff
@@ -42,8 +42,6 @@
           @viewAll={{viewAll}}
           @wildcardLabel={{wildcardLabel}}
         />
-      {{else if isLink}}
-        <LinkTo @route="vault.cluster.secrets.backend.show" @model={{value.[0]}} >{{value}}</LinkTo>
       {{else}}
```
Monkeychip, line 45: This is not used, instead because the value comes in as an array, it's handled by the `infoTableItemArray` .  (e.g. for connections and roles in the show view).

ui/tests/integration/components/info-table-row-test.js
```diff
@@ -0,0 +1,73 @@
+import { module, test } from 'qunit';
```
chelshaw, line 1: 👏

chelshaw (approved):
Nice!

PR #10998, opened 2021-02-25
agent/cache: KeyManager refactor

calvn:
Refactors the `KeyManager` interface to simply return a `wrapping.Wrapper` so that key operations can be handed off to a sub-subsystem such as the persistent storage layer. It also provides a way to get a retrieval token which can be loaded back to an instance of the KeyManager. For the case of `PassthroughKeyManager`, this is the encryption key itself.


command/agent/cache/cacheboltdb/bolt.go
```diff
@@ -242,7 +245,12 @@ func (b *BoltStorage) GetAutoAuthToken(ctx context.Context) ([]byte, error) {
 		return nil, nil
 	}
 
-	plainText, err := b.keyMgr.Decrypt(ctx, token, []byte(b.aad))
+	var blob wrapping.EncryptedBlobInfo
+	if err := proto.Unmarshal(token, &blob); err != nil {
+		return nil, err
+	}
+
+	plainText, err := b.wrapper.Decrypt(ctx, &blob, []byte(b.aad))
```
calvn, line 253: Nit: we should rename this to `plaintext` (same with `ciphertext` where applicable). We can do this on the main branch after merging of this PR to avoid merge conflicts.

command/agent/cache/cacheboltdb/bolt.go
```diff
@@ -207,7 +205,12 @@ func (b *BoltStorage) GetByType(ctx context.Context, indexType string) ([][]byte
 			return fmt.Errorf("bucket %q not found", b.rootBucket)
 		}
 		top.Bucket([]byte(indexType)).ForEach(func(id, cipherText []byte) error {
-			plainText, err := b.keyMgr.Decrypt(ctx, cipherText, []byte(b.aad))
+			var blob wrapping.EncryptedBlobInfo
+			if err := proto.Unmarshal(cipherText, &blob); err != nil {
+				return err
+			}
+
+			plainText, err := b.wrapper.Decrypt(ctx, &blob, []byte(b.aad))
```
tvoran, line 213: Looks like there's two cases of this unmarshal and decrypt block, so it may be worth moving into it's own function?

command/agent.go
```diff
@@ -507,25 +508,40 @@ func (c *AgentCommand) Run(args []string) int {
...
 				})
 				if err != nil {
 					c.UI.Error(fmt.Sprintf("Error opening persistent cache: %v", err))
 					return 1
 				}
+
 				// Get the token from bolt for retrieving the encryption key,
 				// then setup encryption so that restore is possible
-				token, err := ps.GetRetrievalToken()
+				key, err := ps.GetRetrievalToken()
```
tvoran, line 519: Just noticing that we we're using `key` here with GetRetrievalToken(), but using `token` down on line 610 with StoreRetrievalToken(): https://github.com/hashicorp/vault/blob/673cf14ac7ce30ea5543403d9d6119b6ac2f385f/command/agent.go#L609-L618
We may want to use the same term both places?

tvoran (approved):
Minor comments, but looks good.

Issue #5304, opened 2018-09-07
bindfilter, i.e. expand userattr to a full LDAP filter


ghenry:
**Is your feature request related to a problem? Please describe.**

At SureVoIP we do an initial LDAP search with an ldap entity that only has limited permissions to search to find a valid user DN and then re-bind with the supplied password. Your LDAP config almost allows this. Instead of just passing (userattr=uid) as a filter, it would be ideal for that to be supplied as a full LDAP filter, like (uid=%s) etc. where you replace with sprintf. For example from one of our Apps and used in all our services (user_filter):

          <store>
                class LDAP
                ldap_server xxx
                <ldap_server_options>
                        timeout 30
                </ldap_server_options>
                binddn xxx
                bindpw xxx
                start_tls 1
                user_basedn xxx
                user_filter (&(objectClass=person)(uid=%s)(SuretecAccountActive=TRUE)(SuretecServiceActive=TRUE)(SuretecServiceEnabled=SUREVOIPPORTAL))
                user_scope one
                user_field xxx
                <user_search_options>
                    deref always
                </user_search_options>
            </store>

This means we can toggle any things in the filter to FALSE and kick out a user or let them into a service.
 
**Describe the solution you'd like**
rename userattr to userfilter in config and update code here:

https://github.com/hashicorp/vault/blob/b65961e3b695105192b1b518b159d6c59eef0221/helper/ldaputil/config.go#L70
https://github.com/hashicorp/vault/blob/b65961e3b695105192b1b518b159d6c59eef0221/helper/ldaputil/client.go#L116

It's almost there, you just limit the filter to (userattr=username)

**Describe alternatives you've considered**
Using groups, but I don't want to let the user get that far.

**Explain any additional use-cases**
A complete LDAP solution like is possible with almost all LDAP auth options in applications.

**Additional context**
None. Happy to do a patch, test and doc update.

Thanks!

Lasering:
I would also find this change very helpful. The `userattr` should be changed to `userfilter` with the default value of `(cn=%s)` and this line to:

`filter := fmt.Sprintf("(%s=%s)", cfg.UserAttr, ldap.EscapeFilter(username))`

Changed to:

`filter := fmt.Sprintf(cfg.UserFilter, ldap.EscapeFilter(username))`

This way the LDAP authentication would be way more powerfull. I can make a PR if it helps.

Lasering:
Can I help making progress in this issue?

mnsh:
[...]

PR #11000, opened 2021-02-25
Support for a LDAP user search filter. Documentation, tests and UI included

ixe013:
This pull request adds support for a LDAP user search filter, requested in #5304. Unit tests, documentation and UI code included. 

Using a user search filter gives more control over who should even be allowed to authenticate to Vault. You can be liberal and accept more than one username (usefull when migrating ldap usernames after a merger, for example) or more strict and prevent contractors or interns from logging in.

The change is backward compatible with existing ldap configurations. If `upndomain` is set, it takes precedence over the user filter. If it is not set, the filter defaults to the current (as of 1.6.2) `({{.UserAttr}}={{.Username}})` filter.

If the filter is too large and more than one entry is returned, access is denied with an specific error message.

Link with the Identity backend is unchanged regardless of user filter. Ultimately, you will bind to a single distinguished name (DN). The value of the attribute identified by `userattr` in the configuration will still be used to create the entity alias, regardless of how the filter got you there. It allows to add a custom search filter to an existing configuration without having to recreate entities (which would loose audit tracability before-after the change).

Roles and groups are not impacted by this change.


ixe013:
Rebased to 1.7.0

ixe013:
Just keeping the PR alive with a rebase...

Issue #10918, opened 2021-02-12
Vault agent re-authenticates on Vault server outages


andrejvanderzee:
We are on Vault 1.6.2 and are experiencing problems with vault agent in production environments. When Vault is down for longer than 1 minute, it seems that vault agent re-authenticates and eventually revokes the old token and secrets. Preferably, vault agent would at least retry renewing the lease of the vault token until its TTL expires, instead of restating the lifetime watcher and consul template manager.

It seems that this logic could be tuned so that vault agent can cope with Vault outages longer than one minute, without revoking token and secrets:
https://github.com/hashicorp/vault/blob/master/api/lifetime_watcher.go#L263

We have tried VAULT_MAX_RETRIES but it bails out after 1 minute.

ncabatoff:
I think the problematic behaviour is here:

```
		// We keep evaluating a new grace period so long as the lease is
		// extending. Once it stops extending, we've hit the max and need to
		// rely on the grace duration.
		if leaseDuration > priorDuration {
			r.calculateGrace(leaseDuration)
		}
```

When we get there due to a renew failure, this condition will be false, so we don't recalculate grace.  Grace remains what it was at the outset, something like 10-20% of the original lease TTL.  So let's say the auth method TTL is configured at 3m, as it looks to be the case in your agent example in #10898.  That would make the grace be something like 30s, and the sleep duration a bit over 2m.

We immediately attempt a renew when the LifetimeWatcher starts running, then sleep for about 2m.  We try to renew then, and suppose we fail.  So now leaseDuration is about 1m (time remaining), grace remains 30s, and sleep duration is 

```
		// The sleep duration is set to 2/3 of the current lease duration plus
		// 1/3 of the current grace period, which adds jitter.
		sleepDuration := time.Duration(float64(leaseDuration.Nanoseconds())*2/3 + float64(r.grace.Nanoseconds())/3)
```

so about 40s + 10s.  This means that we will indeed return nil here as you saw, because leaseDuration-sleepDuration is about 10s, smaller than grace of 30s:

```
		if leaseDuration <= r.grace || leaseDuration-sleepDuration <= r.grace {
			return nil
		}
```

[...]

PR #11008, opened 2021-02-25
Let lifetime watcher retry until end of lease

andrejvanderzee:
This pull request retries renewals on error until the end of the lease (minus grace), instead of abruptly restarting the lifetime watcher (and revoking tokens/secrets as a result). 

The problem is described in detail here: https://github.com/hashicorp/vault/issues/10918.

@ncabatoff

### DIFF
--- a/api/lifetime_watcher.go
+++ b/api/lifetime_watcher.go
@@ -270,6 +270,7 @@ func (r *LifetimeWatcher) doRenew() error {
 
 		var leaseDuration time.Duration
 		fallbackLeaseDuration := initialTime.Add(priorDuration).Sub(time.Now())
+		var retryOnError bool
 
 		switch {
 		case nonRenewable || r.renewBehavior == RenewBehaviorRenewDisabled:
@@ -290,7 +291,9 @@ func (r *LifetimeWatcher) doRenew() error {
 					}
 				}
 
-				leaseDuration = fallbackLeaseDuration
+				// Calculate remaining duration until inital token lease expires
+				leaseDuration = initialTime.Add(time.Duration(initLeaseDuration) * time.Second).Sub(time.Now())
+				retryOnError = true
 				break
 			}
 
@@ -300,6 +303,9 @@ func (r *LifetimeWatcher) doRenew() error {
 			default:
 			}
 
+			// Reset initial time
+			initialTime = time.Now()
+
 			// Possibly error if we are not renewable
 			if ((tokenMode && !renewal.Auth.Renewable) || (!tokenMode && !renewal.Renewable)) &&
 				r.renewBehavior == RenewBehaviorErrorOnErrors {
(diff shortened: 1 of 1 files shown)


andrejvanderzee:
@ncabatoff Any change that someone can look at this? We do have a production problem and our current workaround is a rolling restart of all our k8s production deployments every week, just to make sure we do not run into this vault-agent sidecar edge-case. This MR makes vault agents more resilient against temporary problems not being able to renew leases of tokens and secrets.

andrejvanderzee:
@ncabatoff  @jefferai 

Can someone please take a look at this Vault Agent resilience issue? Our workaround now is doing rolling upgrades of all our Kubernetes deployments when it nears the TTL of the vault token. Insane.

The current implementation of `func (r *LifetimeWatcher) doRenew() error` is just not OK, I hope you can acknowledge that? 

I think the whole `doRenew()` requires a proper refactoring, but this PR at least makes Vault Agent more resilient with a patch on the _error/problem-path_ only, which causes vault token and secrets to be unnecessarily renewed to soon!

andrejvanderzee:
@ncabatoff  @jefferai 

Can someone please take a look at this Vault Agent resilience issue? Our workaround now is doing rolling upgrades of all our Kubernetes deployments when it nears the TTL of the vault token. Insane.

The current implementation of `func (r *LifetimeWatcher) doRenew() error` is just not OK, I hope you can acknowledge that? 

I think the whole `doRenew()` requires a proper refactoring, but this PR at least makes Vault Agent more resilient with a patch on the _error/problem-path_ only, which causes vault token and secrets to be unnecessarily renewed to soon!

andrejvanderzee:
@ncabatoff @jefferai 

Ping...

andrejvanderzee:
@ncabatoff @jefferai

Any chance that this will be picked up?

ncabatoff:
Hey @andrejvanderzee,

I'm sorry to have left you hanging for so long.  I've added a story to the current sprint to tackle this issue.  I think this PR is on the right track, though I would like to see some tests.  Maybe factor out the first bit of doRenew so that we can provide our own artificial renewFunc when testing.  Would you like to work on those, or shall I?

andrejvanderzee:
Hi @ncabatoff, no problem at all, we are just very happy that it is being picked up now, as it occasionally is causing problems in our prd environments. I won't have much time, so if you could add the tests that would be wonderful :-) 

As a side note: To mitigate the problem, we put VAULT_MAX_TRIES on 12. The problem with this approach is that every common error will end up in a "context deadline exceeded", because retrying is handled inside the Vault client and after 12 retries you exceed  VAULT_CLIENT_TIMEOUT. 

Thanks for putting this in your sprint :-)

ncabatoff:
Closing this in favour of #11445.

PR #11011, opened 2021-02-25
UI: fix KMIP bug and test

Monkeychip:
The KMIP secrets had a bug where if you went from one KMIP engine that had a configuration model and then to another KMIP secret engine that did not have a configuration model it would pull from the Ember data store the first KMIP engine's `kmip/configuration` model.  To solve this, we clear the model using a mixin.

The second issue was a test failure because the ports needed to be random when setting up the configuration, this was fixed by creating a new method called `getRandomPort`.

Because Circle CI won't run the enterprise tests, I ran the enterprise test suite three times, and no failures.

### DIFF
--- a/ui/lib/kmip/addon/routes/configuration.js
+++ b/ui/lib/kmip/addon/routes/configuration.js
@@ -1,11 +1,13 @@
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
+import UnloadModel from 'vault/mixins/unload-model-route';
 
-export default Route.extend({
+export default Route.extend(UnloadModel, {
   store: service(),
   secretMountPath: service(),
   pathHelp: service(),
-  beforeModel() {
+  async beforeModel() {
+    await this.store.unloadAll('kmip/config');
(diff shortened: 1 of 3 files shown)


ui/lib/kmip/addon/routes/configuration.js
```diff
@@ -1,11 +1,13 @@
...
+import UnloadModel from 'vault/mixins/unload-model-route';
 
-export default Route.extend({
+export default Route.extend(UnloadModel, {
   store: service(),
   secretMountPath: service(),
   pathHelp: service(),
-  beforeModel() {
+  async beforeModel() {
+    await this.store.unloadAll('kmip/config');
```
chelshaw, line 10: We don't need this line AND extend UnloadModel, right?

chelshaw (commented):
One small fix, then it looks good

ui/lib/kmip/addon/routes/configuration.js
Monkeychip: Ah yes, you're correct.  I'll remove

chelshaw (approved):

PR #11019, opened 2021-02-26
Autopilot server removals map

vishalnayak:


physical/raft/raft_autopilot.go
```diff
@@ -389,21 +389,31 @@ func (d *Delegate) KnownServers() map[raft.ServerID]*autopilot.Server {
 // goroutine.
 func (d *Delegate) RemoveFailedServer(server *autopilot.Server) {
 	go func() {
-		if _, ok := d.deadServerRemovals.Load(server.ID); ok {
+		success := false
+		d.l.Lock()
```
ncabatoff, line 393: Why not add a new mutex to Delegate instead?  It gets harder to reason about a lock the more things it's used for and the more different goroutines use it.

physical/raft/raft_autopilot.go
```diff
@@ -389,21 +389,31 @@ func (d *Delegate) KnownServers() map[raft.ServerID]*autopilot.Server {
...
+		defer func() {
+			if success {
```
ncabatoff, line 395: Why only on success?  The goal here is to avoid proliferation of duplicate goroutines, so as long as we're returning I'd say we can remove from the map.

physical/raft/raft_autopilot.go
ncabatoff: Regardless of whether you use this mutex or another, I'd be inclined to only hold the luck while accessing/mutating the map.  I don't know how raft.RemoveServer works, so it might not make a difference, but I don't see any reason to serialize parallel removal attempts of different servers.

physical/raft/raft_autopilot.go
vishalnayak: We want to do it only on success. Autopilot library might call `RemoveFailedServer` for the same server ID after every reconcile interval until it is actually removed from the raft config. In that case, we want the already initiated removal to be there in the map.

physical/raft/raft_autopilot.go
vishalnayak: With each invocation of this method, we access the map. So we need to hold the lock each time.

This now returns early if the removal for a server is already triggered. If it is not triggered, it will initiate one in parallel and block on that.

physical/raft/raft_autopilot.go
ncabatoff: Say that we get a failure on our call to RemoveServer so we return and leave the entry in inflightRemovals.  Doesn't that mean that on the next reconcile interval, when autopilot asks us again to remove the server, we'll check the map, see there's an entry, and do nothing?

vishalnayak: Ah, right. Fixed.

physical/raft/raft_autopilot.go
ncabatoff: I don't think we need to hold the lock while we're waiting for RemoveServer, but I also don't want to hold up this PR on that point. It only really matters if RemoveServer could get stuck on serverA but work fine on a subsequent call with serverB, which seems unlikely.

ncabatoff (approved):

physical/raft/raft_autopilot.go
vishalnayak: I appreciate the thoroughness. You are not holding up this PR. I want to get this right. Pushed up. Let me know how it looks.

ncabatoff (approved):
Looks good!

PR #11021, opened 2021-02-26
Add extra check for persistent cache when routing templating

jasonodonnell:
We only want to route the templating engine through the local cache if persistence is enabled, so adding another check here for that.

### DIFF
--- a/changelog/10927.txt
+++ b/changelog/10927.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-agent: Route templating server through cache when enabled.
+agent: Route templating server through cache when persistent cahce is enabled.
 ```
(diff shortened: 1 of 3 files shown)


tvoran (approved):

calvn (approved):
Looks good, though you may want to update the CL file on the original PR: https://github.com/hashicorp/vault/pull/10927/files#diff-1f5eeb55c9c46cf1e78b0d40488d02f75e4ff9e785136931f6de6b8cb112ccfc

changelog/10927.txt
```diff
@@ -1,3 +1,3 @@
 ```release-note:improvement
-agent: Route templating server through cache when enabled.
+agent: Route templating server through cache when persistent cahce is enabled.
```
calvn, line 2:
```suggestion
agent: Route templating server through cache when persistent cache is enabled.
```

jasonodonnell:
Slightly altered the change log from #10927.

PR #11022, opened 2021-02-26
Add Partial Month Client Count API for Activity Log

swayne275:


vault/activity_log.go
```diff
@@ -1771,3 +1771,27 @@ func (c *Core) activeEntityGaugeCollector(ctx context.Context) ([]metricsutil.Ga
...
+	entityCount := len(a.activeEntities)
+	var tokenCount int
+	for _, countByNS := range a.currentSegment.tokenCount.CountByNamespaceID {
+		tokenCount += int(countByNS)
+	}
+	clientCount := entityCount + tokenCount
+
+	responseData := make(map[string]interface{})
+	responseData["clients"] = clientCount
+
```
mgritter, line 1795: I think you should report the two components separately, i.e., have "clients" and "active_entities" (or whatever we use in the other API) and "non_entity_tokens".

swayne275: 23836de7ae

mgritter (approved):

PR #11023, opened 2021-02-26
agent: skip restoring leases that are expired

tvoran:
When restoring persisted leases, checks that the elapsed time since the last renewal is less than the lease duration.

Adds new fields to Index: a LastRenewal timestamp, and a Type (used when updating in bolt db). The LastRenewal timestamp is updated whenever there's a notification on a lease's watcher.RenewCh.

### DIFF
--- a/command/agent/cache/lease_cache.go
+++ b/command/agent/cache/lease_cache.go
@@ -299,6 +299,13 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
 		RequestPath: req.Request.URL.Path,
 	}
 
+	lastRenewed, err := http.ParseTime(resp.Response.Header.Get("Date"))
+	if err != nil {
+		c.logger.Error("failed to parse response header date", "error", err)
+		return nil, err
+	}
+	index.LastRenewed = lastRenewed
+
 	secret, err := api.ParseSecret(bytes.NewReader(resp.ResponseBody))
 	if err != nil {
 		c.logger.Error("failed to parse response as secret", "error", err)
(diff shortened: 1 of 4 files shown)


jasonodonnell (approved):
LGTM

command/agent/cache/lease_cache.go
```diff
@@ -994,6 +1020,25 @@ func (c *LeaseCache) restoreLeases(leases [][]byte) error {
...
+			c.logger.Error("failed to deserialize response", "error", err)
+			return err
+		}
+		secret, err := api.ParseSecret(resp.Body)
+		if err != nil {
+			c.logger.Error("failed to parse response as secret", "error", err)
+			return err
+		}
+		elapsed := time.Now().Sub(newIndex.LastRenewed)
+		if int(elapsed.Seconds()) > secret.LeaseDuration {
```
calvn, line 1037: For auth leases you might have to check `secret.Auth.LeaseDuration` instead.

command/agent/cache/lease_cache.go
```diff
@@ -299,6 +299,13 @@ func (c *LeaseCache) Send(ctx context.Context, req *SendRequest) (*SendResponse,
 		RequestPath: req.Request.URL.Path,
 	}
 
+	lastRenewed, err := http.ParseTime(resp.Response.Header.Get("Date"))
+	if err != nil {
+		c.logger.Error("failed to parse response header date", "error", err)
+		return nil, err
+	}
+	index.LastRenewed = lastRenewed
```
calvn, line 307: Brought up offline, but we might want to use `time.Now()`for the initial value as well since the renewal update and calculation going forward uses the OS's time.

calvn (approved):

command/agent/cache/lease_cache.go
tvoran: Added in b3c32c1ec918d083a795b882c656f773b84604ba

command/agent/cache/lease_cache.go
tvoran: Done in b3c32c1ec918d083a795b882c656f773b84604ba

PR #11031, opened 2021-03-02
Backport 1.6.4: Add partial month client count API

swayne275:


swayne275:
didn't set the right branch

PR #11032, opened 2021-03-02
Backport 1.6.4: Add partial month client count API

swayne275:


pbernal:
@swayne275  - i think we can close this.

PR #11033, opened 2021-03-02
Reset agent backoff on successful auth

kalafut:
The existing code would retain the previous backoff value even after the system had recovered. This PR fixes that issue and improves the structure of the backoff code.

### DIFF
--- a/command/agent/auth/auth.go
+++ b/command/agent/auth/auth.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"math/rand"
 	"net/http"
 	"time"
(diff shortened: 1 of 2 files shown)


command/agent/auth/auth.go
```diff
@@ -281,6 +280,7 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
 			}
 
 			am.CredSuccess()
+			backoff.reset()
 		}
```
calvn, line 284: Could we have just reset the backoff here with `backoff = initialBackoff`?

kalafut: My main motivation for the function was that the action of resetting is sprinkled about in various places so I wanted to ensure consistent handling.

command/agent/auth/auth.go
```diff
@@ -326,15 +326,41 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
...
-	return maxBackoff - time.Duration(trim)
+	b.current = maxBackoff - time.Duration(trim)
+}
+
+func (b *agentBackoff) reset() {
+	b.current = initialBackoff
+}
+
+func (b agentBackoff) String() string {
+	return fmt.Sprintf("%.2fs", b.current.Seconds())
```
calvn, line 365: You could `return b.current.String()` and let the stringer implementation from time.Duration handle the formatting.

calvn (approved):

command/agent/auth/auth.go
kalafut: Updated in https://github.com/hashicorp/vault/pull/11033/commits/eba830eb62fbb30881235183bac4aef5357e7b36. I did this since I prefer fewer sig figs... `backoff=1.91s` instead of `backoff=8.682942552s`. Looks like `time.Truncate` does that job just as well.

briankassouf (approved):

PR #11043, opened 2021-03-04
Bug fix: change default root rotation on DBSE static role from 5s to 24hr

Monkeychip:
Per discussion with backend team. 

Also, remove console log that snuck in there.


chelshaw (approved):
:chef-kissing-fingers:

PR #11048, opened 2021-03-05
Add a reasonable application_name for PostgreSQL connection

xxorde:
I noticed that Vault sets no `application_name` for PostgreSQL connections.

If this is set to an good identifier like the name of the software which created the connection, it makes it much easier for DBAs and other database users to understand what is happening on the database.
The go-to solution to see what connections are open on your PostgreSQL instance and what they are doing is to look at pg_stat_activity `SELECT * FROM pg_stat_activity;`.
 `application_name` is the value used to identify what application is connected.

I set  `application_name` to "vault" which is a lot more helpful than the default empty string.
In the future it could also be useful to deliver more information like current version, but this is optional and has pros and cons.

### DIFF
--- a/sdk/database/helper/connutil/sql.go
+++ b/sdk/database/helper/connutil/sql.go
@@ -128,12 +128,18 @@ func (c *SQLConnectionProducer) Connection(ctx context.Context) (interface{}, er
 	// Otherwise, attempt to make connection
 	conn := c.ConnectionURL
 
-	// Ensure timezone is set to UTC for all the connections
+	// PostgreSQL specific settings
 	if strings.HasPrefix(conn, "postgres://") || strings.HasPrefix(conn, "postgresql://") {
+		// Ensure timezone is set to UTC for all the connections
 		if strings.Contains(conn, "?") {
 			conn += "&timezone=UTC"
 		} else {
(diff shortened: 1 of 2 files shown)


xxorde:
I am not quite sure if my commit broke the CI or if there are other causes.
Help to find out is appreciated.

xxorde:
I brought my commit up too date. Would be great if you could review it, should be strait forward.

sgmiller (approved):
Nice little addition.

sgmiller:
One small hurdle left, can you run "go mod vendor" and commit the resulting change?

sgmiller:
Also, you may need to refresh permissions in order to have the CI trigger when you commit.

xxorde:
> @xxorde: One small hurdle left, can you run "go mod vendor" and commit the resulting change?

Ah thank you. I didn't see the failed CI job, when I looked at them the last time they were not started.

xxorde:
Th "test-go-race" job looks a bit racy, I just push my same code again to trigger a new run.

xxorde:
@sgmiller what is blocking the merge, can I do something?

xxorde:
@sgmiller thank you! :+1:

PR #11050, opened 2021-03-05
Make sure we sanitize the rotation config on each clone

briankassouf:


sgmiller (approved):

vault/barrier_aes_gcm_test.go
```diff
@@ -61,6 +61,37 @@ func TestAESGCMBarrier_Rotate(t *testing.T) {
...
+	oldKeyring := b.keyring.Clone()
+	oldKeyring.rotationConfig = KeyRotationConfig{}
+	b.persistKeyring(context.Background(), oldKeyring)
+
+	b.ReloadKeyring(context.Background())
+
+	// At this point, the rotation config should match the default
+	rc := b.keyring.rotationConfig
+
+	if rc.Disabled != defaultRotationConfig.Disabled || rc.MaxOperations != defaultRotationConfig.MaxOperations ||
```
briankassouf, line 89: rc.Equals could be used here

sgmiller: good call.

PR #11060, opened 2021-03-08
Update go version to 1.15.8

swayne275:
Note: I did some manual editing of files that I think should be automated (like the go version in `.circleci/config.yml` and `.circleci/config/executors/@executors.yml` - @samsalisbury please comment if this was something I shouldn't be doing).


briankassouf (approved):

raskchanky (approved):

swayne275:
okay @samsalisbury that attempt failed `build-common-layers` so i went back, just updated `Dockerfile/Dockerfile.ui` and `packages-oss.yml` to see if it works. It seems running `make packages` leaves remnants of the old versions throughout the code (see above mentioned files)

samsalisbury:
Hi @swayne275, I was out Mon & Tue this week so just saw this, sorry.

The Go versions in `.circleci/config/executors/@executors.yml` and in `Dockerfile/Dockerfile.ui` need to be updated manually, because these are not Packagespec-managed files. Updating `packages-oss.yml` updates all the packagespec-managed parts of the CircleCI config though.

The new CD pipeline work released in packagespec 0.2.0 would obviate the need for the `.circleci/config/executors/@executors.yml` file, because it is capable of generating all the CI config needed to run tests as well as build/stage/publish, but it looks as though we won't be allowed to use that sadly.

swayne275:
Could you tell me more about not being able to use that @samsalisbury ? Thanks!

samsalisbury:
Sure will DM you @swayne275

PR #11077, opened 2021-03-10
Add changelog entry for #10953

Valarissa:


changelog/10953.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+secret/aws: add IAM tagging support for iam_user roles
```
catsby, line 2:
```suggestion
secrets/aws: add IAM tagging support for iam_user roles
```

catsby: I believe this is "more" consistent with other entries. There are in fact some entries that are `secret/` but most I believe are `secrets/`

catsby (approved):
👍

PR #10233, opened 2020-10-24
add consul service metadata

Chrisss93:
An extra configuration parameter "meta" in the Vault configuration's storage block to specify metadata for Consul service registration. Addresses #9121

PR #11084, opened 2021-03-11
Allow setting of Consul ServiceMeta tags from config file

### LINKED ISSUE #9121: Add Metadata to Service Registration in Consul
We have some internal automation that sets up networking for our Consul services based on the metadata attached to our services. Vault currently doesn't have the option to set metadata in the `service_registration` stanza and so we can't use our internal automation to setup additional networking.

We would like the ability to configure the metadata for when Vault registers itself with Consul. This could by done by extending the `service_registration` stanza for using Consul.

```
service_registration "consul" {
    metadata = {
        tag1 = "value1"
        tag2 = "value2"
    }
 }
```

We've tried updating Vault self registration using the Consul HTTP API, but realized that Vault would override those updates.

We could also disable to self registration, but we would then lose some of the automatic configuration of the HA redirect address.

We've resorted to manually setting up our three Vault services in our networking and it works for now, but we'd like to bring Vault into the folds of our existing automation.

This is a nice to have for customizing Vault's self registration. There is already the option for setting `service_tags` in the `consul` service_registration parameters.

**Additional context**
None

unRob:
Hi there!

I'd love to be able to set [`ServiceMeta`](https://www.consul.io/api-docs/catalog#servicemeta) on the Consul service registered by Vault by specifying the desired tags in the config file, and I'm submitting this PR to that effect.

I'm not happy about unmarshalling json during `NewServiceRegistration` but considered that a simpler change than the couple of alternatives I came up from a quick glance at the codebase. That is to say, I'd love feedback on the approach (and perhaps suggestions on a proper testing strategy should folks feel this approach works).

fixes #9121
see also #10233

### DIFF
--- a/serviceregistration/consul/consul_service_registration.go
+++ b/serviceregistration/consul/consul_service_registration.go
@@ -5,6 +5,7 @@ package consul
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math/rand"
@@ -73,6 +74,7 @@ type serviceRegistration struct {
 	redirectPort        int64
 	serviceName         string
 	serviceTags         []string
+	serviceMeta         map[string]string
 	serviceAddress      *string
 	disableRegistration bool
 	checkTimeout        time.Duration
@@ -229,6 +231,20 @@ func (c *serviceRegistration) merge(conf map[string]string) error {
 		c.logger.Debug("config service_tags set", "service_tags", tags)
 	}
 
+	// Get user-defined meta tags to attach to the registered service name
+	metaTagsJSON, ok := conf["service_meta"]
+	var metaTags = map[string]string{}
+	if ok {
+		if err := json.Unmarshal([]byte(metaTagsJSON), &metaTags); err != nil {
+			return errors.New("service tags must be a dictionary of string keys and values")
+		}
+		metaTags["external-source"] = metaExternalSource
+
+		if c.logger.IsDebug() {
+			c.logger.Debug("config service_meta set", "service_meta", metaTags)
+		}
+	}
+
 	// Get the service-specific address to override the use of the HA redirect address
 	var serviceAddr *string
 	serviceAddrStr, ok := conf["service_address"]
@@ -294,6 +310,7 @@ func (c *serviceRegistration) merge(conf map[string]string) error {
 	c.config = consulConf
 	c.serviceName = service
 	c.serviceTags = strutil.ParseDedupAndSortStrings(tags, ",")
+	c.serviceMeta = metaTags
 	c.serviceAddress = serviceAddr
 	c.checkTimeout = checkTimeout
 	c.disableRegistration = disableRegistration
@@ -586,12 +603,10 @@ func (c *serviceRegistration) reconcileConsul() (serviceID string, err error) {
 		ID:                serviceID,
 		Name:              c.serviceName,
 		Tags:              tags,
+		Meta:              c.serviceMeta,
 		Port:              int(c.redirectPort),
 		Address:           serviceAddress,
 		EnableTagOverride: false,
-		Meta: map[string]string{
-			"external-source": metaExternalSource,
-		},
 	}
 
 	checkStatus := api.HealthCritical
--- a/serviceregistration/consul/consul_service_registration.go
+++ b/serviceregistration/consul/consul_service_registration.go
@@ -232,17 +232,15 @@ func (c *serviceRegistration) merge(conf map[string]string) error {
 	}
 
 	// Get user-defined meta tags to attach to the registered service name
-	metaTagsJSON, ok := conf["service_meta"]
 	var metaTags = map[string]string{}
-	if ok {
+	if metaTagsJSON, ok := conf["service_meta"]; ok {
 		if err := json.Unmarshal([]byte(metaTagsJSON), &metaTags); err != nil {
 			return errors.New("service tags must be a dictionary of string keys and values")
 		}
-		metaTags["external-source"] = metaExternalSource
-
-		if c.logger.IsDebug() {
-			c.logger.Debug("config service_meta set", "service_meta", metaTags)
-		}
+	}
+	metaTags["external-source"] = metaExternalSource
+	if c.logger.IsDebug() {
+		c.logger.Debug("config service_meta set", "service_meta", metaTags)
 	}
 
 	// Get the service-specific address to override the use of the HA redirect address
--- a/serviceregistration/consul/consul_service_registration.go
+++ b/serviceregistration/consul/consul_service_registration.go
@@ -232,7 +232,7 @@ func (c *serviceRegistration) merge(conf map[string]string) error {
 	}
 
 	// Get user-defined meta tags to attach to the registered service name
-	var metaTags = map[string]string{}
+	metaTags := map[string]string{}
 	if metaTagsJSON, ok := conf["service_meta"]; ok {
 		if err := json.Unmarshal([]byte(metaTagsJSON), &metaTags); err != nil {
 			return errors.New("service tags must be a dictionary of string keys and values")
--- a/serviceregistration/consul/consul_service_registration_test.go
+++ b/serviceregistration/consul/consul_service_registration_test.go
@@ -492,22 +492,32 @@ func TestConsul_serviceTags(t *testing.T) {
 	}
 }
 
-func TestConsul_serviceMeta(t *testing.T) {
+// TestConsul_ServiceMeta tests whether consul service meta registration works
+func TestConsul_ServiceMeta(t *testing.T) {
 	tests := []struct {
-		conf map[string]string
-		pass bool
+		conf   map[string]string
+		pass   bool
+		expect map[string]string
 	}{
 		{
-			conf: map[string]string{},
-			pass: true,
+			conf:   map[string]string{},
+			pass:   true,
+			expect: map[string]string{"external-source": "vault"},
 		},
 		{
-			conf: map[string]string{"service_meta": "true"},
-			pass: false,
+			conf:   map[string]string{"service_meta": "true"},
+			pass:   false,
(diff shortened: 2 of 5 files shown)


unRob:
Still very interested in seeing this through, any chance it could be reviewed?

VioletHynes:
Hey there @unRob ! I'm really very sorry this one has waited so long, but I'm pretty eager to get it merged in as long as you are.

There's just a few things I'd like to confirm/address before we do. I'll be leaving a review shortly with some comments, but I'll be watching this closely for updates, and please feel free to tag me just in case :)

serviceregistration/consul/consul_service_registration_test.go
```diff
@@ -492,6 +492,43 @@ func TestConsul_serviceTags(t *testing.T) {
...
+		{
+			conf: map[string]string{},
+			pass: true,
+		},
+		{
+			conf: map[string]string{"service_meta": "true"},
+			pass: false,
+		},
+		{
+			conf: map[string]string{"service_meta": "[{\"key\":\"value\"}]"},
```
VioletHynes, line 509: How come the config must be a list? Since we're only taking the first of these values, I'd expect to be able to configure this as a set of JSON strings, e.g.
```
conf: map[string]string{"service_meta": "{\"key\":\"value\", \"bar\":\"baz\"}"},
```

serviceregistration/consul/consul_service_registration.go
```diff
@@ -229,6 +232,22 @@ func (c *serviceRegistration) merge(conf map[string]string) error {
...
+  // Get user-defined meta tags to attach to the registered service name
+	metaTagsJSON, ok := conf["service_meta"]
+	var metaTags = map[string]string{}
+	if ok {
+		metaTagList := []map[string]string{}
+		err := json.Unmarshal([]byte(metaTagsJSON), &metaTagList)
+		if err != nil {
+			return nil, errors.New("service tags must be a dictionary of string keys and values")
+		}
+		metaTags = metaTagList[0]
```
VioletHynes, line 244: As mentioned elsewhere, I'm a little confused why we're unmarshaling this into a list first and then taking the first element. Could this be changed so that we don't expect a list?

serviceregistration/consul/consul_service_registration.go
```diff
@@ -229,6 +232,22 @@ func (c *serviceRegistration) merge(conf map[string]string) error {
...
+
+		if logger.IsDebug() {
+			logger.Debug("config service_meta set", "service_meta", metaTags)
+		}
```
VioletHynes, line 248:
```suggestion
		c.logger.Debug("config service_meta set", "service_meta", metaTags)
```

website/content/docs/configuration/service-registration/consul.mdx
```diff
@@ -85,6 +85,9 @@ at Consul's service discovery layer.
 - `service_tags` `(string: "")` – Specifies a comma-separated list of case-sensitive
   tags to attach to the service registration in Consul.
 
+- `service_meta` `(map[string]string: {})` – Specifies a key-value list of meta tags to
+  attach to the service registration in Consul.
```
VioletHynes, line 89: I think an example here would be great

VioletHynes, line 89:
```suggestion
  attach to the service registration in Consul. See [ServiceMeta](https://developer.hashicorp.com/consul/api-docs/catalog#servicemeta) in the Consul docs for more information.
```

website/content/docs/configuration/storage/consul.mdx
```diff
@@ -92,6 +92,9 @@ and [`cluster_addr`][cluster-addr] ([example][listener-example]).
 - `service_tags` `(string: "")` – Specifies a comma-separated list of tags to
   attach to the service registration in Consul.
 
+- `service_meta` `(map[string]string: {})` – Specifies a key-value list of meta tags to
+  attach to the service registration in Consul.
```
VioletHynes, line 96:
```suggestion
  attach to the service registration in Consul. See [ServiceMeta](https://developer.hashicorp.com/consul/api-docs/catalog#servicemeta) in the Consul docs for more information.
```

serviceregistration/consul/consul_service_registration.go
```diff
@@ -229,6 +232,22 @@ func (c *serviceRegistration) merge(conf map[string]string) error {
 		c.logger.Debug("config service_tags set", "service_tags", tags)
 	}
 
+  // Get user-defined meta tags to attach to the registered service name
```
VioletHynes, line 235: The formatting looks a little strange here -- could you give this a `make fmt`/align this?

serviceregistration/consul/consul_service_registration.go
```diff
@@ -229,6 +232,22 @@ func (c *serviceRegistration) merge(conf map[string]string) error {
...
+	metaTagsJSON, ok := conf["service_meta"]
+	var metaTags = map[string]string{}
+	if ok {
+		metaTagList := []map[string]string{}
+		err := json.Unmarshal([]byte(metaTagsJSON), &metaTagList)
+		if err != nil {
+			return nil, errors.New("service tags must be a dictionary of string keys and values")
```
VioletHynes, line 242: It seems the code has changed around this, and this should just be
`return errors.New("service tags must be a dictionary of string keys and values")`

VioletHynes (commented):
Thanks for bearing with us on this. A few comments and things I'd like to see answered/addressed, but I'm eager to get this merged. I agree that the approach of deserializing the JSON where you have seems to be the easiest place and I don't see it as too problematic.

I know this took a while to be reviewed but I'm committed to helping you see this to the finish line, as long as you're still willing :)

VioletHynes:
Oh, one last thing, we would require a changelog for this. I'm happy to do it myself if you don't want to fiddle, but it'd be a file in the `changelog` directory called `11084.txt` with something like the following text:

```
```release-note:improvement
serviceregistration: Added support for Consul ServiceMeta tags from config file from the new `service_meta` config field.
```
```

File: serviceregistration/consul/consul_service_registration.go (hunk 6 above)
Comment (unRob) [reply serviceregistration/consul/consul_service_registration.go:235]: Yeah, I used github's UI to resolve the refactor of `NewServiceRegistration` / `serverRegistration.merge` and did not really pay attention, while I had hope, I was not really expecting someone to bring this PR back from the dead 😂 . Fixing!

File: serviceregistration/consul/consul_service_registration.go (hunk 2 above)
Comment (unRob) [reply serviceregistration/consul/consul_service_registration.go:244]: I don't recall why I made this choice in the context of the codebase I originally submitted this PR to, but agree it does not make any sense at all as in config, these values—as part of `storage` or `service_registration`— will be maps. Vaguely remember having trouble understanding if I should be using HCL blocks or objects.

I think my original intention was to allow something like the following config snippet

```hcl

storage "consul" {
  service_meta = {some_key = "some value"}
}
```

It's been a while, and I eventually ended up hacking around this shortcoming, so don't really have an informed opinion. Fixing.

serviceregistration/consul/consul_service_registration.go
```diff
@@ -229,6 +231,20 @@ func (c *serviceRegistration) merge(conf map[string]string) error {
...
 	}
 
+	// Get user-defined meta tags to attach to the registered service name
+	metaTagsJSON, ok := conf["service_meta"]
+	var metaTags = map[string]string{}
+	if ok {
+		if err := json.Unmarshal([]byte(metaTagsJSON), &metaTags); err != nil {
+			return errors.New("service tags must be a dictionary of string keys and values")
+		}
+		metaTags["external-source"] = metaExternalSource
```
VioletHynes, line 241: We should do this irrespective of `if ok`, since otherwise there will be a behaviour change as part of this PR. Previously, the tag would have been set, but now, it won't be set unless I provide my own meta tags.

VioletHynes: To put it another way: if I don't provide my own `service_meta`, it should still be:
```
map[string]string{
	"external-source": metaExternalSource,
},
```

serviceregistration/consul/consul_service_registration_test.go
```diff
@@ -492,6 +492,43 @@ func TestConsul_serviceTags(t *testing.T) {
 	}
 }
 
+func TestConsul_serviceMeta(t *testing.T) {
```
VioletHynes, line 495: Nit: We should have a godoc describing what this test does and is about

VioletHynes, line 495:
```suggestion
func TestConsul_ServiceMeta(t *testing.T) {
```

serviceregistration/consul/consul_service_registration_test.go
VioletHynes, line 1: I'd love to see a test in some respect here that validates that if I don't provide my own `service_meta`, that it's still set to what it was before (`"external-source": metaExternalSource`)

VioletHynes (commented):
Looks great! Just a few more things before we can get this over the finish line. Thanks for the quick response so far!

serviceregistration/consul/consul_service_registration.go
VioletHynes: I'd suggest moving this line until after the `if ok`, that way the behaviour is preserved.

serviceregistration/consul/consul_service_registration_test.go
VioletHynes: One way we could do this is by improving `TestConsul_ServiceMeta` to test the resultant map for the three configurations, but I'm open to whatever you feel is easiest :)

serviceregistration/consul/consul_service_registration.go
unRob: good catch, fixed!

VioletHynes (approved):
Thank you! I appreciate your patience for bearing with us on this, but thanks for being so understanding and great to work with. Approved! I'll look to get this merged as soon as all the CI passes etc.

PR #11096, opened 2021-03-11
UI/box radio updates

chelshaw:
Cleans up the box-radio component and addresses some odd behavior on the backend and auth method selection screen where after clicking many options, at a certain point the value stops updating. It is best made obvious in this video: 

https://user-images.githubusercontent.com/16182107/110854680-ee443300-827a-11eb-950a-3954f976965a.mov

This update also made the radio inputs more keyboard-navigable:
box-radio


ui/lib/core/addon/templates/components/box-radio.hbs
```diff
@@ -1,64 +1,66 @@
...
-          @radioId={{@type}}
-          @disabled={{@disabled}}
+        <label for={{@key}}>{{@displayName}}</label>
+        <input
+          type="radio"
+          disabled
+          id={{@key}}
+          name={{@groupName}}
+          value={{@key}}
+          tabindex="-1"
```
meirish, line 24: Are we sure changing the native radio group behavior is what you want to do here? I find the default behavior to be strange (you focus the group and then can use arrow keys to navigate through the options like it's one big control), but I don't know if it's surprising to not use the default.

ui/lib/core/addon/templates/components/box-radio.hbs
```diff
@@ -1,64 +1,66 @@
...
+        data-test-mount-type={{@key}}
+        tabindex="0"
       >
         <Icon
           @glyph={{@glyph}}
           @size="xl"
           class="has-text-grey-light"
         />
-        {{@displayName}}
-        <RadioButton
```
meirish, line 20: Is this the only place `RadioButton` was still in use? I think it was provided by an addon, so you may be able to remove the addon if this is the only place.

chelshaw: Unfortunately not, there are a few other places that I intend to update in a different PR

ui/lib/core/addon/templates/components/box-radio.hbs
chelshaw: Since the input itself is hidden, the default behavior was obscured and led to a confusing interface where tabbing through it looks like nothing is selected. Open to other ways of tackling this though

stasquatch: I had this highlighted in my review before I saw this comment. The tabbing vs arrow key functionality seems a little strange to me. I'd assume a disabled input would be skipped without needing to add `tabindex="-1"` here. It seems surprising, in addition, to use tab instead of arrow keys for moving through available options.

meirish: Ah yeah - is hiding it key to the fix? Was that causing the lag in updating?

I think previously the styling for focus vs not focused wasn't super differentiated:

Focused:
CleanShot 2021-03-15 at 11 40 30@2x

Not focused:
CleanShot 2021-03-15 at 11 40 38@2x

chelshaw: I was trying to highlight the box instead of just the mock-input, but with the most recent changes I'm keeping the default radio input because it allows for the arrow (not tab) moving around. 

Thank you all for the discourse! Let me know if the new approach seems unreasonable for any reason

ui/lib/core/addon/components/box-radio.js
```diff
@@ -7,22 +7,29 @@
...
 
 import Component from '@glimmer/component';
 import layout from '../templates/components/box-radio';
+import { action } from '@ember/object';
 import { setComponentTemplate } from '@ember/component';
 
 class BoxRadio extends Component {
   disabled = false;
-  tooltipMessage = 'This option is not available to you at this time.';
+  defaultTooltipMessage = 'This option is not available to you at this time.';
```
Monkeychip, line 26: nice naming

Monkeychip (approved):
Nice work.  Thanks for adding the story and the a11y labeling!

hashishaw:
Will re-open when browserstack is investigated and fixed for this branch

PR #11105, opened 2021-03-15
update transform.mdx

basingh:
Typo on path of creating/updating transformation in api docs. 
Fixed by updating `transform/transformation` to `transform/transformations`


sgmiller (approved):
Quite correct.

PR #11114, opened 2021-03-16
Update go version to 1.15.10

kalafut:


swayne275 (commented):
My guess is that `Dockerfile.ui` should also be updated?

kalafut:
@swayne275 Good catch. Updated!

calvn (approved):

swayne275 (approved):

PR #11116, opened 2021-03-16
Fix handling of minimum operations, and forward rotate/config requests to Primary

sgmiller:


briankassouf (approved):
👍

vault/keyring.go
```diff
@@ -253,9 +253,12 @@ func (c KeyRotationConfig) Clone() KeyRotationConfig {
 }
 
 func (c *KeyRotationConfig) Sanitize() {
-	if c.MaxOperations == 0 || c.MaxOperations > absoluteOperationMaximum || c.MaxOperations < absoluteOperationMinimum {
+	if c.MaxOperations == 0 || c.MaxOperations > absoluteOperationMaximum {
 		c.MaxOperations = absoluteOperationMaximum
 	}
+	if c.MaxOperations < absoluteOperationMinimum {
+		c.MaxOperations = absoluteOperationMinimum
```
HridoyRoy, line 260: Could we just skip this statement, so c.MaxOperations remains the same as it was previously?

HridoyRoy (changes requested):

vault/keyring.go
sgmiller: no, this is making sure it's never below the min.

sgmiller: oh, in the API call we could, yes.

sgmiller: Updated

HridoyRoy (approved):

PR #11122, opened 2021-03-17
Acquire a per-lock lease to make renew and revoke atomic wrt each other.

mgritter:
This means we don't have to hold pendingLock during I/O.


vault/expiration.go
```diff
@@ -1232,6 +1243,11 @@ func (m *ExpirationManager) RenewToken(ctx context.Context, req *logical.Request
 		leaseID = fmt.Sprintf("%s.%s", leaseID, ns.ID)
 	}
 
+	// Acquire lock for this lease
+	leaseLock := m.lockForLeaseID(leaseID)
+	leaseLock.Lock()
```
mgritter, line 1248: I could save a few lines by having lockForLeaseID() also acquire the lock, I guess.

vault/expiration.go
```diff
@@ -1815,7 +1854,16 @@ func (m *ExpirationManager) loadEntry(ctx context.Context, leaseID string) (*lea
 	} else {
 		ctx = namespace.ContextWithNamespace(ctx, namespace.RootNamespace)
 	}
-	return m.loadEntryInternal(ctx, leaseID, restoreMode, true)
+
+	//
```
swayne275, line 1858: is this blank comment line intentional?

vault/expiration.go
```diff
@@ -1815,7 +1854,16 @@ func (m *ExpirationManager) loadEntry(ctx context.Context, leaseID string) (*lea
...
+	// If a lease entry is nil, proactively delete the lease lock, in case we
+	// created one erroneously.
+	leaseEntry, err := m.loadEntryInternal(ctx, leaseID, restoreMode, true)
+	if err != nil && leaseEntry == nil {
+		m.deleteLockForLease(leaseID)
```
swayne275, line 1863: why wouldn't we also want to do this if error is nil?

swayne275 (approved):
Approving, but did you ever get to the bottom of the potential deadlock you were seeing in tests?

mgritter:
> Approving, but did you ever get to the bottom of the potential deadlock you were seeing in tests?

Yes, https://github.com/hashicorp/vault/pull/11122/commits/1704a8d1c32378cc2c60f022aaecbd9b6ffe93e4 fixes it.
The problem is that when a token is revoked, revokeCommon is called on its lease *twice*.  There's already a flag to prevent an infinite recursion, so that seems like the right thing to do to avoid the deadlock too.  But it feels a little hacky, or like the double-revoke thing means that there's a design problem that could be refactored away.

vault/expiration.go
mgritter: Hm.... this is backwards, actually, thanks for asking about it. 

If err != nil, then we don't know whether the entry is there or not, so we should not assume that it isn't.

vault/expiration.go
mgritter: Fixed.

sgmiller (approved):

mladlow:
@sgmiller @swayne275 could one of you resolve the conflicts and merge this PR, since it seems like something we want, and Mark can't merge it?

vault/expiration.go
```diff
@@ -937,6 +943,17 @@ func (m *ExpirationManager) LazyRevoke(ctx context.Context, leaseID string) erro
 func (m *ExpirationManager) revokeCommon(ctx context.Context, leaseID string, force, skipToken bool) error {
 	defer metrics.MeasureSince([]string{"expire", "revoke-common"}, time.Now())
 
+	if !skipToken {
+		// Acquire lease for this lock
```
swayne275, line 947:
```suggestion
		// Acquire lock for this lease
```

vault/expiration.go
```diff
@@ -1157,6 +1160,11 @@ func (m *ExpirationManager) revokePrefixCommon(ctx context.Context, prefix strin
 func (m *ExpirationManager) Renew(ctx context.Context, leaseID string, increment time.Duration) (*logical.Response, error) {
 	defer metrics.MeasureSince([]string{"expire", "renew"}, time.Now())
 
+	// Acquire lock for this lease
+	leaseLock := m.lockForLeaseID(leaseID)
+	leaseLock.Lock()
+	defer leaseLock.Unlock()
+
 	// Load the entry
 	le, err := m.loadEntry(ctx, leaseID)
```
swayne275, line 1169: since we're dealing with races between revoke and renew, do we need to check for `le != nil` here? What if we revoke just before a renew (deleting it so it can't be loaded), and then renew acquires the lock and goes through this code?

vault/expiration.go
```diff
@@ -1299,6 +1302,11 @@ func (m *ExpirationManager) RenewToken(ctx context.Context, req *logical.Request
 		leaseID = fmt.Sprintf("%s.%s", leaseID, ns.ID)
 	}
 
+	// Acquire lock for this lease
+	leaseLock := m.lockForLeaseID(leaseID)
+	leaseLock.Lock()
+	defer leaseLock.Unlock()
+
 	// Load the entry
 	le, err := m.loadEntry(ctx, leaseID)
```
swayne275, line 1311: same `le != nil` question here

vault/expiration.go
```diff
@@ -1716,6 +1731,30 @@ func (m *ExpirationManager) uniquePoliciesGc() {
 	}
 }
 
+// Placing a lock in pendingMap means that we need to work very hard on reload
+// to only create one lock.  Instead, we'll create locks on-demand in an atomic fashion.
+//
+// Acquiring a lock from a leaseEntry is a bad idea because it could change
+// between loading and acquirung the lock. So we only provide an ID-based map, and the
```
swayne275, line 1738:
```suggestion
// between loading and acquiring the lock. So we only provide an ID-based map, and the
```

swayne275 (commented):
looks good, just a few questions

vault/expiration.go
briankassouf: Is leaking memory your concern here? The loadEntry code path will delete the lock if the leaseID doesn't exist.

swayne275: i was concerned about trying to renew le if it had been deleted by a revoke call that came in just before it. i just checked the code for `*leaseEntry.renewable()` though and it returns an error if `le == nil`, so i think we're good here

vault/expiration.go
briankassouf: This is the same answer as above!

vault/expiration.go
```diff
@@ -937,6 +943,17 @@ func (m *ExpirationManager) LazyRevoke(ctx context.Context, leaseID string) erro
 func (m *ExpirationManager) revokeCommon(ctx context.Context, leaseID string, force, skipToken bool) error {
 	defer metrics.MeasureSince([]string{"expire", "revoke-common"}, time.Now())
 
+	if !skipToken {
+		// Acquire lease for this lock
+		// If skipToken is true, then we're either being (1) called via RevokeByToken, so
+		// probably the lock is already held, and if we re-acquire we get deadlock, or
+		// (2) called by tidy, in which case the lock is not held.
+		// Is it worth separating those cases out, or is (2) OK to proceed unlocked?
```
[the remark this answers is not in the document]
briankassouf: addressed in d983c06

PR #11126, opened 2021-03-17
Add note that static role is rotated upon creation

assareh:
It does not appear to be documented that Vault must rotate the password upon static role creation in order to know the password, as it is not provided.


catsby (approved):
👍

PR #11127, opened 2021-03-17
UI/fix backend db connection

chelshaw:
Fixes bug where only database secret engine connection actions backend was hardcoded. This change allows any named database to accomplish connection actions such as edit and rotate.


Monkeychip (approved):
Nice catch!

PR #11130, opened 2021-03-17
Packagespec 0.2.2

99:
This PR upgrades to the packagespec 0.2.2


swayne275 (approved):

samsalisbury (approved):
LGTM! This correctly only updates a single line in the `packages-oss.lock/pkgs.yml` file.

(Sometimes other files are changed when upgrading packagespec, if the upgrade includes _build_ system changes for example, but this upgrade only affects private release workflows, so the packagespec version is the only thing this repo needs to know to have the upgraded versions of them applied correctly at release time.)

claire-labry (approved):
LGTM

PR #11134, opened 2021-03-17
Backport 1.7: Add a test for server stabilization (#11128)

ncabatoff:


vishalnayak (approved):

physical/raft/raft.go
```diff
@@ -666,6 +681,7 @@ func (b *RaftBackend) SetupCluster(ctx context.Context, opts SetupOpts) error {
 
 	// Setup the raft config
 	raftConfig := raft.DefaultConfig()
+	raftConfig.SnapshotInterval = 5 * time.Second
```
briankassouf, line 684: What's the rational for setting this to 5s?

ncabatoff: Can't set it too high or we get a timeout on the writer side.  I wanted it higher than the stabilization time so that we'd see that node continue to not be promoted for a bit after the stabilization+reconciliation had occurred, then eventually see it promoted.

briankassouf (approved):

PR #11142, opened 2021-03-18
Update MMMM Do yyyy to MMMM do yyyy per date-fns format docs

chelshaw:
This change corrects the output of the token expiration warning, which was using the key for day of year instead of day of month: 

image (1)


Monkeychip (approved):
Nice catch.  There were a lot of small changes like this with the Date fns library update.  Doesn't look like I got this one correct.

PR #11143, opened 2021-03-18
UI/control groups kv

Monkeychip:
Addressing a ticket where the error message was not displaying correctly.  

error

### DIFF
--- a/ui/app/components/secret-edit.js
+++ b/ui/app/components/secret-edit.js
@@ -226,6 +226,7 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
           let errorMessage = this.controlGroup.logFromError(error);
           this.set('error', errorMessage.content);
         }
+        throw error;
       });
   },
   saveComplete(callback, key) {
--- a/ui/app/components/secret-edit.js
+++ b/ui/app/components/secret-edit.js
@@ -193,10 +193,12 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
     let secretData = this.modelForData;
     let isV2 = this.isV2;
     let key = secretData.get('path') || secret.id;
+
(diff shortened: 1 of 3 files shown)


ui/app/components/secret-edit.js
```diff
@@ -197,26 +199,34 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
...
-              this.set(e, e.errors.join(' '));
-            });
-        } else {
-          this.saveComplete(successCallback, key);
+      })
+      .catch(error => {
+        if (error instanceof ControlGroupError) {
+          let errorMessage = this.controlGroup.logFromError(error);
+          this.set('error', errorMessage.content);
         }
```
chelshaw, line 228: To keep the original functionality here, we should `throw error` if it's not an instance of control group error

ui/app/routes/vault/cluster/access/identity/show.js
```diff
@@ -20,11 +20,6 @@ export default Route.extend({
     // if the record is in the store use that
     let model = this.store.peekRecord(modelType, params.item_id);
 
-    // if we don't have creationTime, we only have a partial model so reload
-    if (model && !model.get('creationTime')) {
```
chelshaw, line 24: Is this block part of this fix or a different bug? If part of the other bug we should describe it in the PR description

chelshaw (changes requested):
Couple small changes, great work here! 💪

ui/app/routes/vault/cluster/access/identity/show.js
Monkeychip: I'm actually going to remove it.  It's apart of a larger issue that has several other bugs attached to it. I've got the other bugs documented and will add this to the list.  Good call.

chelshaw (approved):

PR #11147, opened 2021-03-19
Replace non-inclusive language in AWS Auth

kalafut:
These occurrences were missed in the original PR. The changes here are limited to local variable and struct names.


VioletHynes:
Will do! 

Closing as per previous comment. There are also merge conflicts and CLA signing that would be required to merge this.

PR #11155, opened 2021-03-19
storage/raft: Ensure peers are informed of their correct suffrage when added with AutoPilot

briankassouf:
Prior to this fix if a node was added as a non-voter by AutoPilot we were not correctly informing the follower of their suffrage. 

On the leader we see:
```
core0.raft: updating configuration: command=AddNonvoter server-id=core-1 server-addr=inmem-cluster_node_1 servers="[{Suffrage:Voter ID:core-0 Address:inmem-cluster_node_0} {Suffrage:Nonvoter ID:core-1 Address:inmem-cluster_node_1}]"
```

and on the follower, notice how both nodes are voters:

```
core1.raft: initial configuration: index=1 servers="[{Suffrage:Voter ID:core-0 Address:inmem-cluster_node_0} {Suffrage:Voter ID:core-1 Address:inmem-cluster_node_1}]"
```

### DIFF
--- a/vault/external_tests/raft/raft_autopilot_test.go
+++ b/vault/external_tests/raft/raft_autopilot_test.go
@@ -340,3 +340,54 @@ func TestRaft_Autopilot_Stabilization_Delay(t *testing.T) {
 	}
 	require.Equal(t, state.Voters, []string{"core-0", "core-1", "core-2"})
 }
+
+func TestRaft_AutoPilot_Peersets_Equivalent(t *testing.T) {
+	cluster := raftCluster(t, &RaftClusterOpts{
+		InmemCluster:         true,
+		EnableAutopilot:      true,
+		DisableFollowerJoins: true,
+	})
+	defer cluster.Cleanup()
+	testhelpers.WaitForActiveNode(t, cluster)
+
+	// Create a very large stabilization time so we can test the state between
+	// joining and promotions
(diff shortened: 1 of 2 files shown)


vishalnayak (approved):

ncabatoff (approved):

vishalnayak (approved):

vault/external_tests/raft/raft_autopilot_test.go
```diff
@@ -340,3 +340,54 @@ func TestRaft_Autopilot_Stabilization_Delay(t *testing.T) {
...
+		"server_stabilization_time": "1h",
+	})
+	require.NoError(t, err)
+
+	joinFunc := func(core *vault.TestClusterCore) {
+		_, err := core.JoinRaftCluster(namespace.RootContext(context.Background()), []*raft.LeaderJoinInfo{
+			{
+				LeaderAPIAddr: client.Address(),
+				TLSConfig:     cluster.Cores[0].TLSConfig,
+				Retry:         true,
```
ncabatoff, line 366: With Retry=true, it looks like JoinRaftCluster spawns a goroutine and always returns nil, so this test may not be deterministic.

briankassouf: fixed in https://github.com/hashicorp/vault/pull/11158

PR #11160, opened 2021-03-19
Expose snapshot_interval tunable 

ncabatoff:
...instead of setting it in prod code for the sake of a test.


briankassouf (approved):
Want to add docs? You can mostly copy from consul: https://www.consul.io/docs/agent/options#_raft_snapshot_interval

PR #11167, opened 2021-03-20
Autopilot Docs

vishalnayak:
Still a WIP, but putting it out for early feedback.


website/content/docs/commands/operator/raft.mdx
```diff
@@ -184,3 +184,112 @@ Usage: vault operator raft snapshot restore <snapshot_file>
 
 	  $ vault operator raft snapshot restore raft.snap
 ```
+
+## autopilot
+
+This command groups subcommands for operators interacting with the autopilot
+functionality of the integrated Raft storage backend. There are 3 subcommands
+supported: `get-config`, `set-config` and `state`.
```
mladlow, line 192: It's useful to know (although maybe this is a thing that belongs more in the API docs) what kind of node can run these commands? Only the active? Perf standbys? DR clustes?

website/content/docs/commands/operator/raft.mdx
```diff
@@ -184,3 +184,112 @@ Usage: vault operator raft snapshot restore <snapshot_file>
...
+
+Subcommands:
+    get-config    Returns the configuration of the autopilot subsystem under integrated storage
+    set-config    Modify the configuration of the autopilot subsystem under integrated storage
+    state         Displays the state of the raft cluster under integrated storage as seen by autopilot
+```
+
+### autopilot state
+
+Displays the state of the raft cluster under integrated storage as seen by autopilot.
```
mladlow, line 208: As this just repeats the usage information, I think it would be better to explain what we are seeing in the output.

I suggested name, but is that nodeID?

I don't actually know if we have a link where non-voters are described. If not, we should add somewhere? Concepts page?

```suggestion
Displays the state of the raft cluster under integrated storage as seen by autopilot. It shows whether autopilot thinks the cluster is healthy or not, and how many nodes could fail before the cluster becomes unhealthy ("Failure Tolerance").

State includes a list of all servers by name and IP address. Last Index can indicate how close the state on each node is to the leader.

A node can have a status of "leader", "voter", and "non-voter". Non-voters are described in <link>
```

website/content/docs/commands/operator/raft.mdx
```diff
@@ -184,3 +184,112 @@ Usage: vault operator raft snapshot restore <snapshot_file>
...
+
+- `max-trailing-logs` `(int)` - Amount of entries in the Raft Log that a server
+  can be behind before being considered unhealthy.
+
+- `min-quorum` `(int)` - Minimum number of servers allowed in a cluster before
+  autopilot can prune dead servers. This should at least be 3. Applicable only for
+  voting nodes.
+
+- `server-stabilization-time` `(string)` - Minimum amount of time a server must
+  be in a stable, healthy state before it can be added to the cluster.
```
mladlow, line 295: We do mention this in the concepts page but it's the kind of thing that bears repeating.

```suggestion
- `server-stabilization-time` `(string)` - Minimum amount of time a server must
  be in a stable, healthy state before it can be added to the cluster. Before it is
  added, its status will be "non-voter".
```

website/content/docs/commands/operator/raft.mdx
mladlow, line 192:
```suggestion
supported: `get-config`, `set-config` and `state`.

For a more detailed overview of autopilot features, see the [concepts page](/docs/concepts/integratedstorage/autopilot).
```

website/content/docs/concepts/integratedstorage/autopilot.mdx
```diff
@@ -0,0 +1,65 @@
...
+existing cluster, autopilot adds it as a non-voter instead, and waits for a
+pre-configured amount of time to monitor it's health. If the node remains to be
+healthy for the entire duration of stabilization, then that node will be
+promoted as a voter. The server stabilization period can be tuned using
+`server_stabilization_time` (see below).
+
+## Dead Server Cleanup
+
+Dead server cleanup automatically removes those nodes from the Raft, that are
+deemed failed, avoiding the manual operator intervention. This feature can be
```
mladlow, line 27:
```suggestion
Dead server cleanup automatically removes nodes deemed unhealthy from the
Raft cluster, avoiding the manual operator intervention. This feature can be
```

website/content/docs/concepts/integratedstorage/autopilot.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+Follower node health is determined by 2 factors.
+- Its ability to heartbeat to leader node at regular intervals. Tuned using
+`last_contact_threshold` (see below).
+- Its ability to keep up with data replication from the leader node. Tuned using
+`max_trailing_logs` (see below).
+
+### Default Configuration
+
+By default, Autopilot will be enabled with clusters using Vault 1.7+. Upgrade of
```
mladlow, line 46: I know this is apparent because we show the default value of `cleanup_dead_servers`, but I think it's worth calling out twice?

```suggestion
By default, Autopilot will be enabled with clusters using Vault 1.7+,
although dead server cleanup is not enabled by default. Upgrade of
```

website/data/api-navigation.js
```diff
@@ -151,7 +151,7 @@ export default [
       'step-down',
       {
         category: 'storage',
-        content: ['raft', 'raftautosnapshots'],
+        content: ['raft', 'raftautosnapshots', 'raftautopilot'],
```
mladlow, line 154: If we want them alphabetized in the sidebar, it should be like this.

```suggestion
        content: ['raft', 'raftautopilot', 'raftautosnapshots'],
```

website/content/api-docs/system/storage/raftautopilot.mdx
```diff
@@ -0,0 +1,164 @@
...
+sidebar_title: <code>/sys/storage/raft/autopilot</code>
+description: |-
+
+  The `/sys/storage/raft/autopilot` endpoints are used to manage raft clusters using autopilot with Vault's Raft storage backend.
+
+---
+
+## Get Cluster State
+
+This endpoint is used to retrieve the raft cluster state.
```
mladlow, line 13: Maybe? Or we could repeat it here.

```suggestion
This endpoint is used to retrieve the raft cluster state. See the [docs page](/docs/commands/operator/raft#autopilot-state) for a description of the output.
```

mladlow (changes requested):
I care most about the suggestions on describing the state output more, but I also think we should add a section on the overview concepts page to describe what a non-voter is and when a node might be a non-voter.

@ncabatoff do you have enough information to add that section, do you think? It could also wait until Vishal is back.

website/content/docs/commands/operator/raft.mdx
ncabatoff: In terms of active vs standbys: we strive to ensure that the user doesn't need to know, i.e. if a standby can't handle a certain request, it should forward transparently to the active node.

As to the DR cluster question: this is something we don't document elsewhere AFAIK, primarily because there's so few endpoints supported in DR that it probably makes more sense to highlight the ones that *do* work on DR rather than add a note to 99% of the APIs saying "this doesn't work on DR."

website/content/api-docs/system/storage/raftautopilot.mdx
ncabatoff: I'm confused.  Both pages (this one and the one you link) contain appropriate example output.  What do you feel is missing?

website/content/docs/commands/operator/raft.mdx
ncabatoff: How about we rephrase it instead:

Minimum amount of time a server must be in a stable, healthy state before it can become a voter.  Until that happens, it will be visible as a peer in the cluster, but as a non-voter, meaning it won't contribute to quorum.

website/content/docs/concepts/integratedstorage/autopilot.mdx
vishalnayak: We don't remove nodes that are deemed unhealthy. We only mark those as unhealthy. We only remove those that are deemed failed/dead (via a dedicated threshold).

website/content/api-docs/system/storage/raftautopilot.mdx
mladlow: We show the output but don't describe it.

ncabatoff: Ah makes sense.

ncabatoff (approved):

mladlow (approved):
Looks good, thanks @ncabatoff and @vishalnayak!

kylemac:
hey @vishalnayak - this PR changes the URL path from `.../integrated-storage` to `.../integratedstorage` -- was this intentional or unexpected? if it wasn't intentional, we probably want to go back to the original structure as there have been existing links there (i'd prefer reverting versus setting up a redirect, since it's more likely to be links to the older link than to the newer one).

PR #11171, opened 2021-03-22
fix typo "force" to "false"

samgabrail:


pmmukh:
Hi! Thanks for submitting this PR! I looked into the config parameter in the code, and it looks like "force" is a valid input here https://github.com/hashicorp/vault/blob/main/command/agent/config/config.go#L338, the field can be both a booleans and a string. Hence I'll be closing this PR out.

PR #11173, opened 2021-03-22
Update version to 1.7.0 for release, update Go version to 1.15.10

mgritter:


vendor/github.com/hashicorp/vault/sdk/version/version_base.go
```diff
@@ -9,6 +9,6 @@ var (
 	CgoEnabled bool
 
 	Version           = "1.7.0"
-	VersionPrerelease = "rc2"
+	VersionPrerelease = ""
```
mladlow, line 12: I don't think you need to change this file - the `go mod vendor` does that as well as updating the dep in go.mod and modules.txt.

mgritter: Yes, this is the change that "go mod vendor" performed as part of step 1c

mladlow (approved):

PR #10819, opened 2021-02-01
`vault operator diagnose` stub command

mgritter:
This adds the new command to the CLI, and a flag to `vault server`.
It performs some light refactoring to allow us to see the storage setup status, but startup is not completely covered yet (nor meant to be.)

As part of Vault Server:
```
mgritter@ubuntu:~/vault$ VAULT_DIAGNOSE=1 bin/vault server -diagnose= -config=../vault-enterprise/diagnose-proto/test-config.hcl
Vault v1.6.0-dev ('cbbc0d815b131c97da59692a4cb0289f075d23e2+CHANGES')
[  ok  ] Parse configuration
[  ok  ] Access storage
Error parsing Seal configuration: Put "https://test-vault.example.com:8300/v1/transit/encrypt/autounseal": dial tcp 127.0.0.1:8300: connect: connection refused
2021-02-01T15:21:07.168-0800 [INFO]  proxy environment: http_proxy= https_proxy= no_proxy=
2021-02-01T15:21:10.951-0800 [INFO]  seal.transit: unable to renew token, disabling renewal: err="Put "https://test-vault.example.com:8300/v1/auth/token/renew-self": dial tcp 127.0.0.1:8300: connect: connection refused"
```

As standalone command:
```
mgritter@ubuntu:~/vault$ VAULT_DIAGNOSE=1 bin/vault operator diagnose -config bad-config.hcl
Vault v1.6.0-dev ('cbbc0d815b131c97da59692a4cb0289f075d23e2+CHANGES')
[  ok  ] Parse configuration
[failed] Access storage
Error initializing storage of type raft: failed to create fsm: failed to open bolt file: open /var/doesnotexist/raft/vault.db: no such file or directory
```

PR #11180, opened 2021-03-23
Diagnose Stub Command Tests

HridoyRoy:
This PR has a couple tests for: https://github.com/hashicorp/vault/pull/10819/files


mgritter (approved):
This seems like an acceptable way of testing the current code.  But let me give you some thoughts about the future:

It seems likely that the details of the text will change often, so a "golden master" approach like this will break often.  One way to help that is to put the strings in an external file so that they can be more easily diff'ed and recreated.

One reasons to work on the tree structure, instead of raw output, is that the tree structure separates representation (the text) and the values, so a unit test can check the values for equality instead of a big string compare.  The tree structure's output formatting can be tested using content provided by the test.  For end-to-end testing, what I suggested in the RFC was a debug mode that dumped out the key/value pairs in the tree structure, again to avoid depending on matching text exactly.

HridoyRoy:
> This seems like an acceptable way of testing the current code. But let me give you some thoughts about the future:
> 
> It seems likely that the details of the text will change often, so a "golden master" approach like this will break often. One way to help that is to put the strings in an external file so that they can be more easily diff'ed and recreated.
> 
> One reasons to work on the tree structure, instead of raw output, is that the tree structure separates representation (the text) and the values, so a unit test can check the values for equality instead of a big string compare. The tree structure's output formatting can be tested using content provided by the test. For end-to-end testing, what I suggested in the RFC was a debug mode that dumped out the key/value pairs in the tree structure, again to avoid depending on matching text exactly.

I think that makes sense. So, once we have the output templates for each command (in external files), we can test the operator command like I'm doing now, but just compare the values that are being put into the templates, instead of the actual strings themselves. Is that fair to say?

mgritter:
Yup, I think we're on the same page:
(a) when running an end-to-end test use easily-parseable "debug mode" output, like JSON or a table of key-value pairs.
(b) separately test that the templating code works using fixed templates and variables

That way, the end-to-end tests, like this one, do not depend on the exact text that Diagnose outputs, but only on the semantic content (pass/fail, what details were collected.)  You can still run an end-to-end test by hand to see what the output looks like (and should), but including that exact match in an automated test feels very fragile and hard to work with.

sgmiller (approved):

PR #11182, opened 2021-03-23
UI/namespace bug

Monkeychip:
Only on the binary build, if the user logged in with an empty namespace (e.g. root) they would not see the top header.  Here is the error.
image
Here is the missing header
image

Note: Locally we could not reproduce, but we could on the binary build.


chelshaw (approved):
🙏

PR #11185, opened 2021-03-23
Create 1.7.0.mdx

darshanasivakumar:
Release notes for Vault 1.7


pbernal (changes requested):
we're going to need to make an update with respect to a namespace item.

mladlow:
@pbernal I'm going to go ahead and assume that's the UI thing that we no longer need to update, but let me know if that's not the case.

PR #11194, opened 2021-03-24
Fix: sign SSH key with extensions on web UI

### LINKED ISSUE #11191: SSH Sign is not including extensions in signed certificate (UI)
**Describe the bug**

When signing an SSH key via the web UI, extensions are ignored.

I think the error lies in this [line](https://github.com/hashicorp/vault/blob/403b94eb340ff1863f95d418fe95904c55723a34/ui/app/models/ssh-sign.js#L38) the JSON being sent from the UI to the server is (snippet only)

```json
{
  "public_key": ....,
  "cert_type": "user",
  "critical_options": {},
  "extension": {
    "dummy": "value"
  }
}
```

whereas the [docs](
https://www.vaultproject.io/api-docs/secret/ssh#extensions) the attribute name to be `extensions` and not `extension`

**To Reproduce**
1. Add SSH engine
2. Configure a Role to accept at least one extension
3. Sign a SSH public key, and enter an extension in the extension box eg:
 ```json
  { "dummy":"value"}
 ```
4. examine the signed key with `ssh-keygen -L` and notice the extension is not there.

**Expected behavior**

The signed key to contain the extensions.

The UI seems to be sending the wrong payload as stated before. I've testing making the same call the UI is making but replaced `extension` with `extensions` and it worked as expected.

**Environment:**

I've tested this on hashicorp cloud

* Vault Server Version (retrieve with `vault status`): Vault 1.7.0-rc2+ent
* Server Operating System/Architecture:

Don't have. used hashicorp cloud

**Additional context**

tspascoal:
closes #11191

SSH signing was sending an `extension` attribute instead of `extensions` as [documented](https://www.vaultproject.io/api-docs/secret/ssh#extensions) which meant extensions (if defined) were being ignored and not included in the signed certificate

### DIFF
--- a/ui/app/models/ssh-sign.js
+++ b/ui/app/models/ssh-sign.js
@@ -7,7 +7,7 @@ const CREATE_FIELDS = [
   'validPrincipals',
   'certType',
   'criticalOptions',
-  'extension',
+  'extensions',
   'ttl',
 ];
 
@@ -35,7 +35,7 @@ export default Model.extend({
     label: 'Key ID',
   }),
   criticalOptions: attr('object'),
-  extension: attr('object'),
+  extensions: attr('object'),
 
   leaseId: attr('string', {
     label: 'Lease ID',


Monkeychip (approved):

sgmiller:
Thanks for the fix @tspascoal!

alokbcg:
@sgmiller Thanks for the fix on the ssh key signing using UI. Can you please help if there is any plan for including this fix in the upcoming release?

PR #11201, opened 2021-03-25
Update upgrade guide and changelog with AWS Auth deprecations

kalafut:
These were missing from the initial publication.

### DIFF
--- a/website/content/docs/upgrading/upgrade-to-1.7.0.mdx
+++ b/website/content/docs/upgrading/upgrade-to-1.7.0.mdx
@@ -25,8 +25,9 @@ recommended by
 
 AWS Auth concepts and endpoints that use the "whitelist" and "blacklist" terms
 have been updated to more inclusive language (e.g. `/auth/aws/identity-whitelist` has been
-updated to`/auth/aws/identity-accesslist`). The legacy endpoint names are considered **deprecated**.
-The old and new endpoints are aliases, sharing the same underlying data. The complete list of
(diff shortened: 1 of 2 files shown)


website/content/docs/upgrading/upgrade-to-1.7.0.mdx
```diff
@@ -14,13 +14,21 @@ for Vault 1.7.0 compared to 1.6. Please read it carefully.
...
 [NIST SP 800-38D](https://csrc.nist.gov/publications/detail/sp/800-38d/final).
 
+## AWS Auth Endpoint Changes and Deprecations
+
+AWS Auth concepts and endpoints that use the "whitelist" and "blacklist" terms
+have been updated to more inclusive language (e.g. `/auth/aws/identity-whitelist` has been
+updated to`/auth/aws/identity-accesslist`). The legacy endpoint names are considered **deprecated**.
+The old and new endpoints are aliases, sharing the same underlying data. The complete list of
+endpoint changes is available in the [AWS Auth API docs](/api-docs/auth/aws#deprecations-effective-in-vault-1-7).
+
```
mladlow, line 31: The question we'll get is "will these still work" and "when will they stop working". I think adding "Though deprecated, the old aliases will continue to function but will be removed in a future release (not before Vault 1.9)".

kalafut: Updated

mladlow (approved):

PR #11208, opened 2021-03-25
Replace go-bindata-assetfs build dependency with native go:embed

catsby:
This PR replaces Vault's use of [elazarl/go-bindata-assetfs][go-bindata-assetfs] in building the UI with Go's native [Embed package][embed]. 

[go-bindata-assetfs]: https://github.com/elazarl/go-bindata-assetfs
[embed]: https://golang.org/pkg/embed/


kalafut:
Awesome! The bootstrap process can probably be simplified too (https://github.com/hashicorp/vault/blob/master/tools/tools.go#L25-L29)

catsby:
> Awesome! The bootstrap process can probably be simplified too (https://github.com/hashicorp/vault/blob/master/tools/tools.go#L25-L29)

Indeed, I did that in cff9f7a shortly after you mentioned it. Thanks!

packages-oss.yml
```diff
@@ -241,7 +241,7 @@ layers:
   - name: build-static-assets
     source-include: Makefile
     dockerfile: |-
-      RUN go-bindata-assetfs -o bindata_assetfs.go -pkg http -prefix pkg -modtime 1480000000 -tags ui ./pkg/web_ui/...
+      RUN go-bindata-assetfs -o bindata_assetfs.go -pkg http -prefix pkg -modtime 1480000000 -tags ui ./http/web_ui/...
```
briankassouf, line 244: Do we still need to run go-bindata-assetfs with this new system?

catsby: No. I believe this PR needs some additional polishing, it was time-boxed and functional when I opened it. I had planned on reviewing and continuing some cleanup depending on interest.

briankassouf:
I think there are a few more references in the Makefile to remove: https://github.com/hashicorp/vault/blob/1b96d2c44eb17451b5708a8e9d19602d408824d2/Makefile#L11-L12

catsby:
Hey @briankassouf I believe this is ready for another review!

packages-oss.yml
```diff
@@ -103,7 +103,6 @@ build-command: VERSION_PKG_PATH=github.com/hashicorp/vault/vendor/github.com/has
 # is a useful data source for building CI/CD pipelines.
 packages:
   - inputs: { GOOS: darwin, GOARCH: amd64 }
-  - inputs: { GOOS: darwin, GOARCH: arm64 }
```
briankassouf, line 106: Accidental deletion?

packages-oss.lock/pkgs.yml
```diff
@@ -1,15 +1,15 @@
...
 packagespec-version: 0.2.2
 cache-version: 10
 packages:
-  - packagespecid: a8bdeecb59f5f37dc83943709455b9fd277df5f8
+  - packagespecid: b6c12ff70f70cf013c61fafabe18ce98025cb6de
     inputs:
       BINARY_NAME: vault
       CGO_ENABLED: "0"
-      GO_VERSION: 1.16.2
+      GO_VERSION: 1.15.7
```
briankassouf, line 12: It looks like some things in this file were reverted. Maybe a bad merge caused this?

briankassouf (commented):
A few things, but i'd like @kalafut to take a look when he gets a chance too

packages-oss.lock/pkgs.yml
catsby: I think this is from a bad merge, yes. I re-re-re-ran `make packages` in 61e63b4d4 to fix this.

packages-oss.yml
catsby: I suppose this is from a bad merge as well? Fixed in 861abee11 and followed by yet another `make packages`

ncabatoff:
This looks great!  I think you missed scripts/cross/Dockerfile, but that looks stale anyway so maybe we should just get rid of it.

My only question/concern is: why move from pkg to http folders?  Probably not an issue, but it feels like an unnecessary change that could conceivably bite us.

ncabatoff (approved):

packages-oss.yml
catsby: These parts should be cleaned up now

briankassouf (approved):
Looks good! I tested it locally, and it works well! We may want to try building with package spec too!

catsby:
Hey @ncabatoff - 

> My only question/concern is: why move from pkg to http folders? Probably not an issue, but it feels like an unnecessary change that could conceivably bite us.

Some background: the previous bindata-assetFS approach converted the assets into `[]bytes` slices and wrote them to a file, which is how the ui files became included in the binary. The embed approach includes the files as-is, and so they need to be in a folder that the go file invoking them can reach them, and are referenced from the location `//go:embed` is referenced. The documentation mentions forbidding the use of `.` and `..` in paths:

> Patterns may not contain ‘.’ or ‘..’ or empty path elements

(from https://golang.org/pkg/embed/#hdr-Directives)

I found that I couldn't reference a directory _above_ the directory I was in, so I couldn't do this:

```go
//go:embed ../pkg/web_ui/*
var content embed.FS
```

As a result, I had to move things into the `http/` folder. I used the `web_ui` subfolder so that the generated files wouldn't need to be tracked in git. 

We could alternatively start to track them if we want, since now they'll be included in the binaries with the UI instead of converted by bindata-assetFS. What do you think?

kalafut:
cc: @chelshaw @Monkeychip for awareness

scripts/assetcheck.sh
```diff
@@ -1,11 +1,7 @@
 #!/usr/bin/env bash
 
-if [[ ! -e http/bindata_assetfs.go ]]
+if [[ ! -e http/web_ui/index.html ]]
 then
   printf "Compiled UI assets not found. They can be built with: make static-dist\n\n"
-else
```
kalafut, line 6: Why was this else block removed?

catsby: It went on to specifically reference the generated bindata file:

https://github.com/hashicorp/vault/blob/6b2f99cd6e1c0bb654c72ac7459f7956a429f927/scripts/assetcheck.sh#L6-L11

I didn't find it useful. We could restore it and test `index.html` I suppose?

kalafut: Sorry for the delay. But yes, having a check against a representative generated file like `index.html` would still be useful. We added this after a number of cases of people digging into and reporting some odd UI-related behavior, only to find they were running very out of date UI code.

.circleci/config/@build-release.yml.back
```diff
@@ -0,0 +1,2049 @@
+version: "2.1"
```
briankassouf, line 1: Did you intend to commit this file? There are a handful of unresolved merge conflicts in it

briankassouf (approved):
Just one comment, but looks good otherwise!

kalafut (approved):

.circleci/config/@build-release.yml.back
hghaf099: ooops, nope!

heppu:
This actually breaks vendoring the vault as a go module since those ui files are not part of the repo.
```sh
$ go get -d github.com/hashicorp/vault@main
go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.
$ go mod vendor
go mod vendor: pattern web_ui/*: no matching files found
```

pmmukh:
@heppu We don’t really support vendoring Vault as a go module, as that can often run into conflicts when importing both Vault, along with the `api` and `sdk` modules. We understand you may have your reasons to do this, but we don’t offer much in the way of guarantees that this will work.

heppu:
Yes I have encountered those issues previously but then again those are issues that user can always resolve with replace rules and some detective work on github. This how ever is something that goes straight against the Go's tooling and hence there is no nice way around. I guess what I'm asking here is that is there something preventing us from adding those web ui files in the repository?

pmmukh:
@heppu if you'd like to raise a PR fixing this issue, please feel free to do so, and we'll be happy to review it. I'm not sure if there's something preventing us from including the web_ui files, I think that was a decision we'd be ok revisiting if there's a PR making a case for that to be changed.

heppu:
@pmmukh Thanks for quick responses, I will do that!

heppu:
@pmmukh I have now created the PR #12523 if you or some one else has time to take a look at it.

PR #11213, opened 2021-03-26
UI: Fix status menu bug

Monkeychip:
Fix an error where the status menu does not show on the initial login.  The workaround was to refresh.  This fix gets us back to the expected behavior.


chelshaw (approved):
such a small change, such destructive behavior 🐛 🪓

PR #11216, opened 2021-03-26
certutil: select appropriate hash algorithm for ECDSA signature

### LINKED ISSUE #11006: pki: select appropriate hash algorithm for ecdsa signed certificates

Currently, when issuing certificates, vault always sets the signature algorithm to `ECDSAWithSHA256` if an ecdsa key is used.
The key length is not taken into consideration:

https://github.com/hashicorp/vault/blob/056dd26fa815affcdb5e029cdeeadbe1eb5525c0/sdk/helper/certutil/helpers.go#L602-L603
https://github.com/hashicorp/vault/blob/056dd26fa815affcdb5e029cdeeadbe1eb5525c0/sdk/helper/certutil/helpers.go#L622-L623

The [BSI recommendations (section 4.2)](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TR03111/BSI-TR-03111_V-2-0_pdf.pdf?__blob=publicationFile&v=2) are to use a hash size of the same length as the key size.

The [Mozilla root story policy (section 5.1.2)](https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/policy/) even forbids `ECDSAWithSHA256` for a key of length 384.

Instead of setting the `SignatureAlgorithm` explicitly in the certificate template, vault should leave the `SignatureAlgorithm` unspecified. The `x509` library will select the appropriate algorithm:

https://github.com/golang/go/blob/37ca84a9cd6a8a76dfe91263a17d2b92b17a24b3/src/crypto/x509/x509.go#L2024-L2038

Alternatively, the `pki` engine could be extended to take the hash algorithm as input, similar to [transit](https://www.vaultproject.io/api-docs/secret/transit#hash_algorithm)

N/A

I would like to be able to issue certificates with a `p384` or a `p521` key and use the appropriate hash algorithm. 

**Additional context**

## BSI:
image

## Mozilla:
image

Oncilla:
Select the appropriate signature algorithm for certificates signed
with an ECDSA private key.

The algorithm is selected based on the curve:

- P-256 -> x509.ECDSAWithSHA256
- P-384 -> x509.ECDSAWithSHA384
- P-521 -> x509.ECDSAWithSHA512
- Other -> x509.ECDSAWithSHA256

fixes #11006

### DIFF
--- a/sdk/helper/certutil/helpers.go
+++ b/sdk/helper/certutil/helpers.go
@@ -600,7 +600,7 @@ func createCertificate(data *CreationBundle, randReader io.Reader) (*ParsedCertB
 		case RSAPrivateKey:
 			certTemplate.SignatureAlgorithm = x509.SHA256WithRSA
 		case ECPrivateKey:
-			certTemplate.SignatureAlgorithm = x509.ECDSAWithSHA256
+			certTemplate.SignatureAlgorithm = selectSignatureAlgorithmForECDSA(data.SigningBundle.PrivateKey.Public())
 		}
 
 		caCert := data.SigningBundle.Certificate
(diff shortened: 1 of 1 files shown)


oncilla:
@sgmiller should be fixed now

PR #11218, opened 2021-03-26
pki: calculate Subject Key Identifier according to RFC 5280

### LINKED ISSUE #11153: pki: calculate Subject Key Identifier according to RFC 5280

Currently, when issuing certificates, Vault sets the subject key identifier as the sha1 over the public key encoded as a SubjectPublicKeyInfo structure. This deviates from the standard way of calculating the subject key identifier as described in [RFC 5280, Section 4.2.1.2](https://tools.ietf.org/html/rfc5280#section-4.2.1.2), and implemented by OpenSSL and the likes.

RFC 5280 describes the calculation as:
> (1) The keyIdentifier is composed of the 160-bit SHA-1 hash of the
           value of the BIT STRING subjectPublicKey (excluding the tag,
           length, and number of unused bits).

I would like Vault to calculate the subject key identifier according to RFC 5280.

none

I want to find certificates that match my private key in an easy way. This can be done by calculating the subject key identifier of the public key, and searching for certificates that contain that identifier. I would prefer if I do not have to distinguish whether the certificates were created via OpenSSL or Vault.

**Additional context**

none

oncilla:
Calculate the Subject Key Identifier as suggested in RFC 5280, Section 4.2.1.2

> (1) The keyIdentifier is composed of the 160-bit SHA-1 hash of the
value of the BIT STRING subjectPublicKey (excluding the tag,
length, and number of unused bits).

fixes #11153

---
This change is [Reviewable](https://reviewable.io/reviews/hashicorp/vault/11218)

### DIFF
--- a/sdk/helper/certutil/helpers.go
+++ b/sdk/helper/certutil/helpers.go
@@ -69,21 +69,42 @@ func ParseHexFormatted(in, sep string) []byte {
 	return ret.Bytes()
 }
 
-// GetSubjKeyID returns the subject key ID, e.g. the SHA1 sum
-// of the marshaled public key
+// GetSubjKeyID returns the subject key ID. The computed ID is the SHA-1 hash of
+// the marshaled public key according to
+// https://tools.ietf.org/html/rfc5280#section-4.2.1.2 (1)
 func GetSubjKeyID(privateKey crypto.Signer) ([]byte, error) {
 	if privateKey == nil {
 		return nil, errutil.InternalError{Err: "passed-in private key is nil"}
 	}
+	return getSubjectKeyID(privateKey.Public())
+}
 
-	marshaledKey, err := x509.MarshalPKIXPublicKey(privateKey.Public())
-	if err != nil {
-		return nil, errutil.InternalError{Err: fmt.Sprintf("error marshalling public key: %s", err)}
+func getSubjectKeyID(pub interface{}) ([]byte, error) {
+	var publicKeyBytes []byte
+	switch pub := pub.(type) {
+	case *rsa.PublicKey:
+		type pkcs1PublicKey struct {
+			N *big.Int
+			E int
+		}
+
+		var err error
+		publicKeyBytes, err = asn1.Marshal(pkcs1PublicKey{
+			N: pub.N,
+			E: pub.E,
+		})
+		if err != nil {
+			return nil, errutil.InternalError{Err: fmt.Sprintf("error marshalling public key: %s", err)}
+		}
+	case *ecdsa.PublicKey:
+		publicKeyBytes = elliptic.Marshal(pub.Curve, pub.X, pub.Y)
+	case ed25519.PublicKey:
+		publicKeyBytes = pub
+	default:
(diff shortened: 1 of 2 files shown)


Oncilla:
Note that this change could potentially be considered a breaking change.

cipherboy:
Hey @oncilla Do you mind rebasing this? I've begun taking a look at this PR. :-)

oncilla:
@cipherboy Rebased and added the changelog entry. 

I wonder if this classifies as change, or as improvement. Are there any docs on this?

cipherboy:
Hi @oncilla, no I don't think we have any docs on it. I personally feel (and, from historical conversations with the team, think we're generally aligned) that this is a bug more than anything. Dogtag, Let's Encrypt's Boulder, and EJBCA (via BouncyCastle) -- to name a few I looked at -- all appear to use this method for SKID calculation. My 2c.

sdk/helper/certutil/helpers.go
```diff
@@ -69,21 +69,42 @@ func ParseHexFormatted(in, sep string) []byte {
...
+	switch pub := pub.(type) {
+	case *rsa.PublicKey:
+		type pkcs1PublicKey struct {
+			N *big.Int
+			E int
+		}
 
-	return subjKeyID[:], nil
+		var err error
+		publicKeyBytes, err = asn1.Marshal(pkcs1PublicKey{
```
cipherboy, line 92: At least from my playing in the Playground, it [should be OK](https://play.golang.org/p/ZwfM_tKizsi) to replace this with:

```go
publicKeyBytes, err = asn1.Marshal(*pub)
```

since the field ordering is the same and we can avoid needing to duplicate the struct. In [regular Go](https://github.com/golang/go/blob/master/src/crypto/rsa/rsa.go#L41-L45) the struct matches this local struct. In [BoringCrypto fork of Go](https://github.com/golang/go/blob/dev.boringcrypto/src/crypto/rsa/rsa.go#L46-L52), it merely contains an unsafe pointer, which because it is private, shouldn't be accessible to `asn1.Marshal`.

Thoughts? I agree this is more explicit and the less risky approach (in case `rsa.PublicKey` changes public fields in the future), but I'm expecting that to be unlikely. 

Perhaps @sgmiller could weigh in as well.

sgmiller: I don't have a strong opinion, but I'd always trade more safety for terseness.

sgmiller: Re: release note, I definitely wouldn't call it a bug, since the original implementation isn't in violation of the standard.

oncilla: I agree, I wouldn't classify it as a bug either. Technically, there is nothing wrong with the existing computation.

cipherboy: Let's leave as-is then.

cipherboy (approved):

cipherboy:
Looks good; thanks for this @oncilla and sorry about the delay :-)

PR #11221, opened 2021-03-29
Use a 15s instead of 10s timeout in TestAgent_Template_Retry

ncabatoff:
The `default` subtest takes 9s on my laptop, so pad that for CI.

Example failure: https://app.circleci.com/pipelines/github/hashicorp/vault/16245/workflows/203601da-94bf-4ec2-be25-d8e6dd4315b8/jobs/174151


catsby (approved):
👍

PR #11223, opened 2021-03-29
Replace Reference to Version 1.6.4 with 1.7

in0rdr:
This replaces a reference to version 1.6.4 with the reference to the released version 1.7:

https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#170

As far as I could see the endpoint was added in version 1.7.

Not sure if version 1.6.4. was ever released. At least I was unable to find a build here:

https://releases.hashicorp.com/vault/


sgmiller:
You're partially correct.  That endpoint was introduced in Vault 1.6, but 1.6.4 is unreleased so that line is incorrect.

sgmiller (approved):

PR #11230, opened 2021-03-29
Hide fields that are empty in configuration

arnav28:
Updated the existing **is-empty-object** to **is-empty-value** helper to check for an empty object, string,
null or undefined. Used the helper to hide fields that are empty from secrets and auth configuration views.

Previous:
Screen Shot 2021-03-31 at 9 53 02 AM

New:
Screen Shot 2021-03-31 at 9 51 18 AM


ui/app/helpers/is-empty-value.js
```diff
@@ -0,0 +1,8 @@
+import { helper } from '@ember/component/helper';
+
+export default helper(function isEmptyValue([value] /*, hash*/) {
+  if (typeof value === 'object' && value !== null) {
+    return Object.keys(value).length === 0;
+  }
+  return value == null || value === '';
```
Monkeychip, line 7: Nice work on a use case for double equals null here.

ui/tests/integration/helpers/is-empty-value-test.js
```diff
@@ -0,0 +1,51 @@
+import { module, test } from 'qunit';
```
Monkeychip, line 1: Nice work on making this test more robust!

Monkeychip (approved):
This looks good to me! Nice work.

One comment for next time (doesn't need to change for this PR) we prepend our UI branch with `ui`.  So for example, in this case it would be `ui/empty-fields`.  The reason is that by doing this circle ci scans for the ui/ and only runs tests relevant/required for a ui change.

PR #11231, opened 2021-03-29
UI/database mssql

chelshaw:
Adds support for MSSQL database secrets engine, adds username template field to connections, and streamlines the process for adding a new database plugins. All you need to do to add a new database type is update two files: 
- [ ] `models/database/connection.js` add to `AVAILABLE_PLUGIN_TYPES` array with plugin's connection attributes
- [ ] `components/database-role-setting-form` update `STATEMENT_FIELDS` if the plugin does not use the default role fields

create-connection
create-role

Testing these changes require a bit of setup, so here are the steps I took to make sure we could connect to a MSSQL database through the UI:

1. Run MSSQL Docker: `docker run -e 'ACCEPT_EULA=Y' -e 'MSSQL_SA_PASSWORD=<YourStrong!Passw0rd>' -p 1401:1433 -d mcr.microsoft.com/mssql/server:2017-latest-ubuntu`
2. Enable database secret engine and create a new connection, with the following details:
```
connection_url: sqlserver://{{username}}:{{password}}@<your.ip.address>:1401
username: SA
password: <YourStrong!Passw0rd>
```
3. Create dynamic role from the new connection, and make sure it has at least one creation statement (examples below)
```
CREATE LOGIN [{{name}}] WITH PASSWORD = '{{password}}';
CREATE USER [{{name}}] FOR LOGIN [{{name}}];
GRANT SELECT ON SCHEMA::dbo TO [{{name}}]
```
4. You should now be able to generate credentials with your new role


ui/app/models/database/connection.js
```diff
@@ -9,66 +9,134 @@ const AVAILABLE_PLUGIN_TYPES = [
...
+      { attr: 'username_template', group: 'pluginConfig' },
+      { attr: 'max_open_connections', group: 'pluginConfig' },
+      { attr: 'max_idle_connections', group: 'pluginConfig' },
+      { attr: 'max_connection_lifetime', group: 'pluginConfig' },
     ],
   },
 ];
 
+/**
+ * fieldsToGroups helper fn
```
Monkeychip, line 45: any chance this would be a good candidate for a helper? e.g. could this be used elsewhere, or is this pretty specific to this model?

ui/app/templates/components/database-role-edit.hbs
```diff
@@ -95,7 +95,7 @@
         @roleType={{@model.type}}
         @model={{@model}}
         @mode={{@mode}}
-        @dbType={{this.databaseType}}
+        @dbType={{await this.databaseType}}
```
Monkeychip, line 98: I'm all about adding this helper with ember-promise-helpers. I think you were there for the ui-guild meeting where we discussed the argument against having promises in templates but I think we have enough use cases for it. Happy to see it implemented!

Monkeychip (changes requested):
Looks great, aside for some test failures that are component field-value test failures. I first noticed they said PKI and KMIP, but I don't think they're failing because they're flaky. I think it might be something with the form-field changes that were made, but that's only a guess.

ui/app/models/database/connection.js
chelshaw: Not exactly, since it's pretty specific to the shape of the object defined above, `AVAILABLE_PLUGIN_TYPES`. If we use that pattern in the future, we could definitely move it to a shared helper

Monkeychip (approved):

PR #11234, opened 2021-03-30
Update CHANGELOG.md

calvn:


mladlow:
Why do we need two commits? Also you can just directly merge to the changelog I think.

calvn:
11114 is the main PR/commit that made the bulk of the changes due to the upgrade, and 11173 (I believe) is a smaller PR that made some changes that were left out by the initial PR. Just  referencing both in the changelog in case we need to trace the go upgrade back to the appropriate PRs.

PR #11243, opened 2021-03-31
UI: Skip flaky PKI test

chelshaw:
The PKI tests have been flaky since our ember version upgrade; we will most likely need to rewrite them, but hopefully for now to unblock CI this will do


Monkeychip (approved):
I have some ideas on maybe how to help with this test.  I'll spend some time playing with these, but I don't want to hold anyone up.  Thanks for tackling this.

PR #11247, opened 2021-03-31
Autopilot: Return leader info via delegate

vishalnayak:
`go.mod` changes needs to be removed after the upstream PR for the lib is merged. https://github.com/hashicorp/raft-autopilot/pull/7

Update: Upstream PR, merged. `go.mod` updated and the dependency pulled.


vishalnayak:
I did a `go get -u github.com/hashicorp/raft-autopilot` and `go mod vendor` on the change.

ncabatoff (approved):

PR #11249, opened 2021-03-31
Initial Diagnose Command for TLS and Listener [VAULT-1896, VAULT-1899]

HridoyRoy:
This PR contains the framework for running Diagnose to check TLS and Listener functionality. Some of the test methods are stubs to fill in later, and there's an ActiveProbe stub method for use in phase 2 of Diagnose. 

I also added a helper method in core to fetch the unsanitized vault config, as the sanitized config returns a map[string]interface and not a struct.


command/server.go
```diff
@@ -871,6 +871,94 @@ func (c *ServerCommand) setupStorage(config *server.Config) (physical.Backend, e
...
+	if c.reloadFuncsLock != nil {
+		c.reloadFuncsLock.Unlock()
+	}
+	if !disableClustering {
+		if c.logger.IsDebug() {
+			c.logger.Debug("cluster listener addresses synthesized", "cluster_addresses", clusterAddrs)
+		}
+	}
+	// Question: Do I need to return the address of the listeners?
+	return 0, lns, clusterAddrs
```
HridoyRoy, line 959: // Question: Do I need to return a pointer to the listeners to properly close the listeners in the calling function?

vault/core_util.go
```diff
@@ -55,6 +56,27 @@ func (c *Core) setupReplicationResolverHandler() error {
 	return nil
 }
 
+// GetCoreConfigInternal returns the server configuration
+// in struct format. This function should only be used internally
+// to retrieve the core config.
+func (c *Core) GetCoreConfigInternal() *server.Config {
```
HridoyRoy, line 62: I added this initially because I was parsing listeners from core. I removed that functionality from Diagnose, but I think this helper method should still go in for later use.

vault/diagnose/diagnose_tls.go
```diff
@@ -0,0 +1,47 @@
...
+			KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		})
+	}
+	return nil
+}
+
+// ServerListenerActiveProbe attempts to use TLS information to set up a TLS server with each listener
+// and generate a successful request through to the server.
+// TODO
+func ServerListenerActiveProbe(core *vault.Core) {}
```
HridoyRoy, line 47: Kept the activeProbe method as a stub, just as a reminder that we should add this into phase 2. I'll make a Jira and link here as well.

vault/diagnose/diagnose_tls_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	err := TLSConfigChecks(listeners)
+	if err == nil {
+		t.Errorf("TLS Config check on fake certificate should fail")
+	}
+	if err.Error() != "asn1: syntax error: trailing data" {
+		t.Errorf("Bad error message: %s", err.Error())
+	}
+}
+
```
HridoyRoy, line 134: Everything below this are stub tests. All this functionality should exist, but I'm having trouble with creating appropriate certs to check these. For example, I'm not able to get an expired TLS cert for TestTLSExpiredCert. I was hoping we could leave these as stubs for now (so we have the Diagnose framework in), but I'm happy to fill these in in this PR as well!

command/server/test-fixtures/nostore_config.hcl
```diff
@@ -4,8 +4,8 @@ disable_mlock = true
 ui = true
 
 listener "tcp" {
-    address = "127.0.0.1:443"
```
HridoyRoy, line 7: 443 is a privileged port, which is why I needed something >= 1024

command/server/test-fixtures/tls_config_ok.hcl
```diff
@@ -0,0 +1,48 @@
+disable_cache = true
+disable_mlock = true
+
+ui = true
+
+listener "tcp" {
+    address = "127.0.0.1:1025"
```
HridoyRoy, line 7: If I change this to 1024, I get "port already in use" when running the test. This is slightly concerning, since it may point to the fact that the listener may not be closed in the previous runs, but I'm hoping it's just because of the parallelism when running tests.

command/operator_diagnose_test.go
```diff
@@ -21,7 +21,6 @@ func testOperatorDiagnoseCommand(tb testing.TB) (*cli.MockUi, *OperatorDiagnoseC
 }
 
 func TestOperatorDiagnoseCommand_Run(t *testing.T) {
-	t.Parallel()
```
HridoyRoy, line 24: These parallels cause listeners to collide with one another if they share ports (eg 2 tests using port 1024 will cause a race failure in which either test will randomly fail with a "port already in use" error). So I removed them, allowing listeners to be cleaned up between runs via the defer listener.close block.

sgmiller: Could you assign them each a different incrementing port #?

sgmiller: Oh, possibly not since we don't control the port #s

command/server.go
sgmiller: I would imagine if the listener contains a bound net.Listener, it's pointer would be copied by value.  You can test this though with nestat -tl and a breakpoint before and after you close them.

vault/core_util.go
sgmiller: Does this and the next function repeatedly load the config?  Or is rawConfig set only once?

command/server.go
```diff
@@ -871,6 +871,93 @@ func (c *ServerCommand) setupStorage(config *server.Config) (physical.Backend, e
...
+
+	if c.reloadFuncsLock != nil {
+		c.reloadFuncsLock.Lock()
+	}
+
+	for i, lnConfig := range config.Listeners {
+		ln, props, reloadFunc, err := server.NewListener(lnConfig, c.gatedWriter, c.UI)
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error initializing listener of type %s: %s", lnConfig.Type, err))
+			return 1, nil, nil
```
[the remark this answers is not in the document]
sgmiller: It looks like it's like an exit code.  Perhaps this function shouldn't use c.UI.Error, but return an error, and the Run form would then display it via c.UI.Error, while the diagnose tests wouldn't need to do so.

command/server.go
```diff
@@ -871,6 +871,93 @@ func (c *ServerCommand) setupStorage(config *server.Config) (physical.Backend, e
 	return backend, nil
 }
 
+func (c *ServerCommand) InitListeners(config *server.Config, disableClustering bool, infoKeys *[]string, info *map[string]string) (int, []listenerutil.Listener, []*net.TCPAddr) {
+	clusterAddrs := []*net.TCPAddr{}
+
+	// Initialize the listeners
+	lns := make([]listenerutil.Listener, 0, len(config.Listeners))
+
+	if c.reloadFuncsLock != nil {
```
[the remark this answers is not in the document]
HridoyRoy: Good point, I can remove the nil check.

vault/diagnose/diagnose_tls.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"crypto/tls"
+	"crypto/x509"
+
+	"github.com/hashicorp/vault/internalshared/listenerutil"
+	"github.com/hashicorp/vault/vault"
+)
+
+// TLSConfigChecks contains manual error checks against the TLS configuration
+func TLSConfigChecks(listeners []listenerutil.Listener) error {
```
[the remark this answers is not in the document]
HridoyRoy: I can add a helper for 1 listener, but I think we shouldn't separate out the TLS check for listeners from the ConfigUtil.Listener check. The reason is that most of the TLS verification is done for us via helper methods in the x509 library which rely on creating a TLS config from all the different TLS and http related fields in the listener. 

I don't necessarily want to rebuild that -- instead, I want to use these helper methods (as these are what are being used in the server run command).

command/operator_diagnose_test.go
```diff
@@ -45,16 +44,22 @@ func TestOperatorDiagnoseCommand_Run(t *testing.T) {
 			[]string{"Parse configuration\n\x1b[F\x1b[32m[  ok  ]\x1b[0m Parse configuration\n[      ] Access storage\n\x1b[F\x1b[31m[failed]\x1b[0m Access storage\nA storage backend must be specified\n"},
 			1,
 		},
+		{
+			"diagnose_listener_config_ok",
+			[]string{
+				"-config", "./server/test-fixtures/tls_config_ok.hcl",
+			},
+			[]string{"Parse configuration\n\x1b[F\x1b[32m[  ok  ]\x1b[0m Parse configuration\n[      ] Access storage\n\x1b[F\x1b[32m[  ok  ]\x1b[0m Access storage\n"},
```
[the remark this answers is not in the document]
HridoyRoy: Definitely agree. Do you want that in this PR, or can I do that in a separate PR and create a jira for it? I wanted to package that into the task to have Diagnose use a tree structure of templates.

command/operator_diagnose_test.go
HridoyRoy: We do control the port numbers, and I could assign different port numbers (initially I did that to bypass this problem), but I changed it to use the same port and remove the parallel. I don't like the idea that these tests are going to open a bunch of listeners at the same time on different ports, and removing parallel fixes that issue. 

Also it seems like using Parallel() in our tests is an anti-pattern that we want to avoid.

vault/core_util.go
HridoyRoy: Only once.

command/server.go
HridoyRoy: Oh, ok -- let me try that out. Thanks!

command/operator_diagnose_test.go
sgmiller: Parallel is actually good so long as it doesn't change the semantics of the tests, helps CI run more quickly.

command/operator_diagnose_test.go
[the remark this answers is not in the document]
sgmiller: Doing it later is fine.  Just want to have something type safe and machine oriented in the future.

vault/core_util.go
sgmiller: Ok, np then.

vault/diagnose/diagnose_tls.go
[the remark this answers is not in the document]
sgmiller: Okay, makes sense

command/operator_diagnose_test.go
HridoyRoy: I spoke too soon! Parallel and different port numbers may be the way to go here after all!

command/operator_diagnose_test.go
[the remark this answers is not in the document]
HridoyRoy: Awesome! Jira: https://hashicorp.atlassian.net/browse/VAULT-1972

vault/diagnose/diagnose_tls.go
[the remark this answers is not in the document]
HridoyRoy: Actually upon closer inspection, I think you're absolutely right. A lot of the x509 helper methods only need filePaths, and construct a TLSConfig object from those paths. So, I don't necessarily need the entire listener.Config. I separated out a helper function that just takes generic file paths to the TLS info (just the key and cert for now), and performs generic checks on that.

command/server.go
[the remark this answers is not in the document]
HridoyRoy: Done!

command/server.go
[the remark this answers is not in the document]
HridoyRoy: Done!

command/server.go
HridoyRoy: Looks ok!

command/operator_diagnose.go
```diff
@@ -17,9 +21,15 @@ var _ cli.CommandAutocomplete = (*OperatorDiagnoseCommand)(nil)
...
-	flagSkips   []string
-	flagConfigs []string
+	flagDebug    bool
+	flagSkips    []string
+	flagConfigs  []string
+	cleanupGuard sync.Once
+
+	reloadFuncsLock *sync.RWMutex
+	reloadFuncs     *map[string][]reloadutil.ReloadFunc
+	startedCh       chan (struct{}) // for tests
```
HridoyRoy, line 31: Remove parenthesis

vault/diagnose/diagnose_tls_test.go
```diff
@@ -0,0 +1,146 @@
+package diagnose
```
HridoyRoy, line 1: change filename to tls.go instead of diagnose_tls.go

vault/diagnose/diagnose_tls.go
HridoyRoy: Throw an error on the stub in case it gets called.

vault/diagnose/diagnose_tls_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"testing"
+
+	"github.com/hashicorp/vault/command/server"
+	"github.com/hashicorp/vault/internalshared/configutil"
+	"github.com/hashicorp/vault/internalshared/listenerutil"
+	"github.com/hashicorp/vault/vault"
+)
+
+func setup(t *testing.T) *vault.Core {
+	serverConf := &server.Config{
```
swayne275, line 13: it doesn't look like this logs or errors, so we don't need `t.Helper()`?

vault/diagnose/diagnose_tls_test.go
```diff
@@ -0,0 +1,146 @@
...
+				TLSClientCAFile:               "./../../api/test-fixtures/root/rootcacert.pem",
+				TLSMinVersion:                 "0",
+				TLSRequireAndVerifyClientCert: true,
+				TLSDisableClientCerts:         false,
+			},
+		},
+	}
+	err := ListenerChecks(listeners)
+	if err != nil {
+		t.Fail()
```
swayne275, line 77: i think it would be helpful to fail with the error here. maybe a `t.Errorf()` with the error included?

vault/diagnose/diagnose_tls_test.go
```diff
@@ -0,0 +1,146 @@
...
+				TLSRequireAndVerifyClientCert: true,
+				TLSDisableClientCerts:         false,
+			},
+		},
+	}
+	err := ListenerChecks(listeners)
+	if err == nil {
+		t.Errorf("TLS Config check on fake certificate should fail")
+	}
+	if err.Error() != "tls: failed to find any PEM data in certificate input" {
```
swayne275, line 101: is this error "type" included in some library we're using under the hood, or does it need to be a string?

vault/diagnose/diagnose_tls_test.go
```diff
@@ -0,0 +1,146 @@
...
+				TLSRequireAndVerifyClientCert: true,
+				TLSDisableClientCerts:         false,
+			},
+		},
+	}
+	err := ListenerChecks(listeners)
+	if err == nil {
+		t.Errorf("TLS Config check on fake certificate should fail")
+	}
+	if err.Error() != "asn1: syntax error: trailing data" {
```
swayne275, line 130: does this error string exist in some library, or must it be hard coded this way?

vault/core_util.go
swayne275: what do you mean by "should only be used internally"? to me that implies internal to the package, which would mean it might be better off as a private/unexported function?

command/operator_diagnose.go
```diff
@@ -110,7 +120,7 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
 	}
 
 	c.UI.Output(version.GetVersion().FullVersionNumber(true))
-
+	rloadFuncs := make(map[string][]reloadutil.ReloadFunc, 5)
```
swayne275, line 123: does the `5` in your `make()` do anything here? my understanding is you don't include a length for a map. i was wondering if maybe there was some special case where that applies to the length of the slice value, but i tried it out in go playground and that doesn't seem to be the case.

command/operator_diagnose.go
```diff
@@ -140,6 +152,61 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
 		return 1
 	}
 
+	// Check Listener Information
+	// TODO: Run Diagnose checks on the actual net.Listeners
+
+	disableClustering := config.HAStorage.DisableClustering
+	infoKeys := make([]string, 0, 10)
```
swayne275, line 159: why do we specify a capacity here?

command/operator_diagnose.go
```diff
@@ -140,6 +152,61 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
...
+	// Make sure we close all listeners from this point on
+	listenerCloseFunc := func() {
+		for _, ln := range lns {
+			ln.Listener.Close()
+		}
+	}
+
+	defer c.cleanupGuard.Do(listenerCloseFunc)
+
+	sanitizedListeners := make([]listenerutil.Listener, 0, len(config.Listeners))
```
swayne275, line 178: how helpful is it to specify a capacity here? it appears that the required capacity can be lower than `len(config.Listeners)` (if `ln.Config.TLSDisable`, for example)

command/operator_diagnose_test.go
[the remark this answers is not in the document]
swayne275: cool

command/server.go
```diff
@@ -871,6 +871,90 @@ func (c *ServerCommand) setupStorage(config *server.Config) (physical.Backend, e
 	return backend, nil
 }
 
+func (c *ServerCommand) InitListeners(config *server.Config, disableClustering bool, infoKeys *[]string, info *map[string]string) (int, error, []listenerutil.Listener, []*net.TCPAddr) {
```
swayne275, line 874: is the first return parameter a return code? it might be helpful to indicate the intent there

command/server.go
```diff
@@ -871,6 +871,90 @@ func (c *ServerCommand) setupStorage(config *server.Config) (physical.Backend, e
...
+	clusterAddrs := []*net.TCPAddr{}
+
+	// Initialize the listeners
+	lns := make([]listenerutil.Listener, 0, len(config.Listeners))
+
+	c.reloadFuncsLock.Lock()
```
swayne275, line 880: this looks problematic. there are multiple returns below that don't look like they go through a `c.reloadFuncsLock.Lock()`

i think even if the program exits shortly after one of these returns, we should still be correct with our locking

swayne275 (changes requested):
There's a lot of good stuff here!

I would like to see some of the comments addressed before I approve though (and the ones we discussed on the call).

I'm mostly concerned about the comment about locking

vault/diagnose/diagnose_tls_test.go
HridoyRoy: Yep! That was my understanding as well.

vault/diagnose/diagnose_tls_test.go
HridoyRoy: It's already of type string.

vault/diagnose/diagnose_tls_test.go
HridoyRoy: It exists in a library.

command/operator_diagnose.go
HridoyRoy: Copied from server.go.

vault/diagnose/diagnose_tls_test.go
swayne275: i'm wondering if you could do some error comparison like

`if err == storage.ErrObjectNotExist {
	return nil, nil
}`

rather than directly comparing to a string. like does the library have something like this defined

`var (
	// ErrBucketNotExist indicates that the bucket does not exist.
	ErrBucketNotExist = errors.New("storage: bucket doesn't exist")
	// ErrObjectNotExist indicates that the object does not exist.
	ErrObjectNotExist = errors.New("storage: object doesn't exist")
)`

swayne275: oops that formatting is awful, sorry!

vault/diagnose/diagnose_tls_test.go
swayne275: so is that string signifying trailing data not defined somewhere in the library then?

command/operator_diagnose.go
swayne275: hmm i think i'm a little confused. what's the reason for having that capacity here too? is there a technical reason to have it?

command/operator_diagnose.go
HridoyRoy: Same as above -- we should avoid using dynamically sized arrays where possible. In this case, I'm willing to preallocate the listener length, even if we don't necessarily use all the space. It would be more costly I think to append individually.

command/operator_diagnose.go
HridoyRoy: Good catch! I used that for testing to explicitly initialize to 5, and forgot to remove it.

command/operator_diagnose.go
swayne275: is this a pretty costly operation?

command/server.go
HridoyRoy: Good catch! I think this was a bug with the run command as well, up until now. I added a defer unlock to fix this.

command/server.go
HridoyRoy: Done!

vault/diagnose/diagnose_tls_test.go
HridoyRoy: Done!

command/server.go
swayne275: nice! i'll mark resolved then

command/operator_diagnose.go
HridoyRoy: It doubles the array size every time and copies all elements into a new array. https://blog.golang.org/slices-intro for more info!

swayne275: i meant for this code specifically. like how many listeners are typical?

HridoyRoy: I think either way it's not too big a deal. In the case of 3 listeners, we'd end up with an array of 4 with dynamic sizing, versus 3 in the exact case. In the case of 3 listeners with 1 tls_disable listener, we'd end up with 3 in the exact case versus 2 in the dynamic resizing case. 

So in this particular case where the size of the number of listeners is small, I'd say it's more of an idiomatic/personal preference matter.

swayne275: okay cool on the personal preference bit. i was just curious why that choice was made

vault/diagnose/diagnose_tls_test.go
HridoyRoy: It's not defined as a static string that I can import. It's defined like so: `err = asn1.SyntaxError{Msg: "trailing data"}` in pkcs7.go

swayne275: ah, bummer! so either way you'd still have to have some static string in there. alright!

command/operator_diagnose.go
HridoyRoy: Yeah. I think infoKeys here isn't actually relevant -- I kept it mainly to keep the parity between this and server.go, so I could make this dynamic. But I'd want parity with server.go and preexisting code over necessarily cutting out a few extraneous allocations, which is why I went with this.

vault/core_util.go
HridoyRoy: Yeah, that does seem a bit confusing. I just wanted to signify somehow that using the sanitized config is safer, but I'll remove that part of the comment.

vault/diagnose/diagnose_tls_test.go
HridoyRoy: Yeah, I know what you mean, but in this case I can't do it. The error is `return fail(errors.New("tls: failed to find any PEM data in certificate input"))` in src/crypto/tls/tls.go.

vault/core_util.go
HridoyRoy: Removed! Resolving this conversation.

sgmiller (approved):

command/operator_diagnose.go
```diff
@@ -140,6 +152,61 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
...
+		}
+
+		sanitizedListeners = append(sanitizedListeners, listenerutil.Listener{
+			Listener: ln.Listener,
+			Config:   ln.Config,
+		})
+	}
+	err = diagnose.ListenerChecks(sanitizedListeners)
+	if err != nil {
+		c.UI.Output("Diagnose caught configuration errors: ")
```
swayne275, line 205: I'm not familiar with `c.UI.Output` - is the trailing space intended, or will that be auto added? Basically, is this making us have two spaces between "Diagnose..." and the error?

command/server.go
```diff
@@ -871,6 +871,92 @@ func (c *ServerCommand) setupStorage(config *server.Config) (physical.Backend, e
...
+	var errMsg error
+	for i, lnConfig := range config.Listeners {
+		ln, props, reloadFunc, err := server.NewListener(lnConfig, c.gatedWriter, c.UI)
+		if err != nil {
+			errMsg = fmt.Errorf("Error initializing listener of type %s: %s", lnConfig.Type, err)
+			return 1, nil, nil, errMsg
+		}
+
+		if reloadFunc != nil {
+			relSlice := (*c.reloadFuncs)["listener|"+lnConfig.Type]
```
swayne275, line 894: is there any way the value for this key could be uninit? i.e. do you want to check if it exists using the two parameter return form?

command/server.go
```diff
@@ -871,6 +871,92 @@ func (c *ServerCommand) setupStorage(config *server.Config) (physical.Backend, e
...
+
+	var errMsg error
+	for i, lnConfig := range config.Listeners {
+		ln, props, reloadFunc, err := server.NewListener(lnConfig, c.gatedWriter, c.UI)
+		if err != nil {
+			errMsg = fmt.Errorf("Error initializing listener of type %s: %s", lnConfig.Type, err)
+			return 1, nil, nil, errMsg
+		}
+
+		if reloadFunc != nil {
```
swayne275, line 893: hmm, why are we checking for nil on `reloadFunc` but then using `c.reloadFunc`?

command/server.go
```diff
@@ -871,6 +871,92 @@ func (c *ServerCommand) setupStorage(config *server.Config) (physical.Backend, e
...
+
+		if lnConfig.MaxRequestSize == 0 {
+			lnConfig.MaxRequestSize = vaulthttp.DefaultMaxRequestSize
+		}
+		props["max_request_size"] = fmt.Sprintf("%d", lnConfig.MaxRequestSize)
+
+		if lnConfig.MaxRequestDuration == 0 {
+			lnConfig.MaxRequestDuration = vault.DefaultMaxRequestDuration
+		}
+		props["max_request_duration"] = fmt.Sprintf("%s", lnConfig.MaxRequestDuration.String())
```
swayne275, line 932: what's the advantage to using `fmt.Sprintf` here rather than just assigning `lnConfig.MaxRequestDuration.String()`?

vault/core_util.go
```diff
@@ -55,6 +56,26 @@ func (c *Core) setupReplicationResolverHandler() error {
...
+}
+
+// SanitizedConfig returns a sanitized version of the current config.
+// See server.Config.Sanitized for specific values omitted.
+func (c *Core) SanitizedConfig() map[string]interface{} {
+	conf := c.rawConfig.Load()
+	if conf == nil {
+		return nil
+	}
+	return conf.(*server.Config).Sanitized()
```
swayne275, line 76: is it possible/likely for `*server.Config` to be nil?

swayne275 (approved):
I'm approving because these are all questions (not critiques), but I would like to have them answered.

Looks good!

command/operator_diagnose.go
HridoyRoy: The trailing space is intended -- from my testing, it isn't added automatically.

command/server.go
HridoyRoy: We don't want to append nil reload functions to `relSlice` which is why we check for nil on reloadFunc. c.reloadFuncs can't be nil because it's initialized to an empty map.

command/server.go
HridoyRoy: Probably fine to remove the extra fmt.Sprintf wrapping this, but I didn't want to make changes that weren't necessary for this PR. This code was moved from the run function into a helper method, but it's not new.

vault/core_util.go
HridoyRoy: This just asserts that conf is of type `*server.Config` -- if conf is nil, the above nil check will catch it.

command/server.go
HridoyRoy: Yeah, I think that's a good idea!

command/server.go
swayne275: ah alright. your call then but i think it would be an improvement to simplify and remove the fmt.Sprintf()

whichever way you go, feel free to hit the resolve button!

vault/core_util.go
swayne275: ah okay, is it safe to do the `.Sanitized()` for any realistic case here?

I guess I'm concerned because if we're using this as an assertion on a `*server.Config` and the underlying object isn't `*server.Config`, we will panic

basically what i'm asking is: is this safe? if so, go ahead and mark resolved!

command/server.go
HridoyRoy: I went ahead and removed the wrapping Sprintf.

command/server.go
swayne275: great! feel free to mark resolved when you do, and then go ahead with the approval!

PR #11256, opened 2021-04-01
Updated show lease with toggle ttl picker

arnav28:
1. Updated the TTL picker
2. Passed increment param instead of interval

https://user-images.githubusercontent.com/2932708/113443198-55e72d00-93a6-11eb-96d2-c113acb7ac08.mov


ui/app/templates/vault/cluster/access/leases/show.hbs
```diff
@@ -52,25 +52,31 @@
...
-          </button>
+    <form {{action "renewLease" model increment on="submit"}}>
+      <TtlPicker2 
+        @label="Increment"
+        @helperTextEnabled="Lease will expire after" 
+        @helperTextDisabled="Vault will use the default lease duration" 
+        @labelClass="is-label" 
+        @onChange={{action (mut increment)}} @unit="s" 
+      />
+        <div class="field has-top-margin-s">
```
chelshaw, line 70: nit: spacing on this div should be inline with the TTL picker above (since that one is self-closing)

chelshaw (changes requested):
Looks great! One small nit and then add a changelog please 👍

ui/app/templates/vault/cluster/access/leases/show.hbs
arnav28: Done.

chelshaw (approved):
🎉

PR #11258, opened 2021-04-01
UI/fix kvv2 version

chelshaw:
Unselected dropdown option caused a bug when enabling a v2 secrets engine and creating a secret, where the UI didn't recognize that the new secrets engine was `version 2` until a refresh caused a new network request. In addition to correctly setting the default kv version to 2, we moved that option under the Method Options dropdown:

Version selection is in Method Options toggle

This PR also updates the `MessageError` component to better handle the type of error that was resulting in a message like `[object Object]`, so that arrays of objects are instead passed into the component as strings

Better error message than [object Object]

### DIFF
--- a/changelog/11258.txt
+++ b/changelog/11258.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix bug where the UI does not recognize version 2 KV until refresh, and fix [object Object] error message
+```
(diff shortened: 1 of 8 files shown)


Monkeychip (approved):
Thanks for tackling this!

PR #11261, opened 2021-04-01
add Vault Enterprise license page

rberlind:
Add a page that explains how to deploy a Vault Enterprise license file.  This is currently a gap for Vault compared to Consul and Nomad both of which have explicit `license` CLI commands with corresponding documentation. While Vault does describe the `sys/license` API endpoint under the Vault API docs, many customers either don't find that or don't understand how to convert that into a Vault CLI command that uses `vault write sys/license text="<license>"`.

[Deploy Preview](https://vault-git-add-license-page-hashicorp.vercel.app/docs/enterprise)


yhyakuna:
My recommendation is to add this to the [index.mdx](https://github.com/hashicorp/vault/blob/master/website/content/docs/enterprise/index.mdx) file instead. 

license

website/content/docs/enterprise/license.mdx
```diff
@@ -0,0 +1,22 @@
...
+## Overview
+
+A Vault Enterprise license needs to be applied to a Vault cluster
+in order to use Vault Enterprise features.
+
+The license can be applied to a cluster from any server in it with this command.
+```
+vault write sys/license text="01MV4UU43BK5HGYYTOJZWFQM..."
+```
+
```
yhyakuna, line 20:
````suggestion
The license can be applied to a cluster from any server in it with this command.

```shell-session
$ vault write sys/license text="01MV4UU43BK5HGYYTOJZWFQM..."
```

````

rberlind:
Thanks @yhyakuna: So, I think you're suggesting that we not add a license.mdx page at all but just put the part you had outlined in red on the index.mdx page.  Right?

yhyakuna:
> Thanks @yhyakuna: So, I think you're suggesting that we not add a license.mdx page at all but just put the part you had outlined in red on the index.mdx page. Right?

Yes.  Otherwise, you would need to update the [doc-navigation.js file](https://github.com/hashicorp/vault/blob/master/website/data/docs-navigation.js) to add this to the side menu. But I couldn't think of where to inject this. So, I thought the overview page may be appropriate since that's the starting point.

rberlind:
My assumption had been that the addition to the left-hand menu was triggered by the sidebar_title setting near the top of the new page. But if we did have to add a new Enterprise page to the javascript file, it would be done under https://github.com/hashicorp/vault/blob/master/website/data/docs-navigation.js#L428. And I suspect now that the items in docs-navigation.js need to match the sidebar_title settings on the pages.

In any case, I'm fine with putting on index.mdx.  I'll make that change soon.

rberlind:
OK, @yhyakuna: I moved the content to index.mdx.
Thanks for the suggestion.

yhyakuna (approved):
LGTM

PR #11265, opened 2021-04-02
Backport #11190 (Proxy fix) to 1.6.x

sgmiller:
Backports the update of the armon/go-protoproxy which fixes listener close on proxy error.


mladlow:
Can you add a different reviewer? I don't feel qualified to think about any possible implications.

mladlow:
But thank you for the backport!

PR #11271, opened 2021-04-05
bumping alpine version, improving security

Shahard2:
Security update,

Alpine 3.10 will be out of scope for security updates at the end of the month.

### DIFF
--- a/.circleci/config/executors/@executors.yml
+++ b/.circleci/config/executors/@executors.yml
@@ -20,7 +20,7 @@ python:
   working_directory: /go/src/github.com/hashicorp/vault
 alpine:
   docker:
-    - image: docker.mirror.hashicorp.services/alpine:3.10.2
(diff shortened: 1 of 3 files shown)


jasonodonnell (approved):
LGTM!

Our [official docker images](https://github.com/hashicorp/docker-vault/blob/master/0.X/Dockerfile#L1) are already using 3.13 but these dev images weren't updated. Thanks for doing that.

Shahard2:
> LGTM!
> 
> Our [official docker images](https://github.com/hashicorp/docker-vault/blob/master/0.X/Dockerfile#L1) are already using 3.13 but these dev images weren't updated. Thanks for doing that.

Sure!
when will it be merged?

jasonodonnell:
@Shahard2 I was waiting for CI to finish but there seems to be an error. Looking into it on our end.

PR #11279, opened 2021-04-06
Revert "Remove use of go cache from go test CI", bump Go to 1.15.11

samsalisbury:
Reverts hashicorp/vault#10493

Reviving this work to see if it helps reduce our CI bills.

Also updated an old docker image ref that was not using our mirror, and bumped the Go version to 1.15.11 to make it consistent everywhere (previously 1.15.10 and 1.15.4 were in use).


.circleci/config/commands/go_test.yml
```diff
@@ -6,13 +6,26 @@ parameters:
...
+  cache_dir:
+    type: string
+    default: /tmp/go-cache
+  save_cache:
+    type: boolean
+    default: false
   go_image:
     type: string
-    default: "circleci/golang:1.15.4-buster"
+    default: "docker.mirror.hashicorp.services/circleci/golang:1.15.11-buster"
```
ncabatoff, line 17: I think we're using 1.15.10 elsewhere, e.g. packages-oss.

ncabatoff (approved):

PR #11283, opened 2021-04-06
Bug Fix: OIDC with hcp flag

Monkeychip:
There was an issue using OIDC when coming from a Vault HCP cluster.  This PR fixes that by adding a conditional using the feature-flag service.

Tested locally by flipping the flag.

### DIFF
--- a/ui/app/components/auth-jwt.js
+++ b/ui/app/components/auth-jwt.js
@@ -16,6 +16,7 @@ export { ERROR_WINDOW_CLOSED, ERROR_MISSING_PARAMS, ERROR_JWT_LOGIN };
 
 export default Component.extend({
   store: service(),
+  featureFlagService: service('featureFlag'),
   selectedAuthPath: null,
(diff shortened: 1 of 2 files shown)


ui/app/components/auth-jwt.js
```diff
@@ -131,7 +132,7 @@ export default Component.extend({
     // The namespace can be either be passed as a query paramter, or be embedded
     // in the state param in the format `<state_id>,ns=<namespace>`. So if
     // `namespace` is empty, check for namespace in state as well.
-    if (namespace === '') {
+    if (namespace === '' || this.featureFlagService.managedNamespaceRoot) {
```
Monkeychip, line 135: see [here](https://github.com/hashicorp/vault/blob/master/ui/app/services/feature-flag.js#L17) in code for what this returns.

austingebauer (approved):
LGTM. Thank you!

PR #11284, opened 2021-04-06
UI/obscure secret on input

Monkeychip:
This PR is a project to Obscure Secret Values on input, and to also obscure certificates on display. There is an option in both of these to toggle on/off the obscured view. To accomplish this, design created a new font called "obscure.woff".  We now default to using this font on any input or text-area for either masked-input components and text-file components.

The following are screenshots of places were secrets are now obscured by default:
- Shamir keys

- Anytime we display a secret value, via `masked-input @displayOnly=true`

- Generate credentials (database secret engine example)

- KV Secret engine Copy -> Generate wrap

- No longer displaying certificates on default, but masking them and truncating the output.  The user can then opt to see the certificate. 

**Notes:**
- I was able to glimmerize/update the text-file component, but I was not able to glimmerize the masked-input component.  I suspect the latter is because of the use of a partial in the component chain.
- For the `pgp-file.hbs` component, eventually, we will want to replace this with the text-file component.  See start of code [here](https://github.com/hashicorp/vault/blob/master/ui/app/templates/components/pgp-file.hbs#L29). However, it would require some additional refactoring and be beyond the scope of this project.  Currently, this is the only area in the application where you find the upload file but do not see the option to mask/unmask when entering as text.


ui/app/models/auth-config/ldap.js
```diff
@@ -9,7 +9,7 @@ export default AuthConfig.extend({
   useOpenAPI: true,
   certificate: attr({
     label: 'Certificate',
-    editType: 'textarea',
+    editType: 'file',
```
Monkeychip, line 12: This is a certificate and needed to be a file upload with option to add as text with masked input.

ui/app/styles/components/masked-input.scss
```diff
@@ -1,13 +1,12 @@
+.masked-font {
+  color: $ui-gray-300;
```
Monkeychip, line 2: I have the masked-font defined in the form.scss however the color is overridden when it cascades to this level.

ui/app/styles/components/text-file.scss
```diff
@@ -0,0 +1,10 @@
+.text-file {
+  .has-icon-right {
```
Monkeychip, line 2: Adding toggle display icon
image

ui/app/templates/components/config-pki-ca.hbs
```diff
@@ -9,7 +9,32 @@
...
+        (eq attr.name "caChain")
+        (eq attr.name "privateKey")
+      )}}
+        <InfoTableRow @label={{capitalize (or attr.options.label (humanize (dasherize attr.name)))}} @value={{get model attr.name}}>
+          <MaskedInput
+            @value={{get model attr.name}}
+            @displayOnly={{true}}
+          />
+        </InfoTableRow>
+      {{else if (eq attr.name "expiration")}}
```
Monkeychip, line 25: While I was in here, I made it so the expiration was human readable. See Expiration row.
image

ui/app/templates/vault/cluster/policies/create.hbs
```diff
@@ -46,7 +46,7 @@
         </div>
       </div>
       {{#if showFileUpload}}
-        <TextFile @inputOnly={{true}} @index="" @file={{file}} @onChange={{action "setPolicyFromFile"}} />
+        <TextFile @inputOnly={{true}} @file={{file}} @onChange={{action "setPolicyFromFile"}} />
```
Monkeychip, line 49: index was one of those paramaters in TextFile that was defaulted to null and always passed in as empty string.  Changed so that it is always an empty string.

ui/lib/core/addon/helpers/date-format.js
```diff
@@ -4,6 +4,12 @@ import { format, parseISO } from 'date-fns';
 export function dateFormat([date, style]) {
   // see format breaking in upgrade to date-fns 2.x https://github.com/date-fns/date-fns/blob/master/CHANGELOG.md#changed-5
   let number = typeof date === 'string' ? parseISO(date) : date;
+  if (!number) {
+    return;
+  }
+  if (number.toString().length === 10) {
```
Monkeychip, line 10: the timestamp coming in from pki expiration was a ten digit number, missing milliseconds. This catches those numbers.

ui/lib/core/addon/templates/components/form-field.hbs
```diff
@@ -110,11 +110,9 @@
 {{else if (eq attr.options.editType "file")}}
 {{!-- File Input --}}
   {{text-file
-    index=""
     helpText=attr.options.helpText
     file=file
     onChange=(action "setFile")
-    warning=attr.options.warning
```
Monkeychip, line 117: warning was a parameter never used in the .js or .hbs text-file component.

ui/lib/core/addon/templates/components/masked-input.hbs
```diff
@@ -1,35 +1,41 @@
-<div class="masked-input {{if shouldObscure "masked"}} {{if displayOnly "display-only"}} {{if allowCopy "allow-copy"}}"
+
+<div class="masked-input {{if displayOnly "display-only"}} {{if allowCopy "allow-copy"}}"
   data-test-masked-input data-test-field>
   {{#if displayOnly}}
-    <pre class="masked-value display-only is-word-break">{{displayValue}}</pre>
-  {{else if maskWhileTyping}}
+    <pre class="masked-value display-only is-word-break {{unless showValue "masked-font"}}">{{unless showValue (truncate value 20) value}}</pre>
```
Monkeychip, line 5: truncate for when looking at very long displayOnly secrets.  E.g. certificates.
truncate

ui/lib/core/addon/templates/components/masked-input.hbs
```diff
@@ -1,35 +1,41 @@
...
+  {{else if inputField}}
     <input
       autocomplete="off"
       spellcheck="false"
-      type={{if isMasked "password" "text"}}
```
Monkeychip, line 9: previous work on LDAP bindpass secret. Basically a quick fix for this larger obscuring project. See PR [here](https://github.com/hashicorp/vault/pull/10087
).

vault/ui.go
```diff
@@ -32,7 +32,7 @@ type UIConfig struct {
 // NewUIConfig creates a new UI config
 func NewUIConfig(enabled bool, physicalStorage physical.Backend, barrierStorage logical.Storage) *UIConfig {
 	defaultHeaders := http.Header{}
-	defaultHeaders.Set("Content-Security-Policy", "default-src 'none'; connect-src 'self'; img-src 'self' data:; script-src 'self'; style-src 'unsafe-inline' 'self'; form-action 'none'; frame-ancestors 'none'")
+	defaultHeaders.Set("Content-Security-Policy", "default-src 'none'; connect-src 'self'; img-src 'self' data:; script-src 'self'; style-src 'unsafe-inline' 'self'; form-action 'none'; frame-ancestors 'none'; font-src 'self'")
```
Monkeychip, line 35: I read up on this, and we should be okay to add font-src: self to the header.  But please feel free to read up on it, if you're concerned. [MDN article here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/font-src).

ui/app/styles/components/masked-input.scss
```diff
@@ -50,6 +49,7 @@
   border-left: 0;
   color: $grey;
   box-shadow: 0 3px 1px 0px rgba(10, 10, 10, 0.12);
+  display: flex;
```
hashishaw, line 52: Do we always want this display: flex? What changed?

ui/app/styles/core/forms.scss
```diff
@@ -11,6 +11,13 @@ label {
   }
 }
 
+.masked-font {
+  font-family: obscure;
+  font-size: $size-medium;
+  letter-spacing: 2px;
+  color: $ui-gray-300;
```
hashishaw, line 18: If this should never be another color, you could put `!important` so you don't have to overwrite it later. Up to you.

ui/app/templates/components/config-pki-ca.hbs
```diff
@@ -9,7 +9,32 @@
   </h2>
   {{#if (or model.certificate model.csr)}}
     {{#each model.attrs as |attr|}}
-      {{info-table-row data-test-table-row label=(capitalize (or attr.options.label (humanize (dasherize attr.name)))) value=(get model attr.name)}}
+      {{#if (or 
+        (eq attr.name "csr")
```
hashishaw, line 13: Is there something we could grab from the `attr.options` instead to indicate it should be masked? Instead of listing them out explicitly?

ui/lib/core/addon/helpers/date-format.js
hashishaw: Can you add a test to cover this case?

ui/tests/acceptance/secrets/backend/pki/cert-test.js
```diff
@@ -56,7 +55,9 @@ elRplAzrMF4=
     await settled();
     await generatePage.issueCert('foo');
     await settled();
-    assert.ok(generatePage.hasCert, 'displays the cert');
+    let firstUnMaskButton = document.querySelectorAll('.masked-input-toggle')[0];
+    await click(firstUnMaskButton);
```
hashishaw, line 59: Part of this test should be making sure it is masked before clicking unmask

ui/tests/integration/components/masked-input-test.js
```diff
@@ -54,36 +52,12 @@ module('Integration | Component | masked input', function(hooks) {
     assert.notOk(component.copyButtonIsPresent);
   });
```
hashishaw, line 54: Should we add a test for `Does not unmask text on focus`?

hashishaw (changes requested):
Looks great overall! A couple small updates, and I'm going to pull down the branch and try to break it 😁

ui/app/templates/components/config-pki-ca.hbs
Monkeychip: Great idea.  Fixed.

ui/tests/integration/components/masked-input-test.js
Monkeychip: Good call.  I added a test to cover this.

ui/app/templates/components/pki-cert-show.hbs
```diff
@@ -15,7 +15,23 @@
     {{#if (eq attr.type "object")}}
       <InfoTableRow @label={{capitalize (or attr.options.label (humanize (dasherize attr.name)))}} @value={{stringify (get model attr.name)}} />
     {{else}}
-      <InfoTableRow @label={{capitalize (or attr.options.label (humanize (dasherize attr.name)))}} @value={{get model attr.name}} />
+      {{#if (or 
+        (eq attr.name "csr")
```
hashishaw, line 19: Could probably use `attr.options.masked` here too

hashishaw (approved):
Looks great! Thanks for making these changes

PR #11295, opened 2021-04-07
Backport 10181 1.7.x

sgmiller:


physical/dynamodb/dynamodb.go
```diff
@@ -497,15 +500,39 @@ func (d *DynamoDBBackend) HAEnabled() bool {
...
+		boff := backoff.NewExponentialBackOff()
+		boff.MaxElapsedTime = 600 * time.Second
+
+		for len(batch) > 0 {
+			output, err := d.client.BatchWriteItem(&dynamodb.BatchWriteItemInput{
+				RequestItems: batch,
+			})
+
+			if err != nil{
+				break
```
swayne275, line 519: i don't think this is setting the outside-of-loop `err` to the inside-of-loop `err`, was that your intention?

I was curious about this so I spun up a quick example, and I believe it validates that the `err` on line 506 would remain `nil` in this case

example: https://play.golang.org/p/94zycqbFglQ

thoughts?

physical/dynamodb/dynamodb.go
```diff
@@ -497,15 +500,39 @@ func (d *DynamoDBBackend) HAEnabled() bool {
...
+			}
+
+			if len(output.UnprocessedItems) == 0 {
+				break
+			} else {
+				duration := boff.NextBackOff();
+				if (duration != backoff.Stop) {
```
swayne275, line 526: nitpick: parens not needed

physical/dynamodb/dynamodb.go
```diff
@@ -497,15 +500,39 @@ func (d *DynamoDBBackend) HAEnabled() bool {
...
+					batch = output.UnprocessedItems
+					time.Sleep(duration)
+				} else {
+					err = errors.New("dynamodb: timeout handling UnproccessedItems")
```
swayne275, line 530: i believe this is also assigning to the local `err` instance, not the one outside the loop

swayne275 (changes requested):
requesting changes based on the assumption that the line 506 error should capture stuff inside the loop, which I do not believe is what is happening

swayne275 (approved):

PR #11300, opened 2021-04-07
Add More TLS Tests and Verification of TLS Root Certificate

HridoyRoy:
This PR expands upon the use of the Verify method to make sure that the root CA, when specified in the listener config, is checked by the diagnose command. 

The PR also implements some additional tests verifying prior functionality, and adds a check in for the TLS version configuration in the listener. 

Related tickets:

https://hashicorp.atlassian.net/browse/VAULT-1998

https://hashicorp.atlassian.net/browse/VAULT-2002


vault/diagnose/tls.go
```diff
@@ -4,24 +4,52 @@ import (
...
 )
 
+const minVersionError = "'tls_min_version' value %q not supported, please specify one of [tls10,tls11,tls12,tls13]"
+const maxVersionError = "'tls_max_version' value %q not supported, please specify one of [tls10,tls11,tls12,tls13]"
+
 func ListenerChecks(listeners []listenerutil.Listener) error {
 	for _, listener := range listeners {
 		l := listener.Config
-		err := TLSFileChecks(l.TLSCertFile, l.TLSKeyFile)
+		var err error
```
swayne275, line 20: nitpick/style thing, but this could be declared closer to where it's actually used

vault/diagnose/tls.go
```diff
@@ -41,12 +69,36 @@ func TLSFileChecks(certFilePath, keyFilePath string) error {
...
+			return fmt.Errorf("failed to verify certificate: " + err.Error())
+		}
+	} else {
+		if rootFilePath == "" {
+			return fmt.Errorf("TLS Root CA file not specified, but TLS chain verification is required")
+		}
+		caPool := x509.NewCertPool()
+		data, err := ioutil.ReadFile(rootFilePath)
+		if err != nil {
+			return fmt.Errorf("failed to read tls_client_ca_file: %s", err.Error())
```
swayne275, line 86: you could use `%w` verb with `err` so that this error can be unwrapped

vault/diagnose/tls.go
```diff
@@ -41,12 +69,36 @@ func TLSFileChecks(certFilePath, keyFilePath string) error {
...
+		if !caPool.AppendCertsFromPEM(data) {
+			return fmt.Errorf("failed to parse CA certificate in tls_client_ca_file")
+		}
+		// Check that certificate isn't expired and is of correct usage type
+		if _, err = cert.Leaf.Verify(x509.VerifyOptions{
+			Roots:         caPool,
+			Intermediates: x509.NewCertPool(),
+			KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		}); err != nil {
+			return fmt.Errorf("failed to verify certificate: " + err.Error())
```
swayne275, line 98: you're using a formatted error here without formatting. i think the same `%w` verb bit could be good here too

vault/diagnose/tls_test.go
```diff
@@ -132,15 +87,183 @@ func TestTLSTrailingData(t *testing.T) {
 	}
 }
 
-func TestTLSExpiredCert(t *testing.T) {
+func TestTLSExpiredCert(t *testing.T) {}
```
swayne275, line 90: what is this intended to do?

vault/diagnose/tls_test.go
```diff
@@ -132,15 +87,183 @@ func TestTLSTrailingData(t *testing.T) {
...
+				TLSClientCAFile:               "./../../api/test-fixtures/root/rootcacert.pem",
+				TLSMinVersion:                 "tls10",
+				TLSRequireAndVerifyClientCert: true,
+				TLSDisableClientCerts:         false,
+			},
+		},
+	}
+	err := ListenerChecks(listeners)
+	if err == nil {
+		t.Errorf("TLS Config check on fake certificate should fail")
```
swayne275, line 112: nit: this could be `t.Error()`

vault/diagnose/tls_test.go
```diff
@@ -132,15 +87,183 @@ func TestTLSTrailingData(t *testing.T) {
...
+				TLSClientCAFile:               "./../../api/test-fixtures/root/rootcacert.pem",
+				TLSMinVersion:                 "tls10",
+				TLSRequireAndVerifyClientCert: true,
+				TLSDisableClientCerts:         false,
+			},
+		},
+	}
+	err = ListenerChecks(listeners)
+	if err == nil {
+		t.Errorf("TLS Config check on fake certificate should fail")
```
swayne275, line 135: nit: `t.Error()`

vault/diagnose/tls_test.go
```diff
@@ -132,15 +87,183 @@ func TestTLSTrailingData(t *testing.T) {
...
+				TLSClientCAFile:               "./../../api/test-fixtures/root/rootcacert.pem",
+				TLSMinVersion:                 "tls10",
+				TLSRequireAndVerifyClientCert: true,
+				TLSDisableClientCerts:         false,
+			},
+		},
+	}
+	err := ListenerChecks(listeners)
+	if err == nil {
+		t.Errorf("TLS Config check on fake certificate should fail")
```
swayne275, line 161: nit: `t.Error()`

vault/diagnose/tls_test.go
```diff
@@ -132,15 +87,183 @@ func TestTLSTrailingData(t *testing.T) {
...
```
swayne275, line 187: nit: `t.Error()`

swayne275: i'm going to stop commenting on all of these haha

swayne275 (approved):
just a few nitpicks or code quality things, but overall looks good

vault/diagnose/tls.go
```diff
@@ -4,24 +4,52 @@ import (
...
+		if !ok {
+			return fmt.Errorf(minVersionError, l.TLSMinVersion)
+		}
+		_, ok = tlsutil.TLSLookup[l.TLSMaxVersion]
+		if !ok {
+			return fmt.Errorf(maxVersionError, l.TLSMaxVersion)
+		}
+
+		// Perform checks on the TLS Cryptographic Information.
+		if l.TLSRequireAndVerifyClientCert {
```
HridoyRoy, line 39: This is a mistake on my part -- this flag checks whether the listener verifies client certificates, but I'm using it to decide whether a root CA cert for the server, or intermediates, were passed in. I'm not sure how to split out the root CAs and inters from one single PEM file though.

vault/diagnose/tls_test.go
HridoyRoy: This makes sure an expired cert is caught with the appropriate error message. Filled in now! I had to wait a day for my openssl certificate to expire as openssl only allows > day expiration times on certs.

vault/diagnose/tls.go
HridoyRoy: `fmt.Errorf` is the only Error call in the `fmt` package, so I have to use it. But to your point, I'm happy to change all of these error messages to use `%w` instead of `%s` with `err.Error()` :).

vault/diagnose/tls_test.go
HridoyRoy: Will do (for all of these errors that don't have formatting).

vault/diagnose/tls_test.go
HridoyRoy: No worries! I'm going to resolve all of these but one, and make the change throughout this file wherever I'm not using formatting explicitly in the errors.

vault/diagnose/tls.go
swayne275: yeah, i guess i just thought it looked a little strange with a `"" + <var>` thing instead of a `"<some verb>", <var>` thing, but great!

vault/diagnose/tls_test.go
swayne275: oh cool :)

vault/diagnose/tls.go
HridoyRoy: Fixed.

vault/diagnose/tls_verification.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/hashicorp/vault/internalshared/listenerutil"
+	"github.com/hashicorp/vault/sdk/helper/tlsutil"
+	"github.com/hashicorp/vault/vault"
+)
+
+const minVersionError = "'tls_min_version' value %q not supported, please specify one of [tls10,tls11,tls12,tls13]"
+const maxVersionError = "'tls_max_version' value %q not supported, please specify one of [tls10,tls11,tls12,tls13]"
+
+func ListenerChecks(listeners []listenerutil.Listener) error {
+	for _, listener := range listeners {
```
HridoyRoy, line 20: Check that client related certificate paths exist.

HridoyRoy: This should incorporate reading and loading.

vault/diagnose/tls_verification.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	// After verify passes, we need to check the values on the certificate itself.
+	// This is a separate check beyond the certificate expiry and chain checks.
+
+	cert, err := tls.LoadX509KeyPair(certFilePath, keyFilePath)
+	if err != nil {
+		return err
+	}
+	x509Cert, err := x509.ParseCertificate(cert.Certificate[0])
```
HridoyRoy, line 124: This can be removed.

vault/diagnose/tls_verification.go
```diff
@@ -0,0 +1,138 @@
...
+		// Perform checks on the TLS Cryptographic Information.
+		if err = TLSFileChecks(l.TLSCertFile, l.TLSKeyFile); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// TLSFileChecks contains manual error checks against the TLS configuration
+func TLSFileChecks(certFilePath, keyFilePath string) error {
```
HridoyRoy, line 49: Modularize this to load the cert from a file path separately, and have the checks in a separate function so that we can load from urls as well.

vault/diagnose/tls_verification.go
```diff
@@ -0,0 +1,138 @@
...
+			return fmt.Errorf(minVersionError, l.TLSMinVersion)
+		}
+		_, ok = tlsutil.TLSLookup[l.TLSMaxVersion]
+		if !ok {
+			return fmt.Errorf(maxVersionError, l.TLSMaxVersion)
+		}
+
+		var err error
+		// Perform checks on the TLS Cryptographic Information.
+		if err = TLSFileChecks(l.TLSCertFile, l.TLSKeyFile); err != nil {
```
swayne275, line 41: is there a reason to not declare `err` here with `:=`?

vault/diagnose/tls_verification_test.go
```diff
@@ -0,0 +1,317 @@
...
+				TLSKeyFile:                    "./../../api/test-fixtures/keys/key.pem",
+				TLSMinVersion:                 "tls10",
+				TLSRequireAndVerifyClientCert: true,
+				TLSDisableClientCerts:         false,
+			},
+		},
+	}
+	err := ListenerChecks(listeners)
+	if err == nil {
+		t.Error("TLS Config check on fake certificate should fail")
```
swayne275, line 53: i think these should be fails (not errors), since you can't do the next check if `err` is nil (presumably that's a caught panic with the test infra)

swayne275: there are multiple cases like this

vault/diagnose/tls_verification_test.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+	if err.Error() != "tls: private key type does not match public key type" {
```
swayne275, line 135: did you want these to be `!string.Contains()` as you've done for your other tests?

swayne275 (commented):
the comments in your functions and tests were SUPER helpful!

vault/diagnose/tls_verification_test.go
HridoyRoy: I'm so sorry! I didn't realized you had left some comments @swayne275 with the approval. Yeah, fails here would be better. I'll push a PR out today cleaning that up. 

I can also use `:=` for the error var in the cleanup PR. For the `!string.Contains()` comment below, I can't do that for this particular test, because the error message thrown by the library is a concatenation of a few error messages. I didn't want to copy that whole string here, so I just checked the necessary part to distinguish this failure from the other cases.

Thanks for the comments, and sorry about the premature merge! I'll have followup PR with the improvements ASAP!

swayne275: No problem! I wouldn't have approved it if I wasn't okay with it going without fixes/responses to the comments

The panic in tests isn't that big of a deal, it's just a lot cleaner to avoid it

HridoyRoy: I'm changing these to Fatals, not Fails -- Error is equivalent to a Log followed by a Fail, and Fail continues execution. Fatal stops the test with a FailNow.

swayne275: oh yeah fatal is probably what i meant, cool!

Issue #11319, opened 2021-04-08
HA MySQL/MariaDB: Deadlock on leader, if HA DB connection is lost


thingstad:
**Deadlock on leader if HA DB connection is lost**
In a setup with MySQL/MariaDB HA back-end, if the leader looses it's DB connection it has a fair chance of mitigating into a deadlock state.
Loosing the connection releases the lock, and your standby assumes leadership.
You end up with multiple masters.
The previous leader will still report itself as master; as will your former standby node - which now has grabbed leadership.

**To Reproduce**
1. Prepare a MySQL/MariaDB instance.
2. Configure two Vault instances to each use MySQL as HA back-end
3. Start and unseal Vault instances
4. On MySQL/MariaDB instance run `sudo systemctl restart mariadb` to restart DB and hence drop current connections
5. Wait a few seconds till the standby assumes leadership.
6. Issue `curl http://node-addr:8200/v1/sys/leader` on each Vault node.
7. Observe multiple masters repported.

**Expected behavior**
Expectation is that the leader will recover from having lost the leadership.
Either by trying to grab leadership once again, or becoming a standby.

**Environment:**
* Vault Server Version: **1.7**
* HA back-end DB: **MariaDB 10.5.9**

Vault server configuration file(s):

```hcl
storage "mysql" {
  address = "10.10.10.4:3306"
  database = "vault"
  table = "vault_data"
  username = "vault-username"
  password = "vault-password"
  ha_enabled = "true"
  lock_table = "vault_lock"    
}

listener "tcp" {
  address       = "10.10.10.2:8200"
  tls_disable = 1
}

api_addr = "http://10.10.10.2:8200"
ui = true
log_level = "info"
```

**Additional ways to observe the issue**

Adding one line of logging to `physical/mysql/mysql.go`

Change

```go
// hasLock will check if a lock is held by checking the current lock id against our known ID.
func (i *MySQLHALock) hasLock(key string) error {
	var result sql.NullInt64
	err := i.in.statements["used_lock"].QueryRow(key).Scan(&result)
	if err == sql.ErrNoRows || !result.Valid {
		// This is not an error to us since it just means the lock isn't held
		return nil
	}
```

Into

```go
// hasLock will check if a lock is held by checking the current lock id against our known ID.
func (i *MySQLHALock) hasLock(key string) error {
	var result sql.NullInt64
	err := i.in.statements["used_lock"].QueryRow(key).Scan(&result)
[...]

PR #11320, opened 2021-04-08
Fix mysql deadlock

thingstad:
Fix for https://github.com/hashicorp/vault/issues/11319

### DIFF
--- a/changelog/11319.txt
+++ b/changelog/11319.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+physical/mysql: Fix multi master issue after lost connection.
+```
--- a/physical/mysql/mysql.go
+++ b/physical/mysql/mysql.go
@@ -576,8 +576,8 @@ func (i *MySQLHALock) hasLock(key string) error {
 	var result sql.NullInt64
 	err := i.in.statements["used_lock"].QueryRow(key).Scan(&result)
 	if err == sql.ErrNoRows || !result.Valid {
-		// This is not an error to us since it just means the lock isn't held
-		return nil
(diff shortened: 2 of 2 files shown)


hsimon-hashicorp:
Hi @thingstad! Can you add a small changelog entry for this? We'll get it reviewed (ping @imthaghost). Thanks!

ncabatoff:
Hi @thingstad,

Do you suppose you could write a test to demonstrate the problem and the fix?  Typically our physical backend implementations have integration tests.  Sadly the mysql physical tests are relying on the user to provide a MYSQL_ADDR pointing to an existing DB they manage, and the tests are skipped otherwise.  What they should be doing instead is use https://github.com/hashicorp/vault/blob/main/helper/testhelpers/mysql/mysqlhelper.go#L21-L21 to create a docker container for mysql.  This will allow the tests to run in CI, which would already be a great improvement.  It looks like this was already done for TestMySQLHABackend_LockFailPanic, but we failed to add use of PrepareTestContainer to the other tests at the time.

You could make that change as part of this PR, which would be nice but not required.  What I would really like to see though is a test that shows that your change solves the problem, i.e. something that reproduces the steps from #11319 using docker containers.  I realize that the bug may not manifest every time the db is restarted, but even if it takes many invocations of the tests to see the failure, that's still something.  If it takes 20 attempts on average to make the failure happen before your change, and then we run the test 100 times with your change without seeing a failure, that's good enough evidence that it fixes the problem.

hsimon-hashicorp:
Due to the age of this PR, I will go ahead and close it at this point. Please feel free to re-base and continue if you're willing to do so. Thanks!

Issue #10845, opened 2021-02-05
Unable to trace profile standby mode servers


tionebsalocin:
**Describe the bug**
Using /sys/pprof endpoints on standby mode servers does not work due to token check issue

**To Reproduce**
1. Run the following command targeting a standby mode server:
```
curl \
    --header "X-Vault-Token: ..." \
    http://standby-server:8200/v1/sys/pprof/trace
```

2. Get the following error;
`{"errors":["error performing token check: Vault is in standby mode"]}`

**Expected behavior**

**Environment:**
* Vault Server Version (retrieve with `vault status`): `1.5.7`
* Vault CLI Version (retrieve with `vault version`): `1.5.7`
* Server Operating System/Architecture: `CentOS 7 5.4.93-1.el7.x86_64`

Vault server configuration file(s):

```json
{
  "api_addr": "http://server1:8200",
  "disable_mlock": true,
  "ui": true,
  "plugin_directory": "/var/opt/vault/plugins/",
  "listener": {
    "tcp": {
      "address": "0.0.0.0:8200",
      "tls_disable": "true"
    }
  },
  "storage": {
    "consul": {
      "address": "http://localhost:9500",
      "token": "..."
    }
  },
  "ha_storage": {
    "consul": {
      "address": "http://localhost:9500",
      "path": "ha-vault",
      "token": "..."
    }
  },
  "telemetry": {
    "prometheus_retention_time": "24h",
    "disable_hostname": true
  }
}
```

**Additional context**
Vault runs on 3 different servers
Storage backend is consul

HridoyRoy:
I _think_ this is a bug, and the correct behavior should be that the request is routed to the leader.

ncabatoff:
I disagree with the solution being routing the request to the leader: we want to be able to profile standby nodes too.  I see two easy options: add an "unauthenticated" listener-level option to allow profiling without a token, like we do for metrics.  Option 2: listen for SIGUSR1 like go-metrics does, and in addition to our current behaviour of dumping metrics to the logs, if some new config is enabled, also generate a `vault debug` style dump in the filesystem where the config specifies to put it.

Another option that's more involved but shouldn't be ruled out: add an RPC method that allows standbys to ask the active node to verify a token.  Then we can support the same pprof interface that's already being used on the active node and on perf standbys, without needing new config.

ncabatoff:
[...]

PR #11324, opened 2021-04-08
Add support for unauthenticated pprof access on a per-listener basis,…

ncabatoff:
… as we do for metrics.  Note that since this change doesn't affect the endpoint `sys/config/state/sanitized`, by default `vault debug` will fail with an error when no token is provided, even with the new option enabled.  However, using e.g. `-target=pprof` will work around that.

### DIFF
--- a/command/debug.go
+++ b/command/debug.go
@@ -748,6 +748,8 @@ func (c *DebugCommand) collectPprof(ctx context.Context) {
 			}(target)
 		}
 
+		// As a convenience, we'll also fetch the goroutine target using debug=2, which yields a text
+		// version of the stack traces that don't require using `go tool pprof` to view.
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
--- a/command/debug.go
+++ b/command/debug.go
@@ -985,6 +985,7 @@ func (c *DebugCommand) captureError(target string, err error) {
 
 func (c *DebugCommand) writeLogs(ctx context.Context) {
 	out, err := os.Create(filepath.Join(c.flagOutput, "vault.log"))
+	defer out.Close()
 	if err != nil {
 		c.captureError("log", err)
 		return
--- a/command/debug.go
+++ b/command/debug.go
@@ -674,15 +674,11 @@ func (c *DebugCommand) collectMetrics(ctx context.Context) {
 		}
 
 		// Check replication status. We skip on processing metrics if we're one
-		// of the following (since the request will be forwarded):
-		// 1. Any type of DR Node
-		// 2. Non-DR, non-performance standby nodes
+		// a DR node, though non-perf standbys will fail if they aren't using
+		// unauthenticated_metrics_access.
 		switch {
 		case healthStatus.ReplicationDRMode == "secondary":
 			c.logger.Info("skipping metrics capture on DR secondary node")
-			continue
-		case healthStatus.Standby && !healthStatus.PerformanceStandby:
-			c.logger.Info("skipping metrics on standby node")
 			continue
 		}
 
--- a/command/debug.go
+++ b/command/debug.go
@@ -582,7 +582,10 @@ func (c *DebugCommand) capturePollingTargets() error {
 
 	if strutil.StrListContains(c.flagTargets, "log") {
 		g.Add(func() error {
-			_ = c.writeLogs(ctx)
+			c.writeLogs(ctx)
+			// If writeLogs returned earlier due to an error, wait for context
+			// to terminate so we don't abort everything.
+			<-ctx.Done()
 			return nil
 		}, func(error) {
 			cancelFunc()
@@ -982,16 +985,17 @@ func (c *DebugCommand) captureError(target string, err error) {
 	c.errLock.Unlock()
 }
 
-func (c *DebugCommand) writeLogs(ctx context.Context) error {
+func (c *DebugCommand) writeLogs(ctx context.Context) {
 	out, err := os.Create(filepath.Join(c.flagOutput, "vault.log"))
 	if err != nil {
-		return err
+		c.captureError("log", err)
+		return
 	}
-	defer out.Close()
 
 	logCh, err := c.cachedClient.Sys().Monitor(ctx, "trace")
 	if err != nil {
-		return err
+		c.captureError("log", err)
+		return
 	}
 
 	for {
@@ -999,10 +1003,11 @@ func (c *DebugCommand) writeLogs(ctx context.Context) error {
 		case log := <-logCh:
 			_, err = out.WriteString(log)
 			if err != nil {
-				return err
+				c.captureError("log", err)
+				return
 			}
 		case <-ctx.Done():
-			return nil
+			return
 		}
 	}
 }
--- a/command/debug.go
+++ b/command/debug.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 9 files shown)


ncabatoff:
Fixes #10845.

ncabatoff:
> Simple and neat! The only thought I had on this is if we could enable SIGHUP for this config to be loaded on the fly or not.

That would be great, I agree.  I looked into it based on your suggestion and it looks like it would be a bigger time investment than I can spare today, so I think I'll defer that.

http/sys_metrics.go
```diff
@@ -47,3 +50,22 @@ func handleMetricsUnauthenticated(core *vault.Core) http.Handler {
...
+		req, _, status, err := buildLogicalRequestNoAuth(core.PerfStandby(), w, r)
+		if err != nil || status != 0 {
+			respondError(w, status, err)
+			return
+		}
+		if origBody != nil {
+			r.Body = ioutil.NopCloser(origBody)
+		}
+
+		resp, err := core.HandleRequest(r.Context(), req)
```
briankassouf, line 68: On standby nodes this call should fail here: https://github.com/hashicorp/vault/blob/3d0bc374335dfbc6a83d9f4d3fbb468d5957b74f/vault/request_handling.go#L402-L404

The system backend isn't mounted on standby nodes, so this change may be a bit harder than this. You may need to move these handlers into helper functions that are either called directly here or in the system backend

briankassouf (changes requested):

briankassouf:
> Note that since this change doesn't affect the endpoint sys/config/state/sanitized, by default vault debug will fail with an error when no token is provided, even with the new option enabled

Can we make debug ignore this error and continue? The sanitized config step failing shouldn't cause the whole debug package to fail.

http/sys_metrics.go
ncabatoff: Nice catch, thanks!  I had tested at the CLI, but inadequately (I hadn't inspected the resulting debug packages.)

I'm on the fence as to whether my new approach is correct.  Instead of moving the existing handlers as you proposed, I'm using the pprof handlers directly.  Since our sys handlers are very thin wrappers around those that don't currently add anything to what they do, this is technically correct in that the user sees the same behaviour with and without the unauth toggle.  On the other hand, if in future we make enhancements to the sys wrappers, they will diverge.  What do you think?

http/handler.go
```diff
@@ -177,6 +177,18 @@ func Handler(props *vault.HandlerProperties) http.Handler {
 			mux.Handle("/v1/sys/metrics", handleLogicalNoForward(core))
 		}
 
+		if props.ListenerConfig != nil && props.ListenerConfig.Profiling.UnauthenticatedPProfAccess {
+			for _, name := range []string{"goroutine", "threadcreate", "heap", "allocs", "block", "mutex"} {
```
ncabatoff, line 181: Some of these aren't supported by the existing code: threadcreate, allocs, block, and mutex.  I'll see if I can find out if they were excluded deliberately or not, and if not, add them to the existing sys handlers.

ncabatoff:
> > Note that since this change doesn't affect the endpoint sys/config/state/sanitized, by default vault debug will fail with an error when no token is provided, even with the new option enabled
> 
> Can we make debug ignore this error and continue? The sanitized config step failing shouldn't cause the whole debug package to fail.

We could, but I'm not sure I agree with you on this.  Why is the config target okay to tolerate errors for, but not e.g server-status?  If there weren't an easy workaround I'd be inclined to change this, but as it is I think the current behaviour makes sense.  Alternatively, we could go the other way and tolerate failures in any target, just make everything best-effort.

http/handler.go
ncabatoff: I've added them pre-emptively.  @calvn do you know if they were omitted deliberately?

command/debug.go
```diff
@@ -731,35 +732,35 @@ func (c *DebugCommand) collectPprof(ctx context.Context) {
...
+				}
+			}(target)
+		}
 
-		// Capture heap
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			data, err := pprofHeap(ctx, c.cachedClient)
+			data, err := pprofTarget(ctx, c.cachedClient, "goroutine", url.Values{"debug": []string{"2"}})
```
briankassouf, line 755: Nice!

http/handler.go
calvn: The initial implementation was mostly a passthrough of the handlers exposed on the `net/http/pprof` package which includes cmdline, profile, symbol, and trace, but not the ones above since they were not exposed by it directly. 

I see that you're now routing them directly in here instead. Does it make `vault/logical_system_pprof.go` moot?

ncabatoff: `heap` and `goroutine` were already supported, I'm adding `allocs`, `block`, and `mutex`.  All of these are invoked in the same way, hence my question.

As to "does this obviate the need for logical_system_pprof" - no, because for authenticated requests we still want to go through our own request handlers for auth check, auditing, etc.

calvn: Ah, I see that it still routes to those paths on the authenticated case. Would it make sense to add profiling on `[]string{"goroutine", "threadcreate", "heap", "allocs", "block", "mutex"}` for the authenticated scenario as well?

ncabatoff: Absolutely, think I meant to then forgot.  Thanks for suggestion.

ncabatoff: Added

command/debug.go
```diff
@@ -731,35 +731,35 @@ func (c *DebugCommand) collectPprof(ctx context.Context) {
...
-		// Capture goroutines
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			data, err := pprofGoroutine(ctx, c.cachedClient)
-			if err != nil {
-				c.captureError("pprof.goroutine", err)
-				return
-			}
+		for _, target := range []string{"threadcreate", "allocs", "block", "mutex", "goroutine", "heap"} {
```
vishalnayak, line 734: Are the `profile`, `symbol` and `trace` intentionally omitted?

command/debug.go
```diff
@@ -731,35 +731,35 @@ func (c *DebugCommand) collectPprof(ctx context.Context) {
...
+				}
+			}(target)
+		}
 
-		// Capture heap
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			data, err := pprofHeap(ctx, c.cachedClient)
+			data, err := pprofTarget(ctx, c.cachedClient, "goroutine", url.Values{"debug": []string{"2"}})
```
vishalnayak, line 754: I looked up what the `2` represents here and it is indeed nice that it is added. Can you please add a comment here to that effect so it is easier to know that it does?

command/debug.go
```diff
@@ -731,35 +731,35 @@ func (c *DebugCommand) collectPprof(ctx context.Context) {
...
 			if err != nil {
-				c.captureError("pprof.heap", err)
+				c.captureError("pprof.goroutines-text", err)
```
vishalnayak, line 756: Out of curiosity, why do we have the `-text` suffix here?

command/debug.go
```diff
@@ -994,27 +981,29 @@ func (c *DebugCommand) captureError(target string, err error) {
...
 
-func (c *DebugCommand) writeLogs(ctx context.Context) error {
+func (c *DebugCommand) writeLogs(ctx context.Context) {
 	out, err := os.Create(filepath.Join(c.flagOutput, "vault.log"))
 	if err != nil {
-		return err
+		c.captureError("log", err)
+		return
 	}
-	defer out.Close()
```
vishalnayak, line 1002: Why is this being removed?

command/server/config_test_helpers.go
```diff
@@ -711,6 +711,12 @@ listener "tcp" {
 	tls_max_version = "tls13"
 	tls_require_and_verify_client_cert = true
 	tls_disable_client_certs = true
+    telemetry {
+      unauthenticated_metrics_access = true
```
vishalnayak, line 715: There are a bunch of config fields where we are using `enable_` and `disable_` as prefixes, including the newly added `enable_response_header_raft_node_id` and the `enable_response_header_hostname`.

These ones also seem to fall under the same category and could take `enable_` prefixes. WDYT?

vishalnayak: Also, just to be sure that this is not something are overlooking. Is there a reason why these config shouldn't live at the top level (as opposed to being in the listener blocks)?

http/sys_metrics_test.go
```diff
@@ -56,3 +56,41 @@ func TestSysMetricsUnauthenticated(t *testing.T) {
...
+				UnauthenticatedPProfAccess: true,
+			},
+		},
+	}
+	TestServerWithListenerAndProperties(t, ln, addr, core, props)
+	defer ln.Close()
+	TestServerAuth(t, addr, token)
+
+	// Test without token
+	resp = testHttpGet(t, "", addr+"/v1/sys/pprof/cmdline")
```
vishalnayak, line 90: Do we want to add test steps for the paths that exercise the pprof handlers instead of the http handler funcs? At least for one of them maybe?

command/debug.go
ncabatoff: Yes, they're handled using a different mechanism.

command/debug.go
ncabatoff: Good call, done.

command/debug.go
ncabatoff: To distinguish it from the `pprof.goroutine` target.

command/server/config_test_helpers.go
ncabatoff: Re enable/disable: a boolean type already implies enabling/disabling something to me, and we have many bools already that don't have such prefixes.  In any event, in this case the new setting mirrors unauthenticated_metrics_access and I'd rather be consistent with that.

These configs belong in listener blocks rather than top-level so that people can enable them in e.g. localhost listener but not in public listener, as a safety precaution.

vishalnayak: Sounds good!

http/sys_metrics_test.go
ncabatoff: This test is specifically for the new functionality (unauthenticated pprof) which doesn't/cannot use the http handler funcs.  There are pre-existing tests for the http handler funcs.

vishalnayak: Ah, I see. Alright.

vishalnayak (approved):
Looking great!

command/debug.go
ncabatoff: Sorry I missed this.  I think it's a mistake, I'll restore it.

briankassouf (approved):
Looks good!

PR #11325, opened 2021-04-08
Backport 1.6: Add log gathering to debug command. (#10609)

ncabatoff:


vishalnayak (approved):
I hadn't reviewed the original PR, but the diff looks the same and there seems to be no manual change here. Hence looks good.

PR #11331, opened 2021-04-09
Fixes for db connection file type fields

arnav28:
1. Fixed helper text for file type form fields

Screen Shot 2021-04-09 at 10 58 28 AM

2. Added padding to bottom form section

Screen Shot 2021-04-09 at 10 58 17 AM


Monkeychip (approved):
Nice work.  I've gone ahead and added the milestone and the label ui. These are two labels we try to add for each PR and they're easy to forget :) 

image

PR #11333, opened 2021-04-09
Expiration Manager: Handle Presumed Irrevocable Leases Separately

swayne275:
This PR allows leases that are deemed irrevocable to be marked as "zombies", taking them out of the expiration rotation. This is to avoid wasting work on leases that are unlikely to be revoked without manual intervention. This can happen if a backing storage system is removed or misconfigured (e.g. leases stored in a database that cannot be accessed anymore).

Leases are currently only marked as zombies if they fail revocation 6 times. In the future they will also be marked as zombies if the backing systems return an irrevocable error. See VLT-145 for more details.

Question: I don't think zombies need to be included in `walkLeases` or `WalkTokens` - these appear to be for metrics, and I don't think zombies should be included in the metrics for potentially valid leases (there are/will be separate zombie metrics - again, see RFC).

Question: for those of you that have worked on this in the past: please let me know if I "missed a spot" - from my analysis it looks like this should do it!


vault/expiration.go
```diff
@@ -94,6 +98,7 @@ type ExpirationManager struct {
 	// TTL zero, which we want to count but have no timer associated.
 	pending     sync.Map
 	nonexpiring sync.Map
+	zombies     sync.Map
```
HridoyRoy, line 101: A short description of what we expect this to contain might be useful here?

HridoyRoy, line 101: A short description of what we expect this to contain might be useful here?

vault/expiration.go
```diff
@@ -2340,6 +2408,11 @@ func (le *leaseEntry) nonexpiringToken() bool {
 	return !le.Auth.LeaseEnabled()
 }
 
+// TODO maybe lock RevokeErr once this goes in: https://github.com/hashicorp/vault/pull/11122
+func (le *leaseEntry) isZombie() bool {
```
HridoyRoy, line 2412: Is the criteria for a lease being a zombie just that it wasn't able to be revoked? I thought it would be if it had failed to be revoked a certain number of times?

HridoyRoy: I think this is the part that is confusing me. If we can check whether a lease is a zombie or not without storing it, why would we save these leases separately instead of just using that dynamic check?

vault/expiration_test.go
```diff
@@ -2491,3 +2491,165 @@ func TestExpiration_FairsharingEnvVar(t *testing.T) {
 		}
 	}
 }
+
+// register one lease ID and return the leaseID
+func registerOneLease(t *testing.T, ctx context.Context, exp *ExpirationManager) string {
```
HridoyRoy, line 2496: The general question I have, looking at these tests, is that I'm not sure why we are tracking zombies separately in a separate struct. From what I can tell, we're still treating them in a similar way, modulo a few operations where we aren't doing that operation on the zombie lease (because it's stored separately). Is that understanding correct? If so, then wouldn't it be more clear to keep zombies as pending leases, but check whether a lease is a zombie using isZombie(), and skip actions for that lease if it is? 
Thanks!

HridoyRoy (commented):
Looks good to me! Some overarching design questions and a nit, but I feel that the design questions are secondary (though I'd appreciate answers of course!). 

Everything looks correct from a code perspective!

vault/expiration.go
swayne275: 894f249bab

vault/expiration.go
swayne275: in the current code a lease can only become a zombie by failing revocation a certain number of times. in future code, it can also become a zombie if the backing data store tells us it failed to be revoked and is unlikely to be revoked without manual intervention (this is what VAULT-1978 will set us up for).

The goal of zombification is to save cycles attempting to revoke leases that we know can not likely be automatically revoked. A lot of this lease revocation happens shortly after vault startup, and we would like to not attempt revocation if it's likely to fail. Therefore, we must persist some information about the lease (indicating revocation failure) so that on the next vault start we don't attempt to re-revoke it. In bad cases, there can be 100s of 1000s of zombie leases (e.g. if  a lot were stored in a database which became misconfigured or was disconnected from the network).

Does that make sense?

swayne275: the zombie lease cache is actually more for the future API work than anything (zombie status is checked on startup when leases are being loaded). we had discussed removing the cache (therefore hitting storage every time an API request to list zombies is made), but we decided to keep the cache for now

vault/expiration.go
```diff
@@ -922,17 +938,9 @@ func (m *ExpirationManager) revokeCommon(ctx context.Context, leaseID string, fo
...
-		m.pending.Delete(leaseID)
-		m.leaseCount--
-		// Log but do not fail; unit tests (and maybe Tidy on production systems)
-		if err := m.core.quotasHandleLeases(ctx, quotas.LeaseActionDeleted, []string{leaseID}); err != nil {
-			m.logger.Error("failed to update quota on revocation", "error", err)
-		}
-	}
+	m.removeFromPending(ctx, leaseID)
 	m.nonexpiring.Delete(leaseID)
+	m.zombies.Delete(leaseID)
```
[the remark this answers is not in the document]
swayne275: if we get to this point we've either successfully revoked a lease, **or** we've failed to truly revoke a lease, but are forcing the revocation (deleting the lease from Vault, even if we aren't able to remove the actual lease from whereever it is stored)

to clear out zombie leases, you must use the force revoke option

vault/expiration_test.go
swayne275: I think the key difference is that zombies are a separate class of lease. On Vault start we load leases in from storage, and handle zombies completely differently than non-zombies. I think it makes it a little bit more clear and less prone to bugs to not re-use the pending map (since pending is so deeply integrated). The zombies also won't have a timer to trigger revocation attempts.

Basically the expiration manager is actively managing the pending leases, but is doing basically nothing with the zombie leases (besides (in the future) exposing them via an API call to be force revoked).

swayne275: to some extent you can think of marking a lease as a zombie as being similar to tomb-stoning in a database (except that zombies must be manually deleted). we want to get it out of rotation and there's no reason to expose it with all of the other valid leases in rotation

HridoyRoy (approved):

vault/expiration.go
```diff
@@ -2275,6 +2300,44 @@ func (m *ExpirationManager) walkLeases(walkFn leaseWalkFunction) error {
...
+	if le.isZombie() {
+		m.logger.Info("attempted to re-mark lease as zombie", "original_error", le.RevokeErr, "new_error", err.Error())
+		return
+	}
+
+	errStr := err.Error()
+	if len(errStr) == 0 {
+		errStr = "no error message given"
+	}
+	if len(errStr) > maxZombieErrorLength {
```
sgmiller, line 2329: Could we 'intern' these strings, e.g dedupe them in a separate data structure?  Or is that overkill and we're confident in the memory usage being limited by character length without losing info?

sgmiller (approved):
Small comment but otherwise lgtm

vault/expiration.go
swayne275: hmm, i'm not 100% sure what you're asking. could you phrase it differently or expand on it?

the main concern here is we don't want to accidentally over-consume storage (and memory) with many zombies that might have long (i.e. wrapped) errors

swayne275:
temporarily closing

swayne275:
closing to see if we can get tests running

swayne275:
closing for CI issues. copy at https://github.com/hashicorp/vault/pull/11452

PR #11342, opened 2021-04-12
TLS Diagnose Formatting Fixes 

HridoyRoy:
These are to address some comments on: https://github.com/hashicorp/vault/pull/11300

### DIFF
--- a/vault/diagnose/tls_verification_test.go
+++ b/vault/diagnose/tls_verification_test.go
@@ -28,7 +28,7 @@ func TestTLSValidCert(t *testing.T) {
 	}
 	err := ListenerChecks(listeners)
 	if err != nil {
-		t.Error(err.Error())
+		t.Fatalf(err.Error())
 	}
 }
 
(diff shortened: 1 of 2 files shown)


vault/diagnose/tls_verification_test.go
```diff
@@ -309,9 +309,9 @@ func TestTLSInvalidMaxVersion(t *testing.T) {
 	}
 	err := ListenerChecks(listeners)
 	if err == nil {
-		t.Error("TLS Config check on fake certificate should fail")
+		t.Fatalf("TLS Config check on fake certificate should fail")
 	}
 	if err.Error() != fmt.Errorf(maxVersionError, "0").Error() {
-		t.Errorf("Bad error message: %w", err)
```
swayne275, line 315: i think error here was fine since nothing after this depended on the results of that check, but fatal is good too!

swayne275 (approved):

PR #11345, opened 2021-04-12
feat(aws): add ability to provide a role session name when generating STS credentials

RXC001:
We have a need for an IAM session ARN to register an AWS CodeDeploy on-premises instance, more info [here](https://docs.aws.amazon.com/codedeploy/latest/userguide/register-on-premises-instance-iam-session-arn.html); the problem comes with renewing the temporary credentials. In Vault, the IAM session ARN associated with newly generated credentials (using [this endpoint](https://www.vaultproject.io/api/secret/aws#generate-credentials)) is dynamically built with the current time and a random int. Because of this, renewing temporary credentials using the same IAM session ARN would be impossible, as the ARN would be changed when Vault generates new credentials.

The proposed solution comes in two parts. The first is to add a new parameter called `role_session_name` when generating new credentials. With this, users can pass in a role session name and expect an ARN with a static role session name every time new STS credentials are read from Vault. If `role_session_name` isn't provided, then the previous behavior of dynamically creating a session name is followed.

The second part of this solution is to return the IAM session ARN generated when Vault assumed the role. In the return payload, the IAM session ARN will be in the `arn` field. This provides transparency on which IAM user/role the returned credentials are associated with.

### DIFF
--- a/changelog/11345.txt
+++ b/changelog/11345.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-feat(aws): add ability to provide a role session name when generating STS credentials
+secrets/aws: add ability to provide a role session name when generating STS credentials
 ```
--- a/changelog/11345.txt
+++ b/changelog/11345.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+feat(aws): add ability to provide a role session name when generating STS credentials
+```
--- a/changelog/11345.txt
+++ b/changelog/11345.txt
@@ -1,3 +1,3 @@
-```release-note:feature
+```release-note:improvement
 feat(aws): add ability to provide a role session name when generating STS credentials
 ```
--- a/builtin/logical/aws/path_user.go
+++ b/builtin/logical/aws/path_user.go
@@ -33,6 +33,10 @@ func pathUser(b *backend) *framework.Path {
 				Description: "Lifetime of the returned credentials in seconds",
 				Default:     3600,
 			},
+			"role_session_name": &framework.FieldSchema{
+				Type:        framework.TypeString,
+				Description: "Session name to use when assuming role. Max chars: 64",
+			},
 		},
 
 		Callbacks: map[logical.Operation]framework.OperationFunc{
@@ -81,6 +85,7 @@ func (b *backend) pathCredsRead(ctx context.Context, req *logical.Request, d *fr
 	}
 
 	roleArn := d.Get("role_arn").(string)
+	roleSessionName := d.Get("role_session_name").(string)
 
 	var credentialType string
 	switch {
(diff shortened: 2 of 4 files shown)


tomhjp (commented):
Hi @RXC001, thanks for the contribution! I have a few minor suggestions, but a couple of questions too. Forgive me if I'm missing some context as CodeDeploy is new to me.

* You mention you can't renew the generated temporary credentials, but I'm unsure why you can't renew them using the Vault lease ID: https://www.vaultproject.io/api-docs/system/leases#renew-lease, the equivalent of `vault lease renew aws/creds/my-role/Dmlmy...` using the CLI. With the features in this PR, how will you use the ARN to renew the credentials; will it be through Vault or a different mechanism?
* Assuming you can't use Vault's built-in lease renewal, why do you need the generated ARN to be stable in addition to having the ARN returned in the response? As above, I don't have context on how you plan to handle renewals, but it seems to me like either feature would be enough for you to know the ARN. I feel like I'm probably missing a separate requirement here.

ibspoof:
I work with @RXC001 and was involved in the feature and code.

@tomhjp Will have to check on lease renew method.  We are using Vault Agent and haven't tried that method before.

> Assuming you can't use Vault's built-in lease renewal, why do you need the generated ARN to be stable in addition to having the ARN returned in the response? As above, I don't have context on how you plan to handle renewals, but it seems to me like either feature would be enough for you to know the ARN. I feel like I'm probably missing a separate requirement here.

CodeDeploy registers an on premise instance based on the [session arn](https://docs.aws.amazon.com/codedeploy/latest/userguide/register-on-premises-instance-iam-session-arn.html#register-on-premises-instance-iam-session-arn-5) (role arn + sessionid) and as long as the session arn remains the same across multiple STS provided access key, secret key, session token we can refresh the aws credentials without re-registering the instance. This is why we require a session arn that has a consistent session name associated.  

We also have a need outside of code deploy to be able to set a session name on our end when creating new STS creds to track the usage for the specific client in AWS CloudTrail as we have multiple VMs using the same AppRole and therefore currently the same session name with timestamp added.  Right now we have to correlate Vault Audit logs and and timestamps to getting STS creds to find one when X VM requested from vault and then match to cloudtrail.  From a security standpoint this is both painful and not a process we can automate.  With having a settable session name that vault passes through to AWS through VaultAgent we can easily tag the instance requesting the sts creds and match the usage in CloudTrail with a quickly identifying session name.  We can also trigger alerts in AWS for non expected session name or a session coming from an unexpected IP with ease.

The reason we added the arn to the response is:
1) In the current setup vault truncates the session arn with a warning, but doesn't return what the session arn is requiring an additional call to AWS APIs to retrieve the session name vault generated
2) Allows for validation by the client the session arn/name in the assumed role matches the value expected from vault without having to call AWS API to get the arn
3) The key details are already present in the response of the key, secret, and session token adding the arn isn't leaking any information

tomhjp:
Thanks for that extra context @ibspoof, that makes sense and I can see how this all fits together now. The lease renewal can be considered a separate issue as far as I'm concerned. I'll try and get this merged, although it looks like we have some unrelated failures in CI 😞

tomhjp (approved):

changelog/11345.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+feat(aws): add ability to provide a role session name when generating STS credentials
```
kalafut, line 2:
```suggestion
secrets/aws: add ability to provide a role session name when generating STS credentials
```

ibspoof:
@tomhjp Thanks!

Issue #9943, opened 2020-09-14
version_base is stuck to 1.5.0 for the entire 1.5.x release


aieri:
Any vault version from the 1.5.x release series reports itself as being version 1.5.0.

This appears to be due to [version_base](https://github.com/hashicorp/vault/blob/v1.5.3/sdk/version/version_base.go#L111) not having been updated as minor versions got released.

mladlow:
Hi @aieri,

Thanks for flagging this. It's an oversight on our part and we'll fix it for the next 1.5 release.

mladlow:
Hey @aieri - no worries if you don't get back to this, but do you happen to have or remember a description of how this was reported? I believe it should be corrected now for Vault versions after 1.5.4, but even with earlier 1.5 versions, `vault version` would report the correct version.

We're trying to understand where the incorrect version was visible.

aieri:
Hi @mladlow, no problem: we had originally (incorrectly) reported this as a bug in the nrpe check deployed by the vault charm. See https://bugs.launchpad.net/vault-charm/+bug/1895548.

The incorrect response was being returned when [querying via curl](https://bugs.launchpad.net/vault-charm/+bug/1895548/comments/1).

ncabatoff:
I'm unable to reproduce this issue.  I've tested on 1.5.2 and 1.5.3.  Here's an example with the former:

```
==> Vault server configuration:

             Api Address: http://0.0.0.0:8200
                     Cgo: disabled
         Cluster Address: https://0.0.0.0:8201
              Go Version: go1.14.7
              Listener 1: tcp (addr: "0.0.0.0:8200", cluster address: "0.0.0.0:8201", max_request_duration: "1m30s", max_request_size: "33554432", tls: "disabled")
               Log Level: trace
                   Mlock: supported: false, enabled: false
           Recovery Mode: false
                 Storage: inmem
                 Version: Vault v1.5.2
             Version Sha: 685fdfa60d607bca069c09d2d52b6958a7a2febd
...

$ curl -s http://127.0.0.1:8200/v1/sys/health?standbyok=true 
{"initialized":true,"sealed":false,"standby":false,"performance_standby":false,"replication_performance_mode":"disabled","replication_dr_mode":"disabled","server_time_utc":1618250753,"version":"1.5.2","cluster_name":"vault-cluster-9bdad743","cluster_id":"44c6fe1b-f626-a7dd-ed9b-338e9a6018c4"}
```

[...]

PR #11350, opened 2021-04-13
Remove hard-coded version string.  

ncabatoff:
We'll rely solely on ldflags set by packagespec.  Doing this on 1.7 to start with so that I can prove the concept before applying it more widely.

### DIFF
--- a/sdk/version/version_base.go
+++ b/sdk/version/version_base.go
@@ -8,7 +8,7 @@ var (
 	// Whether cgo is enabled or not; set at build time
 	CgoEnabled bool
 
-	Version           = "1.7.0"
+	Version           = ""
 	VersionPrerelease = ""
 	VersionMetadata   = ""
 )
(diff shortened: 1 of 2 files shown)


raskchanky (approved):

calvn:
@ncabatoff we do a version constraint check [in here](https://github.com/hashicorp/vault/blob/master/sdk/helper/pluginutil/env.go#L50-L63), not sure if this new value would break that.

ncabatoff:
> @ncabatoff we do a version constraint check in here, so I think this new value would break that.

@calvn The tests pass, so at the moment it doesn't break anything, though I agree if we added more comprehensive tests for plugins it might.

The `TBD` value won't be used in any Vault binaries.  The new plan is to have the version that's checked into Git be managed via `Makefile`.  This doesn't get used with our official packagespec-built binaries, but they'll do the equivalent of what the Makefile does.  The Makefile change is just for the sake of non-HashiCorp binaries of Vault, as I believe was the case in https://github.com/hashicorp/vault/issues/9943, otherwise we'd rely solely on the packagespec inputs and not have anything in Git anymore.

We're moving the version value from the .go file to the Makefile just to make the process of updating it a tiny bit easier, so that we don't have to `go mod vendor`.

vishalnayak (approved):

PR #11357, opened 2021-04-14
Grammar typo fix

hellobontempo:
Hi there! First time contributing, just noticed this _possible_ grammar error and made a PR for clarification. 

Is `certificate/key` intended to be interchangeable (as in "certificate or key") or are both created?

If the latter, then perhaps adjust text to explicitly say `certificate and key` ?

Cheers and thanks!
Claire

Also thank you @s-christoff for helping navigate HashiCorp contributing docs


hashishaw (approved):
Thanks for the catch and quick bugfix!

PR #11358, opened 2021-04-14
Update Changelog

kalafut:
These two C/L entries were not backported to the 1.7 release branch.


mladlow (approved):
Should we remove changelog/10644.txt if it exists?

kalafut:
> Should we remove changelog/10644.txt if it exists?

Yep, good call.

PR #11363, opened 2021-04-15
issue-11067 updated region eu-west-2 with credential and region

prassawant:
what I found out that in default.go the sts services is not defined for eu-west-2, so added the code


vishalnayak:
This feels like it should go in as a PR on the upstream repository to update this file. https://github.com/aws/aws-sdk-go/blob/main/aws/endpoints/defaults.go

Then, the dependency to that library in Vault can be updated to bring in that change.

We don't manually make any changes to the vendored files. Sorry.

PR #11365, opened 2021-04-15
Validate hostnames when using TLS in Cassandra

pcman312:
Updates Cassandra database engine and physical backend to validate hostnames when configured to use TLS.

This also updates the gocql library to the latest version.


plugins/database/cassandra/connection_producer.go
```diff
@@ -269,6 +230,50 @@ func (c *cassandraConnectionProducer) createSession(ctx context.Context) (*gocql
...
+			return nil, fmt.Errorf("found certificate for TLS authentication but no private key")
+		}
+
+		parsedCertBundle, err := certBundle.ToParsedCertBundle()
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse certificate bundle: %w", err)
+		}
+
+		tlsConfig, err := parsedCertBundle.GetTLSConfig(certutil.TLSClient)
+		if err != nil || tlsConfig == nil {
```
calvn, line 249: Looking at `GetTLSConfig()`, it doesn't seem that we can get a nil `tlsConfig` if the error is nil so a check on `err != nil` should be sufficient.

plugins/database/cassandra/connection_producer.go
```diff
@@ -269,6 +230,50 @@ func (c *cassandraConnectionProducer) createSession(ctx context.Context) (*gocql
 	return session, nil
 }
 
+func getSslOpts(certBundle *certutil.CertBundle, minTLSVersion string, insecureSkipVerify bool) (*gocql.SslOptions, error) {
+	var tlsConfig *tls.Config
```
calvn, line 234: Can we start off with `tlsConfig := &tls.Config{}` if this is never expected to be nil?

plugins/database/cassandra/connection_producer_test.go
```diff
@@ -0,0 +1,109 @@
...
+				cassandra.CopyFromTo(secureFileMounts),
+				cassandra.SslOpts(&gocql.SslOptions{
+					Config:                 &tls.Config{InsecureSkipVerify: true},
+					EnableHostVerification: false,
+				}),
+			)
+			defer cleanup()
+
+			// This is fine if it panics - the connection URL should include the port and if it doesn't, this should fail
+			port := strings.Split(connURL, ":")[1]
```
calvn, line 71: Should we handle this failure a bit more gracefully by doing a length check first instead of letting it panic?

helper/testhelpers/cassandra/cassandrahelper.go
```diff
@@ -2,43 +2,68 @@ package cassandra
...
+	"path/filepath"
 	"testing"
 	"time"
 
 	"github.com/gocql/gocql"
 	"github.com/hashicorp/vault/helper/testhelpers/docker"
 )
 
-func PrepareTestContainer(t *testing.T, version string) (func(), string) {
+type containerConfig struct {
```
calvn, line 15: This is nice!

plugins/database/cassandra/connection_producer.go
```diff
@@ -269,6 +230,50 @@ func (c *cassandraConnectionProducer) createSession(ctx context.Context) (*gocql
...
+		if certBundle.Certificate != "" && certBundle.PrivateKey == "" {
+			return nil, fmt.Errorf("found certificate for TLS authentication but no private key")
+		}
+
+		parsedCertBundle, err := certBundle.ToParsedCertBundle()
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse certificate bundle: %w", err)
+		}
+
+		tlsConfig, err := parsedCertBundle.GetTLSConfig(certutil.TLSClient)
```
austingebauer, line 248: Should this be using `=` instead of `:=` when assigning to `tlsConfig`? I think using `:=` here would result in a nil pointer dereference down on L256.

calvn: Nice catch!

austingebauer: Ya, was thinking it'd be nil pointer dereference if we don't enter that else branch (since certBundle != nil). In any case, I agree that the correct `tlsConfig` won't be set due to the scoping.

plugins/database/cassandra/connection_producer.go
kalafut: Funnily enough we just fixed this in another DB: https://github.com/hashicorp/vault/pull/10899/files

plugins/database/cassandra/connection_producer.go
pcman312: Good catch, fixed.

plugins/database/cassandra/connection_producer.go
pcman312: Yes. Fixed.

plugins/database/cassandra/connection_producer_test.go
pcman312: I don't think an additional check is valuable here because it's in a test and it needs to fail, so having it be a test failure rather than a panic doesn't really give us much. It's an irrecoverable error due to the test construct failing which seems like a panic is well suited for.

helper/testhelpers/cassandra/cassandrahelper.go
pcman312: Thanks! I figured since I needed to add another optional field that it would be better to switch to the functional arguments pattern for ease of use downstream.

plugins/database/cassandra/connection_producer.go
pcman312: I think you're right. Since both variables exist before this line they are shadowed by this call when using `:=`. Or at least that's what my IDE is telling me.

plugins/database/cassandra/connection_producer_test.go
pcman312: Further investigation revealed that panicking in a test does not behave the same as running `t.Fatal`. It will stop all tests that are running in the package. I didn't realize this was the behavior. I've updated this so the `PrepareTestContainer` function instead returns a new `Host` type that contains the hostname, the port, and has a function to build the connection URL. It also performs the equivalent check in `PrepareTestContainer` rather than forcing the callers to make any additional checks.

austingebauer (approved):
LGTM

PR #11371, opened 2021-04-16
Fix goroutine leak caused by updating rate quotas

ncabatoff:


vishalnayak (approved):

helper/metricsutil/wrapped_metrics.go
```diff
@@ -88,8 +88,9 @@ func (m *ClusterMetricSink) MeasureSinceWithLabels(key []string, start time.Time
 
 // BlackholeSink is a default suitable for use in unit tests.
 func BlackholeSink() *ClusterMetricSink {
-	sink, _ := metrics.New(metrics.DefaultConfig(""),
-		&metrics.BlackholeSink{})
+	conf := metrics.DefaultConfig("")
+	conf.EnableRuntimeMetrics = false
```
briankassouf, line 92: Was this intentionally disabled?

ncabatoff: Yes, to avoid it creating a goroutine that interferes with the leak test.  I believe BlackholeSink is only used in tests.

Issue #11178, opened 2021-03-23
Constant growing consul db size (seems related to entries not deleted in path sys/expire/token/$hex1/$hex2)


bom-d-van:
Hi,

We are seeing our vault storage backend consul snapshot size growing constantly. We notice it's since around the time that we started using AWS secret engine. Thus we are suspecting that it might be caused by sys/expire/token/ expiration clean-up logics not dealing entries created by **non-orphan batch token**.

1. In our vault+consul installation, we are seeing there are over 4.7 millions entries in consul within one path: `vault/sys/expire/token/h6600xxx/xxx`
2. And by reading raw value from some of the entries, it looks like an aws secrets lease id. 
3. However, if we try to look up the lease id, our vault instance returns: `{"errors":["invalid lease"]}`.
4. What's more, instead of having 4.7 million leases, there are only less than 96k aws leases under path `vault/sys/expire/id/aws`.

**To Reproduce**

1. Create a new aws secret engine: https://www.vaultproject.io/docs/secrets/aws#setup
2. Create a new aws role: https://www.vaultproject.io/api-docs/secret/aws#create-update-role
3. Create a non-orphan batch token using the root token: https://www.vaultproject.io/api-docs/auth/token#create-token
4. Use the batch token to create a new sts credential: https://www.vaultproject.io/api-docs/secret/aws#generate-credentials
5. After the lease returned by the previous call expired (or revoke it manually), the lease entry in sys/expire/id is deleted, but the corresponding entry in sys/expire/token isn't deleted.

**Expected behavior**

`vault/sys/expire/token/h6600xxx/xxx` entries created by using batch token should also be cleaned up.

**Environment:**

* Vault Server Version: 1.6.1 (6d2db3f033e02e70202bef9ec896360062b88b03)
* Vault CLI Version (retrieve with `vault version`): various versions, some are using vault cli (1.6.1), some are using go library github.com/hashicorp/vault (v0.11.1)
* Server Operating System/Architecture: CentOS 3.10.0

Vault server configuration file(s):

```hcl
max_lease_ttl = "219600h"

cluster_name = "production"

plugin_directory = "/etc/vault/plugin"

api_addr = "https://xxx:8200"

cluster_addr = "https://xxx:8201"

raw_storage_endpoint = true

listener "tcp" {
  address = ":8200"
  cluster_address = ":8201"
  tls_cert_file = "xxx"
  tls_key_file = "xxx"
}

telemetry {
[...]

PR #11377, opened 2021-04-16
On lease deletion, also delete non-orphan batch token parent index

ncabatoff:
…ries for non-orphan batch tokens.

### DIFF
--- a/vault/token_store.go
+++ b/vault/token_store.go
@@ -1122,7 +1122,7 @@ func (ts *TokenStore) lookupTainted(ctx context.Context, id string) (*logical.To
 	return ts.lookupInternal(ctx, id, false, true)
 }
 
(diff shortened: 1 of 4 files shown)


ncabatoff:
Fixes #11178

vault/token_store.go
```diff
@@ -1160,7 +1170,6 @@ func (ts *TokenStore) lookupBatchToken(ctx context.Context, id string) (*logical
 		}
 	}
 
-	te.ID = id
```
vishalnayak, line 1163: I wonder why this was here in the first place.

vishalnayak: Oh, you have moved this to the internal func. Cool.

vishalnayak (approved):

briankassouf (approved):

HridoyRoy (approved):

PR #11378, opened 2021-04-16
Cassandra: root_rotation_statements workaround for special chars

pcman312:
Updates the docs for Cassandra DB engine to include a workaround for using special characters in the root username.


Valarissa (approved):
LGTM!

PR #11389, opened 2021-04-16
Go upgrade for 1.5.x

HridoyRoy:


samsalisbury (changes requested):
Hi @HridoyRoy this looks right in terms of the files that are manually updated, however you will also need to run `make packages` and commit the changes, since the spec file (packages-oss.yml) was updated.

The `build-common-layers` job is also failing because this branch is on an older version of Packagespec that points to the old (now unsupported) jfrog Docker registry. Therefore this branch will also need upgrading to packagespec v0.2.2 which points at the new Docker registry mirror. To upgrade:

1. Download [packagespec v0.2.2](https://github.com/hashicorp/packagespec/releases/tag/v0.2.2) and unzip it into your path.
2. Run `packagespec version` to check that you have v0.2.2 installed correctly.
3. On this branch run `packagespec upgrade -version 0.2.2` and commit the changes.

Please @ me on Slack if you have any issues running the above commands.

HridoyRoy:
Closing as duplicate.

PR #11397, opened 2021-04-19
Update 1.5.x go 1.14.15

vishalnayak:


scripts/cross/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN apt-get update -y && apt-get install -y -q nodejs yarn
 RUN rm -rf /var/lib/apt/lists/*
 
 
-ENV GOVERSION 1.13.8
+ENV GOVERSION 1.14.15
```
vishalnayak, line 24: This isn't a documented change and also felt a bit off because its modified from `1.13.8` and not `1.14.7`. Is this correct? Unlike the newer release branches, there isn't a Dockerfile in `release/1.4.x/scripts/docker`, which is why I went this route, but I am unsure.

vishalnayak: I'm reverting this, as previous PRs on this branch to update Go didn't modify this file.

briankassouf (approved):

PR #11413, opened 2021-04-19
Document unauth pprof and the new pprof endpoints.

ncabatoff:


website/content/api-docs/system/pprof.mdx
```diff
@@ -26,6 +26,44 @@ $ curl \
...
+    --header "X-Vault-Token: ..." \
+    http://127.0.0.1:8200/v1/sys/pprof/allocs
+```
+
+## Block
+
+This endpoint returns a sampling of goroutines involved in blocking on
+synchronization primitives.
+
+It is included for completeness, but since Vault doesn't normally enable
```
vishalnayak, line 51: Will Vault allow collection of this data in any circumstance (if not normally)?

website/content/api-docs/system/pprof.mdx
```diff
@@ -51,6 +89,12 @@ This endpoint returns stack traces of all current goroutines.
 | :----- | :--------------------- |
 | `GET`  | `/sys/pprof/goroutine` |
 
+### Parameters
+
+- `debug` `(int: 0)` - Specifies special arguments for the collection.
```
vishalnayak, line 94: Do we want to default to 2?

website/content/api-docs/system/pprof.mdx
```diff
@@ -75,6 +119,26 @@ $ curl \
     http://127.0.0.1:8200/v1/sys/pprof/heap
 ```
 
+## Mutex
+
+This endpoint returns a sampling of goroutines holding contended mutexes.
+
+It is included for completeness, but since Vault doesn't normally enable
```
vishalnayak, line 126: Same question here. Do we need to satisfy completeness if this is not going to return anything?

website/content/api-docs/system/pprof.mdx
ncabatoff: For now only in custom builds, but I intend to add config settings to configure profiling rate.

website/content/api-docs/system/pprof.mdx
ncabatoff: I don't feel strongly one way or another.  But since this is a docs PR, if you think there's value in that, you should open a ticket.  Note however that most users aren't going to use this endpoint directly, they're going to use `vault debug`.  Also this would break backwards compat, though probably not a big deal here.

vishalnayak: Ah, I see. It would indeed break backwards compatibility. Leaning towards leaving it as-is.

website/content/api-docs/system/pprof.mdx
vishalnayak: I see. Sounds good. You may want to mention its usefulness in the current scenario and the possibility that you envision, to avoid people thinking that this is useless.

website/content/api-docs/system/pprof.mdx
vishalnayak: Addressed in the previous comment. Marking as resolved.

website/content/api-docs/system/pprof.mdx
ncabatoff: Well, it is useless for our primary docs audience atm.  And we don't usually talk about future plans in our docs.

raskchanky (approved):

totongcute (commented):
hi

PR #11430, opened 2021-04-21
builtin: deprecate errwrap.Wrapf() throughout

alrs:
`github.com/hashicorp/errwrap` has deprecated `Wrapf()`, and recommends using `fmt.Errorf()`.

`fmt.Errorf()` added support for error wrapping in go1.13. Vault has a hard requirement of go1.16.2, so this isn't an issue.

This PR deprecates `errwrap.Wrapf` throughout `github.com/hashicorp/vault/builtin` and all of its subpackages. I organized the PR as one-commit-per-package to aid anyone who should ever need to do a `git bisect`, but I'm just as happy to squash this into a single commit.

### DIFF
--- a/audit/format.go
+++ b/audit/format.go
@@ -10,7 +10,6 @@ import (
 
 	squarejwt "gopkg.in/square/go-jose.v2/jwt"
 
-	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/vault/helper/namespace"
 	"github.com/hashicorp/vault/sdk/helper/salt"
 	"github.com/hashicorp/vault/sdk/logical"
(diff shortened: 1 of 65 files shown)


ncabatoff:
Hey @alrs,

This is great, thanks!  I'm going to have to do some investigation / due diligence to reassure myself that this won't break anything, but I hope to merge this soon.

alrs:
> I'm going to have to do some investigation / due diligence to reassure myself that this won't break anything

Makes sense to me, thanks for taking a look at this instead of letting it sink to the bottom of the tank. `:)`.

ncabatoff (approved):
I didn't find a good example of a changed line in this PR that was being consumed by other code which was testing for an embedded error.  So to reassure myself I modified reloadutil.Reload to use fmt.Errorf, and checked that TestReload_KeyWithPassphrase still worked. Since all the "query" errwrap func are built on errwrap.Walk, I don't feel the need to check each of them.  And of course there were already tests in https://github.com/hashicorp/errwrap/pull/9, I just wanted to be doubly sure.

PR #11437, opened 2021-04-22
Fix endpoint path in example

wiget:
It's small bug, but took me few hours to get correct endpoint. So let's save someone else time.


sgmiller (approved):
Thanks for the spot!

PR #11439, opened 2021-04-22
Do not allow OIDC key rotation to be more frequent than the RollbackManager/PeriodicFunc will enforce

### LINKED ISSUE #11438: OIDC key rotation periods below 2m are not honored
**Describe the bug**
The identity backend allows users to configure signing key rotations as [frequently as once a minute](https://github.com/hashicorp/vault/blob/a24653cc5cca475962b27b71dcba29a4aa94d6ee/vault/identity_store_oidc.go#L463), but in practice cannot rotate keys more frequently than 1m, and will only be rotated every 2 minutes.

I believe this is due to the key rotation being handled by the `PeriodicFunc` for the identity backend, which is in turn managed/scheduled by the RollbackManager, which is fixed at a once a minute tick speed: https://github.com/hashicorp/vault/blob/a24653cc5cca475962b27b71dcba29a4aa94d6ee/vault/rollback.go#L19

when the periodic frequency is set to 60-121 seconds (< 120), when the rollback manager runs at time + 60 seconds, it is not time to rotate the keys yet. It isn't until the RollbackManager ticks a second time (time + 120 total seconds) that the [key will be eligible for rotation](https://github.com/hashicorp/vault/blob/a24653cc5cca475962b27b71dcba29a4aa94d6ee/vault/identity_store_oidc.go#L1546-L1558)

**To Reproduce**
run a vault dev server with debug log level set:
```
vault server -dev -dev-root-token-id=dev-root-token  -log-level=debug                  
```

In another terminal/tab configure a signing key with a rotation period of 1m (or any value from 60-119s):
```
VAULT_ADDR=http://127.0.0.1:8200
VAULT_TOKEN=dev-root-token
vault write identity/oidc/key/a-key rotation_period=1m verification_ttl=1m allowed_client_ids="*"  
```

if you watch the dev server logs at that point, it will print rotation/deletion secrets every 2 minutes, rather than every 1 minute:
```
2021-04-22T11:42:53.687-0400 [DEBUG] secrets.identity.identity_218e930c: rotating OIDC key: key=notary
2021-04-22T11:44:53.688-0400 [DEBUG] secrets.identity.identity_218e930c: rotating OIDC key: key=notary
2021-04-22T11:44:53.792-0400 [DEBUG] secrets.identity.identity_218e930c: deleted OIDC public key: key_id=080f4b88-8104-8159-97b9-542d07195614
2021-04-22T11:46:53.687-0400 [DEBUG] secrets.identity.identity_218e930c: rotating OIDC key: key=notary
2021-04-22T11:46:53.959-0400 [DEBUG] secrets.identity.identity_218e930c: deleted OIDC public key: key_id=a79baa4a-c76c-a38e-9968-58d436a56525
2021-04-22T11:48:53.687-0400 [DEBUG] secrets.identity.identity_218e930c: rotating OIDC key: key=notary
2021-04-22T11:48:54.010-0400 [DEBUG] secrets.identity.identity_218e930c: deleted OIDC public key: key_id=91eacb84-9cf5-d2e1-b973-9fc61aa09c41
2021-04-22T11:50:53.688-0400 [DEBUG] secrets.identity.identity_218e930c: rotating OIDC key: key=notary
2021-04-22T11:50:53.907-0400 [DEBUG] secrets.identity.identity_218e930c: deleted OIDC public key: key_id=0ad16d14-3a2a-2db4-ce68-93c82488c122
```

Additionally, when polling the server's JWKS endpoint with curl, key ids will be visible for longer than the configured (`verification_ttl` + `rotation_period`) time (1m each, for a total of 2m) that they should be visible for
```
watch -n10 'curl --silent $VAULT_ADDR/v1/identity/oidc/.well-known/keys | jq -r ".keys[].kid"'
```

**Expected behavior**
Keys configured to rotate every minute are rotated every minute, or Vault does not allow rotation frequency to be set lower than 2 minutes 

**Environment:**
* Vault Server Version (retrieve with `vault status`): v1.6.2
* Vault CLI Version (retrieve with `vault version`): v1.6.2
* Server Operating System/Architecture: OS X

_none_, can be done using `vault server -dev`

**Additional context**
It is unlikely this causes many problems in practice -- setting signing key rotation to a period of 1m seems surprisingly frequent.

ianferguson:
Fixes #11438

#11438 documents how Vault will currently not rotate an OIDC signing key more frequently than every 2 minutes.

This PR addresses that by altering the current configuration check on rotation period from a minimum of 1 minute to a minimum of 2 minutes

### DIFF
--- a/vault/identity_store_oidc.go
+++ b/vault/identity_store_oidc.go
@@ -459,8 +459,8 @@ func (i *IdentityStore) pathOIDCCreateUpdateKey(ctx context.Context, req *logica
 		key.RotationPeriod = time.Duration(d.Get("rotation_period").(int)) * time.Second
 	}
 
-	if key.RotationPeriod < 1*time.Minute {
-		return logical.ErrorResponse("rotation_period must be at least one minute"), nil
+	if key.RotationPeriod < 2*time.Minute {
+		return logical.ErrorResponse("rotation_period must be at least two minutes"), nil
 	}
 
 	if verificationTTLRaw, ok := d.GetOk("verification_ttl"); ok {


ianferguson:
Thinking about this some more over lunch, and I think this may not be a good fix at all. I think the semantic around signing key rotation is that it will be rotated at least as frequently as specified, rather than no more frequently than specified.

i.e. if the rotation frequency is 24 hours, it is better to rotate at 23 hours, 59 minutes and 59 seconds than at 24 hours and 1 second since the last rotation, all things considered

If that is the case, I think the correct fix is to update the [scheduling logic](https://github.com/hashicorp/vault/blob/master/vault/identity_store_oidc.go#L1542-L1559) to rotate any keys _due to be rotated within the next minute_, or more simply to remove 1 minute (or a value derived from the RollbackManager period setting) from the  nextRotation calculation here: https://github.com/hashicorp/vault/blob/20492a144d747c75df56fe9b528797f992adf249/vault/identity_store_oidc.go#L1269, which would ensure that for lower key rotation periods like 2m, that vault would rotate _at least_ every 2 minutes rather than every 2-3 minutes

ianferguson:
After further investigation I don't think this fix is correct, and have updated #11439 with additional context on the impact

PR #11442, opened 2021-04-22
UI/okta duo push notification

Monkeychip:
This PR adds a push notification for Okta authentication.  There have been issues with people waiting for the loader, not knowing that it was actually the loader waiting on them to hit their okta push notifications.

**Functionality**: If you select the auth method "okta" a function called `delayAuthMessageReminder` is fired _concurrently_ while the `authenticate` concurrency task is also running.  If after five seconds the `authenticate` concurrency task is still running a message appears to remind folks to check their push notifications.  If the auth method takes less than five seconds (ex: you have okta setup with no MFA and its successful) you won't see the message reminder.

Here is a gif of a situation in which we wait 90 seconds for someone to hit their push notifications. 

Here is a gif of a situation where it takes less than 5 seconds to login to okta and you do not see the message.

Here is a screenshot of the message:

**Note:**
While I was in there, I removed the overlay loader per Design's request— we show this with a grey out readOnly button that has a spinning loading inside of it.  I also cleaned up the look of the `AlertInline` component that display messages for the JWT method and for under "More options".  The messages now match the font-size and padding that Design had specked the Okta notification message for.

### DIFF
--- a/ui/app/components/auth-form.js
+++ b/ui/app/components/auth-form.js
@@ -1,3 +1,4 @@
+import Ember from 'ember';
 import { next } from '@ember/runloop';
 import { inject as service } from '@ember/service';
 import { match, alias, or } from '@ember/object/computed';
@@ -241,6 +242,11 @@ export default Component.extend(DEFAULTS, {
   }).withTestWaiter(),
 
   delayAuthMessageReminder: task(function*() {
+    if (Ember.testing) {
+      this.showLoading = true;
+      yield timeout(0);
+      return;
+    }
     yield timeout(5000);
   }),
 
--- a/ui/app/components/auth-form.js
+++ b/ui/app/components/auth-form.js
@@ -1,3 +1,4 @@
+import Ember from 'ember';
 import { next } from '@ember/runloop';
 import { inject as service } from '@ember/service';
 import { match, alias, or } from '@ember/object/computed';
@@ -210,6 +211,10 @@ export default Component.extend(DEFAULTS, {
 
   delayPushMessageReminder: task(function*(backendType) {
     if (backendType === 'okta') {
+      if (Ember.testing) {
+        this.showPushNotificationMessage = true;
+        return;
+      }
       yield timeout(3000); // wait 3 seconds before displaying reminder about checking for push notifications
     }
   }),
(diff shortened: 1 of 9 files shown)


ui/app/templates/components/auth-form.hbs
```diff
@@ -1,9 +1,4 @@
 <div class="auth-form">
-  {{#if showLoading}}
```
Monkeychip, line 2: Removed per Design's request. We show a loader on the submit button and grey it out.

kalafut:
Hi @Monkeychip. I'd like to chat about this since the design has moved a bit from the original idea of displaying a message after a certain amount of time.

hashishaw (approved):

ui/app/components/auth-form.js
```diff
@@ -237,6 +241,15 @@ export default Component.extend(DEFAULTS, {
     }
   }).withTestWaiter(),
 
+  delayAuthMessageReminder: task(function*() {
+    if (Ember.testing) {
+      this.showLoading = true;
```
hashishaw, line 246: Why do we showLoading in testing scenario but not otherwise?

ui/app/styles/core/message.scss
```diff
@@ -119,6 +119,15 @@
   &.has-top {
     margin-top: 1rem;
   }
+
+  &.size-small {
+    padding-bottom: $size-8;
```
hashishaw, line 124: Mixing the padding into the classname `size-small` feels a bit conflated to me. I wonder if instead we should keep all padding/margins out of this component, and instead wrap with a div that has the padding? especially since there's no background color here, it should probably be a margin anyway.

ui/tests/acceptance/auth-test.js
```diff
@@ -105,4 +105,17 @@ module('Acceptance | auth', function(hooks) {
     await visit('/vault/access');
     assert.dom('[data-test-allow-expiration="true"]').doesNotExist('hides beacon when the api is used again');
   });
+
+  test('it shows the push notification warning only for okta auth method after submit', async function(assert) {
```
hashishaw, line 109: 👏

hashishaw (commented):
A couple questions here, likely not blocking but I'd like to hear your thoughts before I approve!

ui/app/components/auth-form.js
Monkeychip: Because conditionally, to show the message we have to have both showLoading === true &&& this task function no longer running.

hashishaw (approved):
Great work! Thanks 🚀

PR #11445, opened 2021-04-23
LifetimeWatcher should retry renew failures until end of lease

ncabatoff:
This is based on @andrejvanderzee's work in #11008.


andrejvanderzee:
Any reason why you are using `github.com/cenkalti/backoff/v3` instead of `github.com/cenkalti/backoff/v4`?

ncabatoff:
> @ncabatoff Any reason why you are using github.com/cenkalti/backoff/v3 instead of github.com/cenkalti/backoff/v4?

@andrejvanderzee We're already using v3 elsewhere in Vault and I didn't see anything revolutionary in v4 I wanted.

api/lifetime_watcher.go
```diff
@@ -290,42 +292,67 @@ func (r *LifetimeWatcher) doRenew() error {
 					}
 				}
 
-				leaseDuration = fallbackLeaseDuration
+				// Calculate remaining duration until inital token lease expires
```
vishalnayak, line 295:
```suggestion
				// Calculate remaining duration until initial token lease expires
```

vishalnayak (approved):
Looking great!

andrejvanderzee:
Thank you @ncabatoff and @vishalnayak for picking this up!

tvoran (approved):

api/lifetime_watcher.go
```diff
@@ -290,42 +292,67 @@ func (r *LifetimeWatcher) doRenew() error {
...
+				r.calculateGrace(leaseDuration)
+			}
+			priorDuration = leaseDuration
 
-		// The sleep duration is set to 2/3 of the current lease duration plus
-		// 1/3 of the current grace period, which adds jitter.
-		sleepDuration := time.Duration(float64(leaseDuration.Nanoseconds())*2/3 + float64(r.grace.Nanoseconds())/3)
+			// The sleep duration is set to 2/3 of the current lease duration plus
+			// 1/3 of the current grace period, which adds jitter.
+			sleepDuration = time.Duration(float64(leaseDuration.Nanoseconds())*2/3 + float64(r.grace.Nanoseconds())/3)
```
tomhjp, line 354: Looks like the `sleepDuration` function below is unused? It looks safe to delete to me.

api/lifetime_watcher.go
```diff
@@ -270,6 +270,8 @@ func (r *LifetimeWatcher) doRenew() error {
 
 		var leaseDuration time.Duration
```
tomhjp, line 271: This is inherited, but there are a lot of duration variables, and I think renaming this one to `remainingLeaseDuration` would help make its intention and mutability a bit clearer. If you think that's too long, `lease` and `duration` are kind of implying the same concept each, and perhaps we could drop one of those words from each of the variables. WDYT?

api/lifetime_watcher.go
```diff
@@ -290,42 +292,67 @@ func (r *LifetimeWatcher) doRenew() error {
...
+			errorBackoff = nil
 
 			// Push a message that a renewal took place.
 			select {
 			case r.renewCh <- &RenewOutput{time.Now().UTC(), renewal}:
 			default:
 			}
 
+			// Reset initial time
+			initialTime = time.Now()
```
tomhjp, line 319: It might be nice to combine this with the block of code resetting the initial lease duration below (L328), as they are logically very closely related.

tomhjp (approved):
LGTM! Just some very minor suggestions.

api/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 replace github.com/hashicorp/vault/sdk => ../sdk
 
 require (
+	github.com/cenkalti/backoff/v3 v3.0.0
```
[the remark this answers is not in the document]
ncabatoff: We're already using v3 elsewhere and IIRC the differences between the two versions are disruptive yet insignificant.  Like adopting the new version would require changes elsewhere but doesn't fix any bugs or provide any new functionality.  At least that's my recollection from when I looked into it.

api/lifetime_watcher.go
ncabatoff: Removed.

api/lifetime_watcher.go
ncabatoff: Done.

api/lifetime_watcher.go
ncabatoff: I didn't like the suggestion when I read it, but I tried making the change and I liked it.

api/lifetime_watcher.go
```diff
@@ -268,18 +268,20 @@ func (r *LifetimeWatcher) doRenew() error {
 		default:
 		}
 
-		var leaseDuration time.Duration
+		var remainingLeaseDuration time.Duration
 		fallbackLeaseDuration := initialTime.Add(priorDuration).Sub(time.Now())
```
tomhjp, line 272: One last thought, I think the value of this only gets used in one place now, so you could delete the variable to simplify a bit.

jasonodonnell (approved):
LGTM and tested well in both K8s (with caching/templating) and locally.

PR #11452, opened 2021-04-23
Expiration Manager: Handle Presumed Irrevocable Leases Separately

swayne275:
Note: this is a copy of pull #11333 (@HridoyRoy @sgmiller the one you already approved). I had to make a new branch/pr for CI to work after the Okta changes

This PR allows leases that are deemed irrevocable to be marked as "zombies", taking them out of the expiration rotation. This is to avoid wasting work on leases that are unlikely to be revoked without manual intervention. This can happen if a backing storage system is removed or misconfigured (e.g. leases stored in a database that cannot be accessed anymore).

Leases are currently only marked as zombies if they fail revocation 6 times. In the future they will also be marked as zombies if the backing systems return an irrevocable error. See VLT-145 for more details.

Question: I don't think zombies need to be included in walkLeases or WalkTokens - these appear to be for metrics, and I don't think zombies should be included in the metrics for potentially valid leases (there are/will be separate zombie metrics - again, see RFC).

Question: for those of you that have worked on this in the past: please let me know if I "missed a spot" - from my analysis it looks like this should do it!

### DIFF
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -65,6 +65,10 @@ const (
 	numExpirationWorkersTest = 10
 
 	fairshareWorkersOverrideVar = "VAULT_LEASE_REVOCATION_WORKERS"
+
+	// limit zombie error messages to 240 characters to be respectful of storage
+	// requirements
+	maxZombieErrorLength = 240
 )
 
 type pendingInfo struct {
@@ -96,6 +100,11 @@ type ExpirationManager struct {
 	nonexpiring sync.Map
 	leaseCount  int
 	pendingLock sync.RWMutex
+
+	// Track expired leases that have been determined to be irrevocable (without
+	// manual intervention). These irrevocable leases are referred to as
+	// "zombies" or "zombie leases"
+	zombies sync.Map
 
 	// The uniquePolicies map holds policy sets, so they can
 	// be deduplicated. It is periodically emptied to prevent
@@ -210,6 +219,13 @@ func (r *revocationJob) OnFailure(err error) {
 
 	if pending.revokesAttempted >= maxRevokeAttempts {
 		r.m.logger.Trace("lease has consumed all retry attempts", "lease_id", r.leaseID)
+		le, loadErr := r.m.loadEntry(r.nsCtx, r.leaseID)
+		if loadErr != nil {
+			r.m.logger.Warn("failed to mark lease as zombie - failed to load", "lease_id", r.leaseID, "err", loadErr)
+			return
+		}
+
+		r.m.markLeaseAsZombie(r.nsCtx, le, errors.New("lease has consumed all retry attempts"))
 		return
 	}
 
@@ -824,6 +840,10 @@ func (m *ExpirationManager) Stop() error {
 		return true
 	})
 	m.uniquePolicies = make(map[string][]string)
+	m.zombies.Range(func(key, _ interface{}) bool {
+		m.zombies.Delete(key)
+		return true
+	})
 	m.pendingLock.Unlock()
 
 	if m.inRestoreMode() {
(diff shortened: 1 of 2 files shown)


swayne275:
closing to see if i can trigger circle

HridoyRoy (approved):

vault/expiration.go
```diff
@@ -210,6 +219,13 @@ func (r *revocationJob) OnFailure(err error) {
 
 	if pending.revokesAttempted >= maxRevokeAttempts {
 		r.m.logger.Trace("lease has consumed all retry attempts", "lease_id", r.leaseID)
+		le, loadErr := r.m.loadEntry(r.nsCtx, r.leaseID)
+		if loadErr != nil {
+			r.m.logger.Warn("failed to mark lease as zombie - failed to load", "lease_id", r.leaseID, "err", loadErr)
+			return
+		}
+
+		r.m.markLeaseAsZombie(r.nsCtx, le, errors.New("lease has consumed all retry attempts"))
```
briankassouf, line 228: If we're just going to be using the same error message for each lease, it will be cheaper to just set a boolean instead. Do you think we should switch?

vault/expiration.go
```diff
@@ -2291,6 +2316,47 @@ func (m *ExpirationManager) walkLeases(walkFn leaseWalkFunction) error {
...
+	}
+	if len(errStr) > maxZombieErrorLength {
+		errStr = errStr[:maxZombieErrorLength]
+	}
+
+	le.RevokeErr = errStr
+	m.persistEntry(ctx, le)
+
+	m.zombies.Store(le.LeaseID, le)
+	m.removeFromPending(ctx, le.LeaseID)
```
briankassouf, line 2356: If we remove from pending here we are no longer counting these leases in the `num_leases` metric and they no longer count towards lease quotas. Do you have thoughts on how zombie leases should interact with those two systems?

vault/expiration.go
```diff
@@ -2324,6 +2396,9 @@ func (le *leaseEntry) renewable() (bool, error) {
 	case le == nil:
 		return false, fmt.Errorf("lease not found")
 
+	case le.isZombie():
+		return false, fmt.Errorf("lease is not renewable and has failed previous revocation attempts")
```
briankassouf, line 2400: `lease is expired and has failed previous revocation attempts`

vault/expiration.go
```diff
@@ -1859,6 +1879,11 @@ func (m *ExpirationManager) loadEntryInternal(ctx context.Context, leaseID strin
 	}
 	le.namespace = ns
 
+	if le.isZombie() {
+		m.zombies.Store(le.LeaseID, le)
+		return le, nil
+	}
```
briankassouf, line 1885: I wonder if we can push this into the updatePending function, which will allow us to have one place where we can update the pending, nonexpiring, and zombie maps.

vault/expiration.go
swayne275: hmm, what would this look like? I'm all for cheaper - even if that means just shortening the string in the error. good suggestion!

vault/expiration.go
swayne275: yes. i'm now thinking that zombies should be included in the lease count quota (mostly for storage reasons). i need to do some discovery on the `(ExpirationManager).leaseCount` system

vault/expiration.go
swayne275: good copy pasta catch!

vault/expiration.go
swayne275: https://hashicorp.atlassian.net/browse/VAULT-2303

vault/expiration.go
swayne275: https://hashicorp.atlassian.net/browse/VAULT-2303

PR #11454, opened 2021-04-23
Create helpers which integrate with OpenTelemetry for diagnose collection

sgmiller:


vault/diagnose/output.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if r.Message != "" || len(r.Warnings) > 0 {
+		sb.WriteString(": ")
+	}
+	sb.WriteString(r.Message)
+	for _, w := range r.Warnings {
+		for i := 0; i < depth; i++ {
+			sb.WriteRune('\t')
+		}
+		sb.WriteString("  ")
```
swayne275, line 166: I'm working on visualizing what is going on here.

It looks like general nesting is happening with the `\t` parts, and then warnings are further nested under the `r.Message` by writing the empty spaces.

Do we want the first of the warning to have a newline between it and the `r.Message`? It looks like there will just be spaces, then a warning (on the same line as the message), then more warnings on each line after.

vault/diagnose/output.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func (t *TelemetryCollector) OnStart(parent context.Context, s sdktrace.ReadWriteSpan) {
+	t.spans[s.SpanContext().SpanID()] = s
+
+}
+
+func (t *TelemetryCollector) OnEnd(e sdktrace.ReadOnlySpan) {
+	if !e.Parent().HasSpanID() {
+		// Deep first walk the span structs to construct the top down tree results we want
```
HridoyRoy, line 55: I'm a bit confused about how this works. Why does this call `getOrBuildResult` 2 levels deep, as opposed to recursively until `p == nil` ?

vault/diagnose/output.go
```diff
@@ -0,0 +1,174 @@
...
+					}
+				} else {
+					t.RootResult = r
+				}
+			}
+		}
+	}
+}
+
+func (t *TelemetryCollector) Shutdown(ctx context.Context) error {
```
HridoyRoy, line 72: Could you leave some quick comments on these stub methods detailing what we'd want them to do?

HridoyRoy (approved):
Looks great! Mainly left comments around adding some descriptions for some of the functions, and had a couple clarifying questions. Overall though, makes sense!

vault/diagnose/output.go
```diff
@@ -0,0 +1,174 @@
...
+	return nil
```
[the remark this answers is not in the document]
sgmiller: good call

vault/diagnose/output.go
sgmiller: Possibly, this was quick and dirty.  but that does look like a bug as written.  Maybe I should throw up a unit test so we can take a look.

vault/diagnose/helpers.go
```diff
@@ -0,0 +1,83 @@
...
+var tc *TelemetryCollector
+
+// Init initializes a Diagnose tracing session.  In particular this wires a TelemetryCollector, which
+// synchronously receives and tracks OpenTelemetry spans in order to provide a tree structure of results
+// when the outermost span ends.
+func Init() {
+	tc = NewTelemetryCollector()
+	//so, _ := stdout.NewExporter(stdout.WithPrettyPrint())
+	tp = sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.AlwaysSample()),
```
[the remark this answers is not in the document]
sgmiller: We'd only want to use this in Diagnose runs, likely, yes.  Not in general.

vault/diagnose/helpers.go
```diff
@@ -0,0 +1,83 @@
...
+	span.AddEvent(warningEventName, trace.WithAttributes(attribute.String("message", msg)))
+}
+
+func Action(actionName string) trace.LifeCycleOption {
+	return trace.WithAttributes(attribute.String(actionKey, actionName))
+}
+
+// Test creates a new named span, and executes the provided function within it.  If the function returns an error,
+// the span is considered to have failed.
+func Test(ctx context.Context, spanName string, function func(context.Context) error, options ...trace.SpanOption) error {
```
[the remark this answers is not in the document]
sgmiller: Test introduces a span, so you probably only need the one. You can of course have more spans nested under the Test (started by the function or sub-calls).

vault/diagnose/helpers.go
```diff
@@ -0,0 +1,83 @@
...
+// Fail records an error in the current span (but unlike Fail, doesn't set the overall span status to Error)
+func Error(ctx context.Context, err error, options ...trace.EventOption) error {
+	span := trace.SpanFromContext(ctx)
+	span.RecordError(err, options...)
+	return err
+}
+
+// Warn records a warning on the current span
+func Warn(ctx context.Context, msg string) {
+	span := trace.SpanFromContext(ctx)
```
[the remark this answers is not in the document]
sgmiller: These ones yes.  Let me throw up a unit test with a working example which may help.

vault/diagnose/helpers.go
```diff
@@ -0,0 +1,83 @@
...
+
+	span.AddEvent(warningEventName, trace.WithAttributes(attribute.String("message", msg)))
+}
+
+func Action(actionName string) trace.LifeCycleOption {
+	return trace.WithAttributes(attribute.String(actionKey, actionName))
```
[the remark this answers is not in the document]
sgmiller: It's actually from an earlier effort, removing.

vault/diagnose/output.go
sgmiller: It doesn't need to since the span map is all spans, so all nodes that are parents will be visited once.  Some Results will have already been built when they're visited if they're a parent to an earlier span.

vault/diagnose/output.go
swayne275: great!

vault/diagnose/helpers.go
```diff
@@ -0,0 +1,79 @@
...
+	tc = NewTelemetryCollector()
+	//so, _ := stdout.NewExporter(stdout.WithPrettyPrint())
+	tp = sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.AlwaysSample()),
+		//sdktrace.WithSpanProcessor(sdktrace.NewSimpleSpanProcessor(so)),
+		sdktrace.WithSpanProcessor(tc),
+	)
+	otel.SetTracerProvider(tp)
+	tracer = tp.Tracer("vault-diagnose")
+}
```
swayne275, line 38: oh i was also meaning to ask, why is this a public `Init()` (that presumably is called somewhere else) vs just using a go `init()`?

sgmiller: Because it shouldn't happen all the time, just on runs of diagnose.  It also may need to be more stateful than it currently is, but there are some considerations about the global OpenTelemetry trace provider in that case.

vault/diagnose/output.go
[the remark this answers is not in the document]
swayne275: did we decide to leave these (this and `ForceFlush`) as returning nil instead of an unimplemented error?

swayne275 (approved):
let me know your thoughts on that comment. it's not that big of a deal, but i'd prefer errors (unless that impl is coming like very shortly so it can't really become a potential issue)

command/operator_diagnose.go
```diff
@@ -145,87 +157,68 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
...
+		}
+		return diagnose.ListenerChecks(sanitizedListeners)
+	})
 
-	return 0
+	return diagnose.Test(ctx, "storage", func(ctx context.Context) error {
+		_, err = server.setupStorage(config)
+		return err
+	})
+	return nil
```
swayne275, line 223: hmm, it looks like you're returning right after another return?

swayne275 (changes requested):

command/operator_diagnose.go
swayne275: this github preview doesn't show enough code to show what i'm pointing out

HridoyRoy (approved):

swayne275 (approved):

vault/diagnose/output.go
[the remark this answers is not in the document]
sgmiller: Yes, they do get called as part of the lifecycle of a SpanProcessor, so returning an error would break stuff.

command/operator_diagnose.go
sgmiller: Can you take another look?  It's tricky because there is an anonymous function there as an argument to diagnose.Test, so you're probably seeing that functions return followed by the other one.

swayne275: Screen Shot 2021-04-29 at 9 41 55 AM


yeah, this is what i'm seeing. I'm not thinking about the `return err` bit inside the `func(ctx context.Context) error`, i'm thinking about the `return diagnose.Test...` on line 219

thoughts?

swayne275: basically i think the return nil is a noop - and i'm guessing the compiler doesn't complain about that if the tests are running?

PR #11455, opened 2021-04-23
Expose unknown fields and duplicate sections as diagnose warnings

sgmiller:


command/server/config_test_helpers.go
```diff
@@ -456,6 +457,41 @@ func testLoadConfigFile(t *testing.T) {
...
+		},
+	}
+	errors := config.Validate("./test-fixtures/config.hcl")
+
+outer:
+	for _, er1 := range errors {
+		for _, ex := range expected {
+			// Only test the string, pos may change
+			if ex.Problem == er1.Problem {
+				continue outer
```
swayne275, line 479: could these be refactored to just `break` and not have the tags?

Also, what are those tags called? I've only seen them with `GOTO`

swayne275: And I ask about refactoring (at least for this loop) because i think it might be easier for another eng to understand down the road (i think it would have taken me less time to get it)

internalshared/configutil/config.go
```diff
@@ -13,6 +14,10 @@ import (
 
 // SharedConfig contains some shared values
 type SharedConfig struct {
+	FoundKeys  []string               `hcl:",decodedFields"`
+	UnusedKeys map[string][]token.Pos `hcl:",unusedKeyPositions"`
```
swayne275, line 18: did you want this to be of type `UnusedKeyMap` as defined in `config_util.go`?

internalshared/configutil/config_util.go
```diff
@@ -3,15 +3,71 @@
...
 func (ec *EntSharedConfig) ParseConfig(list *ast.ObjectList) error {
 	return nil
 }
 
 func ParseEntropy(result *SharedConfig, list *ast.ObjectList, blockName string) error {
 	return nil
 }
+
+// Creates the ConfigErrors for unused fields, which occur in various structs
+func ValidateUnusedFields(unusedKeyPositions map[string][]token.Pos, source string) []ConfigError {
```
swayne275, line 39: did you want the type to be `UnusedKeyMap` here too?

swayne275: and in a few other places in this file

internalshared/configutil/config_util.go
```diff
@@ -3,15 +3,71 @@
...
+				Position: pos,
+			})
+		}
+	}
+	return errors
+}
+
+func UnusedFieldDifference(a, b map[string][]token.Pos, foundKeys []string) map[string][]token.Pos {
+	if a == nil {
+		return nil
```
swayne275, line 61: we have different behavior with a being nil vs b. could we document what a and b are supposed to be so that future users of this code know that order matters?

internalshared/configutil/telemetry.go
```diff
@@ -33,8 +34,10 @@ const (
 
 // Telemetry is the telemetry configuration for the server
 type Telemetry struct {
-	StatsiteAddr string `hcl:"statsite_address"`
-	StatsdAddr   string `hcl:"statsd_address"`
+	FoundKeys    []string               `hcl:",decodedFields"`
+	UnusedKeys   map[string][]token.Pos `hcl:",unusedKeyPositions"`
```
swayne275, line 38: same about `UnusedKeyMap`

sdk/helper/strutil/strutil.go
```diff
@@ -32,6 +32,16 @@ func StrListContains(haystack []string, needle string) bool {
 	return false
 }
 
+// StrListContainsCaseInsensitive looks for a string in a list of strings.
+func StrListContainsCaseInsensitive(haystack []string, needle string) bool {
```
swayne275, line 36: love the variable names :)

swayne275 (commented):
A lot of it looks good, just have a few comments

command/server/config_test_helpers.go
```diff
@@ -456,6 +457,41 @@ func testLoadConfigFile(t *testing.T) {
...
+func testUnknownFieldValidation(t *testing.T) {
+	config, err := LoadConfigFile("./test-fixtures/config.hcl")
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
+	expected := []configutil.ConfigError{
+		{
+			Problem:  "unknown field bad_value found in configuration",
+			Position: token.Pos{},
```
HridoyRoy, line 469: I don't think the Pos is being checked in the test. If possible, could we see this in action?

command/server/config_test.go
```diff
@@ -57,3 +57,7 @@ func TestConfigRaftRetryJoin(t *testing.T) {
 func TestParseSeals(t *testing.T) {
 	testParseSeals(t)
 }
+
+func TestUnknownFieldValidation(t *testing.T) {
```
HridoyRoy, line 61: This is great! Can we also have a test that checks the filename for unusedKeys is correct, for configs between multiple files?

HridoyRoy (commented):
Just a few comments. Was primarily wondering about the tests. Overall, looks awesome!

command/server/config_test_helpers.go
sgmiller: They're block labels.  The idea there was if you just break on the inner one you'll hit the t.Fatal, which isn't what we want.  That said I can do it with a found bool instead.

internalshared/configutil/config.go
sgmiller: Ah nice, I must have added that type later but didn't loop back.

command/server/config_test_helpers.go
swayne275: cool, i haven't seen that used before! closest i've seen is with goto. i'm gonna read about them!

do whatever you think is most readable, i'm not too concerned about this

sdk/helper/strutil/strutil.go
sgmiller: Yeah, that's pretty common in other APIs.  Suprised Go didn't use it for things like strings.Contains

command/operator_diagnose.go
```diff
@@ -155,7 +155,11 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
 		c.UI.Output(err.Error())
 		return 1
 	}
-
+	errors := config.Validate("test")
```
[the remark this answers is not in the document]
sgmiller: yep

internalshared/configutil/config_util.go
```diff
@@ -3,15 +3,71 @@
...
+	}
+	if b == nil {
+		return a
+	}
+	res := make(map[string][]token.Pos)
+	for k, v := range a {
+		if _, ok := b[k]; !ok && !strutil.StrListContainsCaseInsensitive(foundKeys, govalidator.UnderscoreToCamelCase(k)) {
```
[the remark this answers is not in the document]
sgmiller: Because HCL's AST refers to the struct member name, while the found keys uses the string from the .hcl file (in lower case with underscores)

command/server/config_test_helpers.go
swayne275: https://medium.com/golangspec/labels-in-go-4ffd81932339

HridoyRoy (dismissed):
Lgtm!

HridoyRoy (approved):
lgtm!

PR #11467, opened 2021-04-26
Consul TLS Checks for Diagnose [draft]

HridoyRoy:
This has TLS checks for Storage, HAStorage, and ServiceRegistration. 

The tests just ensure that the correct validation function is called in all scenarios by operator_diagnose.go. Detailed tests around TLS verification are in the diagnose package (previously merged PR).


command/operator_diagnose.go
```diff
@@ -217,12 +219,29 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
...
 	c.UI.Output(status_unknown + phase)
 	_, err = server.setupStorage(config)
+	storageErrors = append(storageErrors, err.Error())
 	if err != nil {
-		c.UI.Output(same_line + status_failed + phase)
-		c.UI.Output(err.Error())
+
+	}
+
+	// Initialize the Service Discovery, if there is one
```
sgmiller, line 232: If storage setup above fails with an error, does it make sense proceed to the tls portion?

command/server.go
```diff
@@ -1154,7 +1154,7 @@ func (c *ServerCommand) Run(args []string) int {
 	}
 
 	// Prevent server startup if migration is active
-	// TODO: how to incorporate this check into Diagnose?
+	// TODO: Use logs with OpenTelemetry to Integrate this into Diagnose
```
sgmiller, line 1157: Hopefully logs aren't needed, but just general otel instrumentation

serviceregistration/consul/consul_service_registration.go
```diff
@@ -178,47 +213,33 @@ func NewServiceRegistration(conf map[string]string, logger log.Logger, state sr.
 	}
 
 	if consulConf.Scheme == "https" {
+
```
sgmiller, line 216: I think this operation is quick enough to do no matter what.  Then if we like the otel approach you just call diagnose.Error if it fails.

sgmiller (commented):
Not much to comment on here, this all makes sense barring the recording structure we decide on.

command/operator_diagnose.go
HridoyRoy: It does, good call!

serviceregistration/consul/consul_service_registration.go
HridoyRoy: It's definitely quick enough, but if the user doesn't specify tls wouldn't they be surprised by the diagnose TLS error?

command/operator_diagnose.go
```diff
@@ -146,27 +151,25 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
...
+		c.outputErrorsAndWarnings(fmt.Sprintf("Error while reading configuration files: %s", err.Error()), phase, nil)
 		return 1
 	}
+	c.UI.Output(same_line + status_ok + phase)
 
 	// Check Listener Information
+	phase = "Check Listeners"
 	// TODO: Run Diagnose checks on the actual net.Listeners
-
+	var warnings []string
```
sgmiller, line 165: One of the things that'll be nice with the otel version is you can just emit warnings in real time rather than having to build up a list of them, but I'm reviewing this one independently of that.

sgmiller (approved):

PR #11468, opened 2021-04-26
Diagnose V0: Storage End to End Checks

HridoyRoy:
This PR writes, reads, and deletes a secret to physical storage in the root path during the operator diagnose command, with a timeout of 20 seconds. The check aims to diagnose latency issues to storage when vault is down.


command/operator_diagnose.go
```diff
@@ -223,18 +227,41 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
...
 	phase = "Access storage"
 	c.UI.Output(status_unknown + phase)
-	_, err = server.setupStorage(config)
-	storageErrors = append(storageErrors, err.Error())
+	b, err := server.setupStorage(config)
 	if err != nil {
+		storageErrors = append(storageErrors, err.Error())
+	}
 
+	// Attempt to use storage backend
```
sgmiller, line 235: Might break this out into it's own function for tidiness.

sgmiller (commented):
Looks good, we may have to do some error parsing to test the permissions failure case but the best test for storage configuration and operation working is to just try it out.  Cheap to do as well.

vault/diagnose/mock_storage_backend.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/vault/sdk/physical"
+)
+
+const timeoutCallRead string = "lag Read"
```
swayne275, line 11: could you do a `const` block here?

e.g.
```
const (
    thing1 = "hello"
    thing2 = "world"
   ...
)
```

swayne275: i'm only gonna put one comment for this class of thing, but could you also do the same with other consts and vars? i think it makes it a bit cleaner

vault/diagnose/storage_checks.go
```diff
@@ -0,0 +1,82 @@
...
+const wrongRWValsPrefix string = "Storage get and put gave wrong values: "
+
+// StorageEndToEndLatencyCheck calls Write, Read, and Delete on a secret in the root
+// directory of the backend.
+// Note: Just checking read, write, and delete for root. It's a very basic check,
+// but I don't think we can necessarily do any more than that. We could check list,
+// but I don't think List is ever going to break in isolation.
+func StorageEndToEndLatencyCheck(ctx context.Context, b physical.Backend) error {
+
+	c2 := make(chan string, 1)
```
swayne275, line 25: i don't think this channel needs to be buffered with the current code. you could also do a channel of errors (with nil check instead of success stirng) which would simplify the code a bit

same with `c5`

swayne275 (approved):
A few style type things, but i think it looks good!

No reason to change if you disagree, i just thought those suggestions would be helpful.

vault/diagnose/mock_storage_backend.go
HridoyRoy: Yes, will do!

vault/diagnose/storage_checks.go
HridoyRoy: True, will do!

PR #11469, opened 2021-04-26
ECDSA secp256k1 Support (for ETH crypto currency)

adastreamer:
Hello,

I'm CTO of cryptocurrencies processing company and we plan to use Vault for management of crypto keys alongside with signing actions. I have the first version of ECDSA secp256k1 support in transit engine. Looking forward to hearing any feedback / code review. Implementation for Bitcoin also in progress now.

### Usage is the same as for the existing keys types:

**Create Key**
`curl -H "X-Vault-Token: ..." -s -X POST http://127.0.0.1:8200/v1/transit/keys/eth-test -d '{"type":"ecdsa-secp256k1","exportable":true}'`

**Get Key**
`curl -H "X-Vault-Token: ..." -s -X GET http://127.0.0.1:8200/v1/transit/keys/eth-test`
> {"request_id":"5dbde53c-035d-0c36-3ab2-d2da096a6667","lease_id":"","renewable":false,"lease_duration":0,"data":{"allow_plaintext_backup":false,"deletion_allowed":false,"derived":false,"exportable":true,"keys":{"1":{"creation_time":"2021-04-26T18:16:05.150646+03:00","name":"ecdsa-secp256k1","public_key":"**0x0C32AB3e4f256537870A52113fFE1DbEf9f2E6Ad**"}},"latest_version":1,"min_available_version":0,"min_decryption_version":1,"min_encryption_version":0,"name":"eth-test","supports_decryption":false,"supports_derivation":false,"supports_encryption":false,"supports_signing":true,"type":"ecdsa-secp256k1"},"wrap_info":null,"warnings":null,"auth":null}`

**Export Private Key** (for exportable keys)
`curl -H "X-Vault-Token: " -s -X GET http://127.0.0.1:8200/v1/transit/export/signing-key/eth-test`
> {"request_id":"9a2b4c5e-e2c3-8f42-c843-f75929fbd9b8","lease_id":"","renewable":false,"lease_duration":0,"data":{"keys":{"1":"**840fefe4f0fa67fd6791f566f3dc1df5beb9e2a76d209223edb671101fa1ee18**"},"name":"eth-test","type":"ecdsa-secp256k1"},"wrap_info":null,"warnings":null,"auth":null}

**Sign**
`curl -H "X-Vault-Token: ..." -X POST http://127.0.0.1:8200/v1/transit/sign/eth-test -d '{"input":"MTIz","hash_algorithm":"keccak-256"}'`
> {"request_id":"bab73481-0418-37b6-b07f-ab24a2fdaa3d","lease_id":"","renewable":false,"lease_duration":0,"data":{"key_version":1,"signature":"vault:v1:**0x6087159ea3a42d87ee4cb16c3593be4dc6e14f94e875f994fe357f1cab6878a11203431350d812c3a599893cc5a5ad97cc68e163baedb357988a966a469d134f01**"},"wrap_info":null,"warnings":null,"auth":null}

`base64 -D <<< "MTIz" && echo`
> 123

**Verify Signature** (WIP)
Signature accepted by 3rd party services, such as https://etherscan.io/verifySig

Screenshot 2021-04-26 at 18 28 43

Any feedback and review are really appreciated.
Thank you.


A57058M:
nice try!)

VladVladyslav:
awesome!❤️

SorryImNotMike:
Wow! Unbelievable, thank you very much!

AeonHarmony:
Great job!

artamonovdev:
nice job

RemiFokz:
looks good

DashaGera:
Wow! I was waiting for this! ❤️

sdk/helper/keysutil/policy.go
```diff
@@ -1163,6 +1170,37 @@ func (p *Policy) Sign(ver int, context, input []byte, hashAlgorithm HashType, si
...
+
+		key := keyParams.Key
+		hexKey := hexutil.Encode(key)[2:]
+
+		privateKey, err := ethcrypto.HexToECDSA(hexKey)
+		if err != nil {
+			return nil, err
+		}
+
+		hash := ethcrypto.Keccak256Hash(input)
```
intech, line 1188: This not working for bitcoin-based, keccak hash work only for ethereum network. For bitcoin-base double sha256.

intech (changes requested):

sdk/helper/keysutil/policy.go
adastreamer: Correct, I’m working on Bitcoin now, this implementation for Ethereum only

intech: It just so happened that a week ago we did this task for ether and bitcoin as a separate plugin, let's help :)

intech: `entry.FormattedPublicKey` this field is not universal, in the same bitcoin networks there are different formats, it is better not to make a rigid connection to a specific network.

If you remove hash at the initial stage, and receive already hashed data in transit, then there will be a universal solution for any cryptocurrency.

intech: and the `hexutil` from `go-ethereum` use the `0x` prefix, which is not required in bitcoin-based.

intech (approved):

sdk/helper/keysutil/policy.go
intech: and the `hexutil` from `go-ethereum` use the `0x` prefix, which is not required in bitcoin-based.

arnabmitra: > It just so happened that a week ago we did this task for eth and bitcoin as a separate plugin, let's help :)

@intech is the code to the separate plugin somewhere on github? seem's like an interesting one to use if available already

Kirill35:
Interesting! Thank you!

sdk/helper/keysutil/policy.go
adastreamer: Got your point about "prehashed", that makes sense, thanks, doing changes now

hsimon-hashicorp:
Hello! Can you expand upon the use cases, and what problems this PR would solve? Thanks!

hsimon-hashicorp:
Update: Hello! Vault has no plans to implement this in the foreseeable future, but it is an open source product, so please feel free to fork it to develop for your needs. Thanks for understanding.

abhamai:
Based on my deep diving in secp256k1, there is an update.

First of all, the code above doesn't fit cryptocurrencies-related cryptography at all, **please do not use it**.

Here are some key points for whom might be interested in this functionality:

1. Usage of ethereum/go-ethereum as a secp256k1 curve is not efficient enough based on benchmarks in comparison for example to golang-native implementation in decred/dcrd. go-ethereum use links to Bitcoin C++ libs and honestly saying there is a huge mess in this regard, I strictly suggest avoid using go-ethereum as a base for generic secp256k1 cryptography.

2. ethereum/go-ethereum contains of multiple ethereum-specific params, especially in regards to V parameter (see https://eips.ethereum.org/EIPS/eip-155) which is supposed to be on a data-layer (based on logic), but not in signatures, but this is important to avoid replay attack, so cryptography for cryptocurrencies now is more than just using some certain curve.

3. Usage of RFC6979 (https://www.rfc-editor.org/rfc/rfc6979) is important for cryptocurrencies, and there are rules for the curve params for signature generation. Honestly saying I do not remember, what is used in Vault by default, but there is no deterministic nonce by default in crypto/ecdsa or crypto/elliptic packages (might be achieved using custom generator of course).

4. Signatures in blockchains might be recoverable (like Ethereum) and non-recoverable (like Bitcoin) based on a signature format (RSV, or just RS). This is also blockchain-related behaviour which affects curves algorithms and this is something should be parameterized.

5. Addresses in blockchains are not the same as public keys, so this a misconvention is representation layer of functionality this issue is gonna solve. Moreover, some blockchains use compressed or uncompressed keys for some certain addresses types. It's incorrect if we call crypto functions like secp256k1 + keccak256 and use ethereum address as a public key. Address should be moved on another abstraction level.

Right now I'm working on a golang library which covers all possible curves parameters as well as addresses representations, and multiple hashing functions. I will post update and maybe another pull request for Vault once I have the lib done.

Thank you.

adastreamer:
Based on my deep diving in secp256k1, there is an update.

First of all, the code above doesn't fit cryptocurrencies-related cryptography at all, **please do not use it**.

Here are some key points for whom might be interested in this functionality:

1. Usage of ethereum/go-ethereum as a secp256k1 curve is not efficient enough based on benchmarks in comparison for example to golang-native implementation in decred/dcrd. go-ethereum use links to Bitcoin C++ libs and honestly saying there is a huge mess in this regard, I strictly suggest avoid using go-ethereum as a base for generic secp256k1 cryptography.

2. ethereum/go-ethereum contains of multiple ethereum-specific params, especially in regards to V parameter (see https://eips.ethereum.org/EIPS/eip-155) which is supposed to be on a data-layer (based on logic), but not in signatures, but this is important to avoid replay attack, so cryptography for cryptocurrencies now is more than just using some certain curve.

3. Usage of RFC6979 (https://www.rfc-editor.org/rfc/rfc6979) is important for cryptocurrencies, and there are rules for the curve params for signature generation. Honestly saying I do not remember, what is used in Vault by default, but there is no deterministic nonce by default in crypto/ecdsa or crypto/elliptic packages (might be achieved using custom generator of course).

4. Signatures in blockchains might be recoverable (like Ethereum) and non-recoverable (like Bitcoin) based on a signature format (RSV, or just RS). This is also blockchain-related behaviour which affects curves algorithms and this is something should be parameterized.

5. Addresses in blockchains are not the same as public keys, so this a misconvention is representation layer of functionality this issue is gonna solve. Moreover, some blockchains use compressed or uncompressed keys for some certain addresses types. It's incorrect if we call crypto functions like secp256k1 + keccak256 and use ethereum address as a public key. Address should be moved on another abstraction level.

Right now I'm working on a golang library which covers all possible curves parameters as well as addresses representations, and multiple hashing functions. I will post update and maybe another pull request for Vault once I have the lib done.

Thank you.

intech:
@adastreamer Thank you for sharing this valuable knowledge here!

PR #11473, opened 2021-04-26
Update Agent Auth with GCP to use new SignJWT endpoint

catsby:
Updates Vault Agent's GCP auth to use GCP's IAM Service Account Credential endpoint for signing JWTs, as the IAM endpoint versions are deprecated. See https://cloud.google.com/iam/docs/migrating-to-credentials-api for more information on the deprecation and migration. 

See also for more backstory:

- https://github.com/hashicorp/vault-plugin-auth-gcp/issues/100
- https://github.com/hashicorp/vault-plugin-auth-gcp/pull/108
- https://github.com/hashicorp/go-gcp-common/pull/4


command/agent/auth/gcp/gcp.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/hashicorp/vault/command/agent/auth"
 	"github.com/hashicorp/vault/sdk/helper/parseutil"
 	"golang.org/x/oauth2"
-	iam "google.golang.org/api/iam/v1"
+	iam "google.golang.org/api/iamcredentials/v1"
```
calvn, line 20: Nit: Can we reference directly `iamcredentials` instead of overwriting the package name, especially since `google.golang.org/api/iam/v1` exists?

calvn (commented):
Small nit (that would result in a slightly larger diff but ends up being less confusing), otherwise looks good.

calvn (approved):

tvoran (approved):

austingebauer (approved):

command/agent/auth/gcp/gcp.go
catsby: Yeah we should, thanks. I made that change in e0172e5bd

changelog/11473.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:changes
```
tvoran, line 1: It looks like this should be `change`? https://github.com/hashicorp/vault/blob/5aa9047ff282b4fd9fcb865ddbe5c0bac2588896/changelog/changelog.tmpl#L17-L17
```suggestion
```release-note:change
```

catsby: I've seen others with `changes` so I used that, but thanks for pointing out the `tmpl` file. I made the suggested change, thanks!

PR #11477, opened 2021-04-26
Introduce Logical Unrecoverable Error, Use it in Expiration Manager

swayne275:
This one isn't actually that big. It's built on top of https://github.com/hashicorp/vault/pull/11452, and the `...pb.go` files can be ignored. Should be able to just look at 01a406b and beyond

Note: we'll likely need to run a `make proto` and a `go mod vendor` when this auto-merges to ent


vault/expiration.go
```diff
@@ -2291,6 +2345,47 @@ func (m *ExpirationManager) walkLeases(walkFn leaseWalkFunction) error {
...
+	if le == nil {
+		m.logger.Warn("attempted to mark nil lease as zombie")
+		return
+	}
+	if le.isZombie() {
+		m.logger.Info("attempted to re-mark lease as zombie", "original_error", le.RevokeErr, "new_error", err.Error())
+		return
+	}
+
+	errStr := err.Error()
```
HridoyRoy, line 2373: Should we have a nil check here to avoid a potential panic?

vault/expiration.go
```diff
@@ -2291,6 +2345,47 @@ func (m *ExpirationManager) walkLeases(walkFn leaseWalkFunction) error {
...
+	if len(errStr) == 0 {
+		errStr = "no error message given"
```
HridoyRoy, line 2375: Nit: can we put `"no error message given"` in a const and set to the const?

vault/expiration.go
```diff
@@ -2291,6 +2345,47 @@ func (m *ExpirationManager) walkLeases(walkFn leaseWalkFunction) error {
...
+	}
+	if len(errStr) > maxZombieErrorLength {
```
HridoyRoy, line 2377: This seems like an interesting case to log as well. What do you think?

vault/expiration.go
```diff
@@ -205,11 +230,31 @@ func (r *revocationJob) OnFailure(err error) {
...
-
 	if pending.revokesAttempted >= maxRevokeAttempts {
 		r.m.logger.Trace("lease has consumed all retry attempts", "lease_id", r.leaseID)
+		le, loadErr := r.m.loadEntry(r.nsCtx, r.leaseID)
+		if loadErr != nil {
+			r.m.logger.Warn("failed to mark lease as zombie - failed to load", "lease_id", r.leaseID, "err", loadErr)
+			return
+		}
+
+		r.m.markLeaseAsZombie(r.nsCtx, le, errors.New("lease has consumed all retry attempts"))
```
HridoyRoy, line 257: Can this error string be a const?

vault/expiration.go
```diff
@@ -2291,6 +2345,47 @@ func (m *ExpirationManager) walkLeases(walkFn leaseWalkFunction) error {
...
+	}
+	if len(errStr) > maxZombieErrorLength {
+		errStr = errStr[:maxZombieErrorLength]
+	}
+
+	le.RevokeErr = errStr
+	m.persistEntry(ctx, le)
+
+	m.zombies.Store(le.LeaseID, le)
+	m.removeFromPending(ctx, le.LeaseID)
```
HridoyRoy, line 2385: I don't understand quotas enough to comment on correctness here, but this does seem t one adding in an extra call to `quotasHandleLeases`. Could you explain why you're adding that call in? Thanks!

HridoyRoy (changes requested):
Looks good -- some cosmetic nits that are entirely optional (not blocking). 

I'm requesting changes mainly because I'd like a little more information around calling `quotasHandleLeases` when marking leases as zombies before approving. 
Thanks!

vault/expiration.go
```diff
@@ -205,11 +230,31 @@ func (r *revocationJob) OnFailure(err error) {
...
+		r.m.logger.Trace("lease revocation returned unrecoverable error", "lease_id", r.leaseID, "err", err)
+
+		// TODO 1978 below here should be incorporated into markLeaseAsZombie
+		le, loadErr := r.m.loadEntry(r.nsCtx, r.leaseID)
+		if loadErr != nil {
+			r.m.logger.Warn("failed to mark lease as zombie - failed to load", "lease_id", r.leaseID, "err", loadErr)
+			return
+		}
+
+		r.m.markLeaseAsZombie(r.nsCtx, le, err)
```
HridoyRoy, line 243: I see that the nil check is in the `markLeaseAsZombie` function, but I think it would be helpful to have the lease check in the caller as well, so you can log the leaseID for which the corresponding loaded lease was nil.

vault/expiration.go
swayne275: sure! i'm often torn on things like this. like if a nil error is passed in, should we even be marking as a zombie??? anyway, enough pondering. i'll do the same as if `len(err.Error()) == 0`

vault/expiration.go
swayne275: hmm, if you feel strongly that it should be logged i can put a log in. i don't think it's that important though - this is mostly to avoid over-consuming storage/memory, and i think 240 characters is enough to get a sense of what's going on

vault/expiration.go
swayne275: i'll make the error a var

vault/expiration.go
swayne275: yup!

vault/expiration.go
HridoyRoy: But shouldn't zombie leases count to towards the lease count quota? Zombie leases seem exactly like the thing that quotas were guarding against -- namely, vault performance being degraded due to rogue applications "generating an unbounded number of leases and/or loading too much data into Vault leading to exhausting Consul storage backend's available memory" (from https://learn.hashicorp.com/tutorials/vault/resource-quotas).

vault/expiration.go
HridoyRoy: That's fine, definitely don't feel strongly about it. Feel free to mark this one as resolved!

vault/expiration.go
HridoyRoy: Like, if this list of zombies is growing at an alarming rate, I'd want vault to start reporting issues based on lease quotas, rather than silently falling into a degraded state.

swayne275: this feels a bit grey, but you might be right... i'll have to think on that

either way, i'd like to address it in a more related PR (this section of code was from the zombie leases PR) - there may be more work than just skipping the quota update

swayne275: is it alright with you if I create a Jira ticket, move forward with this PR without a fix, and then address the possible fix right after?

swayne275: https://hashicorp.atlassian.net/browse/VAULT-2303

swayne275:
quotas improvements are included in this ticket: https://hashicorp.atlassian.net/browse/VAULT-2303

vault/expiration.go
```diff
@@ -169,6 +175,22 @@ func newRevocationJob(nsCtx context.Context, leaseID string, ns *namespace.Names
...
+// errIsUnrecoverable returns true if the logical error is unlikely to resolve
+// automatically or with additional retries
+func errIsUnrecoverable(err error) bool {
+	switch {
+	case errors.Is(err, logical.ErrUnrecoverable),
+		errors.Is(err, logical.ErrUnsupportedOperation),
+		errors.Is(err, logical.ErrUnsupportedPath),
+		errors.Is(err, logical.ErrInvalidRequest),
+		errors.Is(err, logical.ErrPermissionDenied),
+		errors.Is(err, logical.ErrMultiAuthzPending):
```
briankassouf, line 187: I don't think that `ErrPermissionDenied` or `ErrMultiAuthzPending` should ever happen while sending revoke calls to the mount

vault/expiration.go
```diff
@@ -214,9 +236,26 @@ func (r *revocationJob) OnFailure(err error) {
 		return
 	}
 
+	if errIsUnrecoverable(err) {
```
briankassouf, line 239: Could we move this below and make it:

`if pending.revokesAttempted >= maxRevokeAttempts || errIsUnrecoverable(err) {`

vault/expiration.go
```diff
@@ -225,7 +264,7 @@ func (r *revocationJob) OnFailure(err error) {
 			return
 		}
 
-		r.m.markLeaseAsZombie(r.nsCtx, le, errors.New("lease has consumed all retry attempts"))
+		r.m.markLeaseAsZombie(r.nsCtx, le, errOutOfRetries)
```
briankassouf, line 267: We are storing the entire lease entry in memory here but with pending & nonexpiring we only store a subset of information to save memory. Should we be calling `inMemoryLeaseInfo` in `markLeaseAsZombie`?

vault/expiration.go
swayne275: we can, but then we'll need logic to decide which error to pass to `markLeaseAsZombie()`

swayne275: i've got an idea for a clean combo, i'll comment the commit when done

vault/expiration.go
swayne275: ab9143ab04

vault/expiration.go
swayne275: 6f82bb338f

vault/expiration.go
swayne275: yes, good call ddd285b41a

briankassouf (approved):

PR #11488, opened 2021-04-27
Pass mount accessor rather than inferring from not always present group aliases when refreshing group memberships

### LINKED ISSUE #11487: External Group memberships incorrectly deleted when using multiple auth mounts
### Summary

External Group memberships are incorrectly deleted when using entity aliases from multiple auth mounts mapped to a single entity, and one or more mounts return 0 group aliases for its respective entity alias

### Abstract description
`Entity 1` is an entity is tied to an entity alias from auth mount `Mount A` (`Entity Alias 1A`) and auth mount `Mount B` (`Entity Alias 1B`). Both `Entity Alias 1A` and `Entity Alias 1B` are associated with `Entity 1`.

If the user behind `Entity 1` logs in using `Mount A`, and `Mount A` returns an auth response for `Entity Alias 1A` that includes 2 external group aliases (`Group Alpha-A`, `Group Beta-A`), the vault [login handler will call IdentityStore.refreshExternalGroupMembershipsByEntityID](https://github.com/hashicorp/vault/blob/a24653cc5cca475962b27b71dcba29a4aa94d6ee/vault/request_handling.go#L1206) and `Entity 1` will be added to the groups linked to those external aliases (`Group Alpha`, `Group Beta`).

If the same user behind `Entity 1` then logs in using `Mount B`, and `Mount B` returns an auth response for `Entity Alias 1B` and no `GroupAlias`es, are returned by `Mount B` for that entity alias, `Entity 1` is removed from `Group Alpha` and `Group Beta`, even though the group aliases for those external groups are tied to `Mount A` and not `Mount B`.

Conversely, If `Mount B` returns 1 or more group aliases, Entity 1's membership in `Group Alpha` and `Group Beta` is correctly preserved.

### Real world description

We have an OIDC backend and a backup userpass backend when use if the OIDC provider is down. We link entity aliases from each back to a single entity, and rely on the cached group memberships from the OIDC auth mount to assign policies to identities at login time.

When we enable the userpass backend and a user logs in, the userpass backend [returns no group aliases](https://github.com/hashicorp/vault/blob/master/builtin/credential/userpass/path_login.go#L117-L125), and the entity for that user is removed from all external groups from the OIDC auth mount that they are a member of, losing access we wanted them to maintain.

### Why this happens

The group refresh code does attempt to only filter removals to only be from the same auth mount as the current token was issued from, but it infers the mount by peeking at the first GroupAlias in the group alias slice and taking the mount from it: https://github.com/hashicorp/vault/blob/a24653cc5cca475962b27b71dcba29a4aa94d6ee/vault/identity_store_util.go#L1908-L1911

As a result, if a mount returns no group aliases for an entity alias, `mountAccessor` remains `""` in that code, and the filtering conditional here evaluates to false, and the entity is removed from any external groups it is in from other mounts: https://github.com/hashicorp/vault/blob/a24653cc5cca475962b27b71dcba29a4aa94d6ee/vault/identity_store_util.go#L1967-L1969

## To Reproduce

This is unfortunately less self contained than some of my past issue reproduction steps, but roughly:
1. Configure an LDAP or OIDC auth backend, or any backend that supports external groups/group aliases
1. Configure at least 1 external group+group alias on the LDAP or OIDC backend
1. Login to Vault using the LDAP or OIDC auth backend with an entity that is in the configured external group
1. `vault auth enable userpass && vault write auth/userpass/users/dave.matthews password=testpassword`
1. Use the Vault UI or cli to find the entity alias created for `dave.matthews` on the `userpass` backend above.
1. Link that entity alias to an entity from the LDAP or OIDC auth backend in the external roup
1. `vault login -method=userpass username=dave.matthews password=testpassword`
1. Use the Vault UI or CLI to retrieve group memberships for the aforementioned entity -- they will no longer include the external group previously attached by the login using the LDAP or OIDC backend

## Expected behavior
External groups derived from one auth mount are preserved when an entity logs in using another auth mount that does not support/use groups.

**Environment:**
* Vault Server Version (retrieve with `vault status`): v1.6.2 (but I believe would impact all versions of Vault with external groups)
* Vault CLI Version (retrieve with `vault version`): v1.6.2
* Server Operating System/Architecture: OS X, Linux

**Additional context**

ianferguson:
Fixes #11487

This addresses a bug causing entities to be incorrectly removed from external groups when multiple auth backends are used and one of those auth backends returns zero group aliases for an entity.

Full explanation of bug impact and cause can be found in #11487

### DIFF
--- a/vault/expiration.go
+++ b/vault/expiration.go
@@ -1301,7 +1301,12 @@ func (m *ExpirationManager) RenewToken(ctx context.Context, req *logical.Request
 
 	// Refresh groups
 	if resp.Auth.EntityID != "" && m.core.identityStore != nil {
-		validAliases, err := m.core.identityStore.refreshExternalGroupMembershipsByEntityID(ctx, resp.Auth.EntityID, resp.Auth.GroupAliases)
+		mountAccessor := ""
+		if resp.Auth.Alias != nil {
+			mountAccessor = resp.Auth.Alias.MountAccessor
+		}
+
+		validAliases, err := m.core.identityStore.refreshExternalGroupMembershipsByEntityID(ctx, mountAccessor, resp.Auth.EntityID, resp.Auth.GroupAliases)
 		if err != nil {
 			return nil, err
 		}
--- a/vault/request_handling.go
+++ b/vault/request_handling.go
@@ -1203,7 +1203,7 @@ func (c *Core) handleLoginRequest(ctx context.Context, req *logical.Request) (re
 			}
 
 			auth.EntityID = entity.ID
-			validAliases, err := c.identityStore.refreshExternalGroupMembershipsByEntityID(ctx, auth.EntityID, auth.GroupAliases)
+			validAliases, err := c.identityStore.refreshExternalGroupMembershipsByEntityID(ctx, auth.Alias.MountAccessor, auth.EntityID, auth.GroupAliases)
 			if err != nil {
 				return nil, nil, err
 			}
--- a/vault/identity_store_util.go
+++ b/vault/identity_store_util.go
@@ -1884,7 +1884,7 @@ func (i *IdentityStore) MemDBGroupByAliasID(aliasID string, clone bool) (*identi
 	return i.MemDBGroupByAliasIDInTxn(txn, aliasID, clone)
 }
 
-func (i *IdentityStore) refreshExternalGroupMembershipsByEntityID(ctx context.Context, entityID string, groupAliases []*logical.Alias) ([]*logical.Alias, error) {
+func (i *IdentityStore) refreshExternalGroupMembershipsByEntityID(ctx context.Context, mountAccessor string, entityID string, groupAliases []*logical.Alias) ([]*logical.Alias, error) {
 	defer metrics.MeasureSince([]string{"identity", "refresh_external_groups"}, time.Now())
 
 	if entityID == "" {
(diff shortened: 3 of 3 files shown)


vault/expiration.go
```diff
@@ -1301,7 +1301,12 @@ func (m *ExpirationManager) RenewToken(ctx context.Context, req *logical.Request
 
 	// Refresh groups
 	if resp.Auth.EntityID != "" && m.core.identityStore != nil {
-		validAliases, err := m.core.identityStore.refreshExternalGroupMembershipsByEntityID(ctx, resp.Auth.EntityID, resp.Auth.GroupAliases)
+		mountAccessor := ""
+		if resp.Auth.Alias != nil {
```
ianferguson, line 1305: I was not entirely sure if resp.Auth.Alias could be nil here, but if it ever were to be, this conditional avoids the nil pointer issue and would preserve the current behavior described in #11487 for token renewals

vault/request_handling.go
```diff
@@ -1203,7 +1203,7 @@ func (c *Core) handleLoginRequest(ctx context.Context, req *logical.Request) (re
 			}
 
 			auth.EntityID = entity.ID
-			validAliases, err := c.identityStore.refreshExternalGroupMembershipsByEntityID(ctx, auth.EntityID, auth.GroupAliases)
+			validAliases, err := c.identityStore.refreshExternalGroupMembershipsByEntityID(ctx, auth.Alias.MountAccessor, auth.EntityID, auth.GroupAliases)
```
ianferguson, line 1206: auth.Alias.MountAccessor is set above on line 1180 (`auth.Alias.MountAccessor = req.MountAccessor`) and can be safely referenced here

sgmiller:
Closing this as it was repurposed here: https://github.com/hashicorp/vault/pull/11506

Thanks for your help!

PR #11489, opened 2021-04-28
Minor fix to the raft docs

vishalnayak:
This is the follow up for the work done by @martinhristov90 in https://github.com/hashicorp/vault/pull/10517.


ncabatoff (approved):

mladlow:
I'm going to remove the milestone and add the docs-cherrypick label so this shows up in the documentation now. This doesn't seem like it should need a backport, but feel free to change things back if I'm wrong.

PR #11490, opened 2021-04-29
Make TestActivityLog_MultipleFragmentsAndSegments timeout on its own

ncabatoff:
The main driver for this change was to make the read from a.newFragmentCh timeout quickly rather than waiting for the test timeout (much longer).  While testing the change I observed a panic during shutdown, but it was swallowed and moreover there was no stack trace so it wasn't obvious.  I don't like having that recover so I've removed it and made changes necessary to prevent panics during shutdown.

### DIFF
--- a/vault/activity_log.go
+++ b/vault/activity_log.go
@@ -1523,14 +1523,20 @@ func (a *ActivityLog) precomputedQueryWorker() error {
 
 	// Cancel the context if activity log is shut down.
 	// This will cause the next storage operation to fail.
-	go func() {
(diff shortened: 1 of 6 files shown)


ncabatoff:
Associated test failure: https://app.circleci.com/pipelines/github/hashicorp/vault-enterprise/4185/workflows/5aedf7c1-30a1-4b16-a693-c06b18724b4a/jobs/150845/parallel-runs/7

swayne275 (approved):

PR #7548, opened 2019-10-02
identity propagation in ssh secrets engine #7547

fredrikhgrelland:
To test you may run:
```
pkill -f -9 vault
make dev-ui
export VAULT_DEV_ROOT_TOKEN_ID=test
export VAULT_TOKEN=test
vault server -dev -dev-no-store-token&
sleep 5
export VAULT_ADDR=http://127.0.0.1:8200
vault token lookup
cat <<EOF > user-policy.hcl
# Configure auth methods
path "/*" {
  capabilities = [ "create", "update", "read", "delete", "list", "sudo" ]
}
EOF
vault policy write user user-policy.hcl
vault auth enable userpass
vault write auth/userpass/users/ubuntu password=test policies=user
vault secrets enable -path=ssh ssh
vault write ssh/config/ca generate_signing_key=true
cat <<EOF > signer-clientrole.hcl
{
    "allow_user_certificates": true,
    "allowed_users": "root,{{identity.entity.aliases.$(vault auth list -format=json | jq -r '.["userpass/"].accessor').name}}",
    "default_user": "",
    "allow_user_key_ids": "false",
    "default_extensions": [
        {
          "permit-pty": ""
        }
    ],
    "key_type": "ca",
    "ttl": "60m0s"
}
EOF
vault write ssh/roles/clientrole @signer-clientrole.hcl
vault token lookup
vault write \
    -field=signed_key \
    ssh/sign/clientrole \
    valid_principals="root" \
    public_key=@$HOME/.ssh/id_rsa.pub \
    > $HOME/.ssh/cert-signed.pub
ssh-keygen -L -f $HOME/.ssh/cert-signed.pub
export VAULT_TOKEN=$(vault login -token-only -method=userpass username=ubuntu password=test)
vault token lookup
vault write \
    -field=signed_key \
    ssh/sign/clientrole \
    valid_principals="ubuntu" \
    public_key=@$HOME/.ssh/id_rsa.pub \
    > $HOME/.ssh/cert-signed.pub
ssh-keygen -L -f $HOME/.ssh/cert-signed.pub

```

PR #11495, opened 2021-04-29
Add support for templated values in SSH CA DefaultExtensions.

robison:
This PR adds support for using identity templates in values used in Default Extensions. The primary focus here is to allow the Vault SSH CA backend to function as a point of identity/authentication for GitHub Enterprise (and Github.com) users, by supporting their identity semantics, as [documented here](https://docs.github.com/en/enterprise-server@2.22/organizations/managing-git-access-to-your-organizations-repositories/about-ssh-certificate-authorities).

I've based a large chunk of this on #7548 - happy to make changes as requested/needed to get this into the tree.

### DIFF
--- a/builtin/logical/ssh/path_sign.go
+++ b/builtin/logical/ssh/path_sign.go
@@ -155,7 +155,7 @@ func (b *backend) pathSignCertificate(ctx context.Context, req *logical.Request,
 		return logical.ErrorResponse(err.Error()), nil
 	}
 
-	extensions, err := b.calculateExtensions(data, role)
+	extensions, err := b.calculateExtensions(data, req, role)
 	if err != nil {
 		return logical.ErrorResponse(err.Error()), nil
 	}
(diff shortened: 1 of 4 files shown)


builtin/logical/ssh/path_sign.go
```diff
@@ -356,27 +356,51 @@ func (b *backend) calculateCriticalOptions(data *framework.FieldData, role *sshR
...
 		}
+	} else {
+		extensions = role.DefaultExtensions
+	}
 
-		if len(notAllowed) != 0 {
-			return nil, fmt.Errorf("extensions %v are not on allowed list", notAllowed)
+	if role.DefaultExtensionsTemplate {
+		templatedExtensions := make(map[string]string)
+		for extensionKey, extensionValue := range extensions {
```
briankassouf, line 384: At this point in time the `extensions` map could contain user provided extensions (from the `data.Get("extensions")` call above). The boolean you added only (by name) covers the DefaultExtensions. It may be safer to only allow default extensions to be templated so users cannot query unintended identity data. We could return early above if user provided extensions are passed.

briankassouf (commented):
Thanks for submitting this enhancement!

builtin/logical/ssh/path_sign.go
robison: Ah, excllent point. Let me re-think this portion; originally I had struggled a bit with working out all of the possible unintended consequences (like this).

briankassouf: I think that sounds right. The existing logic prioritizes user input and does not try to merge it with the default extensions. I think we should continue to have the same semantics. So if user supplied extensions exist process those (without templating), else use the defaults with optional templating.

robison: Thanks for the feedback - I've sent a commit to rework the logic, and also add a basic test - can extend this test structure, if it looks acceptable.

robison:
@briankassouf Is there anything else I can do on this PR?

briankassouf (commented):
Thanks for adding the test and Changelog entry! Everything is looking great except for this one comment

builtin/logical/ssh/path_sign.go
```diff
@@ -356,27 +356,50 @@ func (b *backend) calculateCriticalOptions(data *framework.FieldData, role *sshR
...
-
-	extensions := convertMapToStringValue(unparsedExtensions)
-
-	if role.AllowedExtensions != "" {
-		notAllowed := []string{}
-		allowedExtensions := strings.Split(role.AllowedExtensions, ",")
+	extensions := make(map[string]string)
+
+	if len(unparsedExtensions) > 0 {
+		parsedExtensions := convertMapToStringValue(unparsedExtensions)
```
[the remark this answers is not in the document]
robison: Got it - added a commit to fix up this inadvertent change in behavior, and added a test case to cover it.

robison:
As an aside: any thoughts/preferences on cleaning up the tests I've introduced here? I think they've gotten to the point of needing to be broken out into separate cases, but also don't want to start crowding `backend_test.go` with a ton of additional helper functions. Feedback requested!

briankassouf:
Thank you for the latest commit. My only feedback on the test would be the user supplied tests could be broken into a separate test from the templating tests

robison:
I'll do that right now, and send an additional commit shortly.

briankassouf (approved):
Looks great! Thanks for submitting this enhancement

PR #11498, opened 2021-04-29
Update Vault 1.6.x to use GCP Auth v0.8.1

catsby:
Updates Vault 1.6.x to use GCP Auth Plugin v0.8.1, which contains updated source for signing JWTs that were using deprecated endpoints. See hashicorp/vault-plugin-auth-gcp#108 for more information.


Valarissa (approved):
LGTM

PR #11500, opened 2021-04-29
Updated code mirror component for consistency

arnav28:
- Hide gutters, line number and selection while read only
- Show toolbar with copy functionality for all instances

### DIFF
--- a/ui/app/components/json-editor.js
+++ b/ui/app/components/json-editor.js
@@ -14,6 +14,10 @@ const JSON_EDITOR_DEFAULTS = {
 };
 
 export default Component.extend({
+  showToolbar: true,
+  title: null,
+  subTitle: null,
+  helpText: null,
   value: null,
   options: null,
   valueUpdated: null,
--- a/ui/app/components/json-editor.js
+++ b/ui/app/components/json-editor.js
@@ -1,5 +1,5 @@
-import { assign } from '@ember/polyfills';
-import IvyCodemirrorComponent from './ivy-codemirror';
+import Component from '@ember/component';
+
 const JSON_EDITOR_DEFAULTS = {
   // IMPORTANT: `gutters` must come before `lint` since the presence of
   // `gutters` is cached internally when `lint` is toggled
@@ -13,19 +13,37 @@ const JSON_EDITOR_DEFAULTS = {
   showCursorWhenSelecting: true,
 };
 
-export default IvyCodemirrorComponent.extend({
-  'data-test-component': 'json-editor',
-  updateCodeMirrorOptions() {
-    const options = assign({}, JSON_EDITOR_DEFAULTS, this.options);
-    if (options.autoHeight) {
-      options.viewportMargin = Infinity;
-      delete options.autoHeight;
-    }
+export default Component.extend({
+  value: null,
+  options: null,
+  valueUpdated: null,
+  onFocusOut: null,
+  readOnly: false,
 
-    if (options) {
-      Object.keys(options).forEach(function(option) {
-        this.updateCodeMirrorOption(option, options[option]);
-      }, this);
+  init() {
+    this._super(...arguments);
+    this.options = { ...JSON_EDITOR_DEFAULTS, ...this.options };
+    if (this.options.autoHeight) {
+      this.options.viewportMargin = Infinity;
+      delete this.options.autoHeight;
     }
+    if (this.options.readOnly) {
+      this.options.readOnly = 'nocursor';
+      this.options.lineNumbers = false;
+      delete this.options.gutters;
+    }
+  },
+
+  actions: {
+    updateValue(...args) {
+      if (this.valueUpdated) {
+        this.valueUpdated(...args);
+      }
+    },
+    onFocus(...args) {
+      if (this.onFocusOut) {
+        this.onFocusOut(...args);
+      }
+    },
   },
 });
--- a/ui/app/templates/components/secret-edit-display.hbs
+++ b/ui/app/templates/components/secret-edit-display.hbs
@@ -36,17 +36,22 @@
 
   {{#if @showAdvancedMode}}
     <div class="form-section">
-      <label class="title is-5">
-        {{#if isV2}}
-        Version data
-        {{else}}
-        Secret data
-        {{/if}}
-      </label>
+      <Toolbar>
+        <label class="is-label">
+          {{ if isV2 "Version Data" "Secret Data" }}
+        </label>
+        <ToolbarActions>
+          <div class="toolbar-separator"></div>
+          <CopyButton class="button is-transparent" @clipboardText={{@codemirrorString}}
+              @buttonType="button" @success={{action (set-flash-message 'Secret data copied!')}}>
+              <Icon @glyph="copy-action" aria-label="Copy" />
+            </CopyButton>
+        </ToolbarActions>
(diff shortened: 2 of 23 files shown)


arnav28:
Screen Shot 2021-04-29 at 2 44 16 PM

Screen Shot 2021-04-29 at 2 35 26 PM

Screen Shot 2021-04-29 at 3 14 58 PM

ivana-mcconnell:
Would it be possible to see an example where there's help text included?

ui/app/templates/components/transit-key-action/hmac.hbs
```diff
@@ -6,16 +6,26 @@
...
-        <IvyCodemirror @valueUpdated={{action (mut input)}} @options={{hash
-            lineNumbers=true
-            tabSize=2
-            mode='ruby'
-            theme='hashi'
-          }} @data-test-transit-input="input" />
+        <JsonEditor
+            @valueUpdated={{action (mut input)}} 
+            @options={{hash
+              lineNumbers=true
```
hashishaw, line 23: On these instances where we're swapping out IvyCodemirror for the JsonEditor component, we can simplify the options passed since many of these are in the default settings for the JsonEditor component

ui/lib/core/addon/templates/components/form-field.hbs
```diff
@@ -223,6 +241,7 @@
           theme=(or attr.options.theme 'hashi')
         }}
       />
+      <p class="sub-text">{{ attr.options.helpText }}</p>
```
hashishaw, line 244: We should only show this if `attr.options.helpText` exists. This also might error out if there are no options passed, so I would double check that

hashishaw (changes requested):
A few things to revise, and since the previous component was just an extension of a third party library we didn't have any tests for it. We should add tests for this one since we're doing custom logic (esp to make sure the options join with the defaults correctly, show or don't show the toolbar, etc)

ui/app/components/json-editor.js
```diff
@@ -13,19 +13,37 @@ const JSON_EDITOR_DEFAULTS = {
...
+  onFocusOut: null,
+  readOnly: false,
 
-    if (options) {
-      Object.keys(options).forEach(function(option) {
-        this.updateCodeMirrorOption(option, options[option]);
-      }, this);
+  init() {
+    this._super(...arguments);
+    this.options = { ...JSON_EDITOR_DEFAULTS, ...this.options };
```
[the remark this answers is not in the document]
arnav28: **options** is a param on this component. Line 25 will merge the object properties from options passed in the child component with the default ones defined in JSON_EDITOR_DEFAULTS.

ui/app/templates/components/secret-edit-display.hbs
```diff
@@ -36,17 +36,22 @@
...
   {{#if @showAdvancedMode}}
     <div class="form-section">
-      <label class="title is-5">
-        {{#if isV2}}
-        Version data
-        {{else}}
-        Secret data
-        {{/if}}
-      </label>
+      <Toolbar>
```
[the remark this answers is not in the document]
arnav28: I did that but later realized we have a comment with an additional action for the upload file as well. Need to come up with a better design so we can create a separate ToolbarJsonEditor component with support for custom actions.

ui/app/templates/components/transit-key-action/hmac.hbs
arnav28: Good point will fix it.

ui/lib/core/addon/templates/components/form-field.hbs
arnav28: True, will add defensive checks.

ui/app/components/json-editor.js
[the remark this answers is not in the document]
hashishaw: Oh my bad, I was thinking this was a glimmer component 👍

ui/app/templates/components/secret-edit-display.hbs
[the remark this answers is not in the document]
hashishaw: Since the actions for each button will need to be passed in anyway, you could have the JSON component take "children" that get yielded within the toolbar. Or, if we know we want a discreet number of options (probably not) we could to a contextual component. Passing in the yielded content would be simpler, but it does depend on the developer using the correct class names for consistency in styling

arnav28:
https://user-images.githubusercontent.com/2932708/117081650-29039e00-acf5-11eb-9659-1949147b4b9a.mov

ui/app/templates/components/secret-edit-display.hbs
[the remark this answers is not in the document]
arnav28: Done. Moved the Toolbar to json-editor component itself and added a yield block for custom actions.

ivana-mcconnell:
Thanks for the recording! I noticed a couple of things with the cursor hover behaviour. It's backwards in a couple of spots: 

- The hover on the "read" state is the text cursor at 0:05. The cursor's hover on the read state should be "default." This is also the case at 0:24, when viewing the policy. The cursor is "text" in the read state when hovering, and should be "default". 
- When looking at signing the SSH key (0:17), it seems like the cursor is "default" when hovering, but this is an edit state. In this case, the cursor should be "text" 

Additionally, would it be possible to see an example of one of the code blocks with help text?

ui/tests/integration/components/json-editor-test.js
```diff
@@ -0,0 +1,36 @@
...
+
+  test('it renders', async function(assert) {
+    let value = '';
+    await setup(this, 'Test title', value, null);
+    assert.equal(component.title, 'Test title', 'renders the title');
+    assert.equal(component.hasJSONEditor, true, 'renders the ivy code mirror component');
+    assert.equal(component.canEdit, true, 'json editor is in read only mode');
+  });
+
+  test('it renders in read only mode', async function(assert) {
```
hashishaw, line 28: Can we add one more test that ensures the toolbar shows only when showToolbar = true?

hashishaw (commented):
Looking great! I think we should just add one more test, and I'll pull it down and play with it in the meantime

ivana-mcconnell:
After a chat with @arnav28  to clarify cursor behaviour, we've decided to leave them as they are.

The current behaviour is consistent with how the cursor behaves in the browser: when hovering over text, it's the `text` cursor. When it isn't hovering over text, the cursor is `default`. We'll keep this. The original reasoning for changing it was that users would get confused that the read-only code was editable. However, we've made other modifications which should help solve this problem. Those are:

- Removing the 'flashing' cursor in the read-only mode. Previously, the user could click into the read-only version and see an 'editing' cursor (see image below). This gave the impression that editing was possible when it wasn't. 
image
- We've removed the line numbering in the read-only view, which also contributes to the assumption that the code is editable.

The hope is that these visual changes will be enough to prevent customers assuming that the read-only code component is editable. There's no need to change cursor behaviour. If there's still confusion, we can look at making more visual changes to the code display in read-only mode. 

Other than that, this looks good to me now!

ui/tests/integration/components/json-editor-test.js
arnav28: Done.

hashishaw (approved):
Great work! 🚀

PR #11502, opened 2021-04-30
Allow Agent auto auth to read symlinked JWT files

tomhjp:
This came up when helping someone to set up JWT auto auth with Vault Agent in Kubernetes. They had lots of Kubernetes clusters and so didn't want to have to set up a Kubernetes auth mount for each cluster, and instead wanted to use projected Service Account tokens with the JWT auth method. Kubernetes mounts projected tokens using symlinks from an immutable volume mount to help it rotate JWTs when they expire. In this case, the fact the JWT was a symlink stopped JWT auto auth from working, but I see no reason we can't support symlinks in this case.

Currently, the tests are a little fragile in relying on log output. I opted not to change the signature of `ingressToken` to return an error because the actual production code would never pay attention to or change its behaviour based on the return value, but I'm very open to feedback on that.

Lastly, I have opted to allow `ioutil.ReadFile` to handle any errors from cases such as a symlink pointing to a directory for simplicity, as when experimenting I found the error message was clear enough without additional handling, e.g. `read sym-dir: is a directory`. It's possible I've missed a separate motivation for the `IsRegular()` check though.

### DIFF
--- a/command/agent/auth/jwt/jwt.go
+++ b/command/agent/auth/jwt/jwt.go
@@ -164,7 +164,7 @@ func (j *jwtMethod) ingressToken() {
 	j.logger.Debug("new jwt file found")
 
 	// Check that the path refers to a file.
-	// If it's a sym link, it could still be a sym link to a directory,
+	// If it's a sym link, it could still be a symlink to a directory,
 	// but ioutil.ReadFile below will return a descriptive error.
 	if !fi.Mode().IsRegular() && (fi.Mode()&fs.ModeSymlink) == 0 {
 		j.logger.Error("jwt file is not a regular file or symlink")
--- a/command/agent/auth/jwt/jwt.go
+++ b/command/agent/auth/jwt/jwt.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -31,6 +32,8 @@ type jwtMethod struct {
 	latestToken     *atomic.Value
 }
 
+// NewJWTAuthMethod returns an implementation of Agent's auth.AuthMethod
+// interface for JWT auth.
 func NewJWTAuthMethod(conf *auth.AuthConfig) (auth.AuthMethod, error) {
 	if conf == nil {
 		return nil, errors.New("empty config")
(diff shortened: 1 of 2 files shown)


command/agent/auth/jwt/jwt_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+		jwtAuth.ingressToken()
+
+		if tc.errString != "" {
+			if !strings.Contains(logBuffer.String(), tc.errString) {
+				t.Fatal("logs did no contain expected error", tc.errString, logBuffer.String())
+			}
+		} else {
+			if strings.Contains(logBuffer.String(), "[ERROR]") || strings.Contains(logBuffer.String(), "[WARN]") {
+				t.Fatal("logs contained unexepected error", logBuffer.String())
```
tvoran, line 104:
```suggestion
				t.Fatal("logs contained unexpected error", logBuffer.String())
```

command/agent/auth/jwt/jwt.go
```diff
@@ -160,8 +163,11 @@ func (j *jwtMethod) ingressToken() {
...
 
-	if !fi.Mode().IsRegular() {
-		j.logger.Error("jwt file is not a regular file")
+	// Check that the path refers to a file.
+	// If it's a sym link, it could still be a symlink to a directory,
+	// but ioutil.ReadFile below will return a descriptive error.
+	if !fi.Mode().IsRegular() && (fi.Mode()&fs.ModeSymlink) == 0 {
+		j.logger.Error("jwt file is not a regular file or symlink")
 		return
 	}
```
tvoran, line 172: I was thinking a switch statement might make the logic clearer here, though now that I've written it out I'm not so sure. Leaving it here for posterity 🙂 

```suggestion
	switch mode := fi.Mode(); {
	case mode.IsRegular():
		// regular file
	case mode&fs.ModeSymlink != 0:
		// symlink
	default:
		j.logger.Error("jwt file is not a regular file or symlink")
		return
	}
```

tvoran (approved):

command/agent/auth/jwt/jwt.go
tomhjp: I actually prefer that, thanks for the suggestion.

PR #11505, opened 2021-04-30
Diagnoses Consul Direct Access Check

HridoyRoy:
The logic checks if address is not localhost for any storage or SR configuration stanza. Warnings are moved before errors so they can be relayed even when a particular error occurs.


vault/diagnose/storage_checks.go
```diff
@@ -80,3 +81,14 @@ func StorageEndToEndLatencyCheck(ctx context.Context, b physical.Backend) error
 	}
 	return nil
 }
+
+// ConsulDirectAccess verifies that consul is connecting to local agent,
+// versus directly to a remote server. We can only assume that the local address
+// is a server, not a client.
+func ConsulDirectAccess(config map[string]string) string {
+	serviceRegistrationAddr := config["address"]
```
swayne275, line 89: do we want to handle the case where `address` isn't in `config`?

vault/diagnose/storage_checks.go
```diff
@@ -80,3 +81,14 @@ func StorageEndToEndLatencyCheck(ctx context.Context, b physical.Backend) error
...
+	if !strings.Contains(serviceRegistrationAddr, "localhost") && !strings.Contains(serviceRegistrationAddr, "127.0.0.1") {
+		return "consul storage does not connect to local agent, but directly to server"
```
swayne275, line 91: does it make sense to `const` this since we check for it in tests?

swayne275 (approved):
few comments but it looks good

vault/diagnose/storage_checks.go
HridoyRoy: Yes, good point!

vault/diagnose/storage_checks.go
HridoyRoy: Absolutely.

PR #11517, opened 2021-05-03
When running under systemd, send notifications about server startup, shutdown, and config reload

sgmiller:
This is a noop on systems without systemd, and when not spawned by systemd.  Note that "Type=notify" must be specified in the systemd service definition in order for systemd to spawn a notification socket and set the corresponding environment variable.


command/server.go
```diff
@@ -1992,6 +2001,18 @@ CLUSTER_SYNTHESIS_COMPLETE:
...
 
+func (c *ServerCommand) notifySystemd(status string) {
+	sent, err := systemd.SdNotify(false, status)
+	if err != nil {
+		c.logger.Error("error notifying systemd", "error", err)
+	}
+	if sent {
+		c.logger.Debug("sent systemd notification", "notification", status)
+	} else {
+		c.logger.Debug("would have sent systemd notification (systemd not present)", "notification", status)
```
swayne275, line 2012: these are likely very infrequent, but why do we want this log?

swayne275: also i think this can be inaccurate if `err != nil` - I looked at `SdNotify` and it'll return `false` with a non-nil error so this log might be confusing?

swayne275 (changes requested):
requesting changes because i don't think the log in question is accurate if `err != nil`

command/server.go
sgmiller: The logging was partially to assist implementation, but could also help people debugging since it requires correct configuration in systemd to work as well.   You're right that the if sent/else block should only happen if err is nil.

PR #11391, opened 2021-04-18
website: resolve unlinked content

zchsh:
This PR fixes an issue where pages that were previously not included in `*-nav-data.json` files were not rendered.

This PR introduces a temporary fix to ensure these pages are rendered, but are not shown in the sidebar navigation, to match previous behaviour. We plan to revisit this issue more intentionally in the near future, once the [Vault branding PR]() lands.

This PR also introduces a hotfix for brand-related link coloration issues in `docs-sidenav`. On the live site, the "active" sidebar item is not clearly highlighted; with this fix in place, we use our brand blue color to ensure the highlight is visible (and to match a similar hotfix, see `style.css` for details).

## Affected Pages

`api`

- [/api-docs/secret/cassandra](https://vault-git-fork-zchsh-zsfix-missing-pages-hashicorp.vercel.app/api-docs/secret/cassandra) (compare to [live](https://www.vaultproject.io/api-docs/secret/cassandra))
- [/api-docs/system/generate-recovery-token](https://vault-git-fork-zchsh-zsfix-missing-pages-hashicorp.vercel.app/api-docs/system/generate-recovery-token) (compare to [live](https://www.vaultproject.io/api-docs/system/generate-recovery-token))

`docs`

- [/docs/configuration/storage/alicloudoss](https://vault-git-fork-zchsh-zsfix-missing-pages-hashicorp.vercel.app/docs/configuration/storage/alicloudoss) (compare to [live](https://www.vaultproject.io/docs/configuration/storage/alicloudoss))
- [/docs/commands/help](https://vault-git-fork-zchsh-zsfix-missing-pages-hashicorp.vercel.app/docs/commands/help) (compare to [live](https://www.vaultproject.io/docs/commands/help))
  - Note: redirects to `/docs/commands/path-help`, we likely just want to delete the `website/content/docs/commands/help.mdx` file.
- [/docs/commands/lease/lookup](https://vault-git-fork-zchsh-zsfix-missing-pages-hashicorp.vercel.app/docs/commands/lease/lookup) (compare to [live](https://www.vaultproject.io/docs/commands/lease/lookup))
- [/docs/secrets/cassandra](https://vault-git-fork-zchsh-zsfix-missing-pages-hashicorp.vercel.app/docs/secrets/cassandra) (compare to [live](https://www.vaultproject.io/docs/secrets/cassandra))
  - Note: redirects to `/docs/secrets/databases/cassandra`, we likely just want to delete the `website/content/docs/secrets/cassandra.mdx` file.
[...]

PR #11528, opened 2021-05-04
website: support hidden pages in nav-data

zchsh:
👀 [Preview](https://vault-git-fork-zchsh-zssupport-hidden-pages-hashicorp.vercel.app)
🎟️ [Asana task](https://app.asana.com/0/1100423001970639/1200197752405255/f)

This PR is a follow-up to #11391, and implements a more permanent and clean solution for "hidden" pages, via updates to our `docs-page` component.

---

Items in `navData` now support an optional `"hidden": true` property, which will hide the item (and any descendants) from the docs sidebar.

Alongside this change, all `.mdx` files in each `content` folder must be included in the corresponding `nav-data` file. The changes to `docs-page` included an automated check which will throw an error during the website build process if there are any unlinked content files.


website/data/api-docs-nav-data-hidden.json
```diff
@@ -1,20 +0,0 @@
-[
```
zchsh, line 1: The `*-nav-data-hidden.json` files were part of the temporary solution, so removing them. These nav-data items have been added to the proper `nav-data` file, with the `"hidden"` prop in place to reflect current state.

website/redirects.next.js
```diff
@@ -34,6 +34,11 @@ module.exports = [
     destination: '/docs/auth/aws',
     permanent: true,
   },
+  {
+    source: '/docs/auth/jwt_oidc_providers',
+    destination: '/docs/auth/jwt/oidc_providers',
+    permanent: true,
+  },
```
zchsh, line 41: Needed this redirect to account for the moved file. Should note as well here: the reason this happened is that `docs-page` was previously not actually performing the "nesting structure" check - it now is, so these issues will be properly flagged and fail builds going forward.

website/package.json
```diff
@@ -11,7 +11,7 @@
     "@hashicorp/react-case-study-slider": "6.0.0",
     "@hashicorp/react-code-block": "4.0.1",
     "@hashicorp/react-content": "7.0.1",
-    "@hashicorp/react-docs-page": "13.1.4",
+    "@hashicorp/react-docs-page": "13.1.7-alpha.9",
```
zchsh, line 14: Will bump this to a stable release once these PRs are finalized and https://github.com/hashicorp/react-components/pull/220 is merged and published.

website/data/docs-nav-data.json
```diff
@@ -1064,7 +1089,7 @@
           },
           {
             "title": "OIDC Providers",
-            "path": "auth/jwt_oidc_providers"
+            "path": "auth/jwt/oidc_providers"
```
zchsh, line 1092: This was a nesting issue that should have been caught previously, but wasn't due to an issue within `docs-page` which has since been resolved. I've added a corresponding `redirect` to ensure the old URL still works.

website/package.json
zchsh: ✅ Bumped to stable

kylemac (approved):

PR #11530, opened 2021-05-04
KV 2 Toolbar delete redesign

Monkeychip:
This PR redesigns the Delete functionality of the KV2 secret engine.  Before we didn't cover all the situations for the various deletes based on permissions.  Under this new design, we make sure to cover all the various combinations of deletes based on permissions as well as make it a little more intuitive on how to reach the delete actions and what they each do.

Different Delete options:

**1. Delete latest version only.**
```
path "secret/data/*" {
  capabilities = [ "delete"]
}
```
In this scenario we only show the delete dropdown option when the user is on the current version, it disappears if the user is on an older version.
one_soft_delete

**2. Delete any version**
In this scenario, we show the delete dropdown option for all versions.
```
path "secret/delete/*" {
 capabilities = ["update"]
}
```
2deleteanyversion

**3. Destroy version**
Whenever a user has the destroy permissions we show a modal instead of the dropdown.  The modal will then either show all three options or a mix of them based on permissions but it will always show a destroy option. In the example, the user also can delete any version so that option is also shown.
```
path "secret/destroy/*" {
 capabilities = ["update"]
}
```
destroy

**4. Permanently remove all versions**
This policy allows you to permanently remove all the versions, essentially a destroy on all versions.  

```
path "secret/metadata/*" {
 capabilities = ["delete"]
}
```
This is the modal with all permissions.

**5. Undelete only shows if you have the following permissions**
```
path "secret/undelete/*" {
 capabilities = ["update"]
}
```

When the delete/destroy/undelete action occurs the page is refreshed.  Below is a gif.  I couldn't find a better solution for refreshing the model and closing the modal, but I'm very open to suggestions.  However, as it currently works the functionality is clean.  
refresh

Icons were also added to the version dropdown menu as well as the version page (where we removed deleted actions).


ui/app/components/secret-edit.js
```diff
@@ -112,8 +112,8 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
     'model.id',
     'mode'
   ),
-  canDelete: alias('model.canDelete'),
```
Monkeychip, line 115: changed the wording here because "update" matches the actual capabilities word you use to make this work and in this context it was helpful to keep this syntax the same while trying to understand how to properly set the conditionals.

ui/app/adapters/secret-v2-version.js
```diff
@@ -70,14 +70,26 @@ export default ApplicationAdapter.extend({
...
-
-    // deleteType should be 'delete', 'destroy', 'undelete'
-    return this.ajax(this._url(backend, path, deleteType), 'POST', { data: { versions: [version] } }).then(
-      () => {
-        let model = store.peekRecord('secret-v2-version', id);
-        return model && model.rollbackAttributes() && model.reload();
-      }
-    );
+    // deleteType should be 'delete', 'destroy', 'undelete', 'delete-latest-version', 'destroy-version'
+    if ((!version && deleteType === 'delete') || deleteType === 'delete-latest-version') {
```
Monkeychip, line 74: before we were not handling the delete-latest-version which made it impossible for some people with proper permissions to execute this action.  They'd get a permissions denied, because we weren't actually deleting as those permissions allowed.

Monkeychip: We also weren't handling errors, so they would silently fail.

ui/app/components/secret-delete-menu.js
```diff
@@ -0,0 +1,147 @@
+import Ember from 'ember';
```
Monkeychip, line 1: moved all deletion actions into one glimmerized component.

ui/app/components/secret-delete-menu.js
```diff
@@ -0,0 +1,147 @@
...
+  return errorMessage;
+};
+export default class SecretDeleteMenu extends Component {
+  @service store;
+  @service router;
+  @service flashMessages;
+
+  @tracked showDeleteModal = false;
+
+  @maybeQueryRecord(
```
Monkeychip, line 20: TIL that if you have a macro (even a custom one like maybeQueryRecord, you just @'it and then put the name below it.  This took me a while to figure out, so I thought it worth pointing out.

ui/app/components/secret-delete-menu.js
```diff
@@ -0,0 +1,147 @@
...
+            return;
+          }
+          if (!resp) {
+            location.reload();
+          }
+          if (resp.isAdapterError) {
+            const errorMessage = getErrorMessage(resp.errors);
+            this.flashMessages.danger(errorMessage);
+          } else {
+            location.reload();
```
Monkeychip, line 142: here is the reload.  Curious if folks have other suggestions. I tried all sorts of things and nothing cleared the modal.

ui/app/components/secret-version-menu.js
```diff
@@ -1,60 +1,5 @@
-import { maybeQueryRecord } from 'vault/macros/maybe-query-record';
-import Component from '@ember/component';
-import { inject as service } from '@ember/service';
-import { alias, or } from '@ember/object/computed';
+import Component from '@glimmer/component';
```
Monkeychip, line 2: secret-version-menu used to house all the deletion actions, now it only handles version things.

ui/app/styles/components/masked-input.scss
```diff
@@ -66,7 +66,6 @@
     height: auto;
     line-height: 1rem;
     min-width: $spacing-l;
```
Monkeychip, line 68: the icons were showing in front of the modal, we determined it was the z-index causing this.

ui/app/templates/components/secret-edit.hbs
```diff
@@ -33,75 +33,15 @@
     </ToolbarFilters>
   {{/unless}}
   <ToolbarActions>
-    {{#if (and (eq @mode "show") this.isV2 (not @model.failedServerRead))}}
```
Monkeychip, line 36: the ordering of toolbar-actions changed.  So now deletion comes first and version is scooted to the far right.

hashishaw (commented):
Really great work so far, and good questions. I'll pull down the branch to play around with it too. 👏

ui/app/components/secret-edit.js
```diff
@@ -112,8 +112,8 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
...
-  canEdit: alias('updatePath.canUpdate'),
+  canDelete: alias('updatePath.canDelete'),
+  canUpdate: alias('updatePath.canUpdate'),
```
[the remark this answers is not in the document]
Monkeychip: Oh, good catch. It's easy enough for me to change back.

ui/app/templates/components/secret-delete-menu.hbs
```diff
@@ -0,0 +1,132 @@
...
+        >
+        {{if (and (not @modelForData.deleted) (not @modelForData.destroyed)) "Delete" "Destroy"}}
+      </button>
+      <div class="toolbar-separator"/>
+    {{else}}
+      {{#if (or this.canDeleteAnyVersion (and this.isLatestVersion this.canDelete))}}
+        <ConfirmAction
+          @buttonClasses="toolbar-link"
+          @confirmTitle="Delete"
+          @confirmMessage="This version will no longer be able to be read, but the underlying data will not be removed and can still be undeleted."
```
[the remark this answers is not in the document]
Monkeychip: Agreed.  I double check with Ivana.  I checked before (this was the language we used previously) but I'll confirm specifically on the language.

ui/app/components/secret-delete-menu.js
```diff
@@ -0,0 +1,147 @@
...
+      let { id } = this.args.model;
+      this.args.model.destroyRecord().then(() => {
+        this.args.navToNearestAncestor.perform(id);
+      });
+    } else {
+      return this.store
+        .adapterFor('secret-v2-version')
+        .v2DeleteOperation(this.store, this.args.modelForData.id, deleteType)
+        .then(resp => {
+          if (Ember.testing) {
```
johncowen, line 132: Ah so I'm guessing `resp` is where the `undefined` comes in from the comment I made above for the adapter bits?

`resp` is only ever set when its come in via the:

```javascript
catch((e) => { 
  return e
})
```
...in the adapter?

ui/app/templates/components/secret-delete-menu.hbs
```diff
@@ -0,0 +1,132 @@
...
+            <p>All secret versions and metadata are permanently destroyed and cannot be read or recovered later.</p>
+          </div>
+        </div>
+      {{/if}}
+    </section>
+    <footer class="modal-card-foot modal-card-foot-outlined">
+      <button
+        type="button"
+        class="button has-text-danger"
+        {{on "click" (fn this.handleDelete this.deleteType)}}
```
johncowen, line 117: Potential Model Closing Solution Part 1:

Add a third 'success' function here:

```hbs
{{on "click" (fn this.handleDelete this.deleteType (action (mut this.showDeleteModal) false))}}
```

ui/app/components/secret-delete-menu.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+    return true;
+  }
+
+  @action toggleDeleteModal() {
+    this.showDeleteModal = !this.showDeleteModal;
+  }
+
+  @action
+  handleDelete(deleteType) {
```
johncowen, line 117: Potential Model Closing Solution Part 2:

Add a second parameter here called `success`:

```javascript
handleDelete(deleteType, success) {
```

ui/app/components/secret-delete-menu.js
```diff
@@ -0,0 +1,147 @@
...
+    } else {
+      return this.store
+        .adapterFor('secret-v2-version')
+        .v2DeleteOperation(this.store, this.args.modelForData.id, deleteType)
+        .then(resp => {
+          if (Ember.testing) {
+            return;
+          }
+          if (!resp) {
+            location.reload();
```
johncowen, line 136: Potential Model Closing Solution Part 3:

Remove the location.reload() and use `success()` instead:

```javascript
if(!resp) {
  success()
} 
```

johncowen (commented):
Hey @Monkeychip !

I took a quick look at this and noticed something that looked a little strange, but not massively consequential. Also, I checked it out and tried a different approach to closing the modal which 'seemed' to work, but I might have missed something due to lack of context. I tried to detail it in the comments.

Happy to hook up separately to chat through this if it helps though, feel free to give me a ping when you are around if that helps.

ui/app/components/secret-delete-menu.js
Monkeychip: @johncowen this is very helpful!  Thank you.  I'll start working through some of these solutions. Closing the modal was the whole reason I was forcing a reload of the page, but in this case I may not have to.  Thanks again!

johncowen: Awesome 🙌

ui/app/components/secret-delete-menu.js
```diff
@@ -0,0 +1,147 @@
...
+
+  @maybeQueryRecord(
+    'capabilities',
+    context => {
+      if (!context.args.model) {
+        return;
+      }
+      let backend = context.args.model.backend;
+      let id = context.args.model.id;
+      let path = context.args.isV2 ? `${backend}/data/${id}` : `${backend}/${id}`;
```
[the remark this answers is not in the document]
Monkeychip: Great point.  Let me check this out.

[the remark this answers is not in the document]
Monkeychip: Note to self: it errors if unsafe id or backend before this point.  Something to revisit.

ui/app/adapters/secret-v2-version.js
```diff
@@ -70,14 +70,26 @@ export default ApplicationAdapter.extend({
...
+          return model && model.rollbackAttributes() && model.reload();
+        })
+        .catch(e => {
+          return e;
+        });
+    } else {
+      return this.ajax(this._url(backend, path, deleteType), 'POST', { data: { versions: [version] } })
+        .then(() => {
+          let model = store.peekRecord('secret-v2-version', id);
+          return model && model.rollbackAttributes() && model.reload();
```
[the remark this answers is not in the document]
Monkeychip: Investigated and John is potentially right.  However, in order to not cause an unknown bug, I added a comment here instead of a complete removal.  I'll discuss with Chelsea.

ui/app/components/secret-delete-menu.js
Monkeychip: correct. We only get a resp if it's an adapter error.

ui/app/components/secret-delete-menu.js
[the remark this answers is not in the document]
hashishaw: I think it's worth urlEncoding here even though it currently fails upstream, so that in the future if/when the upstream thing is fixed we don't have unsafe strings here

ui/app/components/secret-delete-menu.js
```diff
@@ -0,0 +1,146 @@
...
+  @alias('updatePath.canDelete') canDelete;
+  @alias('updatePath.canUpdate') canUpdate;
+
+  @maybeQueryRecord(
+    'capabilities',
+    context => {
+      if (!context.args || !context.args.modelForData || !context.args.modelForData.id) return;
+      let [backend, id] = JSON.parse(context.args.modelForData.id);
+      return {
+        id: `${backend}/delete/${id}`,
```
hashishaw, line 48: here too probably (re: encoding)

hashishaw (approved):
Great work on this!

PR #11544, opened 2021-05-06
Remove vendor/ (RELENG-355)

samsalisbury:
**UPDATE** ~Do not merge this yet. We are doing similar work for Vault Enterprise, and don't want to merge this until that work is ready.~
**UPDATE 2021-06-10** This is no longer blocked on enterprise changes, just updated it with latest from master.

- Deletes the vendor/ dir
- Updates the spec file packages-oss.yml to use go modules instead
- Updates CI config to use modules not vendor
- Updates `make update-plugins` to update go.mod not vendor/

Purposefully leaves in some references to `vendor/` in the makefile, as they are not harmful, and should people vendor locally for whatever reason they will still be useful to have in (mostly `grep -v vendor/` to filter our vendored files).

Having done this, there is likely further cleanup possible in packages-oss.yml, CI config, and Makefile, but leaving that out until this bit is merged.


tomhjp (approved):
LGTM!

PR #11548, opened 2021-05-06
Update container image tag used in SSH secrets tests

austingebauer:
This PR updates the container image tag used in the SSH backend tests to [8.4_p1-r3-ls48](https://hub.docker.com/layers/linuxserver/openssh-server/8.4_p1-r3-ls48/images/sha256-cde9b6e03c72b2a8b149c374bdf99d8d1556143c5e87736ef8784b534cef42ec?context=explore). The prior tag, [8.3_p1-r0-ls21](https://hub.docker.com/layers/linuxserver/openssh-server/8.3_p1-r0-ls21/images/sha256-777813ff07cd601956d2fc207c1cb420095761bd82aa580598bacaf668757151?context=explore), doesn't have SCP installed, which the test depends on. 

This fixes two tests that are currently failing in CI and locally:
- `TestSSHBackend_DynamicKeyCreate`
- `TestSSHBackend_CredsForZeroAddressRoles_dynamic`

It's not obvious when SCP was removed from the image. It appears the the image hasn't been updated in ~1 year (looking at docker hub).


ncabatoff (approved):

calvn:
I'm a bit confused/curious as to why this started breaking, especially since we pin it to a specific image version.

PR #11565, opened 2021-05-07
Fix activity log fragment race test

swayne275:
functionally, this shouldn't do anything. but i think it'll make the race detector not mad anymore

goal is to fix these races: https://app.circleci.com/pipelines/github/hashicorp/vault/17270/workflows/716efca0-26cc-46b3-9b49-d73fbfa75eba/jobs/199011


ncabatoff (approved):

swayne275:
i race tested it with untilfail and it ran 175 times (no error) before i stopped the test. i'll merge this assuming CI passes

PR #11569, opened 2021-05-10
internalshared: deprecate errwrap.Wrap()

alrs:
This removes `errwrap.Wrapf()` throughout the `internalshared` subpackages and replaces it with `fmt.Errorf()`.


sgmiller (approved):
Good stuff!

PR #11573, opened 2021-05-10
Modularize Run Command

HridoyRoy:
This PR modularizes the run command entirely and adds every step to operator-diagnose, so now operator-diagnose almost fully mimics the actual run command. I also changed diagnose to not return on errors, but just report the error and continue onto the next step, so users can get the output of every step when they run the command.


command/operator_diagnose.go
```diff
@@ -257,22 +253,24 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
+			err = diagnose.StorageEndToEndLatencyCheck(ctx, *backend)
 			if err != nil {
 				return err
 			}
 		}
 
 		return nil
 	}); err != nil {
-		return err
+		diagnose.Error(ctx, err)
```
sgmiller, line 264: I'm not sure you'll need these errors, as they'd be caught by Test as well.  I think you can just drop the if entirely

command/operator_diagnose.go
```diff
@@ -284,7 +282,238 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
+		if sealConfigError != nil {
+			return fmt.Errorf("seal could not be configured: seals may already be initialized")
+		}
+
+		barrierSeal = &bS
+		barrierWrapper = &bW
+		unwrapSeal = &uS
+		if seals != nil {
+			for _, seal := range seals {
+				// Ensure that the seal finalizer is called, even if using verify-only
```
sgmiller, line 317: This is maybe a good place to add another span with the seal type.

command/operator_diagnose.go
```diff
@@ -284,7 +282,238 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
+		if coreConfig.RawConfig == nil {
+			return fmt.Errorf(CoreConfigUninitializedErr)
+		}
+		core, newCoreError = vault.NewCore(&coreConfig)
+		if newCoreError != nil {
+			if vault.IsFatalError(newCoreError) {
+				return fmt.Errorf("Error initializing core: %s", newCoreError)
+			}
+			diagnose.Warn(ctx, wrapAtLength(
+				"WARNING! A non-fatal error occurred during initialization. Please "+
```
sgmiller, line 402: Let's drop the "WARNING" text, that'll be redundant in the output

sgmiller (changes requested):
Just a few comments

command/operator_diagnose.go
```diff
@@ -284,7 +282,238 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
+		err = runListeners(server, &coreConfig, config, configSR)
+		if err != nil {
+			return err
+		}
+		return nil
+	}); err != nil {
+		diagnose.Error(ctx, err)
+	}
+
+	if err := diagnose.Test(ctx, "start-servers", func(ctx context.Context) error {
```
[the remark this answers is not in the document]
HridoyRoy: Good point -- I don't think these will get shut back down as things currently are. I definitely will look into changing that if I can. 

These will conflict with an already running vault server, so I'm not sure what to do with that. We could potentially skip this step for now. It's the same question with starting the listeners though; they may conflict with listeners that are already running.

command/operator_diagnose.go
HridoyRoy: Ooh, that would be a lot nicer. I'll give that a try!

command/operator_diagnose.go
```diff
@@ -173,119 +186,270 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
+					return err
+				}
+				err = diagnose.EndToEndLatencyCheckRead(ctx, veryRandomUuid, *backend)
+				if err != nil && strings.Contains(err.Error(), diagnose.LatencyWarning) {
+					diagnose.Warn(ctx, err.Error())
+				} else if err != nil {
+					return err
+				}
+				err = diagnose.EndToEndLatencyCheckDelete(ctx, veryRandomUuid, *backend)
+				if err != nil && strings.Contains(err.Error(), diagnose.LatencyWarning) {
```
sgmiller, line 247: It's pretty likely if latency is high for one of these, it's high for all of them.  Would it be possible to create a custom error with the latency as a field, then just report once with the highest observed latency?

command/operator_diagnose.go
```diff
@@ -173,119 +186,270 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
+				return nil
 			})
 		}
-		return diagnose.ListenerChecks(sanitizedListeners)
-	}); err != nil {
-		return err
+		return nil
+	})
+
+	sealcontext, sealspan := diagnose.StartSpan(ctx, "create-seal")
```
sgmiller, line 287: This is just the modularization part right?  Testing the seal encrypt/decrypt is later?

command/operator_diagnose.go
```diff
@@ -173,119 +186,270 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
+	}
 
-		if config.Storage != nil && config.Storage.Type == storageTypeConsul {
-			err = physconsul.SetupSecureTLS(api.DefaultConfig(), config.Storage.Config, server.logger, true)
+SEALFAIL:
+	sealspan.End()
+	var coreConfig vault.CoreConfig
+	if err := diagnose.Test(ctx, "setup-core", func(ctx context.Context) error {
+		var secureRandomReader io.Reader
+		diagnose.Test(ctx, "init-randreader", func(ctx context.Context) error {
```
sgmiller, line 326: This one could just be a SpotCheck rather than a test, it's not likely to take long or be complicated.

command/operator_diagnose.go
```diff
@@ -173,119 +186,270 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
+	})
 
-		// Attempt to use storage backend
-		if !c.skipEndEnd {
-			err = diagnose.StorageEndToEndLatencyCheck(ctx, b)
-			if err != nil {
-				return err
-			}
+	// // Determine the redirect address from environment variables
+	diagnose.Test(ctx, "determine-redirect", func(ctx context.Context) error {
```
sgmiller, line 377: Same, this is a tiny thing, maybe just SpotCheck or SpotOk/Error depending on the err return for less code.

command/operator_diagnose.go
```diff
@@ -173,119 +186,270 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
+
+		err = determineRedirectAddr(server, &coreConfig, config)
+		if err != nil {
+			return err
 		}
+		return nil
+	})
 
+	diagnose.Test(ctx, "find-cluster-addr", func(ctx context.Context) error {
+		err = findClusterAddress(server, &coreConfig, config, disableClustering)
```
sgmiller, line 387: Ditto above.

vault/diagnose/storage_checks.go
```diff
@@ -14,66 +14,55 @@ const (
...
-		return fmt.Errorf(timeOutErr + "operation: Put")
+func EndToEndLatencyCheckWrite(ctx context.Context, uuid string, b physical.Backend) error {
+	start := time.Now()
+	err := b.Put(context.Background(), &physical.Entry{Key: secretKey, Value: []byte(secretVal)})
+	duration := time.Since(start)
+	if err != nil {
+		return err
 	}
+	if duration > latencyThreshold {
+		return fmt.Errorf(LatencyWarning + "operation: put")
```
sgmiller, line 32: Per comment above, it'd be nice to have some structure and actually include the latency.  It'll be helpful to folks to report if the latency is 101ms or 3000ms.

vault/init.go
```diff
@@ -461,9 +462,11 @@ func (c *Core) UnsealWithStoredKeys(ctx context.Context) error {
 	// This usually happens when auto-unseal is configured, but the servers have
 	// not been initialized yet.
 	if len(keys) == 0 {
```
sgmiller, line 464: Are we actually calling into UnsealWithStoredKeys from diagnose?  I think we want to avoid actually unsealing.

sgmiller (changes requested):
Nothing huge, just a few pointers.

vault/init.go
HridoyRoy: I don't think we are. Is there somewhere you suspect this might be getting called?

vault/diagnose/storage_checks.go
HridoyRoy: Will do!

command/operator_diagnose.go
HridoyRoy: Sounds good!

command/operator_diagnose.go
HridoyRoy: The seal encrypt/decrypt is in a separate PR (https://github.com/hashicorp/vault/pull/11627). This creates all the seals specified in the config file.

command/operator_diagnose.go
HridoyRoy: Sure!

command/operator_diagnose.go
sgmiller: Nah, sorry, not for the finalize part.  In the next effort where we test each seal, we might want that.

command/operator_diagnose.go
```diff
@@ -173,119 +189,270 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
...
-			// Check ciphersuite and load ca/cert/key files
-			// TODO: TLSConfig returns a reloadFunc and a TLSConfig. We can use this to
-			// perform an active probe.
-			_, _, err := listenerutil.TLSConfig(ln.Config, make(map[string]string), c.UI)
+		// Attempt to use storage backend
+		if !c.skipEndEnd {
+			diagnose.Test(ctx, "test-access-storage", diagnose.WithTimeout(30*time.Second, func(ctx context.Context) error {
+				maxDurationCrudOperation := "write"
+				maxDuration := time.Duration(0)
+				uuidSuffix := strconv.Itoa(rand.Intn(int(math.Pow(2, 10))))
```
sgmiller, line 235: Nit: You can just do 2<<10, or use uuid.GenerateUUID.

HridoyRoy: Done!

sgmiller (approved):

Issue #10835, opened 2021-02-03
Vault agent certificate authentication error


datadude816:
**Describe the bug**
When using the vault agent for auto-auth with cert auth

**To Reproduce**
1. create and sign a certificate
2. enable cert auth in vault
3. upload the certificate to vault
4. configure vault agent to use certificate 
5. run the the agent with the configuration
`sudo -u vault /usr/local/bin/vault agent -config=/etc/vault.d/vault.hcl`

**Expected behavior**
I expect the successful authentication with the token populated in the file specified in the config file.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.6.2 and 1.5.3
* Vault CLI Version (retrieve with `vault version`): 1.6.2 and 1.5.3
* Server Operating System/Architecture: RHEL 8

Vault server configuration file(s):

``` Vault agent config ```
 `vault {
tls_disable = false
client_key = “/etc/pki/tls/certs/machine.key”
client_cert = “/etc/pki/tls/certs/machine.pem”
ca_cert = “/etc/pki/ca-trust/source/anchors/CA.pem”
address = “https://active.vault.service.consul.domain.us:8200”
}
pid_file = “/var/vault/.pidfile”
exit_after_auth = false
auto_auth {
method “cert” {
name = “p5520”
mount_path = “cert/certs/test0”
}
sink “file” {
config = {
path = “/var/vault/token”
}
}
}`

**Additional context**
Using the same certificate as the agent I can authenticate just fine with the CLI and the API using curl

`sudo -u vault curl \

--request POST \
--cert "/etc/pki/tls/certs/machine.pem" \
--key "/etc/pki/tls/certs/machine.key" \
--data '{"name": "p5520"}' \
https://active.vault.service.consul.domain.us:8200/v1/auth/cert/certs/test0/login
[...]

PR #11576, opened 2021-05-11
agent/cert: properly return the cached client on AuthClient

calvn:
This PR addresses a bug on the agent cert auto-auth method where the client returned was not the cached client if `AuthClient` was called on subsequent auth requests.

Possibly fixes https://github.com/hashicorp/vault/issues/10835. This fixes an issue that had the same error message as the one reported there, but the error only appeared if `auth.cert` was expecting a cached client to be returned (e.g. due to a re-auth or retry after backoff)

Test fixtures were copied over from the `api` package in order to keep the tests lean and avoid instantiating a test cluster to dynamically generate them like we do on the end-to-end test (and thus carrying over a bunch of imports to the cert package).


command/agent/auth/cert/cert_test.go
```diff
@@ -0,0 +1,133 @@
...
+			"name": "foo",
+		},
+	}
+
+	method, err := NewCertAuthMethod(config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	client, err := api.NewClient(api.DefaultConfig())
```
tomhjp, line 29: `api.DefaultConfig()` has an odd interface because it may set an error internally but not return it as a return value. Although it's unlikely in a test environment, I think we could miss errors by using the return value without checking.

command/agent/auth/cert/cert_test.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	clientToUse, err := method.(auth.AuthMethodWithClient).AuthClient(client)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !reflect.DeepEqual(client, clientToUse) {
```
tomhjp, line 76: nit: This is a pretty good assertion to make, but might it be better to compare the actual pointer values? After all, if we are looking to test that we receive the same client back, the strongest assertion we can make is that it occupies the same area of memory which we know hasn't been garbage collected in the meantime. The `DeepEqual` assertion is more of a proxy for what we're trying to test, but we could get false negatives due to a deep copy.

command/agent/auth/cert/cert_test.go
calvn: Good point. `NewClient` actually uses `DefaultConfig` internally if I don't pass in any, and also performs the error checking for us. I'll update this.

tomhjp (approved):

PR #11577, opened 2021-05-11
vault: deprecate errwrap.Wrapf()

alrs:
This deprecates `errwrap.Wrapf()` throughout `vault` and its subpackages, replacing it with `fmt.Errorf()`.

### DIFF
--- a/vault/dynamic_system_view.go
+++ b/vault/dynamic_system_view.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/vault/helper/identity"
 	"github.com/hashicorp/vault/helper/namespace"
 	"github.com/hashicorp/vault/helper/random"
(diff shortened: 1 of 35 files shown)


vault/dynamic_system_view.go
```diff
@@ -219,7 +218,7 @@ func (d dynamicSystemView) LookupPlugin(ctx context.Context, name string, plugin
 		return nil, err
 	}
 	if r == nil {
-		return nil, errwrap.Wrapf(fmt.Sprintf("{{err}}: %s", name), ErrPluginNotFound)
+		return nil, fmt.Errorf("%s: %w", name, ErrPluginNotFound)
```
ncabatoff, line 221: While I understand that the format used in this error is unconventional, I would prefer that this PR be a drop-in replacement with only s/wrapf/errorf/.  Makes it easier to have confidence in it despite it being such a sweeping change.

ncabatoff (changes requested):

vault/dynamic_system_view.go
alrs: I understand that. I've made the change.

ncabatoff (approved):
Looks good, thanks again for all your work on this!

PR #11582, opened 2021-05-11
Implement a diagnose check for open files on unix

sgmiller:
https://hashicorp.atlassian.net/browse/VAULT-2421

This has the diagnose-ux branch merged into it, so it may be worth ignoring this until that gets in.


sgmiller:
Closing as too intermixed with the CLI UI

PR #11585, opened 2021-05-11
Check ErrPluginStaticUnsupported for fallback to RotateRootCredentials

pcman312:
When using the V4 DB engine and the user wants to rotate the root credentials, the `SetCredentials` call is attempted. If `SetCredentials` comes back with an "unimplemented" error, the DB engine should fall back to running `RotateRootCredentials`. Previously, it was making the check to determine if it should fall back by checking for an `Unimplemented` gRPC status code. This is never being hit because `ErrPluginStaticUnsupported` is returned instead: https://github.com/hashicorp/vault/blob/master/sdk/database/dbplugin/grpc_transport.go#L341-L345

This PR fixes this by adding `ErrPluginStaticUnsupported` as an additional check to determine if it should fall back to `RotateRootCredentials`.

### DIFF
--- a/builtin/logical/database/version_wrapper.go
+++ b/builtin/logical/database/version_wrapper.go
@@ -152,7 +152,7 @@ func (d databaseVersionWrapper) changePasswordLegacy(ctx context.Context, userna
 	err = d.changeUserPasswordLegacy(ctx, username, passwordChange)
 
 	// If changing the root user's password but SetCredentials is unimplemented, fall back to RotateRootCredentials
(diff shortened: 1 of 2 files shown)


calvn (approved):

pcman312:
@austingebauer and I talked about this offline, but we think this issue came up because the plugin in question was an external plugin. The external plugin goes through the gRPC layer which was converting the gRPC status to the `ErrPluginStaticUnsupported` type.

PR #11588, opened 2021-05-11
Add an exponential backoff to TCP listeners to avoid fast loops in error scenarios

sgmiller:
In cases like too many open files, the TCP listener Accept loop retries as
fast as it can, leading to excessive resource consumption.  This adds a
short (<1s) exponential backoff to those cases.


swayne275 (approved):

briankassouf (approved):
Looks good!

PR #11593, opened 2021-05-12
Add infrastructure and helpers for skipping diagnose checks

sgmiller:


HridoyRoy (approved):

vault/diagnose/helpers.go
```diff
@@ -47,15 +49,39 @@ func New() *Session {
...
+func (s *Session) IsSkipped(skipName string) bool {
+	return s.skip[skipName]
+}
+
 // Context returns a new context with a defined diagnose session
 func Context(ctx context.Context, sess *Session) context.Context {
 	return context.WithValue(ctx, diagnoseSession, sess)
 }
 
+// CurrentSession retrieves the active diagnose section from the context, or nil if none.
```
swayne275, line 74:
```suggestion
// CurrentSession retrieves the active diagnose session from the context, or nil if none.
```

vault/diagnose/helpers.go
```diff
@@ -154,7 +184,7 @@ func Test(ctx context.Context, spanName string, function func(context.Context) e
 // complete within the timeout, e.g.
 //
 // diagnose.Test(ctx, "my-span", diagnose.WithTimeout(5 * time.Second, myTestFunc))
-func WithTimeout(d time.Duration, f func(context.Context) error) func(ctx context.Context) error {
+func WithTimeout(d time.Duration, f func(context.Context) error) func(context.Context) error {
```
swayne275, line 187: might be helpful to the reader to define a type for `func(context.Context) error`?

vault/diagnose/helpers_test.go
```diff
@@ -70,6 +75,7 @@ func makeCoffee(ctx context.Context) error {
 
 	SpotCheck(ctx, "pick-scone", pickScone)
 
+	Test(ctx, "dispose-grounds", Skippable("dispose-grounds", disposeGrounds))
```
swayne275, line 78: just to verify, Test does the appropriate thing if given a function that returns `nil`?

swayne275 (approved):
looks good! approving but have a few comments/questions

vault/diagnose/helpers.go
sgmiller: For code cleanliness, yeah, good idea.

vault/diagnose/helpers_test.go
sgmiller: Yeah, marks the span successful (but will show up as Skipped if the skipped event was added)

PR #11607, opened 2021-05-13
Vault 1979: Query API for Irrevocable Leases

swayne275:


vault/expiration.go
```diff
@@ -2418,6 +2415,140 @@ func (m *ExpirationManager) markLeaseIrrevocable(ctx context.Context, le *leaseE
 	m.nonexpiring.Delete(le.LeaseID)
 }
 
+func (m *ExpirationManager) getNamespaceFromLeaseID(ctx context.Context, leaseID string) (*namespace.Namespace, error) {
+	_, nsID := namespace.SplitIDFromString(leaseID)
+	leaseNS := namespace.RootNamespace
+	var err error
```
swayne275, line 2421: this looks like an unneeded declaration, but it avoids creating a `leaseNS` with scope inside of the if (and not actually setting the one we want)

vault/logical_system_paths.go
```diff
@@ -1210,6 +1210,59 @@ func (b *SystemBackend) leasePaths() []*framework.Path {
...
+				// currently only works for irrevocable leases with param: type=irrevocable
+				logical.ReadOperation: b.handleLeaseCount,
+			},
+
+			HelpSynopsis:    strings.TrimSpace(sysHelp["count-leases"][0]),
+			HelpDescription: strings.TrimSpace(sysHelp["count-leases"][1]),
+		},
+
+		{
+			Pattern: "leases(/)?$",
```
briankassouf, line 1239: We already have a leases/lookup API. Should we repurpose that instead of using the top level /lease path?

website/content/api-docs/system/leases.mdx
```diff
@@ -243,3 +243,68 @@ $ curl \
...
+
+```shell-session
+$ curl \
+    --header "X-Vault-Token: ..." \
+    --request GET \
+    http://127.0.0.1:8200/v1/sys/leases/count \
+    -d type=irrevocable
+```
+
+## Irrevocable Leases List
```
briankassouf, line 277: given the `type` parameter, should the documentation around this API be a little more general?

vault/logical_system_paths.go
```diff
@@ -1210,6 +1210,59 @@ func (b *SystemBackend) leasePaths() []*framework.Path {
...
+			Fields: map[string]*framework.FieldSchema{
+				"type": {
+					Type:        framework.TypeString,
+					Required:    true,
+					Description: "Type of leases to get counts for (currently only supporting irrevocable).",
```
briankassouf, line 1244:
```suggestion
					Description: "Type of leases to retrieve (currently only supporting irrevocable).",
```

vault/logical_system_paths.go
```diff
@@ -1210,6 +1210,59 @@ func (b *SystemBackend) leasePaths() []*framework.Path {
...
+				},
+				"include_child_namespaces": {
+					Type:        framework.TypeBool,
+					Default:     false,
+					Description: "Set true if you want counts for this namespace and its children.",
```
briankassouf, line 1249:
```suggestion
					Description: "Set true if you want leases for this namespace and its children.",
```

vault/expiration.go
```diff
@@ -2418,6 +2415,142 @@ func (m *ExpirationManager) markLeaseIrrevocable(ctx context.Context, le *leaseE
...
+		matchingLeasesPerMount[mountAccessor] = append(matchingLeasesPerMount[mountAccessor], &leaseResponse{
+			LeaseID: leaseID,
+			ErrMsg:  leaseInfo.RevokeErr,
+		})
+
+		return true
+	})
+
+	if retErr != nil {
+		return nil, retErr
```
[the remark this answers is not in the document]
swayne275: i wasn't aware of that, i think that's a much better option!

vault/expiration.go
```diff
@@ -2418,6 +2415,142 @@ func (m *ExpirationManager) markLeaseIrrevocable(ctx context.Context, le *leaseE
...
+		if !force && (numMatchingLeases >= MaxIrrevocableLeasesToReturn) {
+			m.logger.Warn("hit max irrevocable leases without force flag set")
+			retErr = errHitMaxIrrevocableLeases
+			return false
+		}
+
+		mountAccessor := m.getLeaseMountAccessor(ctx, leaseID)
+
+		numMatchingLeases++
+		matchingLeasesPerMount[mountAccessor] = append(matchingLeasesPerMount[mountAccessor], &leaseResponse{
```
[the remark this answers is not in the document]
swayne275: if you think this is a more common use case/better experience then I can definitely go that route

vault/expiration.go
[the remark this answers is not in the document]
swayne275: cb47c4a691

vault/logical_system_paths.go
swayne275: i checked it out and i'm not sure if this has the same ethos as the lookup api. i could be convinced, though

website/content/api-docs/system/leases.mdx
swayne275: e5dafb4cd8

vault/expiration.go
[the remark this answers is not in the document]
briankassouf: I think so, it's also what was written up in the RFC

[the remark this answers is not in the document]
swayne275: yeah, i was deciding if that should be changed or not

alright, can do!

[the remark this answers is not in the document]
swayne275: 6f2abe7ae8

vault/logical_system_paths.go
```diff
@@ -1210,6 +1210,59 @@ func (b *SystemBackend) leasePaths() []*framework.Path {
...
+				},
+				"force": {
```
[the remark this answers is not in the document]
swayne275: da988f93d8

good call. i changed it to `include_large_results`, although am open to a better name

I'd prefer to keep it as a bool unless we think there will be enough customer benefit to changing to an int - keep things simple if we can

vault/expiration.go
```diff
@@ -2418,6 +2415,142 @@ func (m *ExpirationManager) markLeaseIrrevocable(ctx context.Context, le *leaseE
...
+	if err != nil {
+		m.logger.Error("could not get namespace from context", "error", err)
+		return nil, err
+	}
+
+	// map of mount point : lease info
+	matchingLeasesPerMount := make(map[string][]*leaseResponse)
+	numMatchingLeases := 0
+	var retErr error
+	m.irrevocable.Range(func(k, v interface{}) bool {
```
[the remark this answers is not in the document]
swayne275: yeah, i think there's benefit to that consistency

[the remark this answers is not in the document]
swayne275: i've also thought that getting away from `sync.Map` could be nice so that when there are >10,000 results we can tell the client how many to expect (without doing the work of going through all of them)... a story for another day perhaps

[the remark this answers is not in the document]
swayne275: d06e92305c

vault/expiration.go
```diff
@@ -2418,6 +2414,146 @@ func (m *ExpirationManager) markLeaseIrrevocable(ctx context.Context, le *leaseE
...
+			MountID: mountAccessor,
+			ErrMsg:  leaseInfo.RevokeErr,
+		})
+
+		return true
+	})
+
+	// sort the results for consistent API response
+	sort.Slice(matchingLeases, func(i, j int) bool {
+		return matchingLeases[i].LeaseID < matchingLeases[j].LeaseID
```
sgmiller, line 2547: I think this is fine, but we could also sort by expiration time then lease id, so the operator can get a sense of how old the oldest irrevocable leases are.

vault/logical_system_paths.go
sgmiller: The leases lookup does segment by path, which _could_ be useful?  That has the advantage of allowing operators to ACL against specific paths for irrevocable lookups.

vault/expiration.go
swayne275: yeah that sounds like a good idea

vault/logical_system_paths.go
swayne275: @sgmiller and i discussed in slack and decided not to use lease lookup

vault/expiration.go
swayne275: 23a10820e4

sgmiller: Other way around on that sort? I think you want to do it in just one pass with the sort predicate testing expiration time and returning if non-equal, then testing leaseid if the expiration times are equal.

sgmiller (changes requested):

vault/expiration.go
swayne275: can you clarify how you'd like this sorted? the current code will display the oldest-expiring leases at the top of the list, and if multiple leases have the same expiration time it will then display them in increasing order by the lease string comparison (a comes before b)

i can certainly do it in one slice sort path, but let's verify how we want the results displayed first

swayne275: we chatted in slack. sorting is correct, but made more efficient here: 585caeafff

sgmiller (approved):

PR #11627, opened 2021-05-17
Diagnose resource creation checks

HridoyRoy:
1. Static checks for configuration during resource creation are added to diagnose.
2. Sealing/unsealing a dummy value is checked.


command/operator_diagnose.go
```diff
@@ -450,5 +481,43 @@ SEALFAIL:
...
 
 	// TODO: Diagnose logging configuration
+
+	// The unseal diagnose check will simply attempt to use the barrier to encrypt and
+	// decrypt a mock value. It will not call runUnseal.
+	if err := diagnose.Test(ctx, "unseal", func(ctx context.Context) error {
+		if barrierWrapper == nil {
+			return fmt.Errorf("Diagnose could not create a barrier seal object")
+		}
+		barrierEncValue := "diagnose-0481"
```
sgmiller, line 491: Lets make this random too just to be sure.

sgmiller (changes requested):
One small thing

sgmiller:
But can those be done without unsealing?

command/operator_diagnose.go
```diff
@@ -427,6 +428,26 @@ SEALFAIL:
...
 
+	// Run all the checks that are utilized when initializing a core object
+	// without actually calling core.Init. These are in the init-core section
+	// as they are runtime checks.
+	diagnose.Test(ctx, "init-core", func(ctx context.Context) error {
+		var newCoreError error
+		if coreConfig.RawConfig == nil {
+			return fmt.Errorf(CoreConfigUninitializedErr)
+		}
+		_, newCoreError = vault.NewCoreUninit(&coreConfig)
```
HridoyRoy, line 439: CreateCore and StartCore as opposed to NewCoreUninit and NewCoreInit

vault/core.go
```diff
@@ -896,8 +897,18 @@ func NewCore(conf *CoreConfig) (*Core, error) {
...
-	if err := coreInit(c, conf); err != nil {
+// NewCore is used to construct a new core
+func NewCore(conf *CoreConfig) (*Core, error) {
+	var err error
+	c, err := NewCoreUninit(conf)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = coreInit(c, conf); err != nil {
```
HridoyRoy, line 911: If we don't init the core object, we can't call the helper that checks if the autoloader license exists.

However, `VAULT_LICENSE` should contain the license if not empty, and `VAULT_LICENSE_PATH` should contain a filesystem license path to a license file (regardless of the state vault is in). 

There's also a field on the hcl config called `license_path` which should also contain the license path (same as env var able).

One of those three fields have to have something in it, otherwise autoloading cannot work.

PR #11637, opened 2021-05-17
Backport 1.7.x: Add ability to customize some timeouts in MongoDB database plugin

pcman312:
Backport to 1.7.x: https://github.com/hashicorp/vault/pull/11600
---
This PR adds the ability to customize `SocketTimeout`, `ConnectTimeout`, and `ServerSelectionTimeout` within MongoDB connections. Vault passes these values to the MongoDB library when it makes a client connection.

This also makes some improvements around throughput. Previously, when an operation came to the plugin (creating/destroying a user, etc.), it would lock a mutex for the entirety of the operation. This resulted in the plugin forcing all requests to be single-threaded. This moves the mutex lock a bit lower in the plugin to try to limit its scope and allow multiple threads to be manipulating users at the same time. I did run a bunch of tests to try to produce races and didn't find any within the plugin code. However, I did find a race within the DB engine which I think is fixed in `builtin/logical/database/path_rotate_credentials.go`. The lock was being unlocked before the connection was being being closed & removed from the DBI map (`b.connections`). This probably wasn't a serious issue in practice as the map isn't changed in this manner very frequently.

### DIFF
--- a/builtin/logical/database/path_rotate_credentials.go
+++ b/builtin/logical/database/path_rotate_credentials.go
@@ -78,6 +78,14 @@ func (b *databaseBackend) pathRotateRootCredentialsUpdate() framework.OperationF
 			return nil, err
 		}
 
+		// Take out the backend lock since we are swapping out the connection
+		b.Lock()
+		defer b.Unlock()
+
+		// Take the write lock on the instance
+		dbi.Lock()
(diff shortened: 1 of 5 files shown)


calvn (approved):

NicoleS021:
oracle.go

PR #11638, opened 2021-05-17
AWS Auth: Update error message to include underlying error

pcman312:
Updates an error message to include the underlying error to help users debug what the actual problem is instead of guessing based on an incomplete message.

### DIFF
--- a/builtin/credential/aws/path_login.go
+++ b/builtin/credential/aws/path_login.go
@@ -1364,7 +1364,7 @@ func (b *backend) pathLoginUpdateIam(ctx context.Context, req *logical.Request, 
 	if roleEntry.InferredEntityType == ec2EntityType {
 		instance, err := b.validateInstance(ctx, req.Storage, entity.SessionInfo, roleEntry.InferredAWSRegion, callerID.Account)
 		if err != nil {
-			return logical.ErrorResponse(fmt.Sprintf("failed to verify %s as a valid EC2 instance in region %s", entity.SessionInfo, roleEntry.InferredAWSRegion)), nil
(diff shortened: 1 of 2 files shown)


builtin/credential/aws/path_login.go
```diff
@@ -1364,7 +1364,7 @@ func (b *backend) pathLoginUpdateIam(ctx context.Context, req *logical.Request,
 	if roleEntry.InferredEntityType == ec2EntityType {
 		instance, err := b.validateInstance(ctx, req.Storage, entity.SessionInfo, roleEntry.InferredAWSRegion, callerID.Account)
 		if err != nil {
-			return logical.ErrorResponse(fmt.Sprintf("failed to verify %s as a valid EC2 instance in region %s", entity.SessionInfo, roleEntry.InferredAWSRegion)), nil
+			return logical.ErrorResponse("failed to verify %s as a valid EC2 instance in region %s: %s", entity.SessionInfo, roleEntry.InferredAWSRegion, err), nil
```
calvn, line 1367: nit:
```suggestion
			return logical.ErrorResponse("failed to verify %s as a valid EC2 instance in region %s: %w", entity.SessionInfo, roleEntry.InferredAWSRegion, err), nil
```

calvn (approved):

builtin/credential/aws/path_login.go
pcman312: Unfortunately this won't work here. The format string is interpreted with `fmt.Sprintf`, not `fmt.Errorf`: https://github.com/hashicorp/vault/blob/master/sdk/logical/response.go#L116

calvn: Nvm, this this on `logical.ErrorResponse`...

calvn:
Almost forgot -- can we add a CL for this?

pcman312:
Whoops, forgot about the CL. Added.

Issue #11557, opened 2021-05-07
Group Membership does not appear in UI


mattcopas:
**Describe the bug**
I've added an entity to a group via the REST API, and can see that the entity has been added to the group if I query the group or entity via the rest api (that is, `member_entity_ids` and `group_ids` are populated as expected).

However, when viewing the group or entity in the UI, the group appears to have no members, and the entity appears be in no groups.

This seems to be a bug in the UI itself, as looking at the Network tab in Chrome dev tools, I can see from the response that the group has members, and the entity belongs to the group.

**To Reproduce**
1. Create a Group `example-group` and an Entity `example-entity` (I did this via Terraform but the REST API will do the job)
2. Add example-entity to example-group via the REST API
3. Query example-group/example-entity by name via the REST API to confirm this has worked
4. View example-group or example-entity in the UI - it will appear as if example-group has no members, and example-entity belongs to no groups.

**Expected behavior**
In the UI, example-group should have example-entity as a member, and example-entity should be a member of example-group.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.7.1
* Vault CLI Version (retrieve with `vault version`): Vault v1.7.1 (917142287996a005cb1ed9d96d00d06a0590e44e)
* Server Operating System/Architecture: Docker Image (`vault:latest`)

mladlow:
Hi @mattcopas - I see a PR associated with this issue has been merged and should be included in the Vault 1.7.2 release, so I'm going to close this issue. Thanks for reporting it!

PR #11641, opened 2021-05-17
UI/fix identity model

hashishaw:
This PR addresses issue #11557 where group membership and metadata were not showing in the UI. This was due to the model not being invoked correctly in the component template.


ui/app/templates/components/identity/item-metadata.hbs
```diff
@@ -1,4 +1,5 @@
-{{#each-in model.metadata as |key value|}}
+{{log "model" model this.model @model}}
```
arnav28, line 1: Do we need to log this or just forgot to remove?

hashishaw: Just forgot! Thanks

arnav28 (approved):

PR #11646, opened 2021-05-18
Vault 1.5.9 Pre-staging

HridoyRoy:


sdk/version/version_base.go
```diff
@@ -8,7 +8,7 @@ var (
 	// Whether cgo is enabled or not; set at build time
 	CgoEnabled bool
 
-	Version           = "1.5.8"
-	VersionPrerelease = ""
+	Version           = "1.5.9"
+	VersionPrerelease = "rc"
```
vishalnayak, line 12: This is not an RC. Why is this set?

HridoyRoy: Oh, do minor releases not have release candidates? Should I just set it back to `""` ?

vishalnayak: I don't think so. In my opinion, yes, it should be set to "".

HridoyRoy: Sounds good!

vishalnayak (approved):

briankassouf (approved):

Issue #11570, opened 2021-05-10
vault does not allow changing approle token_num_use=1 to batch token in a single request


juliantaylor:
**Describe the bug**
When you have an approle with default token type and token_num_uses>0 and you want to change it to a batch token (which implies token_num_uses=0) the update fails when done in a single requset

**To Reproduce**
```
$ vault write auth/approle/role/rolename bind_secret_id=false secret_id_bound_cidrs=10.90.28.0/22 secret_id_num_uses=0 secret_id_ttl=10m token_policies=policy token_ttl=5m token_max_ttl=10m token_num_uses=2 token_type=default
Success! Data written to: auth/approle/role/rolename
$ vault write auth/approle/role/rolename bind_secret_id=false secret_id_bound_cidrs=10.90.28.0/22 secret_id_num_uses=0 secret_id_ttl=10m token_policies=policy token_ttl=5m token_max_ttl=10m token_num_uses=0 token_type=batch
Code: 400. Errors:

* 'token_type' cannot be 'batch' or 'default_batch' when set to generate tokens with limited use count
```

The workaround is to update the approle first to token_num_uses=0 and default token_type and then change the token type (or delete and recreate which cycles the roleid)

**Expected behavior**
The change can be done in one call instead of two.

**Environment:**
* Vault Server Version: 1.6.3
* Vault CLI Version: 1.6.3
* Server Operating System/Architecture: debian buster

PR #11647, opened 2021-05-18
Tokenutil: Perform num uses check earlier

vishalnayak:
Fixes: #11570

### DIFF
--- a/builtin/credential/approle/backend_test.go
+++ b/builtin/credential/approle/backend_test.go
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/hashicorp/vault/sdk/logical"
 )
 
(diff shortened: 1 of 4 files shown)


builtin/credential/approle/backend_test.go
```diff
@@ -26,6 +27,39 @@ func createBackendWithStorage(t *testing.T) (*backend, logical.Storage) {
...
+		"token_ttl":             "5m",
+		"token_max_ttl":         "10m",
+		"token_num_uses":        2,
+		"token_type":            "default",
+	}
+	requestFunc(logical.CreateOperation, data)
+
+	data["token_num_uses"] = 0
+	data["token_type"] = "batch"
+	requestFunc(logical.UpdateOperation, data)
```
ncabatoff, line 60: Would it make sense to try using the token after making the change?

vishalnayak: I am not sure I understand. This is when updating a role that is used to issue a token. Do you mean that that test should issue a token and try to use it?

vishalnayak: The tests stops here because without the fix, the update step used to fail and now it doesn't.

ncabatoff: Maybe?  I just figure if this wasn't possible before, the resulting role might not yield usable tokens due to other latent bugs, so it would be nice to validate that they work.

vishalnayak: Sure. I'll add those steps.

vishalnayak: This is a backend level unit test. So I couldn't ensure that the token is usable. However the test now does ensure that a valid Auth block is returned from the engine. The token creation and usability should not be affected in my opinion. Pushed up the changes.

ncabatoff (approved):

Issue #11572, opened 2021-05-10
Docs are rendered with wrong YAML indentation on website


mbrancato:
**Describe the bug**
The website has not been updated to the latest version of the docs, but starting with this line in the Vault Kubernetes documentation, the fields are outdented and should be nested under `injector` in the current website docs version, and `agentDefaults` in the next release:
https://github.com/hashicorp/vault/blame/e27b43c4829223555484b2fb7f0049b3ec2881fc/website/content/docs/platform/k8s/helm/configuration.mdx#L128

Here is the beginning of the `injector` block:
![image](https://user-images.githubusercontent.com/8493275/117680286-03f0ae80-b17f-11eb-9aa1-748c8c2e0d69.png)

And here is where the `injector.extraLabels` begins:
![image](https://user-images.githubusercontent.com/8493275/117680388-1e2a8c80-b17f-11eb-862a-e42a181b85ff.png)

Note that the `injector.objectSelector` is correctly indented.

I've tested in Chrome and Safari.

**To Reproduce**
1. View docs page: https://www.vaultproject.io/docs/platform/k8s/helm/configuration#objectselector

**Expected behavior**
The indentation should match the documentation.

SohamJ:
This also seems to apply to all configs that follow `extraLabels` like  `certs`, `failurePolicy`, `annotations`, etc based on the [injector-deployment](https://github.com/hashicorp/vault-helm/blob/dcb4b102835d1c9911cfc9a47fb9bfc2084ffdbf/templates/injector-deployment.yaml) template

jasonodonnell:
Thanks, we'll fix this today!

jasonodonnell:
We found the root cause of the issues, this should be fixed now.

kitography:
This looks fixed to me (looking at the documentation now), but feel free to reopen if I've missed something.

PR #11657, opened 2021-05-19
docs: fix rendering issue with helm configs

jasonodonnell:
Rendering is a little off when YAML source code blocks are indented.

Fixes https://github.com/hashicorp/vault/issues/11572.


tvoran:
Personally I think it makes more sense visually to have the yaml lined up with the text instead of the bullet points, but I guess we can choose one way or the other.

But in this case it looks like the indentation problem is being caused by an extra line in the stable-website branch (which is now behind master for this file):
https://github.com/hashicorp/vault/blob/beff5544dd926793794fade0b62b0a9ab460830a/website/content/docs/platform/k8s/helm/configuration.mdx#L125-L133

mbrancato:
Because it is kind of difficult to read, it might be helpful to move from bullets to dotted syntax. That would make things much clearer IMO.

PR #11672, opened 2021-05-19
UI/tools partial

Monkeychip:
In an effort to get ahead of future Ember upgrade depreciations we are trying to replace our use of partials with components.  This tackles the Tools area.

In the end there is slightly more code, but it is much easier to read.


ui/app/components/tool-actions-form.js
```diff
@@ -135,14 +135,11 @@ export default Component.extend(DEFAULTS, {
       this.reset();
     },
 
-    updateTtl(evt) {
-      const ttl = evt.enabled ? `${evt.seconds}s` : '30m';
```
Monkeychip, line 139: the ttl is determined in the tool-hash component.

ui/app/components/tool-actions-form.js
```diff
@@ -135,14 +135,11 @@ export default Component.extend(DEFAULTS, {
...
+    updateTtl(ttl) {
       set(this, 'wrapTTL', ttl);
     },
 
-    codemirrorUpdated(val, codemirror) {
-      codemirror.performLint();
```
Monkeychip, line 144: codemirror things need to happen inside the component where codemirroris being used, thus the property `codemirror` is used in the tool-hash component.

ui/app/components/tool-hash.js
```diff
@@ -0,0 +1,9 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class HashTool extends Component {
+  @action
+  onClear() {
+    this.args.onClear();
```
Monkeychip, line 7: for all new components the `onClear()` action is passed through to the child as a property and then called again via an action.

ui/app/components/tool-wrap.js
```diff
@@ -0,0 +1,27 @@
...
+    const ttl = evt.enabled ? `${evt.seconds}s` : '30m';
+    this.args.updateTtl(ttl);
+  }
+  @action
+  codemirrorUpdated(val, codemirror) {
+    codemirror.performLint();
+    const hasErrors = codemirror.state.lint.marked.length > 0;
+    this.data = val;
+    this.buttonDisabled = hasErrors;
+    this.args.codemirrorUpdated(val, hasErrors);
```
Monkeychip, line 25: now call parent's codemirrorUpdated function and instead of passing in `val, codemirror` we pass in `val, hasErrors`.  Again, the codemirror property is only available inside the instance of the component it is used, it cannot be passed through as a property.

ui/app/templates/components/tool-actions-form.hbs
```diff
@@ -1,3 +1,59 @@
...
+      @creation_time={{creation_time}}
+      @creation_ttl={{creation_ttl}}
+      @creation_path={{creation_path}}
+      @expirationDate={{expirationDate}}
+      @selectedAction={{selectedAction}}
+      @token={{token}}
+      @onClear={{action "onClear"}}
+      @errors={{errors}}
+    />
+  {{else if (eq selectedAction 'wrap')}}
```
Monkeychip, line 48: notice these are all else ifs, there is no fallback component.  Technically there doesn't need to be, but curious if folks take issue with this.

ui/app/components/tool-hash.js
```diff
@@ -0,0 +1,9 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class HashTool extends Component {
```
hashishaw, line 4: Might be nice to add JSDocs here and include onClear as a required attr, as well as anything else used on the template

ui/app/templates/components/tool-actions-form.hbs
hashishaw: I think this is the right call, unless we want to add an empty state (which would most likely be viewed by a developer updating this space)

ui/app/components/tool-lookup.js
```diff
@@ -0,0 +1,9 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class HashTool extends Component {
```
hashishaw, line 4: Sort of a nit, but can we name the class to match the file and how they're invoked (eg this one would be `ToolLookup`)

Monkeychip: I totally missed that, not a nit at all.  Nice catch, thanks!

ui/app/components/tool-wrap.js
```diff
@@ -0,0 +1,27 @@
...
+  @action
+  updateTtl(evt) {
+    if (!evt) return;
+    const ttl = evt.enabled ? `${evt.seconds}s` : '30m';
+    this.args.updateTtl(ttl);
+  }
+  @action
+  codemirrorUpdated(val, codemirror) {
+    codemirror.performLint();
+    const hasErrors = codemirror.state.lint.marked.length > 0;
```
hashishaw, line 22: I wonder if we should use optional chaining here in case `marked` or anything upstream comes back `undefined`?

ui/app/templates/components/tool-actions-form.hbs
```diff
@@ -1,3 +1,59 @@
 <form onsubmit={{action "doSubmit"}}>
-  {{partial (concat "partials/tools/" selectedAction)}}
+  {{#if (eq selectedAction 'hash')}}
+    <ToolHash
```
hashishaw, line 3: 😍

ui/app/components/tool-wrap.js
Monkeychip: Good point.

ui/app/templates/components/tool-actions-form.hbs
Monkeychip: Good idea with the empty state.  It feels strange to have no else so I'll go ahead and add it.

ui/app/components/tool-hash.js
Monkeychip: Great idea.

ui/app/templates/components/tool-rewrap.hbs
```diff
@@ -6,35 +6,35 @@
...
   <div class="field is-grouped box is-fullwidth is-bottomless">
     <div class="control">
-      <CopyButton @clipboardText={{rewrap_token}} @class="button is-primary" @buttonType="button" @success={{action (set-flash-message "Token copied!")}}>
+      <CopyButton @clipboardText={{@rewrap_token}} @class="button is-primary" @buttonType="button" @success={{action (set-flash-message "Token copied!")}}>
         Copy
       </CopyButton>
     </div>
     <div class="control">
-      <button {{action 'onClear'}} type="button" class="button">
+      <button {{on "click" (action this.onClear)}} type="button" class="button">
```
hashishaw, line 25: we *should* be able to just do `{{on "click" this.onClear}}` but if that doesn't work for some reason I'd be happy to pair about it

ui/app/templates/components/tool-wrap.hbs
```diff
@@ -6,51 +6,51 @@
...
   <div class="field is-grouped box is-fullwidth is-bottomless">
      <div class="control">
-      <CopyButton @clipboardText={{token}} @class="button is-primary" @buttonType="button" @success={{action (set-flash-message "Token copied!")}}>
+      <CopyButton @clipboardText={{@token}} @class="button is-primary" @buttonType="button" @success={{action (set-flash-message "Token copied!")}}>
         Copy
       </CopyButton>
     </div>
     <div class="control">
-      <button {{action 'onClear'}} type="button" class="button">
+      <button {{on "click" (action this.onClear)}} type="button" class="button">
```
hashishaw, line 25: same here

ui/app/templates/components/tool-wrap.hbs
```diff
@@ -6,51 +6,51 @@
...
-    <MessageError @errors={{errors}} />
+    <NamespaceReminder @mode="perform" @noun={{@selectedAction}} />
+    <MessageError @errors={{@errors}} />
     <div class="field">
       <div class="control">
         <JsonEditor @title="Data to wrap" 
           @subTitle="json-formatted"
-          @value={{data}} 
+          @value={{this.data}} 
           @valueUpdated={{action "codemirrorUpdated"}} />
```
hashishaw, line 39: This one can be `{{this.codemirrorUpdated}}` as we discussed as well

ui/app/components/tool-wrap.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+  @action
+  updateTtl(evt) {
+    if (!evt) return;
+    const ttl = evt.enabled ? `${evt.seconds}s` : '30m';
+    this.args.updateTtl(ttl);
+  }
+  @action
+  codemirrorUpdated(val, codemirror) {
+    codemirror.performLint();
```
arnav28, line 45: Can codemirror be undefined or null?

ui/app/components/tool-actions-form.js
```diff
@@ -135,14 +134,11 @@ export default Component.extend(DEFAULTS, {
...
```
arnav28, line 144: Same, can codemirror be undefined or null?

ui/app/components/tool-hash.js
```diff
@@ -0,0 +1,28 @@
...
+ *  @sum={{sum}}
+ *  @algorithm={{algorithm}}
+ *  @format={{format}}
+ *  @errors={{errors}}/>
+ * ```
+ * @param onClear {Function} - parent action that is passed through. Must be passed as {{action "onClear"}}
+ * @param sum=null {String} - property passed from parent to child and then passed back up to parent
+ * @param algorithm {String} - property returned from parent.
+ * @param format {String} - property returned from parent.
+ * @param error=null {Object} - errors passed from parent as default then from child back to parent.
```
arnav28, line 21: Nice job with js doc, I should add it as well 😄

ui/app/components/tool-actions-form.js
Monkeychip: nope, it's always an object that's returned.

ui/app/components/tool-wrap.js
Monkeychip: This is also how we handle it in other parts of the app.

arnav28 (approved):

PR #11678, opened 2021-05-20
Add ulimit check, and tidy unit test cases

sgmiller:
...to avoid needing to have all results and perfect ordering


command/operator_diagnose_test.go
```diff
@@ -66,17 +70,6 @@ func TestOperatorDiagnoseCommand_Run(t *testing.T) {
 				"-config", "./server/test-fixtures/nostore_config.hcl",
 			},
 			[]*diagnose.Result{
-				{
```
HridoyRoy, line 69: Why are we removing these stanzas from the tests?

command/operator_diagnose_test.go
```diff
@@ -237,10 +169,20 @@ func TestOperatorDiagnoseCommand_Run(t *testing.T) {
 				cmd.Run(tc.args)
 				result := cmd.diagnose.Finalize(context.Background())
 
-				for i, exp := range tc.expected {
-					act := result.Children[i]
-					if err := compareResult(t, exp, act); err != nil {
-						t.Fatalf("%v", err)
+				for _, exp := range tc.expected {
+					found := false
+					// Check them all so we don't have to be order specific
```
HridoyRoy, line 174: I really like this improvement to the tests!

command/operator_diagnose_test.go
```diff
@@ -237,10 +169,20 @@ func TestOperatorDiagnoseCommand_Run(t *testing.T) {
...
+					for _, act := range result.Children {
+						if exp.Name == act.Name {
```
HridoyRoy, line 176: I think this won't account for improperly nested checks. For example, if a stanza exists as a child of another stanza and is meant instead to be a top level stanza, this test will still pass because of the nesting here.

command/operator_diagnose.go
```diff
@@ -196,6 +196,11 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error
 
 	ctx, span := diagnose.StartSpan(ctx, "initialization")
 	defer span.End()
+
+	// OS Specific checks
+	// Check open file count
+	diagnose.OSChecks(ctx)
```
HridoyRoy, line 202: If I'm not mistaken, I think this won't open a new top level span. It'll append the OSChecks span to the child of whatever span was currently open. Maybe in this case it's ok since this is the first nested span of the `initialization` span structurally, but then all the future spans will be subspans of this one, which would change the structure. 

I vote we start a new Span and use that context in `diagnose.OSChecks`.

HridoyRoy (changes requested):

command/operator_diagnose.go
NicoleS021: **[](https://wealthsimple.com/invite/NHO6EA)**

NicoleS021 (approved):

command/operator_diagnose_test.go
sgmiller: Ooh, good point.

command/operator_diagnose.go
sgmiller: Your right, it won't, but the tests in OSChecks do spot tests that will add sub spans.  Not sure if it makes sense to have a container span called os-checks or something if it just has one sub check. Maybe in the future when we have more?

command/operator_diagnose_test.go
NicoleS021: #- [ ] []()

NicoleS021:
- [ ] []()[]()

command/operator_diagnose.go
HridoyRoy: Ah, my mistake!

command/operator_diagnose_test.go
sgmiller: In order not to have to update every test case when we add a new diagnose test, the stanzas are only including the results that are relevant to the test case itself.  This means we only have to add a new test case per test rather than updating all of the test case result trees.

HridoyRoy (approved):

PR #11680, opened 2021-05-20
Update cluster status partial to component

arnav28:


changelog/11680.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
Monkeychip, line 1: So Chelsea and I talked about doing one large status update.  Maybe just change this to something more generic that we can use for all the partial updates.  e.g. update partials to components.

ui/app/components/auth-info.js
```diff
@@ -1,42 +1,54 @@
...
+ */
+export default class AuthInfoComponent extends Component {
+  @service auth;
+  @service wizard;
+  @service router;
+
+  @tracked
+  fakeRenew = false;
+
+  get isRenewing() {
```
Monkeychip, line 26: nice!

ui/app/templates/components/cluster-info.hbs
```diff
@@ -1,30 +1,30 @@
 <div class="popup-menu-content">
   <div class="box">
-    {{#unless version.isOSS}}
-      {{#if (and activeCluster.unsealed auth.currentToken)}}
-        {{#unless cluster.dr.isSecondary}}
+    {{#unless this.version.isOSS}}
+      {{#if (and @cluster.unsealed this.auth.currentToken)}}
```
hashishaw, line 4: Based on looking at the code, I'm not sure `activeCluster` as defined on (old) status-menu.js is exactly swappable with `currentCluster.cluster` which is what's being passed into the component here. My guess is that with the full model (activeCluster) we get the replication info and we may not on the cluster service version, but we'll need to investigate

ui/app/templates/components/auth-info.hbs
```diff
@@ -2,38 +2,38 @@
...
     <div class="menu-label">
-      {{auth.authData.displayName}}
+      {{this.auth.authData.displayName}}
     </div>
     <nav class="menu">
       <ul class="menu-list">
-        {{#if canExpire}}
-          <li class="action">
+        {{#if this.auth.allowExpiration }}
+         <li class="action">
```
Monkeychip, line 10: I generally keep an indentation after an if statement.  Double check in the code and see if that's a consistent pattern, otherwise keep as is.

ui/app/templates/components/auth-info.hbs
```diff
@@ -54,7 +54,7 @@
           {{/if}}
         {{/if}}
         <li class="action">
-          <LinkTo @route="vault.cluster.logout" @model={{activeClusterName}} @id="logout" class="is-destroy" @invokeAction={{onLinkClick}}>
+          <LinkTo @route="vault.cluster.logout" @model={{@activeClusterName}} @id="logout" class="is-destroy" @invokeAction={{@onLinkClick}}>
```
Monkeychip, line 57: I noticed elsewhere that you had the action keyword in front of this `@onLinkClick` Maybe confirm that this link is working?

ui/app/templates/components/cluster-info.hbs
```diff
@@ -38,13 +38,13 @@
           {{/if}}
         {{else}}
           {{#if (has-permission 'status' routeParams='replication')}}
-            <nav class="menu">
+           <nav class="menu">
```
Monkeychip, line 41: same here, it looks like you removed the indendentation but then added two indentations below the `nav` element.  I would expect one indendentation after the if and then one after the nav.

Monkeychip (changes requested):
One other check, in case you haven't already, is to run `yarn run test -f="enterprise"` locally.  When we push a branch to OSS it skips all the enterprise test.  This area seems to touch some enterprise feature so it's good practice to run those test just in case.

ui/app/templates/components/auth-info.hbs
arnav28: Yes, it works. This action is coming from the parent.

ui/app/templates/components/cluster-info.hbs
arnav28: I added back the active cluster for now.
In meantime, I will dig further and if required get in touch with Matthew. We can clean up later if required.

Monkeychip (approved):

PR #11682, opened 2021-05-21
Initial list for CODEOWNERS

kalafut:
An initial cut at some CODEOWNERS rules.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,8 @@
+# Each line is a file pattern followed by one or more owners.
+# More on CODEOWNERS files: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
+
+
+/plugins/             @hashicorp/vault-ecosystem
+/builtin/credential/  @hashicorp/vault-ecosystem
+/builtin/logical/     @hashicorp/vault-ecosystem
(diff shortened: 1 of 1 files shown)


calvn (approved):
Since we already have a `.github` directory, can we move this into there?

kalafut:
I debated the location... Terraform has theirs at the root. If we want it to also serve as a more visible bit of documentation I think it's better there, but I'm fine with either location really.

CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# Each line is a file pattern followed by one or more owners.
+# More on CODEOWNERS files: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
+
+
+/plugins/             @hashicorp/vault-ecosystem
```
pcman312, line 5: nit: what do you think about alphabetizing the list? I'm thinking it will help with future additions to the list.

pcman312 (approved):

CODEOWNERS
kalafut: Done

pcman312 (approved):

pcman312:
re: file location: I like it better at the root because it makes it more visible rather than putting it in a metaphorical corner.

PR #11690, opened 2021-05-22
Only fail auth plugin backend mounting for custom plugins

### LINKED ISSUE #11687: Custom plugins mounted over built in plugins can lead to a bricked installation
**Describe the bug**
If a custom plugin is mounted over a built in plugin then removed or otherwise altered, vault will refuse to unseal.

**To Reproduce**
Assuming a working plugin at $PLUGINDIR and a binary at $PLUGINBIN.  In this example vault-plugin-auth-jwt
```
PLUGINDIR=/vault/plugins
PLUGINBIN=vault-plugin-auth-jwt
# Register
vault plugin register -command $PLUGINBIN -sha256 $(openssl dgst -sha256 $PLUGINDIR/$PLUGINBIN | cut -d '=' -f2 | tr -d ' ') oidc
# Verify its our plugin and not builtin
vault read /sys/plugins/catalog/auth/oidc
# Mount it
vault write auth/oidc/config oidc_discovery_url="..." oidc_client_id="..." oidc_client_secret="..." default_role="default"

# stop server however applicable (docker, service, CTRL-C)
mv $PLUGINDIR/$PLUGINBIN $PLUGINDIR/$PLUGINBIN.old

# start vault server however applicable
vault operator unseal ...
```
Logs:
```
[ERROR] core: failed to create credential entry: path=oidc/ error="error verifying checksum: open /vault/plugins/vault-plugin-auth-jwt: no such file or directory"
[INFO]  core: pre-seal teardown starting
[INFO]  core: pre-seal teardown complete
[ERROR] core: post-unseal setup failed: error="failed to setup auth table"
```
Seemingly no recovery from this point.  Vault cannot be unsealed to remove the plugin or in the case of a sha256 mismatch, reregister it.  Very similar to #3602

**Expected behavior**
Vault starts albeit in a degraded state but from which it can be repaired by removing or reregistering the plugin.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.7.1
* Vault CLI Version (retrieve with `vault version`): v1.7.1
* Server Operating System/Architecture: docker amd64 https://hub.docker.com/_/vault

```hcl
storage "file" {
  path = "/mnt/vault-data"
}

listener "tcp" {
  address         = "0.0.0.0:80"
  tls_disable     = "true"
}

plugin_directory = "/vault/plugins"
disable_mlock   = true
api_addr = "http://example.com:8080"

disable_clustering  = true

log_level           = "Trace"

default_lease_ttl   = "1h"
max_lease_ttl       = "1h"

plugin_directory    = "/vault/plugins"
ui                  = true

```

**Additional context**
Having now reviewed the relevant code sections, this might be caused by the logic of: https://github.com/hashicorp/vault/blob/6d6ea1840a11fa9f921b49282c1bac30f3715f97/vault/auth.go#L688 

As I read this, the intended purpose is to continue on with mounting if the plugin is not a builtin.  Unfortunately the lookup here only matches the plugin type to the builtin registry rather than actually checking to see if it's a built in or not.  That information is available, just not being used here.  A possible solution would be to query the plugin catalog instead.  This builds but I have not yet tested it resolves my issue as a complete build of vault with the ui takes a lifetime on my machine.
```
diff --git a/vault/auth.go b/vault/auth.go
index 5b5394716..39c4b19bc 100644
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -685,7 +685,7 @@ func (c *Core) setupCredentials(ctx context.Context) error {
                backend, err = c.newCredentialBackend(ctx, entry, sysView, view)
                if err != nil {
                        c.logger.Error("failed to create credential entry", "path", entry.Path, "error", err)
-                       if !c.builtinRegistry.Contains(entry.Type, consts.PluginTypeCredential) {
+                       if plug, plugerr := c.pluginCatalog.Get(ctx, entry.Type, consts.PluginTypeCredential); plugerr == nil && !plug.Builtin {
                                // If we encounter an error instantiating the backend due to an error,
                                // skip backend initialization but register the entry to the mount table
                                // to preserve storage and path.
```
The need to mount this particular custom plugin or top of the existing builtin is to retain the ui configuration and authentication listing visibility of the original oidc plugin.  Mounting this with an alternate path and preserving those features does not seem possible.

corybolar:
Checking if a plugin is a builtin by comparing it's type to those in the builtin
registry allows for a custom plugin loaded with the same name to be considered a
builtin during error handling of the mounting process.  This can cause the vault
installation to brick itself because it cannot be unsealed to register a new
sha256 or file path for a previously loaded custom plugin.  Improve this logic
by checking the plugin catalog rather than the builtin registry.

Fixes #11687

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -685,7 +685,7 @@ func (c *Core) setupCredentials(ctx context.Context) error {
 		backend, err = c.newCredentialBackend(ctx, entry, sysView, view)
 		if err != nil {
 			c.logger.Error("failed to create credential entry", "path", entry.Path, "error", err)
-			if !c.builtinRegistry.Contains(entry.Type, consts.PluginTypeCredential) {
+			if plug, plugerr := c.pluginCatalog.Get(ctx, entry.Type, consts.PluginTypeCredential); plugerr == nil && !plug.Builtin {
 				// If we encounter an error instantiating the backend due to an error,
 				// skip backend initialization but register the entry to the mount table
 				// to preserve storage and path.
--- a/vault/logical_system_integ_test.go
+++ b/vault/logical_system_integ_test.go
@@ -245,26 +245,34 @@ func testPlugin_CatalogRemoved(t *testing.T, btype logical.BackendType, testMoun
 func TestSystemBackend_Plugin_continueOnError(t *testing.T) {
 	t.Run("secret", func(t *testing.T) {
 		t.Run("sha256_mismatch", func(t *testing.T) {
-			testPlugin_continueOnError(t, logical.TypeLogical, true, consts.PluginTypeSecrets)
+			testPlugin_continueOnError(t, logical.TypeLogical, true, "mock-plugin", consts.PluginTypeSecrets)
 		})
 
 		t.Run("missing_plugin", func(t *testing.T) {
-			testPlugin_continueOnError(t, logical.TypeLogical, false, consts.PluginTypeSecrets)
+			testPlugin_continueOnError(t, logical.TypeLogical, false, "mock-plugin", consts.PluginTypeSecrets)
 		})
 	})
 
 	t.Run("auth", func(t *testing.T) {
 		t.Run("sha256_mismatch", func(t *testing.T) {
-			testPlugin_continueOnError(t, logical.TypeCredential, true, consts.PluginTypeCredential)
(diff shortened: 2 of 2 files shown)


sgmiller (approved):

sgmiller:
Looks good, thanks for your work on this subtle issue.

PR #11694, opened 2021-05-24
cleanup unused code and fix t.Fatal usage in goroutine in testing

mayocream:
Description:
- cleanup unused code
- simply code (like `errors.New(fmt.Errorf())` to `fmt.Errorf()`)
- change t.Fatal to "t.Error and return" in goroutine in testing, why: https://github.com/ipfs/go-ipfs/issues/2043 , how: I followed this PR https://github.com/minio/minio/pull/6682

These changes should not change the behavior of code.

Appended changelog.


mayocream:
Sorry about adding harmful error handler in `agent_test.go`, I didn't noticed these comments that need time waiting for template files to be generated and loop for retrying.
I am trying fix it.

changelog/11694.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
ncabatoff, line 1: We don't usually do changelog entries if there's no user-visible impact.  Since as you say these changes should be no-ops (ignoring test changes), I suggest you delete this file.

command/server.go
```diff
@@ -828,14 +827,14 @@ type quiescenceSink struct {
...
-func (q quiescenceSink) Accept(name string, level log.Level, msg string, args ...interface{}) {
+func (q quiescenceSink) Accept(name string, level hclog.Level, msg string, args ...interface{}) {
 	q.t.Reset(100 * time.Millisecond)
 }
 
 func (c *ServerCommand) setupStorage(config *server.Config) (physical.Backend, error) {
 	// Ensure that a backend is provided
 	if config.Storage == nil {
-		return nil, fmt.Errorf("A storage backend must be specified")
+		return nil, errors.New("A storage backend must be specified")
```
ncabatoff, line 837: I prefer to use `fmt.Errorf` over `errors.New`, except in the case where I'm declaring a variable to hold an error which I will compare against in code to see if an error is of a particular type, since this is a widespread convention.  Rationale: consistency, as most returned errors will include parameters, why have a second way of returning an error that's only very rarely used?

ncabatoff (commented):
Two minor changes requested, but overall looks good.  And yes, the pgp test failures aren't your fault, and we're looking at them.

changelog/11694.txt
mayocream: I'll delete this changelog file, thanks!

command/server.go
mayocream: Consistency is important. In my opinion, `fmt.Errorf` is useful when these are other args to include, as this function call `errors.New` if no args provided. 
I found this file that include both `fmt.Errorf` and `errors.New` usages: https://github.com/mayocream/vault/blob/0ef61459b82d1c4287e3398540c0e3114fc2a3c8/builtin/logical/ssh/path_sign.go#L310
This could be different contributors using their way to return an error.

ncabatoff: I would reject this change as gratuitous if it were on its own - since both ways of returning a parameter-less error are equally valid, and it largely comes down to taste.  I wouldn't want to flip-flop between styles based on who the latest author is.  However, since there's a lot of other good stuff in this PR I'll allow it.

In future if I can get consensus internally for either approach (i.e. either "always use errors.New when there are no parameters" or "always use fmt.Errorf regardless") I'll look at adding a semgrep check to enforce that style.  Like with gofmt, it's better if there's only one way to do it so that people don't have to choose.

ncabatoff (approved):

ncabatoff:
Sorry @mayocream, I know that the only reason this is stale is because I took a long time to get back to you.  All the same, would you mind pulling in the latest changes from the `main` branch?  I feel a little uncomfortable merging it as-is.

mayocream:
> Sorry @mayocream, I know that the only reason this is stale is because I took a long time to get back to you. All the same, would you mind pulling in the latest changes from the `main` branch? I feel a little uncomfortable merging it as-is.

Sorry for replying late, I merged `main` branch into this PR.
And there's no any changes since last commit except merge things.

I think one of the tests failed is not related to this PR: https://app.circleci.com/pipelines/github/hashicorp/vault/21173/workflows/e7c3298c-38c0-43a8-8167-5631f09cd857/jobs/296348

ncabatoff:
Thanks!

PR #11700, opened 2021-05-25
Replace auth-form/* partials

Monkeychip:
Replace all partials for the auth-form.  I double checked and userpass-form is only used in this area.  Also, I ran enterprise test locally several times and they all pass.


hashishaw (approved):
Great stuff! So much more legible too

PR #11702, opened 2021-05-25
UI/partials replace partials/role-ssh partials/role-aws partials/role-pki

Monkeychip:
Looking at the convention in other components it made sense to move all of these partials inline on their original components.  This extended the file length, but I think makes it easier to read in the end.

I also ran enterprise test locally and tested and things looked good.


NicoleS021:
https://github.com/hashicorp/vault/commit/fbff9bbf1be87da0d78a430084af8986f1ffba17#r51307027[](https://github.com/hashicorp/vault/commit/fbff9bbf1be87da0d78a430084af8986f1ffba17#r51307027)

NicoleS021:
[]()

hashishaw (approved):
Nice work!

PR #11705, opened 2021-05-26
UI unseal screen updates

hashishaw:
This change updates the style and language on the Vault unseal screen, and adds a specific view when a login attempt fails with the error `licensing is in an invalid state`. There are now links to relevant documentation on both the unseal form and the license error view.

As you can see from the gif below, it still shows other errors correctly on the unseal form itself, and then disallows the ability to try to unseal once a licensing error is detected:

unseal-license-terminated

### DIFF
--- a/ui/lib/core/addon/components/shamir-flow.js
+++ b/ui/lib/core/addon/components/shamir-flow.js
@@ -42,7 +42,7 @@ export default Component.extend(DEFAULTS, {
   },
 
   onUpdate() {},
-  onError() {},
+  onLicenseError() {},
   onShamirSuccess() {},
   // can be overridden w/an attr
   isComplete(data) {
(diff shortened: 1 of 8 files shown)


ui/lib/core/addon/components/shamir-flow.js
```diff
@@ -152,6 +157,7 @@ export default Component.extend(DEFAULTS, {
     },
 
     onSubmit(data) {
+      console.log('on submit', data);
```
Monkeychip, line 160: Looks like this was accidently left in.

Monkeychip (approved):
Looks great.  Only additional comments are a changelog and you currently have the label at 1.8 but shouldn't it be 1.7 if we're backporting it to 1.7?

ui/lib/core/addon/components/shamir-flow.js
hashishaw: Good catch!

arnav28 (approved):

PR #11710, opened 2021-05-26
secret-list/item partial removed

Monkeychip:
Removing all partials and turning into components from the folder `partials/secret-list/*`.  Only a couple of little catches, the coolest being the use of `let` and nested components to conditionally render the component, very similar to how we previously rendered the partials.

Other catches included passing very specific properties like `itemType` and `itemPath` which are only used in the transform component.  However, after playing with various options it's cleaner to have all the logic stay in the `list.js` file and pass these properties through even though they are only used by one of the components.  Open for discussion on this if folks think otherwise.


ui/app/helpers/options-for-backend.js
```diff
@@ -8,7 +8,7 @@ const DEFAULT_DISPLAY = {
   create: 'Create secret',
   navigateTree: true,
   editComponent: 'secret-edit',
-  listItemPartial: 'partials/secret-list/item',
+  listItemPartial: 'secret-list/item',
```
hashishaw, line 11: the naming for this key is no longer accurate, maybe we rename to `listItemComponent`?

ui/app/templates/partials/secret-list/transform-list-item.hbs
```diff
@@ -1,5 +0,0 @@
-<TransformListItem
```
hashishaw, line 1: Ha! We were already moving this direction 😎

ui/app/templates/vault/cluster/secrets/backend/list.hbs
```diff
@@ -57,8 +57,21 @@
 
   {{#if model.meta.total}}
     {{#each model as |item|}}
-      {{partial options.listItemPartial}}
-    {{else}}
+    {{!-- Because of the component helper cannot use glimmer nested SecretList::Item --}}
+      {{#let (component options.listItemPartial) as |Component|}}
```
hashishaw, line 61: Very nice solution 👍 Not sure I totally understand the comment above, but I'm gathering it's justifying the `#let`?

hashishaw (approved):
Great work! Thanks for tackling this, just a couple non-blocking comments

ui/app/templates/vault/cluster/secrets/backend/list.hbs
Monkeychip: It's more about having the `options.listItemPartial` return `SecretList::Item` which is the new ember way of nested components, we have to use the old way of `secret-list/item` because of the `component` helper.  It was a TIL thing as I went down the rabbit hold of not understanding why it couldn't find my component.

PR #11113, opened 2021-03-16
Rework agent retry config, extend it to cover proxy cache as well

ncabatoff:
Remove template_retry config section.  Add new vault.retry section which only has num_retries field; if num_retries is 0 or absent, default it to 12 for backwards compat.

Configured retries are used for both templating and api proxy, though if template requests go through proxy (currently requires persistence enabled) we'll only configure retries for the latter to avoid duplicate retrying.  Though there is some duplicate retrying already because whenever the template server does a retry when not going through the proxy, the Vault client it uses allows for two behind-the-scenes retries for some 400/500 http error codes.

PR #11711, opened 2021-05-26
agent: allow templating to indefinitely retry if explicitly set to zero

calvn:
This PR re-introduces the ability for agent's template subsystem to perform indefinite retries that was removed in 1.6.0 through #9670. Indefinite retry can be toggled on by setting `vault.retry.num_retries = 0`. 

Since `vault.retry.num_retries` is used by both the cache client and the template config, the changeset moves setting these values closer to where the update is called. This reduces the behavioral change to the bare minimum, but introduces the difference on what `0` means for template vs cache.

Templating:
| vault.retry.num_retries  | Old | New |
| ------------- | ------------- | ----- |
| -1  | Disabled  | Disabled   |
| 0  | Default (12x)  |  **∞ retries** |
| N > 0 | N retries | N retries |
| Absent | Default (12x) | Default (12x) |

Cache:
| vault.retry.num_retries | Old           | New           |
|-------------------------|---------------|---------------|
| -1                      | Disabled      | Disabled      |
| 0                       | Default (12x) | **API Client default (2x)**     |
| N > 0                   | N retries     | N retries     |
| Absent                  | Default (12x) | Default (12x) |

Related to #9670
Related to #11113

Backport note:
We need to backport https://github.com/hashicorp/vault/pull/11696 if/when we backport this PR.

TODO:
- [ ] Update/add tests
- [ ] Update docs
- [ ] Include changelog

### DIFF
--- a/command/agent.go
+++ b/command/agent.go
@@ -55,6 +55,11 @@ import (
 var (
 	_ cli.Command             = (*AgentCommand)(nil)
 	_ cli.CommandAutocomplete = (*AgentCommand)(nil)
+
+	// defaultClientRetryAttempts is the default number of retries that the
+	// client will make if vault.retry.num_retries is not specified in the
+	// config. 12 is chosen as the value for backward compatibility.
+	defaultClientRetryAttempts = 12
 )
 
 type AgentCommand struct {
@@ -420,7 +425,17 @@ func (c *AgentCommand) Run(args []string) int {
 	// confuse the issue of retries for auth failures which have their own
 	// config and are handled a bit differently.
 	if os.Getenv(api.EnvVaultMaxRetries) == "" {
-		client.SetMaxRetries(config.Vault.Retry.NumRetries)
+		retries := config.Vault.Retry.NumRetries
+
+		// Overrides here to maintain backward compatibility in behavior
+		switch config.Vault.Retry.NumRetries {
+		case -1:
+			retries = 0
+		case 0:
+			retries = defaultClientRetryAttempts
+		}
+
+		client.SetMaxRetries(retries)
 	}
 
 	enforceConsistency := cache.EnforceConsistencyNever
--- a/command/agent.go
+++ b/command/agent.go
@@ -58,7 +58,8 @@ var (
 
 	// defaultClientRetryAttempts is the default number of retries that the
 	// client will make if vault.retry.num_retries is not specified in the
-	// config. 12 is chosen as the value for backward compatibility.
+	// config. 12 is chosen as the value for backward compatibility and was
+	// based on the value from ctconfig.DefaultRetryAttempts.
 	defaultClientRetryAttempts = 12
 )
 
@@ -425,17 +426,20 @@ func (c *AgentCommand) Run(args []string) int {
 	// confuse the issue of retries for auth failures which have their own
 	// config and are handled a bit differently.
 	if os.Getenv(api.EnvVaultMaxRetries) == "" {
-		retries := config.Vault.Retry.NumRetries
 
 		// Overrides here to maintain backward compatibility in behavior
 		switch config.Vault.Retry.NumRetries {
 		case -1:
-			retries = 0
+			client.SetMaxRetries(0)
 		case 0:
-			retries = defaultClientRetryAttempts
+			// If the value was not explicitly provided, we set to it the
+			// internal default for backward compatibility.
+			if config.Vault.Retry.NumRetriesRaw == nil {
+				client.SetMaxRetries(defaultClientRetryAttempts)
+			}
+		default:
+			client.SetMaxRetries(config.Vault.Retry.NumRetries)
 		}
-
-		client.SetMaxRetries(retries)
 	}
 
 	enforceConsistency := cache.EnforceConsistencyNever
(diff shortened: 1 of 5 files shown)


calvn:
One case that may look odd -- if `vault.retry.num_retries = 0` and `persist` is non-nil, it means that the client will try 12x per request. Functionally, this doesn't matter too much since the number of attempts by template is infinite, but does make each template attempt be performed 12x by the cache client.

command/agent.go
```diff
@@ -420,7 +425,17 @@ func (c *AgentCommand) Run(args []string) int {
...
 	if os.Getenv(api.EnvVaultMaxRetries) == "" {
-		client.SetMaxRetries(config.Vault.Retry.NumRetries)
+		retries := config.Vault.Retry.NumRetries
+
+		// Overrides here to maintain backward compatibility in behavior
+		switch config.Vault.Retry.NumRetries {
+		case -1:
+			retries = 0
+		case 0:
+			retries = defaultClientRetryAttempts
```
calvn, line 435: I'm tempted to change this to be `api.Client` default of 2 retries here (for a total of 3 calls). Even though this is the current value, the client actually times out before making all 12 attempts due to the default request timeout value of 60s.

```
[WARN] (view) vault.read(database/creds/readonly): vault.read(database/creds/readonly): context deadline exceeded (retry attempt 1 after "250ms")
[TRACE] (view) vault.read(database/creds/readonly) starting fetch
[TRACE] vault.read(database/creds/readonly): starting renewer
2021-05-26T15:54:39.425-0700 [INFO]  cache: received request: method=PUT path=/v1/sys/leases/renew
2021-05-26T15:54:39.425-0700 [DEBUG] cache.leasecache: forwarding request: method=PUT path=/v1/sys/leases/renew
2021-05-26T15:54:39.425-0700 [INFO]  cache.apiproxy: forwarding request: method=PUT path=/v1/sys/leases/renew
[WARN] vault.read(database/creds/readonly): failed to renew: context deadline exceeded
[WARN] vault.read(database/creds/readonly): renewer done (maybe the lease expired)
[TRACE] vault.read(database/creds/readonly): GET /v1/database/creds/readonly
2021-05-26T15:55:39.428-0700 [INFO]  cache: received request: method=GET path=/v1/database/creds/readonly
2021-05-26T15:55:39.428-0700 [DEBUG] cache.leasecache: forwarding request: method=GET path=/v1/database/creds/readonly
2021-05-26T15:55:39.428-0700 [INFO]  cache.apiproxy: forwarding request: method=GET path=/v1/database/creds/readonly
[WARN] (view) vault.read(database/creds/readonly): vault.read(database/creds/readonly): context deadline exceeded (retry attempt 2 after "500ms")
[TRACE] (view) vault.read(database/creds/readonly) starting fetch
[TRACE] vault.read(database/creds/readonly): starting renewer
```

Another suggestion that was made was to break retry values by the cache's client and the template to be different values, e.g. `vault.retry.cache_retries` and `vault.retry.template_retries`. We'd still have to support the existing `num_retries` even if we deprecate it though.

calvn:
Another suggestion that was made was to break retry values by the cache's client and the template to be different values, e.g. `vault.retry.cache_retries` and `vault.retry.template_retries`. We'd still have to support the existing `num_retries` even if we deprecate it though.

command/agent.go
jasonodonnell: I'm OK with this change

command/agent/template/template.go
```diff
@@ -247,12 +247,18 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
...
+	// that the client will retry 12 times per request/attempt but this
+	// shouldn't matter since template will indefinitely retry.
+	enabled := attempts >= 0 && sc.AgentConfig.Vault.Retry.NumRetriesRaw != nil
+
 	// Use the cache if available or fallback to the Vault server values.
 	// For now we're only routing templating through the cache when persistence
 	// is enabled. The templating engine and the cache have some inconsistencies
 	// that need to be fixed for 1.7x/1.8
 	if sc.AgentConfig.Cache != nil && sc.AgentConfig.Cache.Persist != nil && len(sc.AgentConfig.Listeners) != 0 {
-		attempts = 0
```
ncabatoff, line 255: Don't we still want to disable retries if using persistent cache?

calvn: We want to support the ability for template to indefinitely retry regardless of whether persistent cache is enabled. Since the cache client doesn't support indefinite retries and 0 on template really mean infinite retries, we can't hardcode 0 here.

I think that the logic that's missing here is something along the lines of:

```go
// Disable template retries if attempts is greater than zero
// so that we don't end up performing n^2 retries by both
// template and the cache client.
if attempts > 0 {
    enabled = false
}
```

calvn: Addressed in 9fab0c1

command/agent.go
calvn: I've gone ahead an made this change on 03d248f, and also updated the table on the PR description.

calvn:
Closing in favor of https://github.com/hashicorp/vault/pull/11775.

PR #11719, opened 2021-05-27
Dependabot security alert fixes

arnav28:


website/package.json
```diff
@@ -42,6 +42,10 @@
     "inquirer": "^8.1.0",
     "prettier": "^2.3.0"
   },
+  "resolutions": {
+    "trim": "^0.0.3",
+    "node-fetch": "^2.6.1"
+  },
```
jescalan, line 48: I'm not sure this is the way we'd want to go about fixing these - this is a fairly unusual fix. Is there any way we could leave this one out of this specific patch?

jescalan (changes requested):

hashishaw (approved):

website/package.json
hashishaw: Out of curiosity, how do you go about fixing these when they aren't direct dependencies? Our resolutions section is getting fairly large and I'd love to find a different way to address this

arnav28: Sure, I will remove it from the website for now and we can discuss it offline. Thanks :)

jescalan: Generally, we try to update the dependency and/or get an update merged upstream wherever possible 😀

jescalan (approved):

PR #11729, opened 2021-05-31
(OSS) VAULT-2446: Re-enable server tests using env var VAULT_LICENSE_CI

ncabatoff:
We'd disabled these tests on enterprise temporarily, but now we're ready to make them run there again.

Vault cares about the VAULT_LICENSE env var, but we don't want to set that in CI because it would change behaviour of tests that don't intend for it to be set.  Instead, we use VAULT_LICENSE_CI so that only packages/tests that opt-in will use it.

### DIFF
--- a/command/server_test.go
+++ b/command/server_test.go
@@ -1,4 +1,4 @@
-// +build !race,!hsm,!enterprise
+// +build !race,!hsm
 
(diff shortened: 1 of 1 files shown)


victorr (approved):
👍🏽

PR #11730, opened 2021-05-31
Restore lost server license config propagation

ncabatoff:
Fixes regression from #11573.


victorr (approved):
I diffed the new CoreConfig against what it was before the changes in #11573 and there are no differences now.

PR #11731, opened 2021-05-31
go_test.yml on ent has needlessly diverged from oss; make them identical again

ncabatoff:
This is actually not 100% identical to what's on ent, but that's ok, once we merge to ent they will be.


victorr:
The changes look good to me.

Are both cases (using a Docker containers vs not using one) used by the circleci jobs?

victorr (approved):

ncabatoff:
> Are both cases (using a Docker containers vs not using one) used by the circleci jobs?

@victorr yes everything under .circleci is for CircleCI.

PR #11735, opened 2021-06-01
Fixing issue with FPE read and delete api docs

basingh:
Read and Delete FPE api-docs point to path `/transform/transformations/:name` instead it should be `/transform/transformation/:name`


sgmiller (approved):
Thanks for spotting this.

PR #11736, opened 2021-06-01
Put go_test.yml back how it was (on ent) re licensing

ncabatoff:
There was a good reason for doing it the convoluted way we were doing it before, restore that and explain why in a comment.


victorr (approved):
👍🏽

PR #11737, opened 2021-06-01
Partials Transit removal

Monkeychip:
Very similar to the other processes: removal of a transit-form-* partials.


hashishaw (approved):
👏 Great work!

PR #11739, opened 2021-06-01
Removing all secret-form-*

Monkeychip:
These three partials `secret-form-create, secret-form-edit, secret-form-show` were not nested inside a folder but instead in the generic partials folder.  For the create and edit partial I ended up putting them inline on the `secret-edit.hbs` for a couple of reasons: I originally tried making them separate components, but the number of properties I had to pass in—largely because of the nested `SecretEditDisplay` component— made the actual declaration of the component pretty long.  Second, it also didn't work.  There are funky things with updating the model that were not properly being passed through.  After trying to problem solve I realized I was only saving about 20 lines of inline code by using a component and decided it was safer to remove the component and move the partial code inline.

For the `secret-form-show` partial this I was able to put into a component and the declaration was clean.


hashishaw (approved):
Thanks for including your reasoning in the description, it makes a lot of sense. Great work!

PR #11749, opened 2021-06-02
Segment out disk checks to disable on openbsd/arm

sgmiller:
Also add a spot skipped helper.

### DIFF
--- a/command/operator_diagnose.go
+++ b/command/operator_diagnose.go
@@ -211,9 +211,7 @@ func (c *OperatorDiagnoseCommand) offlineDiagnostics(ctx context.Context) error 
 	defer span.End()
 
 	// OS Specific checks
-	// Check open file count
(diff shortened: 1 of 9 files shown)


swayne275 (commented):
I think it looks good but I have a couple questions

1) why do we need to skip the disk check on openbsd and arm?
2) where does the windows version of diskUsage() live? I didn't see a version of it in `vault/diagnose/os_windows.go`, and it looks like the original version being used only got moved to the `../os_unix_other.go` file

sgmiller:
The library that we use for disk usage checks doesn't build on openbsd/arm, there are various constants that aren't defined for that platform.  We've opened a bug with the library authors.  Also there aren't any good alternatives to that library with licenses we can use.

Good catch on diskUsage which should be built for Windows, fixed.

swayne275 (approved):

PR #11759, opened 2021-06-03
UI/license banners

hashishaw:
Add license expiry warning and danger messages when `sys/health` endpoint returns license info. If no `license` key supplied, falls back to normal behavior (no banner)

**When license will expire within 30 days, warning shown globally:**
Screen Shot 2021-06-03 at 10 36 14 AM

**When license expiry date has passed, danger shown globally:**
Screen Shot 2021-06-03 at 10 35 49 AM

### DIFF
--- a/changelog/11759.txt
+++ b/changelog/11759.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: show site-wide banners for license warnings if applicable
+```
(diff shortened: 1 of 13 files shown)


Monkeychip (approved):
This looks really great.  Nice work!!  Only thing, do we need to add a changelog?

PR #11760, opened 2021-06-03
Add Advice and Advise functions for adding an advice section to any span.

sgmiller:
Also finishes implementing indentation and adds word wrapping to messages,
warnings, and advice.

### DIFF
--- a/command/operator_diagnose.go
+++ b/command/operator_diagnose.go
@@ -161,7 +161,6 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
 			c.diagnose = diagnose.New(os.Stdout)
 		}
 	}
-	c.UI.Output(version.GetVersion().FullVersionNumber(true))
 	ctx := diagnose.Context(context.Background(), c.diagnose)
 	c.diagnose.SetSkipList(c.flagSkips)
 	err := c.offlineDiagnostics(ctx)
--- a/command/operator_diagnose.go
+++ b/command/operator_diagnose.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"golang.org/x/term"
 	"io"
 	"os"
 	"strings"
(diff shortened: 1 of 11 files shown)


command/operator_diagnose.go
```diff
@@ -176,7 +176,12 @@ func (c *OperatorDiagnoseCommand) RunWithParsedFlags() int {
 		c.UI.Output(string(resultsJS))
 	} else {
 		c.UI.Output("\nResults:")
-		results.Write(os.Stdout)
+		w, _, err := term.GetSize(0)
```
swayne275, line 179: i'm not sure what's going on here.

i looked up the doc for `term.GetSize()` and it looks like it gives the width, height, and error of the current terminal window. I'm not sure what the `0` input means, or why we'd want to return the width?

vault/diagnose/output.go
```diff
@@ -310,62 +273,107 @@ func (t *TelemetryCollector) getOrBuildResult(id trace.SpanID) *Result {
 	return r
 }
 
+func findAttributes(e trace.Event, attr1, attr2 attribute.Key) (string, string) {
+	var av1, av2 string
```
swayne275, line 277: why are these called `av1` and `av2`?

vault/diagnose/output.go
```diff
@@ -310,62 +273,107 @@ func (t *TelemetryCollector) getOrBuildResult(id trace.SpanID) *Result {
...
-			sb.WriteString(status_failed)
+			prelude = status_failed
 		case SkippedStatus:
-			sb.WriteString(status_skipped)
+			prelude = status_skipped
 		}
-		sb.WriteString(r.Name)
+		prelude = prelude + r.Name
 
 		if r.Message != "" || len(r.Warnings) > 0 {
```
swayne275, line 334: why do we check message or warnings, but then only include the message?

swayne275 (commented):
looks pretty good overall

command/operator_diagnose.go
sgmiller: Yeah, we're trying to determine the with of the terminal we're running in.  The 0 is the file descriptor for "stdout", which identifies the correct TTY for the terminal.  We're only interested in width, which is what we pass to set the wordwrap limit.

vault/diagnose/output.go
sgmiller: attribute value 1 and 2.  This function is generically grabbing up two two attribute values given their keys.

command/operator_diagnose.go
swayne275: ohhhhh totally makes sense now. thanks!

vault/diagnose/output.go
sgmiller: yeah, that's not quite right.  The intent is to output the message if present, or the first warning if there are warnings but no message.  But this is outputting a blank message if there is no message but some warnings.  Will fix.

swayne275 (approved):

PR #11765, opened 2021-06-03
replace last two partials

Monkeychip:
Removing the last of the partials.

Notes:
- the `paritals/backend-details.hbs` I could not find any evidence of it being called anywhere or being used anywhere.  Double checked by running the test suite with a `log` on the file and got nothing.
- the form-field-groups-loop partial i had to add a js file because of the use of `this` in the togglebutton.  `This` only works if there's a js file present.  Even if it's empty.
- Added to the ssh-test to cover the new component functionality. Checking that it renders properly and that the actions are passed through okay (testing specifically the ttl action).
- made some small changes to the enterprise-replication-test.  I was getting local failures, made some small changes and now I'm not getting local failures.


ui/app/templates/components/config-pki-ca.hbs
```diff
@@ -83,6 +86,7 @@
     </form>
   {{/if}}
 {{else if signIntermediate}}
+{{log "here1"}}
```
hashishaw, line 89: remove

ui/app/templates/components/form-field-from-model.hbs
```diff
@@ -0,0 +1,84 @@
+<div class="field">
```
hashishaw, line 1: Doesn't really work this time because there's no corresponding js file, but one day I'd like to have JSDocs or the like describing what type of attributes this accepts and what it does. 
Also now that it's a component maybe we should add a test?

hashishaw (changes requested):
A couple changes to make. Really awesome to see all the partials disappear ✨

ui/app/templates/components/form-field-from-model.hbs
Monkeychip: Good call.  I'll work on the test.

ui/app/templates/components/config-pki-ca.hbs
Monkeychip: thank you.

ui/tests/acceptance/enterprise-replication-test.js
```diff
@@ -275,10 +275,10 @@ module('Acceptance | Enterprise | replication', function(hooks) {
     await click('[data-test-secondary-add]');
     await settled();
     await pollCluster(this.owner);
-
+    await settled();
```
Monkeychip, line 278: was running into flakyness locally on this test and tackled it quickly here.

ui/app/components/form-field-groups-loop.js
```diff
@@ -0,0 +1,3 @@
+import Component from '@glimmer/component';
+
+export default class FormFieldGroupsLoop extends Component {}
```
hashishaw, line 3: Can we add JSDocs here so it's easily discoverable what the component takes as args?

hashishaw (approved):

PR #11768, opened 2021-06-03
Update secret backend partial to components

arnav28:


hashishaw (approved):
Looks great! Nice work

PR #11771, opened 2021-06-04
Diagnose checks for raft quorum status and file backend permissions

HridoyRoy:


vault/diagnose/raft_checks.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		SpotError(ctx, "raft file permission checks", fmt.Errorf("error computing file permissions: %w", err))
+	}
+
+	mode := info.Mode()
+
+	canRead := false
+	canWrite := false
+	for i := 1; i < 4; i++ {
+		perm := string(mode.String()[i])
```
sgmiller, line 32: You should be able to test these using bitwise and.  For example, to test for o:rw, mode & 0700 != 0.  Might be a touch cleaner than string comparison.

vault/diagnose/raft_checks.go
```diff
@@ -0,0 +1,102 @@
...
+	"strings"
+
+	"github.com/hashicorp/vault/physical/raft"
+)
+
+// RaftFilePermsChecks inputs the raft file path and outputs:
+// 1. A bool, which dictates whether Owner has rw permissions
+// 2. A list of errors
+func RaftFilePermsChecks(ctx context.Context, path string) {
+	owner := "owner"
```
sgmiller, line 16: Maybe go with consts for these rather than vars?

vault/diagnose/raft_checks.go
sgmiller: On the other hand, you're not having to re-synthesize the perm strings, so maybe it's a wash.  Something about this just feels a touch off...

vault/diagnose/raft_checks.go
```diff
@@ -0,0 +1,102 @@
...
+		okFlag = false
+		SpotWarn(ctx, "raft file permission checks", "raft storage backend file is a symlink")
+	}
+	if len(overPermissions) > 0 {
+		okFlag = false
+		SpotWarn(ctx, "raft file permission checks", fmt.Sprintf("too many permissions -- %s", strings.Join(overPermissions, ",")))
+	}
+	if !canRead || !canWrite {
+		okFlag = false
+		SpotError(ctx, "raft file permission checks", fmt.Errorf("error: not enough permissions -- owner read: %v, owner write: %v", canRead, canWrite))
```
sgmiller, line 71: Instead of not enough permissions, maybe just "owner needs read and write permissions"

vault/diagnose/raft_checks.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	if okFlag {
+		SpotOk(ctx, "raft file permission checks", "correct permissions to raft storage backend file path")
+	}
```
sgmiller, line 75: One other thing I'd like if you can pull it off is to check whether the target files are owned by root and only owner has read/write perms.  That would imply Vault has to run as root which we should warn about.  Something like "raft backend files owned by root and only accessible as root" with an Advise of "This prevents Vault from running as a non-privileged user".

vault/diagnose/raft_checks.go
```diff
@@ -0,0 +1,102 @@
...
+		SpotError(ctx, "raft quorum", fmt.Errorf("error retrieving server configuration: %w", err))
+		return fmt.Sprintf("error retrieving server configuration: %s", err.Error())
+	}
+	voterCount := 0
+	for _, s := range conf.Servers {
+		if s.Voter {
+			voterCount++
+		}
+	}
+	if voterCount == 1 || voterCount == 3 || voterCount == 5 || voterCount == 7 {
```
sgmiller, line 94: Isn't there a warning case when there is only one voter.  Something about not being highly available?

sgmiller (changes requested):
Few comments...

vault/diagnose/raft_checks.go
HridoyRoy: Yeah, we can definitely do that!

vault/diagnose/raft_checks.go
HridoyRoy: That sounds valuable for sure, I can look into it!

vault/diagnose/raft_checks.go
HridoyRoy: 💯

vault/diagnose/raft_checks.go
HridoyRoy: Yeah, I did find the bitwise way to do this as well, but I wondered if this would be more human-readable. Either way is okay with me though -- I'll change this to use bitwise operations for now, and we can see if that looks cleaner. If not, I can always revert.

vault/diagnose/raft_checks.go
```diff
@@ -0,0 +1,93 @@
...
+		nonHAWarning := "warning: only one server node found. Vault is not running in high availability mode"
+		SpotWarn(ctx, "raft quorum", nonHAWarning)
+		return nonHAWarning
+	}
+	if voterCount == 3 || voterCount == 5 || voterCount == 7 {
+		okMsg := fmt.Sprintf("voter quorum exists: %d voters", voterCount)
+		SpotOk(ctx, "raft quorum", okMsg)
+		return okMsg
+	}
+	warnMsg := fmt.Sprintf("error: even number of voters found: %d", voterCount)
```
sgmiller, line 90: Not *quite* accurate, if voterCount == 9 you'll report it as even.  Maybe voter % 2 == 0

physical/raft/raft_test.go
```diff
@@ -334,6 +334,86 @@ func TestRaft_Backend_ThreeNode(t *testing.T) {
...
+	raft3.fsm.upgradeLocalNodeConfig()
+
+	// Take everything down and ensure that GetConfigurationOffline yields 2 voters and 1 nonvoter
+	raft3.TeardownCluster(nil)
+	raft2.TeardownCluster(nil)
+	raft1.TeardownCluster(nil)
+
+	conf, err := raft1.GetConfigurationOffline()
+	c2, _ := raft2.GetConfigurationOffline()
+	c3, _ := raft2.GetConfigurationOffline()
```
briankassouf, line 407:
```suggestion
	c3, _ := raft3.GetConfigurationOffline()
```

vault/diagnose/file_checks.go
```diff
@@ -0,0 +1,90 @@
...
+	//check that owners have read and write permissions
+	if !hasSomeRead || !hasSomeWrite {
+		errors = append(errors, fmt.Sprintf(FilePermissionsMissingWarning+": perms are %s", mode.String()))
+	}
+
+	if mode&os.ModeSymlink != 0 {
+		errors = append(errors, FileIsSymlinkWarning)
+	}
+
+	if hasOnlyOwnerRW {
```
sgmiller, line 86: Group r/w is probably also okay.  Logic should basically be "owner or group has rw, other has nothing".

vault/diagnose/file_checks_windows.go
```diff
@@ -0,0 +1,10 @@
+// +build windows
+
+package diagnose
+
+import "io/fs"
+
+// IsOwnedByRoot does nothing on windows
+func IsOwnedByRoot(info fs.FileInfo) bool {
```
sgmiller, line 8: Lets add a TODO here that there may be some similar check on Windows.  But we can worry about it later.

sgmiller (changes requested):
Just a small comment, but raft stuff looks good!

vault/diagnose/file_checks.go
HridoyRoy: Yep, I don't throw an error in the case where we have group rw. `hasOnlyOwnerRW` is returned so that the caller can check if `hasOnlyOwnerRW` is true and if the folder is owned by root. Check out https://github.com/hashicorp/vault/pull/11771/files/dc731cc748b1ce842066e485484c1b09b1e31bad#diff-8f7d332a7fe17314b0213fc3bcc6163b8c29c5764e4e8b26c941a9c51f37f8c2R35 .

vault/diagnose/file_checks_windows.go
HridoyRoy: Sounds good!

sgmiller (approved):

PR #11774, opened 2021-06-04
vault 1979: bug fixes and improvements from ent namespace work

swayne275:
this is basically a port from some of the namespace testing i'm doing on the ent side

rough change summary
- i wasn't nil checking in a helper function (which can be convenient)
- mounting backends for various namespaces requires the namespace itself, so i defined a struct so it can work with and without custom namespaces
- re-organized some helpers now that i have a test utilities file


sgmiller (approved):
Afaict looks good.

Issue #11349, opened 2021-04-13
error_on_missing_key no longer working in vault 1.7


adamstrawson:
**Describe the bug**
Upgrading from Vault 1.4 to 1.7, the behaviour of the template option `error_on_missing_key` no longer works as expected.

Previously, `error_on_missing_key = false` would result in a `WARN` for a missing key, however on Vault 1.7, it is returned as an `ERROR`

vault 1.4
```
2021/04/13 17:09:35.001560 [WARN] (view) vault.read(secret/foobar/service.v1): no secret exists at secret/foobar/service (retry attempt 11 after "1m0s")
```

vault 1.7
```
2021/04/13 17:09:35.001560 [ERROR] (view) vault.read(secret/foobar/service.v1): no secret exists at secret/foobar/service (retry attempt 11 after "1m0s")
```

**To Reproduce**

Using the below config, which remains the same between Vault 1.4 and 1.7 during the upgrade.

```hcl
exit_after_auth = true
pid_file = "/home/vault/pidfile"

auto_auth {
    method "kubernetes" {
        mount_path = "auth/kubernetes-foobar"
        config = {
            role = "vault-auth-foobar"
        }
    }

    sink "file" {
        config = {
            path = "/home/vault/.vault-token"
        }
    }
}

template {
  source               = "/home/vault/template/vault-secrets.json"
  destination          = "/etc/secrets/vault-secrets.json"
  error_on_missing_key = false
  left_delimiter       = "<<"
  right_delimiter      = ">>"
}
```

```hcl
vault-config-json.tmpl: <<- with secret "secret/foobar/service?version=1">> <<- end >>
```

**Expected behavior**
The secret key `secret/foobar/service` doesn't exist, and since `error_on_missing_key` is `false` a `WARN` should be returned.

**Environment:**
* Vault Server Version (retrieve with `vault status`): `1.7.0`
* Vault CLI Version (retrieve with `vault version`): `Vault v1.7.0 (4e222b85c40a810b74400ee3c54449479e32bb9f)`
* Server Operating System/Architecture: EKS v1.16.15, using the vault helm chart v0.10.0

adamstrawson:
Hey! Hate to be the guy to nudge on an issue, but would be great to get some eyes/feedback on this? It's creating quite a few headaches/noise for us and ideally don't want to have to rollback from `1.7.0`

kalafut:
cc: @calvn as this may be related to the retry changes

kalafut:
[...]

PR #11775, opened 2021-06-04
agent: restart template runner on retry for unlimited retries

calvn:
This PR takes an alternate approach compared to the one from #11711 on re-introducing unlimited template retry behavior that existed on Vault 1.5.x and earlier. A separate flag controls whether agent restarts the templating server on error, which is  [how it used to work](https://github.com/hashicorp/vault/blob/release/1.5.x/command/agent/template/template.go#L154-L162) up until 1.5.x.

This approach maintains backward compatibility on how retry works with Agent 1.6.x and onward while optionally allowing users to revert to the old (1.5.x and earlier) behavior if desired. ~~I've chosen the value to reside at `vault.retry.template_unlimited_retries = <bool>` since the `template` stanza is an actual CT config and scoped on a per-template basis, but I'm also open to suggestions where else it could go or what else it could be named.~~

Update: See https://github.com/hashicorp/vault/pull/11775#issuecomment-858190780 for the latest update on the config value and format.

Supersedes https://github.com/hashicorp/vault/pull/11711

TODO:
- [x] Update/add tests
- [x] Update docs
- [x] Include changelog


ncabatoff:
I think I'm okay with the idea of having a new key governing the unlimited-retry behaviour.  I'm not sure I like where it is or the fact that the default would be limited-retries - though I get that agent has had limited retries since 1.5 and that going back and forth isn't great.  Since traditionally Agent was meant to be a long-running process, exiting due to some errors talking to Vault is unusual/surprising behaviour, which is why I'm advocating for making indefinite retries the default.  

The `vault` section is supposed to be about Agent's interface to Vault, so it doesn't seem ideal to have a key in there that's template-specific.  I see the new key being added as being similar to the top-level `exit_after_auth` key: something that governs the agent process's lifecycle.

Proposal 1:  add a `template_default` section.   The new key is similar/related to the template-specific `error_on_missing_key` at least the way that consul_template uses it, i.e. to exit on a missing key rather than log a message like Agent apparently does? (https://github.com/hashicorp/vault/issues/11349).  

This new `template_default` section would allow users to reduce duplication by specifying values that should apply to all `template` sections unless overridden.  There would be two new keys in addition to those that are already defined:

- exit_on_missing_key: this would replace `error_on_missing_key`, and when set true would make Agent exit immediately if a key is missing from a successful response
- exit_on_retry_failure: this would be the inverse of the `template_unlimited_retries` key in this PR

Proposal 2:
An alternate approach would be add a single top-level key, e.g. `template_server_lifecycle` or a key within a new section, e.g. `template_server.lifecycle`, which would have values like:
- long-lived: run indefinitely, log errors
- one-shot: exit after rendering all templates successfully
- until-failure: exit if a template isn't rendering successfully after exhausting configured retries

calvn:
The reason that `template_unlimited_retries` is  within `vault.retry` is because we intend to push out this change on a patch release of Vault and so it'd be ideal to avoid introducing a new stanza and move things around or doing things like reverting the default behavior if possible. I'd be a proponent of making the changes from proposal 1 as a separate task that's slated for a major Vault release.

I'm open to moving `template_unlimited_retries` to be a top-level parameter similar to how `error_on_missing_key` is, but I'm a bit hesitant to flip the default to be unlimited retries. This re-introduces issues such as containers/pods being stuck until its eviction timeout because agent doesn't signal that templates can't be rendered (see VAULT-134).
 
Thanks for the review and feedback!

calvn:
@ncabatoff chatted with Product and it seems that we're okay with releasing this for the next major release. I'm open to the idea of having a new `template_default` stanza in this case.

I'm still unsure if we should change the default back to be unlimited though. Thoughts on this @jasonodonnell and @tvoran, especially around the k8s use case?

calvn:
Providing a public update on where we stand on the default behavior after a few rounds of internal discussions:

We're going with Nick's proposal 1, with some slight modifications. We will be adding a new stanza `template_config` for template-related behaviors. For the moment there's  going to be just one new parameter introduced - `exit_on_retry_failure` which controls whether the template runner should exit on failure. In the future, we may use this stanza to tune other template-related behaviors that doesn't quite fit in Agent's existing `template` stanza, since these are declared on a per-template basis.

```hcl
template_config {
  exit_on_retry_failure = <bool>
}
```

As for the omission of `exit_on_retry_failure`, there's already an existing parameter within `template.error_on_missing_key` which conjunction with  `exit_on_retry_failure=false` should restore the old behavior of not exiting Agent if there are errors including, but not limited to, missing key. Past behavior which used to restart the template runner once its internal retry were exhausted will be restored in this PR. This should resolve https://github.com/hashicorp/vault/issues/11349.

It's worth noting that the default value for `exit_on_retry_failure` will be `false`, which means that we will be re-introducing the behavior of unlimited retries that existed on Vault Agent 1.5.x and earlier.

Our rationale for reverting back to unlimited template retries by default is that there's a valid use case brought to us by our users and customers for Agent to continue running when other subsystems such as caching are defined, regardless of whether the templates rendered successfully or not. We also understand that there are other use cases that want the exact opposite. For instance, in the kubernetes use case having unlimited retries can lead to containers and pods being forever stuck during deployments. However, this time around the unlimited template retries behavior can be turned off since this is not always desirable. Separately, our vault-helm chart will make this behavior configurable.

website/content/docs/agent/template-config.mdx
```diff
@@ -0,0 +1,48 @@
...
+Note that a missing key from a secret's response is different from a missing or
+non-existent secret. The former's behavior can be toggled through
+`error_on_missing_key`, while the latter always results in an error returned by
+the templating engine.
+
+## Configuration
+
+The top level `template_config` block has the following configuration entries:
+
+- `exit_on_retry_failure` `(bool: false)` - This option tells Vault Agent to create
```
jasonodonnell, line 47:
```suggestion
- `exit_on_retry_failure` `(bool: false)` - This option configures Vault Agent to create
```

website/content/docs/agent/template-config.mdx
```diff
@@ -0,0 +1,48 @@
...
+---
+
+# Vault Agent Template Config
+
+Template Config allows users of Vault Agent to determine the behavior that
+pertains to the templating engine.
+
+## Functionality
+
+The `template_config` stanza. Note that `template_config` can only be defined
```
jasonodonnell, line 15: > The `template_config` stanza.

Feels like this is missing some words.

website/content/docs/agent/template-config.mdx
```diff
@@ -0,0 +1,48 @@
...
+### Interaction with `error_on_missing_key`
+
+The parameter `error_on_missing_key` can be specified within the `template`
+stanza which determines if a template should ignore rendering a missing key
+within the response of a secret or error out. When `error_on_missing_key` is
+not specified or set to `false` and the key to render is not in the secret's
+response, the templating engine will ignore it (or render `"<no value>"`)
+and continue on with its rendering.
+
+If the desired is to have Agent fail and exit on a missing key, both
```
jasonodonnell, line 32:
```suggestion
If the desire is to have Agent fail and exit on a missing key, both
```

website/content/docs/agent/template-config.mdx
calvn: Whoops, will get this updated!

command/agent/config/test-fixtures/config-template_config-empty.hcl
```diff
@@ -0,0 +1,13 @@
+vault {
+  address = "http://127.0.0.1:1111"
+  retry {
+    num_retries = 5
+  }
+}
+
+template_config {}
```
jasonodonnell, line 8: Is an empty `template_config` now required or can this be omitted if templating is not being configured?

calvn: It can be omitted entirely.

website/content/docs/agent/template-config.mdx
```diff
@@ -0,0 +1,48 @@
...
+layout: docs
+page_title: Vault Agent Template Config
+description: |-
+  Vault Agent's Template Config to set Templating Engine behavior
+---
+
+# Vault Agent Template Config
+
+Template Config allows users of Vault Agent to determine the behavior that
+pertains to the templating engine.
```
tomhjp, line 11:
```suggestion
Template Config configures default behavior for [templates](/docs/agent/template).
```

tomhjp: Perhaps also worth adding a note and the reverse link from the template page?

command/agent_test.go
```diff
@@ -1691,3 +1695,304 @@ vault {
...
+
+	for tcName, tc := range testCases {
+		t.Run(tcName, func(t *testing.T) {
+			// create temp dir for this test run
+			tmpDir, err := ioutil.TempDir(tmpDirRoot, tcName)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			listenAddr := "127.0.0.1:18123"
```
tomhjp, line 1863: Fixed ports make tests sensitive to failures due to port clashes (even if with itself on a VM where two test suites are run in parallel for whatever reason). This case is a tricky one to do reliably, given we can't ask the kernel to reserve a random port for us in the way we normally would in tests, because the port is for use in a separate process. Could we reduce the chance of failure substantially by generating a random high port each time though?

website/content/docs/agent/template-config.mdx
```diff
@@ -0,0 +1,48 @@
...
+once, and is different from the `template` stanza. Unlike `template` which
+focuses on where and how a specific secret is rendered, `template_config`
+contains parameters affects how the templating engine as a whole behaves and its
+interaction with the rest of Agent. This includes, but is not limited to,
+program exit behavior. Other parameters that applies to the templating engine as
+a whole may be added over time.
+
+### Interaction with `error_on_missing_key`
+
+The parameter `error_on_missing_key` can be specified within the `template`
```
tomhjp, line 25:
```suggestion
The parameter [`error_on_missing_key`](/docs/agent/template#error_on_missing_key) can be specified within the `template`
```

website/content/docs/agent/template-config.mdx
```diff
@@ -0,0 +1,48 @@
...
+stanza which determines if a template should ignore rendering a missing key
+within the response of a secret or error out. When `error_on_missing_key` is
```
tomhjp, line 27:
```suggestion
stanza which determines if a template should error when a key is missing
in the secret. When `error_on_missing_key` is
```

nit: Just trying to make this a little more succinct.

website/content/docs/agent/template-config.mdx
```diff
@@ -0,0 +1,48 @@
...
+If the desire is to have Agent fail and exit on a missing key, both
+`template.error_on_missing_key` and `template_config.exit_on_retry_failure` must
+be set to true. Otherwise, the templating engine will error and render to its
+destination, but agent will not exit and retry until it's able to do so or until
+the process is terminated.
+
+Note that a missing key from a secret's response is different from a missing or
+non-existent secret. The former's behavior can be toggled through
+`error_on_missing_key`, while the latter always results in an error returned by
+the templating engine.
```
tomhjp, line 41:
```suggestion
non-existent secret. The templating engine will always error if a secret is missing, but will only error for a missing key if `error_on_missing_key` is set. Whether Vault Agent will exit when the templating engine errors depends on the value of `exit_on_retry_failure`.
```

I had a quick go at reworking this because I found it a bit confusing. It sounded a bit like "always results in an error" means `exit_on_retry_failure` might be ignored in some cases.

website/content/docs/agent/template-config.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+## Configuration
+
+The top level `template_config` block has the following configuration entries:
+
+- `exit_on_retry_failure` `(bool: false)` - This option configures Vault Agent to create
+  the parent directories of the destination path if they do not exist.
```
tomhjp, line 48: Looks like a copy-paste block that needs updating.

tomhjp (commented):
Sorry for the delay reviewing this. I think the top-level config and behaviour changes are good, mostly just some suggestions on the docs 👍

command/agent_test.go
calvn: This bit of the code is borrowed from `TestAgent_Cache_Retry`, but as long as we're not parallelizing this test with any other tests (by explicitly calling `t.Parallel()`) it should be fine.

website/content/docs/agent/template-config.mdx
calvn: I referenced "templating engine" on purpose instead of `template` because the params within the `template` stanza defines the rendering behavior for templates themselves so I didn't want to confuse the two. The current wording might not be the best though, so still open to suggestions!

website/content/docs/agent/template-config.mdx
calvn: Yeah this note was meant purely for `error_on_missing_key` as there was some misunderstanding of how it worked for me in the past. I like how you tie it back to `exit_on_retry_failure` though, so I'll go ahead and accept your suggestion!

command/agent/template/template.go
```diff
@@ -253,6 +265,20 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
...
+		//
+		// Note: This is a fixed value (12) that ends up being a multiplier to
+		// retry.num_retires (i.e. 12 * N total retries per runner restart).
+		// Since we are performing retries indefinitely this base number helps
+		// prevent agent from spamming Vault if retry.num_retries is set to a
+		// low value by forcing exponential backoff to be high towards the end
+		// of retries during the process.
+		if sc.AgentConfig.TemplateConfig != nil && !sc.AgentConfig.TemplateConfig.ExitOnRetryFailure {
+			attempts = ctconfig.DefaultRetryAttempts
+		}
```
calvn, line 280: This logic over here is what allows agent to perform a reasonable number of retries with enough backoff pause towards the tail end of the run before restarting. I've chosen `ctconfig.DefaultRetryAttempts` as the value since that was always the case on Agent 1.5.x and earlier. We can't completely hand off retries to the cache's proxy because the API client only performs retries on 5xx errors whereas the builtin retry logic from consul-template performs retries on any errors. A non-existent secret (i.e. a 404) without retry attempts from consul-template itself will spam apiproxy, and in turn the server, with requests indefinitely which can easily DDoS Vault.

The alternative here could be to set `attempts = sc.AgentConfig.Vault.Retry.NumRetries` but it has the danger of spamming the server with requests if the value is too low or zero due to low initial backoff values.

jasonodonnell (approved):
LGTM and tested fine in K8s with persistent caching.

website/content/docs/agent/template-config.mdx
kalafut: I'm reading the suggesting as adding something to `/docs/agent/template` referencing _this_ page. I can imagine being very confused when someone is supposed to "set xyz in the template config", and they're on the template docs page looking and a big section called "Configuration" and not finding it. It is still intrinsically confusing, IMO, but as least some extra guidance might route them to the right page.

kalafut:
I think we need to update https://www.vaultproject.io/docs/agent#retry-stanza, especially around:

> The vault stanza may contain a retry stanza that controls how failing Vault requests are handled, whether these requests are issued in order to render templates...

website/content/docs/agent/template-config.mdx
calvn: My reply was to Tom's commit suggestion and not on the reverse link. I thin it makes sense to add something to `/docs/agent/template` that references this page though so I'll get that updated!

calvn: Ended up backlinking both ways. I realized how confusing it can be otherwise. Thanks for the suggestion Tom and Jim!

calvn:
@kalafut ended up appending rather than modifying the retry statement under `/docs/agent`. Let me know what you think.

website/content/docs/agent/index.mdx
```diff
@@ -94,12 +96,17 @@ configuration entries:
...
 
-The `vault` stanza may contain a `retry` stanza that controls how failing
-Vault requests are handled, whether these requests are issued in order to render
+The `vault` stanza may contain a `retry` stanza that controls how failing Vault
+requests are handled, whether these requests are issued in order to render
 templates, or are proxied requests coming from the proxy cache subsystem.
 Auto-auth, however, has its own notion of retrying and is not affected by this
 section.
 
+For requests from the templating engine, Agent will restart its internal runner
```
kalafut, line 105: This gets into details (e.g. "its internal runner") that the user should not need to know about. I also don't think I confidently understand what's going to happen. Does "perform retries again once all retries are exhausted" mean that if I have`num_retries` at 5, I get 5 retries from the `retry` stanza and another 5 from the template rendering?

calvn: It means that the templating engine will restart again with a count=0. It's mostly a backoff reset that doesn't cause agent to exit unless explicitly stated to do so. The scenario you gave is more of chained retries which _can_ happen if templating is going through the cache, but that's different from what's being described here.

The internal runner might be too much detail, though. What's a good way to describe this behavior without having to pull back the curtain too much on how it works under the hood?

calvn: Removed the mentions of the internal runner.

website/content/docs/agent/index.mdx
```diff
@@ -114,7 +121,13 @@ result codes: any 50x code except 501 ("not implemented"), as well as 412
 stale read due to eventual consistency. Requests coming from the template
 subsystem are retried regardless of the failure.
 
-Second, the backoff algorithm used to set the time between retries differs for
+Second, templating retries may be performed by both the templating engine _and_
+the cache proxy if Agent [persistent
+cache][persitent-cache] is enabled. This is due to the
```
kalafut, line 126:
```suggestion
cache][persistent-cache] is enabled. This is due to the
```

website/content/docs/agent/index.mdx
```diff
@@ -198,7 +211,9 @@ template {
 [vault]: /docs/agent#vault-stanza
 [autoauth]: /docs/agent/autoauth
 [caching]: /docs/agent/caching
+[persitent-cache]: /docs/agent/caching/persistent-caches
```
kalafut, line 214:
```suggestion
[persistent-cache]: /docs/agent/caching/persistent-caches
```

tvoran (commented):
Just some small suggestions.

command/agent_test.go
```diff
@@ -1691,3 +1695,304 @@ vault {
...
+			templateErrorOnMissingKey: true,
+			expectExit:                true,
+			expectError:               true,
+		},
+		"false, no template error": {
+			exitOnRetryFailure:        pointerutil.BoolPtr(false),
+			templateContents:          templateContents(0),
+			expectTemplateRender:      templateRendered(0),
+			templateErrorOnMissingKey: false,
+			expectExit:                false,
```
briankassouf, line 1818: Why would we expect to not exit in this case if there isn't an error?

command/agent_test.go
```diff
@@ -1691,3 +1695,304 @@ vault {
...
+			expectError:               true,
+		},
+		"false, with missing key": {
+			exitOnRetryFailure:        pointerutil.BoolPtr(false),
+			templateContents:          missingKeyTemplateContent,
+			expectTemplateRender:      missingKeyTemplateRender,
+			templateErrorOnMissingKey: false,
+			expectExit:                false,
```
briankassouf, line 1834: Same here, this seems like a behavior change on non-error cases

command/agent_test.go
```diff
@@ -1691,3 +1695,304 @@ vault {
...
+			err = verify()
+			close(cmd.ShutdownCh)
+			wg.Wait()
+
+			switch {
+			case (code != 0 || err != nil) && tc.expectError:
+				if exited != tc.expectExit {
+					t.Fatalf("expected program exit to be '%t', got '%t'", tc.expectExit, exited)
+				}
+			case code == 0 && err == nil && !tc.expectError:
```
briankassouf, line 1988: Should we still test if we exited in this case?

command/agent_test.go
calvn: `expectExit` is a bit misnamed. It's really more like `expectExitByErrors` since it's only checked if the exit code is non-zero.

command/agent_test.go
calvn: This is the happy path. The exit code was 0 and we didn't expect an error.

command/agent_test.go
calvn: Refer to the comment above.

command/agent_test.go
calvn: Renamed it to `expectExitFromError` to make it more explicit on how this should be used.

briankassouf: I see, so it has no meaning if `expectError == false`? I think i was just confused that it was a different value from the same template values above.

calvn: Yup, that's correct. I realized that two cases were flipped even if they were not applicable, so I've gone ahead and fixed that in b3dea46.

command/agent_test.go
calvn: I added an additional check within the happy path to make sure that we are not exiting early for any reason.

website/content/docs/agent/template-config.mdx
```diff
@@ -0,0 +1,55 @@
...
+layout: docs
+page_title: Vault Agent Template Config
+description: |-
+  Vault Agent's Template Config to set Templating Engine behavior
+---
+
+# Vault Agent Template Config
+
+Template Config allows users to determine the behavior that
+pertains to the templating engine within Vault Agent.
```
tomhjp, line 11: > I referenced "templating engine" on purpose instead of template because the params within the template stanza defines the rendering behavior for templates themselves so I didn't want to confuse the two.

Oops, yeah I definitely missed that distinction! It's a tricky one to communicate well, because it's kind of spilling over into explaining the internals of Vault Agent rather than just explaining the behavior. WDYT of this suggestion? Possibly too simplistic, but I think it's complemented very well by the paragraph after it.

```suggestion
Template Config configures Vault Agent behavior common to all `template` stanzas.
```

tomhjp: I've just noticed the "Functionality" section below as well, which goes into slightly more detail on the same subject. Perhaps worth merging these two sections together?

briankassouf (approved):

website/content/docs/agent/template-config.mdx
calvn: This initial section is meant to be a 1-2 sentence general description of the feature. The functionality section goes a bit deeper into the details, and also has "Interactions with ..." sub-section.

PR #11779, opened 2021-06-07
errwrap.Wrapf() mop-up

alrs:
This picks off a couple of spots where `errwrap.Wrapf()` got reintroduced as I was finishing its deprecation.

### DIFF
--- a/serviceregistration/consul/consul_service_registration.go
+++ b/serviceregistration/consul/consul_service_registration.go
@@ -15,7 +15,6 @@ import (
 	"time"
 
 	"github.com/hashicorp/consul/api"
-	"github.com/hashicorp/errwrap"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault/sdk/helper/consts"
 	"github.com/hashicorp/vault/sdk/helper/parseutil"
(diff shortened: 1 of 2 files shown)


vishalnayak:
@alrs Can you provide more context into why this change is needed, and which deprecation you are referring to?

`errwrap.Wrapf` is used in a ton of places in Vault and it is a recommended way to handle errors that are embedded in another. This way the original error will still be preserved.

alrs:
> @alrs Can you provide more context into why this change is needed, and which deprecation you are referring to?

Hi, @vishalnayak. From the godoc of `errwrap.Wrapf`: `Deprecated: Use fmt.Errorf()`.

`fmt.Errorf()` has been able to wrap errors itself with `%w` for some versions now, and every version supported by Vault is included in that list.

Should you want to grep, you'll find that `errwrap.Wrapf()` has been removed everywhere in the project, save for these two.

vishalnayak:
> > @alrs Can you provide more context into why this change is needed, and which deprecation you are referring to?
> 
> Hi, @vishalnayak. From the godoc of `errwrap.Wrapf`: `Deprecated: Use fmt.Errorf()`.
> 
> `fmt.Errorf()` has been able to wrap errors itself with `%w` for some versions now, and every version supported by Vault is included in that list.
> 
> Should you want to grep, you'll find that `errwrap.Wrapf()` has been removed everywhere in the project, save for these two.

@alrs Yep! Got to know. I was typing up a response here, but you beat me to it. Looks good!

vishalnayak (approved):

PR #11785, opened 2021-06-07
Implement ember-cp-validations on KV secret engine

Monkeychip:
One PR of many to come where we'll be adding form validations to various forms in the application.  This first PR tackles the KV secret engine.  

In an earlier demo, I used the ember-changeset-validations library.  In this, I'm using the ember-cp-validations library and I much prefer the latter.  It does not modify the whole model object but instead adds to it using a mixin.  It's much easier to use, and more importantly, it should be usable in new glimmer components.

**Notes:**
- the kv secret engine model is unique in that it gets path and secret-key from a mixin, therefore the validation is custom for these two and does not use the new cp-validations library.  There was some previous custom validation that I modified to make more consistent with this new pattern/design.
- I noticed the `max_versions` is not saving on create.  This will be addressed in another ticket. 

**Path**
path

**Create and Edit Secret**
validation

### DIFF
--- a/ui/app/components/secret-edit.js
+++ b/ui/app/components/secret-edit.js
@@ -57,8 +57,10 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
   hasLintError: false,
   isV2: false,
 
-  validationError: null,
-  validationMessage: '',
+  validationMessagePath: '',
+  validationMessageKey: '',
+  validationMessageMaxVersions: '',
+  validationErrorCount: 0,
 
   init() {
     this._super(...arguments);
@@ -81,31 +83,19 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
     if (this.mode === 'edit') {
       this.send('addRow');
     }
-    let errorObject = {
-      key: '',
-      path: '',
-      maxVersions: '',
-    };
-    this.set('validationError', errorObject);
   },
 
   waitForKeyUp: task(function*(name, value) {
     // path and key are not on the model (there via key-mixin) doing custom validations here instead of cp validations
     if (name === 'path') {
       // no value indicates missing presence
-      if (!value) {
-        this.set('validationMessagePath', `Secret path can't be blank`);
-      } else {
-        this.set('validationMessagePath', '');
-      }
+      !value
+        ? this.set('validationMessagePath', `Secret path can't be blank`)
+        : this.set('validationMessagePath', '');
     }
     if (name === 'key') {
       // no value indicates missing presence
-      if (!value) {
-        this.set('validationMessageKey', `Key can't be blank`);
-      } else {
-        this.set('validationMessageKey', '');
-      }
+      !value ? this.set('validationMessageKey', `Key can't be blank`) : this.set('validationMessageKey', '');
     }
     if (name === 'maxVersions') {
       // checking for value because value is blank on first loading, no keyup event has occurred and default is 10.
@@ -120,7 +110,12 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
         this.set('validationMessageMaxVersions', '');
       }
     }
-
+    this.set(
+      'validationErrorCount',
+      (this.validationMessagePath === '' ? 0 : 1) +
+        (this.validationMessageKey === '' ? 0 : 1) +
+        (this.validationMessageMaxVersions === '' ? 0 : 1)
+    );
     while (true) {
       let event = yield waitForEvent(document.body, 'keyup');
       this.onEscape(event);
--- a/ui/app/components/secret-edit.js
+++ b/ui/app/components/secret-edit.js
@@ -57,6 +57,8 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
   hasLintError: false,
   isV2: false,
 
+  validationError: null,
+
   init() {
     this._super(...arguments);
     let secrets = this.model.secretData;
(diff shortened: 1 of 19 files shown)


ui/app/templates/components/secret-edit-display.hbs
```diff
@@ -1,92 +1,115 @@
...
-        @class="is-marginless"
-        data-test-v2-write-without-read
-      />
-    {{else}}
-      <AlertBanner
-        @type="warning"
-        @message="Your policies prevent you from reading the current secret data. Saving using this form will overwrite the existing values."
-        @class="is-marginless"
-        data-test-v1-write-without-read
+{{#if (and (or @model.isNew @canEditV2Secret) @isV2 (not @model.failedServerRead))}}
```
Monkeychip, line 1: fixing spacing here, hence the big diff.

ui/app/styles/components/form-field.scss
```diff
@@ -0,0 +1,3 @@
+.has-error-border {
```
hashishaw, line 1: I think we have a forms.scss this might be better suited to go in, rather than a new page.

ui/app/components/secret-edit.js
```diff
@@ -81,7 +87,36 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
     }
   },
 
-  waitForKeyUp: task(function*() {
+  waitForKeyUp: task(function*(name, value) {
+    // because path and key are not on the model performing custom validations instead of cp-validations
+    if (name === 'path') {
+      // no value indicates missing presence
+      !value
+        ? this.set('validationMessagePath', `Secret path can't be blank`)
```
hashishaw, line 95: Rather than passing in each of these as separate values, what if we had a `validationMessages` object and then you could do `this.set('validationMessages.path', 'error message')`? Then we could access the object in the template

ui/lib/core/addon/components/alert-inline.js
```diff
@@ -16,7 +16,7 @@ import layout from '../templates/components/alert-inline';
  * @param [message=null]{String} - The message to display within the alert.
  * @param [sizeSmall=false]{Boolean} - Whether or not to display a small font with padding below of alert message.
  * @param [paddingTop=false]{Boolean} - Whether or not to add padding above component.
- *
+ * @param [noMarginBottom=false]{Boolean} - Whether or not to remove margin bottom below component.
```
hashishaw, line 19: I think our previous pattern for this is something like `isMarginless` to avoid a double negative (no margin = false)

ui/lib/core/addon/components/form-field.js
```diff
@@ -18,11 +18,13 @@ import layout from '../templates/components/form-field';
  * ```
  *
  * @param [onChange=null] {Func} - Called whenever a value on the model changes via the component.
+ * @param [onKeyUp=null] {Func} - Called whenever cp-validations is being used and you need to validation on keyup.  Send name of field and value of input.
  * @param attr=null {Object} - This is usually derived from ember model `attributes` lookup, and all members of `attr.options` are optional.
  * @param model=null {DS.Model} - The Ember Data model that `attr` is defined on
  * @param [disabled=false] {Boolean} - whether the field is disabled
  * @param [showHelpText=true] {Boolean} - whether to show the tooltip with help text from OpenAPI
  * @param [subText] {String} - Text to be displayed below the label
+ * @param [validationMessage] {String} - AlertInline error message to be displayed if cp-validation error.
```
hashishaw, line 27: 💯

ui/lib/core/addon/templates/components/form-field.hbs
```diff
@@ -248,22 +248,40 @@
...
         {{#if
           (and
             (get model valuePath) (not (get model attr.options.validationAttr))
           )
         }}
           <AlertInline @type="danger" @message={{attr.options.invalidMessage}} />
-
         {{/if}}
       {{/if}}
+      {{!-- Cp Validations Library --}}
```
hashishaw, line 277: This isn't specific to the validations library, since anytime you pass in `validation message` this will render

ui/tests/acceptance/secrets/backend/kv/secret-test.js
```diff
@@ -56,6 +64,22 @@ module('Acceptance | secrets/secret/create', function(hooks) {
     assert.ok(showPage.editIsPresent, 'shows the edit button');
   });
 
+  test('it disables save when validation errors occur', async function(assert) {
```
hashishaw, line 67: 👏

hashishaw (changes requested):

ui/app/components/secret-edit.js
```diff
@@ -81,7 +87,36 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
     }
   },
 
-  waitForKeyUp: task(function*() {
+  waitForKeyUp: task(function*(name, value) {
+    // because path and key are not on the model performing custom validations instead of cp-validations
+    if (name === 'path') {
```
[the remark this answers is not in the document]
Monkeychip: Good idea.

ui/app/models/secret-engine.js
```diff
@@ -2,12 +2,20 @@ import Model, { attr } from '@ember-data/model';
...
+import { validator, buildValidations } from 'ember-cp-validations';
 
 //identity will be managed separately and the inclusion
 //of the system backend is an implementation detail
 const LIST_EXCLUDED_BACKENDS = ['system', 'identity'];
 
-export default Model.extend({
+const Validations = buildValidations({
+  path: validator('presence', {
+    presence: true,
```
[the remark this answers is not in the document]
Monkeychip: It's just the way they're written.  See the example from documentation [here](http://offirgolan.github.io/ember-cp-validations/docs/classes/Presence.html).

ui/app/styles/components/form-field.scss
Monkeychip: Smart! WIll do.

ui/app/components/secret-edit.js
Monkeychip: Good call.  It was a bit of working around, and learned a good amount about the get helper in handlebars, but I got it.

hashishaw (approved):
Great work!

PR #11787, opened 2021-06-07
Removes space on overview page with empty state

hellobontempo:
Previously, during empty state a white space appeared below tabs:

image-20210226-155821

Added empty toolbar to conditionally render when state is empty to match the designs:

Screen Shot 2021-06-07 at 3 24 06 PM

When a database is connected, the empty toolbar does not display:

Screen Shot 2021-06-07 at 3 32 08 PM


hashishaw (approved):
🚀

PR #11789, opened 2021-06-08
Vault-1983: Use fairsharing to distribute workers between queues

swayne275:
This PR upgrades the fairsharing library to assign workers to queues based on the fairsharing algorithm detailed in VLT-145.

I've left a couple comments on remaining questions/things to discuss.


helper/fairshare/jobmanager.go
```diff
@@ -179,10 +177,92 @@ func (j *JobManager) getNextJob() Job {
...
+func (j *JobManager) sortByNumWorkers() []string {
+	out := make([]string, 0, len(j.queues))
+	for queueID := range j.queues {
+		out = append(out, queueID)
+	}
+
+	workersPerQueue := j.workerCount
+
+	sort.Slice(out, func(i, j int) bool {
+		// TODO do we want this explicity, or do we want some randomness?
```
swayne275, line 233: i'd love some thoughts on this

helper/fairshare/jobmanager.go
```diff
@@ -179,10 +177,94 @@ func (j *JobManager) getNextJob() Job {
...
+func (j *JobManager) sortByNumWorkers() []string {
+	out := make([]string, 0, len(j.queues))
+	for queueID := range j.queues {
+		out = append(out, queueID)
+	}
+
+	workersPerQueue := j.workerCount
+
+	sort.Slice(out, func(i, j int) bool {
+		// TODO should we be explicitly breaking times with the queueID, or
```
swayne275, line 234: i'd love some thoughts on this

helper/fairshare/jobmanager.go
```diff
@@ -203,57 +285,48 @@ func (j *JobManager) assignWork() {
...
-			// listen for a wake-up when an emtpy job manager has been given
-			// new work
 			select {
 			case <-j.quit:
 				j.wg.Done()
 				return
 			case <-j.newWork:
-				break
+				// listen for wake-up when an empty job manager has been given work
+			case <-time.After(50 * time.Millisecond):
```
swayne275, line 308: happy to discuss this bit

helper/fairshare/jobmanager.go
HridoyRoy: I think you're right, randomness here wouldn't necessarily give us anything as it wouldn't decrease the stress on any of the workers (since this is just a tiebreaker).

helper/fairshare/jobmanager.go
```diff
@@ -179,10 +177,94 @@ func (j *JobManager) getNextJob() Job {
...
+	}
+
+	workersPerQueue := j.workerCount
+
+	sort.Slice(out, func(i, j int) bool {
+		// TODO should we be explicitly breaking times with the queueID, or
+		// might we want some randomness?
+		// I think it's probably fine since it only breaks ties between queues
+		// with equal workers assigned, and makes testing much easier
+		if workersPerQueue[out[i]] == workersPerQueue[out[j]] {
```
HridoyRoy, line 238: I don't think you need this check. If you want the sorting to be stable, you could call `SliceStable` , which I think should allow you to remove this code. But, by virtue of this being a tiebreaker, I don't think we need stability.

helper/fairshare/jobmanager.go
```diff
@@ -179,10 +177,94 @@ func (j *JobManager) getNextJob() Job {
...
+	j.workerCount[queueID]++
+}
+
+// decrement the worker count for this queue
+// this also removes worker tracking for this queue if needed
+func (j *JobManager) decrementWorkerCount(queueID string) {
+	j.l.Lock()
+	defer j.l.Unlock()
+
+	j.workerCount[queueID]--
```
HridoyRoy, line 262: Can this become negative?

helper/fairshare/jobmanager.go
```diff
@@ -203,57 +285,48 @@ func (j *JobManager) assignWork() {
...
-	if j.lastQueueAccessed > 0 {
-		j.lastQueueAccessed--
-	} else {
-		j.lastQueueAccessed = len(j.queuesIndex) - 1
+	// it's possible the queue ran out of work and was pruned, but there were
+	// still workers operating on data formerly in that queue, which were still
+	// being tracked. if that is the case, we don't want to wipe out that worker
+	// count when the queue is re-initialized.
+	if _, ok := j.workerCount[queueID]; !ok {
+		j.workerCount[queueID] = 0
```
HridoyRoy, line 330: I'm not sure I understand this part -- if there are still operating workers in the queue, isn't their count still at 0?

helper/fairshare/jobmanager.go
swayne275: i read the doc for `Slice` and `SliceStable` and didn't see anything about them breaking ties in this way. do you know how they break ties?

Since this data comes from a map (which has no guaranteed ordering) i think we either accept the inherent randomness of that map's order (i believe the iter is random), or we explicitly sort.

My thought process is explicitly sorting is nice for the testing process so we don't have flakiness if the map gets unordered?

helper/fairshare/jobmanager.go
swayne275: if there are 5 workers operating on data from a particular queue, the count of how many workers will be 5 (and that will decrease as workers finish their tasks). the edge case is if there is no upcoming work in that queue, but workers are still operating on stuff that came from that queue

helper/fairshare/jobmanager.go
HridoyRoy: Oh, I see, no that makes perfect sense! I didn't realize that the comment above this code was talking about the inherent randomness in the unstable sorting. Feel free to resolve this convo!

helper/fairshare/jobmanager.go
HridoyRoy: Oh, I see. So is the reason this is being explicitly set to 0 in this case because `DecrementWorkers` starts by decreasing the workerCount, and then checks `j.workerCount[queueID]` ?

swayne275: well, this is more about only re-initializing if we aren't already tracking data there. i think the more important bit is the if statement - setting to 0 just initializes the `queueID` k/v in the map (but we have to be careful to not wipe out data if it's already there)

does that make sense? is there a way i could improve the comment to make it more clear?

helper/fairshare/jobmanager.go
```diff
@@ -179,10 +177,94 @@ func (j *JobManager) getNextJob() Job {
...
+	return nextQueue, canAssignWorker
+}
+
+// returns true if there are already too many workers on this queue
+// note: this must be called with j.l held (at least for read).
+// note: we may want to eventually factor in queue length relative to num queues
+func (j *JobManager) queueWorkersSaturated(queueID string) bool {
+	numActiveQueues := float64(len(j.queues))
+	numTotalWorkers := float64(j.workerPool.numWorkers)
+	maxWorkersPerQueue := math.Ceil(0.9 * numTotalWorkers / numActiveQueues)
```
swayne275, line 216: it might make sense to bump this to 0.95. It's 0.9 in the RFC, mainly to keep workers in reserve for new mounts (especially when the number of mounts is low). But 0.9 might keep too many workers in reserve. Hard to know without empirical data i think

HridoyRoy (approved):

helper/fairshare/jobmanager.go
HridoyRoy: Yeah, that makes sense! I'm not sure about the comment reformatting, but in light of this conversation the comment looks good.

helper/fairshare/jobmanager.go
sgmiller: You can't use randomness purely, as some sorts won't ever complete if the tests aren't consistent.

swayne275: the randomness i was discussing is from the order from the map iter, so whatever is in the sort less function would be consistent

basically, should we leave the tiebreaker in or not?

swayne275, line 234:
```suggestion
		// TODO should we be explicitly breaking ties with the queueID, or
```

helper/fairshare/jobmanager.go
```diff
@@ -179,10 +177,94 @@ func (j *JobManager) getNextJob() Job {
...
+	out := make([]string, 0, len(j.queues))
+	for queueID := range j.queues {
+		out = append(out, queueID)
+	}
+
+	workersPerQueue := j.workerCount
+
+	sort.Slice(out, func(i, j int) bool {
+		// TODO should we be explicitly breaking times with the queueID, or
+		// might we want some randomness?
```
swayne275, line 235:
```suggestion
		// might we want some randomness from the map iter order?
```

helper/fairshare/jobmanager.go
```diff
@@ -179,10 +182,81 @@ func (j *JobManager) getNextJob() Job {
...
+	return (currentIdx + 1) % len(j.queuesIndex)
+}
+
+// returns true if there are already too many workers on this queue
+// note: this must be called with j.l held (at least for read).
+// note: we may want to eventually factor in queue length relative to num queues
+func (j *JobManager) queueWorkersSaturated(queueID string) bool {
+	numActiveQueues := float64(len(j.queues))
+	numTotalWorkers := float64(j.workerPool.numWorkers)
+	maxWorkersPerQueue := math.Ceil(0.9 * numTotalWorkers / numActiveQueues)
```
swayne275, line 233: it might make sense to reduce the 0.9, or ration workers out based on number of available workers (rather than total workers)?

briankassouf (approved):
One small thing, looks good otherwise!

helper/fairshare/jobmanager.go
pmmukh: one doubt here. if the job manager is periodically waking up to check if there is work available, is the newWork channel still required ?

swayne275: new work will save it from having to wait if the queues are running really low and new work is being added about as fast as it can be consumed. really though, i'd love to get rid of the timer channel to begin with. going to give it a bit more thought, and open to ideas if suggested

swayne275: granted, in that case it probably isn't a huge deal, buuuuutttt....

pmmukh: yeah not removing the newWork makes sense, it does shave some ms off a time sensitive process, if i can think of any suggestions to make this more elegant shall let you know! ( i tried thinking about it a little, but all thoughts were grossly complex for the problem at hand )

swayne275: yeah i could probably take some time to re-build the signaling stuff to make it more cohesive, but when i wrote this i didn't see a simple close-to-drop-in solution at the time. i'd really like to get rid of the time bit, unless it makes the code really complex

pmmukh: roger. so this is the first solution off the top of my head, nothing fancy so i bet you might've already considered something similar
- Make the newWork channel be a `processableWork` channel, indicating that work is available and can be processed. This would be for both scenarios when nothing is being processed, or work has been lined up and a queue just got unsaturated
- The place where an empty struct is passed in, in `AddJob`, stays the same
- In `getNextQueue`, if all queues are non-empty and saturated, set a flag value protected with a write lock
- In the cleanup function `decrement..` passed to `dispatch`, we can check that flag, if it is set and the queue just got unsaturated, unset the flag and pass in a struct to the `processableWork` channel
I haven't considered feasibilities like `can everything access the things i've set out there`, `does combo-ing newWork into this make sense` and `will this even work`, so it may not be a great idea to rabbithole this, but bleh i had fun thinking this through lol

helper/fairshare/jobmanager.go
swayne275: could you explain why this suggestion? i think that would do an integer divide before a float multiplication (and order of operations makes the math correct as is)?

see https://play.golang.org/p/48DlknViyjk for an example

swayne275: i'm going to mark this as resolved without changing it, but happy to discuss if there's something i've missed!

PR #11795, opened 2021-06-08
Displays Auth Method description on Vault UI login page

hellobontempo:
Adds description beneath path name
auth_description


ui/app/templates/components/auth-form.hbs
```diff
@@ -26,6 +26,10 @@
       @errorMessage={{if (and cluster.standby hasCSPError) cspErrorText error}}
       data-test-auth-error
     />
+    <div class="has-bottom-margin-s ">
+      <p class="is-label"> {{this.selectedAuthBackend.path}}</p>
```
Monkeychip, line 30: super small comment, but no spaces after the closing element.  In the line below too, I noticed a space before and after the opening and closing span tags.  e.g `>space{{this.selectedAuthBackend....`

Monkeychip (approved):
nice work!!!

Monkeychip:
Also, I forgot you'll need to do a changelog.  If you checkout the vault (e.g. it's a folder higher than the ui folder) files called: 11500.txt. with the following inside: 
```release-note:improvement
ui: Updated ivy code mirror component for consistency
```

The 11500 is the PR number.  In this case your pr number is 11795 so you'll make a .txt file called 11795.txt in the changelog folder with a description of this pr. This will also be an improvement like my example.

ui/app/templates/components/auth-form.hbs
hellobontempo: @Monkeychip - thank you for pointing this out! I wasn't sure which was correct 😅

PR #11796, opened 2021-06-08
InfluxDB - Add username customization

MilenaHC:
Adds the ability to customize username generation for dynamic users in InfluxDB.

### DIFF
--- a/plugins/database/influxdb/influxdb.go
+++ b/plugins/database/influxdb/influxdb.go
@@ -3,13 +3,13 @@ package influxdb
 import (
 	"context"
 	"fmt"
-	"github.com/hashicorp/vault/sdk/helper/template"
 	"strings"
 
 	multierror "github.com/hashicorp/go-multierror"
 	dbplugin "github.com/hashicorp/vault/sdk/database/dbplugin/v5"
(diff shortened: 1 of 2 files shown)


tomhjp (approved):
LGTM! 👍

kalafut:
Nice! Will this be for 1.8? If so, please add the 1.8 Milestone to the PR.

plugins/database/influxdb/influxdb.go
```diff
@@ -3,11 +3,11 @@ package influxdb
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/vault/sdk/helper/template"
```
austingebauer, line 6: You may not have set this up already, but the [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) tool is useful to organize imports in a consistent way. I suggest to try running that tool on this file! I would expect to see this import grouped with the other group (`"github.com/*"`) below.

This is just a stylistic suggestion. We usually try to organize imports consistently in Vault.

austingebauer (approved):
Looks good to me! Great job 👍

calvn (approved):
Same thoughts on imports as Austin mentioned, but otherwise looks great!

PR #11799, opened 2021-06-09
vault: token store dropped test errors

alrs:
This fixes dropped `error` variables in the token store tests.


vishalnayak (approved):
Thank you!

PR #11807, opened 2021-06-09
Updated transit key action to components

arnav28:


ui/app/components/transit-key-action/rewrap.js
```diff
@@ -0,0 +1,14 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class RewrapComponent extends Component {
```
hashishaw, line 4: One thing we may want to explore, to DRY (don't repeat yourself) this up is to make a component js file like `transit-key-action/base.js` with these actions defined, and then the rest of these component files can instead extend that base component. I know we've done similar patterns elsewhere, and I'm not totally sure it will work on glimmer components, but it's worth a try

ui/app/templates/components/transit-key-actions.hbs
```diff
@@ -15,7 +15,34 @@
   <MessageError @errors={{errors}} />
   {{#if selectedAction}}
     <div data-test-transit-action={{selectedAction}}>
-    {{partial (concat 'components/transit-key-action/' selectedAction)}}
+      {{ component (concat 'transit-key-action/' selectedAction)
+          key=key
```
hashishaw, line 19: 😮

hashishaw (commented):
Looks good overall! I added one suggestion, if that doesn't work for some reason I think this looks good to go 👍

ui/app/components/transit-key-action/export.js
```diff
@@ -0,0 +1,5 @@
+import Component from '@glimmer/component';
+export default class ExportComponent extends Component {
+  wrapTTL = null;
```
arnav28, line 3: Do these need tracked attribute?

ui/app/templates/components/transit-key-actions.hbs
arnav28: Yeah too many properties to pass. No wonder they used partials earlier!!
The second option is to add a bunch of if/else-if statements and pass the required properties for each component. But the dynamic component approach seems cleaner.

ui/app/components/transit-key-action/export.js
hashishaw: Yes, i believe so since we're mutating on the template

hashishaw (approved):
Nice work!

PR #11821, opened 2021-06-10
physical/raft: Add a function that gets the offline, stale configuration

briankassouf:


HridoyRoy (approved):
👍 lgtm, going to use and add tests for this in: https://github.com/hashicorp/vault/pull/11771

PR #11826, opened 2021-06-10
Omit wrapping tokens and control groups from client counts

briankassouf:


vault/activity_log_test.go
```diff
@@ -91,6 +91,54 @@ func TestActivityLog_Creation(t *testing.T) {
...
+		Path:         "test",
+		Policies:     []string{controlGroupPolicyName},
+		CreationTime: time.Now().Unix(),
+		TTL:          3600,
+		NamespaceID:  namespace_id,
+	})
+
+	a.fragmentLock.Lock()
+	if a.fragment != nil {
+		t.Fatal("fragment created")
```
swayne275, line 137:
```suggestion
		t.Fatal("fragment not created")
```

swayne275 (approved):
I'm approving because I'll be heading out for a flight soon, but i think that code suggestion i added needs to go in first. other than that, looks good!

vault/activity_log_test.go
swayne275: wait a sec, do we want the fragment to be created here for non-wrapped token? if the `a.fragment != nil` check is correct, ignore my suggestion!

briankassouf: Yeah we do not want it to be created here since this is a control group token

PR #11832, opened 2021-06-11
Makefile: don't go get -u for plugin upgrades

samsalisbury:
- Not needed to upgrade the plugin itself.
- Can cause issues when transitive modules are updated.


tomhjp (approved):
Thanks!

PR #11835, opened 2021-06-11
Reorganize request handling code so that we don't touch storage until we have the stateLock

ncabatoff:


briankassouf (approved):
Looks good, thanks!

PR #11839, opened 2021-06-11
Add more detailed documentation on setting issuer for CSI provider

tomhjp:
This has been a common tripping point for first-time setup of the Vault CSI provider, and isn't well addressed by existing documentation, so this is an attempt to explain the topic and help people find the correct configuration.

Also adds documentation for command line arguments now that the provider has some non-trivial configuration available from 0.3.0+.

### DIFF
--- a/website/content/docs/platform/k8s/csi/index.mdx
+++ b/website/content/docs/platform/k8s/csi/index.mdx
@@ -18,7 +18,7 @@ Driver to be installed.
 At a high level, the CSI Secrets Store driver allows users to create `SecretProviderClass` objects.
 This object defines which secret provider to use and what secrets to retrieve. When pods requesting CSI volumes
 are created, the CSI Secrets Store driver will send the request to the Vault CSI Provider if the provider
-is `vault`. The Vault CSI Provider will then use Secret Class Provider specified and the pod's service account to retrieve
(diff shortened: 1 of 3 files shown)


website/content/docs/platform/k8s/csi/index.mdx
```diff
@@ -48,6 +48,52 @@ A service account must be present to use the Vault CSI Provider with the Kuberne
...
+  kubernetes_ca_cert=@/var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+  issuer="\"test-aks-cluster-dns-d6cbb78e.hcp.uksouth.azmk8s.io\""
+```
+
+Importantly, this means most common configurations of Vault Agent Injector and
+Vault CSI Provider **cannot share the same Kubernetes auth mount**. Vault
+Agent sidecars will most commonly be configured to authenticate using a
+long-lived default service account token, with an issuer different to the tokens
+Vault CSI Provider will create. But one Kubernetes auth mount can only be
+configured to validate a single issuer value.
```
tvoran, line 95: Good shout, I hadn't thought of this before.

jasonodonnell (approved):
Great addition to the doc, LGTM!

tvoran (approved):

PR #11843, opened 2021-06-11
Clarify new node should not be init'd

mikegreen:


jasonodonnell (approved):

website/content/docs/commands/operator/raft.mdx
```diff
@@ -30,10 +30,10 @@ to join, there must be at least one existing member of the cluster. If Shamir
 seal is in use, then unseal keys are to be supplied before or after the
 join process, depending on whether it's being used exclusively for HA.
 
-If raft is used for `storage`, the node must be joined before unsealing and the
-`leader-api-addr` argument must be provided. If raft is used for `ha_storage`,
-the node must be first unsealed before joining and the `leader-api-addr` must
-_not_ be provided.
+If raft is used for `storage`, the node must be joined before initializing or
```
ncabatoff, line 33: This is a little confusing.  You can't initialize a cluster twice, and you can't join a node to a cluster until the cluster has been initialized.  So how would you join a node *before* initializing?

jasonodonnell: Good point, only one node should be initialized and unsealed, the others should simply join and unseal.

mikegreen: This section is in the context of joining a node to an existing raft cluster. 
Does it less confuse with this context above?
"This command is used to join a new node as a peer to the Raft cluster..."
(https://www.vaultproject.io/docs/commands/operator/raft#join)

Or maybe change to "the new node must be joined" ?

ncabatoff: Maybe break it down into the different cases:

1. First node in a cluster: start node, initialize
2. Second and subsequent nodes, `storage`: start node, join node, unseal node
3. Second and subsequent nodes, `ha_storage`: start node, unseal node, join node

PR #11850, opened 2021-06-14
Add expiration warning to certificate checking for diagnose [VAULT-1018]

HridoyRoy:
- Reformatting TLS checks for diagnose to return warnings and an error
- Rework TLS checks for storage and serviceRegistration to warn in production code using by passing in Otel context to the production code directly
- Split out Certificate parsing into subfunction
- Add warnings for all found certificates if certificate will expire within the week

### DIFF
--- a/serviceregistration/consul/consul_service_registration.go
+++ b/serviceregistration/consul/consul_service_registration.go
@@ -1,6 +1,7 @@
 package consul
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"math/rand"
@@ -146,7 +147,7 @@ func NewServiceRegistration(conf map[string]string, logger log.Logger, state sr.
 	// Set MaxIdleConnsPerHost to the number of processes used in expiration.Restore
 	consulConf.Transport.MaxIdleConnsPerHost = consts.ExpirationRestoreWorkerCount
 
-	SetupSecureTLS(consulConf, conf, logger, false)
+	SetupSecureTLS(context.Background(), consulConf, conf, logger, false)
 
 	consulConf.HttpClient = &http.Client{Transport: consulConf.Transport}
 	client, err := api.NewClient(consulConf)
@@ -178,7 +179,7 @@ func NewServiceRegistration(conf map[string]string, logger log.Logger, state sr.
 	return c, nil
 }
 
-func SetupSecureTLS(consulConf *api.Config, conf map[string]string, logger log.Logger, isDiagnose bool) error {
+func SetupSecureTLS(ctx context.Context, consulConf *api.Config, conf map[string]string, logger log.Logger, isDiagnose bool) error {
 	if addr, ok := conf["address"]; ok {
 		consulConf.Address = addr
 		if logger.IsDebug() {
@@ -216,13 +217,17 @@ func SetupSecureTLS(consulConf *api.Config, conf map[string]string, logger log.L
 			certPath, okCert := conf["tls_cert_file"]
 			keyPath, okKey := conf["tls_key_file"]
 			if okCert && okKey {
-				err := diagnose.TLSFileChecks(certPath, keyPath)
+				warnings, err := diagnose.TLSFileChecks(certPath, keyPath)
+				for _, warning := range warnings {
+					diagnose.Warn(ctx, warning)
+				}
 				if err != nil {
 					return err
 				}
-			} else {
-				return fmt.Errorf("key or cert path: %s, %s, cannot be loaded from consul config file", certPath, keyPath)
+				return nil
+
 			}
+			return fmt.Errorf("key or cert path: %s, %s, cannot be loaded from consul config file", certPath, keyPath)
 		}
 
 		// Use the parsed Address instead of the raw conf['address']
--- a/vault/diagnose/tls_verification.go
+++ b/vault/diagnose/tls_verification.go
@@ -2,11 +2,13 @@ package diagnose
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
+	"time"
 
 	"github.com/hashicorp/vault/internalshared/listenerutil"
 	"github.com/hashicorp/vault/sdk/helper/tlsutil"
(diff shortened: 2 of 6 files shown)


serviceregistration/consul/consul_service_registration.go
```diff
@@ -216,13 +217,17 @@ func SetupSecureTLS(consulConf *api.Config, conf map[string]string, logger log.L
...
+				for _, warning := range warnings {
+					diagnose.Warn(ctx, warning)
+				}
 				if err != nil {
 					return err
 				}
-			} else {
-				return fmt.Errorf("key or cert path: %s, %s, cannot be loaded from consul config file", certPath, keyPath)
+				return nil
+
```
swayne275, line 228:
```suggestion
```

vault/diagnose/tls_verification.go
```diff
@@ -28,64 +37,119 @@ func ListenerChecks(listeners []listenerutil.Listener) error {
 		}
 		_, ok := tlsutil.TLSLookup[l.TLSMinVersion]
 		if !ok {
-			return fmt.Errorf(minVersionError, l.TLSMinVersion)
+			err := fmt.Errorf("listener at address: %s has error %s: ", listenerID, fmt.Sprintf(minVersionError, l.TLSMinVersion))
```
swayne275, line 40:
```suggestion
			err := fmt.Errorf("listener at address: %s has error %w: ", listenerID, fmt.Sprintf(minVersionError, l.TLSMinVersion))
```

swayne275, line 40: is it valuable to wrap the underlying error here (rather than insert a string)? like would that ever feasibly benefit calling code or tests?

vault/diagnose/tls_verification.go
```diff
@@ -102,23 +166,45 @@ func TLSFileChecks(certFilePath, keyFilePath string) error {
...
 
-	// After verify passes, we need to check the values on the certificate itself.
-	// This is a separate check beyond the certificate expiry and chain checks.
+	return nil
+}
 
-	_, err = tls.LoadX509KeyPair(certFilePath, keyFilePath)
-	if err != nil {
-		return err
+func TLSFileWarningChecks(leafcerts, interCerts, rootCerts []*x509.Certificate) ([]string, error) {
```
swayne275, line 182: nit: `leafCerts` for consistency?

vault/diagnose/tls_verification_test.go
```diff
@@ -48,12 +53,15 @@ func TestTLSFakeCert(t *testing.T) {
...
 	}
-	err := ListenerChecks(listeners)
-	if err == nil {
+	_, errs := ListenerChecks(context.Background(), listeners)
+	if errs == nil {
 		t.Fatalf("TLS Config check on fake certificate should fail")
 	}
-	if !strings.Contains(err.Error(), "could not decode cert") {
-		t.Fatalf("Bad error message: %s", err)
+	if len(errs) != 1 {
```
swayne275, line 60: might be useful to print out `errs`

swayne275 (commented):
just a few small things. i think it looks good, but probably best to get some eyes from Scott too

vault/diagnose/tls_verification.go
```diff
@@ -102,23 +166,45 @@ func TLSFileChecks(certFilePath, keyFilePath string) error {
...
+	var warnings []string
+	for _, c := range leafcerts {
+		if NearExpiration(c) {
+			warnings = append(warnings, fmt.Sprintf("certificate %d is expired or near expiry", c.SerialNumber))
```
hghaf099, line 186: Since the messages in this section are all similar, would it make sense to add cert type in the message as well? Like here, we can add "Leaf certificate ... ". This would make it easier to spot the issue.

vault/diagnose/tls_verification.go
```diff
@@ -28,64 +37,119 @@ func ListenerChecks(listeners []listenerutil.Listener) error {
...
-func TLSFileChecks(certFilePath, keyFilePath string) error {
+// ParseTLSInformation parses certficate information and returns it from a cert path.
+func ParseTLSInformation(certFilePath string) ([]*x509.Certificate, []*x509.Certificate, []*x509.Certificate, error) {
+	leafCerts := []*x509.Certificate{}
+	interCerts := []*x509.Certificate{}
+	rootCerts := []*x509.Certificate{}
 	data, err := ioutil.ReadFile(certFilePath)
 	if err != nil {
-		return fmt.Errorf("failed to read tls_client_ca_file: %w", err)
+		return leafCerts, interCerts, rootCerts, fmt.Errorf("failed to read tls_client_ca_file: %w", err)
```
hghaf099, line 101: I think certFilePath is for tls_cert_file, the error, however, states tls_client_ca_file.

vault/diagnose/tls_verification.go
HridoyRoy: There isn't an underlying error constant here -- `minVersionError` is a string.

vault/diagnose/tls_verification.go
HridoyRoy: Sounds good!

vault/diagnose/tls_verification_test.go
HridoyRoy: Yep!

vault/diagnose/tls_verification.go
HridoyRoy: Makes sense to me!

vault/diagnose/tls_verification.go
HridoyRoy: Oops, my mistake -- good catch!

vault/diagnose/tls_verification.go
swayne275: yes, but it was previously an error. i'm wondering if we'd like it to still be an error, and be wrapped?

HridoyRoy: I'm not sure I follow -- `minVersionError` was always a string as far as I'm aware.

swayne275: we discussed in slack, and it sounds like doing so won't be useful. resolving!

swayne275 (approved):

sgmiller (changes requested):
LGTM, but having worked with TLS certs in the past in disfunctional orgs, 1 week is probably way too short for people to take action.  I'd make it 1 month.  They can shrug it off if they want but at least that gives them time to start internal processes.

PR #11854, opened 2021-06-14
Wire configuration checks into diagnose and fix resulting bugs.

sgmiller:


internalshared/configutil/config.go
```diff
@@ -98,36 +98,42 @@ func ParseConfig(d string) (*SharedConfig, error) {
...
 
 	if o := list.Filter("seal"); len(o.Items) > 0 {
+		result.found("seal", "Seal")
 		if err := parseKMS(&result.Seals, o, "seal", 3); err != nil {
 			return nil, fmt.Errorf("error parsing 'seal': %w", err)
 		}
 	}
 
 	if o := list.Filter("kms"); len(o.Items) > 0 {
+		result.found("kms", "Seal")
```
HridoyRoy, line 115: Is this a typo?

HridoyRoy (changes requested):
Hey Scott, I wasn't able to pull this code in locally and take a look due to the build issues. I noticed that these are also on CI, so if you could take a look at those and re-request a review from me, I'll be able to pull locally and dig into the code some more. 
Thanks!

internalshared/configutil/config.go
sgmiller: No, strangely, for the seals, they're all 'seal "sealtype" {' , and we iterate on sealtype but want to remove the "seal" element as one that wasn't found.  TBH it's all a bit of a hack, specifically how HCL 1.0 works with these stanzas.

HridoyRoy (dismissed):
lgtm, but there are some tests that are failing which use fake keys (like "my-stuff" and are having issues.

HridoyRoy (changes requested):
the code lgtm, but there are some tests that are failing which use fake keys (like "my-stuff" and are having issues. These are in both `test-remote-docker` and `test-go-remote-docker`.

`test-go-remote-docker` is actually more worrisome because it seems like `sentinel` isn't recognized by the config checking. If there are unrecognized keys in the config, should it error in the parsing itself?

command/server/test-fixtures/config_diagnose_ok.hcl
```diff
@@ -37,10 +37,6 @@ telemetry {
     metrics_prefix = "myprefix"
 }
 
-sentinel {
```
HridoyRoy, line 40: It would be good to add this back in.

HridoyRoy (approved):

PR #11861, opened 2021-06-14
Cassandra: Refactor PEM parsing logic

pcman312:
# Description

This fixes an issue in the Cassandra database plugin TLS logic where providing only a custom CA doesn't work when using `pem_bundle` and partially works with `pem_json`. This now supports setting either a client certificate & private key, a custom CA, or both.

# How does this work?

The `certutil.ParsePEMBundle` function makes assumptions about the PEM data it is parsing:

1) It assumes the PEM bundle will be used for mTLS and therefore assumes the presence of both a client certificate and a custom CA, though it doesn't error if only the CA is provided. If only a CA is provided to this function, the CA certificate ends up in the `Certificate` field rather than `CAChain` which ends up not being interpreted correctly when converting to a `tls.Config`.
2) It assumes the PEM data is in a specific order (client certificate before CA chain)

A new function has been written (`pemBundleToTLSConfig`) that parses the PEM bundle very similarly to `certutil.ParsePEMBundle` but does not make either of those assumptions. 

Similarly, the `certutil.ParsePKIJSON` function ends up having some similar assumptions, however it does parse the CA correctly and thus the `tls.Config` works when connecting to Cassandra. However, the CA is also included as a client certificate when it shouldn't be. However, it appears that Cassandra will ignore a provided client certificate if not configured to use mTLS.

A new function was created (`jsonBundleToTLSConfig`) that parses the certificate data into a `tls.Config` object. 

# Testing

Additional automated tests were added in `connection_producer_test.go` that start up a Cassandra instance using TLS with a custom CA and connects to it using the same custom CA.

Also this was tested with a real Cassandra instance (also configured with a custom CA) and a real Vault instance configured to use just the custom CA when talking to Cassandra.


plugins/database/cassandra/tls.go
```diff
@@ -0,0 +1,118 @@
...
+				return nil, errutil.UserError{Err: "multiple leaf certificates not supported"}
+			}
+			certificate = string(blockBytes)
+
+		case "PRIVATE KEY":
+			if privateKey != "" {
+				return nil, errutil.UserError{Err: "multiple private keys not supported"}
+			}
+			privateKey = string(blockBytes)
+		}
```
calvn, line 77: What about PKCS#1 (`RSA PRIVATE KEY`) or EC (`EC PRIVATE KEY`) `Block` types? Should there also be a default for any other type that's considered invalid/unsupported for our use case?

plugins/database/cassandra/connection_producer.go
```diff
@@ -83,38 +82,46 @@ func (c *cassandraConnectionProducer) Initialize(ctx context.Context, req dbplug
 		return fmt.Errorf("username cannot be empty")
 	case len(c.Password) == 0:
 		return fmt.Errorf("password cannot be empty")
+	case len(c.PemJSON) > 0 && len(c.PemBundle) > 0:
+		return fmt.Errorf("cannot specify both pem_json and pem_bundle")
```
calvn, line 86: I think that this is good to do, but note that we previously allowed both to be specified at the same time, with preference on `pem_json` so it's technically a breaking change. Would be good to get the docs updated as well to indicate that this is a `one of` between the two parameters.

plugins/database/cassandra/tls.go
```diff
@@ -0,0 +1,118 @@
...
+
+	pemBytes := []byte(pemBundle)
+	var pemBlock *pem.Block
+
+	certificate := ""
+	privateKey := ""
+	caChain := []string{}
+
+	count := 0
+	for len(pemBytes) > 0 {
```
calvn, line 45: Does the ordering while parsing matter? For example, should we check/ensure that there's no private key being inserted in the middle of a chain?

helper/testhelpers/docker/testhelpers.go
```diff
@@ -29,9 +29,13 @@ type Runner struct {
...
 type RunOptions struct {
-	ImageRepo       string
-	ImageTag        string
-	ContainerName   string
+	ImageRepo string
+	ImageTag  string
+
+	ContainerName string
+	// DoNotAppendUUID to the container name for uniqueness
+	DoNotAppendUUID bool
```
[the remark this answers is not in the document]
pcman312: Updated. I like this way better than my original.

plugins/database/cassandra/cassandra_test.go
```diff
@@ -55,16 +55,14 @@ func getCassandra(t *testing.T, protocolVersion interface{}) (*Cassandra, func()
...
+	// getCassandra performs an Initialize call
 	db, cleanup := getCassandra(t, 4)
 	defer cleanup()
 
 	err := db.Close()
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
-
-	db, cleanup = getCassandra(t, "4")
```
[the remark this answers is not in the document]
pcman312: Ah, I missed that it was a string. I'll add this as a separate test as I do want the cleanup function to work correctly.

plugins/database/cassandra/tls.go
```diff
@@ -0,0 +1,118 @@
...
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+
+	"github.com/hashicorp/vault/sdk/helper/certutil"
+	"github.com/hashicorp/vault/sdk/helper/errutil"
+)
+
+func jsonBundleToTLSConfig(rawJSON string, tlsMinVersion uint16, serverName string, insecureSkipVerify bool) (*tls.Config, error) {
+	certBundle := certutil.CertBundle{}
```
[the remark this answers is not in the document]
pcman312: :+1:

plugins/database/cassandra/tls.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	pemBytes := []byte(pemBundle)
+	var pemBlock *pem.Block
+
+	certificate := ""
+	privateKey := ""
+	caChain := []string{}
+
+	count := 0
```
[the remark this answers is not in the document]
pcman312: Whoops. That was for some debugging at one point that I missed. Removed.

plugins/database/cassandra/tls.go
pcman312: Good catch! I'll add `RSA PRIVATE KEY` and `EC PRIVATE KEY` to the `PRIVATE KEY` case.
Re: default: We could either ignore an unsupported block type (current), or error if we find something we don't support. I like the idea of erroring on it to make it clear to the user that what they provided isn't supported rather than failing silently.

plugins/database/cassandra/connection_producer.go
pcman312: True, this is a breaking change. I'm not a fan of allowing for both as it can create confusion ("which one is being used?"). I'll update the docs to describe this change unless anyone has any objections to making this change.

plugins/database/cassandra/tls.go
pcman312: As far as I can tell there is no required ordering down stream. The current `ParsePEMBundle` function also doesn't enforce any ordering (other than the implicit assertion that the client certificate is before the CA cert). We can enforce ordering if we want, but the current implementation doesn't require it.

jasonodonnell (approved):
LGTM and works as expected!

website/content/api-docs/secret/databases/cassandra.mdx
```diff
@@ -42,23 +42,35 @@ has a number of parameters to further configure a connection.
...
-   following structure: `{"ca_chain": ["-----BEGIN CERTIFICATE-----\nMIIEFjC...FNYakP7I\n-----END CERTIFICATE-----"]}`
-   Also make sure the PEM data is properly JSON encoded with `\n` instead of newlines.
+  certificate; or just a CA certificate. Only one of `pem_bundle` or `pem_json` can be specified.
 
 - `pem_json` `(string: "")` – Specifies JSON containing a certificate and
   private key; a certificate, private key, and issuing CA certificate; or just a
   CA certificate. For convenience format is the same as the output of the
   `issue` command from the `pki` secrets engine; see
-  [the pki documentation](/docs/secrets/pki).
+  [the pki documentation](/docs/secrets/pki).  Only one of `pem_bundle` or `pem_json` can be specified.
```
jasonodonnell, line 56: Should we have a note that the certs/key can't have newlines?

pcman312: I've added a note that the PEM data must be properly JSON encoded.

plugins/database/cassandra/tls.go
calvn: I noticed that `ParsePEMBundle` performs validity checks around the client certs and the CA chain, [via](https://github.com/hashicorp/vault/blob/4e0fa83bcda67ea3548d73f4016bbe5a03f3119e/sdk/helper/certutil/helpers.go#L187) `parsedBundle.Verify()`. Wondering if we should do the same.

pcman312: I looked into the `Verify` function and it is specific to mTLS so we can't do the exact same validation. I'm suspicious that we don't need to do much of what it's doing in that function as it's probably handled downstream in the TLS code, but I'll try to double check that.

website/content/api-docs/secret/databases/cassandra.mdx
```diff
@@ -42,23 +42,35 @@ has a number of parameters to further configure a connection.
...
+
 - `pem_bundle` `(string: "")` – Specifies concatenated PEM blocks containing a
   certificate and private key; a certificate, private key, and issuing CA
-  certificate; or just a CA certificate.
-
-!> **Known Issue:** There is a known issue when using `pem_bundle` with only a CA (no client certificate & key)
-   where Vault will not parse the CA certificate correctly. To work around this, use `pem_json` with the
-   following structure: `{"ca_chain": ["-----BEGIN CERTIFICATE-----\nMIIEFjC...FNYakP7I\n-----END CERTIFICATE-----"]}`
-   Also make sure the PEM data is properly JSON encoded with `\n` instead of newlines.
+  certificate; or just a CA certificate. Only one of `pem_bundle` or `pem_json` can be specified.
```
calvn, line 50: Do you think there's value in providing a sample command for constructing this concatenated bundle to minimize the chances of a malformed bundle being provided?

calvn (approved):
Minor comment and a reminder to add a changelog file, but otherwise looks good!

website/content/api-docs/secret/databases/cassandra.mdx
pcman312: I lean towards no because such a command is going to be fairly specific to my development environment.

Issue #11863, opened 2021-06-15
Panic when creating/updating approle role if token_type set to default-service or default-batch


marcboudreau:
**Describe the bug**
When creating or updating an approle role with the token_type parameter set to either `default-service` or `default-batch`, the server (Vault) returns an empty reply and the role is not created/updated.

**To Reproduce**
Setup Vault is dev mode: `docker run -d --name vault -p 8200:8200 -e SKIP_SETCAP=1 vault:1.7.2 vault server -dev -dev-root-token-id=root -dev-listen-address=0.0.0.0:8200`

Any of the following commands will reproduce the behavior:
1. Run `vault write auth/approle/role/new token_type=default-service` 
2. Run `vault write auth/approle/role/new token_type=default-batch`
3. Run `vault write auth/approle/role/new ;  vault write auth/approle/role/new token_type=default-service`
4. Run `vault write auth/approle/role/new2 ; vault write auth/approle/role/new2 token_type=default-batch`

**Effects**
For any of the above commands, the final vault command exits with a status code of 2 with the following message written to stderr: `Error writing data to auth/approle/role/new: Put http://localhost:8200/v1/auth/approle/role/new: EOF`

In addition, the following stack trace is written to the Vault logs (`docker logs vault`):
```
2021-06-15T00:26:00.186Z [INFO]  http: panic serving 172.17.0.1:64822: runtime error: invalid memory address or nil pointer dereference
goroutine 1807 [running]:
net/http.(*conn).serve.func1(0xc0002d3cc0)
	/goroot/src/net/http/server.go:1801 +0x147
panic(0x4abdd60, 0x8b6cad0)
	/goroot/src/runtime/panic.go:975 +0x47a
github.com/hashicorp/vault/vendor/github.com/hashicorp/vault/sdk/logical.(*Response).AddWarning(...)
	/gopath/src/github.com/hashicorp/vault/vendor/github.com/hashicorp/vault/sdk/logical/response.go:78
github.com/hashicorp/vault/builtin/credential/approle.(*backend).pathRoleCreateUpdate(0xc0000e8240, 0x5e9d740, 0xc00138dbc0, 0xc000b7e140, 0xc00093df30, 0x0, 0x0, 0x0)
	/gopath/src/github.com/hashicorp/vault/builtin/credential/approle/path_role.go:891 +0x1ba4
[...]

PR #11864, opened 2021-06-15
Fix for Issue 11863 - Panic when creating/updating approle role with token_type

marcboudreau:
This PR offers a fix for Issue #11863. The cause of the panic was that the variable **resp** was declared but not initialized.  When the warning was added if the token_type provided to the create or update request was either `default-service` or `default-batch`.

This change creates a new **logical.Response** instance and assigns its address to the **resp** variable prior to either cases where the warning is added because of the token_type value.  Since those cases are exclusive, there's no possibility that the **resp** variable be anything but nil at either of those points.

The only other location in the function where the **resp** variable is used is to add the warning about the max_ttl exceeding the mount's tuned max_ttl.  In this case, the code was creating a new **logical.Response**.  I've added a condition around it, so that if a new instance has already been created and a warning about the token_type added, the same instance gets reused.


builtin/credential/approle/path_role_test.go
```diff
@@ -1820,6 +1820,136 @@ func TestAppRole_RoleWithTokenBoundCIDRsCRUD(t *testing.T) {
...
+	roleData = map[string]interface{}{
+		"role_id":            "test_role_id",
+		"policies":           "a,b,c,d",
+		"secret_id_num_uses": 100,
+		"secret_id_ttl":      3000,
+		"token_ttl":          4000,
+		"token_max_ttl":      5000,
+		"token_type":         "default-service",
+	}
+	roleReq.Data = roleData
```
tomhjp, line 1895: It would be nice to turn these into table-based tests so that the inputs for each test case can be treated as immutable and make it a bit clearer to see exactly what each test's inputs are.

changelog/11863.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
```
tomhjp, line 1: Thanks for adding a changelog note - one small update though, the number in the file name should actually be the PR number.

builtin/credential/approle/path_role_test.go
```diff
@@ -1820,6 +1820,136 @@ func TestAppRole_RoleWithTokenBoundCIDRsCRUD(t *testing.T) {
...
+	}
+
+	var expectedStruct roleStorageEntry
+	err = mapstructure.Decode(expected, &expectedStruct)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var actualStruct roleStorageEntry
+	err = mapstructure.Decode(resp.Data, &actualStruct)
```
tomhjp, line 1876: I may be missing something obvious, but can you skip decoding and just compare the two `map[string]interface{}` values? (along with setting the `RoleID` still)

tomhjp (commented):
Thanks for finding and patching this, including the thorough tests! The main change LGTM, just a couple of nits elsewhere.

builtin/credential/approle/path_role_test.go
marcboudreau: To be honest, this test function is just a copy of the above test function **TestAppRole_RoleWithTokenBoundCIDRsCRUD** and then I removed the unnecessary bits.

builtin/credential/approle/path_role_test.go
marcboudreau: Do you mean for just the various cases within this test function?  The table-based tests are what I'm more used to doing, so I don't mind giving it a go.  Seems like this test function could be combined with the TestAppRole_RoleCRUD function at the very least.

changelog/11863.txt
marcboudreau: My bad.  I'll fix that right away.

vishalnayak:
@marcboudreau The changes looks good. The comments Tom left are very valid. However, I'd understand if you prefer to work on refactoring as a subsequent new PR after this one gets in. Up to you. Let us know.

marcboudreau:
@vishalnayak I'm willing to take a stab at refactoring the test code a bit.  If it turns into a rabbit hole, we can backtrack and go with this original PR.

builtin/credential/approle/path_role_test.go
marcboudreau: I also don't really know why the tests go through that additional mapstructure decode operation for the comparison.  If I were to venture a guess, I suppose it does ensure that the fields being compared (some fields are being compared directly aside from the DeepEquals comparison) are properly tagged in the struct definition.  That said, since it's clearly the established pattern in the test file, it feels wrong to break from the norm in a single test function.

marcboudreau:
On second thought @vishalnayak, I'd like to keep the PR as is.  I've looked at the code and found that there is definitely some repetition among a few of the test functions, namely around creating the role, and updating it, but then there's also quite a bit of unique quarks.  Like in the TestAppRole_RoleWithTokenBoundCIDRsCRUD.

vishalnayak:
> On second thought @vishalnayak, I'd like to keep the PR as is. I've looked at the code and found that there is definitely some repetition among a few of the test functions, namely around creating the role, and updating it, but then there's also quite a bit of unique quarks. Like in the TestAppRole_RoleWithTokenBoundCIDRsCRUD.

Totally understandable. I've seen that part of the code and had a feeling that you might hit the thoughts you did.

Thanks for the submission!

Hoping to see the refactoring PR from you as well, but no rush.

vishalnayak (approved):

PR #11867, opened 2021-06-15
vault: fix dropped identity store test errors

alrs:
This fixes two dropped test `err` variables in the `vault` package.


vishalnayak (approved):
Thank you!

PR #11870, opened 2021-06-15
chore:  add local vault server wrappers in yarn and document in readme

randallmorey:
This PR adds convenient yarn wrappers around commonly-used local Vault server commands.  Since Vault UI development requires a running Vault server, documenting these commands in the README is helpful to get new developers up and running quickly.


hashishaw (approved):
Thanks for improving the docs! 🚀

PR #11872, opened 2021-06-15
Mongo doesnt allow periods in usernames

mr-miles:
We found that the default username template can create invalid usernames which aren't accepted by mongo.  Replacing the dots with an allowed character fixes it.

### DIFF
--- a/plugins/database/mongodb/mongodb.go
+++ b/plugins/database/mongodb/mongodb.go
@@ -21,7 +21,7 @@ import (
 const (
 	mongoDBTypeName = "mongodb"
 
-	defaultUserNameTemplate = `{{ printf "v-%s-%s-%s-%s" (.DisplayName | truncate 15) (.RoleName | truncate 15) (random 20) (unix_time) | truncate 100 }}`
+	defaultUserNameTemplate = `{{ printf "v-%s-%s-%s-%s" (.DisplayName | replace "." "-" | truncate 15) (.RoleName | replace "." "-" | truncate 15) (random 20) (unix_time) | truncate 100 }}`
 )
 
 // MongoDB is an implementation of Database interface
--- a/plugins/database/mongodb/mongodb.go
+++ b/plugins/database/mongodb/mongodb.go
@@ -21,7 +21,7 @@ import (
 const (
 	mongoDBTypeName = "mongodb"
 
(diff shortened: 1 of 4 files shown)


vishalnayak (approved):

vishalnayak:
@mr-miles The changes looks good to me. Would you be able to add a test to this effect?

kalafut:
@pcman312  Any issues with this change? I also wonder if there are other special characters to avoid. Should be listed in CHANGES too?

mr-miles:
Will attempt to add a test.

SO says the one other restriction on keys is that they can't start with a dollar
https://stackoverflow.com/questions/7975185/mongodb-query-over-a-hash-with-special-chars-in-keys/7976235#7976235

so I think the default template is safe

plugins/database/mongodb/mongodb.go
```diff
@@ -21,7 +21,7 @@ import (
 const (
 	mongoDBTypeName = "mongodb"
 
-	defaultUserNameTemplate = `{{ printf "v-%s-%s-%s-%s" (.DisplayName | truncate 15) (.RoleName | truncate 15) (random 20) (unix_time) | truncate 100 }}`
+	defaultUserNameTemplate = `{{ printf "v-%s-%s-%s-%s" (.DisplayName | replace "." "-" | truncate 15) (.RoleName | replace "." "-" | truncate 15) (random 20) (unix_time) | truncate 100 }}`
```
pcman312, line 24: This looks good! What do you think about moving the replace to the end so it only needs to be done once instead of for each of the two fields?

```
{{ printf "v-%s-%s-%s-%s" (.DisplayName | truncate 15) (.RoleName | truncate 15) (random 20) (unix_time) | replace "." "-" | truncate 100 }}
```

pcman312 (commented):
This looks good to me. Can you also update the API docs to reflect the updated default template?
https://github.com/hashicorp/vault/blob/master/website/content/api-docs/secret/databases/mongodb.mdx#parameters

Thanks!

plugins/database/mongodb/mongodb.go
mr-miles: Docs and template updated

mr-miles:
Test added also

kalafut:
Thanks for the updates. Can you please add a changelog file to this PR?  The /changelog folder has examples.

mr-miles:
done (and hopefully correctly!)

vishalnayak (approved):

vishalnayak:
Thanks for submitting this!

PR #11873, opened 2021-06-15
Fix error log typo for failing to decrypt keys

kinghrothgar:
Fixes small typo that can lead to significant confusion when debugging seal key decryption failures.

### DIFF
--- a/vault/seal.go
+++ b/vault/seal.go
@@ -488,7 +488,7 @@ func readStoredKeys(ctx context.Context, storage physical.Backend, encryptor *se
 
 	pt, err := encryptor.Decrypt(ctx, blobInfo, nil)
 	if err != nil {
-		return nil, &ErrDecrypt{Err: fmt.Errorf("failed to encrypt keys for storage: %w", err)}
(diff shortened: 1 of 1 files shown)


vault/seal.go
```diff
@@ -488,7 +488,7 @@ func readStoredKeys(ctx context.Context, storage physical.Backend, encryptor *se
 
 	pt, err := encryptor.Decrypt(ctx, blobInfo, nil)
 	if err != nil {
-		return nil, &ErrDecrypt{Err: fmt.Errorf("failed to encrypt keys for storage: %w", err)}
+		return nil, &ErrDecrypt{Err: fmt.Errorf("failed to decrypt stored keys: %w", err)}
```
vishalnayak, line 491:
```suggestion
		return nil, &ErrDecrypt{Err: fmt.Errorf("failed to decrypt keys from storage: %w", err)}
```

kinghrothgar: Suggested edit applied in force push

vishalnayak (approved):

PR #11877, opened 2021-06-16
[RELENG-444] Update the README to include EULA and Terms of Evaluation legal blurb

claire-labry:
As outlined in ENGSRV-062 RFC, we will need to include a legal blurb that outlines the End-User License Agreement and Terms of Evaluation to be able to use Vault Enterprise in the README.md.


mladlow (approved):

claire-labry:
Hi there @mladlow --  I just recently chatted with my team and we came to the conclusion that we should not update the OSS repos with the EULA + TOE blurb as it isn't one of the requirements explicitly called in our PRD (and is causing some confusion regarding updating this blurb in a OSS repo, not the ENT repo), it was a confusion on my part. I will be closing this PR and will be going along my way! 😅 Thank you for taking the time to review this.

PR #11878, opened 2021-06-16
UI/cp validations kv duplicate path

Monkeychip:
This is the last validation piece for the KV secret engine.  

duplicate-paths

The reason Design wanted to include this validation error is because of the Require check and set option, which is on by default in the UI.  If you name the path the same, we currently show the following API error, which is not very helpful. So instead of hitting this API error, we are instead returning a validation error that's more clear.

image

**Notes:**
- I did not include a changelog because it's covered under the previous changelog "adding validation to the kv secret engine"
- I did not include this in the last PR because I wanted to get that through so I would unblock another team member.
- I could have used peekRecord however it would not have checked the multiple pages of path names nor would it return values if you cold refreshed from the create secret page.  You had to have navigated from the list of paths page first.


hashishaw (approved):
nice!

PR #11880, opened 2021-06-16
Update website latest version

kalafut:


mladlow (approved):
I _think_ this should actually be RC1 (from past experience).

PR #11884, opened 2021-06-17
UI/fix safari oidc login

hashishaw:
Due to a [Safari localStorage bug](https://twitter.com/jaffathecake/status/1389493762129375232?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E1389493762129375232%7Ctwgr%5E%7Ctwcon%5Es1_c10&ref_url=https%3A%2F%2Fappleinsider.com%2Farticles%2F21%2F05%2F04%2Fmacos-bugs-causing-sporadic-browsing-issues-with-safari-firefox-others), logging into Vault with OIDC was not working because the login tab never saw the `localStorage` write from the popup tab. To resolve, we are replacing `localStorage` in this scenario with a [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) call between the two tabs. 

**Before:**
safari-before

**After:**
safari-after

### DIFF
--- a/ui/app/components/auth-jwt.js
+++ b/ui/app/components/auth-jwt.js
@@ -82,13 +82,21 @@ export default Component.extend({
     this.onError(err);
   },
 
-  prepareForOIDC(oidcWindow) {
+  prepareForOIDC: task(function*(oidcWindow) {
+    const thisWindow = this.getWindow();
     // show the loading animation in the parent
     this.onLoading(true);
     // start watching the popup window and the current one
     this.watchPopup.perform(oidcWindow);
     this.watchCurrent.perform(oidcWindow);
-  },
+    // wait for message posted from popup
+    const event = yield waitForEvent(thisWindow, 'message');
+    if (event.origin === thisWindow.origin && event.isTrusted) {
+      this.exchangeOIDC.perform(event.data, oidcWindow);
+    } else {
+      this.handleOIDCError();
+    }
+  }),
 
   watchPopup: task(function*(oidcWindow) {
     while (true) {
@@ -177,10 +185,7 @@ export default Component.extend({
         'vaultOIDCWindow',
         `width=${POPUP_WIDTH},height=${POPUP_HEIGHT},resizable,scrollbars=yes,top=${top},left=${left}`
       );
-      win.addEventListener('message', event => {
-        this.exchangeOIDC.perform(event.data, oidcWindow);
-      });
-      this.prepareForOIDC(oidcWindow);
+      this.prepareForOIDC.perform(oidcWindow);
     },
   },
 });
--- a/ui/app/components/auth-jwt.js
+++ b/ui/app/components/auth-jwt.js
@@ -82,17 +82,13 @@ export default Component.extend({
     this.onError(err);
   },
 
-  prepareForOIDC: task(function*(oidcWindow) {
+  prepareForOIDC(oidcWindow) {
     // show the loading animation in the parent
     this.onLoading(true);
     // start watching the popup window and the current one
     this.watchPopup.perform(oidcWindow);
     this.watchCurrent.perform(oidcWindow);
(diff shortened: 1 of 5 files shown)


ui/tests/integration/components/auth-jwt-test.js
```diff
@@ -203,56 +203,88 @@ module('Integration | Component | auth jwt', function(hooks) {
...
-    this.window.localStorage.getItem.returns(null);
-    this.window.trigger('storage', { storageArea: this.window.localStorage });
+    this.window.trigger('message', buildMessage({ data: { state: 'state', foo: 'bar' } }));
     run.cancelTimers();
-    assert.ok(this.window.localStorage.getItem.calledOnce, 'calls getItem');
-    assert.notOk(this.window.localStorage.removeItem.called, 'never calls removeItem');
+    assert.equal(this.error, ERROR_MISSING_PARAMS, 'calls onError with params missing error');
   });
 
-  test('oidc: storage event fires with state key, wrong params', async function(assert) {
```
hashishaw, line 221: Removed this test because we don't fire en event

ui/app/components/auth-jwt.js
```diff
@@ -114,20 +119,14 @@ export default Component.extend({
...
-      ? event.storageArea.getItem('oidcState')
-      : yield timeout(1000).then(() => event.storageArea.getItem('oidcState'));
-
+  exchangeOIDC: task(function*(oidcState, oidcWindow) {
     if (oidcState === null || oidcState === undefined) {
       return;
     }
     this.onLoading(true);
     // get the info from the event fired by the other window and
     // then remove it from localStorage
```
Monkeychip, line 128: amend comment as it looks like we're no longer removing from localstorage

ui/app/templates/components/auth-form.hbs
```diff
@@ -55,7 +55,7 @@
       @roleName={{this.roleName}}
       @selectedAuthType={{this.selectedAuthBackend.type}}
       @selectedAuthPath={{or this.customPath this.selectedAuthBackend.id}}
-      @disabled={{authenticate.isRunning}}
+      @disabled={{or authenticate.isRunning this.isLoading}}
```
hashishaw, line 58: Added this because once the popup is closed, while things are processing but before authenticate runs, the button is clickable again and it was confusing

ui/tests/integration/components/auth-jwt-test.js
```diff
@@ -15,6 +15,12 @@ import { ERROR_WINDOW_CLOSED, ERROR_MISSING_PARAMS, ERROR_JWT_LOGIN } from 'vaul
 
 const component = create(form);
 const windows = [];
+const buildMessage = opts => ({
+  isTrusted: true,
+  origin: 'http://localhost:8200',
```
Monkeychip, line 20: I guess we'll see, but I'd be surprised if this works on CircleCI, or maybe I wouldn't be because Circle CI is a bit of a black box.

Monkeychip (approved):
Code looks good! Two comments: changelog and I just want to confirm that this works fine on other browsers.

ui/app/templates/components/auth-form.hbs
Monkeychip: Smart

ui/tests/integration/components/auth-jwt-test.js
hashishaw: it works! The window and events are all mocked, but the logic in the component would fail if origin from the event doesn't match which is why I added this. It could be anything, I'll change it to something generic so that is clear

Monkeychip: Rad.  I learned something new today. Thanks!

picatz (approved):
LGTM 👍

PR #11891, opened 2021-06-17
test(website): a11y report automation (🛹)

kendallstrautman:
## 🚧 Experimental 🚧

This PR adds a GH action for running [pa11y-ci](https://github.com/pa11y/pa11y-ci) to asses accessibility errors on the website. This PR also adds a sitemap generator, as that output is what `pa11y-ci` uses as a basis to test routes. The output from the test is posted as a comment to the PR. 

My main concerns is the feedback being too large / illegible. Of course, that the feedback is so noisy is indicative of the amount of accessibility improvements we need to make. 

This approach could also be done locally with our sites to generate baseline artifacts on our a11y conformance.

### DIFF
--- a/.github/workflows/website-a11y-tests.yml
+++ b/.github/workflows/website-a11y-tests.yml
@@ -16,14 +16,14 @@ jobs:
       - name: Set up node
         uses: actions/setup-node@v1
 
-      - name: Install pa11y-ci dependencies.
-        run: npm install
+      - name: Install pa11y-ci dependencies 
+        run: cd website && npm install 
 
-      - name: Start up dev server.
-        run: npm run start
+      - name: Start dev server
+        run: cd website && npm run start
 
       - name: Run pa11y-ci.
-        run: npm run a11y-test:dev 2>&1 | tee pa11y_output.txt
+        run: cd website && npm run a11y-test:dev 2>&1 | tee pa11y_output.txt
         
       - name: Read pa11y_output file.
         id: pa11y_output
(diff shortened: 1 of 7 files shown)


website/scripts/generate_sitemap.js
```diff
@@ -0,0 +1,46 @@
...
+
+/** Based on this blog by Lee Robinson: https://leerob.io/blog/nextjs-sitemap-robots */
+
+;(async () => {
+  const prettierConfig = await prettier.resolveConfig('./.prettierrc.js')
+
+  // Ignore Next.js specific files (e.g., _app.js) and API routes.
+  const pages = await globby([
+    'pages/**/*{.js,.tsx,.jsx}',
+    //  'content/**/*.mdx', // docs routes
```
kendallstrautman, line 13: Only printing out the static routes currently as running the a11y test on all docs routes would take a really long time. Obviously the sitemap should have all routes so maybe I shouldn't use that as the base for a11y testing.

.github/workflows/website-a11y-tests.yml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          message: '<details><summary>Pa11y testing results</summary>
+
+
+```${{ steps.pa11y_output.outputs.content }}```
+
+</details>'
+
+      - name: Check for pa11y failures. Still pass even if there are failures.
```
kendallstrautman, line 50: This isn't great but there are _too_ many errors right now, it will always fail. But at least we have the feedback for a baseline.

kendallstrautman:
The formatting of the output from `pa11y-cli` is not very easy to read 🤨

kendallstrautman:
Closing this PR. As a team we decided that unit tests may be a better approach for this type of feedback for now. And we could use this broad approach routinely to get zoomed out perspective on conformance.

PR #11894, opened 2021-06-17
Adding github action to check for changelog file

mladlow:
This might have to be slightly different on ENT, where changelog files should be prefixed with an underscore, but I thought having the OSS version would be useful to do in the meantime.

This won't be a required check, and is skippable with the "pr/no-changelog" label.

### DIFF
--- a/.github/workflows/changelog-checker.yml
+++ b/.github/workflows/changelog-checker.yml
@@ -31,7 +31,7 @@ jobs:
           if [ -z "$changelog_files" ]; then
             echo "Did not find a changelog entry named ${{ github.event.pull_request.number }}.txt"
             echo "If your changelog file is correct, skip this check with the 'pr/no-changelog' label"
-            echo "Reference - https://github.com/hashicorp/vault/pull/10363"
+            echo "Reference - https://github.com/hashicorp/vault/pull/10363 and https://github.com/hashicorp/vault/pull/11894"
             exit 1
           elif grep -q ':enhancement$' $changelog_files; then
             # "Enhancement is not a valid type of changelog entry, but it's a common mistake.
--- a/.github/workflows/changelog-checker.yml
+++ b/.github/workflows/changelog-checker.yml
@@ -6,7 +6,7 @@ name: Check Changelog
 on:
   pull_request:
     types: [opened, synchronize, labeled, unlabeled]
-    # Runs on PRs to master and all release branches
+    # Runs on PRs to master
     branches:
       - master/*
 
--- a/.github/workflows/changelog-checker.yml
+++ b/.github/workflows/changelog-checker.yml
@@ -6,10 +6,9 @@ name: Check Changelog
 on:
   pull_request:
     types: [opened, synchronize, labeled, unlabeled]
-    # Runs on PRs to master and all release branches
+    # Runs on PRs to master
     branches:
-      - master/*
-      - release/*
+      - master
 
 jobs:
   # checks that a changelog entry is present for a PR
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog-checker.yml
```diff
@@ -0,0 +1,47 @@
...
+# or there is a changelog/<pr number>.txt file associated with a PR for a changelog entry
+
+name: Check Changelog
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled]
+    # Runs on PRs to master
+    branches:
+      - master/*
```
ncabatoff, line 11: Why limit to master, don't we want them on releases too?

.github/workflows/changelog-checker.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0 # by default the checkout action doesn't checkout all branches
+      - name: Check for changelog entry in diff
+        run: |
+          # check if there is a diff in the changelog directory
+          changelog_files=$(git --no-pager diff --name-only HEAD "$(git merge-base HEAD "origin/${{ github.event.pull_request.base.ref }}")" -- changelog/${{ github.event.pull_request.number }}.txt)
```
ncabatoff, line 28: Might be clearer to break this up into multiple steps.

.github/workflows/changelog-checker.yml
```diff
@@ -0,0 +1,47 @@
...
+
+          # If we do not find a file in changelog/, we fail the check
+          if [ -z "$changelog_files" ]; then
+            # Fail status check when no changelog entry was found on the PR
+            echo "Did not find a changelog entry and the 'pr/no-changelog' label was not applied. Reference - https://github.com/hashicorp/vault/pull/10363"
+            exit 1
+          else
+            # Don't want square brackets bc not using test
+            if $(grep -q 'enhancement' $changelog_files); then
```
ncabatoff, line 37: You don't even need to create a subshell, it's valid to just say:
```
if grep -q needle haystack; then
```

.github/workflows/changelog-checker.yml
```diff
@@ -0,0 +1,47 @@
...
+              # "Enhancement is not a valid type of changelog entry, but it's a common mistake.
```
ncabatoff, line 38: Do we want to constrain this check further?  Right now it forbids the use of the word altogether, maybe we can instead grep for e.g. `:enhancement$`?

.github/workflows/changelog-checker.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0 # by default the checkout action doesn't checkout all branches
+      - name: Check for changelog entry in diff
+        run: |
+          # check if there is a diff in the changelog directory
+          changelog_files=$(git --no-pager diff --name-only HEAD "$(git merge-base HEAD "origin/${{ github.event.pull_request.base.ref }}")" -- changelog/${{ github.event.pull_request.number }}.txt)
+
```
ncabatoff, line 29: Should it be an error to have more than one changelog file?  Similarly, do we want to try to catch an incorrectly numbered changelog file?

.github/workflows/changelog-checker.yml
mladlow: Theoretically that should already be caught by the master check?

.github/workflows/changelog-checker.yml
mladlow: This part I've copied from Consul's version (https://github.com/hashicorp/consul/pull/9781/files), and I'm inclined to keep it identical although I agree it's not very readable.

.github/workflows/changelog-checker.yml
mladlow: I don't think it should be an error to have more than one. I tested this and it does wind up checking that the number matches - https://github.com/hashicorp/vault-plugin-mock/pull/76. You can see various permutations in the history on this PR.

.github/workflows/changelog-checker.yml
mladlow: But also why not, I guess.

kalafut (approved):
🚢 🇮🇹

.github/workflows/changelog-checker.yml
calvn: How does this work for changelog file(s) that is referencing an external PR, e.g. `_12345.txt`?

.github/workflows/changelog-checker.yml
```diff
@@ -0,0 +1,48 @@
...
+          # If we do not find a file in changelog/, we fail the check
+          if [ -z "$changelog_files" ]; then
+            # Fail status check when no changelog entry was found on the PR
+            echo "Did not find a changelog entry and the 'pr/no-changelog' label was not applied. Reference - https://github.com/hashicorp/vault/pull/10363"
+            exit 1
+          else
+            # Don't want square brackets bc not using test
+            if grep -q ':enhancement$' $changelog_files; then
```
calvn, line 38: Another common one that I've seen in the past is `bug` (correct) vs `bugs` (incorrect).

calvn (approved):
Couple of small things, but 🎉  for this!

.github/workflows/changelog-checker.yml
briankassouf: Will this break on backport PRs to the `release/*` branches since we are cherry-picking and not renaming this file?

briankassouf (approved):
Just the one comment about backport PRs and the PR number, looks good otherwise!

.github/workflows/changelog-checker.yml
ncabatoff: Also, I think that trailing `/*` shouldn't be there for master.

.github/workflows/changelog-checker.yml
ncabatoff: Yeah, what I meant was that it looks like this code wouldn't see a cl file for a different pr#, so we'd say "did not find a changelog entry" when we could give a more specific error like "incorrectly named changelog entry".

.github/workflows/changelog-checker.yml
mladlow: > How does this work for changelog file(s) that is referencing an external PR, e.g. `_12345.txt`?

The check will fail. The check isn't required, and can be skipped by adding a label, if you don't want a failing check when you merge. This check is intended as a gentle nudge: "hey, did you want a changelog for this?", so hopefully it would prompt someone to verify that their _ changelog file is present and formatted correctly. I have updated the error message though.

Based on this comment I took a look at some options for handling more variations on changelog entries - with an _, maybe with multiple files, maybe you're adding a changelog file to a PR where the number doesn't match for some other reason, etc. We could certainly handle more cases with this check, but there are a lot of possible edge cases, and I think it might be best to just have this handle the simple case of: I have a PR, and if it should have a changelog entry, the changelog entry should probably match the PR number.

> Will checking the PR number break on backport PRs to the release/* branches since we are cherry-picking and not renaming this file?

Good point, I'll remove it for release branches.

.github/workflows/changelog-checker.yml
mladlow: I'll update the error to be more descriptive, also keeping in mind Calvin's thoughts in the above convo.

.github/workflows/changelog-checker.yml
```diff
@@ -0,0 +1,49 @@
...
+# or there is a changelog/<pr number>.txt file associated with a PR for a changelog entry
+
+name: Check Changelog
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled]
+    # Runs on PRs to master
+    branches:
+      - master
```
vishalnayak, line 11: This can now be renamed to `main` I think.

mladlow: Jim's done this on `main` already, as part of his PR to do the branch switchover.

PR #11897, opened 2021-06-18
Rename some references from master to main

kalafut:
Prepare to rename the Vault repo master branch to main.

### DIFF
--- a/scripts/update_plugin_modules.sh
+++ b/scripts/update_plugin_modules.sh
@@ -32,7 +32,7 @@ do
 		cd ..
 	fi
 	cd vault
-	go get github.com/hashicorp/$plugin@master
+	go get github.com/hashicorp/$plugin@main
 	cd ..
 done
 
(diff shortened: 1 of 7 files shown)


mladlow (approved):
I'm happy to approve, get this merged, and fix whatever breaks, but the changelog action I just added (.github/workflows/changelog-checker.yml) references master if you don't mind adding that?

scripts/update_plugin_modules.sh
```diff
@@ -32,7 +32,7 @@ do
 		cd ..
 	fi
 	cd vault
-	go get github.com/hashicorp/$plugin@master
+	go get github.com/hashicorp/$plugin@main
```
calvn, line 35: I wonder if we should do `... || go get github.com/hashicorp/$plugin@master` as a fallback mechanism in the near term in case this updated script gets run before all repos are updated.

kalafut: Do we even use this script? If so, I'd rather see the failure so we can go fix them (the rename is a button click in the Github UI).

PR #11907, opened 2021-06-21
raft: Set BatchApplyCh for more consistent batch sizes

briankassouf:
See https://github.com/hashicorp/raft/pull/445 and https://github.com/hashicorp/raft/pull/124 for more details

### DIFF
--- a/changelog/11907.txt
+++ b/changelog/11907.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+raft: Improve raft batch size selection
+```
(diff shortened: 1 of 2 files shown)


ncabatoff (approved):
I don't really understand the implications of this breaking the timeout contract in Apply, but I take it you do and aren't concerned about it?

briankassouf:
That's correct, we don't currently use the timeout feature so that shouldn't affect us 

https://github.com/hashicorp/vault/blob/43ccb633b5e0df4b0b67b77e026cfbdf0d37a8f2/physical/raft/raft.go#L1334

PR #11910, opened 2021-06-21
TLS Verification Bugfixes

HridoyRoy:
The crux of the matter is:

1. diagnose.Error does NOT set the error status of the span or fail the span. diagnose.Fail does that. 
2. sanitizedListeners simply doesn't include listeners that failed the initListeners check. We are using instead the unsanitized listeners for the TLS verification checks. 
3. diagnose.Fail combined with diagnose.Error in a particular check caused the output to only print out warnings, but NOT set status. This is because in the output code, if there were warnings the check name and status of the span were not printed (they were skipped).

In addition to these 3 things, I also removed some diagnose calls in production code (that code was meant to be cleaned up in a separate PR), and moved the listener TLS checks that were in operator_diagnose.go into the diagnose package.


vault/diagnose/helpers_test.go
```diff
@@ -63,27 +85,51 @@ func TestDiagnoseOtelResults(t *testing.T) {
...
 	SpotCheck(ctx, "pick-scone", pickScone)
 
 	Test(ctx, "dispose-grounds", Skippable("dispose-grounds", disposeGrounds))
 	return nil
 }
 
+// buildMicrowave will throw an error in the function itself to fail the span,
+// but will return nil so the caller test doesn't necessarily throw an error.
+// The intended behavior is that the superspan will detect the failed subspan
+// and fail regardless.This happens when Fail is used to fail the span, but not
```
swayne275, line 117:
```suggestion
// and fail regardless. This happens when Fail is used to fail the span, but not
```

vault/diagnose/helpers_test.go
```diff
@@ -63,27 +85,51 @@ func TestDiagnoseOtelResults(t *testing.T) {
...
+// when Error is used. See the comment in the function itself.
+func buildMicrowave(ctx context.Context) error {
+	ctx, span := StartSpan(ctx, "buy-parts")
+
+	Fail(ctx, "no stores sell microwave parts, please buy a microwave instead.")
+
+	// The error line here does not actually yield an error in the output.
+	// TODO: Debug this. In the meantime, always use Fail over Error.
```
swayne275, line 125: Does this TODO need to be addressed?

vault/diagnose/tls_verification.go
```diff
@@ -21,17 +20,32 @@ const maxVersionError = "'tls_max_version' value %q not supported, please specif
...
 	var listenerWarnings []string
 	var listenerErrors []error
 
-	for _, listener := range listeners {
-		l := listener.Config
+	for _, l := range listeners {
 		listenerID := l.Address
 
+		if l.TLSDisable {
+			Warn(ctx, fmt.Sprintf("listener at address: %s has error %s. ", listenerID, "TLS is disabled in a Listener config stanza"))
```
swayne275, line 37: did you want these trailing spaces here (for like concat future strings on or something), or is it a typo?

swayne275: also, does it make sense to just const the string that's being added (or just include it in the first string if not)?

swayne275 (commented):
I think it looks good! A few small comments then re-request and I'll approve

vault/diagnose/helpers_test.go
HridoyRoy: It's not urgent since we can and are just avoiding the use of `Error` for now, but we'd need to fix it eventually. I could put a `deprecated` warning above the Error method, but I figure we'll be fixing it post code-freeze, so it's fine.

vault/diagnose/tls_verification.go
HridoyRoy: Nope! I'll remove the trailing spaces! Yeah, good point -- I'll just include it in the first string.

vault/diagnose/tls_verification.go
```diff
@@ -21,17 +20,32 @@ const maxVersionError = "'tls_max_version' value %q not supported, please specif
...
+			continue
+		}
+		if l.TLSDisableClientCerts {
+			Warn(ctx, fmt.Sprintf("listener at address: %s has error %s. ", listenerID, "TLS for a listener is turned on without requiring client certs"))
+
+		}
+		status, warning := TLSMutualExclusionCertCheck(l)
+		if status != 0 {
```
hghaf099, line 45: minor, but, would it make sense to not use the negation? like : " if status == 1"
Since we are printing a warning, and we have three status values. This would avoid confusion.

vault/diagnose/tls_verification.go
swayne275: cool! mark as resolved whenever you do and i'll re-review

vault/diagnose/tls_verification.go
HridoyRoy: Definitely!

swayne275 (approved):

PR #11912, opened 2021-06-21
Prototype track number of inflight lease revocations

swayne275:
This is for VAULT-1304 to provide better insight into lease revocation status/load/progress for operators and support for 1.6.x releases


StymieB (approved):

swayne275:
FYI failing `ActivityLog` race tests are because we didn't backport the race test fixes to `release/1.6.x*`

PR #11920, opened 2021-06-22
KV 2 fix number as key

Monkeychip:
Right now if you enter a number as a path you get an error when viewing the value later.  I modified an test to cover this scenario.

Also, while I was in here, I did a quick fix on my previous validations pr with duplicate paths.  I needed to add a conditional so it only checked the path and not also the key.

**See previous behavior:**
error

**See behavior with fix:**
fix


Monkeychip:
This was fixed with a revert as it was caused by a regression bug.

PR #11921, opened 2021-06-22
Backport [1.7.x]: Cassandra: Refactor PEM parsing logic (#11861)

pcman312:
Backports #11861 to 1.7.x

Original PR description:
===
# Description

This fixes an issue in the Cassandra database plugin TLS logic where providing only a custom CA doesn't work when using `pem_bundle` and partially works with `pem_json`. This now supports setting either a client certificate & private key, a custom CA, or both.

# How does this work?

The `certutil.ParsePEMBundle` function makes assumptions about the PEM data it is parsing:

1) It assumes the PEM bundle will be used for mTLS and therefore assumes the presence of both a client certificate and a custom CA, though it doesn't error if only the CA is provided. If only a CA is provided to this function, the CA certificate ends up in the `Certificate` field rather than `CAChain` which ends up not being interpreted correctly when converting to a `tls.Config`.
2) It assumes the PEM data is in a specific order (client certificate before CA chain)

A new function has been written (`pemBundleToTLSConfig`) that parses the PEM bundle very similarly to `certutil.ParsePEMBundle` but does not make either of those assumptions. 

Similarly, the `certutil.ParsePKIJSON` function ends up having some similar assumptions, however it does parse the CA correctly and thus the `tls.Config` works when connecting to Cassandra. However, the CA is also included as a client certificate when it shouldn't be. However, it appears that Cassandra will ignore a provided client certificate if not configured to use mTLS.

A new function was created (`jsonBundleToTLSConfig`) that parses the certificate data into a `tls.Config` object. 

# Testing

Additional automated tests were added in `connection_producer_test.go` that start up a Cassandra instance using TLS with a custom CA and connects to it using the same custom CA.

Also this was tested with a real Cassandra instance (also configured with a custom CA) and a real Vault instance configured to use just the custom CA when talking to Cassandra.


austingebauer (approved):
LGTM

jasonodonnell (approved):

PR #11923, opened 2021-06-23
Diagnose Storage Panic Bugfixes

HridoyRoy:
This PR fixes panics in the storage and raft checks by:

1. Skipping server traversal in the raft config in `GetConfigOffline` if no servers exist
2. Returning from Diagnose early if storage could not be configured


physical/raft/raft.go
```diff
@@ -1014,6 +1014,11 @@ func (b *RaftBackend) GetConfigurationOffline() (*RaftConfigurationResponse, err
 	config := &RaftConfigurationResponse{
 		Index: state.Index,
 	}
+
+	if config.Servers == nil {
+		return config, nil
```
swayne275, line 1019: can you explain this? is no servers not an error?

command/server/test-fixtures/config_raft.hcl
```diff
@@ -9,7 +9,7 @@ listener "tcp" {
 }
 
 storage "raft" {
-  path = "./server/test-fixtures/raft_storage_file.db"
+  path = "./server/test-fixtures/"
```
swayne275, line 12: could you explain or link a reason for this config change?

swayne275 (approved):
A few questions for my own knowledge, but approved!

command/server/test-fixtures/config_raft.hcl
HridoyRoy: Absolutely! the boltDB file is created by the raft backend -- the config is meant to have the path to the directory where vault can create this file.

physical/raft/raft.go
HridoyRoy: Nope, because this is essentially a cache of the previous raft state that the fsm records. If vault was never run for example, there'd be no servers. Diagnose will throw a warning downstream where it says `even number of servers detected: 0`, which I'm hoping will be a red flag to operators that have been running a vault server, but won't be as strong a nuisance as an error to those that have a fresh binary.

briankassouf (approved):

PR #11935, opened 2021-06-24
ci: cache go modules

samsalisbury:
Avoid downloading modules so often by caching them.

Download and cache go modules early in the workflow.
Disable GOPROXY during build and test runs so that they rely on pre-downloaded modules.

Fail if any of these files are modified by go mod download:

- go.sum
- sdk/go.sum
- api/go.sum


vishalnayak (approved):

.circleci/config/commands/go_test.yml
```diff
@@ -116,8 +117,14 @@ steps:
...
 
+          # Copy the downloaded modules inside the container.
+          docker exec testcontainer sh -c 'mkdir -p /go/pkg'
+          docker cp "$(go env GOPATH)/pkg/mod" testcontainer:/go/pkg/mod
+
           docker exec -w /go/src/github.com/hashicorp/vault/ \
-            -e CIRCLECI -e GOCACHE=/tmp/gocache -e VAULT_CI_GO_TEST_RACE \
+            -e CIRCLECI -e VAULT_CI_GO_TEST_RACE \
+            -e GOCACHE=/tmp/gocache \
+            -e GOPROXY="off" \
```
ncabatoff, line 127: Why no GOPRIVATE?

samsalisbury: Good point, I intended to replace this file with the ENT one. I'll open a PR to that effect now @ncabatoff

samsalisbury: Please see https://github.com/hashicorp/vault/pull/11951 and its sister ent PR

PR #11937, opened 2021-06-24
Update plugin-portal.mdx

acahn:
Adding in a few community supported plugins

### DIFF
--- a/website/content/docs/plugin-portal.mdx
+++ b/website/content/docs/plugin-portal.mdx
@@ -122,11 +122,13 @@ Plugin authors who wish to have their plugins listed may file a submission via a
 
 ### Auth
 
+- [Github Actions](https://github.com/Mongey/vault-plugin-auth-github-actions)
 - [Jenkins](https://plugins.jenkins.io/hashicorp-vault-plugin)
 - [Terraform Enterprise/Terraform Cloud](https://github.com/gitrgoliveira/vault-plugin-auth-tfe)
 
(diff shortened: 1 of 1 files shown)


website/content/docs/plugin-portal.mdx
```diff
@@ -122,11 +122,13 @@ Plugin authors who wish to have their plugins listed may file a submission via a
 
 ### Auth
 
+- [Github Actions](https://github.com/Mongey/vault-plugin-auth-github-actions)
```
calvn, line 125: How does this compare to `vault-action`? At a glance, it seems that it's an auth method that's trading in a GH token (along with some other data from the run), but it's also able to fetch secrets from vault once authenticated which is a bit non-standard for an auth method to do.

narayan-iyengar: I believe the `vault-action` auth creds are stored either in the repo or environment variables.
the plugin listed here is a different auth mechanism and uses the assigned GITHUB_TOKEN as a source of identity as opposed to app role or a personal access token which can be long lived

narayan-iyengar: to add...maybe the overlap here is the fetching secrets part?

acahn: @calvn, I've removed the plugin from the PR.  We can chat about it later if we still want to add it in.

hsimon-hashicorp (approved):

PR #11951, opened 2021-06-28
ci: update go_test to match one from ent

samsalisbury:
These files do not need to be different, so making them the same.

### DIFF
--- a/.circleci/config/commands/go_test.yml
+++ b/.circleci/config/commands/go_test.yml
@@ -92,10 +92,6 @@ steps:
         export VAULT_LICENSE_CI="$VAULT_LICENSE"
         VAULT_LICENSE=
 
-        # Download modules so we can copy them to testcontainer. This is easier than
-        # setting up testcontainer to be able to auth to GitHub to download them.
-        go mod download
-
         # Create a docker network for our testcontainer
         if [ $USE_DOCKER == 1 ]; then
           # Despite the fact that we're using a circleci image (thus getting the
--- a/.circleci/config/commands/go_test.yml
+++ b/.circleci/config/commands/go_test.yml
@@ -21,8 +21,9 @@ parameters:
   arch:
     type: string
     # Only supported for use_docker=false, and only other value allowed is 386
-    default: amd64
+    default: amd64 # must be 386 or amd64
 steps:
+  - configure-git
   - run:
       name: Compute test cache key
       command: |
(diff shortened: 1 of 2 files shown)


.circleci/config/commands/go_test.yml
```diff
@@ -103,9 +106,17 @@ steps:
...
           fi
 
+
+
           # Start a docker testcontainer to run the tests in
-          docker run -d -e TEST_DOCKER_NETWORK_ID \
-            -e DOCKER_CERT_PATH -e DOCKER_HOST -e DOCKER_MACHINE_NAME -e DOCKER_TLS_VERIFY -e NO_PROXY \
+          docker run -d \
+            -e TEST_DOCKER_NETWORK_ID \
+            -e GOPRIVATE \
```
ncabatoff, line 114: Feels like this one would fit better with the other "GO" variables in the docker exec below, but I don't suppose it matters.

.circleci/config/commands/go_test.yml
```diff
@@ -21,8 +21,9 @@ parameters:
   arch:
     type: string
     # Only supported for use_docker=false, and only other value allowed is 386
-    default: amd64
+    default: amd64 # must be 386 or amd64
 steps:
+  - configure-git
```
ncabatoff, line 26: I assume this will simply do nothing useful on OSS?

ncabatoff: I mean with respect to the add_ssh_keys

ncabatoff (approved):

.circleci/config/commands/go_test.yml
samsalisbury: Right, it's not _needed_ for OSS, but having these files be the same seems easier to manage for now..

ncabatoff (approved):

.circleci/config/commands/go_test.yml
samsalisbury: And yes, the referenced SSH key doesn't exist on OSS, so that part is just ignored.

.circleci/config/commands/go_test.yml
samsalisbury: True.. For now I would like to merge as-is just to get these files to match, then changes can flow OSS->ENT like normal. If this gets confusing, we could try to isolate the ENT-specific portions and use an _ent.yml suffixed file for the non-OSS parts, but it seems like it's worth trying it this way round first as it's simpler.

samsalisbury: Addressed in https://github.com/hashicorp/vault/pull/11952

Issue #3792, opened 2018-01-15
[improvement] Add prefix_filter option for telemetry


stevenscg:
Consul has the "prefix_filter" option to allow filtering of the telemetry, which is great for controlling costs and noise.

We use a hosted metrics provider that bills on the number of unique metrics streams submitted.

There are some telemetry items like `vault.route.*` that we don't use right now and it would be ideal to not have to pay for those until or unless we need them.

pmmukh:
This change has been added in https://github.com/hashicorp/vault/pull/12025

PR #11954, opened 2021-06-28
Augment lease revocation metrics

pmmukh:
- Adds a gauge metric to track number of irrevocable leases present in the system
- Augments expiration tests for checking the new counter


helper/fairshare/jobmanager.go
```diff
@@ -243,6 +243,9 @@ func (j *JobManager) incrementWorkerCount(queueID string) {
 	defer j.l.Unlock()
 
 	j.workerCount[queueID]++
+	if j.metricSink != nil {
+		j.metricSink.IncrCounterWithLabels([]string{"expire", "queue_workers"}, 1, []metrics.Label{{"queue_id", queueID}})
```
swayne275, line 247: Since the fair sharing layer is currently general-purpose we probably shouldn't be assuming it's always used for expiration (as this metric does). What does that mean? One option would be to keep it in line with other fair share metrics (`...[]string{"job_manager", "workers"}...`) and then update documentation on how to get the data in the case of expiration. Thoughts @briankassouf ?

vault/expiration.go
```diff
@@ -118,6 +118,10 @@ type ExpirationManager struct {
 	// manual intervention). We retain a subset of the lease info in memory
 	irrevocable sync.Map
 
+	// 	Track count for metrics reporting
+	// 	Mutations to this value should be under pendingLock
```
swayne275, line 122:
```suggestion
	// Track count for metrics reporting
	// Mutations to this value should be under pendingLock
```

vault/expiration.go
```diff
@@ -884,8 +889,10 @@ func (m *ExpirationManager) Stop() error {
 	m.uniquePolicies = make(map[string][]string)
 	m.irrevocable.Range(func(key, _ interface{}) bool {
 		m.irrevocable.Delete(key)
+		m.irrevocableLeaseCount--
```
swayne275, line 892: I'm not sure that there's a point to this decrement if we are setting it back to 0 anyway on line 895

vault/expiration.go
```diff
@@ -1812,6 +1820,7 @@ func (m *ExpirationManager) updatePendingInternal(le *leaseEntry) {
 
 		m.removeFromPending(m.quitContext, le.LeaseID, false)
 		m.irrevocable.Store(le.LeaseID, m.inMemoryLeaseInfo(le))
+		m.irrevocableLeaseCount++
```
swayne275, line 1823: we might want to qualify this based on `!leaseInIrrevocable` to avoid potentially over counting.

On that note, we could be slightly more efficient/clear by also qualifying `m.removeFromPending` on `leaseInPending`.

I'm not sure if we should qualify `m.irrevocable.Store()` on `!leaseInIrrevocable` as it might be possible this call is being used to update in-memory data (something to look at, or ignore for now).

These last two points are less important, but it looks like we want to qualify the increment count on if it's being added to the `irrevocable` map

vault/expiration.go
```diff
@@ -2295,16 +2304,19 @@ func (m *ExpirationManager) lookupLeasesByToken(ctx context.Context, te *logical
...
 	m.pendingLock.RLock()
-	num := m.leaseCount
+	allLeases := m.leaseCount
+	irrevocableLeases := m.irrevocableLeaseCount
 	m.pendingLock.RUnlock()
 
-	metrics.SetGauge([]string{"expire", "num_leases"}, float32(num))
+	metrics.SetGauge([]string{"expire", "num_leases"}, float32(allLeases))
+
+	metrics.SetGauge([]string{"expire", "num_irrevocable_leases"}, float32(irrevocableLeases))
```
swayne275, line 2315: we should either change this to `irrevocable_leases` or update VLT-145 (in an Appendix -> Changes bullet point?) to say the metric is `num_irrevocable_leases`. I'm fine with either, but I like having the `num_` prefix since it's more in-line with what we are doing for all leases

vault/expiration_test.go
```diff
@@ -2839,6 +2839,9 @@ func TestExpiration_MarkIrrevocable(t *testing.T) {
 	if _, ok := exp.irrevocable.Load(leaseID); !ok {
 		t.Fatalf("irrevocable lease not included in irrevocable map")
 	}
+	if exp.irrevocableLeaseCount != 1 {
```
swayne275, line 2842: I just noticed this in previous code (line 2831) - we might want to hold pending lock for some of these updates for things that are supposed to be called/accessed locked. I'm not sure why the race tests didn't go off on this earlier?

vault/expiration_test.go
```diff
@@ -3086,6 +3092,9 @@ func TestExpiration_getIrrevocableLeaseCounts(t *testing.T) {
 		t.Fatalf("error getting irrevocable lease counts: %v", err)
 	}
 
+	if exp.irrevocableLeaseCount != len(backends)*expectedPerMount {
```
swayne275, line 3095: do we want `pendingLock` held for this access too?

website/content/docs/internals/telemetry.mdx
```diff
@@ -150,13 +150,15 @@ These metrics cover measurement of token, identity, and lease operations, and co
...
+| `vault.expire.num_irrevocable_leases`                                                           | Number of irrevocable leases which are present in the system                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | leases   | gauge   |
```
swayne275, line 153:
```suggestion
| `vault.expire.num_irrevocable_leases`                                                           | Number of irrevocable leases that cannot be revoked automatically                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | leases   | gauge   |
```

this phrasing might be more clear? if you're ever in doubt about phrasing, Ivana is a great resource!

website/content/docs/internals/telemetry.mdx
```diff
@@ -150,13 +150,15 @@ These metrics cover measurement of token, identity, and lease operations, and co
...
+| `vault.expire.queue_workers` (cluster, queue_id)                                                | Count of expiration workers per auth mount                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          | leases   | counter |
```
swayne275, line 161: if you update the naming be sure to update here. the form for the suggestion I gave would be `vault.expire.job_manager.workers`. I believe what you currently have would lead to `vault.expire.expire.queue_workers` (I think `expire` is prefixed to job manager metrics that are called by the expiration manager

swayne275 (commented):
Looks good! Just a number of comments/suggestions/questions

website/content/docs/internals/telemetry.mdx
swayne275: now that I'm looking at it, I'm not sure that the metric sink is prefixed with `expire` since we're passing in the one from core. so perhaps `vault.expire.job_manager.queue_length` would have actually been `vault.job_manager.queue_length`. Let's look into this.

helper/fairshare/jobmanager.go
pmmukh: ah yes this makes sense, will get the name in line with the other metrics in this file!

vault/expiration.go
pmmukh: Roger that, qualifying the increment makes sense! I'm thinking those other 2 points can be ignored for now, as they would both be no-op functions if the data isn't present in pending / is present in irrevocable maps, so they should be safe without the qualifying checks.

vault/expiration_test.go
pmmukh: hmm i think maybe not ? so in this test, one lease was registered then marked as irrevocable on L2831, so I don't _think_ anything else will be trying to concurrently access this value, so it should be safe to access in this section without the lock. does that make sense ?

vault/expiration_test.go
pmmukh: i'm thinking not based on the same logic as the above, when we're accessing this value here the irrevocable leases have all been added to the system then the counts were fetched, so at this point nothing else should be trying to access this value other than this test.

website/content/docs/internals/telemetry.mdx
pmmukh: ah nice! yeah I'm gonna use this wording, thanks!

website/content/docs/internals/telemetry.mdx
pmmukh: ah so I think the metricSink struct is not prefixed with `expire` itself, but when emitting the metrics this is done by starting the metric name with `j.name` https://github.com/hashicorp/vault/blob/main/helper/fairshare/jobmanager.go#L120 which is `expire` in the case of the jobmanager used by expiration.go.
So once I update this new metric to be on par with those, it should then come out to `vault.expire.job_manager.workers` ( also changing the `queue_workers` back to `workers`, cause externally nobody would understand what this internal queue is for anyways so guess it'd be confusing )

pmmukh: ah looks like I can't make a different comment on this line so just gonna reuse this. I have 2 doubts on this line
1. Should the unit of the metric be `leases` ? I'm thinking it should be `workers`, but that wouldn't mean much for external people, so not sure what a more external-friendly unit should be here.
2. In the labels for this metric, I've used `queue_id`. This could just as easily be `auth_method` or `mount_point` or `mount_id`, based on other words we've used in the telemetry docs, i believe. So I'm not sure if `queue_id` is the right choice here.
Note: The metric `vault.expire.job_manager.queue_length` on L157 seems to be missing a note on the label, as it also uses a `queue_id` label in the code.

ncabatoff: I worry about the cardinality of metrics that have a label based on mounts.  There are customers out there with 1000s of mounts, and most metrics systems don't cope well with cardinality over ~100.  We do have some per-mount metrics, but those I believe are constrained by settings like https://www.vaultproject.io/docs/configuration/telemetry#usage_gauge_period and https://www.vaultproject.io/docs/configuration/telemetry#maximum_gauge_cardinality.

pmmukh: that concern makes a lotta sense! I did think briefly about emitting this metric as a gauge (off of the workerCount map) to reduce the volume of metrics emission, but that ended up missing workers that would do their job and quit in between metric emissions, plus it wouldn't address the cardinality issue. 
So doing this as a gauge would probably limit the effectiveness of this metric. In that case, would it just make more sense to leave this out for now, and only add it in a later release if there is customer demand for it, so the situation is more `can we please have this` versus `why is this blowing up my grafana budget` ?

vault/expiration_test.go
swayne275: i think some good advice i've heard is if you have to justify why something doesn't need concurrent protection in a specific instance, it's best and most clear to just lock it

website/content/docs/internals/telemetry.mdx
swayne275: irrevocable is a bit redundant in this case, but probably nice for a user CTL/CMD-F-ing through the documentation

vault/expiration_test.go
pmmukh: hah that does sound like a very good rule of thumb, gotcha, i'll go ahead and lock up the `irrevocableLeaseCount` accesses in the test!

website/content/docs/internals/telemetry.mdx
pmmukh: Ah yeah I was thinking of getting rid of `irrevocable` here, making it `Number of leases that cannot be revoked automatically`, and since that is in the metric name itself, users should still be able to find it on searching i think.

swayne275: yeah true, sounds good!

website/content/docs/internals/telemetry.mdx
pmmukh: Fixed up the metric name here 2558e12, although the question around whether we should have this metric, and the couple doubts on language are still open.

website/content/docs/internals/telemetry.mdx
pmmukh: 2558e12

vault/expiration_test.go
pmmukh: fixed in 2558e12

vault/expiration_test.go
pmmukh: 2558e12

vault/expiration.go
pmmukh: Updated the RFC with this metric name

vault/expiration.go
pmmukh: 2558e12

vault/expiration.go
pmmukh: 2558e12

vault/expiration.go
pmmukh: 2558e12

helper/fairshare/jobmanager.go
pmmukh: 2558e12

vault/expiration_test.go
swayne275: cool. while you're in here, could you also lock around the stuff that was on line 2831 (if i remember right i must have forgotten to do that before calling a function that should be called under lock)

feel free to do this anywhere else you see that might be relevant!

pmmukh: ah gotcha, can do!

pmmukh: fixed this and a couple other tests here e6eda3d

website/content/docs/internals/telemetry.mdx
pmmukh: Aha, did not notice that, good spot!
Roger that, will go ahead with this for now! Btw, I'm gonna be working on https://github.com/hashicorp/vault/issues/3792 soon as my current plate empties a little, think this might be helpful for customers to just turn off metrics if they aren't gaining value from them.

website/content/docs/internals/telemetry.mdx
```diff
@@ -150,10 +150,12 @@ These metrics cover measurement of token, identity, and lease operations, and co
...
+| `vault.expire.job_manager.workers` (cluster, queue_id)                                          | Count of expiration workers per auth mount                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          | leases   | counter |
```
pmmukh, line 158: couple things to point out for reviewers
1. Is `queue_id` the correct term to use here? ( as opposed to `mount_point`, `auth_method` or `mount_id` )
2. Is `leases` the correct unit ? Technically it should be workers, but that won't mean much externally, so not sure.

pmmukh:
Based on the discussion with the team, will be removing the expiration workers per mount metric, as it has high cardinality, and not necessarily a lot of value beyond what is already provided by the revocation jobs per mount metric.

website/content/docs/internals/telemetry.mdx
pmmukh: removed in 1df235f

vault/expiration.go
```diff
@@ -1018,6 +1024,7 @@ func (m *ExpirationManager) revokeCommon(ctx context.Context, leaseID string, fo
 	m.removeFromPending(ctx, leaseID, true)
 	m.nonexpiring.Delete(leaseID)
 	m.irrevocable.Delete(leaseID)
+	m.irrevocableLeaseCount--
```
swayne275, line 1027: could this lead to under-counting? there are likely scenarios where we hit this code block, but the lease doesn't exist in the `irrevocable` map (the above call would just be a noop)

vault/expiration.go
```diff
@@ -1812,6 +1819,11 @@ func (m *ExpirationManager) updatePendingInternal(le *leaseEntry) {
 
 		m.removeFromPending(m.quitContext, le.LeaseID, false)
 		m.irrevocable.Store(le.LeaseID, m.inMemoryLeaseInfo(le))
+
+		// Only increment this counter if the lease was not in the irrevocable map before
```
swayne275, line 1823:
```suggestion
```

vault/expiration.go
```diff
@@ -2472,6 +2487,7 @@ func (m *ExpirationManager) markLeaseIrrevocable(ctx context.Context, le *leaseE
 	m.persistEntry(ctx, le)
 
 	m.irrevocable.Store(le.LeaseID, m.inMemoryLeaseInfo(le))
+	m.irrevocableLeaseCount++
```
swayne275, line 2490: can you update the comment on line 2463 ("...no total lease count metrics/...") to be accurate given the new metrics?

vault/expiration_test.go
```diff
@@ -2828,7 +2828,11 @@ func TestExpiration_MarkIrrevocable(t *testing.T) {
 	}
 
 	irrevocableErr := fmt.Errorf("test irrevocable error")
+
+	exp.pendingLock.Lock()
 	exp.markLeaseIrrevocable(ctx, loadedLE, irrevocableErr)
+	irrevocableLeaseCount := exp.irrevocableLeaseCount
```
swayne275, line 2834: since this is a test and we aren't overly-concerned about performance, it's probably fine to declare `irrevocableLeaseCount` closer to where it's used and just lock/unlock twice

vault/expiration_test.go
```diff
@@ -2943,6 +2956,14 @@ func TestExpiration_StopClearsIrrevocableCache(t *testing.T) {
 	if _, ok := exp.irrevocable.Load(leaseID); ok {
 		t.Error("expiration manager irrevocable cache should be cleared on stop")
 	}
+
+	exp.pendingLock.Lock()
```
swayne275, line 2960: all of these reads can just be under `RLock()` instead of `Lock()` (and then `RUnlock()`)

swayne275 (commented):
I think it's pretty close. we want to ensure that irrevocable lease counter increments/decrements only happen when a lease is being added or removed from the `irrevocable` map (and not in the case of noops)

vault/expiration.go
pmmukh: ah this would definitely lead to undercounting!!

pmmukh: df2931a

vault/expiration.go
pmmukh: df2931a

vault/expiration.go
pmmukh: donzo df2931a

vault/expiration_test.go
pmmukh: makes sense, changed here df2931a

vault/expiration_test.go
pmmukh: ah yeah good catch! df2931a

pmmukh:
> at this point, you might as qualify this on lease presence too?

updated here b345922

vault/expiration.go
```diff
@@ -118,6 +118,10 @@ type ExpirationManager struct {
 	// manual intervention). We retain a subset of the lease info in memory
 	irrevocable sync.Map
 
+	// Track count for metrics reporting
+	// Mutations to this value should be under pendingLock
```
swayne275, line 122:
```suggestion
	// This value is protected by pendingLock
```

vault/expiration.go
```diff
@@ -1812,6 +1822,10 @@ func (m *ExpirationManager) updatePendingInternal(le *leaseEntry) {
 
 		m.removeFromPending(m.quitContext, le.LeaseID, false)
 		m.irrevocable.Store(le.LeaseID, m.inMemoryLeaseInfo(le))
+
+		if !leaseInIrrevocable {
```
swayne275, line 1826: might be overly-cautious, but it might be good to note that we care about whether the lease was in the irrevocable map *before* adding it a couple lines above. This could be done either with a comment, or by calling this code just before line 1824

swayne275 (commented):
very close! just a couple clarifications

vault/expiration.go
pmmukh: i think it probably makes sense to have the increment here, after the store succeeds, will add a comment to clarify!

pmmukh: accfda0

vault/expiration.go
pmmukh: accfda0

swayne275 (approved):

PR #11963, opened 2021-06-29
Handle form validation for open api form

arnav28:
- Added required validator for all the default fields

### DIFF
--- a/ui/app/components/generated-item.js
+++ b/ui/app/components/generated-item.js
@@ -46,19 +46,35 @@ export default Component.extend({
   init() {
     this._super(...arguments);
     this.set('validationMessages', {});
+    if (this.mode === 'edit') {
+      // For validation to work in edit mode,
+      // reconstruct the model values from field group
+      this.model.fieldGroups.forEach(element => {
+        if (element.default) {
+          element.default.forEach(attr => {
+            let fieldValue = attr.options && attr.options.fieldValue;
+            if (fieldValue) {
+              this.model[attr.name] = this.model[fieldValue];
+            }
+          });
+        }
+      });
+    }
   },
   actions: {
     onKeyUp(name, value) {
       this.model.set(name, value);
-      // Set validation error message for updated attribute
-      this.model.validations.attrs[name] && this.model.validations.attrs[name].isValid
-        ? set(this.validationMessages, name, '')
-        : set(this.validationMessages, name, this.model.validations.attrs[name].message);
+      if (this.model.validations) {
+        // Set validation error message for updated attribute
+        this.model.validations.attrs[name] && this.model.validations.attrs[name].isValid
+          ? set(this.validationMessages, name, '')
+          : set(this.validationMessages, name, this.model.validations.attrs[name].message);
 
-      // Set form button state
-      this.model.validate().then(({ validations }) => {
-        this.set('isFormInvalid', !validations.isValid);
-      });
+        // Set form button state
+        this.model.validate().then(({ validations }) => {
+          this.set('isFormInvalid', !validations.isValid);
+        });
+      }
     },
     deleteItem() {
       this.model.destroyRecord().then(() => {
--- a/ui/app/components/generated-item.js
+++ b/ui/app/components/generated-item.js
@@ -1,7 +1,7 @@
 import AdapterError from '@ember-data/adapter/error';
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
-import { computed } from '@ember/object';
+import { computed, set } from '@ember/object';
 import { task } from 'ember-concurrency';
 
 /**
@@ -24,6 +24,8 @@ export default Component.extend({
   itemType: null,
   flashMessages: service(),
   router: service(),
+  validationMessages: null,
+  isFormInvalid: true,
   props: computed('model', function() {
     return this.model.serialize();
   }),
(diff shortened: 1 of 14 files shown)


ui/app/components/generated-item.js
```diff
@@ -41,7 +43,23 @@ export default Component.extend({
...
+  },
   actions: {
+    onKeyUp(name, value) {
+      this.model.set(name, value);
+      // Set validation error message for updated attribute
+      this.model.validations.attrs[name] && this.model.validations.attrs[name].isValid
+        ? set(this.validationMessages, name, '')
+        : set(this.validationMessages, name, this.model.validations.attrs[name].message);
+
+      // Set form button state
```
Monkeychip, line 58: I like this!

ui/app/components/mount-backend-form.js
```diff
@@ -108,6 +112,10 @@ export default Component.extend({
       this.mountModel.validations.attrs.path.isValid
         ? set(this.validationMessages, 'path', '')
         : set(this.validationMessages, 'path', this.mountModel.validations.attrs.path.message);
+
+      this.mountModel.validate().then(({ validations }) => {
```
Monkeychip, line 116: thank you for pushing this pattern forward here.

ui/app/models/auth-method.js
```diff
@@ -4,11 +4,19 @@ import { computed } from '@ember/object';
...
+import { validator, buildValidations } from 'ember-cp-validations';
 
 import apiPath from 'vault/utils/api-path';
 import attachCapabilities from 'vault/lib/attach-capabilities';
 
-let ModelExport = Model.extend({
+const Validations = buildValidations({
+  path: validator('presence', {
+    presence: true,
+    message: "Path can't be blank",
```
Monkeychip, line 15: I ran into this issue, but I think there should be a period at the end of the validation error messages.  Might be worth checking with design.

ui/app/services/path-help.js
```diff
@@ -291,6 +292,11 @@ export default Service.extend({
       }
       newModel.reopenClass({ merged: true });
       owner.unregister(modelName);
+
+      if (newModel.proto().fieldGroups) {
```
Monkeychip, line 296: can you add some comments here explaining how this is used?  I just imagine running into this file later and trying to understand this conditional.

ui/app/styles/core/message.scss
```diff
@@ -106,6 +106,7 @@
 
 .message-inline {
   display: flex;
+  align-items: center;
```
Monkeychip, line 109: is this going to be true for all message-inline?  worth checking that this doesn't have consequences elsewhere.

ui/app/templates/components/generated-item.hbs
```diff
@@ -52,12 +52,12 @@
     <div class="box is-sideless is-fullwidth is-marginless">
       <NamespaceReminder @mode="save" @noun={{itemType}} />
       <MessageError @model={{model}} />
-      <FormFieldGroups @model={{model}} @mode={{mode}} />
+      <FormFieldGroups @model={{model}} @mode={{mode}} @onKeyUp={{action "onKeyUp"}} @validationMessages={{validationMessages}}/>
```
Monkeychip, line 55: Something Chelsea has been good at pointing out to me in my PRs, but whenever we add properties to a component like this, document the properties in the JSDOCS on the component itself.  So on `form-field-groups.js` add the onKeyUp and validationMessages to the documentation snippet below.

```
 * <FormFieldGroups
 *  @model={{mountModel}}
 *  @onChange={{action "onTypeChange"}}
 *  @renderGroup="Method Options"
 * />
 * ```
 *
 * @param [renderGroup=null] {String} - An allow list of groups to include in the render.
 * @param model=null {DS.Model} - Model to be passed down to form-field component. If `fieldGroups` is present on the model then it will be iterated over and groups of `FormField` components will be rendered.
 * @param onChange=null {Func} - Handler that will get set on the `FormField` component.
 *
 */
```

ui/app/utils/build-api-validators.js
```diff
@@ -0,0 +1,20 @@
+import { validator, buildValidations } from 'ember-cp-validations';
```
Monkeychip, line 1: Can you add some comments explaining what this util does?  A good example of this is in the uitls/api-path.js file.  Something that would help a future reader understand its purpose and usage.

ui/lib/core/addon/templates/components/form-field.hbs
```diff
@@ -191,7 +191,18 @@
...
     @onChange={{action (action "setAndBroadcast" valuePath)}}
+    onkeyup={{action
+      (action "handleKeyUp" attr.name)
+      value="target.value"
+    }}
   />
+  {{#if (get validationMessages attr.name)}}
+    <AlertInline 
+      @type="danger" 
+      @message={{get validationMessages attr.name}}
```
Monkeychip, line 202: nice!

Monkeychip (changes requested):
This looks really good.  A couple of comments, and the following requests:
- Can you add a gif to the PR that shows the validations and how they work?
- Can you add test coverage for this particular validation? You can look at my earlier PR that does this.  [See file here.](https://github.com/hashicorp/vault/pull/11878/files#diff-589ec28acb244120e7f83605b6a6558ad5d8e3155585f7d543854855d0d08700)

arnav28:
https://user-images.githubusercontent.com/2932708/124649632-9941b380-de4d-11eb-9a16-ff7739e559b8.mov

ui/app/utils/build-api-validators.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Add validation on dynamic form fields generated via open api spec
+ * For fields grouped under default category, add the require/presence validator
+ * @param {Array} fieldGroups
```
hashishaw, line 6: Can we get a little more description here about what shape the objects in the array should be in? Or where to find it?

ui/lib/core/addon/components/form-field.js
```diff
@@ -93,6 +95,11 @@ export default Component.extend({
    */
   valuePath: or('attr.options.fieldValue', 'attr.name'),
 
+  isReadOnly: computed('attr.options.readOnly', 'mode', function() {
+    let readonly = this.attr.options ? this.attr.options.readOnly : false;
```
hashishaw, line 99: If you want, this could be re-written as `let readonly = this.attr.options?.readOnly || false`

Monkeychip: Nice!

ui/app/utils/build-api-validators.js
arnav28: Good point. Added example for fieldgroup structure.

Monkeychip (approved):

Issue #11968, opened 2021-06-30
Vault Agent: make auto-delete of JWT file a configurable option


tdsacilowski:
**Is your feature request related to a problem? Please describe.**
It's possible to use [Kubernetes bound service account tokens](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md) with Vault's JWT auth method. However, the [service account token projected volume](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection) is mounted as read-only. Vault agent's JWT auto-auth method will [attempt to delete the JWT ](https://www.vaultproject.io/docs/agent/autoauth/methods/jwt) after reading it. Because of this, container logs will be flooded with the following every 500ms:

```
...
2021-06-30T21:16:56.654Z [ERROR] auth.jwt: error removing jwt file: error="remove /run/secrets/tokens/sa-token: read-only file system"
2021-06-30T21:16:57.153Z [ERROR] auth.jwt: error removing jwt file: error="remove /run/secrets/tokens/sa-token: read-only file system"
2021-06-30T21:16:57.653Z [ERROR] auth.jwt: error removing jwt file: error="remove /run/secrets/tokens/sa-token: read-only file system"
2021-06-30T21:16:58.154Z [ERROR] auth.jwt: error removing jwt file: error="remove /run/secrets/tokens/sa-token: read-only file system"
2021-06-30T21:16:58.654Z [ERROR] auth.jwt: error removing jwt file: error="remove /run/secrets/tokens/sa-token: read-only file system"
2021-06-30T21:16:59.153Z [ERROR] auth.jwt: error removing jwt file: error="remove /run/secrets/tokens/sa-token: read-only file system"
...
```

**Describe the solution you'd like**
It would be helpful to make auto-deletion of the JWT a configurable option similar to how the AppRole auto-auth method [handles the secret-id file](https://www.vaultproject.io/docs/agent/autoauth/methods/approle#remove_secret_id_file_after_reading).

**Describe alternatives you've considered**
No alternative approach is possible. These types of [volumes are mounted as read-only](https://github.com/kubernetes/kubernetes/pull/58720) and can not be altered.

**Explain any additional use-cases**
N/A

**Additional context**
N/A

PR #11969, opened 2021-06-30
First pass at JWT auto-deletion flag

tdsacilowski:
Fixes https://github.com/hashicorp/vault/issues/11968

Still need to write tests.


calvn:
Are there potential (re-)authentication issues that could arise if we don't delete a stale JWT?

website/content/docs/agent/autoauth/methods/jwt.mdx
```diff
@@ -17,3 +17,7 @@ JWT to perform a reauthentication.
 - `path` `(string: required)` - The path to the JWT file
 
 - `role` `(string: required)` - The role to authenticate against on Vault
+
+- `remove_jwt_after_reading` `(bool: optional, defaults to true)` -
```
calvn, line 21: Can we flip this around and have it be `keep_after_reading` or something along those lines? This way the default value is also the zero value (`false`).

tdsacilowski: I thought about that but I thought it more prudent to retain the current default behavior and opt-out if needed. I'd assume that in _most_ cases, the default behavior is more appropriate. This change is mostly to allow for using Kubernetes bound service account tokens with JWT auth. In this use case, Kubernetes automatically rotates the JWT based on a configurable TTL.

calvn: `keep_after_reading` with it being defaulted `false` should still retain the current behavior though. Note that we're flipping the naming and also the default so keeping the files would still be an opt-in. We'd only _not_ delete the file if this gets set to `true` explicitly.

tdsacilowski:
> Are there potential (re-)authentication issues that could arise if we don't delete a stale JWT?

I wouldn't think so. JWTs _should_ have an expiration set so if re-auth is attempted with an expired JWT auth should fail. In this scenario, the underlying system that is responsible for delivering an updated JWT (e.g. Kubernetes) should have provided a newer JWT. But I may be missing some of the finer points here.

website/content/docs/agent/autoauth/methods/jwt.mdx
tdsacilowski: Ah ok, I see what you mean. I was following the same model that the AppRole auto-auth was using but I can change this around if it makes more sense.

calvn:
It looks like `ingressToken()` [relies on the absence](https://github.com/hashicorp/vault/pull/11969/files#diff-91e3ede29e58c49d358b61b7c7b14faa3585be554040c0fb1a6f4603d7212b7aR170-R172) on the file to short-circuit from the function, so on non-kubernetes workflows in which someone opts in for this, I could see Wgent doing the opposite -- reading the file every 500ms. We do end up comparing the value of this token with the existing one so it's mostly benign, but it does result in more filesystem reads.

aelbarkani:
@arianvp since this PR was not merged, did you find a workaround ?

website/content/docs/agent/autoauth/methods/jwt.mdx
tomhjp: I actually like `remove_jwt_after_reading` for its consistency with AppRole's option: https://www.vaultproject.io/docs/agent/autoauth/methods/approle#remove_secret_id_file_after_reading

IMO the bar to break from this pattern needs to be higher than making it terser. If it wasn't for the prior art, I would 100% agree with both the flip + naming comments above, but they're minor nits and they're secondary to preventing confusion for consumers.

command/agent/auth/jwt/jwt.go
```diff
@@ -73,6 +76,15 @@ func NewJWTAuthMethod(conf *auth.AuthConfig) (auth.AuthMethod, error) {
...
 	}
 
+	removeJWTAfterReadingRaw, ok := conf.Config["remove_jwt_after_reading"]
+	if ok {
+		removeJWTAfterReading, err := parseutil.ParseBool(removeJWTAfterReadingRaw)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing 'remove_jwt_after_reading' value: %w", err)
+		}
+		j.removeJWTAfterReading = removeJWTAfterReading
+	}
```
tomhjp, line 86:
```suggestion
	if removeJWTAfterReadingRaw, ok := conf.Config["remove_jwt_after_reading"]; ok {
		removeJWTAfterReading, err := parseutil.ParseBool(removeJWTAfterReadingRaw)
		if err != nil {
			return nil, fmt.Errorf("error parsing 'remove_jwt_after_reading' value: %w", err)
		}
		j.removeJWTAfterReading = removeJWTAfterReading
	} else {
		j.removeJWTAfterReading = true
	}
```

tomhjp: Seems like the diff doesn't display very well, but the main comment is to default the value to true to maintain existing behaviour.

tomhjp (commented):
Taking another look as there has been some renewed interest in this. @tdsacilowski I'm happy to take this on if you'd prefer not to pick this up again.

command/agent/auth/jwt/jwt.go
```diff
@@ -73,14 +75,32 @@ func NewJWTAuthMethod(conf *auth.AuthConfig) (auth.AuthMethod, error) {
...
 
+	if removeJWTAfterReadingRaw, ok := conf.Config["remove_jwt_after_reading"]; ok {
+		removeJWTAfterReading, err := parseutil.ParseBool(removeJWTAfterReadingRaw)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing 'remove_jwt_after_reading' value: %w", err)
+		}
+		j.removeJWTAfterReading = removeJWTAfterReading
+	} else {
+		// Default to true for maximum backwards compatibility.
+		j.removeJWTAfterReading = true
```
tvoran, line 86: It looks like this is defaulting to `true` in the initial declaration of `j`. Do we need it both places?

tvoran (approved):
I'm ambivalent about the option name.

command/agent/auth/jwt/jwt.go
tomhjp: Oops, good spot. I lost that in the formatting diff noise.

PR #11977, opened 2021-07-01
Vault-2713: Retry Irrevocable Lease Revocation Periodically

swayne275:


vault/expiration.go
```diff
@@ -943,6 +956,27 @@ func (m *ExpirationManager) lazyRevokeInternal(ctx context.Context, leaseID stri
...
 }
 
+// TODO should context passed to revokeCommon contain the lease namespace?
+// should be run on a schedule. something like once a day, maybe once a week
+func (m *ExpirationManager) attemptIrrevocableLeasesRevoke() {
+	m.irrevocable.Range(func(k, v interface{}) bool {
+		leaseID := k.(string)
+		le := v.(*leaseEntry)
+
+		if le.ExpireTime.Add(7 * 24 * time.Hour).Before(time.Now()) {
```
briankassouf, line 966: I think this could be reduced quite a bit, what about just an hour ago?

vault/expiration.go
```diff
@@ -943,6 +956,27 @@ func (m *ExpirationManager) lazyRevokeInternal(ctx context.Context, leaseID stri
...
+			// if the lease has expired at least a week ago, try again
+			if err := m.revokeCommon(context.Background(), leaseID, false, false); err != nil {
```
briankassouf, line 968: Can we derive a context from a.activeContext and add a timeout here?

sgmiller (approved):
Agree with Brian's recs but otherwise lgtm.

vault/expiration.go
swayne275: how much timeout would you recommend? 1 minute?

swayne275: 6197b3c04

let me know what you think about the 1 minute timeout

vault/expiration.go
swayne275: 6197b3c04

briankassouf (approved):

vault/expiration.go
sgmiller: I think 1m is fine since this is the retry code.

PR #11998, opened 2021-07-06
UI: Exit test process if vault fails to initialize

hashishaw:
This is a developer ease-of-life update which has the test script exit with an error code if Vault does not start up when the tests start running (for example, if the vault binary is enterprise but you don't have a license). This avoids UI tests falsely failing due to API calls failing.


Monkeychip (approved):
Nice! Thanks for tackling this.

PR #12017, opened 2021-07-08
sdk/logical/request: fix declaration of operation constants

alexey-igrychev:


ncabatoff:
Hi @alexey-igrychev,

Thank you for the pull request.  As far as I can tell, this is purely a style change:  the syntax was valid Go before, and you're changing it to be different but also valid Go.  Since tastes vary, we do not tend to accept purely stylistic changes.  Please let us know if we've missed something here.

PR #12022, opened 2021-07-08
Update node to latest stable version

arnav28:
- v10 has reached EOL so upgrading node to v14 which is the latest
stable build


Monkeychip (approved):
Would you mind amending your PR description to include the steps you took to update? That way in the future, folks will be able to reference this PR (Seeing the history change on the node version in the UI like we did with Matthew's PR) and have some documentation on the steps we took?

PR #12025, opened 2021-07-08
[VAULT-2776] Add prefix_filter option to Vault

pmmukh:
- Adds a `prefix_filter` option to the telemetry configs, that can explicitly allow or block certain metric prefixes.
- Adds a `filter_default` option to the telemetry configs, which is required in conjunction with the config above, to decide whether all metrics are allowed or denied by default. This is currently always true, but users may want to control this in their vault config along with `prefix_filter`.
- Updates documentation to cover both configs

Note that this attempts to maintain parity with this config in consul https://www.consul.io/docs/agent/options#telemetry-prefix_filter
Also, the logic for parsing the prefixFilter config is pretty much lifted from here https://github.com/hashicorp/consul/blob/main/agent/config/builder.go#L640-L654

Related GH issue: https://github.com/hashicorp/vault/issues/3792

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,7 @@ Vagrantfile
 # Configs
 *.hcl
 !command/agent/config/test-fixtures/*.hcl
-!command/server/test-fixtures/*/*.hcl
+!command/server/test-fixtures/**/*.hcl
 
 
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,7 @@ Vagrantfile
 # Configs
 *.hcl
 !command/agent/config/test-fixtures/*.hcl
-!command/server/test-fixtures/*.hcl
+!command/server/test-fixtures/*/*.hcl
 
 
 .DS_Store
--- a/command/server/config_telemetry_test.go
+++ b/command/server/config_telemetry_test.go
@@ -1,0 +1,41 @@
+package server
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMetricFilterConfigs(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		configFile            string
+		expectedFilterDefault *bool
+		expectedPrefixFilter  []string
+	}{
+		{
+			"./test-fixtures/telemetry/valid_prefix_filter.hcl",
+			nil,
+			[]string{"-vault.expire", "-vault.audit", "+vault.expire.num_irrevocable_leases"},
+		},
+		{
+			"./test-fixtures/telemetry/filter_default_override.hcl",
+			boolPointer(false),
+			[]string(nil),
+		},
+	}
+	t.Run("validate metric filter configs", func(t *testing.T) {
+		t.Parallel()
+
+		for _, tc := range cases {
+			config, err := LoadConfigFile(tc.configFile)
+
+			if err != nil {
+				t.Fatalf("Error encountered when loading config %+v", err)
+			}
+
+			assert.Equal(t, tc.expectedFilterDefault, config.SharedConfig.Telemetry.FilterDefault)
+			assert.Equal(t, tc.expectedPrefixFilter, config.SharedConfig.Telemetry.PrefixFilter)
+		}
+	})
+}
--- a/internalshared/configutil/telemetry.go
+++ b/internalshared/configutil/telemetry.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/hashicorp/vault/sdk/helper/parseutil"
+	"log"
 	"time"
+
+	"github.com/hashicorp/vault/sdk/helper/parseutil"
 
 	monitoring "cloud.google.com/go/monitoring/apiv3"
 	"github.com/armon/go-metrics"
@@ -150,6 +152,10 @@ type Telemetry struct {
 
 	// Whether or not telemetry should add labels for namespaces
 	LeaseMetricsNameSpaceLabels bool `hcl:"add_lease_metrics_namespace_labels"`
+
+	// PrefixFilter is a list of filter rules to apply for allowing
+	// or blocking metrics by prefix.
+	PrefixFilter []string `hcl:"prefix_filter"`
 }
 
 func (t *Telemetry) Validate(source string) []ConfigError {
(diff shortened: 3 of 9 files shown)


internalshared/configutil/telemetry.go
```diff
@@ -388,5 +401,29 @@ func SetupTelemetry(opts *SetupTelemetryOpts) (*metrics.InmemSink, *metricsutil.
...
+			log.Print("Cannot have empty filter rule in prefix_filter")
+			continue
+		}
+		switch rule[0] {
+		case '+':
+			telemetryAllowedPrefixes = append(telemetryAllowedPrefixes, rule[1:])
+		case '-':
+			telemetryBlockedPrefixes = append(telemetryBlockedPrefixes, rule[1:])
+		default:
+			log.Printf("Filter rule must begin with either '+' or '-': %q", rule)
```
ncabatoff, line 425: We never use the stdlib `log` in Vault.  I would add an `error` return value to this func instead.

website/content/docs/configuration/telemetry.mdx
```diff
@@ -51,6 +51,17 @@ The following options are available on all telemetry configurations.
 - `add_lease_metrics_namespace_labels` `(bool: false)` - If this value is set to true, then `vault.expire.leases.by_expiration`
   will break down expiring leases by both time and namespace. This parameter is disabled by default because enabling it can lead
   to a large-cardinality metric.
+- `filter_default` - This controls whether to allow metrics that have not been specified by the filter.
```
ncabatoff, line 54: Please add `(bool: false)` to describe this param.

internalshared/configutil/telemetry_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParsePrefixFilters(t *testing.T) {
+	prefixFilters := []string{"", "+vault.abc", "-vault.abc", "vault.abc"}
+
+	allowedPrefixes, blockedPrefixes := parsePrefixFilter(prefixFilters)
+
+	assert.Equal(t, len(allowedPrefixes), 1)
```
ncabatoff, line 14: You can say `assert.Len` to check the length with testify.

internalshared/configutil/telemetry.go
pmmukh: Ah yeah I wasn't sure if I should use `log` here. I think the only reason I didn't do error return, is cause the equivalent consul code logs out a warning when the filter is improperly formatted versus failing https://github.com/hashicorp/consul/blob/main/agent/config/builder.go#L652, so I was looking to maintain that pattern. Maybe `opts.UI.Warn` would make sense https://github.com/hashicorp/vault/blob/main/internalshared/configutil/telemetry.go#L371 ?

ncabatoff:
I know Consul doesn't treat a config error here as a failure and will start up with warnings, but I would prefer we were stricter.  I think it's a bad UX to ignore config errors.

Also, could you add fixture-based test(s) like we do for most other config parsing code?  Look under `command/server` for some examples.

pmmukh:
> I know Consul doesn't treat a config error here as a failure and will start up with warnings, but I would prefer we were stricter. I think it's a bad UX to ignore config errors.
Roger, i would concur with that tbh, will change it to error out.

> Also, could you add fixture-based test(s) like we do for most other config parsing code? Look under `command/server` for some examples.
Will do!

internalshared/configutil/telemetry.go
pmmukh: 6a28af2

website/content/docs/configuration/telemetry.mdx
pmmukh: 6a28af2

internalshared/configutil/telemetry_test.go
pmmukh: I ended up just doing an `assert.Equal` against an expected array, while refactoring the test
6a28af2

pmmukh:
> Also, could you add fixture-based test(s) like we do for most other config parsing code? Look under `command/server` for some examples.

6a28af2

internalshared/configutil/telemetry_test.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+	}
+	t.Run("validate metric filter configs", func(t *testing.T) {
+		t.Parallel()
+
+		for _, tc := range cases {
+
+			allowedPrefixes, blockedPrefixes, err := parsePrefixFilter(tc.inputFilters)
+
+			if err != nil {
```
ncabatoff, line 43: It's fine with the cases you have now, but I consider it best practice when you have an expected error field in your test case type to do something like

if err != nil { compare with expected error }
else { assert expected error empty}

command/server/config_telemetry_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+func TestMetricFilterConfigs(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		configFile            string
+		expectedFilterDefault *bool
+		expectedPrefixFilter  []string
+	}{
+		{
+			"./test-fixtures/telemetry/valid_prefix_filter.hcl",
```
ncabatoff, line 17: Forgot to add fixtures?

pmmukh: hah yep just noticed that I didn't push them, looks like the dir I created them in got gitignored, fixing that

pmmukh: 6c2ad99

internalshared/configutil/telemetry_test.go
pmmukh: ah that makes sense! added that here 6c2ad99

.gitignore
```diff
@@ -48,7 +48,7 @@ Vagrantfile
 # Configs
 *.hcl
 !command/agent/config/test-fixtures/*.hcl
-!command/server/test-fixtures/*.hcl
+!command/server/test-fixtures/*/*.hcl
```
ncabatoff, line 51: Should this be `**/*.hcl` instead of `*/*.hcl`?

ncabatoff (approved):
Looks good, possibly modulo the .gitignore change.

.gitignore
pmmukh: it should! beae728

PR #12032, opened 2021-07-09
Add alias to approle secret IDs

### LINKED ISSUE #11798: Observing flakiness with oidc-token using identity.entity.aliases.<mount accessor>.metadata.<metadata key>
**Describe the bug**

Sometimes custom claims populated via `identity.entity.aliases.<app_role_mount accessor>.metadata.<metadata key>` , where `metadata` is stamped on the `secret-id` of the aprole, don't get populated. There is no pattern on when this happens. Sometimes it happens on a successive curl request to the token endpoint exposed by the oidc-role using the same vault token that returned an oidc token with all claims correctly set. Sometimes, it happens on a brand new token obtained by logging in using approle creds. Once it happens, successive iterations do not populate those claims at all unless a new vault token is obtained.

**To Reproduce**

Cannot describe a deterministic way to repro as this problem doesn't appear to occur in a deterministic manner. Instead will try our best to describe our setup

**Overview**
We have a custom oidc-token template to identify our agent pollers looking for work from a central server.  All pollers use the same approle but each poller has its own secret-id. This secret-id is stamped with a meta data for uniquely identifying the poller. For ex. Poller-A could have poller-a.json as its secret's meta data

```
{
  "metadata": "{ \"resource_id\": \"foobar\" }"
}
``` 
and its secret created and propagated via `curl --header "X-Vault-Token: ..." --request POST --data @poller-a.json https://my-vault-endpoint/my-approle-role-url/secret-id`. Similarly, we have a poller-b, whose metadata could look like

```
{
  "metadata": "{ \"resource_id\": \"barbaz\" }"
}
```

and its secret-id could have been generated by `curl --header "X-Vault-Token: ..." --request POST --data @poller-b.json https://my-vault-endpoint/my-approle-role-url/secret-id`

**Details**

- Login to vault using one of the poller's approles `vault write <my-app-role-path> role_id=... secret_id=...`
- Verify that the vault-token retrieved from above has `token_meta_resource_id    <my-resource-id>`
- Send a curl request to the oidc-role's token endpoint stamped with the vault token from above ` curl  --header "X-Vault-Token: ..." https://<my-vault-endpoint>/v1/identity/oidc/token/<my-role>`
   -  `<My role>` is defined in tf such that there is one per approle mountpoint and we have one approle mountpoint per env stages. In broad strokes, 
   ```
resource "vault_identity_oidc_role" "my-svc" {  
    for_each = local.my-svc-envs
    name     = "my-svc-${each.key}"
    key      = vault_identity_oidc_key.key[each.key].name
    ttl        = my-ttl  
    template = &lt;&lt;EOF
{  
     <other claims>
    "entity_data": { 
        <other claims>   
        "resource_id": {{identity.entity.aliases.${vault_auth_backend.app[each.key].accessor}.metadata.resource_id}}  }}  
EOF}
  ```
- Now decrypt the received oidc-token via `jwt decode <my-token>`

Repeat the above test in various flavors

- Round robin across poller-A/poller-B vault approles.
- Repeat the `curl` for fetching the oidc-token and the `jwt decode` for the same token **which had previously returned a valid resource_id**
Additionally, performed this test as a golang program which runs all steps above using jose library for decoding the oidc and creating vault clients from scratch every 30 seconds.

**The manual tests showed quite a few instances of resource_id not getting populated and there was no pattern to when it occured. Sometimes it would occur for a token for which resource_id previously got populated. Sometimes it occurred for a brand new token**

**golang tests are run for just one of the pollers in tight loop and every 30 seconds. When run on a tight loop, we saw 3 failures out of 4000 odd. 30 second interval test is ongoing. No failures so far after 36 iterations.**

**Expected behavior**
Expect the `resource_id` to be set every time

**Environment:**
* 
```
<redacted>@<redacted> ~ % vault status
Key                    Value
---                    -----
Seal Type              shamir
Initialized            true
Sealed                 false
Total Shares           5
Threshold              3
Version                1.6.2
Storage Type           postgresql
Cluster Name           vault-cluster-140e09ca
Cluster ID             62dba0b3-dedb-1844-1a16-e873f9cf4847
HA Enabled             true
HA Cluster             https://vault.vault:8201
HA Mode                standby
Active Node Address    https://vault.vault:8200
```

```
sganapathy@C02DM0ZTMD6R ~ % vault version
Vault v1.6.0 ('7ce0bd9691998e0443bc77e98b1e2a4ab1e965d4+CHANGES')
```

peterverraedt:
When using a single approle role_id with multiple secret_ids, each
carrying different metadata, the latest authenticated secret id will
define the identity alias for that role_id. This means that all earlier
obtained tokens for the same role_id will use the metadata of the latest
created token in templating (e.g. policy templating), breaking the
scenario of a single role_id with multiple secret_ids to hand out to
different clients that need different permissions.

We fix this by adding an alias field to secret_id, which should contain
a unique value for each "end client", and which is used to define the
identy alias. I.e. in general secret IDs with different metadata should
have different aliases. Other options would be to use the secret
accessor id as identity alias, or to put simply a hash of the metadata,
but we opted for this solution to have a limit on the entries in the
identity alias table - we would flood it over time if secret IDs change
rapidly or a metadata field like a timestamp would be added.

Fixes #11798

Signed-off-by: Peter Verraedt <peter@verraedt.be>

### DIFF
--- a/builtin/credential/approle/path_login.go
+++ b/builtin/credential/approle/path_login.go
@@ -45,10 +45,72 @@ func (b *backend) pathLoginUpdateAliasLookahead(ctx context.Context, req *logica
 		return nil, fmt.Errorf("missing role_id")
 	}
 
+	// Look for the storage entry that maps the roleID to role
+	roleIDIndex, err := b.roleIDEntry(ctx, req.Storage, roleID)
+	if err != nil {
+		return nil, err
+	}
+	if roleIDIndex == nil {
+		return logical.ErrorResponse("invalid role ID"), nil
+	}
+
+	roleName := roleIDIndex.Name
+
+	roleLock := b.roleLock(roleName)
+	roleLock.RLock()
+
+	role, err := b.roleEntry(ctx, req.Storage, roleName)
+	roleLock.RUnlock()
+	if err != nil {
+		return nil, err
+	}
+	if role == nil {
+		return logical.ErrorResponse("invalid role ID"), nil
+	}
+
+	aliasName := roleID
+
+	if role.BindSecretID {
+		secretID := strings.TrimSpace(data.Get("secret_id").(string))
+		if secretID == "" {
+			return logical.ErrorResponse("missing secret_id"), nil
+		}
+
+		secretIDHMAC, err := createHMAC(role.HMACKey, secretID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create HMAC of secret_id: %w", err)
+		}
+
+		roleNameHMAC, err := createHMAC(role.HMACKey, role.name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create HMAC of role_name: %w", err)
+		}
+
+		secretIDLock := b.secretIDLock(secretIDHMAC)
+		secretIDLock.RLock()
+
+		unlockFunc := secretIDLock.RUnlock
+		defer func() {
+			unlockFunc()
+		}()
+
+		entry, err := b.nonLockedSecretIDStorageEntry(ctx, req.Storage, role.SecretIDPrefix, roleNameHMAC, secretIDHMAC)
+		if err != nil {
+			return nil, err
+		}
+		if entry == nil {
+			return logical.ErrorResponse("invalid secret id"), nil
+		}
+
+		if entry.Alias != "" {
+			aliasName += ":" + entry.Alias
+		}
+	}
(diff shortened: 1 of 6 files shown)


pmmukh:
Closing out this PR based on the discussion here https://github.com/hashicorp/vault/issues/11798#issuecomment-920369995

PR #12042, opened 2021-07-12
[VAULT-2825] Fix erroneous 500 resp for field validation errors

### LINKED ISSUE #11645: Field validation on invalid data from client returns status 500
**Describe the bug**
When the client sends invalid data, the server responds with status 500 instead of 400.

**To Reproduce**
1. Run `vault write secret/config max_versions=abcd`

**Expected behavior**
Return status 400

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.7.1
* Vault CLI Version (retrieve with `vault version`): 1.7.1
* Server Operating System/Architecture: https://hub.docker.com/layers/vault/library/vault/latest/images/sha256-7d07d64480897215cca86ce51750c888fae4d703529708517ecdd5fabba6cbb3

**Additional context**
The issue appears to come from here:
https://github.com/hashicorp/vault/blob/v1.7.1/sdk/framework/backend.go#L266-L268

Note, this is not restricted to just `secret/config`, but any and all field which have a schema defined on them.

ccapurso:
The server will currently respond with a 500 if an endpoint for which a schema is defined is provided with invalid data. The proposed fix is to instead respond with a 400 with an appropriate error message stating that field validation has failed.

Fixes #11645

### DIFF
--- a/sdk/framework/backend_test.go
+++ b/sdk/framework/backend_test.go
@@ -249,7 +249,7 @@ func TestBackendHandleRequest_badwrite(t *testing.T) {
 		t.Fatalf("err: %s", err)
 	}
 
-	if resp.Data["error"] != "Field validation failed: error converting input 3false3 for field \"value\": cannot parse '' as bool: strconv.ParseBool: parsing \"3false3\": invalid syntax" {
+	if !strings.Contains(resp.Data["error"].(string), "Field validation failed") {
 		t.Fatalf("bad: %#v", resp)
 	}
 }
--- a/sdk/framework/backend_test.go
+++ b/sdk/framework/backend_test.go
@@ -240,14 +240,17 @@ func TestBackendHandleRequest_badwrite(t *testing.T) {
 		},
 	}
 
-	_, err := b.HandleRequest(context.Background(), &logical.Request{
+	resp, err := b.HandleRequest(context.Background(), &logical.Request{
 		Operation: logical.UpdateOperation,
 		Path:      "foo/bar",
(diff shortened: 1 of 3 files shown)


sdk/framework/backend_test.go
```diff
@@ -240,14 +240,17 @@ func TestBackendHandleRequest_badwrite(t *testing.T) {
...
 		Path:      "foo/bar",
 		Data:      map[string]interface{}{"value": "3false3"},
 	})
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
 
-	if err == nil {
-		t.Fatalf("should have thrown a conversion error")
+	if resp.Data["error"] != "Field validation failed: error converting input 3false3 for field \"value\": cannot parse '' as bool: strconv.ParseBool: parsing \"3false3\": invalid syntax" {
```
ncabatoff, line 252: Maybe just look for contains "Field validation failed"?  Nice to not have to update tests just because you tweak the wording of an error message.

ncabatoff (approved):
Minor tweak suggested, but it's optional.

sdk/framework/backend_test.go
ccapurso: Good call, thanks

ccapurso:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note block in the PR body, as described in https://github.com/GoogleCloudPlatform/magic-modules/blob/master/.ci/RELEASE_NOTES_GUIDE.md:

~~~
```release-note:TYPE
Release note
```
~~~

PR #12046, opened 2021-07-12
skip seal check for manual sealing

HridoyRoy:


command/operator_diagnose.go
```diff
@@ -629,10 +635,14 @@ SEALFAIL:
...
 	// The unseal diagnose check will simply attempt to use the barrier to encrypt and
 	// decrypt a mock value. It will not call runUnseal.
-	diagnose.Test(ctx, "Check Barrier Encryption", diagnose.WithTimeout(30*time.Second, func(ctx context.Context) error {
-		if barrierWrapper == nil {
+	diagnose.Test(ctx, "Check Autounseal Encryption", diagnose.WithTimeout(30*time.Second, func(ctx context.Context) error {
+		if barrierSeal == nil {
 			return fmt.Errorf("Diagnose could not create a barrier seal object.")
 		}
+		if barrierSeal.BarrierType() == wrapping.Shamir {
+			diagnose.Skipped(ctx, "Skipping barrier encryption for manual sealing, as Diagnose should not have access to unseal keys.")
```
sgmiller, line 643: I think we can simplify this as "Skipping barrier encryption test.  Only supported for auto-unseal."

sgmiller (approved):
Small language change, but otherwise approved.

command/operator_diagnose.go
HridoyRoy: Done!

PR #12049, opened 2021-07-12
Update node to latest stable version

arnav28:
- v10 has reached EOL so upgrading node to v14 which is the latest
stable build

**UI specific changes:**
1. Update engine in the package.json to enforce the node version
2. Update .nvmrc file for nvm use

**Infra/build changes:**
1. Install packagespec: https://github.com/hashicorp/packagespec#installation
2. Update the node version for docker (scripts > docker > dockerfile)
3. Update the node image in .circleci > config > executors > @executors.yml
4. Update node version in packages-oss.yml
5. Run _**make packages**_
6. This will update the circle ci and build configuration files
7. Commit all the changes (including the auto-updated files)


Monkeychip (approved):
Thank you for tackling this and for adding the steps! Not sure if you think we should have a release engineer also preview but might be worth having them look.

sarahethompson (approved):
Looks good on the packagespec side

PR #12051, opened 2021-07-12
Set up Dependabot

mmorel-35:


hsimon-hashicorp:
Hi @mmorel-35! Welcome to the Vault community. We're so glad you're here! The team discussed this change, and decided against including it because we do track npm dependencies internally and update them accordingly. 
While we are interested in possibly adopting Dependabot for our Go dependencies, we're not quite ready to do so yet.  We also note that Dependabot still lists Go support as "alpha", and that they don't yet handle security advisories, which reduces the immediate motivation.
Please don't let this discourage you from creating future PRs or issues, though, and do let me know if you have any questions I can help with. Thanks!

PR #12057, opened 2021-07-13
Add support for testing Azure storage backend in CI using Azurite.

ncabatoff:
This is exposing a problem with the Azurite List implementation and how it handles pagination, so this PR is on hold until they fix that.

### DIFF
--- a/physical/azure/azure.go
+++ b/physical/azure/azure.go
@@ -84,17 +84,29 @@ func NewAzureBackend(conf map[string]string, logger log.Logger) (physical.Backen
 	}
 
 	var environment azure.Environment
+	var URL *url.URL
 	var err error
 
-	if environmentURL != "" {
-		environment, err = azure.EnvironmentFromURL(environmentURL)
-		if err != nil {
-			return nil, fmt.Errorf("failed to look up Azure environment descriptor for URL %q: %w", environmentURL, err)
+	testHost := conf["testHost"]
+	switch {
+	case testHost != "":
+		URL = &url.URL{Scheme: "http", Host: testHost, Path: fmt.Sprintf("/%s/%s", accountName, name)}
+	default:
+		if environmentURL != "" {
+			environment, err = azure.EnvironmentFromURL(environmentURL)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


physical/azure/azure.go
```diff
@@ -84,17 +84,29 @@ func NewAzureBackend(conf map[string]string, logger log.Logger) (physical.Backen
...
 
 	var environment azure.Environment
+	var URL *url.URL
 	var err error
 
-	if environmentURL != "" {
-		environment, err = azure.EnvironmentFromURL(environmentURL)
-		if err != nil {
-			return nil, fmt.Errorf("failed to look up Azure environment descriptor for URL %q: %w", environmentURL, err)
+	testHost := conf["testHost"]
```
benashz, line 90: I wonder if there is a possibility of `testHost` being configured outside the test? Not sure if this is a valid concern though.

benashz (approved):
One minor question/concern, otherwise LGTM!

physical/azure/azure.go
ncabatoff: Yes, this does modify prod code, so a Vault user could make use of this undocumented option.  I'm okay with that personally, but if you foresee a problem speak up.

benashz: Perhaps we could rename `tesHost` to something that denotes that this configuration is not for prod. I would suggest something like `_testHost_` or  `_ciTestHost_`?

benashz:
```suggestion
	// used by integration tests
	testHost := conf["_testHost_"]
```

ncabatoff: Having the `test` prefix (and not documenting it) is good enough on that front for me.  There's plenty of examples of other such conf in our codebase, e.g. Raft's apply_delay: https://github.com/hashicorp/vault/blob/main/physical/raft/raft.go#L337

benashz (approved):

physical/azure/azure.go
benashz: Sounds good!

PR #12061, opened 2021-07-13
Add namespace in error

hghaf099:


vault/logical_system.go
```diff
@@ -2922,8 +2922,8 @@ func (b *SystemBackend) handleMonitor(ctx context.Context, req *logical.Request,
 	if logLevel == log.NoLevel {
 		return logical.ErrorResponse("unknown log level"), nil
 	}
-
-	flusher, ok := w.ResponseWriter.(http.Flusher)
+	nw := w.ResponseWriter.(*logical.NamespaceResponseWriter)
```
ncabatoff, line 2925: Are you sure this is necessary (genuine question, I'm not saying it isn't)?  Since we're embedding the ResponseWriter without giving it an explicit field name, I would've thought we could treat our wrapping type NamespaceResponseWriter more or less as a drop-in replacement.

http/handler.go
```diff
@@ -350,7 +350,7 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
 			return
 		}
 
-		h.ServeHTTP(w, r)
+		h.ServeHTTP(&logical.NamespaceResponseWriter{ResponseWriter: w, NamespacePath: r.Header.Get("X-Vault-Namespace")}, r)
```
ncabatoff, line 353: I'd rather see ResponseWriter and NamespacePath on their own lines, e.g.

```
h.ServeHTTP(&logical.NamespaceResponseWriter{
  ResponseWriter: w, 
  NamespacePath: r.Header.Get("X-Vault-Namespace"),
}, r)
```

This is more in line with our formatting conventions, and to me it looks more readable.

vault/logical_system.go
hghaf099: So, I am getting this error with the old code:
"""
--- FAIL: TestSysMonitorStreamingLogs (3.61s)
    /Users/hamid/repo/vault/http/sys_monitor_test.go:57: Error making API request.
        
        URL: GET https://127.0.0.1:62014/v1/sys/monitor?log_level=DEBUG
        Code: 400. Errors:
        
        * streaming not supported
"""        

and flusher is nil. 

The http.Flusher comments say this: 

"""
The default HTTP/1.x and HTTP/2 ResponseWriter implementations support Flusher, but ResponseWriter wrappers may not. Handlers should always test for this ability at runtime.
"""

I think the comment is talking about our use case as far as I understand.

http/handler.go
hghaf099: Will do! Thanks.

vault/logical_system.go
ncabatoff: Ok, I stand corrected!  I'd feel a little better if you made the cast conditional though.  I know that with the current code it's always the same underlying type, but if things change and that's not always true, this would panic.

hghaf099: Ah, I see. Sure thing, would this be fine ? 
"""
flusher, ok := w.ResponseWriter.(http.Flusher)
if !ok {
        // Casting to logical.ResponseWriter and try http.Flusher again
        nw := w.ResponseWriter.(*logical.NamespaceResponseWriter)
        flusher, ok = nw.ResponseWriter.(http.Flusher)
        if !ok {
	        return logical.ErrorResponse("streaming not supported"), nil
        }
}
"""

ncabatoff: Yeah I had in mind something like that.

hghaf099: Great! Thanks

changelog/12061.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+namespace: show namespace in error message along side Code and URL
```
ncabatoff, line 2: I wonder if we should call out the other change, e.g.

```
core (enterprise): namespace header included in responses, Go client uses it when displaying error messages
```

ncabatoff (approved):
Looks good!  I recommend you test this latest version of your PR using your test in the ent PR before merging.

hghaf099:
I made the change and ran the new test again, and it passed!

PR #12063, opened 2021-07-13
InfluxDB - Update API docs

MilenaHC:
Updating API docs for InfluxDB to include username_template parameter.


jasonodonnell (approved):
LGTM, I think this needs to be back ported to the 1.8 branch.

fairclothjm (approved):

PR #12077, opened 2021-07-14
Clarify step-down functionality in docs 

mikegreen:
Clarify that step-down can be run against any node in the cluster


mikegreen:
Question on docs-only changes that are not specific to a version/release - should I be putting the cherry-pick label on them when created?

ncabatoff:
> Question on docs-only changes that are not specific to a version/release - should I be putting the cherry-pick label on them when created?

Sure, may as well.

ncabatoff (approved):

PR #12079, opened 2021-07-14
Fix KV Version History queryParams on the component LinkedBlock

Monkeychip:
**The bug:** Originally `versions.hbs` was calling the component `ListItem` (which passes params through to LinkedBlock) using an ember helper `query-params`.  However, there appear to be some issues with this helper—[see the issue here](https://github.com/emberjs/ember.js/issues/18076). (As an aside I could not find solid Ember documentation on this helper and suspect though we use it in other places in the app, it's not something we should continue to use). Because of this, the actual query param `?version=X` was not be amended onto the params object that made its way [here](https://github.com/hashicorp/vault/pull/12079/files#diff-f1543c5c355e7974c72abe2ca53e3c37a747a2cfc8ad130cf8b7eb92c52df8f9R66). Thus the `transitionTo` method was borked.

**How I fixed it:** The `LinkedBlock` component was previously setup to handle `queryParams` by passing in the object directly. We followed this pattern and it fixed the bug. Additionally, I added test coverage and documentation on LinkedBlock so it would be a little easier to understand how to use it.

**Bug before** _to reproduce you have to click on the linkedblock. there is another nested link in the block that works. to hit it, just click on the edges of the block, away from the text._
broken

**Fixed bug**
fixed

### DIFF
--- a/changelog/12079.txt
+++ b/changelog/12079.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fix Version History queryParams on LinkedBlock
+```
--- a/ui/app/templates/vault/cluster/secrets/backend/versions.hbs
+++ b/ui/app/templates/vault/cluster/secrets/backend/versions.hbs
@@ -19,7 +19,11 @@
   </p.levelLeft>
 </PageHeader>
 <ListView @items={{reverse model.versions}} @itemNoun="version" as |list|>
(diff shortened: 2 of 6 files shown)


hashishaw (approved):
👏  excellent work, thanks for the extra docs and tests!

PR #12083, opened 2021-07-14
Database + Control Group updates

hashishaw:
Since database roles are coming from two different endpoints, we were doing non-typical things with the ajax request that circumvented our typical control group logic. This makes a few changes that make it easier to take advantage of default behavior, and put in control group checks otherwise:

- Add `type` parameter to secret/show page when linked from the list view
- Only call the relevant query when type is provided on the show page
- Check for control group consistently between GET credentials and GET role


hashishaw:
Branch got borked, opened #12111 instead

PR #12084, opened 2021-07-14
Improve Secret Empty States

hellobontempo:
Previously when a user adds a secret and leaves the key and/or value empty, the visuals are inconsistent. If there's no key but a value does exist, the value appears in the key column. If neither a key or value exist, the entire table row doesn't render. Finally, if there is a key with no value, the 'view' icon appears, but there is no content. Screenshot of old behavior:
image

### Changes reflected below render a 'minus-plain' glyph when a key and/or value does not exist
Screen Shot 2021-07-14 at 2 00 15 PM


ui/tests/integration/components/info-table-row-test.js
```diff
@@ -70,4 +75,27 @@ module('Integration | Component | InfoTableItem', function(hooks) {
 
     assert.dom('[data-test-item="array"]').hasText('valueArray', 'Confirm link with item value exist');
   });
+
+  test('it renders a dash (-) if a label and/or value do not exist', async function(assert) {
```
hashishaw, line 79: We should maybe have one more tests that confirms that any block content overrides any value content

hashishaw (changes requested):
Great work! One more test to add and then I think we're golden

hashishaw (approved):

PR #12086, opened 2021-07-14
Clarify KV 2 API docs

Monkeychip:
This PR updates the language on the KV2 API documentation. The UI team ran into some confusion when comparing the docs to the actual data returned and realized that the payload data needed to be updated on the return of metadata. Additionally, we found some places where slight modifications to the wording would help clarify the information.

### DIFF
--- a/changelog/12086.txt
+++ b/changelog/12086.txt
@@ -1,4 +1,0 @@
-```release-note:improvement
-KV2 API docs: Update the language on the KV2 API documentation to include missing payload data on Read Secret Metadata and 
-clarify language on cas parameter when creating or updating a secret.
-```
--- a/changelog/12086.txt
+++ b/changelog/12086.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+KV2 API docs: Update the language on the KV2 API documentation to include missing payload data on Read Secret Metadata and 
+clarify language on cas parameter when creating or updating a secret.
+```
--- a/website/content/api-docs/secret/kv/kv-v2.mdx
+++ b/website/content/api-docs/secret/kv/kv-v2.mdx
@@ -84,15 +84,17 @@ $ curl \
 {
   "data": {
     "cas_required": false,
-    "max_versions": 0,
-    "delete_version_after": "3h25m19s"
+    "delete_version_after": "3h25m19s",
+    "max_versions": 0
   }
 }
 ```
 
 ## Read Secret Version
 
-This endpoint retrieves the secret at the specified location.
+This endpoint retrieves the secret at the specified location. Metadata returned
+here is version specific metadata. Not to be confused with the Metadata set on the
+metadata endpoint.
 
 | Method | Path                                         |
 | :----- | :------------------------------------------- |
@@ -146,11 +148,9 @@ have an ACL policy granting the `update` capability.
 
 - `options` `(Map: <optional>)` – An object that holds option settings.
 
-  - `cas` `(int: <optional>)` - Set the "cas" value to use a Check-And-Set
-    operation. If not set the write will be allowed. If set to 0 a write will
-    only be allowed if the key doesn’t exist. If the index is non-zero the
-    write will only be allowed if the key’s current version matches the
-    version specified in the cas parameter.
+- `cas` `(int: <optional>)` - A flag required if Check-And-Set is set to true on
+  either the secret or the engine's config. A write will only then be allowed if 
+  key's current version matches the specified in the cas flag.
 
 - `data` `(Map: <required>)` – The contents of the data map will be stored and
   returned on read.
@@ -367,7 +367,7 @@ entries.
 ## Read Secret Metadata
 
 This endpoint retrieves the metadata and versions for the secret at the
-specified path.
+specified path. Metadata is version agnostic.
 
 | Method | Path                     |
 | :----- | :----------------------- |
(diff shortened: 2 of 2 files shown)


website/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -84,15 +84,17 @@ $ curl \
...
   }
 }
 ```
 
 ## Read Secret Version
 
-This endpoint retrieves the secret at the specified location.
+This endpoint retrieves the secret at the specified location. Metadata returned
+here is version specific metadata. Not to be confused with the Metadata set on the
+metadata endpoint.
```
ivana-mcconnell, line 97: This is a good distinction, but the sentences don't read smoothly to me. A suggestion:

> This endpoint retrieves the secret at the specified location. The metadata returned here (`created_time`, `deletion_time`, `destroyed`, and `version`) is version-specific. It should not be confused with the response from the metadata endpoint. 

I might also suggest linking "the metadata endpoint" to that section in the documentation.

website/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -146,11 +148,9 @@ have an ACL policy granting the `update` capability.
...
 - `options` `(Map: <optional>)` – An object that holds option settings.
 
-  - `cas` `(int: <optional>)` - Set the "cas" value to use a Check-And-Set
-    operation. If not set the write will be allowed. If set to 0 a write will
-    only be allowed if the key doesn’t exist. If the index is non-zero the
-    write will only be allowed if the key’s current version matches the
-    version specified in the cas parameter.
+- `cas` `(int: <optional>)` - A flag required if Check-And-Set is set to true on
+  either the secret or the engine's config. A write will only then be allowed if 
+  key's current version matches the specified in the cas flag.
```
ivana-mcconnell, line 153: I think everyone might write this differently so I'll just leave my suggestion here 😂 I think yours makes the point but it could be a little more verbose: 

>  `cas` `(int: <optional>)` - This flag is required if `cas_required` is set to `true` on either the secret or the engine's config. In order for a write to be successful, `cas` must be set to the current version of the secret. If `cas` is set to 0, the write will only be allowed if the key doesn't exist. 

I specified `cas_required` and adjusted the wording slightly. I also did keep the 0 case, since that's in the code sample.

website/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -367,7 +367,7 @@ entries.
 ## Read Secret Metadata
 
 This endpoint retrieves the metadata and versions for the secret at the
-specified path.
+specified path. Metadata is version agnostic.
```
ivana-mcconnell, line 370: I think it's 'version-agnostic' with a dash.

website/content/api-docs/secret/kv/kv-v2.mdx
Monkeychip: Perfect and will do!

website/content/api-docs/secret/kv/kv-v2.mdx
Monkeychip: Well said. I struggled with this one.

website/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -84,15 +84,17 @@ $ curl \
 {
   "data": {
     "cas_required": false,
-    "max_versions": 0,
-    "delete_version_after": "3h25m19s"
+    "delete_version_after": "3h25m19s",
```
Monkeychip, line 87: this was a nit, but I put them in the order they are returned on the request.

website/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -84,15 +84,17 @@ $ curl \
...
   }
 }
 ```
 
 ## Read Secret Version
 
-This endpoint retrieves the secret at the specified location.
+This endpoint retrieves the secret at the specified location. The metadata returned
+here ( `created_time, destroy,` and `version`) is version specific. It should not be
+confused with the response from the [metadata endpoint](/api/secret/kv/kv-v2#read-secret-metadata).
```
Monkeychip, line 97: Is this the correct way to do an anchor? I looked at other files and it appears this is the pattern, but wanted to double check.

website/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -146,11 +148,10 @@ have an ACL policy granting the `update` capability.
...
 
-  - `cas` `(int: <optional>)` - Set the "cas" value to use a Check-And-Set
-    operation. If not set the write will be allowed. If set to 0 a write will
-    only be allowed if the key doesn’t exist. If the index is non-zero the
-    write will only be allowed if the key’s current version matches the
-    version specified in the cas parameter.
+- `cas` `(int: <optional>)` - This flag is required if cas_required is set
+  to true on either the secret or the engine's config. In order for a write
+  to be successful, cas must be set to the current version of the secret.
+  If cas is set to 0, the write will only be allowed if the key doesn't exist.
```
Monkeychip, line 155: I started a new line based on what was there before.  However, I wasn't sure if there was a specific max character count or something I should be following for the length of line descriptions?

changelog/12086.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:improvement
```
ncabatoff, line 1: Docs don't get CL entries, they're really about the product/binary that we ship.

website/content/api-docs/secret/kv/kv-v2.mdx
briankassouf: Yep! looks correct

briankassouf (approved):
Looks good, but i agree this doesn't need a CL entry

changelog/12086.txt
Monkeychip: Thank you, I'll remove!

PR #12093, opened 2021-07-15
Revert #12061 due to failures in TestLogical_RequestSizeLimit

ncabatoff:
Example failure: https://app.circleci.com/pipelines/github/hashicorp/vault-enterprise/5806/workflows/ceee884a-cc93-4f49-af4c-ec9416896bbd/jobs/217709/tests#failed-test-0

This is reproducible locally, but not if you checkout the commit immediately preceding the merge of #12061.  On this branch that includes the revert, no longer reproducible.

### DIFF
--- a/api/response.go
+++ b/api/response.go
@@ -41,14 +41,12 @@ func (r *Response) Error() error {
 
 	r.Body.Close()
 	r.Body = ioutil.NopCloser(bodyBuf)
-	ns := r.Header.Get("X-Vault-Namespace")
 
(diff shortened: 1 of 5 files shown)


vishalnayak (approved):

ncabatoff:
My guess is that the breakage is due to interaction between the new responsewriter wrapper and the one we already had, namely HTTPResponseWriter.  We probably want to fold the namespace field into that, which is what I'd have suggested if I'd known it existed.

PR #12101, opened 2021-07-15
Suppress logging during a diagnose run

sgmiller:


HridoyRoy (approved):
Wow, I'm surprised that the default for an intercepterLogger isn't off, given that it's effectively a required field. Lgtm!

PR #12111, opened 2021-07-16
UI/database cg read role

hashishaw:
Since database roles are coming from two different endpoints, we were doing non-typical things with the ajax request that circumvented our typical control group logic. This makes a few changes that make it easier to take advantage of default behavior, and put in control group checks otherwise:

* Add type parameter to secret/show page when linked from the list view
* Only call the relevant query when type is provided on the show page
* Check for control group consistently between GET credentials and GET role
* Add type parameter to credentials page (and links to them) so that only one call is made when type is known

Example Scenario where `education` is a static type role, and `readonly` is a dynamic type role.

**BEFORE** Control groups on both roles. Notice that I clicked on a *static* type role, which should have had control group access for `database/static-roles/education`
before-cg-fix

**AFTER** Viewing and credentials on role both work correctly when control groups active on those paths
cg-after-dynamic
cg-after-static

**AFTER** Known behavior: when a developer only has CG read access for both credential paths as wildcards, or just `creds/*`:
```
path "/database/creds/*"
path "/database/static-creds/*"
```
The dynamic role will work as expected but the static role does not.
wildcard-cg-creds

Workaround options:
* Update the policy to allow `read` access on `/database/static-roles` so the developer can traverse to the correct path
* Update the policy to use specific paths (eg `/database/creds/readonly`) rather than wildcards
* * link directly with the `roleType` param (eg. `https://my-vault.dev/ui/vault/secrets/database/credentials/education?roleType=static` (example below)
staticrole-creds-cg-workaround


ui/app/adapters/database/role.js
```diff
@@ -24,11 +25,31 @@ export default ApplicationAdapter.extend({
...
+    });
   },
 
   dynamicRoles(backend, id) {
-    return this.ajax(this.urlFor(backend, id), 'GET', this.optionsForQuery(id));
+    return this.ajax(this.urlFor(backend, id), 'GET', this.optionsForQuery(id)).then(resp => {
+      if (id) {
+        return {
+          ...resp,
+          type: 'dynamic',
```
Monkeychip, line 46: smart move on adding a type.

Monkeychip (approved):
I really like you're approach on this. Nice work!

PR #12117, opened 2021-07-19
Add ability to optionally clone an api.Client's headers

benashz:
`client.Clone()` does not provide a convenient way to copy the source instances headers. The following PR adds the ability to optionally include the headers for any whenever the client is cloned.

### DIFF
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


ncabatoff (commented):
I haven't reviewed the tests, but the change LGTM.

tomhjp (approved):
LGTM

api/client_test.go
```diff
@@ -409,63 +410,112 @@ func TestClientNonTransportRoundTripper(t *testing.T) {
...
+		headers *http.Header
 	}
-
-	// Set all of the things that we provide setter methods for, which modify config values
-	err = client1.SetAddress("http://example.com:8080")
-	if err != nil {
-		t.Fatalf("SetAddress failed: %v", err)
+	tests := []struct {
+		name   string
+		fields fields
```
calvn, line 419: Is the intermediate `fields` needed or can we have *Config and *http.Header as part of the struct field directly?

calvn (commented):
One comment, otherwise looks good. Can you also add a changelog entry for this?

api/client_test.go
benashz: Absolutely.

calvn (approved):

PR #12119, opened 2021-07-19
Elaborating on telemetry persistence

mladlow:
Some users understand how an aggregator relates to Vault telemetry, and
some users are approaching this concept for the first time. Those newer
to the concepts benefit from some extra clarification that the metrics
sourced directly from Vault aren't stored anywhere.

Sources:
https://prometheus.io/docs/concepts/metric_types/
https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md
https://docs.splunk.com/observability/metrics-and-metadata/metric-types.html

### DIFF
--- a/website/content/docs/internals/telemetry.mdx
+++ b/website/content/docs/internals/telemetry.mdx
@@ -6,7 +6,7 @@ description: Learn about the telemetry data available in Vault.
 
 # Telemetry
 
-The Vault server process collects various runtime metrics about the performance of different libraries and subsystems. These metrics are aggregated on a ten second interval and are retained for one minute in-memory.
+The Vault server process collects various runtime metrics about the performance of different libraries and subsystems. These metrics are aggregated on a ten second interval and are retained for one minute in-memory. In order to monitor Vault and collect durable metrics, Telemetry from Vault must be stored in metrics aggregation software.
 
 To view the raw data, you must send a signal to the Vault process: on Unix-style operating systems, this is `USR1` while on Windows it is `BREAK`. When the Vault process receives this signal it will dump the current telemetry information to the process's `stderr`.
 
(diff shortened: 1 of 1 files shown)


hsimon-hashicorp (approved):

ncabatoff (approved):

vishalnayak (approved):

website/content/docs/internals/telemetry.mdx
```diff
@@ -49,9 +49,9 @@ The following is an example telemetry dump snippet:
 
 You'll note that log entries are prefixed with the metric type as follows:
 
-- **[C]** is a counter
-- **[G]** is a gauge
-- **[S]** is a summary
+- **[C]** is a counter. Counters are cumulative metrics that are incremented when some event occurs, and are reset at the end of reporting intervals. Vault retains counters and other metrics for one minute in-memory, so to see accurate and persistent counters over time an [aggregation solution][telemetry-stanza] must be configured.
+- **[G]** is a gauge. Gauges provide measurements of current values.
+- **[S]** is a summary. Summaries provide sample observations of values and can measure distributions of discrete events.
```
pmmukh, line 54: Should we note that these types correlate to Prometheus metric types specifically https://prometheus.io/docs/concepts/metric_types/ and link to that? cause say someone familiar with Timers in statsd might not understand that Summary is used for the same thing.

mladlow: I don't think they are specific to Prometheus metric types - they might apply to Splunk too? I pulled the definition from OpenMetrics. Do you have a suggestion for how I could make this description more broadly applicable to other metrics type? I don't 100% understand, is why I ask.

pmmukh: oh interesting, TIL about openMetrics 😀 ! That specification looks to be formalizing Prometheus standards, so maybe it's still specific to Prometheus? I'm not sure if there are other metric types to worry about, so just for this, I think maybe an example might make the use case clearer.
So my suggestion would be along the lines of 
```
Summaries provide sample observations of values, and is most commonly used for measuring timing durations of events.
```
but not sure if this is a kind of generalization that we'd prefer to avoid. It can be used for other things like `vault.autosnapshots.snapshot.size` also, but that is rarer.

mladlow: Ah, I see now. I've tried to summarize based on the kinds of things it seems we use them for.

pmmukh (approved):

website/content/docs/internals/telemetry.mdx
pmmukh: lgtm!

PR #12126, opened 2021-07-20
VAULT-2285 adding capability to accept comma separated entries for au…

hghaf099:
…th enable/tune


hghaf099:
Seems like secrets tune/enable are also missing the same parameters.

website/content/docs/commands/secrets/tune.mdx
```diff
@@ -32,6 +32,9 @@ $ vault secrets tune -audit-non-hmac-request-keys=value1 -audit-non-hmac-request
 The following flags are available in addition to the [standard set of
 flags](/docs/commands) included on all commands.
 
+`-allowed-response-headers` `(string: "")` - response header values that plugins
```
ncabatoff, line 35:
```suggestion
`-allowed-response-headers` `(string: "")` - response header values that the secret engine
```

website/content/docs/commands/secrets/tune.mdx
```diff
@@ -45,7 +48,13 @@ flags](/docs/commands) included on all commands.
...
+- `-description` `(string: "")` - Specifies the description of the mount.
+  This overrides the current stored value, if any.
+
 - `-max-lease-ttl` `(duration: "")` - The maximum lease TTL for this secrets
   engine. If unspecified, this defaults to the Vault server's globally
   configured maximum lease TTL, or a previously configured value for the secrets
   engine.
+
+- `-passthrough-request-headers` `(string: "")` - request header values that will
+  be sent to the plugin.
```
ncabatoff, line 60:
```suggestion
  be sent to the secrets engine.
```

website/content/docs/commands/secrets/enable.mdx
```diff
@@ -90,3 +90,9 @@ flags](/docs/commands) included on all commands.
 - `-path` `(string: "")` Place where the secrets engine will be accessible. This
   must be unique cross all secrets engines. This defaults to the "type" of the
   secrets engine.
+
+- `-passthrough-request-headers` `(string: "")` - request header values that will
+  be sent to the plugin.
```
ncabatoff, line 95:
```suggestion
  be sent to the secrets engine.
```

website/content/docs/commands/secrets/enable.mdx
```diff
@@ -90,3 +90,9 @@ flags](/docs/commands) included on all commands.
...
+
+- `-allowed-response-headers` `(string: "")` - response header values that plugins
```
ncabatoff, line 97:
```suggestion
- `-allowed-response-headers` `(string: "")` - response header values that the secret engine
```

website/content/docs/commands/auth/tune.mdx
```diff
@@ -45,7 +48,16 @@ flags](/docs/commands) included on all commands.
...
+- `-description` `(string: "")` - Specifies the description of the mount.
+  This overrides the current stored value, if any.
+
 - `-max-lease-ttl` `(duration: "")` - The maximum lease TTL for this auth
   method. If unspecified, this defaults to the Vault server's globally
   configured maximum lease TTL, or a previously configured value for the auth
   method.
+
+- `-passthrough-request-headers` `(string: "")` - request header values that will
+  be sent to the plugin.
```
ncabatoff, line 60:
```suggestion
  be sent to the auth method.
```

website/content/docs/commands/auth/tune.mdx
```diff
@@ -45,7 +48,16 @@ flags](/docs/commands) included on all commands.
...
+
+- `-token-type` `(string: "")` - Specifies the type of tokens that should be
+  returned by the mount.
```
ncabatoff, line 63:
```suggestion
  returned by the auth method.
```

website/content/docs/commands/auth/tune.mdx
```diff
@@ -32,6 +32,9 @@ $ vault auth tune -audit-non-hmac-request-keys=value1 -audit-non-hmac-request-ke
 The following flags are available in addition to the [standard set of
 flags](/docs/commands) included on all commands.
 
+- `-allowed-response-headers` `(string: "")` - response header values that plugins
```
ncabatoff, line 35:
```suggestion
- `-allowed-response-headers` `(string: "")` - response header values that the auth method
```

website/content/docs/commands/auth/enable.mdx
```diff
@@ -57,12 +57,27 @@ flags](/docs/commands) included on all commands.
   configured default lease TTL, or a previously configured value for the auth
   method.
 
+- `-passthrough-request-headers` `(string: "")` - request header values that will
+  be sent to the plugin.
```
ncabatoff, line 61:
```suggestion
  be sent to the auth method.
```

website/content/docs/commands/auth/enable.mdx
```diff
@@ -57,12 +57,27 @@ flags](/docs/commands) included on all commands.
...
+
+- `-allowed-response-headers` `(string: "")` - response header values that plugins
```
ncabatoff, line 63:
```suggestion
- `-allowed-response-headers` `(string: "")` - response header values that the auth method
```

command/secrets_tune_test.go
```diff
@@ -166,6 +166,9 @@ func TestSecretsTuneCommand_Run(t *testing.T) {
 				"-max-lease-ttl", "1h",
 				"-audit-non-hmac-request-keys", "foo,bar",
 				"-audit-non-hmac-response-keys", "foo,bar",
+				"-passthrough-request-headers", "authorization",
```
ncabatoff, line 169: Are you planning to add checks below to verify that these arguments had an effect?

website/content/docs/commands/auth/tune.mdx
```diff
@@ -45,7 +48,16 @@ flags](/docs/commands) included on all commands.
   configured default lease TTL, or a previously configured value for the auth
   method.
 
+- `-description` `(string: "")` - Specifies the description of the mount.
+  This overrides the current stored value, if any.
+
 - `-max-lease-ttl` `(duration: "")` - The maximum lease TTL for this auth
```
ncabatoff, line 54: Add listing-visibility?

vault/logical_system.go
```diff
@@ -1309,6 +1317,19 @@ func (b *SystemBackend) handleMountTuneWrite(ctx context.Context, req *logical.R
 	return b.handleTuneWriteCommon(ctx, path, data)
 }
 
+func augmentTuneWriteParams(paramIn interface{}) ([]string, error) {
+	newVal := paramIn.([]string)
```
ncabatoff, line 1321: This will panic if paramIn isn't a string slice.

vault/logical_system.go
```diff
@@ -1309,6 +1317,19 @@ func (b *SystemBackend) handleMountTuneWrite(ctx context.Context, req *logical.R
...
+	var outputSlice []string
+	for _, v := range newVal {
```
ncabatoff, line 1323: Is this the behaviour we want?  My thinking was that we should only expand commas if we were given a single string input, but I haven't looked to see what we do elsewhere.

vault/logical_system.go
```diff
@@ -1869,6 +1897,45 @@ func (b *SystemBackend) handleAuthTable(ctx context.Context, req *logical.Reques
...
+		"audit_non_hmac_request_keys",
+		"audit_non_hmac_response_keys",
+		"passthrough_request_headers",
+		"allowed_response_headers",
+	}
+	for _, paramName := range configParamNameSlice {
+		if raw, ok := configMap[paramName]; ok {
+			outputSlice := []string{}
+
+			rt := reflect.TypeOf(raw)
```
ncabatoff, line 1911: You don't need to use reflect explicitly, you can do type switches: https://tour.golang.org/methods/16

changelog/12126.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
+auth/secret: Adding capability to use comma separated parameters for auth enable/tune
```
ncabatoff, line 2: I don't see this multiline syntax documented anywhere so I wouldn't assume it's valid.

When choosing the string to put before the colon, have a look at https://github.com/hashicorp/vault/blob/master/CHANGELOG.md.  If you don't see any examples of it, it's probably not the right value.  I'm not sure what I'd put here, maybe `api, cli:`?

changelog/12126.txt
```diff
@@ -0,0 +1,4 @@
...
+auth/secret: Adding missing CLI parameters and updating website information accordingly
```
ncabatoff, line 3: Website changes aren't tracked by changelogs.

vault/logical_system.go
hghaf099: Based on what I have understood (please correct me if I am wrong), the schema for these values is TypeCommaStringSlice. Then, for example, the following two commands should work:
1 -   "vault auth tune -audit-non-hmac-request-keys=value1 -audit-non-hmac-request-keys=value2 -audit-non-hmac-request-keys=value3 userpass" 
2-  "vault auth tune -audit-non-hmac-request-keys=value1 -audit-non-hmac-request-keys=value2,value3 userpass"

and the result for the above two commands needs to match.
"audit_non_hmac_request_keys": [
        "value1",
        "value2",
        "value3"
      ],

vault/logical_system.go
hghaf099: Ah, yeah, I will do the type assertion before invoking this!

command/secrets_tune_test.go
hghaf099: Yes, Done

vault/logical_system.go
hghaf099: Sounds good!

website/content/docs/commands/auth/tune.mdx
hghaf099: Done

changelog/12126.txt
hghaf099: I see! Sounds good!

changelog/12126.txt
hghaf099: Sounds good! I will follow those then.
I think cli/api is a good choice for this PR.

command/secrets_tune_test.go
```diff
@@ -200,6 +204,18 @@ func TestSecretsTuneCommand_Run(t *testing.T) {
 			if exp := 3600; mountInfo.Config.MaxLeaseTTL != exp {
 				t.Errorf("expected %d to be %d", mountInfo.Config.MaxLeaseTTL, exp)
 			}
+			if exp := []string{"authorization", "www-authentication"}; !reflect.DeepEqual(exp, mountInfo.Config.PassthroughRequestHeaders) {
```
ncabatoff, line 207: It's fine to do it this way, but for future reference we often use `deep.Equal` instead: it returns a diff, which can allow for more helpful failure messages.

vault/logical_system.go
```diff
@@ -1869,6 +1899,44 @@ func (b *SystemBackend) handleAuthTable(ctx context.Context, req *logical.Reques
...
+		"audit_non_hmac_request_keys",
+		"audit_non_hmac_response_keys",
+		"passthrough_request_headers",
+		"allowed_response_headers",
+	}
+	for _, paramName := range configParamNameSlice {
+		if raw, ok := configMap[paramName]; ok {
+			outputSlice := []string{}
+
+			switch raw.(type) {
```
ncabatoff, line 1913: This is perfectly fine, but you can make it even more succinct by assigning in the switch, e.g. : https://play.golang.org/p/x-RL2pAi1Vl

vault/logical_system.go
```diff
@@ -1869,6 +1899,44 @@ func (b *SystemBackend) handleAuthTable(ctx context.Context, req *logical.Reques
...
+			case []interface{}:
+				rawNew := raw.([]interface{})
+				for _, rawVal := range rawNew {
+					rawValSt := rawVal.(string)
```
ncabatoff, line 1917: Are we safe to assume that the []interface{} will always only contain string elements?  I think so, but if you're not sure, it's always better to do a conditional cast and return an error if it's not the right type.  Better than a surprising panic.

vault/logical_system.go
```diff
@@ -1869,6 +1899,44 @@ func (b *SystemBackend) handleAuthTable(ctx context.Context, req *logical.Reques
 	return resp, nil
 }
 
+func augmentEnableAuthConfigMap(configMap map[string]interface{}) error {
```
ncabatoff, line 1902: It feels like you could find a less generic name for what this does, e.g. "expandStringValsWithCommas" or something.

vault/logical_system.go
```diff
@@ -1895,6 +1963,13 @@ func (b *SystemBackend) handleEnableAuth(ctx context.Context, req *logical.Reque
 	var apiConfig APIMountConfig
 
 	configMap := data.Get("config").(map[string]interface{})
+	// Augmenting configMap for some config options to treat them as comma separated entries
+	err := augmentEnableAuthConfigMap(configMap)
+	if err != nil {
+		return logical.ErrorResponse(
```
ncabatoff, line 1969: Typically when returning an ErrorResponse we return a nil error and let the framework sort out what HTTP error code to return.  It's not wrong to do it this way though.

vault/logical_system.go
ncabatoff: Yeah, I suspect your intuition was correct here.  I think I was just feeling like there should be some way to provide a value that contains a comma if needed, but upon further reflection I can't see why we'd need to.

vault/logical_system.go
```diff
@@ -1418,13 +1437,16 @@ func (b *SystemBackend) handleTuneWriteCommon(ctx context.Context, path string,
 	}
 
 	if rawVal, ok := data.GetOk("audit_non_hmac_request_keys"); ok {
-		auditNonHMACRequestKeys := rawVal.([]string)
+		newVal := rawVal.([]string)
+		auditNonHMACRequestKeys, err := augmentTuneWriteParams(newVal)
```
ncabatoff, line 1441: Why do we need this?  I thought this part was already working, since the param is declared as a TypeCommaStringSlice?

hghaf099: Yes, in the schema, the fields are defined as TypeCommaStringSlice. However, it seems that there is a bug somewhere in the mapstructure where the input is decoded. So, here is the code path: 
data.GetOk.-> data.GetOkError -> data.getPrimitive -> parseutil.ParseCommaStringSlice.

In there, mapstructure.DecoderConfig is instantiated having DecodeHook set to  mapstructure.StringToSliceHookFunc(","). There is a check in StringToSliceHookFunc where the bug show its symptom:
   if f != reflect.String || t != reflect.Slice {
       return data, nil
  }


"f" is defined the type for the input data, and "t" is defined the type for the output data. "f" should be string and "t" should be slice. 

As it seems, the "t" should be set in the DecoderConfig and should not be changed throughout the execution of the decode. However, it seems that it changes to something other than a slice, like a string. And this is where the comma separated strings are not converted properly (because the code returns before it reaches "strings.Split(raw, sep)") . So, the execution goes as follows:

From parseutil.ParseCommaStringSlice -> decoder.Decode -> d.decode : from this point on, some sort of recursive calls go through each element of the input where the output type changes from slice to string. And, if this change does not happen, the execution just loops for ever. It is hard to document all the steps here, let's have a chat when you have time to go through the sequence of function executions. 

I have not found a fix yet in a way that nothing else breaks. :/
So, long story short, the existing old code seems to be not working if the input is some sort of a slice, and that is why if we pass in "foo,bar" in the curl command, it works, but passing ["foo,bar"] it does not. For the cli, it always converts the input to slice and sends the slice to the endpoint.

Here are some examples of the results. I reverted my change for "-audit-non-hmac-request-keys" but kept it for "-allowed-response-headers" to show the difference in the output:
Command: 
vault auth tune   -audit-non-hmac-request-keys="value5,value6" -allowed-response-headers="authorization,www-authenticate" userpass

Result in Json format:
  "userpass/": {
    "uuid": "b01d4748-8490-e357-fb27-a8d4fad81508",
    "type": "userpass",
    "description": "",
    "accessor": "auth_userpass_44f766ab",
    "config": {
      "default_lease_ttl": 0,
      "max_lease_ttl": 0,
      "force_no_cache": false,
      "audit_non_hmac_request_keys": [
        "value5,value6"
      ],
      "passthrough_request_headers": [
        "Authorization"
      ],
      "allowed_response_headers": [
        "authorization",
        "www-authenticate"
      ],
      "token_type": "default-service"
    },
    "options": null,
    "local": false,
    "seal_wrap": false,
    "external_entropy_access": false
  }

Please let me know if you have any suggestions or solutions other than the approach I took to rerun ParseCommaStringSlice for every element in a slice.

vault/logical_system.go
hghaf099: I see, I just tried to be consistent with the use of logical.ErrorResponse in this file.

vault/logical_system.go
hghaf099: Done

vault/logical_system.go
hghaf099: Done

vault/logical_system.go
hghaf099: Done

command/secrets_tune_test.go
hghaf099: Done

website/content/docs/commands/auth/enable.mdx
hghaf099, line 61: Ah, I forgot to pull the committed suggestions first! I will fix these shortly

vault/logical_system.go
ncabatoff: I may have been influenced my intuition that this was how Vault behaves when, previously, I suggested that you should only be expanding comma-separated values when a single string was provided.

I worry that by changing this existing behaviour we may break someone's usage who depends on this somehow.  I also worry that if we start treating this TypeCommaStringSlice argument this way, other endpoints that make use of TypeCommaStringSlice which don't do the extra expansion will start being viewed as bugs by some people, and changing those would also introduce backwards compatibility issues.

My hunch is that we don't want to change this.  It may be a good topic for the pod sync though, feel free to add it to the agenda doc.

vault/logical_system.go
```diff
@@ -1869,6 +1899,47 @@ func (b *SystemBackend) handleAuthTable(ctx context.Context, req *logical.Reques
...
+		"audit_non_hmac_response_keys",
+		"passthrough_request_headers",
+		"allowed_response_headers",
+	}
+	for _, paramName := range configParamNameSlice {
+		if raw, ok := configMap[paramName]; ok {
+			outputSlice := []string{}
+
+			switch t := raw.(type) {
+			case []interface{}:
```
ncabatoff, line 1914: So maybe we do nothing in the slice case, for compatibility reasons.  Rationale: we only expand the commas as a convenience for users, because sometimes it's easier to produce a single comma-separated value than an array; if they can already produce an array, they don't need that convenience.  As a side benefit, it may provide a workaround if for some reason they need a comma in one these inputs: just use an array with an empty element.  ("may" because presumably not all endpoints tolerate an empty array element gracefully.)

vault/logical_system.go
ncabatoff: Yup, fine either way.

vault/logical_system.go
hghaf099: I see, sounds like a good topic for the pod sync. I will remove this part of the change for now then. Do we have some sort of a channel to understand how customers use these options? I have found some tests in our code where "Purpose" is used in KMS stanza for example, but for that we only have an example of an array like this ["foo", "bar"] which still works with this code. 
I will investigate other use cases of TypeCommaStringSlice and maybe even send a general question to a channel to see how people use that.

vault/logical_system.go
hghaf099: To clarify, if they pass in ["abc,def", "ghi"], then we want to preserve the comma in the first element, right? 
Then, both auth enable and auth tune would work the same.

ncabatoff: Yes.

ncabatoff: Do we need to keep the []interface{} case?  It doesn't seem like it does anything now.

website/content/docs/commands/auth/enable.mdx
```diff
@@ -57,12 +57,27 @@ flags](/docs/commands) included on all commands.
   configured default lease TTL, or a previously configured value for the auth
   method.
 
+- `-passthrough-request-headers` `(string: "")` - request header values that will
+  be sent to the auth method.
```
ncabatoff, line 61: I see that the audit options have a note about repeating the option, maybe we should put that on these new docs too?

ncabatoff (approved):
Couple small suggestions but overall looks good!

vault/logical_system.go
hghaf099: not really! Done

website/content/docs/commands/auth/enable.mdx
hghaf099: Done

PR #12138, opened 2021-07-21
Clarify sudo req'd for remount

mikegreen:
Making it obvious which policy remount needs. From https://discuss.hashicorp.com/t/understanding-of-secrets-move-command/26970


mikegreen:
superceded with #12139

PR #12146, opened 2021-07-21
Allow for logical client operations with a context arg

ncabatoff:
We've talked in the past about a `WithContext` helper similar to the `WithRequestCallbacks` method.  The idea would be to add a `context` field to Client, and WithContext take a ctx and return a shallow clone of the Client with `c.context = ctx`.  This would avoid the need to have new methods with a different signature.  I decided against this approach because it's not idiomatic, and in general putting contexts inside structs is frowned upon.  (There are good reasons why we do so inside Vault currently, which don't really apply here.)

### DIFF
--- a/api/logical.go
+++ b/api/logical.go
@@ -39,19 +39,60 @@ var (
 
 // Logical is used to perform logical backend operations on Vault.
 type Logical struct {
-	c *Client
+	c *LogicalContext
 }
 
-// Logical is used to return the client for logical-backend API calls.
 func (c *Client) Logical() *Logical {
-	return &Logical{c: c}
+	return &Logical{c: c.LogCtx()}
 }
 
 func (c *Logical) Read(path string) (*Secret, error) {
-	return c.ReadWithData(path, nil)
+	return c.c.Read(context.Background(), path)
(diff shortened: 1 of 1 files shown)


api/logical.go
```diff
@@ -39,19 +39,60 @@ var (
...
+	return c.c.Unwrap(context.Background(), wrappingToken)
+}
+
+// LogicalContext is used to perform logical backend operations on Vault with a
+// context argument.
+type LogicalContext struct {
+	c *Client
+}
+
+func (c *Client) LogCtx() *LogicalContext {
```
digivava, line 87: I wonder if this method's name might confuse people, since to me "LogCtx" sounds like a context related to logging.

ncabatoff: Yeah, I was pretty iffy about the name.  I was originally going with LogicalContext() but that seemed unwieldy, since people often write stuff like

```
resp, err := c.Logical().Write("/some/rather/long/path", map[string]interface{}{
 ...
```

and between the longer name and the new ctx argument it'll just be worse.  Still, probably better to have long lines than misleading names.

PR #12149, opened 2021-07-22
docs: add barrier key auto-rotation to internals/rotation

mickael-hc:
This functionality was introduced in Vault 1.7.0 and is documented in api-docs/system/rotate-config


yhyakuna (approved):
LGTM

PR #12159, opened 2021-07-23
Diagnose docs + changelog

HridoyRoy:
Docs and changelog for diagnose feature. I'm not adding the `docs/cherrypick` label because, if I'm not mistaken, that releases the docs immediately, whereas we want this to be released as part of the docs push for 1.8.


swayne275:
Will this need to be back ported to the 1.8 release branch?

swayne275 (approved):

HridoyRoy:
> Will this need to be back ported to the 1.8 release branch?

Nope! Docs don't need to be backported.

PR #12164, opened 2021-07-24
Removes grey shadow/line beneath secret form key/value input fields

hellobontempo:
**Previously a grey line displayed beneath each input field.** 
Screen Shot 2021-07-23 at 5 07 13 PM

**No longer shows line**
Screen Shot 2021-07-23 at 5 06 38 PM

### DIFF
--- a/ui/app/styles/components/info-table-row.scss
+++ b/ui/app/styles/components/info-table-row.scss
@@ -2,6 +2,10 @@
   box-shadow: 0 1px 0 $grey-light;
   margin: 0;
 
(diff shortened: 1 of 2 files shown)


hellobontempo:
This was the only use-case I found where removing the grey line for `class="info-table-row"` seemed appropriate? If anyone finds it elsewhere, please let me know!

Monkeychip (approved):
Great work, thank you. And for some context for future PR readers, this styling issue became more apparent in our work on the KV user metadata work. We confirmed with design the change.

PR #12174, opened 2021-07-27
Pin RabbitMQ and Cassandra docker image versions

ncabatoff:
Works around a rabbitmq regression in v3.9 with UserInfo.Tags; looks like they're now returning a string array into of a string?

This should fix these test failures:

https://app.circleci.com/pipelines/github/hashicorp/vault/19403/workflows/6758dc1c-b540-44f3-b4d5-0dae53366f32/jobs/251152/tests#failed-test-3

With the above changes I was seeing unrelated failures in Cassandra: looks like they came out with a 4.0.0 docker image just 7 hours ago.  

Failures: https://app.circleci.com/pipelines/github/hashicorp/vault/19404/workflows/206773e4-f72f-44fa-86dc-ac88ff6329bf/jobs/251201

So now I'm pinning Cassandra to 3.11 in this PR, in addition to pinning RabbitMQ 3.8.

Presumably we'll still need to make changes to Vault to support newer RabbitMQ/Cassandra versions.

### DIFF
--- a/builtin/logical/rabbitmq/backend_test.go
+++ b/builtin/logical/rabbitmq/backend_test.go
@@ -37,7 +37,7 @@ func prepareRabbitMQTestContainer(t *testing.T) (func(), string) {
 
 	runner, err := docker.NewServiceRunner(docker.RunOptions{
 		ImageRepo:     "rabbitmq",
-		ImageTag:      "3-management",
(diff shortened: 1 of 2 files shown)


tomhjp (approved):
LGTM, this seems like the correct level to pin these dependencies at.

PR #12181, opened 2021-07-27
UI: minor typo changes/fixes

hellobontempo:


Monkeychip (approved):
Question, I see there are two instances of overflow-ellipsis. Are we only removing the one because the other one overrides it? Maybe a comment on the PR about why this one is being removed.

ui/app/styles/components/linkable-item.scss
```diff
@@ -10,12 +10,3 @@
   grid-row: 1;
   align-self: center;
 }
-
-.overflow-ellipsis {
```
hellobontempo, line 14: Moved to `read-only.scss` once we refactored `LinkableItem` into separate components and created  a `ReadOnly` component - forgot the remove styling from this `.scss` file

PR #12182, opened 2021-07-27
Match KV key value to kv-object-editor component

Monkeychip:
While working on the kv custom metadata project we realized the kv-object-editor component and the kv key value inputs did not match. This pr matches their styling while keeping their functionality separate.

This matters because in the kv custom metadata project we will have both components on the page. Confirmed with design that the styles should match the gap space of the kv key value.

**New kv-object-editor look.**
image

**Matches with the kv key value.**
image

**Notes:**
I considered moving the secret key value fields into FormFields and using the kv-object-editor component. However, Form Fields works beacuse of model properties. KV key value are not set on the model but instead created in the component under the value secretData. TLDR: it's complicated to mix when we're only concerned with the styles not matching.

### DIFF
--- a/ui/app/styles/core/forms.scss
+++ b/ui/app/styles/core/forms.scss
@@ -13,7 +13,6 @@ label {
 
 .masked-font {
   font-family: obscure, text-security-square;
-  font-size: $size-medium;
   letter-spacing: 2px;
 }
 
--- a/ui/app/templates/components/secret-edit-display.hbs
+++ b/ui/app/templates/components/secret-edit-display.hbs
@@ -59,7 +59,7 @@
       {{/if}}
     </label>
   {{#each @secretData as |secret index|}}
-    <div class="columns is-variable is-1 has-no-shadow">
+    <div class="columns is-variable has-no-shadow">
       <div class="column is-one-quarter ">
         <Input 
           data-test-secret-key={{true}}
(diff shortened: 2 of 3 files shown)


ui/app/styles/core/forms.scss
```diff
@@ -13,7 +13,6 @@ label {
 
 .masked-font {
   font-family: obscure, text-security-square;
-  font-size: $size-medium;
```
Monkeychip, line 16: removed because it was making the masked-input larger than the key column

hellobontempo: thank you for fixing this!

ui/app/templates/components/secret-edit-display.hbs
```diff
@@ -59,8 +59,8 @@
       {{/if}}
     </label>
   {{#each @secretData as |secret index|}}
-    <div class="info-table-row has-no-shadow">
-      <div class="column is-one-quarter info-table-row-edit">
+    <div class="columns is-variable has-no-shadow">
```
hellobontempo, line 62: should this be `column` singular?

hellobontempo (changes requested):
I think "columns" needs to be singular in `secret-edit-display.hbs` then we're good to go!

hellobontempo (approved):

PR #12185, opened 2021-07-27
Update genUsername to cap STS usernames at 32 chars

vinay-gopalan:
This PR updates the `genUsername` function in the AWS secrets engine and updates it to cap STS usernames at 64 characters instead of 32. The PR also refactors the unit tests to reflect the updates and converts them to table tests.

Follow-up to https://github.com/hashicorp/vault/pull/12066


builtin/logical/aws/secret_access_keys.go
```diff
@@ -48,42 +47,26 @@ func secretAccessKeys(b *backend) *framework.Secret {
...
-		up, err := template.NewTemplate(template.Template(usernameTemplate))
-		if err != nil {
-			return "", "", fmt.Errorf("unable to initialize username template: %w", err)
-		}
-
-		um := UsernameMetadata{
-			DisplayName: normalizeDisplayName(displayName),
-			PolicyName:  normalizeDisplayName(policyName),
-		}
+	// IAM and STS usernames are capped at 64 chars; this leaves, after the beginning and
```
pcman312, line 50: Is the second half of this comment still accurate? With a template, I think this could probably go away?

builtin/logical/aws/secret_access_keys_test.go
```diff
@@ -47,53 +48,58 @@ func TestNormalizeDisplayName_NormNotRequired(t *testing.T) {
...
+			name:             "this---is---a---very---long---name",
+			policy:           "long------policy------name",
+			userType:         "iam_user",
+			UsernameTemplate: `{{ printf "%s-%s-%s-%s" (.DisplayName) (.PolicyName) (unix_time) (random 20) }}`,
+			warningExpected:  true,
+		},
+	}
+
+	for k, v := range tests {
+		testUsername, warning, err := genUsername(v.name, v.policy, v.userType, v.UsernameTemplate)
```
pcman312, line 77: You can wrap everything in this loop with a `t.Run(testName, func(t *testing.T) {...})` and it will separate the tests so they can be run independently and show separately in the test output. Then the key in the test map above should just be a description of the test, not used as values within the test itself.

builtin/logical/aws/secret_access_keys_test.go
```diff
@@ -47,53 +48,58 @@ func TestNormalizeDisplayName_NormNotRequired(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf(
```
pcman312, line 79: nit: Can the `t.Fatalf` calls be a single line instead of multi-line? It's much more common to have it on a single line unless the line is particularly long

pcman312: nit: I should also mention that you can instead use `require.Error(t, err)`: https://pkg.go.dev/github.com/stretchr/testify/require?utm_source=godoc#Error

builtin/logical/aws/secret_access_keys_test.go
```diff
@@ -47,53 +48,58 @@ func TestNormalizeDisplayName_NormNotRequired(t *testing.T) {
...
+		`this---is---a---very---long---name-long------policy------name-[0-9][0-9]`: {
+			name:             "this---is---a---very---long---name",
+			policy:           "long------policy------name",
+			userType:         "iam_user",
+			UsernameTemplate: `{{ printf "%s-%s-%s-%s" (.DisplayName) (.PolicyName) (unix_time) (random 20) }}`,
+			warningExpected:  true,
+		},
+	}
+
+	for k, v := range tests {
```
pcman312, line 76: nit: I recommend changing the names of `k` and `v` to something more descriptive like `name` and `test`

pcman312 (commented):
Looks good overall! Just a few small comments/suggestions

fairclothjm (approved):

pcman312 (approved):

builtin/logical/aws/secret_access_keys.go
```diff
@@ -48,42 +47,25 @@ func secretAccessKeys(b *backend) *framework.Secret {
 }
 
 func genUsername(displayName, policyName, userType, usernameTemplate string) (ret string, warning string, err error) {
-	switch userType {
-	case "iam_user":
```
calvn, line 52: Nit: Can we add a case for "assume_role" so make it a bit more clear that this applies to both instead of re-using the same string value?
```
case "iam_user", "assume_role":
```

We'd update this bit to use "assume_role" instead.

https://github.com/hashicorp/vault/blob/86c1cdf9e61897dee14203c51d6ad3d971cfda1a/builtin/logical/aws/secret_access_keys.go#L208

calvn (changes requested):
It looks like for STS federated tokens, the name is capped at 32 chars according to [their docs](https://docs.aws.amazon.com/STS/latest/APIReference/API_GetFederationToken.html), so we should keep the existing default template.

builtin/logical/aws/secret_access_keys.go
vinay-gopalan: @calvn added!

builtin/logical/aws/secret_access_keys.go
```diff
@@ -69,11 +68,10 @@ func genUsername(displayName, policyName, userType, usernameTemplate string) (re
 		// To prevent template from exceeding IAM length limits
 		if len(ret) > 64 {
 			ret = ret[0:64]
-			warning = "the calling token display name/IAM policy name were truncated to 64 characters to fit within IAM username length limits"
+			warning = fmt.Sprintf("the calling token's %s user name was truncated to 64 characters to fit within username length limits", userType)
 		}
 	case "sts":
-		// Capped at 32 chars, which leaves only a couple of characters to play
-		// with, so don't insert display name or policy name at all
+		// Capped at 32 chars
```
calvn, line 74: Can we move this comment to be right above where we declare `defaultSTSTemplate` since it's no longer obvious here??

builtin/logical/aws/secret_access_keys.go
```diff
@@ -69,11 +68,10 @@ func genUsername(displayName, policyName, userType, usernameTemplate string) (re
...
 		up, err := template.NewTemplate(template.Template(usernameTemplate))
 		if err != nil {
 			return "", "", fmt.Errorf("unable to initialize username template: %w", err)
```
calvn, line 77: I can't drop comments past here, but on L84 can you add a 32 char length check, similar to what we have in?

https://github.com/hashicorp/vault/blob/86c1cdf9e61897dee14203c51d6ad3d971cfda1a/builtin/logical/aws/secret_access_keys.go#L70

@pcman312 mentioned offline that returning an error if custom username templates is above character limit is an option, which I think is okay to do.

builtin/logical/aws/secret_access_keys.go
```diff
@@ -69,11 +68,10 @@ func genUsername(displayName, policyName, userType, usernameTemplate string) (re
 		// To prevent template from exceeding IAM length limits
 		if len(ret) > 64 {
 			ret = ret[0:64]
-			warning = "the calling token display name/IAM policy name were truncated to 64 characters to fit within IAM username length limits"
+			warning = fmt.Sprintf("the calling token's %s user name was truncated to 64 characters to fit within username length limits", userType)
```
calvn, line 71: I like Michael's idea of returning an error if custom username template results in usernames that's longer than the limit so perhaps we can turn this into an error (you'll also have to update the tests).

changelog/12185.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+secrets/aws: Refactor genUsername tests into table tests
```
calvn, line 2: If we change to errors instead of warnings for custom username templates char limits, we should call it out in here.

calvn (commented):
Left a few comments, but looking good! Can you also update the PRs title since that's what the squash merge will use?

builtin/logical/aws/secret_access_keys.go
vinay-gopalan: @calvn I can do that, but also wanted to mention that we only ever use the `defaultSTSTemplate` when generating usernames (which is always capped at 32 by default). Since the length limits are different and the feature ask was specifically for custom IAM user/role names, currently STS names are never generated by a custom user template.

pcman312: > currently STS names are never generated by a custom user template

Is there a technical reason for this? If not, this feels like a deficiency in the feature.

vinay-gopalan: I'm not sure if this is a strong enough technical reason, but my understanding is since STS is capped at 32 and IAM is capped at 64, we would have to add another STS template field in the AWS config to give users the ability to set custom templates for both. Otherwise if we expect the same template to work for both the cases, then that template would need to be capped at the minimum length cap and generate both STS and IAM usernames at 32 chars.


At the time of writing this feature the ticket's ask was for IAM, so I didn't add a separate STS custom username field, but should that also be in the scope here?

builtin/logical/aws/secret_access_keys.go
vinay-gopalan: Updated

builtin/logical/aws/secret_access_keys.go
vinay-gopalan: Updated

changelog/12185.txt
vinay-gopalan: Updated

builtin/logical/aws/secret_access_keys.go
calvn: > I didn't add a separate STS custom username field, but should that also be in the scope here?

We should be able to re-use the same `username_template` parameter for both (actually all three workflows).  `genUsername` is already doing the heavy lifting under the hood. So, the behavior on what's the cap on a specific workflow can be conditionally determined by the `userType` that's provided to the func call, i.e. within their own specific switch case.

calvn: Ah, I see the issue here. Since `username_template` is at the config level but the `auth_type` is at the role level, it doesn't work for cases where there are different role types created for the same mount.

calvn: We might have to add a username_template specific for federation tokens, but we can do that in a separate PR to keep this one clean. @pcman312 thoughts on this due to the above points?

pcman312: Having the template at the config level but the credential type at the role level does make this tricky. Off hand I see two options:

1. Have a single template that takes in the credential type as an argument to the template. This would allow the template to choose the max length based on the credential type
2. Specify one template for each credential type

I'm not sure I have a strong opinion either way. I don't really like either option all that much, but I also don't dislike either one. Either way we choose to go, I think we should consider other possible use cases similar to this and make sure we're consistent.

calvn: Option 1 is an interesting take. I assume it takes advantage of conditional within the template based on that value, i.e. `... {{if eq .CredType "sts"}} ... {{else}} ... {{end}}`?

I like it from an API perspective, but it makes crafting a custom username_template that works for both a bit trickier.

pcman312: Yea, that's the general idea. It makes things simpler and more complicated at the same time because a user would need to know that a custom template needs to handle all three cases.

calvn: If we can craft our default to use this pattern and include an example in the docs, I don't mind going down this route. If user feedback is that it's too complex we can re-visit and add a separate param on the API endpoint.

vinay-gopalan: @calvn @pcman312 I have added a template that generates usernames of differing lengths based on a conditional template and updated the tests till they all pass. Let me know what you think!

builtin/logical/aws/secret_access_keys.go
```diff
@@ -47,42 +46,41 @@ func secretAccessKeys(b *backend) *framework.Secret {
...
 
-		um := UsernameMetadata{}
+		um := UsernameMetadata{
+			Type: "STS",
+		}
 		ret, err = up.Generate(um)
 		if err != nil {
-			return "", "", fmt.Errorf("failed to generate username: %w", err)
+			return "", fmt.Errorf("failed to generate username: %w", err)
 		}
```
calvn, line 84: Can we add a 32 length check here (similar to the 64 length above) to guard against custom username templates that might generate longer than allowed usernames?

builtin/logical/aws/secret_access_keys_test.go
```diff
@@ -47,53 +47,61 @@ func TestNormalizeDisplayName_NormNotRequired(t *testing.T) {
...
+		"Truncated to 32. No warnings expected": {
+			name:             "name1",
+			policy:           "policy1",
+			userType:         "sts",
+			UsernameTemplate: defaultUserNameTemplate,
+			expectedError:    "",
+			expectedRegex:    `^vault-[0-9]+-[a-zA-Z0-9]+`,
+			expectedLength:   32,
+		},
+		"Too long. Warning expected": {
```
calvn, line 79:
```suggestion
		"Too long. Error expected": {
```

builtin/logical/aws/secret_access_keys_test.go
```diff
@@ -47,53 +47,61 @@ func TestNormalizeDisplayName_NormNotRequired(t *testing.T) {
...
+		name             string
+		policy           string
+		userType         string
+		UsernameTemplate string
+		expectedError    string
+		expectedRegex    string
+		expectedLength   int
+	}
+
+	tests := map[string]testCase{
```
calvn, line 60: Can we also add an error case for custom template that generates a long username for the `sts` credential type since the cap in that case will be 32 chars?

builtin/logical/aws/path_config_root.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/hashicorp/vault/sdk/logical"
 )
 
-const defaultUserNameTemplate = `{{ printf "vault-%s-%s-%s" (printf "%s-%s" (.DisplayName) (.PolicyName) | truncate 42) (unix_time) (random 20) | truncate 64 }}`
+const defaultUserNameTemplate = `{{ if (eq .Type "STS") }}{{ printf "vault-%s-%s"  (unix_time) (random 20) | truncate 32 }}{{ else }}{{ printf "vault-%s-%s-%s" (printf "%s-%s" (.DisplayName) (.PolicyName) | truncate 42) (unix_time) (random 20) | truncate 64 }}{{ end }}`
```
calvn, line 11: Can we include a comment on why this is done this way, i.e. single template to support different `credential_type`'s that has different length limits?

calvn (commented):
Looking good! Can we update the title of the PR to have a more accurate description of the latest changeset?

builtin/logical/aws/secret_access_keys.go
vinay-gopalan: Added!

builtin/logical/aws/secret_access_keys_test.go
vinay-gopalan: Added!

builtin/logical/aws/path_config_root.go
vinay-gopalan: Added

builtin/logical/aws/secret_access_keys.go
```diff
@@ -47,42 +46,45 @@ func secretAccessKeys(b *backend) *framework.Secret {
...
-		um := UsernameMetadata{}
+		um := UsernameMetadata{
+			Type: "STS",
+		}
 		ret, err = up.Generate(um)
 		if err != nil {
-			return "", "", fmt.Errorf("failed to generate username: %w", err)
+			return "", fmt.Errorf("failed to generate username: %w", err)
+		}
+		// To prevent template from exceeding STS length limits
```
calvn, line 85: Nit (same above):
```suggestion
		// To prevent a custom template from exceeding STS length limits
```

vinay-gopalan: Ah good call! Updated

calvn (approved):

PR #12186, opened 2021-07-27
Add 1.8 upgrade guide

kalafut:
Add 1.8 upgrade guide

### DIFF
--- a/website/content/docs/upgrading/upgrade-to-1.8.0.mdx
+++ b/website/content/docs/upgrading/upgrade-to-1.8.0.mdx
@@ -21,7 +21,7 @@ OSS users. Please see the [License](/docs/enterprise/license) documentation for 
 
 The following API endpoints have been deprecated and will be removed in a future release:
 
-* `PUT sys/license` to store licenses in storage; it is recommended to use
+* `sys/license` to manage licenses in storage; it is recommended to use
    [License Autoloading](/docs/enterprise/license/autoloading) instead.
 
 * `/gcp/token/:roleset` and `/gcp/key/:roleset` paths for generating secrets for rolesets
--- a/website/content/docs/upgrading/upgrade-to-1.8.0.mdx
+++ b/website/content/docs/upgrading/upgrade-to-1.8.0.mdx
@@ -26,3 +26,12 @@ The following API endpoints have been deprecated and will be removed in a future
 
 * `/gcp/token/:roleset` and `/gcp/key/:roleset` paths for generating secrets for rolesets
   in GCP Secrets. Use `/gcp/roleset/:roleset/token` and `/gcp/roleset/:roleset/key` instead.
+
+## Go Version
+
+Vault 1.8.0 is built with Go 1.16. Please review the [Go Release
+Notes](https://golang.org/doc/go1.16) for full details. Of particular note:
+
+- Go 1.16 has added support for darwin/arm64. Vault binaries for this platform are
+  now available supporting the Apple M1 CPU.
+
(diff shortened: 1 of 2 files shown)


website/content/docs/upgrading/upgrade-to-1.8.0.mdx
```diff
@@ -0,0 +1,28 @@
...
+OSS users. Please see the [License](docs/enterprise/license) documentation for more details.
+
+## Deprecations
+
+The following API endpoints have been deprecated and will be removed in a future release:
+
+* `PUT sys/license` to store licenses in storage; it is recommended to use
+   [License Autoloading](/docs/enterprise/license/autoloading) instead.
+
+* `/gcp/token/:roleset` and `/gcp/key/:roleset` paths for generating secrets for rolesets
```
calvn, line 27: Should we add a note to ask users to update ACL policies accordingly?

website/content/docs/upgrading/upgrade-to-1.8.0.mdx
```diff
@@ -0,0 +1,28 @@
...
+description: |-
+  This page contains the list of deprecations and important or breaking changes
+  for Vault 1.8.0. Please read it carefully.
+---
+
+# Overview
+
+This page contains the list of deprecations and important or breaking changes
+for Vault 1.8.0 compared to 1.7. Please read it carefully.
+
```
calvn, line 13: Sub-section here for Go version update, similar to https://www.vaultproject.io/docs/upgrading/upgrade-to-1.6.0#go-version?

kalafut: Good idea

website/content/docs/upgrading/upgrade-to-1.8.0.mdx
kalafut: IMO there are potentially a lot of things that you might need to do to switch over (both in and outside of vault). There isn't anything really specially about ACL policies for this change so I'd be inclined to not call it out.

website/content/docs/upgrading/upgrade-to-1.8.0.mdx
```diff
@@ -0,0 +1,37 @@
...
+Licenses and EULA enhancements have been introduced in the Vault 1.8 release.
+These changes are important for Enterprise customers to review. They do not affect
+OSS users. Please see the [License](/docs/enterprise/license) documentation for more details.
+
+## Deprecations
+
+The following API endpoints have been deprecated and will be removed in a future release:
+
+* `PUT sys/license` to store licenses in storage; it is recommended to use
```
ncabatoff, line 24: I'd assumed we were going to deprecate /sys/license and /sys/license/signed altogether, not just PUT.  Rationale: 
- GET /sys/license/signed doesn't make sense with autoloading, the operator already has the signed license (we also didn't implement anything to make it work with autoloading)
- DELETE /sys/license serves no purpose when you don't have a stored license any longer, though naturally we may want to keep this one around longer than the others
- GET /sys/license isn't needed because GET /sys/license/status provides the same information and more, like what kind of license is in use; we could keep it around anyway, but I figured it was easier for people to reason about if we could just say that all /sys/license endpoints were deprecated

kalafut: I've made the description more generic. We can update it again as needed if we do need to be more precise as to method. cc @aartiiyengar

ncabatoff (approved):

PR #12188, opened 2021-07-27
Set no_output_timeout for build-ui layer

mdeggies:
I added the value at `.circleci/config/@build-release.yml` and then ran `make ci-config` per the instructions in `.circleci/config.yml`. I think it's ok to make the change here directly, rather than updating packagespec, as we plan to migrate vault + vault ENT to CRT within the quarter and remove packagespec. Tagging @sarahethompson for review! Thank you :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -493,6 +493,7 @@ jobs:
 
           make -f packages*.lock/layer.mk 06-build-ui-26a986a756dd09ad76b230768f6d889c031d6fea-image
         name: Build build-ui layer
+        no_output_timeout: 15m
     - run:
         command: |2-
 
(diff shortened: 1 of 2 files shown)


ncabatoff:
We haven't really needed this on branches post 1.7.  Arnav put this same change in https://github.com/hashicorp/vault-enterprise/pull/2055, so I'll just go ahead and merge it.  You can probably close this PR.

sarahethompson (commented):
The only thing would be, the next time `make packages` is run, this line will be removed from the config as this target rewrites @build-release.yml. I need to do some digging into packagespec to see where this could be set. 
I don't think there is any way of doing this for an individual build layer step in packagespec  but we could possibly do it for all of them by editing - https://github.com/hashicorp/packagespec/blob/82602d640e24b866d23ce5729376a1e3496f14ef/internal/workflow/buildworkflow.go#L444 to set a step timeout.

PR #12196, opened 2021-07-28
VAULT-1303 when a request to vault fails, show namespace if set

hghaf099:


hghaf099:
I have also ran the new test I added in vault-ent to test this new functionality and it passes.

api/response.go
```diff
@@ -41,12 +41,14 @@ func (r *Response) Error() error {
 
 	r.Body.Close()
 	r.Body = ioutil.NopCloser(bodyBuf)
+	ns := r.Header.Get("X-Vault-Namespace")
```
ncabatoff, line 44: Can you use consts.NamespaceHeaderName instead of the literal string?

ncabatoff (approved):
Minor change suggested, looks good overall, nice work!

api/response.go
hghaf099: Done

PR #12199, opened 2021-07-29
Fixed typo for 'forward' value in relation to caching

rowansmithhc:
Per https://github.com/hashicorp/vault/blob/861454e0ed1390d67ddaf1a53c1798e5e291728c/command/agent.go#L440 the correct value to use should be documented as 'forward' instead of the two current incorrect values.

The value is referenced in the following two URL's:
https://www.vaultproject.io/docs/agent/caching#when_inconsistent
https://www.vaultproject.io/docs/enterprise/consistency#vault-agent-and-consistency-headers


ncabatoff (approved):

ncabatoff:
Thanks @rowansmithhc!

PR #12204, opened 2021-07-29
Backport 1.8: Pin RabbitMQ and Cassandra docker image versions (#12174)

ncabatoff:
* Work around rabbitmq regression with UserInfo.Tags in rabbitmq 3.9: use v3.8 docker image in tests.

* Also pin cassandra docker image version to 3.11 (4.00 was making tests fail)


calvn (approved):
I'm fine with the pins, though we should look into the regressions and failures due to newer DB versions on both engines.

PR #12206, opened 2021-07-29
docs: updating for vault-helm v0.14.0

tvoran:
Also found a couple missing values. The EnterpriseAlert macro doesn't seem to work anymore, so I removed it from the enterpriseLicense line (since it's pretty clear that's for enterprise only as-is).


tomhjp (approved):
LGTM

benashz (approved):
LGTM!

PR #12212, opened 2021-07-29
 UI/TTL helperEnabled/DisabledText fix 

hellobontempo:
**The UI defaults to generate credentials' TTL and Max-TTL for a role respectively to expire after 1 hr and 24hrs respectively, so this update is to clarify this behavior to customers using the UI by explicitly stating these expiry times.**

 **Old Role Text:**
Screen Shot 2021-07-30 at 2 33 35 PM

---

**New Role Text:**
Screen Shot 2021-07-30 at 2 30 49 PM

---

**Old Rotation TTL Text:**
_toggled off_
Screen Shot 2021-07-30 at 2 33 54 PM

---

_toggled on_
Screen Shot 2021-07-30 at 2 34 09 PM

---

**Corrected Text:**
_toggled off_
Screen Shot 2021-07-30 at 2 31 37 PM

_toggled on_
Screen Shot 2021-07-30 at 2 32 02 PM

### DIFF
--- a/ui/app/models/database/role.js
+++ b/ui/app/models/database/role.js
@@ -29,14 +29,14 @@ export default Model.extend({
     editType: 'ttl',
     defaultValue: '1h',
     label: 'Generated credentials’s Time-to-Live (TTL)',
-    helperTextDisabled: 'Vault will use a TTL of 1 hour',
+    helperTextDisabled: 'Vault will use a TTL of 1 hour.',
     defaultShown: 'Engine default',
   }),
   max_ttl: attr({
     editType: 'ttl',
     defaultValue: '24h',
     label: 'Generated credentials’s maximum Time-to-Live (Max TTL)',
-    helperTextDisabled: 'Vault will use a TTL of 24 hours',
+    helperTextDisabled: 'Vault will use a TTL of 24 hours.',
     defaultShown: 'Engine default',
   }),
   username: attr('string', {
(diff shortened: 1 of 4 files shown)


ui/app/models/database/role.js
```diff
@@ -45,8 +45,9 @@ export default Model.extend({
   rotation_period: attr({
     editType: 'ttl',
     defaultValue: '24h',
-    subText:
+    helperTextDisabled:
```
hashishaw, line 48: Do any other models' attributes with `editType: 'ttl',` have `subtext`? If so we should switch them out

ui/app/models/database/role.js
```diff
@@ -29,14 +29,14 @@ export default Model.extend({
     editType: 'ttl',
     defaultValue: '1h',
     label: 'Generated credentials’s Time-to-Live (TTL)',
-    subText: 'Vault will use the engine default of 1 hour',
+    helperTextDisabled: 'Vault will use a TTL of 1 hour.',
```
Monkeychip, line 32: Did Ivana have any thoughts on helperTextEnabled for these?

ui/app/models/database/role.js
hellobontempo: Nope these were the only TTL cases of `subText`. 

`aws-credential.js` does have a `helpText` attribute but I was planning to address that in another PR so this change could be merged. I was planning to set up AWS in order to test that TTL

ui/app/models/database/role.js
hellobontempo: No, but it'd be good to clarify a final time when she is back tomorrow.

hellobontempo: @Monkeychip - Ivana wants to stick with the default text!

Monkeychip (approved):
Nice work!

PR #12217, opened 2021-07-30
Identity: Percolate case insensitivity to inmem objects

vishalnayak:


aphorise:
@vishalnayak is this still relevant and something only relevant to inmem or needed? - to be perfectly frank I could not understand it's exact relevance or need.

PR #12229, opened 2021-08-02
Fix Diagnose Formatting In Disk Usage Checks

HridoyRoy:


vault/diagnose/os_common.go
```diff
@@ -31,11 +31,11 @@ partLoop:
 			Warn(ctx, fmt.Sprintf("Could not obtain partition usage for %s: %v.", partition.Mountpoint, err))
 		} else {
 			if usage.UsedPercent > 95 {
-				SpotWarn(ctx, testName, fmt.Sprintf(partition.Mountpoint+" is %d percent full.", usage.UsedPercent),
-					Advice("It is recommended to have more than five percent of the partition free."))
+				SpotWarn(ctx, testName, fmt.Sprintf(partition.Mountpoint+" is %f percent full.", usage.UsedPercent))
```
ncabatoff, line 34: I usually like to limit the number of digits of precision for float printfs, e.g. `%.2f`.  Otherwise it can look ungainly.

vault/diagnose/os_common.go
```diff
@@ -31,11 +31,11 @@ partLoop:
...
+				Advise(ctx, "It is recommended to have more than five percent of the partition free.")
 			} else if usage.Free < 2<<30 {
-				SpotWarn(ctx, testName, partition.Mountpoint+" has %d bytes full.",
-					Advice("It is recommended to have at least 1 GB of space free per partition."))
+				SpotWarn(ctx, testName, fmt.Sprintf(partition.Mountpoint+" has %d bytes free.", usage.Free))
```
ncabatoff, line 37: Bytes can be hard to read when we're talking about 8-9 digits.  Maybe show as MBs or MiBs instead?

vault/diagnose/os_common.go
HridoyRoy: Ah, didn't realize there was a simple way to make that nicer. Thanks, will do!

vault/diagnose/os_common.go
HridoyRoy: Will do!

ncabatoff (approved):

PR #12236, opened 2021-08-02
Recommended TOC change: Vault documentation

taoism4504:
This is a minor change to the TOC, but makes a fairly big impact on user documentation journey. I recommend that we provide an overview of Vault first and foremost, before users learn about installation and the preceding sections.

:mag: [Deploy preview](https://vault-plwibjap5-hashicorp.vercel.app/docs)

vault-docs

### DIFF
--- a/website/data/docs-nav-data.json
+++ b/website/data/docs-nav-data.json
@@ -7,10 +7,7 @@
     "title": "Use Cases",
     "path": "use-cases"
   },
-  {
-    "title": "Vault Integration Program",
-    "path": "partnerships"
-  },
+
   {
     "divider": true
   },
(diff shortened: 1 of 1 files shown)


yhyakuna (approved):
**What is Vault?** up front makes a good sense.  LGTM

kalafut:
I think **What is Vault?** and **Use Cases** are a lot better moved up! Not so much for **Vault Integration Program**, however, which is detail information for folks wanting to work with us on plugins and other integrations. I feel that should be in one of the lower sections, though I'm not sure where. cc @narayan-iyengar

narayan-iyengar:
I am okay with leaving the Vault Integration Program right below the plugin portal topic. That to see seems like a good combination of topics. Here is our plugin portal (list of all plugins that work with Vault) and here is the integration Program if you choose to join that list

PR #12246, opened 2021-08-04
Updating SDK version for 1.9 dev

mladlow:
@sgmiller I'm not sure if I should be doing something else for this update?


sgmiller (approved):
As far as I know, that's it. (and I did a deep search to double check)

Issue #9914, opened 2020-09-10
Upgrade github.com/ulikunitz/xz to v0.5.8


tsaarni:
Vault is (indirectly) depending on github.com/ulikunitz/xz v0.5.7 which is now being tagged by vulnerability scanners as security issue.

The security issue is reported in https://github.com/ulikunitz/xz/security/advisories/GHSA-25xm-hr59-7c27

The vulnerability has been fixed in https://github.com/ulikunitz/xz/releases/tag/v0.5.8.

pmmukh:
Closed by https://github.com/hashicorp/vault/pull/12253

PR #12253, opened 2021-08-04
Update github.com/ulikunitz/xz

tsaarni:
* Bump xz which is transitive dependency of github.com/mholt/archiver.
  Fixes known security vulnerability GHSA-25xm-hr59-7c27 #9914


changelog/12253.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+core: Update github.com/ulikunitz/xz
```
pmmukh, line 2: would it make sense to link to the security issue here, stating the reason for the upgrade ?

pmmukh (approved):
one minor comment, otherwise lgtm, and thanks for submitting the PR!

changelog/12253.txt
tsaarni: Done!

tsaarni:
@pmmukh Two tests fail `connection_producer_test.go:43: Could not start docker cassandra`.  This seems like possible flake.  Is it possible to re-trigger those two test cases?

pmmukh:
ah yep, those are flaky tests, will give em a rerun!

tsaarni:
The test seems to fail consistently now, also on my machine. Test is using `bitnami/cassandra:latest` which has been [updated to cassandra 4.0](https://hub.docker.com/r/bitnami/cassandra/tags?page=1&ordering=last_updated). 

I submitted a question if this is to be expected https://github.com/bitnami/bitnami-docker-cassandra/issues/98

tsaarni:
> The test seems to fail consistently now, also on my machine. Test is using bitnami/cassandra:latest which has been updated to cassandra 4.0 and startup is now much slower.

I've now merged main to pick up https://github.com/hashicorp/vault/pull/12311 which I missed in this PR. It forces fixed version `bitnami/cassandra:3.11` instead of `bitnami/cassandra:latest`, so tests should now pass.

Another alternative that worked for me locally, even with `bitnami/cassandra:4.0`, would be to increase this timeout

```diff
diff --git a/helper/testhelpers/docker/testhelpers.go b/helper/testhelpers/docker/testhelpers.go
index 673a0dea7..90698e323 100644
--- a/helper/testhelpers/docker/testhelpers.go
+++ b/helper/testhelpers/docker/testhelpers.go
@@ -146,7 +146,7 @@ func (d *Runner) StartService(ctx context.Context, connect ServiceAdapter) (*Ser

        bo := backoff.NewExponentialBackOff()
        bo.MaxInterval = time.Second * 5
-       bo.MaxElapsedTime = time.Minute
+       bo.MaxElapsedTime = 2 * time.Minute

        pieces := strings.Split(hostIPs[0], ":")
        portInt, err := strconv.Atoi(pieces[1])
```

but I'm still unsure if this long startup time of new cassandra is to be expected or bug.

pmmukh:
@tsaarni thanks for looking into that cassandra startup time issue! For the purposes of this PR, I'm going to merge this in and ignore that, I think as far as upgrading Cassandra tests to support cassandra 4, the ecosystem team will be taking a look at that later. Also, if your open issue there reaches a conclusion that some change in Vault can allow us to run tests against cassandra 4, please feel free to open a PR!

PR #12268, opened 2021-08-05
docs: since Vault 1.0 Unseal is OSS

greut:
Hi, I got confused reading the doc the other day.

### DIFF
--- a/website/content/api-docs/system/init.mdx
+++ b/website/content/api-docs/system/init.mdx
@@ -60,8 +60,6 @@ available when using Vault HSM.
   `secret_shares`. If using Vault HSM with auto-unsealing, this value must be
   the same as `secret_shares`.
 
-Additionally, the following options are only supported on Vault Pro/Enterprise:
-
 - `stored_shares` `(int: <required>)` – Specifies the number of shares that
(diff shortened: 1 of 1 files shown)


ncabatoff:
Hi @greut ,

Thanks for the PR!  Could you please remove the changelog entry though?  The changelog is only used to record changes to the Vault binary itself, it doesn't cover things like tests or docs.

website/content/api-docs/system/init.mdx
```diff
@@ -60,8 +60,6 @@ available when using Vault HSM.
   `secret_shares`. If using Vault HSM with auto-unsealing, this value must be
   the same as `secret_shares`.
 
-Additionally, the following options are only supported on Vault Pro/Enterprise:
```
ncabatoff, line 63: Instead of removing this altogether, how about something like "... only supported when using Auto Unseal"?  

There's also a bit higher up where we say "only available when using Vault HSM" that should also be adjusted similarly.

pmmukh: @greut Just wanted to drop a note, there is one change that was suggested that is pending, which is to reword the `only available when using Vault HSM` bit under the `Start Initialization` section.

ncabatoff (approved):

Issue #11853, opened 2021-06-14
Refactor api module's logical type to be able to read kv v2


adrianord:
**Is your feature request related to a problem? Please describe.**
In the Api module, the Logical type which facilitates logical backend operations cannot read from kv v2 without a lot of additional help.

**Describe the solution you'd like**
Move [`kv_helpers`](https://github.com/hashicorp/vault/blob/f8d255ead6521c02f458a0497183e459e6789f8e/command/kv_helpers.go) into the Api module, possible expose the functions to be public, and align the Api module with the Command module. kv_get has the logic to be able to read kv v1 and kv v2 that `Logical.ReadWithData` should have.

https://github.com/hashicorp/vault/blob/f8d255ead6521c02f458a0497183e459e6789f8e/command/kv_get.go#L89-L181

**Explain any additional use-cases**
The Command module could use the Logical type to read, list, and write kv for v1 and v2 if it were refactored to support v2, rather than having this logic repeated per command.

mahalrs:
@vishalnayak I will work on this. Based on my understanding, we want to move [kv_helpers](https://github.com/hashicorp/vault/blob/f8d255ead6521c02f458a0497183e459e6789f8e/command/kv_helpers.go) to `vault/api` directory, and expose all the functions to be public. Then, update `vault/command/kv_get.go` to use `vault/api/kv_helpers.go`. Please guide me with any additional details that may be helpful.

johnernaut:
@mahalrs You still planning on looking into this?  Otherwise, I'd be happy to give it a shot.  This is a current blocker for me.

pmmukh:
Hi! Thanks for submitting this feature request! After looking through it and the corresponding PR, we don't think this is something we want to handle at this time. There is certainly some validity here, moving the kv_helpers.go file into a layer where both the command and api modules can access it easily, is something we're interested in. However, we don't think moving it directly into the api module is the right call, as well as updating the generic functions `ReadWithData` and `List` to be aware of kv-v2 requests. ( the PR author did note attempting this broke a few tests ) 
Hence, I'll be closing out the linked PR, as I don't think that is the direction we want to go in, and swapping the `good-first-issue` label with an `internal` label, cause while this is something our devex team may want to take a look at in the future, we won't be accepting external contributions on this at this time.

averche:
Hi @adrianord!

[...]

PR #12284, opened 2021-08-09
improvement: refactor api module's logical type to be able to read kv v2

johnernaut:
This PR addresses issue #11853.

`kv_helpers` were moved into the `vault/api` package so that they could
be used by the methods attached to the `Logical` type (namely `List` and
`ReadWithData`) in order to support KV V2 calls.

We could potentially clean this up further by potentially removing the `KvReadRequest`
entirely and sticking with the `ReadWithData` method since there's so
much overlap, but will leave more refactoring suggestions to those more
familiar with the code-base.

### DIFF
--- a/api/kv_helpers.go
+++ b/api/kv_helpers.go
@@ -1,4 +1,4 @@
-package command
+package api
 
 import (
 	"errors"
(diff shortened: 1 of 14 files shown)


johnernaut:
@vishalnayak the test failures here are  making me realize that directly updating the `ReadWithData` method may be a mistake, as a lot of other things rely on this method to work a certain way, such as [token self-lookup](https://github.com/johnernaut/vault/blob/a24653cc5cca475962b27b71dcba29a4aa94d6ee/command/agent/cache_end_to_end_test.go#L346), [rate limiting](https://github.com/johnernaut/vault/blob/a24653cc5cca475962b27b71dcba29a4aa94d6ee/vault/external_tests/quotas/quotas_test.go#L221), etc...  The current problem is that I'm making another call to Vault within that method now to check to see if it should be a KV V1 or V2 request.

I'm wondering if a new method should be introduced here instead to use going forward.  Also, apologies if you weren't the correct person to ping around this.

pmmukh:
Closing this PR out based on https://github.com/hashicorp/vault/issues/11853#issuecomment-921198893

PR #12288, opened 2021-08-09
add known issue for dr secondary lease count quota invalidation

swayne275:


vishalnayak (approved):

mladlow:
It might take me a little while to review this but I'd like to request changes, so if you could hold off on merging, that would be helpful.

website/content/partials/lease-count-quota-upgrade.mdx
```diff
@@ -0,0 +1,5 @@
+### Lease Count Quota Invalidations on DR Secondaries Fixed
+
+Lease count quota invalidation causes DR Secondaries to panic and experience
+a hard shutdown. This issue exists prior to Vault 1.6.6, and 1.7.4. It has been
```
mladlow, line 4:
```suggestion
a hard shutdown. This issue exists prior to Vault 1.6.6, and 1.7.4. It is
```

mladlow: It won't quite be released when this PR merges, so we can leave the tense ambiguous this way.

website/content/docs/upgrading/upgrade-to-1.7.4.mdx
```diff
@@ -0,0 +1,48 @@
+---
+layout: docs
+page_title: Upgrading to Vault 1.7.4 - Guides
```
mladlow, line 3: Like the 1.6.6 upgrade page, we should put this on the 1.7.0 upgrade page at least, and maybe 1.7.1/1.7.2/1.7.3 (but I'm more on the fence about those).

website/data/docs-nav-data.json
```diff
@@ -1620,10 +1620,18 @@
         "title": "Upgrade to 1.6.3",
         "path": "upgrading/upgrade-to-1.6.3"
       },
+      {
```
mladlow, line 1623: If we get rid of the 1.6.6 and 1.7.4 pages, we won't want these links.

mladlow (changes requested):
I think it would be more consistent and more obvious to put the "known issues" on the 1.6.0 and 1.7.0 upgrade pages, as there isn't a known issue with 1.6.6 or 1.7.4. Changing that would mean that you don't need to add a new page for 1.6.6 and 1.7.4.

swayne275:
@mladlow I resolved your comments (thanks!). Let me know if you'd like this info to also be put on the minor 1.6 and 1.7 pages

mladlow (approved):
I think we don't need to add it to the other pages.

PR #12295, opened 2021-08-10
UI/StatText Component

hellobontempo:
A new component to display a label and value beneath, with the option to pass in subtext. If there is subtext, the component can be small or large, if there is no subtext then medium is also valid. 
Screen Shot 2021-08-12 at 9 44 24 AM

### DIFF
--- a/ui/app/styles/components/stat-text.scss
+++ b/ui/app/styles/components/stat-text.scss
@@ -1,10 +1,4 @@
-// size will change conditionally depending on whether or not a description is passed in
-// IF description then title does NOT change size
-// if no description, title gets incrementally smaller
-// there will not be a description in key value use cases
-
 .stat-text-container {
-  border: 1px solid lavender;
   line-height: normal;
 
   &.l {
@@ -17,6 +11,7 @@
       font-size: $size-8;
       font-weight: $font-weight-normal;
       color: $ui-gray-700;
+      line-height: inherit;
     }
     .stat-value {
       font-size: $size-3;
@@ -25,15 +20,17 @@
     }
   }
 
-  &.m {
+  &.s {
     > .stat-label {
-      font-size: $size-5; // which design? stays same size or decrease to 8 (~.857rem)
+      font-size: $size-5;
       font-weight: $font-weight-semibold;
+      line-height: inherit;
     }
     > .stat-text {
       font-size: $size-8;
       font-weight: $font-weight-normal;
       color: $ui-gray-700;
+      line-height: inherit;
     }
     > .stat-value {
       font-size: $size-5;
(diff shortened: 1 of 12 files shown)


ui/app/styles/components/stat-text.scss
```diff
@@ -0,0 +1,100 @@
+.stat-text-container {
+  line-height: normal;
+
+  &.l {
+    > .stat-label {
+      font-size: $size-5;
+      font-weight: $font-weight-semibold;
+      line-height: inherit;
+    }
+    .stat-text {
```
Monkeychip, line 10: curious here why you didn't need the `>` but you have it down below on row 23? Does this work without the carrot, e.g. is it not nested?

ui/app/styles/utils/_bulma_variables.scss
```diff
@@ -28,9 +28,10 @@ $family-sans: system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto'
 $family-monospace: 'SFMono-Regular', Consolas, 'Liberation Mono', Menlo, Courier, monospace;
 $family-primary: $family-sans;
 $body-size: 14px;
-$size-3: (24/14) + 0rem;
-$size-7: (13/14) + 0rem;
-$size-8: (12/14) + 0rem;
+$size-3: (24/14) + 0rem; // ~1.714rem
```
Monkeychip, line 31: thank you for adding the comments, that is helpful

ui/lib/core/addon/components/stat-text.js
```diff
@@ -0,0 +1,21 @@
+/**
```
Monkeychip, line 1: I think this is the right move, adding it to addons because we do want to use it in replication, which is in another engine. But normally, you wouldn't do this unless you knew it needed to be consumed in another engine within the app. e.g. KMIP, replication, or open-api-explorer (which all live outside of the app directory).

ui/tests/integration/components/stat-text-test.js
```diff
@@ -0,0 +1,32 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | StatText', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function(assert) {
+    // Set any properties with this.set('myProperty', 'value');
```
Monkeychip, line 10: I think these are default comments?  if yes, you can remove them.

ui/tests/integration/components/stat-text-test.js
```diff
@@ -0,0 +1,32 @@
...
+    // Handle any actions with this.set('myAction', function(val) { ... });
+
+    await render(hbs`<StatText />`);
+
+    assert.equal(this.element.textContent.trim(), '');
```
Monkeychip, line 15: Maybe instead search for the name of the component?  It's a glimmer component, so you should see it rendered. Ping me if that doesn't make sense.

Monkeychip (changes requested):

Monkeychip (approved):
Nice work!

ui/app/styles/components/stat-text.scss
```diff
@@ -0,0 +1,119 @@
...
+      color: $ui-gray-700;
+      line-height: inherit;
+    }
+    .stat-value {
+      font-size: $size-3;
+      font-weight: $font-weight-normal;
+      margin-top: $spacing-s;
+    }
+  }
+  // medium styling is the same as large
```
hashishaw, line 22: Instead of duplicating the contents, you could do `&.l, &.m { .... }`

PR #12296, opened 2021-08-10
Fix a tokenization docs link fix

sgmiller:


amharpin (approved):
Tested the new link and it works, thanks for doing this for me!

PR #12302, opened 2021-08-10
[Docs] Add code example

yhyakuna:
This PR adds [Go code snippet written by the DevEx team](https://github.com/hashicorp/vault-examples/tree/main/go).

🔍 [Deploy Preview - AppRole auth method](https://vault-git-docs-auth-code-example-hashicorp.vercel.app/docs/auth/approle#code-example)

🔍 [Deploy Preview - AWS auth method](https://vault-git-docs-auth-code-example-hashicorp.vercel.app/docs/auth/aws#code-example)

The code example section appears on the navigation drop-down(?):

code-example-final

### DIFF
--- a/website/content/docs/auth/approle.mdx
+++ b/website/content/docs/auth/approle.mdx
@@ -69,6 +69,89 @@ The response will contain the token at `auth.client_token`:
 }
 ```
 
+### Code example
+
+The following code snippet demonstrates AppRole authentication with response
+wrapping.
+
+<CodeTabs heading="AppRole auth with response wrap">
+
+<CodeBlockConfig highlight="40-48">
+
+```go
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	vault "github.com/hashicorp/vault/api"
+)
+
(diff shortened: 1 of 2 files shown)


digivava:
Yay! One thing I'm thinking is since the code sample is so far down the page, and I've never actually noticed that little "Jump to Section" dropdown (glad I know it now though!), it might be worth adding just a single sentence with a link right after the intro paragraph as well so people can find it quicker. (Especially since the AWS auth docs are so long.)

Something like "For an example of how to use this auth method in application code, check out the **Code Sample**."  (where Code Sample is a link to the heading below)

taoism4504 (approved):
I would image the reason why **Code Example** is showing up in the nav drop-down is because it's an H2, and all H2s appear in the nav drop-down. Not sure if it's a hard requirement to apply an H2 to Code Example, of if you can apply an H3 instead.

Other than that, LGTM.

digivava (approved):

yhyakuna:
For the AWS auth method doc, added a callout note in the beginning to draw attention to the command & code example sections since they are way down the bottom of the doc. 

code-example-callout

PR #12339, opened 2021-08-17
(OSS Port) Restrict Quota Deletion to Primary Cluster [vault-2399] 

HridoyRoy:
Ent testing: https://github.com/hashicorp/vault-enterprise/pull/2051/files

Description: This PR restricts quota deletion when an auth backend is removed to the primary cluster only. The PR above has a test that verifies the bug with quota deletion does not occur when secrets backends are deleted. We do not need to check if the cluster is a DR cluster, as DR clusters cannot utilize the quota deletion path or disable auth paths. Therefore, the bug doesn't show up in those areas.


vault/quotas/quotas.go
```diff
@@ -765,7 +765,7 @@ func (m *Manager) Invalidate(key string) {
 	default:
 		splitKeys := strings.Split(key, "/")
 		if len(splitKeys) != 2 {
-			m.logger.Error("incorrect key while invalidating quota rule")
+			m.logger.Error(fmt.Sprintf("incorrect key while invalidating quota rule. Key is %v", key))
```
vishalnayak, line 768:
```suggestion
			m.logger.Error("incorrect key while invalidating quota rule", "key", key)
```

vishalnayak (approved):
One minor suggestion. Otherwise, LGTM!

PR #12340, opened 2021-08-17
creds/aws: Add support for DSA signature verification for EC2

jasonodonnell:
DSA support was dropped in Go 1.16 which is used by the EC2 auth method for verifying signatures. Since this was dropped from a core library and EC2 needs DSA support until we can switch to RSA, I needed to bring in the external package we were using for verification and alter the code to add support. RSA adoption is a high priority task so we can formally deprecate the DSA support in Vault.


sgmiller (approved):
Looks fine to me, a straight lift of the PKCS7 routines we need.

builtin/credential/aws/pkcs7/README.md
```diff
@@ -0,0 +1,5 @@
+# PKCS7
+
+This code is used to veriy PKCS7 signatures for the EC2 auth method. The code
```
calvn, line 3:
```suggestion
This code is used to verify PKCS7 signatures for the EC2 auth method. The code 
```

builtin/credential/aws/path_login.go
```diff
@@ -348,8 +348,9 @@ func (b *backend) parseIdentityDocument(ctx context.Context, s logical.Storage,
 
 	// Verify extracts the authenticated attributes in the PKCS#7 signature, and verifies
 	// the authenticity of the content using 'dsa.PublicKey' embedded in the public certificate.
-	if pkcs7Data.Verify() != nil {
-		return nil, fmt.Errorf("failed to verify the signature")
+	err = pkcs7Data.Verify()
+	if err != nil {
```
calvn, line 352: Nit:
```suggestion
	if err := pkcs7Data.Verify(); err != nil {
```

calvn (approved):

PR #12345, opened 2021-08-18
Upgrade note for RedHat package issue

mladlow:


website/content/docs/upgrading/upgrade-to-1.8.0.mdx
```diff
@@ -41,3 +41,8 @@ Notes](https://golang.org/doc/go1.16) for full details. Of particular note:
 - AWS Auth using the [EC2 method](https://www.vaultproject.io/docs/auth/aws#ec2-auth-method)
   fails with the error `failed to verify the signature`. This effects 1.8.0 and 1.8.1 and there
   is not a workaround. The issue will be fixed in a future Vault minor release.
+- Configuration files in RedHat packages for Vault were not properly flagged as
+  config files for `fpm`, causing user-edited files on disk to be replaced with
+  the defaults when a new package was installed. This
+  [issue](https://github.com/hashicorp/vault/issues/12275) affects RedHat
+  packages 1.8.0, 1.8.1, and 1.8.1-1, and is fixed in 1.8.1-2 and up.
```
kpenfound, line 48: Should be fixed in 1.8.1-1

mladlow: I wasn't sure what to say about 1.8.1-1. For Vault binaries, we don't re-publish them even if something fairly catastrophic happens during the release because people cache them. For 1.8.1-1, I'd have to say "is fixed in 1.8.1-1 if downloaded after <date> and 1.8.1-2" or something like that, right? On the other hand, if someone is reading this note, it's effectively fixed in 1.8.1-1 downloads now. In the end, I decided to go with the messaging that matches how we release Vault binaries, even though it doesn't map perfectly to the packages. Would you prefer the "if downloaded after" or some other wording though? I don't feel extremely strongly about it.

kpenfound: Ah, so with the package versioning, the initial release of 1.8.1 would techinically be 1.8.1-0, the -0 is just omitted. So right now doing a `apt-get install vault` would get 1.8.1-1 with the fix, or `apt-get upgrade` with 1.8.0 installed would upgrade to `1.8.1-1`.  The 1.8.1-0 package is the one with the issue!

kpenfound (approved):

PR #12346, opened 2021-08-18
Docs: API Docs: Identity tokens

brianshumate:
The sample request in the **Generate a Signed ID Token** section includes references to an unnecessary `--data` flag and `payload.json`.

The endpoint/request does not actually require a payload, so I removed it from the sample as it was reported to be confusing.


aphorise:
Correction to [API example documentation](https://www.vaultproject.io/api-docs/secret/identity/tokens#sample-request-11) still relevant & should be merged.

yhyakuna (approved):

brianshumate:
I'm closing this PR due to age and difficulty with merging in favor of the same change in #16882.

PR #12348, opened 2021-08-18
UI: Allow metrics view without config read

hashishaw:
Fixes a bug where if config is not read, the metrics do not render correctly due to badly assigned default value. Now, the parent component passes in a valid value even if the one read from config is not available. Notice in the screenshot, the config tab is still not visible if the user cannot read the config endpoint

Screen Shot 2021-08-18 at 12 33 41 PM

### DIFF
--- a/ui/tests/acceptance/usage-metrics-test.js
+++ b/ui/tests/acceptance/usage-metrics-test.js
@@ -2,9 +2,22 @@ import { module, test } from 'qunit';
 import { visit, currentURL, findAll } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { create } from 'ember-cli-page-object';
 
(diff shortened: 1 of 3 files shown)


ui/tests/acceptance/usage-metrics-test.js
```diff
@@ -2,9 +2,22 @@ import { module, test } from 'qunit';
 import { visit, currentURL, findAll } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { create } from 'ember-cli-page-object';
 
 import authPage from 'vault/tests/pages/auth';
 import logout from 'vault/tests/pages/logout';
+import consoleClass from 'vault/tests/pages/components/console/ui-panel';
```
Monkeychip, line 9: thanks for adding test coverage

Monkeychip (approved):

PR #12354, opened 2021-08-18
Handle api explorer routing error

arnav28:
- For some reason when routing is done during async process, router transtionTo throws the TransitionAbortedError
- As a fix treat this particular error as success since it doesn't interfere in the routing
- Reference: https://github.com/emberjs/ember-test-helpers/issues/332

### DIFF
--- a/changelog/12354.txt
+++ b/changelog/12354.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixed api explorer routing bug
+```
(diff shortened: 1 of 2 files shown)


arnav28:
Screen Shot 2021-08-18 at 4 05 18 PM

Monkeychip (approved):
Looks good. I added the 1.9 milestone, but unsure if this should actually be backported?

PR #12357, opened 2021-08-19
Fixed overflowing text of flash message container

arnav28:


arnav28:
Screen Shot 2021-08-18 at 5 48 28 PM

Monkeychip (approved):
Nice work, same as before, unsure if this should be backported, and if so change the milestone to whatever version we're backporting to.

arnav28:
> Nice work, same as before, unsure if this should be backported, and if so change the milestone to whatever version we're backporting to.

I don't think backport is needed, it's very minor and edge case (text only overflows when the text is super long).

PR #12365, opened 2021-08-19
Add notes re dangers of identity write endpoints.

ncabatoff:


website/content/api-docs/secret/identity/entity-alias.mdx
```diff
@@ -6,6 +6,10 @@ description: >-
   store.
 ---
 
+~> **NOTE:** Be careful in granting permissions to non-readonly identity group
+endpoints. If they can modify group membership, they can add their entity to
```
mladlow, line 10: You use "If a user" elsewhere and I think that reads more clearly.

```suggestion
endpoints. If a user can modify group membership, they can add their entity to
```

website/content/api-docs/secret/identity/entity-alias.mdx
```diff
@@ -6,6 +6,10 @@ description: >-
   store.
 ---
 
+~> **NOTE:** Be careful in granting permissions to non-readonly identity group
```
mladlow, line 9: I couldn't tell what the word "group" meant here. Should this sentence go on the group page and be switched with the one that's currently on the group page?

website/content/api-docs/secret/identity/group.mdx
```diff
@@ -4,6 +4,11 @@ page_title: 'Identity Secret Backend: Group - HTTP API'
 description: This is the API documentation for managing groups in the identity store.
 ---
 
+~> **NOTE:** Be careful in granting permissions to non-readonly identity endpoints.
+If a user can modify an entity, they can grant it additional privileges through
+policies.  If they can modify an alias they can login with, they can bind it to
+an entity with higher privileges.
```
mladlow, line 10: Maybe?
```suggestion
an entity with higher privileges as long as they have access to the entity with more privileges.
```

mladlow (approved):
I have a couple suggestions, and it looks like the note on the identity/group page should be switched with the entity-alias page?

website/content/api-docs/secret/identity/group.mdx
ncabatoff: Not sure what you're getting at here?

website/content/api-docs/secret/identity/entity-alias.mdx
ncabatoff: Yeah, got it backwards.

website/content/api-docs/secret/identity/group.mdx
mladlow: The first time I read it my initial reading included some "they can bind it to ANY entity with higher privileges" and not "they can bind it to an entity they already have access to". This suggestion definitely gets into areas where my understanding is kind of hazy though, so I don't mind if you deem it nonsensical.

ncabatoff: It is the former, i.e. ANY entity.  That's the motivation for this warning.

PR #12358, opened 2021-08-19
Upgrade go-limiter to fix building on 1.17

sgmiller:
However go-limiter added context fields, so also modify the quota system to
supply them

PR #12369, opened 2021-08-19
Upgrade go-limiter to v0.7.0 to fix build failure in go1.17

johejo:
See https://github.com/sethvargo/go-limiter/pull/25


ncabatoff:
Thanks for the fix @johejo, but we're already on it: https://github.com/hashicorp/vault/pull/12358

Issue #12168, opened 2021-07-26
Incomplete snapshots returned by snapshot save command


needle82:
**Describe the bug**
On Openshift environment I have Vault 1.7.2 running, every night I take a snapshot with REST call. I wanted to restore this snapshot to another environment with posting it into snapshot-force, but getting {"errors":["1 error occurred:\n\t* failed to read snapshot file: failed to read or write snapshot data: unexpected EOF\n\n"]} all the time. I tried different backups and found that old of them are recoverable. I'm not able to restore files that are taken manually by CLI too.

**To Reproduce**
Create new StatefulSet with Vault 1.7.2 which is configured to use Raft. Initialize it with `vault operator init`. Vault becomes unsealed with autounseal schema. Set port forwarding for serving 8200:8200 and call from local machine `curl -vvv -L --header "X-Vault-Token: token" -X POST --data-binary @backup.snap http://localhost:8200/v1/sys/storage/raft/snapshot-force`

**Expected behavior**
Vault is rewritten with old data.

**Environment:**
* Vault 1.7.2
* Openshift 4, Kubernetes v1.20.0+87cc9a4

Vault server configuration file(s):

```hcl
storage "raft" {
  path = "/vault/data"
}

listener "tcp" {
  address     = "[::]:8200"
  tls_disable = 1
}

disable_mlock = true
ui = true

seal "transit" {
          disable_renewal = "false"
          key_name = "autounseal"
          mount_path = "transit/"
          tls_skip_verify = "true"
}

autopilot {
    cleanup_dead_servers = "true"
    last_contact_threshold = "200ms"
    last_contact_failure_threshold = "10m"
    max_trailing_logs = 250
    min_quorum = 3
    server_stabilization_time = "10s"
}
```

**Additional context**
No problems while getting snapshot in logs: 
2021-07-26T13:50:06.361Z [INFO]  storage.raft: starting snapshot up to: index=106948
2021-07-26T13:50:06.362Z [INFO]  storage.raft: compacting logs: from=96701 to=96708
2021-07-26T13:50:06.364Z [INFO]  storage.raft: snapshot complete up to: index=106948

ncabatoff:
Hi @needle82,

It's possible you're timing out, hence the EOF.  Have you tried setting `VAULT_CLIENT_TIMEOUT` as per  https://www.vaultproject.io/api/system/storage/raft#large-raft-snapshots-1 ?

needle82:
[...]

PR #12388, opened 2021-08-20
Add code to api.RaftSnapshot to detect incomplete snapshots

ncabatoff:
An incomplete snapshot is one that doesn't include the final SHA256SUMS.sealed file.


ncabatoff:
To test whether or not my implementation is a memory hog, I added the following to the end of OperatorRaftSnapshotSaveCommand.Run in operator_raft_snapshot_save.go:

```
	var m runtime.MemStats
	runtime.ReadMemStats(&m)
	c.UI.Info(fmt.Sprintf("totallocs=%d sys=%d", m.TotalAlloc, m.Sys))
```

I wrote 100k small keys to a raft cluster and took a snapshot with 1.8.1 and the above change:

```
totallocs=14887512 sys=76366856
```

With my branch and the above change:

```
totallocs=14945856 sys=75842568
```

(there's a bit of variability, but re-runs yielded similar numbers).  The snapshot file was 11MB uncompressed, and I'm pretty sure if we held it all in memory those memory numbers would've gone up substantially.  This confirms my understanding from code inspection of archive/tar, I just wanted to double check.

ncabatoff:
Fixes #12168

sgmiller (approved):
Looks good, love the injection of the error in the seal to trigger it.

PR #12393, opened 2021-08-20
Vault 2823 cc namespace

akshya96:
https://hashicorp.atlassian.net/browse/VAULT-2823

(PR addresses comments on https://github.com/hashicorp/vault-enterprise/pull/2103)

**will be creating a PR for enterprise updating the test for partialmonthlycount with entitites and tokens created using different namespaces** 

**Issue:** Currently, the partial client count API /sys/internal/counters - HTTP API | Vault by HashiCorp does not return the information subdivided by namespaces. The reason for that, is that we do not store entity counts broken down by namespace in memory at the moment. Ideally, we want customers to be able to observe the client counts broken down by namespace for the current month as well as prior months.

**Change:** Using a namespace->entity counter map (a.entityCountByNamespaceID) and incrementing it only if the entity is not in the map
Using non-entity token count (a.currentSegment.tokenCount.CountByNamespaceID) and distinct entity count (a.entityCountByNamespaceID) to divide the response by namespaces

**Testing: (Open Source Vault)**
{"request_id":"21dfb313-80b5-63df-25cf-9ced0cbe389d","lease_id":"","renewable":false,"lease_duration":0,"data":{"by_namespace":[{"namespace_id":"root","namespace_path":"","counts":{"distinct_entities":3,"non_entity_tokens":0,"clients":3}}],"clients":3,"distinct_entities":3,"non_entity_tokens":0},"wrap_info":null,"warnings":null,"auth":null}

**(Enterprise : with namespaces)**
{"request_id":"992b21de-0d7d-650d-5d6f-930b70bb2ecf","lease_id":"","renewable":false,"lease_duration":0,"data":{"by_namespace":[{"namespace_id":"root","namespace_path":"","counts":{"distinct_entities":3,"non_entity_tokens":0,"clients":3}},{"namespace_id":"N3MZO","namespace_path":"education/","counts":{"distinct_entities":1,"non_entity_tokens":0,"clients":1}},{"namespace_id":"kOiIk","namespace_path":"education/training/","counts":{"distinct_entities":1,"non_entity_tokens":0,"clients":1}}],"clients":5,"distinct_entities":5,"non_entity_tokens":0},"wrap_info":null,"warnings":null,"auth":null}


swayne275:
the failed `test-go-remote-docker` looks transient to me - you can probably just give that test a re-run

swayne275:
Also - big fan of the description with the PR!

vault/activity_log_testing_util.go
```diff
@@ -36,6 +36,29 @@ func (c *Core) InjectActivityLogDataThisMonth(t *testing.T) (map[string]struct{}
 	return activeEntities, tokens
 }
 
+// InjectActivityLogDataThisMonth populates the in-memory client store
+// with some entities and tokens for root namespace , overriding what was already there
```
swayne275, line 40:
```suggestion
// with some entities and tokens for root namespace, overriding what was already there
```

vault/activity_log.go
```diff
@@ -1792,26 +1805,91 @@ func (c *Core) activeEntityGaugeCollector(ctx context.Context) ([]metricsutil.Ga
...
+			clientCountTable[nsID] = &clients{distinctEntities: 0, nonEntityTokens: 0}
+		}
+		clientCountTable[nsID].nonEntityTokens += count
+
+	}
+	return clientCountTable
+
+}
+
+func (a *ActivityLog) AddEntity(e *activity.EntityRecord) {
```
swayne275, line 1889: i think this should be private - i didn't see any outside-package usage (or any reason to make it public)

vault/activity_log.go
```diff
@@ -1792,26 +1805,91 @@ func (c *Core) activeEntityGaugeCollector(ctx context.Context) ([]metricsutil.Ga
...
+
+	clientCountTable := createClientCountTable(a.entityCountByNamespaceID, a.currentSegment.tokenCount.CountByNamespaceID)
 
-	entityCount := len(a.activeEntities)
-	var tokenCount int
-	for _, countByNS := range a.currentSegment.tokenCount.CountByNamespaceID {
-		tokenCount += int(countByNS)
+	queryNS, err := namespace.FromContext(ctx)
+	if err != nil {
+		return responseData, err
```
swayne275, line 1825: what are your thoughts on returning this empty `responseData` vs `nil` when there is an `err`? It seems to be a pretty common pattern to return nil with an error

vault/activity_log.go
```diff
@@ -1792,26 +1805,91 @@ func (c *Core) activeEntityGaugeCollector(ctx context.Context) ([]metricsutil.Ga
...
 	}
-	clientCount := entityCount + tokenCount
 
-	responseData := make(map[string]interface{})
-	responseData["distinct_entities"] = entityCount
-	responseData["non_entity_tokens"] = tokenCount
-	responseData["clients"] = clientCount
+	for nsID, clients := range clientCountTable {
+
```
swayne275, line 1829:
```suggestion
```

vault/activity_log.go
swayne275: same in other `err != nil` cases

vault/activity_log.go
```diff
@@ -1792,26 +1805,91 @@ func (c *Core) activeEntityGaugeCollector(ctx context.Context) ([]metricsutil.Ga
...
+	//add distinct entity count
+	clientCountTable := make(map[string]*clients)
+	for nsID, count := range entityMap {
+		if _, ok := clientCountTable[nsID]; !ok {
+			clientCountTable[nsID] = &clients{distinctEntities: 0, nonEntityTokens: 0}
+		}
+		clientCountTable[nsID].distinctEntities += count
+
+	}
+	//add non-entity token count
```
swayne275, line 1877:
```suggestion

```

vault/activity_log.go
```diff
@@ -1792,26 +1805,91 @@ func (c *Core) activeEntityGaugeCollector(ctx context.Context) ([]metricsutil.Ga
...
+	responseData["distinct_entities"] = totalEntities
+	responseData["non_entity_tokens"] = totalTokens
+	responseData["clients"] = totalEntities + totalTokens
+
+	return responseData, nil
+}
+
+//createClientCountTable maps the entitycount and token count to the namespace id
+func createClientCountTable(entityMap map[string]uint64, tokenMap map[string]uint64) map[string]*clients {
+	//add distinct entity count
```
swayne275, line 1868:
```suggestion
```

vault/activity_log.go
```diff
@@ -1792,26 +1805,91 @@ func (c *Core) activeEntityGaugeCollector(ctx context.Context) ([]metricsutil.Ga
...
+	//add non-entity token count
+	for nsID, count := range tokenMap {
+		if _, ok := clientCountTable[nsID]; !ok {
+			clientCountTable[nsID] = &clients{distinctEntities: 0, nonEntityTokens: 0}
+		}
+		clientCountTable[nsID].nonEntityTokens += count
+
+	}
+	return clientCountTable
+
```
swayne275, line 1886:
```suggestion
```

vault/logical_system_activity.go
```diff
@@ -142,7 +142,12 @@ func (b *SystemBackend) handleMonthlyActivityCount(ctx context.Context, req *log
 		return logical.ErrorResponse("no activity log present"), nil
 	}
 
-	results := a.partialMonthClientCount(ctx)
+	results, err := a.partialMonthClientCount(ctx)
```
swayne275, line 145: i made this comment earlier, but it doesn't look like `results` has any info in it when `err != nil` - it might be more clear to show that results has nothing in it and put an empty `Data` object in the response in this error case

vault/activity_log_test.go
```diff
@@ -1390,6 +1390,80 @@ func setupActivityRecordsInStorage(t *testing.T, base time.Time, includeEntities
...
+			t.Fatalf(err.Error())
+		}
+
+		WriteToStorage(t, core, ActivityLogPrefix+"entity/"+fmt.Sprint(monthsAgo.Unix())+"/0", entityData1)
+		WriteToStorage(t, core, ActivityLogPrefix+"entity/"+fmt.Sprint(base.Unix())+"/0", entityData2)
+		WriteToStorage(t, core, ActivityLogPrefix+"entity/"+fmt.Sprint(base.Unix())+"/1", entityData3)
+	}
+
+	var tokenRecords map[string]uint64
+	if includeTokens {
```
swayne275, line 1449: everything up to this point looks identical with `setupActivityRecordsInStorage` - it might be a good idea to break some of that out into a base/helper function, and then just add the namespace flavor on top for the namespace version of the function. this would result in more clear and maintainable code

vault/activity_log_test.go
```diff
@@ -2520,7 +2594,10 @@ func TestActivityLog_partialMonthClientCount(t *testing.T) {
 
 	expectedClientCount := partialMonthEntityCount + partialMonthTokenCount
 
-	results := a.partialMonthClientCount(ctx)
+	results, err := a.partialMonthClientCount(ctx)
+	if err != nil {
+		t.Fatal(err.Error())
```
swayne275, line 2599: no need to call `.Error()` here if you don't want to

swayne275 (commented):
I think it's looking pretty good so far. I left a few comments/suggestions - let me know your thoughts

vault/logical_system_activity.go
swayne275: did you have any thoughts on this?

vault/activity_log_test.go
```diff
@@ -1326,129 +1328,55 @@ func setupActivityRecordsInStorage(t *testing.T, base time.Time, includeEntities
...
 			{
 				EntityID:    "22222222-2222-2222-2222-222222222222",
-				NamespaceID: "root",
+				NamespaceID: namespace.RootNamespaceID,
 				Timestamp:   time.Now().Unix(),
 			},
 			{
 				EntityID:    "33333333-2222-2222-2222-222222222222",
-				NamespaceID: "root",
+				NamespaceID: namespace.RootNamespaceID,
```
swayne275, line 1341: nice!

vault/activity_log_test.go
```diff
@@ -1326,129 +1328,55 @@ func setupActivityRecordsInStorage(t *testing.T, base time.Time, includeEntities
...
 				Timestamp:   time.Now().Unix(),
 			},
 		}
+		if constants.IsEnterprise {
```
swayne275, line 1345: it's probably better to put this in `vault/activity_log_ent_test.go`

vault/logical_system_activity.go
ncabatoff: The Go convention for funcs that return (x,error) is to populate only one of the two return values - I'd prefer if it were exactly one, but sometimes we do return (nil,nil), which tends to cause bugs because people are so used to the more common practice of always returning exactly one non-nil value.

This carries over to our handlers as well.  If they return a non-nil error, they should be returning a nil response.  The complication is that we can return errors in two ways.  We can either return (nil, err) in which case the user will not an error message, they'll just get a 500 internal server error.  Or we can return (logical.ErrorResponse("message"), nil) in which case we'll put "message" into the response's `error` field and return a 400 bad request.  Or we can return (logical.ErrorReponse("message"), err), in which case they'll see "message" in the response's `error` field and the status code of the response will depend on err, see logical.RespondErrorCommon for specifics.

There's no general guidance about when to use these different approaches, though we should probably write some.  One tip I can offer is that if the error is an "expected" error, e.g. the user provided bad input in their request, or asked to do something we don't want to allow, it's good to provide an explanation of the problem in the response's error message.  If it's an internal failure that shouldn't have happened, that isn't predictable based on the inputs, then we usually tend more towards the (nil, err) return where they just a 500.  This is because we don't know what appear in the error, and we might wind up leaking information that the user shouldn't see.  Often in that case we'll log the error.

Anyway, in this case, when err != nil, I would simply return (nil, err).

vault/activity_log.go
```diff
@@ -1792,26 +1808,94 @@ func (c *Core) activeEntityGaugeCollector(ctx context.Context) ([]metricsutil.Ga
...
-	clientCount := entityCount + tokenCount
 
-	responseData := make(map[string]interface{})
-	responseData["distinct_entities"] = entityCount
-	responseData["non_entity_tokens"] = tokenCount
-	responseData["clients"] = clientCount
+	for nsID, clients := range clientCountTable {
+		ns, err := NamespaceByID(ctx, nsID, a.core)
+		if err != nil {
+			return responseData, err
```
ncabatoff, line 1834: I think this should be `return nil, err`.

vault/activity_log_test.go
ncabatoff: Why so?

swayne275: it's an enterprise specific test involving enterprise features (namespaces). that said, it does mean less code duplication by having this as an optional add-on, rather than two separate tests

what do we generally do in this situation?

ncabatoff: We went back and forth on some variations on this theme.  Typically we have two variations on the same test, one oss, one ent, or we have the test only in ent.  I think this approach is fine too, and has some advantages.  We should probably expand the wiki "Writing Tests: Models to Follow" to talk about this topic.

ncabatoff (approved):

PR #12409, opened 2021-08-23
Bug Fix: tab on MaskedInput for GeneratedItems it was clearing the value

Monkeychip:
Only in the case where the `MaskedInput` component was being called via `GeneratedItem` (e.g. auth methods with sensitive field inputs) the data was cleared on tab.

This is because the `onKeyUp` function was not being handled in the `MaskedInput` component but instead in the `FormField` component. I moved this function down a level into the `MaskedInput`. I tested all the other use cases of `MaskedInput` and things were working as expected.

**To replicate:**
- On userpass auth method, create a user and tab on the password field. Before this would clear the data, now it does not.

### DIFF
--- a/ui/lib/core/addon/components/form-field.js
+++ b/ui/lib/core/addon/components/form-field.js
@@ -18,7 +18,7 @@ import layout from '../templates/components/form-field';
  * ```
  *
  * @param [onChange=null] {Func} - Called whenever a value on the model changes via the component.
- * @param [onKeyUp=null] {Func} - A function passed through into MaskedInput to handle validation.
+ * @param [onKeyUp=null] {Func} - A function passed through into MaskedInput to handle validation. It is also handled for certain form-field types here in the action handleKeyUp.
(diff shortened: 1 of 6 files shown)


ui/lib/core/addon/components/form-field.js
```diff
@@ -18,7 +18,7 @@ import layout from '../templates/components/form-field';
  * ```
  *
  * @param [onChange=null] {Func} - Called whenever a value on the model changes via the component.
- * @param [onKeyUp=null] {Func} - Called whenever cp-validations is being used and you need to validation on keyup.  Send name of field and value of input.
+ * @param [onKeyUp=null] {Func} - A function passed through into MaskedInput to handle validation.
```
hashishaw, line 21: This method is also called on the regular input, should that be called out here?

hashishaw (approved):
Nice work! And thanks for adding tests. One small docs change request

ui/lib/core/addon/components/form-field.js
Monkeychip: Good point. I'll add that.

Issue #12064, opened 2021-07-13
Setting transit/cache-config to > 0 causes terminal election/seal cycle


mikegreen:
**Describe the bug**

**To Reproduce**
1. Vault running with integrated storage, single node is fine. But this will also render an entire cluster unusable the same manner as it won't elect a leader. 
2. `$ vault secrets enable transit`
`Success! Enabled the transit secrets engine at: transit/`
3. `$ vault write transit/cache-config size=1`
`WARNING! The following warnings were returned from Vault:`
`  * cache configurations will be applied when this backend is restarted`
4. Restart vault service
5. Node cannot complete full startup. See bottom of issue for full log.
```
2021-07-13T20:44:00.394Z [ERROR] core: failed to create mount entry: path=transit/ error="failed to create cache: Must provide a positive size"
2021-07-13T20:44:00.394Z [INFO]  core: pre-seal teardown starting
2021-07-13T20:44:00.394Z [INFO]  core: stopping raft active node
2021-07-13T20:44:00.394Z [INFO]  core: stopping replication
2021-07-13T20:44:00.394Z [INFO]  core: closed sync connection
2021-07-13T20:44:00.394Z [INFO]  core: replication stopped
```

**Expected behavior**
Regardless of cache-config setting, node and cluster are stable.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.7.3+ent
* Vault CLI Version (retrieve with `vault version`): 1.7.3+ent
* Server Operating System/Architecture: Ubuntu 18 LTS

Vault server configuration file(s):

```hcl
# Paste your Vault config here.
# Be sure to scrub any sensitive values
```

**Additional context**
```
Jul 13 20:43:48 cr3yhx1 vault[38818]: 2021-07-13T20:43:48.208Z [DEBUG] core: starting cluster listeners
Jul 13 20:43:48 cr3yhx1 vault[38818]: 2021-07-13T20:43:48.208Z [INFO]  core.cluster-listener.tcp: starting listener: listener_address=192.168.1.215:8201
Jul 13 20:43:48 cr3yhx1 vault[38818]: 2021-07-13T20:43:48.208Z [INFO]  core.cluster-listener: serving cluster requests: cluster_listen_address=192.168.1.215:8201
[...]

PR #12418, opened 2021-08-24
Enforce minimum cache size for transit backend

divyapola5:
https://hashicorp.atlassian.net/browse/VAULT-2914

This PR includes changes to enforce minimum cache size (10) for transit backend.

**Testing done:**

**_Default size:_**

divyapola@divyapola-C02CF744MD6R vault % vault secrets enable transit
Success! Enabled the transit secrets engine at: transit/
divyapola@divyapola-C02CF744MD6R vault % vault read transit/cache-config
**Key     Value
size    0**

----- ---------

**_Set the cache size to a value less than 10:_**

divyapola@divyapola-C02CF744MD6R vault % vault write transit/cache-config size=6

  * size 6 is less than default minimum 10. Cache size is set to 10

  * cache configurations will be applied when this backend is restarted

divyapola@divyapola-C02CF744MD6R vault % vault read transit/cache-config        
**Key     Value
size    10**

----- ---------

**_Set the cache size to 1:_**

divyapola@divyapola-C02CF744MD6R vault % vault write transit/cache-config size=1

  * size 1 is less than default minimum 10. Cache size is set to 10

  * cache configurations will be applied when this backend is restarted

divyapola@divyapola-C02CF744MD6R vault % vault read transit/cache-config        
**Key     Value
size    10**

----- ---------

**_Successfully restarted vault:_**

divyapola@divyapola-C02CF744MD6R vault % vault server -config=config.hcl
You cannot specify a custom root token ID outside of "dev" mode. Your request
has been ignored.
==> Vault server configuration:

             Api Address: http://127.0.0.1:8200
                     Cgo: disabled
         Cluster Address: https://127.0.0.1:8201
              Go Version: go1.16.6
              Listener 1: tcp (addr: "0.0.0.0:8200", cluster address: "0.0.0.0:8201", max_request_duration: "1m30s", max_request_size: "33554432", tls: "disabled")
               Log Level: info
                   Mlock: supported: false, enabled: false
           Recovery Mode: false
                 Storage: file
                 Version: Vault v1.8.0-dev
             Version Sha: 910da26a2b67dff7727fc01a56392bf4affaa2bb

==> Vault server started! Log data will stream in below:

2021-08-24T10:47:42.258-0500 [INFO]  proxy environment: http_proxy="" https_proxy="" no_proxy=""
2021-08-24T10:47:47.681-0500 [INFO]  core.cluster-listener.tcp: starting listener: listener_address=0.0.0.0:8201
2021-08-24T10:47:47.681-0500 [INFO]  core.cluster-listener: serving cluster requests: cluster_listen_address=[::]:8201
2021-08-24T10:47:47.682-0500 [INFO]  core: post-unseal setup starting
2021-08-24T10:47:47.685-0500 [INFO]  core: loaded wrapping token key
2021-08-24T10:47:47.685-0500 [INFO]  core: successfully setup plugin catalog: plugin-directory=""
2021-08-24T10:47:47.690-0500 [INFO]  core: successfully mounted backend: type=system path=sys/
2021-08-24T10:47:47.692-0500 [INFO]  core: successfully mounted backend: type=identity path=identity/
2021-08-24T10:47:47.693-0500 [INFO]  core: successfully mounted backend: type=transit path=transit/
2021-08-24T10:47:47.693-0500 [INFO]  core: successfully mounted backend: type=cubbyhole path=cubbyhole/
2021-08-24T10:47:47.700-0500 [INFO]  core: successfully enabled credential backend: type=token path=token/
2021-08-24T10:47:47.701-0500 [INFO]  rollback: starting rollback manager
2021-08-24T10:47:47.701-0500 [INFO]  core: restoring leases
2021-08-24T10:47:47.703-0500 [INFO]  identity: entities restored
2021-08-24T10:47:47.703-0500 [INFO]  identity: groups restored
2021-08-24T10:47:47.704-0500 [INFO]  expiration: lease restore complete
2021-08-24T10:47:47.706-0500 [INFO]  core: post-unseal setup complete
2021-08-24T10:47:47.706-0500 [INFO]  core: vault is unsealed

----- ---------

**_Verify that cache size is set to 10 after restarting vault:_**

divyapola@divyapola-C02CF744MD6R vault % vault operator unseal                  
Unseal Key (will be hidden): 
Key             Value
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    1
Threshold       1
Version         1.8.0-dev
Storage Type    file
Cluster Name    vault-cluster-74d7615e
Cluster ID      9a05ee54-e2a3-f709-b33e-c18f27513c4a
HA Enabled      false

divyapola@divyapola-C02CF744MD6R vault % vault read transit/cache-config
**Key     Value
size    10**

----- ---------

**_Set cache size to 0:_**

divyapola@divyapola-C02CF744MD6R vault % vault write transit/cache-config size=0

  * cache configurations will be applied when this backend is restarted

divyapola@divyapola-C02CF744MD6R vault % vault read transit/cache-config        
**Key     Value
size    0**

----- ---------

**_Set the cache size to value greater than 10:_**

divyapola@divyapola-C02CF744MD6R vault % vault write transit/cache-config size=24

  * cache configurations will be applied when this backend is restarted

divyapola@divyapola-C02CF744MD6R vault % vault read transit/cache-config         
**Key     Value
size    24**

----- ---------


builtin/logical/transit/backend.go
```diff
@@ -68,6 +68,12 @@ func Backend(ctx context.Context, conf *logical.BackendConfig) (*backend, error)
 		if err != nil {
 			return nil, fmt.Errorf("Error retrieving cache size from storage: %w", err)
 		}
+
+		minCacheSize := 10
```
sgmiller, line 72: Since this is a constant, break it out as a Go "const" at the beginning of the file.  You can find examples of this in other code.

builtin/logical/transit/backend.go
```diff
@@ -68,6 +68,12 @@ func Backend(ctx context.Context, conf *logical.BackendConfig) (*backend, error)
...
+		if cacheSize != 0 && cacheSize < minCacheSize {
+			b.Logger().Warn("size %d is less than default minimum %d. Cache size is set to %d", cacheSize, minCacheSize, minCacheSize)
```
sgmiller, line 74: strike "default", since the minimum can't be changed.

builtin/logical/transit/path_cache_config.go
```diff
@@ -48,6 +48,12 @@ func (b *backend) pathCacheConfigWrite(ctx context.Context, req *logical.Request
 	if cacheSize < 0 {
 		return logical.ErrorResponse("size must be greater or equal to 0"), logical.ErrInvalidRequest
 	}
+    minCacheSize := 10
+    var resp = &logical.Response{}
+    if cacheSize != 0 && cacheSize < minCacheSize {
+		resp.AddWarning(fmt.Sprintf("size %d is less than default minimum %d. Cache size is set to %d", cacheSize, minCacheSize, minCacheSize))
```
sgmiller, line 54: In the case of config, lets actually error out if they try to set a too small value.

sgmiller (changes requested):
Couple small points only.

sgmiller:
And you should add a unit test for this behavior too.  You can add it to the existing TestTransit_CacheConfig function.

website/content/api-docs/secret/transit.mdx
```diff
@@ -1307,7 +1307,7 @@ using the [`/sys/plugins/reload/backend`][sys-plugin-reload-backend] endpoint.
 
 - `size` `(int: 0)` - Specifies the size in terms of number of entries. A size of
   `0` means unlimited. A _Least Recently Used_ (LRU) caching strategy is used for a
-  non-zero cache size.
+  non-zero cache size. Must be 0 (default) or a value greater or equal to 10 (minimum cache size).
```
victorr, line 1310: Maybe you can make it look like the `size` parameter above, including explaining what 0 means?

victorr (approved):
I added a small suggestion.

Good job!

mladlow:
@divyapola5 we're aiming to include this in Vault 1.8.3, so once this PR has merged we'll need to backport it to the `release/1.8.x` branch.

ncabatoff:
Fixes #12064.

ncabatoff:
This change solves the immediate issue.  From what I can see that was: 
* patchCacheConfigWrite stores the new cache size without making use of it
* at startup we read the non-zero configured cache size from storage and pass it to NewTransitLRU
* NewTransitLRU calls lru.New2Q(size), and uses int(size/2) for one of its internal LRU caches, which works out to 0, which is invalid

There's still the bug that patchCacheConfigWrite isn't making use of the updated cache size, so it only takes effect after a restart.  It feels like we should fix that too, else conceivably some future change to the lru library could re-introduce a variant of this bug.  In other words, failing to apply the change prior to storing the config opens the door to another bug that prevents Vault from coming up.  It's also a bad UX - we don't normally require a vault restart for changes to be applied.

website/content/api-docs/secret/transit.mdx
victorr: Hmm, I was confused thinking that the new text was for a different parameter, so please disregard my comment.

divyapola5:
> This change solves the immediate issue. From what I can see that was:
> 
> * patchCacheConfigWrite stores the new cache size without making use of it
> * at startup we read the non-zero configured cache size from storage and pass it to NewTransitLRU
> * NewTransitLRU calls lru.New2Q(size), and uses int(size/2) for one of its internal LRU caches, which works out to 0 if size==1, which is invalid
> 
> There's still the bug that patchCacheConfigWrite isn't making use of the updated cache size, so it only takes effect after a restart. It feels like we should fix that too, else conceivably some future change to the lru library could re-introduce a variant of this bug. In other words, failing to apply the change prior to storing the config opens the door to another bug that prevents Vault from coming up. It's also a bad UX - we don't normally require a vault restart for changes to be applied.

builtin/logical/transit/path_cache_config.go
```diff
@@ -60,8 +60,11 @@ func (b *backend) pathCacheConfigWrite(ctx context.Context, req *logical.Request
 		return nil, err
 	}
 
-	resp := &logical.Response{
-		Warnings: []string{"cache configurations will be applied when this backend is restarted"},
+	resp := &logical.Response{}
+
+	b.lm, err = keysutil.NewLockManager(true, cacheSize)
```
ncabatoff, line 65: I remember somewhere Scott pointing out that we shouldn't be replacing the locks manager and thereby losing track of any existing locks.  Instead, we should be replacing the LockManager `cache` while maintaining the `keyLocks` member.  Not sure what happened to that comment, but anyway, he's right.

divyapola5: Yes, that was in a slack thread. I will address that in my next commit. Thanks :)

builtin/logical/transit/backend.go
```diff
@@ -100,13 +113,45 @@ func GetCacheSizeFromStorage(ctx context.Context, s logical.Storage) (int, error
 	return size, nil
 }
 
-func (b *backend) invalidate(_ context.Context, key string) {
+// Update cache size and get policy
+func (b *backend) GetPolicy(ctx context.Context, polReq keysutil.PolicyRequest, rand io.Reader) (retP *keysutil.Policy, retUpserted bool, retErr error) {
+	if b.lm.GetUseCache() && b.cacheSizeChanged {
```
sgmiller, line 118: Since cacheSizeChanged can be written in a different thread, you should protect the reading of it with the configMutex as well.  Take a read lock where it's read, and a write lock where it's written.  Take care to unlock the read half before you take the write lock to change it.

builtin/logical/transit/backend.go
```diff
@@ -100,13 +113,45 @@ func GetCacheSizeFromStorage(ctx context.Context, s logical.Storage) (int, error
...
+		if currentCacheSize != storedCacheSize {
+			err = b.lm.RefreshCache(storedCacheSize)
+			if err != nil {
+				return nil, false, err
+			}
+		}
+		// Acquire the lock to modify cacheSizeChanged
+		b.configMutex.Lock()
+		defer b.configMutex.Unlock()
+		b.cacheSizeChanged = false
```
sgmiller, line 134: There is a bug here where if the invalidation happens, but someone calls GetPolicy before the storage change actually arrives over replication, the cache sizes won't have changed but you'll have cleared cacheSizeChanged.  Only reset this if you changed the size.

sdk/helper/keysutil/lock_manager.go
```diff
@@ -101,6 +101,24 @@ func (lm *LockManager) InvalidatePolicy(name string) {
 	}
 }
 
+func (lm *LockManager) RefreshCache(cacheSize int) error {
```
sgmiller, line 104: Can we call this InitCache or something?  Refresh kind of implies you're refreshing the cached data somehow.

sgmiller (changes requested):
A couple of requests plus this needs a Changelog entry, but otherwise looks good!

sdk/helper/keysutil/lock_manager.go
divyapola5: Yes, that makes sense.

sgmiller (approved):

PR #12421, opened 2021-08-24
Vault-2840 Add no-op check for HTTP listener validity

rculpepper:
Adds a validity check for HTTP listeners. Currently the check is just a no-op


command/config/validate_listener.go
```diff
@@ -0,0 +1,7 @@
+package config
```
sgmiller, line 1: This will need to match the one on ENT exactly, so go ahead and add the build tag check here. One other way to do this in dev is to get the OSS side approved and merged, then merge oss->ent to pick up what you need ENT side, then get that one reviewed.  In this case it's fine just to have the file be identical

command/server.go
```diff
@@ -2623,6 +2624,11 @@ func startHttpServers(c *ServerCommand, core *vault.Core, config *server.Config,
 		if ln.Config == nil {
 			return fmt.Errorf("Found nil listener config after parsing")
 		}
+
```
sgmiller, line 2627: Ditto this one.  There is a whitespace difference that will have a conflict when merging oss->ent.

sgmiller (changes requested):

sgmiller (approved):

PR #12422, opened 2021-08-24
Client count config view

arnav28:
- Switched to toggle button from checkbox and updated the design
- Switched to ember octane
- Update ember concurrency dependency

### DIFF
--- a/ui/app/components/pricing-metrics-config.js
+++ b/ui/app/components/pricing-metrics-config.js
@@ -10,19 +10,20 @@
  * @param {string} [mode=show] - mode is either show or edit. Show results in a table with the config, show has a form.
  */
 
+import Component from '@glimmer/component';
 import { computed } from '@ember/object';
-import Component from '@ember/component';
+import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
 import { task } from 'ember-concurrency';
 
-export default Component.extend({
-  router: service(),
-  mode: 'show',
-  model: null,
+export default class PricingMetricsConfigComponent extends Component {
+  @service router;
+  @tracked mode = 'show';
+  @tracked modalOpen = false;
+  error = null;
 
-  error: null,
-  modalOpen: false,
-  infoRows: computed(function() {
+  get infoRows() {
     return [
       {
         label: 'Usage data collection',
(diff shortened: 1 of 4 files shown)


arnav28:
Screen Shot 2021-08-24 at 2 17 09 PM
Screen Shot 2021-08-24 at 2 17 17 PM

ui/app/components/pricing-metrics-config.js
```diff
@@ -40,35 +41,42 @@ export default Component.extend({
         valueKey: 'defaultReportMonths',
       },
     ];
-  }),
-  modalTitle: computed('model.enabled', function() {
+  }
+
+  @computed('args.model.enabled')
```
Monkeychip, line 46: I didn't know computed was a thing in octane?

ui/app/templates/components/pricing-metrics-config.hbs
```diff
@@ -1,13 +1,47 @@
...
+              <label for="{{attr.name}}" class="is-label">
+                {{ attr.options.label }}
+              </label>
+              {{#if attr.options.subText}}
+                <p class="sub-text">{{attr.options.subText}} {{#if attr.options.docLink}}<a href="{{attr.options.docLink}}" target="_blank" rel="noopener noreferrer">See our documentation</a> for help.{{/if}}</p>
+              {{/if}}
+              <div class="control">
+                <input 
+                  data-test-field
+                  id={{attr.name}}
```
Monkeychip, line 33: no chance there would be two of the same fields on the same page?  e.g. two with the same attr.name?

ui/app/templates/components/pricing-metrics-config.hbs
```diff
@@ -1,13 +1,47 @@
...
+                  disabled={{eq @model.enabled "Off"}}
+                  readonly={{isReadOnly}}
```
Monkeychip, line 35: is `isReadOnly` a local property (e.g. this.isReadOnly) or something passed in (e.g. @isReadOnly). With glimmer at least we should always see this or @ prefixing varaibles, unless their helpers.  I _think_ that's the standard, but correct me if I'm wrong.

ui/tests/integration/components/pricing-metrics-config-test.js
```diff
@@ -76,8 +76,9 @@ module('Integration | Component | pricing-metrics-config', function(hooks) {
     `);
 
     assert.dom('[data-test-pricing-metrics-config-form]').exists('Pricing metrics config form exists');
-    const fields = document.querySelectorAll('[data-test-field] input');
-    assert.equal(fields.length, 3, 'renders 3 input fields');
+    const fields = document.querySelectorAll('[data-test-field]');
+    console.log(fields);
```
Monkeychip, line 80: left over console.log

Monkeychip (approved):
Small couple of comments, but all nits.
Nice work.

ui/app/templates/components/pricing-metrics-config.hbs
arnav28: True. Good catch, in fact we don't need readonly attribute for these inputs.

ui/app/templates/components/pricing-metrics-config.hbs
arnav28: Nope, should not happen.

ui/app/components/pricing-metrics-config.js
arnav28: While the computed decorator is available in octane, I don't think its necessary in this case. I will remove it.

PR #12425, opened 2021-08-25
Expose secret_id_accessor as WrappedAccessor when wrapping secret-id creation.

### LINKED ISSUE #12173: Return wrapped accessor when creating wrapped AppRole secret-ids
When provisioning an AppRole secret-id, response wrapping is a recommended best practice to avoid exposure of the actual secret-id to intermediate provisioning systems as per eg https://learn.hashicorp.com/tutorials/vault/approle-best-practices?in=vault/auth-methods#approle-response-wrapping

The problem is when using response wrapping, there is no way for the provisioning system to determine the accessor of the newly created secret-id, making lifecycling of the secret-id nearly impossible (ie, destroying the secret-id when it is no longer needed).

A solution already exists when creating response-wrapped tokens, in addition to the wrapping_token and wrapping_accessor fields, a wrapped_accessor field is also returned in the response that contains the accessor for the token itself. Extending this pattern to also work when creating AppRole secret-id's would allow the same pattern to be used here as well.

Not using response wrapping, but this exposes the sensitive secret-id to the provisioning system.
Using auth tokens directly, but these are otherwise more difficult to manage than AppRole credentials which generally seem to be the preferred option.

The same sort of pattern might also make sense for some other auth methods too, but I'm not familiar enough to say for certain one way or the other.

Likewise for secrets engines that return a lease_id a similar wrapped_lease_id field could make sense too.

**Additional context**
https://github.com/hashicorp/terraform-provider-vault/issues/976 mentions terraform leaking secret-id's when response wrapping is used. I can't see a way to resolve that issue without a change in Vault to expose the wrapped accessor.

melbit-michaelw:
Resolves #12173 for Secret-Ids. 

I'm not a Go developer (nor indeed any kind of developer), and was unable to find the right place to add a test for this functionality. Could you please point me in the right direction as to where this should be tested ?

### DIFF
--- a/vault/external_tests/approle/wrapped_secretid_test.go
+++ b/vault/external_tests/approle/wrapped_secretid_test.go
@@ -1,0 +1,121 @@
+package approle
+
+import (
+	"testing"
+
+	log "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/vault/api"
+	credAppRole "github.com/hashicorp/vault/builtin/credential/approle"
+	vaulthttp "github.com/hashicorp/vault/http"
+	"github.com/hashicorp/vault/sdk/logical"
+	"github.com/hashicorp/vault/vault"
+	"github.com/stretchr/testify/require"
+)
+
+func TestApproleSecretId_Wrapped(t *testing.T) {
+
+	var err error
+	coreConfig := &vault.CoreConfig{
+		DisableMlock: true,
+		DisableCache: true,
+		Logger:       log.NewNullLogger(),
+		CredentialBackends: map[string]logical.Factory{
+			"approle": credAppRole.Factory,
+		},
+	}
+
+	cluster := vault.NewTestCluster(t, coreConfig, &vault.TestClusterOptions{
+		HandlerFunc: vaulthttp.Handler,
+	})
+
+	cluster.Start()
+	defer cluster.Cleanup()
+
+	cores := cluster.Cores
+
+	vault.TestWaitActive(t, cores[0].Core)
+
+	client := cores[0].Client
+	client.SetToken(cluster.RootToken)
+
+	err = client.Sys().EnableAuthWithOptions("approle", &api.EnableAuthOptions{
+		Type: "approle",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = client.Logical().Write("auth/approle/role/test-role-1", map[string]interface{}{
+		"name": "test-role-1",
+	})
+	require.NoError(t, err)
+
+	client.SetWrappingLookupFunc(func(operation, path string) string {
+		return "5m"
+	})
+
+	resp, err := client.Logical().Write("/auth/approle/role/test-role-1/secret-id", map[string]interface{}{})
+	require.NoError(t, err)
+
+	wrappedAccessor := resp.WrapInfo.WrappedAccessor
+	wrappingToken := resp.WrapInfo.Token
+
+	client.SetWrappingLookupFunc(func(operation, path string) string {
+		return api.DefaultWrappingLookupFunc(operation, path)
+	})
+
+	unwrappedSecretid, err := client.Logical().Unwrap(wrappingToken)
+	unwrappedAccessor := unwrappedSecretid.Data["secret_id_accessor"].(string)
+
+	if wrappedAccessor != unwrappedAccessor {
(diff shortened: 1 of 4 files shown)


pmmukh:
Hi @melbit-michaelw ! Thanks for submitting this PR! For adding a test for this functionality, I think you are good to skip unit tests, as we don't really have unit tests on wrapping.go. For integration tests, I think adding a new folder for `approle` under https://github.com/hashicorp/vault/tree/main/vault/external_tests, and adding a test there that exercises setting up a wrapped secret and uses the accessor in the response like how https://github.com/hashicorp/terraform-provider-vault/issues/976 plans to use it, would make sense.

vault/wrapping.go
```diff
@@ -184,6 +184,11 @@ DONELISTHANDLING:
 		resp.WrapInfo.WrappedAccessor = resp.Auth.Accessor
 	}
 
+	// Make secret_id_accessor available as a WrappedAccessor.
```
pmmukh, line 187:
```suggestion
	// Make secret_id_accessor available as a WrappedAccessor.
	// Store the accessor of the approle secret in WrappedAccessor
```

vault/wrapping.go
```diff
@@ -184,6 +184,11 @@ DONELISTHANDLING:
...
+	if secretIdAccessor, ok := resp.Data["secret_id_accessor"]; ok {
```
pmmukh, line 188: would it make sense to be a bit defensive and make this check `ok && resp.Auth == nil` , to be certain we don't overwrite token accessors? ( Unlikely as it is to be an actual problem )

pmmukh:
Think this comment https://github.com/hashicorp/vault/blob/main/sdk/helper/wrapping/wrapinfo.go#L20-L21 might need to be updated, to reflect the dual nature of WrappedAccessor with this change.

vault/wrapping.go
melbit-michaelw: That sounds sensible to me. Added it.

melbit-michaelw:
Thank you very much for the feedback. I've done the minor fixes suggested, and will start looking at the integration testing.

melbit-michaelw:
@pmmukh  - Added tests to confirm WrappedAccessor set as expected, and not set when wrapping isn't performed.

Let me know if anything else is required.

pmmukh:
hey! Just wanted to drop a note that there's one little point we're hashing out internally on this change, after which I'll give this another pass sometime this week!

vault/wrapping.go
```diff
@@ -184,6 +184,11 @@ DONELISTHANDLING:
 		resp.WrapInfo.WrappedAccessor = resp.Auth.Accessor
 	}
 
+	// Store the accessor of the approle secret in WrappedAccessor
+	if secretIdAccessor, ok := resp.Data["secret_id_accessor"]; ok && resp.Auth == nil {
```
pmmukh, line 188: Another check that we discussed and would be nice to have, is a [logical.Request.MountType](https://github.com/hashicorp/vault/blob/main/sdk/logical/request.go#L150) == "approle" check here, to be sure we don't do this for a different backend that also uses a field called `secret_id_accessor`. wdyt?

vault/external_tests/approle/wrapped_secretid_test.go
```diff
@@ -0,0 +1,121 @@
...
+	_, err = client.Logical().Write("auth/approle/role/test-role-1", map[string]interface{}{
+		"name": "test-role-1",
+	})
+	require.NoError(t, err)
+
+	resp, err := client.Logical().Write("/auth/approle/role/test-role-1/secret-id", map[string]interface{}{})
+	require.NoError(t, err)
+
+	if resp.WrapInfo != nil && resp.WrapInfo.WrappedAccessor != "" {
+		t.Fatalf("WrappedAccessor unexpectedly set")
```
pmmukh, line 118: nit, I think this should be t.Fatal as the formatting isn't required

pmmukh: actually ignore this, on second thoughts its a bit too nit-ty, we do this in several places so its fine imo

vault/wrapping.go
melbit-michaelw: That seems sensible.

pmmukh (approved):
lgtm! Thanks for the quick responses on the feedback!

PR #12428, opened 2021-08-25
Add PutAutoPilotRaftConfiguration to api

### LINKED ISSUE #12427: API lacks capability to modify Autopilot configuration
The API package currently only allows for retrieval of the autopilot configuration but doesn't expose any capabilities to modify it.

Add missing API call for `PutRaftAutoPilotConfiguration`

Users are currently forced to use the HTTP API instead, see [here](https://www.vaultproject.io/api-docs/system/storage/raftautopilot#set-configuration)

elsesiy:
The previous implementation only allows retrieving the autopilot state but not modifying it, hence the missing method has been added.

Misc:
- Update api to go1.16
- Update dependencies

Fix #12427

### DIFF
--- a/api/sys_raft.go
+++ b/api/sys_raft.go
@@ -368,3 +368,22 @@ func (c *Sys) RaftAutopilotConfiguration() (*AutopilotConfig, error) {
 
 	return &result, err
 }
+
+// PutRaftAutopilotConfiguration allows modifying the raft autopilot configuration
+func (c *Sys) PutRaftAutopilotConfiguration(opts *AutopilotConfig) error {
+	r := c.c.NewRequest("POST", "/v1/sys/storage/raft/autopilot/configuration")
+
+	if err := r.SetJSONBody(opts); err != nil {
+		return err
+	}
+
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	resp, err := c.c.RawRequestWithContext(ctx, r)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	return nil
+}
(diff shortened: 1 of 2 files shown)


ncabatoff (commented):
The new method looks fine.  Can you revert the other bits?  We can't rename public methods without incrementing major version, since that's a breaking change.  I don't want think that's worth it at this point.  Also, I don't see any reason to include module updates (changes to go.mod/go.sum) in this PR.

hsimon-hashicorp:
Hi @elsesiy - if you can take a look at @ncabatoff's comment and update and let us know you've done so, we'll get this reviewed and merged. Thanks so much!

elsesiy:
@hsimon-hashicorp @ncabatoff I updated the PR according to the comments. Should I submit a separate one for the go toolchain/module changes?

ncabatoff:
> @hsimon-hashicorp @ncabatoff I updated the PR according to the comments. Should I submit a separate one for the go toolchain/module changes?

Sorry, I've already forgotten what they were. :)

Going by
> Update api to go1.16
> Update dependencies

my hunch is no, since we don't want the api to impose go 1.16 (unless you know of a reason why it should - I don't, but I may be missing something) and we don't update dependencies automatically just because they're out of date.  But feel free to open a PR for them if you'd like us to take a closer look.

ncabatoff (approved):

elsesiy:
@ncabatoff When will this be merged? I see it's tagged for 1.9 but it didn't make it into the rc..

ncabatoff:
@elsesiy sorry about that, stuff came up, and CI is stuck and not running the tests.  I don't know of a good way of getting CI unstuck when this happens... the only thing I've found that works in the past sometimes is to add a new commit.  In this case the change is purely additive, so I decided to test that it doesn't break anything myself, and I'll force the commit with admin privs.

ncabatoff:
Thanks for the contribution!  I decided to remove the release label: since this is a change to the client API, there's no reason to backport it, because it won't be compiled in to the Vault binary in any event.

Issue #9400, opened 2020-07-05
Default connect_timeout in cassandra secret plugin is not according to docs


enudler:
**Describe the bug**
When setting up Cassandra secret backend and trying to get creds to Cassandra I'm getting timeouts and errors like this:
```
{
    "errors": [
        "1 error occurred:\n\t* read tcp 192.168.80.3:49848->192.168.80.2:9042: i/o timeout\n\n"
    ]
}

 
```

I suspect that the default "connect_timeout" is not a string of "5s" but an integer with a value of 5.

**To Reproduce**
1. mount secret backend of type 'database'
2. create a role with creation statement
3. configure a connection to Cassandra without "connect_timeout" (/v1/database/config/{dbName})
4. get creds "v1/database/creds/{roleName}"
5. see the error (i/o timeout)
6. configure new connection to Cassandra this time with default "connect_timeout" from the docs which is 5s
7. get creds "v1/database/creds/{roleName}"
8. all good, creds returned 
**Expected behavior**
no timeouts

**Environment:**
* Vault Server Version: 1.4.2
* Server Operating System/Architecture: vault 1.4.2 running in dev mode (docker) and  Cassandra (docker)

aphorise:
This was addressed in: [fix cassandra db plugin timeout to 5s as in docs #12443](https://github.com/hashicorp/vault/pull/12443)
Hey @enudler - thanks for reporting this.

Issue #8527, opened 2020-03-10
Cassandra database plugin intermittent timeouts connection to Cassandra


nmylesnbx:
**Describe the bug**
I'm getting intermittent timeout errors from the Cassandra database plugin connecting to Cassandra. This happens when calling `database/creds/<role>`. Most of the time it works but I'm hitting this every 10 or so calls to get new database credentials.

hvac python stacktrace:
```
File "/usr/local/lib/python3.7/site-packages/hvac/v1/__init__.py", line 195, in read
   return self._adapter.get('/v1/{0}'.format(path), wrap_ttl=wrap_ttl).json()
File "/usr/local/lib/python3.7/site-packages/hvac/adapters.py", line 93, in get
   return self.request('get', url, **kwargs)
File "/usr/local/lib/python3.7/site-packages/hvac/adapters.py", line 276, in request
  utils.raise_for_error(response.status_code, text, errors=errors)
File "/usr/local/lib/python3.7/site-packages/hvac/utils.py", line 39, in raise_for_error
   raise exceptions.InternalServerError(message, errors=errors)
hvac.exceptions.InternalServerError: 1 error occurred:
  * read tcp 192.168.144.4:37610->192.168.144.2:9042: i/o timeout
```

**To Reproduce**
1. Run `vault secrets enable database`
2. Run:
```
vault write database/config/cassandra \
    plugin_name="cassandra-database-plugin" \
    hosts=cassandra \
    protocol_version=4 \
    username=username \
    password=password \
    allowed_roles='*'
```
3. Run:
```
vault write database/roles/role \
    db_name=cassandra \
    creation_statements="CREATE USER '{{username}}' WITH PASSWORD '{{password}}' NOSUPERUSER; \
          GRANT SELECT ON ALL KEYSPACES TO {{username}};" \
    default_ttl="1h" \
    max_ttl="24h"
```
4. Call API multiple times to retrieve creds: `GET http://vault:8200/v1/database/creds/role`
3. See intermittent 500 errors

**Expected behavior**
I would expect timeouts to be retried to avoid intermittent 500's

**Environment:**
This happens both locally (docker-compose) and in a deployed environment 
* Vault Server Version (retrieve with `vault status`): 1.2.2
* Vault CLI Version (retrieve with `vault version`): 1.2.2
* Server Operating System/Architecture: official vault docker image

Vault cassandra plugin configuration was something like this:
```
vault write database/config/cassandra \
    plugin_name="cassandra-database-plugin" \
[...]

PR #12443, opened 2021-08-26
fix cassandra db plugin timeout to 5s as in docs

viljoviitanen:
Documentation says timeout is 5s, but code uses 0s, which is too short any any real world usage, causing issues.
https://www.vaultproject.io/api/secret/databases/cassandra#connect_timeout
issues: https://github.com/hashicorp/vault/issues/8527 https://github.com/hashicorp/vault/issues/9400


imthaghost (approved):
Looks great thanks for the submission!

PR #12463, opened 2021-08-31
improve "x_forwarded_for_hop_skips" example

noelprince:
Currently the example given results in 2.3.4.5 if it is indexed from other side. This new example prevents confusion because it is now clear which side x_forwarded_for_hop_skips is indexing from


akshya96 (approved):

mladlow:
@hsimon-hashicorp this was approved and all the checks have run, so I've merged it, but the docs-cherrypick label won't work because it's a fork. Would you mind cherry picking this to stable website please?

PR #12475, opened 2021-09-01
Allow perf standbys to create batch tokens [vault-1793]

HridoyRoy:
What sort of tests can we add to the handler to ensure that create/token requests aren't being forwarded n a cluster? Is there something in the request that tells us a request has been forwarded, that we can inspect via external test?

Tests I'm planning to add:
- Ensure that "type" can be parsed properly from the request
- Ensure that `TokenTypeBatch` works properly with tokens of type "batch" and tokens of type "service", and that the type is effectively ignored when a role is being used. 

--- 

Also was looking for some preliminary feedback on the code itself to make sure I wasn't missing anything. 
Thanks!


hghaf099:
Closing this in favour of https://github.com/hashicorp/vault-enterprise/pull/3179

PR #12478, opened 2021-09-01
UI: Fix missing nav links on namespace login

hashishaw:
Fixes an issue due to an Ember bug with sticky query params which was fixed in [Ember 3.24](https://github.com/emberjs/ember.js/pull/19448). Since we need to backport this fix to 1.8.x, we're going to go with the workaround outlined in [this comment](https://github.com/emberjs/ember.js/pull/19249#issuecomment-790479034) and then revert and upgrade to Ember 3.24 for 1.9 and forward.

**BEFORE**
nav-before

**AFTER**
nav-after

### DIFF
--- a/changelog/12478.txt
+++ b/changelog/12478.txt
@@ -1,0 +1,3 @@
+```release-note:fix
+ui: fix missing navbar items on login to namespace
+```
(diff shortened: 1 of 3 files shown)


Monkeychip (approved):
Nice work on investigating this.

Issue #12397, opened 2021-08-22
auth > tune readme file -- add information how to view configuration


tograla:
The [auth - tune readme file](https://github.com/hashicorp/vault/blob/main/website/content/docs/commands/auth/tune.mdx) ([official documenation](https://www.vaultproject.io/docs/commands/auth/tune)) does not show how to view configuration prior and post editing.

**Describe the solution you'd like**
I suggest to add information how to view auth configuration. It really wasn't easy and took me a decent while to figure this out but eventually I managed to confirm my settings using the following commands:
Firstly the most verbose output formatted in the default table fashion:
```
$ vault read /sys/auth
Key          Value
---          -----
approle/     map[accessor:auth_approle_48059fa4 config:map[default_lease_ttl:0 force_no_cache:false max_lease_ttl:0 token_type:default-service] description: external_entropy_access:false local:false options:<nil> seal_wrap:false type:approle uuid:e73f9405-4e50-ea19-9c73-228a47c1da73]
token/       map[accessor:auth_token_290aaf74 config:map[default_lease_ttl:0 force_no_cache:false max_lease_ttl:0 token_type:default-service] description:token based credentials external_entropy_access:false local:false options:<nil> seal_wrap:false type:token uuid:21bb6a05-d330-53ee-572d-ba20f83944d0]
userpass/    map[accessor:auth_userpass_8fac179a config:map[default_lease_ttl:3600 force_no_cache:false max_lease_ttl:14400 token_type:default-service] description: external_entropy_access:false local:false options:<nil> seal_wrap:false type:userpass uuid:1e85c903-a948-5f16-b350-f034ba514818]
$
```
Works but isn't very pretty, so I started experimenting with ```-format=json```, and here is what I managed to get:

The below lists ```type``` of every auth method enabled and then lists the ```.config``` section for each. This is more useable than the above ```table``` format providing you have no more than a handful of methods enabled.
```
$ vault read -format=json /sys/auth/ | jq '.data[].type, .data[].config'
"approle"
"token"
"userpass"
{
  "default_lease_ttl": 0,
  "force_no_cache": false,
  "max_lease_ttl": 0,
  "token_type": "default-service"
}
{
  "default_lease_ttl": 0,
  "force_no_cache": false,
  "max_lease_ttl": 0,
  "token_type": "default-service"
}
{
  "default_lease_ttl": 3600,
  "force_no_cache": false,
  "max_lease_ttl": 14400,
  "token_type": "default-service"
}
```
[...]

PR #12479, opened 2021-09-02
Fix issue 12397

yhyakuna:
This PR fixes the reported issue: https://github.com/hashicorp/vault/issues/12397

🔍 [Deploy Preview: CLI doc](https://vault-git-fix-issue-12397-hashicorp.vercel.app/docs/commands/auth/tune)

I also updated the sample HTTP response in the API doc.

🔍 [Deploy Preview: API doc](https://vault-git-fix-issue-12397-hashicorp.vercel.app/api-docs/system/auth#sample-response-1)

### DIFF
--- a/website/content/api-docs/system/auth.mdx
+++ b/website/content/api-docs/system/auth.mdx
@@ -188,8 +188,11 @@ $ curl \
 
 ```json
 {
-  "default_lease_ttl": 3600,
-  "max_lease_ttl": 7200
+  "default_lease_ttl": 2764800,
+  "description": "",
+  "force_no_cache": false,
+  "max_lease_ttl": 2764800,
+  "token_type": "default-service"
 }
 ```
 
--- a/website/content/docs/commands/auth/tune.mdx
+++ b/website/content/docs/commands/auth/tune.mdx
@@ -28,7 +28,8 @@ max_lease_ttl        768h
 token_type           default-service
 ```
 
-Tune the default lease for the auth method enabled at "github/" to be 72 hours.
+The default lease for the auth method enabled at "github/" is currently set to
+768 hours. Tune this value to 72 hours.
 
 ```shell-session
 $ vault auth tune -default-lease-ttl=72h github/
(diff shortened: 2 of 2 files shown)


website/content/api-docs/system/auth.mdx
```diff
@@ -188,8 +188,11 @@ $ curl \
 
 ```json
 {
-  "default_lease_ttl": 3600,
-  "max_lease_ttl": 7200
+  "default_lease_ttl": 2764800,
```
pmmukh, line 191: I'm actually getting a response like this, when I try to hit this endpoint, should we add these other fields?
```
{
    "description": "",
    "default_lease_ttl": 2764800,
    "max_lease_ttl": 2764800,
    "force_no_cache": false,
    "token_type": "default-service",
    "request_id": "b814023c-4f0f-cf19-6ef2-fe666037c1bf",
    "lease_id": "",
    "renewable": false,
    "lease_duration": 0,
    "data": {
        "default_lease_ttl": 2764800,
        "description": "",
        "force_no_cache": false,
        "max_lease_ttl": 2764800,
        "token_type": "default-service"
    },
    "wrap_info": null,
    "warnings": null,
    "auth": null
}
```

taoism4504 (approved):
slight change, but otherwise LGTM.

website/content/api-docs/system/auth.mdx
yhyakuna: For the consistency, we should keep it looks like all other API docs.  It seems that the API docs only shows the HTTP response payload as the sample response.

pmmukh: ah yep, that is true, I did notice this isn't the only endpoint with this difference, so that sounds good to me!

pmmukh (approved):

yhyakuna:
Cannot fix the build issue. I opened a new [PR](https://github.com/hashicorp/vault/pull/12484), and closing this one.

PR #12485, opened 2021-09-02
Customizing HTTP headers in the config file

hghaf099:


command/agent.go
```diff
@@ -878,9 +878,10 @@ func (c *AgentCommand) Run(args []string) int {
 func verifyRequestHeader(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if val, ok := r.Header[consts.RequestHeaderName]; !ok || len(val) != 1 || val[0] != "true" {
-			logical.RespondError(w,
-				http.StatusPreconditionFailed,
-				errors.New(fmt.Sprintf("missing '%s' header", consts.RequestHeaderName)))
+			err := errors.New(fmt.Sprintf("missing '%s' header", consts.RequestHeaderName))
```
swayne275, line 881: could be `fmt.Errorf()` since you're doing a format string

command/agent.go
```diff
@@ -878,9 +878,10 @@ func (c *AgentCommand) Run(args []string) int {
...
+			status := http.StatusPreconditionFailed
+			logical.AdjustErrorStatusCode(&status, err)
+			logical.RespondError(w,	status,	err)
```
swayne275, line 884:
```suggestion
			logical.RespondError(w, status, err)
```

strange, seems like something the ide should have done

http/logical.go
```diff
@@ -482,7 +497,9 @@ WRITE_RESPONSE:
 	if cacheControl, ok := resp.Data[logical.HTTPRawCacheControl].(string); ok {
```
chair6, line 497: Should we consider pulling this Cache-Control header into the set of headers that we allow to be customized as well? The default `Cache-Control: no-store` (set at https://github.com/hashicorp/vault/blob/main/http/handler.go#L303-L305) aligns with current Mozilla guidance (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#preventing_caching) and may be acceptable to many Vault operators, but some may prefer different values depending on security requirements or what vulnerability scanner they're trying to appease.

chair6 (commented):
Nice to see this feature under development! Added a question re. the `Cache-Control` header that is currently set separately...

http/logical.go
hghaf099: Yes, indeed it is customizable. Couple of lines down (line 501) will allow setting that if operators included that into the config file.

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,186 @@
...
+    "4xx",
+    "5xx",
+}
+
+const (
+    contentSecurityPolicy = "default-src 'none';   connect-src 'self';  img-src 'self' data:; script-src 'self'; style-src 'unsafe-inline' 'self'; form-action  'none'; frame-ancestors 'none'; font-src 'self'"
+    xXssProtection = "1; mode=block"
+    xFrameOptions = "Deny"
+    xContentTypeOptions = "nosniff"
+    strictTransportSecurity = "max-age=31536000; includeSubDomains"
```
mickael-hc, line 33: This header won't be honored by browsers on HTTP connections but will definitely please the scanners. I think the default is fine as-is

mickael-hc (commented):
thanks for working on this @hghaf099 , this definitely be helpful to many organizations with stricter requirements. I've left some comments inline

command/agent.go
hghaf099: done

command/agent.go
hghaf099: hmm, I see the spaces in my ide. not sure what has happened! :)

http/logical.go
```diff
@@ -268,17 +273,17 @@ func handleLogicalRecovery(raw *vault.RawBackend, token *atomic.String) http.Han
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		req, _, statusCode, err := buildLogicalRequestNoAuth(false, w, r)
 		if err != nil || statusCode != 0 {
-			respondError(w, statusCode, err)
+			respondError(w, statusCode, err, nil)
```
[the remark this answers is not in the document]
hghaf099: yes, these are related to the recovery mode. Setting custom response headers in recovery mode is not a requirement at this stage based on the discussions we had during RFC development.

vault/custom_response_headers.go
```diff
@@ -0,0 +1,278 @@
...
+					logger.Error("Custom headers starting with X-Vault are not valid", "header", h)
+					continue
+				}
+
+				// Checking for UI headers, if any common header exists, we just log an error
+				if uiHeaders != nil {
+					exist := uiHeaders.Get(h)
+					if exist != "" {
+						logger.Warn("found a duplicate header in UI, note that config file headers take precedence.", "header:", h)
+					}
```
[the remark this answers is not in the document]
hghaf099: At this time, the decision was made such that configured headers in the config file take precedence over the ones set in the UI endpoint. We have intentionally moved away from having an API endpoint to set these custom headers. So, if an operator would like to update the response headers they would need to update their configuration file and send a SIGHUP signal to the Vault and the custom headers will be updated with the new configured headers. 
There is also an option to modify "/sys/config/reload" endpoint in such a way so that the custom headers would be reloaded as well. As it stands, the endpoint only supports renewing licensing AFAIK. Updating that endpoint for custom response headers may come with a future PR depending on the time and priorities. 
As for deprecating the /ui endpoint, unfortunately, I don't have sufficient information.

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,186 @@
...
+var ValidCustomStatusCodeCollection = []string {
+    "default",
+    "1xx",
+    "2xx",
+    "3xx",
+    "4xx",
+    "5xx",
+}
+
+const (
```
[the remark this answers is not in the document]
hghaf099: These headers cannot be unset for now, however, if an operator is not happy with the default values, they can simply configure their desired value in the configuration file. I will discuss the idea of providing a way to unset the defaults or even not set any defaults with our team to see what they think about it! Thanks for raising this up.

internalshared/configutil/http_response_headers.go
hghaf099: Yes, that is one of the intentions of including that here :)

vault/logical_system.go
```diff
@@ -2620,9 +2620,17 @@ func (b *SystemBackend) handleConfigUIHeadersUpdate(ctx context.Context, req *lo
...
 
+	if req.ResponseWriter == nil {
+		return logical.ErrorResponse("no ResponseWriter in the request"), logical.ErrInvalidRequest
+	}
+
 	// Translate the list of values to the valid header string
 	value := http.Header{}
 	for _, v := range values {
+		la := req.ResponseWriter.Header().Get("X-Vault-Listener-Add")
+		if b.Core.ExistCustomResponseHeader(header, la) {
```
ncabatoff, line 2631: Shouldn't we refuse to allow setting a conflicting header for any listener, not just the one the update request came over on?

vault/core.go
```diff
@@ -2632,6 +2651,53 @@ func (c *Core) SetConfig(conf *server.Config) {
...
+	if conf == nil {
+		return fmt.Errorf("failed to load core raw config")
+	}
+
+	tempLH := c.customListenerHeader
+	c.customListenerHeader = nil
+
+	uiHeaders, _ := c.UIHeaders()
+
+	customHeaderLogger := c.GetLogger("customHeader")
```
ncabatoff, line 2684: We usually only create a new sub logger for new subsystems.  I'm not sure what the dividing line should be, but e.g. if there's a new background goroutine, I could see that warranting a new logger.  I don't think this feature warrants one.

internalshared/configutil/listener.go
```diff
@@ -99,6 +99,11 @@ type Listener struct {
 	CorsAllowedOrigins    []string    `hcl:"cors_allowed_origins"`
 	CorsAllowedHeaders    []string    `hcl:"-"`
 	CorsAllowedHeadersRaw []string    `hcl:"cors_allowed_headers,alias:cors_allowed_headers"`
+
+	// Custom Http response headers
+	CustomResponseHeaders 	 map[string]map[string]string `hcl:"-"`
+	CustomResponseHeadersRaw interface{}                  `hcl:"custom_response_headers,alias:custom_response_headers"`
```
ncabatoff, line 105: I don't think you need an alias here.

internalshared/configutil/listener.go
```diff
@@ -361,6 +366,18 @@ func ParseListeners(result *SharedConfig, list *ast.ObjectList) error {
 			}
 		}
 
+		// HTTP Headers
+		{
+			if l.CustomResponseHeadersRaw != nil {
+				customHeadersMap, err := ParseCustomResponseHeaders(l.CustomResponseHeadersRaw)
+				if err != nil {
+					return multierror.Prefix(fmt.Errorf("failed to parse custom_response_headers:%w", err), fmt.Sprintf("listeners.%d", i))
```
ncabatoff, line 374:
```suggestion
					return multierror.Prefix(fmt.Errorf("failed to parse custom_response_headers: %w", err), fmt.Sprintf("listeners.%d", i))
```

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    // setting default custom headers
+    de := h["default"]
+    h["default"] = setDefaultResponseHeaders(de)
+
+	return h, nil
+}
+
+func isValidListDict(in interface{}) bool {
```
ncabatoff, line 117:
```suggestion
func isValidListOfMaps(in interface{}) bool {
```

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,186 @@
...
+    if _, ok := in.([]map[string]interface{}); ok {
```
ncabatoff, line 118: You don't really need an `if`, you can just return `ok`.  Though it's more conventional to just inline one-liners like this instead of creating a func.

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,186 @@
...
+
+    return defaults
+}
+
+func ParseCustomResponseHeaders(r interface{}) (map[string]map[string]string, error) {
+    if !isValidListDict(r) {
+        return nil, fmt.Errorf("response headers were not configured correctly. please make sure they're in a map")
+    }
+
+    customResponseHeader := r.([]map[string]interface{})
```
ncabatoff, line 83: Are we guaranteed not to panic here?

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,186 @@
...
+    h := make(map[string]map[string]string)
+
+    for _, crh := range customResponseHeader {
+        for sc, rh := range crh {
```
ncabatoff, line 87: Short var names are ok when it's obvious from local context what the thing refers to.  I don't think it's obvious here, consider using instead `statusCode, header := range crh`.

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,186 @@
...
+            if !isValidListDict(rh){
+                return nil, fmt.Errorf("invalid response header type")
+            }
+
+            if !IsValidStatusCode(sc) {
+                return nil, fmt.Errorf("invalid status code found in the config file: %v", sc)
+            }
+
+            hvl := rh.([]map[string]interface{})
```
ncabatoff, line 96: Again, will this never panic?

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func parseHeaderValues(h interface{}) (string, error) {
+    var sl []string
+    if !isValidList(h) {
+        return "", fmt.Errorf("headers must be given in a list of strings")
+    }
+    vli := h.([]interface{})
+    for _, vh := range vli {
+        sl = append(sl, vh.(string))
```
ncabatoff, line 181: I wonder if we should do any safety checks on the strings, e.g. to make sure they don't contain any invalid header chars like space or semicolon?

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,186 @@
...
+const (
+    contentSecurityPolicy = "default-src 'none';   connect-src 'self';  img-src 'self' data:; script-src 'self'; style-src 'unsafe-inline' 'self'; form-action  'none'; frame-ancestors 'none'; font-src 'self'"
+    xXssProtection = "1; mode=block"
+    xFrameOptions = "Deny"
+    xContentTypeOptions = "nosniff"
+    strictTransportSecurity = "max-age=31536000; includeSubDomains"
+    contentType = "text/plain; charset=utf-8"
+)
+
+func ParseDefaultHeaders(h string) string {
```
ncabatoff, line 37:
```suggestion
func GetDefaultHeaderValue(h string) string {
```

ncabatoff (commented):
I haven't finished reading through the entire thing, but here are some initial comments.

ncabatoff:
One idea I'm considering: instead of passing core.SetCustomResponseHeaders to respondError/etc, just pass the http.Request.  Then we can extract what we need from the request context.

Alternatively, maybe wrapGenericHandler should invoke h.ServeHTTP with a ResponseWriter object that does the header population when WriteHeader(statusCode) is called.  That feels like the cleanest solution, but maybe I'm missing something.

hghaf099:
Regarding wrapping ResponseWriter in such a way that we modify WriteHeader to add the headers, this would indeed be the cleanest solution. I asked about this approach in slack as well https://hashicorp.slack.com/archives/C028HE7SYJW/p1630441384104600

Wrapping ResponseWriter is error prone since not all underlying functions are known or may not be implemented/accessible for the new wrapper. This was learned through another bug where a max size request caused a broken pipe.

ncabatoff:
> Wrapping ResponseWriter is error prone since not all underlying functions are known or may not be implemented/accessible for the new wrapper. This was learned through another bug where a max size request caused a broken pipe.

I remember, but is this case similar?  It feels like all we need is the standard interface methods here.

hghaf099:
Passing http.Request and extracting what we need to set the headers sounds good! Doing that would still need to satisfy some requirements highlighted in the RFC:
1- we need to log a warning/error if a duplicate header exist or being set in the /ui endpoint
2- we need to log a warning/error if the custom response headers are not properly configured

so, basically, we need to have access to some logger, maybe core.logger? what would you recommend here? core.logger is not available during parsing the config file, and I would need to check for the above requirements after the config file was parsed. 

What I can think of right now in terms of implementation is the following pointers:
1- Parsing the config as is, and then create a function in core to sanitize the parsed headers according the above two requirements and log warnings/errors as needed.
2- one complication is with the /ui endpoint when an operator wants to set the header and we would need to check if the header is already set in the config file. There is not a reference to http.Request as it is well deep in logical layer. Though, There is a context object there, but I am not sure if that would contain the same info as we want. It seems that after creating the logical request, the context is changed/modified. This would need to be carefully tracked to make sure we don't loose anything. Alternatively, we could parse the request path and method and do the check early on. So, wondering what you would think about that?

hghaf099:
for wrapping ResponseWriter, there is already a "copyResponseWriter" struct which implements all the standard interface methods, and I just tested it again, and "TestLogical_RequestSizeLimit" fails.

internalshared/configutil/http_response_headers.go
hghaf099: I am struggling with these naming pattern in GO, and I have not figured out yet how to properly use them. I have seen in go that even for very long functions, these short var names are used which make it hard to search what it is. Appreciate the explanation. I will try to follow that!

internalshared/configutil/http_response_headers.go
hghaf099: "isValidListDict(rh)" function is called before this line to make sure a panic does not happen. would it be enough though?

internalshared/configutil/http_response_headers.go
hghaf099: done, made inline check

internalshared/configutil/listener.go
hghaf099: done! just out of curiosity, what is alias used for?

vault/core.go
hghaf099: So, the reason I did this is that we need to log invalid headers like the ones start with "X-Vault" and the conflicting ones with the /ui endpoint. So, should I just pass in core.logger? There are other warnings/errors which are logged. Not sure if all is needed though.

vault/logical_system.go
hghaf099: Yes indeed!

command/agent/cache/handler.go
```diff
@@ -37,8 +37,11 @@ func Handler(ctx context.Context, logger hclog.Logger, proxier Proxier, inmemSin
 		// Parse and reset body.
 		reqBody, err := ioutil.ReadAll(r.Body)
 		if err != nil {
-			logger.Error("failed to read request body")
-			logical.RespondError(w, http.StatusInternalServerError, errors.New("failed to read request body"))
+			errRet := errors.New("failed to read request body")
```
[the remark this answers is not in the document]
hghaf099: so, logical.RespondError calls logical. AdjustErrorStatusCode in which the status code might change. I moved AdjustErrorStatusCode out of logical.RespondError so that I would not need to pass in a core reference to the logical package as it could cause circular dependency.

http/handler.go
```diff
@@ -334,19 +335,27 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
 			w.Header().Set("X-Vault-Hostname", hostname)
 		}
 
+		// Setting the listener address as a header so that we could set customized
+		// headers configured in the corresponding listener stanza
+		var la string
+		if props.ListenerConfig != nil {
+			la = props.ListenerConfig.Address
+		}
+		w.Header().Set("X-Vault-Listener-Add", la)
```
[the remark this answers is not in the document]
hghaf099: comments related to this are in the PR conversations.

command/agent/cache/handler.go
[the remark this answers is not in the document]
ncabatoff: Sorry, I don't follow.  How is AdjustErrorStatusCode dependent on Core?

internalshared/configutil/http_response_headers.go
ncabatoff: Ah, true.

command/agent/cache/handler.go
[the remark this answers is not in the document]
hghaf099: Sorry for my bad explanation. So, AdjustErrorStatusCode does not depend on Core. However, to set the headers based on a status code, I need to first call AdjustErrorStatusCode and then call the function that sets the headers. The main reason I move AdjustErrorStatusCode out of logical.RespondError is such that I would not need to pass in that function anything extra specially since it is in the sdk/logical package. 
A little bit more context, so, the way that setting custom headers was implemented before was that I passed in a reference to Core and through that I called the SetCustomResponseHeaders. This would mean that I would need to pass in Core to the logical.RespondError which caused circular dependency. However, now I send the core.SetCustomResponseHeaders function itself, and I can define a function type to avoid the circular dependency issue. But, wondering would the change (moving out the AdjustErrorStatusCode out of logical.RespondError) still a bad Idea? 
In any case, I think the decision on this depends on the other comments about wrapping ResponseWriter, or including the headers in the http request context.

hghaf099:
@Nick, I have incorporated your comment on passing http.Request and extract what we need from its context. I am testing it right now. I will push the code tomorrow.

internalshared/configutil/http_response_headers.go
hghaf099: we could do that, however, one thing to consider is that operators may want to just pass in one string with all headers specified or pass in a list of headers values which comprise couple of headers per item. Those could be have spaces and semicolons. We have examples of it in /ui and also I have specified default headers the same way. For example:
strictTransportSecurity = "max-age=31536000; includeSubDomains"
So, searching for semicolon and spaces may cause issues. what do you think? is there any other particular invalid chars that we need to care for?

http/handler.go
```diff
@@ -166,8 +167,8 @@ func Handler(props *vault.HandlerProperties) http.Handler {
 			} else {
 				mux.Handle("/ui/", handleUIHeaders(core, handleUIStub()))
 			}
-			mux.Handle("/ui", handleUIRedirect())
-			mux.Handle("/", handleUIRedirect())
+			mux.Handle("/ui", handleUIRedirect(core))
+			mux.Handle("/", handleUIRedirect(core))
```
hghaf099, line 171: TODO: core param is not needed here

internalshared/configutil/http_response_headers.go
ncabatoff: Ah, fair enough.  No, let's leave it be then.

internalshared/listenerutil/custom_response_headers.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// ChangeListenerAddress is used for tests where the listener address (at least the port)
+// is chosen at random
+func (l *ListenerCustomHeaders) ChangeListenerAddress(la string) {
+	l.Address = la
+	return
+}
+
+func (l *ListenerCustomHeaders) SetCustomResponseHeaders(w http.ResponseWriter, status int) {
```
hghaf099, line 33: Any suggestion on how to handle errors in this function? should we return them, or somehow log them? In the one hand, I am not sure if it should interrupt request processing, on the other hand, it should not silently fail. So, please provide me with your suggestions on this.

http/sys_rekey.go
```diff
@@ -24,7 +24,8 @@ func handleSysRekeyInit(core *vault.Core, recovery bool) http.Handler {
 		repState := core.ReplicationState()
 		if repState.HasState(consts.ReplicationPerformanceSecondary) {
 			respondError(w, http.StatusBadRequest,
-				fmt.Errorf("rekeying can only be performed on the primary cluster when replication is activated"))
+				fmt.Errorf("rekeying can only be performed on the primary cluster when replication is activated"),
```
ncabatoff, line 27: Please revert.

vault/core.go
```diff
@@ -1001,6 +1003,9 @@ func NewCore(conf *CoreConfig) (*Core, error) {
 
 	c.clusterListener.Store((*cluster.Listener)(nil))
 
+	uiHeaders, _ := c.UIHeaders()
```
ncabatoff, line 1006: Why ignore the err?

vault/core.go
```diff
@@ -517,6 +517,8 @@ type Core struct {
 	// clusterListener starts up and manages connections on the cluster ports
 	clusterListener *atomic.Value
 
+	customListenerHeader *ListenersCustomResponseHeadersList
```
ncabatoff, line 520: We'll need to make this field threadsafe, either with a lock or by making it atomic.

vault/core.go
```diff
@@ -2632,6 +2637,84 @@ func (c *Core) SetConfig(conf *server.Config) {
 	c.logger.Debug("set config", "sanitized config", string(bz))
 }
 
+func (c *Core) getCustomResponseHeaders(la string) []*ListenerCustomHeaders {
+	if c.customListenerHeader == nil {
+		c.logger.Debug("failed to find the custom response headers configuration")
```
ncabatoff, line 2642: Looks like this debug message as well as the warn below would be emitted on every request, probably not what we want.

vault/core.go
```diff
@@ -2632,6 +2637,84 @@ func (c *Core) SetConfig(conf *server.Config) {
...
+	lch := c.customListenerHeader.getListenerMap(la)
+	if lch == nil {
+		c.logger.Warn("no listener config found", "address", la)
+		return nil
+	}
+
+	return lch
+}
+
+func (c *Core) GetListenerCustomResponseHeaders(la string) *ListenerCustomHeaders {
```
ncabatoff, line 2655: Can you rename `la` to something more meaningful please?

vault/custom_response_headers.go
```diff
@@ -0,0 +1,193 @@
...
+	"fmt"
+	log "github.com/hashicorp/go-hclog"
+	"net/http"
+	"net/textproto"
+	"strings"
+
+	"github.com/hashicorp/vault/internalshared/configutil"
+)
+
+type ListenersCustomResponseHeadersList struct {
```
ncabatoff, line 13: Why define a type for this?

vault/custom_response_headers.go
```diff
@@ -0,0 +1,193 @@
...
+	CustomHeadersList []*ListenerCustomHeaders
+}
+
+type ListenerCustomHeaders struct {
+	Address string
```
ncabatoff, line 18: Looks like these fields could all be private, since they're only referenced in methods.

vault/custom_response_headers.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	// either looking for a specific listener, or if listener address isn't given,
+	// checking for all available listeners
+	var lch []*ListenerCustomHeaders
+	if address == "" {
+		lch = c.CustomHeadersList
+	} else {
+		for _, l := range c.CustomHeadersList {
+			if l.Address == address {
```
ncabatoff, line 98: If it is presently possible to specify multiple listener config stanzas that reference the same address, that might be a bug.

ncabatoff (commented):
Not done reviewing, but figured I'd give you a batch of comments since I'll be busy for the next little while.

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func parseHeaderValues(h interface{}) (string, error) {
+    var sl []string
+    if _, ok := h.([]interface{}); !ok {
+        return "", fmt.Errorf("headers must be given in a list of strings")
+    }
+    vli := h.([]interface{})
+    for _, vh := range vli {
+        if vh.(string) == "" {
```
ncabatoff, line 167: What if vh isn't a string?  Won't this type assertion panic?  (I don't actually know whether it's possible for the HCL parsing code to return anything other than a list of interfaces that are actually strings, just being paranoid.)

vault/custom_response_headers_test.go
```diff
@@ -0,0 +1,179 @@
...
+		t.Fatalf("should not be able to set a header that is in custom response headers")
+	}
+
+	req = logical.TestRequest(t, logical.UpdateOperation, "config/ui/headers/X-CustomUiHeader")
+	req.Data["values"] = []string{"Ui header value"}
+	req.ResponseWriter = hw
+
+	_, err = b.HandleRequest(namespace.RootContext(nil), req)
+	if err != nil {
+		t.Fatal("request did not fail on setting a header that is present in custom response headers")
```
ncabatoff, line 172: You say "err != nil", but the message is "did not fail" - something seems wrong there.

command/agent/config/config_test.go
```diff
@@ -11,6 +11,15 @@ import (
 	"github.com/hashicorp/vault/sdk/helper/pointerutil"
 )
 
+var DefaultCustomHeaders = map[string]string{
```
ncabatoff, line 14: I hadn't paid attention to the changes in this file before. It looks like you're modifying Agent's listener config as well.  That may make sense, but it's not clear to me that it was included in the scope, and if it is included, I suspect there need to be additional changes and tests.

vault/custom_response_headers_test.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// setting an ui specific header
+	req = logical.TestRequest(t, logical.UpdateOperation, "config/ui/headers/X-CustomUiHeader")
+	req.Data["values"] = []string{"Ui header value"}
+	req.ResponseWriter = hw
+
+	_, err = b.HandleRequest(namespace.RootContext(nil), req)
+	if err != nil {
+		t.Fatal("request failed on setting a header that is not present in custom response headers")
```
ncabatoff, line 167: Please include the error here.

vault/logical_system.go
```diff
@@ -2623,6 +2623,11 @@ func (b *SystemBackend) handleConfigUIHeadersUpdate(ctx context.Context, req *lo
 	// Translate the list of values to the valid header string
 	value := http.Header{}
 	for _, v := range values {
+		// ExistCustomResponseHeader support checking for custom headers per listener address
```
ncabatoff, line 2626: I think this comment is better suited to a godoc on ExistCustomResponseHeader, that way if we wind up using it elsewhere people will still be able to see it by looking at the definition.

vault/core.go
```diff
@@ -2632,6 +2647,91 @@ func (c *Core) SetConfig(conf *server.Config) {
...
+			}
+		}
+		if len(lch) == 0 {
+			return nil
+		}
+	}
+	return lch
+}
+
+func (c *Core) GetListenerCustomResponseHeaders(listenerAdd string) *ListenerCustomHeaders {
```
ncabatoff, line 2680: Not sure how to add to an outdated conversation, so re:

> The "if address is an empty string" comment is in another function though, not sure what to do about that here? would you please elaborate?

I just think it's worth noting in godoc on this method that listenerAddr is optional.

vault/core.go
```diff
@@ -2632,6 +2647,91 @@ func (c *Core) SetConfig(conf *server.Config) {
...
+	if customHeadersList == nil {
+		return nil
+	}
+
+	// either looking for a specific listener, or if listener address isn't given,
+	// checking for all available listeners
+	var lch []*ListenerCustomHeaders
+	if listenerAdd == "" {
+		lch = customHeadersList
+	} else {
```
ncabatoff, line 2666: Totally fine to ignore this, but: it'd be more idiomatic to do early returns instead of declaring lch, using an else, doing a break, etc.

vault/core.go
hghaf099: I see. I will add it to the ExistCustomResponseHeader function

vault/custom_response_headers_test.go
hghaf099: Done

vault/logical_system.go
hghaf099: Done

command/agent/config/config_test.go
ncabatoff: Well, Agent is its own mode, like `vault agent` vs `vault server`, and has its own listener config.  I guess it's using the same listener parsing code, based on what you're saying.  So we need to decide whether to maintain that and thereby include Agent within the scope of this change, or, more likely, decouple them so that Agent's listener config parsing doesn't have expectations re custom response headers.

I do think it makes sense to fix Agent in the same way, but I'm not sure we want to do that for 1.9 (since we weren't required to) and even if we do, I'd prefer it were its own PR so that we can get this one merged sooner.

hghaf099: I am going to decouple it then for now, to do that I figure I could simply add the following in line 177 of command/agent/config/config.go, what would you think? 

`// Pruning custom headers for Agent for now
  for _, ln := range sharedConfig.Listeners {
		ln.CustomResponseHeaders = nil
}
`
well, I will fix the current tests I modified as well. There is a Prune() as well in which I will do the same step as above.

changelog/12485.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+http: Enable users to customize HTTP response headers
```
ccapurso, line 2: The format for a `feature` entry should be as follows:

```
**Customizable HTTP Headers**: Add support to define custom HTTP headers for root path (`/`) and also on API endpoints (`/v1/*`)
```

The description can be whatever you want, I just wanted to put something there that was related to the PR.

http/sys_metrics.go
```diff
@@ -35,12 +35,14 @@ func handleMetricsUnauthenticated(core *vault.Core) http.Handler {
...
 		// Manually extract the logical response and send back the information
-		w.WriteHeader(resp.Data[logical.HTTPStatusCode].(int))
+		status := resp.Data[logical.HTTPStatusCode].(int)
 		w.Header().Set("Content-Type", resp.Data[logical.HTTPContentType].(string))
 		switch v := resp.Data[logical.HTTPRawBody].(type) {
 		case string:
+			w.WriteHeader(status)
 			w.Write([]byte(v))
 		case []byte:
+			w.WriteHeader(status)
```
ccapurso, line 45: I might be missing something subtle but the proposed changes look functionally equivalent to the prior logic. A call to `w.WriteHeader` is being done in both the `string` and `[]byte` cases. Is there a reason that the outer call to `w.WriteHeader` on line 38 cannot be kept as is?

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,179 @@
+package configutil
```
ccapurso, line 1: Before I review this further, the functions in this file could benefit from some godoc comments.

Using more descriptive variables names would create an easier reading experience as they can be somewhat self-documenting. Short variable names can often be difficult to determine the intent of the variable. They are also harder to search for.

Examples:
- `ParseCustomResponseHeaders`: `h`, `r`, `de`, `hv`, `hvl`, and `hvm`
- `setDefaultResponseHeaders`: `c`, `hv` and `hn` are arguably fine here as their usage and scope is short-lived
- `parseHeaderValues`: `h`, especially since it is of type `interface{}`

changelog/12485.txt
hghaf099: hmm, seems like there are various formats for a `feature`. I am following this and not see the suggested format
https://github.com/hashicorp/vault/blob/main/CHANGELOG.md
though, I do see the suggested format in the changelog directory.

http/sys_metrics.go
hghaf099: So, once the WriteHeader is called, setting other headers would be ignored. So, in this case, the "Content-Type" will be ignored which it seems like an old bug. Another pointer is that the responseError function also sets some headers and calls the WriteHeader. So, I made the change to accommodate the two cases here.

internalshared/configutil/http_response_headers.go
hghaf099: Thanks for your comment, I will try to get them fixed. I am still struggling with finding a good habit of naming variables. I have seen this pattern in the Vault code base, but it is indeed confusing to me as well. I will try to fix this!

changelog/12485.txt
ccapurso: I don't think we have been consistent but the "Key Management Secrets Engine (Enterprise)" line in the unreleased 1.9 section aligns with what I mentioned. I commented on it because I was informed that used the incorrect format as well for the "secrets/kv: Add ability to specify version-agnostic custom key metadata" line awhile after it had been merged.

hghaf099: Sounds good! Thanks Chris!

internalshared/configutil/http_response_headers.go
ccapurso: I find that it is typically situational. If the lexical scope that you're dealing with is small (e.g. a small helper function or loop) and the type is descriptive enough (e.g. a frequently used struct) then I often find the variable names are throw away. Modern editors definitely help but reading the code in the browser is pretty common so descriptive names are really beneficial.

Some suggestions would be `header` over `h`, `headerVal` over `hv`, and `headerValList` over `hvl`. I'm not actually sure if I inferred the correct intent there but the examples should hopefully be sufficient.

changelog/12485.txt
mladlow: The feature section of the changelog is our chance to do a fancier formatting and more high-level description of major features in the new major release (see https://github.com/hashicorp/vault/blob/main/CHANGELOG.md#140-april-7th-2020 for an example).

hghaf099: Thanks, Done

internalshared/configutil/http_response_headers.go
hghaf099: I made some changes to the variable names, I hope it is more clear now. I will push a patch shortly

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+		hvMap[hn] = s
+	}
+	return hvMap, nil
+}
+
+func parseHeaderValues(h interface{}) (string, error) {
+	var sl []string
+	if _, ok := h.([]interface{}); !ok {
+		return "", fmt.Errorf("headers must be given in a list of strings")
```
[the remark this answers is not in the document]
hghaf099: This function is called from `ParseCustomResponseHeaders` which its input is an interface parsed using HCL config parser. So, the assertion should happen anyways either in this function or the functions down in the stack. This was the most relevant place for those.

vault/core.go
```diff
@@ -2632,6 +2647,91 @@ func (c *Core) SetConfig(conf *server.Config) {
 	c.logger.Debug("set config", "sanitized config", string(bz))
 }
 
+func (c *Core) getCustomHeadersListenerList(listenerAdd string) []*ListenerCustomHeaders {
+	customHeaders := c.customListenerHeader.Load()
+	if customHeaders == nil {
+		return nil
+	}
+
+	customHeadersList := customHeaders.([]*ListenerCustomHeaders)
```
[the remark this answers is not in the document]
hghaf099: done, I am going to add this with my other changes.

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,179 @@
...
+	if _, ok := r.([]map[string]interface{}); !ok {
+		return nil, fmt.Errorf("response headers were not configured correctly. please make sure they're in a map")
+	}
+
+	customResponseHeader := r.([]map[string]interface{})
+
+	for _, crh := range customResponseHeader {
+		for statusCode, responseHeader := range crh {
+			if _, ok := responseHeader.([]map[string]interface{}); !ok {
+				return nil, fmt.Errorf("response headers were not configured correctly. please make sure they're in a map")
```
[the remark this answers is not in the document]
hghaf099: done

internalshared/configutil/http_response_headers.go
```diff
@@ -0,0 +1,179 @@
...
+	h := make(map[string]map[string]string)
+	// if r is nil, we still should set the default custom headers
+	if r == nil {
+		de := h["default"]
+		h["default"] = setDefaultResponseHeaders(de)
+		return h, nil
+	}
+
+	if _, ok := r.([]map[string]interface{}); !ok {
+		return nil, fmt.Errorf("response headers were not configured correctly. please make sure they're in a map")
```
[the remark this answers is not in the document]
hghaf099: done

PR #12495, opened 2021-09-07
docs: provide clarifications for github auth method and ssh secrets engine

mickael-hc:


sgmiller (approved):
We're probably going to follow on with a feature to prohibit extensions as defense in depth.  But this is good for a docs warning.

PR #12498, opened 2021-09-07
KV: handle permissions on config and mount/sys when enabling a KV engine

Monkeychip:
This PR extends the work already merged for the KV custom metadata project. It handles an issue where if you don't have config access (which most users won't because it wasn't something you needed previously to enable a KV engine) we now show a flash warning instead of stalling on the enable page.

**See gif below:**
no-config

Test policy to reproduce the above:
```
path "secret-test/*"
{
  capabilities = ["create", "read", "delete", "list", "sudo"]
}

# Manage secrets engines
path "sys/mounts/*"
{
  capabilities = ["update", "create", "read","delete", "list", "sudo"]
}

# List existing secrets engines.
path "sys/mounts"
{
  capabilities = ["read"]
}

# List existing secrets engines.
path "kv/*"
{
  capabilities = ["read", "list"]
}
```

This PR also cleans up a situation where if you don't have KV `mount/sys` permissions we stay on the enable page and show an error message. Before we transitioned you to an error page with an incorrect error message.

**See gif below:**
update-permissions

Test policy to reproduce the above is the same policy before but remove "update" on the `path "sys/mounts" `

### DIFF
--- a/ui/app/adapters/secret-engine.js
+++ b/ui/app/adapters/secret-engine.js
@@ -30,6 +30,7 @@ export default ApplicationAdapter.extend({
     try {
       mountModel = await this.ajax(this.internalURL(query.path), 'GET');
       // if kv2 then add the config data to the mountModel
+      // version comes in as a string
       if (mountModel.data.type === 'kv' && mountModel.data.options.version === '2') {
         configModel = await this.ajax(this.urlForConfig(query.path), 'GET');
         let data = mountModel.data;
@@ -47,9 +48,7 @@ export default ApplicationAdapter.extend({
     let data = serializer.serialize(snapshot);
     const path = snapshot.attr('path');
     // for kv2 we make two network requests
-    // ARG TODO make sure to have it equal 2 and it might be a string. Need to double check.
-    if (data.type === 'kv' && data.options.version !== 1) {
-      console.log('here');
+    if (data.type === 'kv' && data.options.version === 2) {
       // data has both data for sys mount and the config, we need to separate them
       let splitObjects = splitObject(data, ['max_versions', 'delete_version_after', 'cas_required']);
       let configData;
--- a/ui/app/adapters/secret-engine.js
+++ b/ui/app/adapters/secret-engine.js
@@ -29,7 +29,7 @@ export default ApplicationAdapter.extend({
     return this.ajax(this.internalURL(query.path), 'GET');
   },
 
-  async createRecord(store, type, snapshot) {
+  createRecord(store, type, snapshot) {
     const serializer = store.serializerFor(type.modelName);
     let data = serializer.serialize(snapshot);
     const path = snapshot.attr('path');
@@ -44,22 +44,30 @@ export default ApplicationAdapter.extend({
         data.id = path;
       }
       // first create the engine
-      await this.ajax(this.url(path), 'POST', { data })
+      return this.ajax(this.url(path), 'POST', { data })
         .then(() => {
           // second create the config request
-          this.ajax(this.urlForConfig(path), 'POST', { data: configData }).catch(() => {
+          return this.ajax(this.urlForConfig(path), 'POST', { data: configData }).catch(() => {
             // error here means you do not have update capabilities to config endpoint. If that's the case we show a flash message in the component and continue with the transition.
             // the error is handled by mount-backend-form component because the save method returns an error and we filter through the error to determine if we should proceed or not.
+            throw new Error('configIssue');
           });
         })
-        .then(() => {
+        .finally(() => {
           return {
             data: assign({}, data, { path: path + '/', id: path }),
           };
         })
-        .catch(() => {
-          // if error here it's because they don't have access to mount engine and should not proceed. Return them out of function.
+        .catch(e => {
           // if not mount error and already returned then you need to assign the id
+          if (e.message === 'configIssue') {
+            // but still proceed with mount sys call and do not throw mountIssue error
+            return;
+          }
+          // if they do not hit a config error then it's a sys/mount error and they should not proceed.
+          if (e.httpStatus === 400) {
+            throw new Error('samePath');
+          }
           throw new Error('mountIssue');
         });
     } else {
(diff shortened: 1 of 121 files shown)


ui/app/controllers/vault/cluster/settings/mount-secret-backend.js
```diff
@@ -22,8 +22,5 @@ export default Controller.extend({
         this.wizard.transitionFeatureMachine(this.wizard.featureState, 'CONTINUE', type);
       });
     },
-    onConfigError: function(modelId) {
```
Monkeychip, line 25: this was not being used, and we don't actually want to transition to an error page. We want to keep the user on the page because it could just be they don't have access to create that secret engine with that name.

ui/app/templates/components/mount-backend-form.hbs
```diff
@@ -29,7 +29,7 @@
       @mode="enable"
       @noun={{if (eq mountType "auth") "Auth Method" "Secret Engine"}}
      />
-    <MessageError @model={{mountModel}} />
```
Monkeychip, line 32: this was causing errors, because we don't create the `mountModel` if we get caught in the catch on save()

ui/app/adapters/secret-engine.js
```diff
@@ -25,34 +25,66 @@ export default ApplicationAdapter.extend({
...
       // data has both data for sys mount and the config, we need to separate them
       let splitObjects = splitObject(data, ['max_versions', 'delete_version_after', 'cas_required']);
       let configData;
       [configData, data] = splitObjects;
+
+      if (!data.id) {
+        data.id = path;
+      }
       // first create the engine
       return this.ajax(this.url(path), 'POST', { data })
```
Monkeychip, line 61: Note: TODO refactor to try catch.

ui/app/components/mount-backend-form.js
```diff
@@ -90,10 +92,32 @@ export default Component.extend({
...
     const mountModel = this.mountModel;
     const { type, path } = mountModel;
+    // because user might not have access to config, do a capabilities check with the path name here
+    let capabilities = yield this.store.findRecord('capabilities', `${path}/config`);
+    if (!capabilities.get('canUpdate')) {
+      // if there is no sys/mount issue then error is config endpoint.
+      this.flashMessages.warning(
+        'You do not have access to the config endpoint. The secret engine was mounted, but the configuration settings were not saved.'
+      );
+      // remove the config data from the model otherwise it will save it even if the network request failed.
```
Monkeychip, line 102: this may get fixed with serializer change that Chelsea is working on.

ui/app/models/secret-engine.js
```diff
@@ -49,7 +49,7 @@ export default Model.extend(Validations, {
   }),
   // KV 2 additional config default options
   maxVersions: attr('number', {
-    defaultValue: 10,
```
Monkeychip, line 52: api docs need to be updated. I double checked this by running the api commands and then reading the config. the default here is 0, though on a secret it's 10.

ui/app/routes/vault/cluster/secrets/backend/configuration.js
```diff
@@ -9,40 +9,15 @@ export default Route.extend({
     if (this.wizard.featureState === 'list') {
       this.wizard.transitionFeatureMachine(this.wizard.featureState, 'CONTINUE', backend.get('type'));
     }
-    // if KV2 then we pull in specific attrs from the config endpoint saved on the secret-engine record and display them
     if (backend.isV2KV) {
-      let secretEngineRecord = this.store.peekRecord('secret-engine', backend.id);
```
Monkeychip, line 14: is set on the model. my bad for not figuring this out in the first PR. I really should have pulled this config work into a separate PR to start with.

ui/app/routes/vault/cluster/secrets/backend/list.js
```diff
@@ -149,6 +149,7 @@ export default Route.extend({
     let backend = this.enginePathParam();
     let backendModel = this.store.peekRecord('secret-engine', backend);
     let has404 = this.has404;
+    let capabilities = this.store.findRecord('capabilities', `${backend}/data/${secret}`);
```
Monkeychip, line 152: TODO: this might have been a part of trying to hide the "create secret work" I believe I can remove this but want to double check.

ui/app/templates/components/mount-backend-form.hbs
Monkeychip: But need to double check this because now there might be other errors that are created after mountModel that we don't capture?

ui/app/adapters/secret-engine.js
hashishaw: Here's how I might do it:
```js
// first create the engine
      try {
        await this.ajax(this.url(path), 'POST', { data });
      } catch (createError) {
        if (createError.httpStatus === 400) {
          throw new Error('samePath');
        }
        throw new Error('mountIssue');
      }

      // second create the config request
      let config;
      try {
        config = await this.ajax(this.urlForConfig(path), 'POST', { data: configData });
        // Do we do anything with the config response?
      } catch (configError) {
        // do nothing, continue with success
      }
      return {
        data: assign({}, data, { path: path + '/', id: path }),
      };
```

ui/app/adapters/secret-engine.js
```diff
@@ -25,35 +25,61 @@ export default ApplicationAdapter.extend({
...
+  async query(store, type, query) {
+    let mountModel, configModel;
+    try {
+      mountModel = await this.ajax(this.internalURL(query.path), 'GET');
+      // if kv2 then add the config data to the mountModel
+      // version comes in as a string
+      if (mountModel.data.type === 'kv' && mountModel.data.options.version === '2') {
+        configModel = await this.ajax(this.urlForConfig(query.path), 'GET');
+        let data = mountModel.data;
+        data = { ...mountModel.data, ...configModel.data };
```
arnav28, line 37: Maybe we can simplify this directly? 
mountModel.data = { ...mountModel.data, ...configModel.data }

arnav28 (approved):
Nice work!!

ui/app/adapters/secret-engine.js
Monkeychip: Good call, thank you!

Issue #12290, opened 2021-08-09
LDAP Auth user/group mapping are cleared when loading from a file with a name containing an equal ('=')


arusso:
**Describe the bug**

When creating / updating an LDAP Auth group (or user) mapping from the command-line using the load from file (`@`) where the filename contains an equal (`=`), the group is either created or updated but the policies in the mapping are emptied.

**To Reproduce**

```

# setup config, works here!
❯ vault auth enable -path ldap-test ldap
Success! Enabled ldap auth method at: ldap-test/

❯ cat ldap-config.json
{
  "anonymous_group_search": false,
  "binddn": "",
  "case_sensitive_names": true,
  "certificate": "-----BEGIN...",
  "deny_null_bind": true,
  "discoverdn": true,
  "groupattr": "memberOf",
  "groupdn": "ou=people,o=example.com",
  "groupfilter": "(uid={{.Username}})",
  "insecure_tls": false,
  "starttls": false,
  "tls_max_version": "tls12",
  "tls_min_version": "tls10",
  "token_bound_cidrs": [],
  "token_explicit_max_ttl": 0,
  "token_max_ttl": 0,
  "token_no_default_policy": false,
  "token_num_uses": 0,
  "token_period": 0,
  "token_policies": [],
  "token_ttl": 0,
  "token_type": "default",
  "upndomain": "",
  "url": "ldaps://ldap-infra.example.com:636",
  "use_pre111_group_cn_behavior": true,
  "use_token_groups": false,
  "userattr": "uid",
  "userdn": "ou=people,o=example.com"
}

❯ vault write auth/ldap-test/config @ldap-config.json
Success! Data written to: auth/ldap-test/config

❯ vault read -format=json -field=data auth/ldap-test/config
{
  "anonymous_group_search": false,
  "binddn": "",
  "case_sensitive_names": true,
  "certificate": "-----BEGIN...",
  "deny_null_bind": true,
  "discoverdn": true,
  "groupattr": "memberOf",
  "groupdn": "ou=people,o=example.com",
  "groupfilter": "(uid={{.Username}})",
  "insecure_tls": false,
  "starttls": false,
  "tls_max_version": "tls12",
  "tls_min_version": "tls10",
  "token_bound_cidrs": [],
  "token_explicit_max_ttl": 0,
  "token_max_ttl": 0,
  "token_no_default_policy": false,
  "token_num_uses": 0,
  "token_period": 0,
  "token_policies": [],
  "token_ttl": 0,
[...]

PR #12500, opened 2021-09-07
Update kv input documentation with edge case

pmmukh:
- Updates the CLI documentation with the edge case that was discovered while looking into https://github.com/hashicorp/vault/issues/12290

### DIFF
--- a/website/content/docs/commands/index.mdx
+++ b/website/content/docs/commands/index.mdx
@@ -172,6 +172,10 @@ Or specify the contents of a file as a value:
 $ vault kv put secret/password value=@data.txt
 ```
 
+Note that if an argument is supplied in a @key=value format, Vault will treat that as a 
+kv pair with the key being `@key`, not a file called `key=value`. This also means that Vault
+does not support filenames with `=` in them.
+
 ### Reading Data
 
 After data is persisted, read it back using `vault kv get`:


website/content/docs/commands/index.mdx
```diff
@@ -172,6 +172,10 @@ Or specify the contents of a file as a value:
 $ vault kv put secret/password value=@data.txt
 ```
 
+Note that if an argument is supplied in a @key=value format, Vault will treat that as a 
+kv pair with the key being `@key`, not a file called `key=value`. This also means that Vault
```
ncabatoff, line 176: Technically we do support filenames with `=` in them in some cases right?

pmmukh: hmm no i don't think so. We first split the input by `=` here https://github.com/hashicorp/go-secure-stdlib/blob/main/kv-builder/builder.go#L52, and if len(parts) == 1 i.e the input has no `=`, then we try to read it as a file here https://github.com/hashicorp/go-secure-stdlib/blob/main/kv-builder/builder.go#L70

ncabatoff: Ah, my mistake then.

ncabatoff (approved):

PR #12517, opened 2021-09-08
Use HCL syntax for code snippet

topfunky:
Use `hcl` so the code snippet is more readable and is highlighted correctly.

Screen Shot 2021-09-08 at 12 50 09 PM


aphorise:
This is still a valid request and we'd do well to check other HCL areas in the docs that may be missing.

taoism4504 (approved):

yhyakuna (approved):

PR #12519, opened 2021-09-08
Fix pkcs7 parsing in some cases

### LINKED ISSUE #12520: EC2 Auth Fails on Some Instances
**Describe the bug**
EC2 Vault login fails with `failed to parse the BER encoded PKCS#7 signature: ber2der: Invalid BER format` for some instances

**To Reproduce**
1. Run `vault write -format=json "auth/aws/login" role="$role" pkcs7="$pkcs7" nonce="$nonce"`
2. Get the error from above

This only happens on a subset of ec2 instances (those that include a certificate in the response from `http://169.254.169.254/latest/dynamic/instance-identity/pkcs7`)  We could not find a pattern as to which instances did and didn't include the certificate, but the bug reliably happens if the certificate is included.

**Expected behavior**
It should not break

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.8.2
* Vault CLI Version (retrieve with `vault version`): 1.6.2
* Server Operating System/Architecture: linux/amd64

**Additional context**
When #12340 / #12361 were pulled in, they happened before this upstream PR: https://github.com/mozilla-services/pkcs7/pull/61

Without the upstream pr, certain valid BER files are rejected as invalid and are mis-parsed.

maths22:
Brings in https://github.com/mozilla-services/pkcs7/pull/61 from upstream

In some cases but not all, aws includes a certificate in the pkcs7 response,
and currently vault fails to parse those certificates:
```
URL: PUT https://vault.example.com/v1/auth/aws/login
Code: 500. Errors
* failed to parse the BER encoded PKCS#7 signature: ber2der: Invalid BER format
```

This fixes logins on those instances.  Note we could not readily ascertain why
some instances have those certificates and others don't.

fixes #12520


jasonodonnell (approved):

ryanmt:
@jasonodonnell I see this was added to the 1.8.3 milestone but wasn't picked, nor has it been added to any subsequent 1.8.X release.

Was that intentional or an oversight? I prefer to not skip a release, but my journey from 1.7 to 1.9 would be a lot easier if there was a fixed version of 1.8 released. 

Thanks!

jasonodonnell:
Hi @ryanmt, thanks for the ping. This was an oversight, so I opened a PR to do so here: https://github.com/hashicorp/vault/pull/13851.

harsimranmaan:
This seems to be affecting 1.7.9 too. Are there plans to backport the patch @jasonodonnell ?

Issue #12507, opened 2021-09-08
Vendoring not working with //go:embed web_ui/*


heppu:
I tried to use main branch of vault as dependecy since it has the fix which is needed for Go 1.17 and that's when I found a new problem. When running go mod vendor Go will now try to copy also all files described with //go:embed directive. Since those files are not part of the vaults git repo vault cannot be vendored.

**To Reproduce**
```sh
$ go get -d github.com/hashicorp/vault@main
go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.
$ go mod vendor
go mod vendor: pattern web_ui/*: no matching files found
```

**Expected behavior**
Running go mod vendor should work with vault as dependecy.

**Environment:**
go1.17 linux/amd64
github.com/hashicorp/vault@main

hsimon-hashicorp:
As this is not a HashiCorp-supported method of using Vault, I will close this issue. If you would like to submit a PR to address this, we'd be happy to review it, but at this time, we will not be making these changes. Thanks for understanding.

heppu:
@hsimon-hashicorp  I have now created the PR so should we reopen this issue or should I create a new one?

PR #12523, opened 2021-09-09
Include generated ui files in git repo.

heppu:
As described in #12507 when `//go:embed` is used Go expects those files to be available in repository. If the files are not available that becomes a problem when we try to vendor the package as dependency.

This PR  should include two things:

- [x] commit `http/web_ui` directory in git repository
- [ ] validate in CI that those files have not been edited manually

For the  validation part in CI we could do following:

1. `make static-dist`
2. `git diff --name-status --exit-code -- ./http/web_ui`

I'm not too familiar with Vault CI pipe and not sure what would the correct place for this so any suggestions would be really helpfull. Ideally the UI files would be already generated in CI pipe and we would only need to check with git command that nothing actually changed.


pmmukh:
Hi @heppu ! Thanks for raising this PR! After taking a look at it, we don't feel like having these files added to Vault is a path we want to go down, with 70 something files it causes unnecessary bloat to the repo to support something we don't intend to support ( vendoring vault as a go module ). As such, I'm going to go ahead and close this PR out.

heppu:
Sure I can understand that. @pmmukh  How would you feel about having that folder with single place holder file? That way it wouldn't bloat the repo, no need to modify tooling and vendoting would work?

pmmukh:
That honestly seems like a bit of a hacky workaround, and not something we'd like to keep around. I would recommend that for now you depend on an older version of vault that works for vendoring. If you need to both use go 1.17 (as I saw mentioned in your issue) and use vault as a vendored dependency, I'm sorry but I don't think that will be possible.

PR #12526, opened 2021-09-09
Use LTS node version

amyrlam:
This change allows nvm to be picked up by asdf as well re: https://github.com/asdf-vm/asdf-nodejs#nvmrc-and-node-version-files

asdf couldn't pick up `v14`


hashishaw (approved):
Thank you!

amyrlam:
I've noticed that this has weird implications with `nvm` so I think this could actually be closed. Could alternatively add a `.tool-versions` for `asdf` if needed, but I've actually stopped using it and went back to `nvm`, ran into too many headaches. If you disagree, please reopen!

PR #12530, opened 2021-09-10
Docs: Seal pkcs11 updated example with actual hex slot reference and …

aphorise:
…notes related to decimal conversion. Minor correction to **Note** area in 'lib' parameter above 'slot'.

Screenshot 2021-09-10 at 15 54 47


sgmiller (approved):
Looks good, thanks for the clarification text.

sgmiller (approved):

aphorise:
Hey @sgmiller - shouldn't this already [be on `vaultproject.io/docs/configuration/seal/pkcs11`](https://www.vaultproject.io/docs/configuration/seal/pkcs11)? - it still reads the same as before.

PR #12534, opened 2021-09-11
vault-agent: copy values retrieved from bolt

tvoran:
[Byte slices returned from Bolt are only valid during a transaction](https://github.com/etcd-io/bbolt#caveats--limitations), and using them outside the transaction can cause an `unexpected fault address` panic so this makes a copy in the persistent caching functions.

Similar to the pattern in fsm.go: https://github.com/hashicorp/vault/blob/861454e0ed1390d67ddaf1a53c1798e5e291728c/physical/raft/fsm.go#L469-L479

### DIFF
--- a/changelog/12534.txt
+++ b/changelog/12534.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-agent: Avoid possible `unexpected fault address` panic caused by using byte slices returned from Bolt outside their transactions; make a copy first.
+agent: Avoid possible `unexpected fault address` panic caused by using byte slices returned from Bolt outside their transactions; make a copy first in the persistent cache restore functions.
 ```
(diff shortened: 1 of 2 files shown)


kalafut (approved):

calvn (approved):

changelog/12534.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+agent: Avoid possible `unexpected fault address` panic caused by using byte slices returned from Bolt outside their transactions; make a copy first in the persistent cache restore functions.
```
tomhjp, line 2: nit: Maybe a bit too much detail for a changelog?

```suggestion
agent: Avoid possible `unexpected fault address` panic when using persistent cache.
```

tomhjp (approved):

benashz (approved):

PR #12538, opened 2021-09-13
Adds OIDC Authorization Endpoint to OIDC providers

austingebauer:
## Overview

This PR adds the [OIDC Authorization Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint) to OIDC providers in the identity store. This allows OIDC clients to request an authorization code to be used for the [Authorization Code Flow](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth).

With this PR, all OIDC clients will be loaded into the memdb instance on the `IdentityStore` struct at [core.go#L2015](https://github.com/hashicorp/vault/blob/main/vault/core.go#L2015). This allows for looking up the OIDC clients by ID in OIDC endpoints.

## Testing

I've added tests in the PR and ran manual tests.

Manual test output:
```sh
# Create an entity
$ vault write -format json identity/entity \
    name="austin" \
    metadata="email=vault@hashicorp.com" \
    metadata="phone_number=123-456-7890"
$ ENTITY_ID=$(vault read -field=id identity/entity/name/austin)

# Create a group
$ vault write identity/group \
    name="engineering" \
    member_entity_ids="$ENTITY_ID"
$ GROUP_ID=$(vault read -field=id identity/group/name/engineering)

# Enable userpass auth method
$ vault auth enable userpass
$ vault write auth/userpass/users/austin \
    password=mypassword \
    policies=default

# Create entity alias for userpass auth method and entity
$ USERPASS_ACCESSOR=$(vault auth list -detailed -format json | jq -r '.["userpass/"].accessor')
$ vault write identity/entity-alias \
    name=austin \
    canonical_id="$ENTITY_ID" \
    mount_accessor="$USERPASS_ACCESSOR"

# Create a key
$ vault write identity/oidc/key/sigkey allowed_client_ids="*"

# Create an assignment
$ vault write identity/oidc/assignment/my-assignment \
    group_ids="$GROUP_ID" \
    entity_ids="$ENTITY_ID"

# Create a client
$ vault write identity/oidc/client/my-webapp \
    redirect_uris="http://127.0.0.1:8251/callback" \
    assignments="my-assignment" \
    key="sigkey" \
    id_token_ttl="30m" \
    access_token_ttl="1h"

# Get client_id and client_secret
$ CLIENT_ID=$(vault read -field=client_id identity/oidc/client/my-webapp)
$ CLIENT_SECRET=$(vault read -field=client_secret identity/oidc/client/my-webapp)

# Create a provider
$ vault write identity/oidc/provider/my-provider \
    allowed_client_ids="$CLIENT_ID"

# Login with userpass auth method
$ vault login -method=userpass username=austin password=mypassword

# GET /authorize with query string serialization
$ curl -X GET \
    --header "X-Vault-Token: $(cat ~/.vault-token)" \
    --header "X-Vault-Namespace: $VAULT_NAMESPACE" \
    -G \
    -d "response_type=code" \
    -d "client_id=$CLIENT_ID" \
    -d "state=af0ifjsldkj" \
    -d "nonce=abcdefghijk" \
    --data-urlencode "scope=openid" \
    --data-urlencode "redirect_uri=http://127.0.0.1:8251/callback" \
    http://127.0.0.1:8200/v1/identity/oidc/provider/my-provider/authorize | jq

{
  "code": "TmX67vWTJdZhQ136iMEGOFT4c7aCQf6h",
  "state": "af0ifjsldkj"
}

# POST /authorize with form serialization (PUT also works)
$ curl -X POST \
    --header "X-Vault-Token: $(cat ~/.vault-token)" \
    --header "X-Vault-Namespace: $VAULT_NAMESPACE" \
    --header 'Content-Type: application/x-www-form-urlencoded' \
    -d "response_type=code" \
    -d "client_id=$CLIENT_ID" \
    -d "state=af0ifjsldkj" \
    -d "nonce=abcdefghijk" \
    -d "scope=openid" \
    -d "redirect_uri=http://127.0.0.1:8251/callback" \
    -d "max_age=120" \
    http://127.0.0.1:8200/v1/identity/oidc/provider/my-provider/authorize | jq

{
  "code": "BDSc9kVYljxND93YpveBuJtSvguM3AWe",
  "state": "af0ifjsldkj"
}
```


vault/policy_store.go
```diff
@@ -150,6 +150,11 @@ path "sys/tools/hash/*" {
 path "sys/control-group/request" {
     capabilities = ["update"]
 }
+
+# Allow a token to make requests to the Authorization Endpoint for OIDC providers.
+path "identity/oidc/provider/+/authorize" {
```
calvn, line 155: Does this work for an already existing node/cluster? There should already be a default policy on storage so I'm not sure if it would get picked up.

calvn (commented):
Great work! Left a few comments, but looking good so far.

vault/identity_store_oidc_provider_test.go
```diff
@@ -10,8 +10,576 @@ import (
...
+						"client_id":     "",
+						"scope":         "openid",
+						"redirect_uri":  "https://localhost:8251/callback",
+						"response_type": "code",
+						"state":         "abcdefg",
+						"nonce":         "hijklmn",
+					},
+				},
+			},
+		},
```
austingebauer, line 463: Adding test coverage for `max_age` parameter.

vault/identity_store_oidc_provider.go
```diff
@@ -1060,3 +1178,454 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+	}
+	if entity == nil {
+		return logical.ErrorResponse("%s: identity entity must be associated with the request",
+			ErrAuthAccessDenied), nil
+	}
+
+	// Validate that the identity entity associated with the request
+	// is a member of the client assignment's groups or entities
+	var entityHasAssignment bool
+AssignmentsLoop:
```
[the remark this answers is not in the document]
austingebauer: Good idea. Added to its own func in 0155cff.

vault/identity_store_oidc_provider.go
```diff
@@ -277,6 +313,57 @@ func oidcProviderPaths(i *IdentityStore) []*framework.Path {
...
 			HelpDescription: "Query this path to retrieve the configured OIDC Issuer and Keys endpoints, response types, subject types, and signing algorithms used by the OIDC backend.",
 		},
+		{
+			Pattern: "oidc/provider/" + framework.GenericNameRegex("name") + "/authorize",
+			Fields: map[string]*framework.FieldSchema{
+				"name": {
+					Type:        framework.TypeString,
+					Description: "Name of the provider",
+				},
+				"client_id": {
```
fairclothjm, line 323: Should all these fields be marked as `Required`?

vault/identity_store_oidc_provider.go
```diff
@@ -1060,3 +1179,461 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+	}
+
+	// Validate the nonce
+	nonce := d.Get("nonce").(string)
+	if nonce == "" {
+		return logical.ErrorResponse("%s: nonce parameter is required", ErrAuthInvalidRequest), nil
+	}
+
+	// Validate that there is an identity entity associated with the request
+	entity, err := i.MemDBEntityByID(req.EntityID, false)
```
fairclothjm, line 1261: Is it possible for the Entity to exist in Storage but not MemDB? If yes, would that be an issue here?

vault/identity_store_oidc_provider.go
```diff
@@ -1060,3 +1179,461 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+
+	txn.Commit()
+
+	return nil
+}
+
+// memDBDeleteClientByNameInTxn deletes the client with name from memdb using the given txn.
+func (i *IdentityStore) memDBDeleteClientByNameInTxn(ctx context.Context, txn *memdb.Txn, name string) error {
+	if name == "" {
+		return nil
```
fairclothjm, line 1548: should this also return Error `missing client name`?

vault/identity_store_oidc_provider.go
```diff
@@ -1060,3 +1179,461 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+		assignment, err := i.getOIDCAssignment(ctx, s, a)
+		if err != nil {
+			return false, fmt.Errorf("%s: internal server error", ErrAuthServerError)
+		}
+		if assignment == nil {
+			return false, fmt.Errorf("%s: client assignment %q not found", ErrAuthServerError, a)
+		}
+
+		// Get the group names that the entity is a member of
+		entityGroups, err := i.MemDBGroupsByMemberEntityID(entity.GetID(), true, false)
```
fairclothjm, line 1359: Same question here as above. Is it possible for the Groups to exist in Storage but not MemDB? If yes, would that be an issue here?

vault/identity_store_oidc_provider.go
```diff
@@ -1219,3 +1151,632 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
 
 	return entry != nil, nil
 }
+
+func (i *IdentityStore) pathOIDCProviderDiscovery(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
```
austingebauer, line 1155: I moved some code around when resolving conflicts from a merge of main. The following methods have been reviewed and aren't being introduced in this PR:
- `pathOIDCProviderDiscovery`
- `pathOIDCReadProviderPublicKeys`
- `keyIDsReferencedByTargetClientIDs`

vault/identity_store_oidc_provider.go
```diff
@@ -1060,3 +1178,454 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+	ns, err := namespace.FromContext(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("%s: internal server error", ErrAuthServerError)
+	}
+
+	// Get the OIDC provider
+	name := d.Get("name").(string)
+	provider, err := i.getOIDCProvider(ctx, req.Storage, name)
+	if err != nil {
+		return nil, fmt.Errorf("%s: internal server error", ErrAuthServerError)
```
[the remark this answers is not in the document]
austingebauer: I'm wondering what's the best approach for formatting and returning these errors. I want to communicate the spec-defined error code (see [OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1), [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#AuthError)) and an error description. The primary consumer of this will be the Vault UI. The Vault UI will need to return this to the OIDC client in the [spec-defined format](https://openid.net/specs/openid-connect-core-1_0.html#AuthError), which will include the `error`, `error_description`, and `state`.

We could do something like:
```go
// Option 1: Include error code and error string
provider, err := i.getOIDCProvider(ctx, req.Storage, name)
if err != nil {
	return nil, fmt.Errorf("%s: %s", ErrAuthServerError, err)
}

// Option 2: Include error code, specific message, and error string
provider, err := i.getOIDCProvider(ctx, req.Storage, name)
if err != nil {
	return nil, fmt.Errorf("%s: Failed to get OIDC provider - %s", ErrAuthServerError, err)
}
```

Something about making the Vault UI parse a loosely defined string format (i.e., split on `:` to get the error code + message) doesn't feel right. I'm wondering if there is a better approach.

My alternative idea is to return the errors in a JSON format (kind of similar to [identity_store_oidc.go#L1287-L1312](https://github.com/hashicorp/vault/blob/main/vault/identity_store_oidc.go#L1287-L1312)). The UI could then use the well-defined keys to pull out what it needs. The format could be something like:
```json
{
    "error": "server_error",
    "error_description": "A description of the error",
    "state": "not strictly needed, but probably good to return this back to the UI"
}
```

vault/identity_store_oidc_provider.go
```diff
@@ -1060,3 +1178,454 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+
+	// Validate the client ID
+	clientID := d.Get("client_id").(string)
+	if clientID == "" {
+		return logical.ErrorResponse("%s: client_id parameter is required",
+			ErrAuthInvalidClientID), nil
+	}
+	client, err := i.clientByID(ctx, req.Storage, clientID)
+	if err != nil {
+		return nil, fmt.Errorf("%s: internal server error", ErrAuthServerError)
```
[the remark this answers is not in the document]
austingebauer: Will address each of these in the above thread. Resolving.

vault/identity_store_oidc_provider.go
```diff
@@ -1060,3 +1178,454 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+	}
+
+	// Validate that the identity entity associated with the request
+	// is a member of the client assignment's groups or entities
+	var entityHasAssignment bool
+AssignmentsLoop:
+	for _, a := range client.Assignments {
+		assignment, err := i.getOIDCAssignment(ctx, req.Storage, a)
+		if err != nil {
+			return nil, fmt.Errorf("%s: internal server error", ErrAuthServerError)
```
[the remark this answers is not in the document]
austingebauer: Will address each of these in the above thread. Resolving.

vault/policy_store.go
austingebauer: You're right that this doesn't get picked up by an existing instance. Looking closer to see how this can be accomplished.

austingebauer: I wonder if we'll have to instruct those with existing instances to update their default policy? I'm not seeing any patterns in past changes to the default policy around upgrading/merging with what's in storage.

@briankassouf  - Do you happen to know what the typical path forward is for existing instances in this case?

vault/identity_store_oidc_provider.go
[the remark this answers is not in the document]
fairclothjm: +1 to return the errors in a JSON format. That should make it much more predictable for the UI to parse.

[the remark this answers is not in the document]
austingebauer: I agree. I reworked the error handling in 03f43eb. I think that'll be easier for the UI to parse. Let me know your thoughts.

vault/identity_store_oidc_provider.go
```diff
@@ -277,6 +313,57 @@ func oidcProviderPaths(i *IdentityStore) []*framework.Path {
...
+				"max_age": {
+					Type:        framework.TypeInt,
+					Description: "The allowable elapsed time in seconds since the last time the end-user was actively authenticated.",
+				},
+			},
+			Operations: map[logical.Operation]framework.OperationHandler{
+				logical.ReadOperation: &framework.PathOperation{
+					Callback:                    i.pathOIDCAuthorize,
+					ForwardPerformanceStandby:   true,
+					ForwardPerformanceSecondary: false,
```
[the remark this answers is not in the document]
jasonodonnell: 👍 Good question

vault/identity_store_oidc_provider.go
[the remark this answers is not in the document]
fairclothjm: LGTM

vault/identity_store_oidc_provider.go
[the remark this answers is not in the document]
austingebauer: The authorization request doesn't write anything to storage, so I don't see a need to forward the request from  secondary to primary cluster. This means that tokens issued by a provider in a given cluster should be validated within that same cluster. I believe this is a good thing, since we don't want to have to forward each request in the auth flow to the primary cluster.

We will forward the authorization request from standbys to active _within_ a cluster. We need to be able to cache and subsequently look up `authCodeCacheEntry`s for the token endpoint. We don't have a distributed cache, so we'll forward within a cluster to the active.

vault/identity_store_oidc_provider.go
austingebauer: Good question! Dug into this a bit to find out that it's not likely for that to happen (if it did, then it'd suggest a bug in the entity code). The reason is that updates to entities write to memdb _and_ storage (see: [handleEntityUpdateCommon](https://github.com/hashicorp/vault/blob/main/vault/identity_store_entities.go#L272), [upsertEntity](https://github.com/hashicorp/vault/blob/main/vault/identity_store_util.go#L482-L503)). The memdb transaction will be aborted if the storage write fails, which effectively rolls back a partial failure. This is a bit of a different strategy than our client update handler, where we simply invalidate the memdb entry for writes. I think either approach would work.

I think we're safe to use `MemDBEntityByID` here. You can see that the entity read operation uses the same method ([pathEntityIDRead](https://github.com/hashicorp/vault/blob/main/vault/identity_store_entities.go#L322-L340)).

vault/identity_store_oidc_provider.go
austingebauer: See https://github.com/hashicorp/vault/pull/12538#discussion_r709606050. Checked that this is also true for `MemDBGroupsByMemberEntityID`.

vault/identity_store_oidc_provider.go
austingebauer: Thinking on this more, if we followed a similar pattern (client updates write to both memdb and storage), then we could probably list all clients from memdb for the `allowed_client_ids="*"` case instead of listing from storage. I think that'd be a nice optimization. Will add a stretch task for this and proceed with what we have here.

vault/identity_store_oidc_provider.go
austingebauer: Done in 8f9893e.

vault/identity_store_oidc_provider.go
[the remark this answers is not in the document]
fairclothjm: Thanks, that makes sense!

vault/identity_store_oidc_provider.go
fairclothjm: Nice! Thanks for the clarification

vault/identity_store_oidc_provider.go
```diff
@@ -1219,3 +1158,649 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+
+	// Cache the authorization code for a subsequent token exchange
+	if err := i.oidcAuthCodeCache.SetDefault(ns, code, authCodeEntry); err != nil {
+		return errorResp(ErrAuthServerError, err.Error(), state)
+	}
+
+	return &logical.Response{
+		Data: map[string]interface{}{
+			"code":  code,
+			"state": state,
```
fairclothjm, line 1482: 👍

vault/policy_store.go
austingebauer: Result from discussion is that we'll write documentation to note that existing deployments should add this to a policy. Will keep this in the default policy for new deployments. Resolving.

vault/identity_store.go
```diff
@@ -96,7 +98,8 @@ func NewIdentityStore(ctx context.Context, core *Core, config *logical.BackendCo
 		},
 	}
 
-	iStore.oidcCache = newOIDCCache()
+	iStore.oidcCache = newOIDCCache(cache.NoExpiration, cache.NoExpiration)
+	iStore.oidcAuthCodeCache = newOIDCCache(5*time.Minute, 10*time.Minute)
```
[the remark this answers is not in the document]
calvn: I see. Would it make sense to set the cleanup to be the same as the expiry? I'm not sure if there's any value in keeping it around.

vault/identity_store_oidc_provider_test.go
```diff
@@ -10,9 +10,587 @@ import (
...
+		client           *client
+		provider         *provider
+		assignment       *assignment
+		authorizeRequest *logical.Request
+		entityID         string
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
```
calvn, line 28: In addition to just `wantErr`, do we also want to test the actual error value to make sure that it contains the errors defined in this package, e.g. `ErrAuthUnsupportedResponseType`?

vault/identity_store_oidc_provider_test.go
calvn: I don't see test for `max_age`, is this a todo note for yourself?

vault/identity_store_oidc_provider_test.go
austingebauer: Good idea! Changed to test against actual error codes in 7dc219d.

vault/identity_store_oidc_provider_test.go
austingebauer: Yep, this is a todo for myself. Adding coverage for max_age.

vault/identity_store.go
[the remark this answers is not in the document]
austingebauer: Ya, I think running the cleanup interval every 5m is okay. Changed in 5b4f1e1.

vault/identity_store_oidc_provider.go
```diff
@@ -1219,3 +1160,657 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+	}
+
+	return authResponse(code, state, "", "")
+}
+
+// authResponse returns the OIDC Authentication Response. An error response is
+// returned if the given error code is non-empty. For details, see spec at
+//   - https://openid.net/specs/openid-connect-core-1_0.html#AuthResponse
+//   - https://openid.net/specs/openid-connect-core-1_0.html#AuthError
+func authResponse(code, state, errorCode, errorDescription string) (*logical.Response, error) {
```
austingebauer, line 1462: @calvn, @fairclothjm - We talked last week about changing the format of a successful authorize response from the standard `logical.Response` (which included Vault-y things like `lease_id`, `renewable`, etc.) to a JSON format similar to what was already done for error responses. The idea was to closer align with the spec. I checked what this looks like in the audit log and think it's okay to proceed with that approach.

This is the response portion of an audit log for a successful authorize response using this approach:
```
"response": {
    "mount_type": "identity",
    "data": {
      "http_content_type": "hmac-sha256:5986a3f279a6e8d417d805f0e9be432afca5fe2dd5a9cb91a6cf2cf761ecaeb1",
      "http_raw_body": "hmac-sha256:f20c15bb779adc8ee97d8006c3672e3f8c0fcebbc0b5201da8e0d45922bd3721",
      "http_status_code": 200
    }
}
```

Let me know your thoughts!

vault/identity_store_oidc_provider_test.go
austingebauer: Added max_age coverage in 290fbaa.

fairclothjm (commented):
LGTM! I believe we are waiting on changes to max_age. I will hold off on approval until then.

vault/identity_store_oidc_provider.go
```diff
@@ -1060,3 +1178,454 @@ func (i *IdentityStore) pathOIDCProviderExistenceCheck(ctx context.Context, req
...
+	// Returning ErrAuthMaxAgeReAuthenticate will enforce the user to re-authenticate via
+	// the user agent.
+	if maxAgeRaw, ok := d.GetOk("max_age"); ok {
+		maxAge := maxAgeRaw.(int)
+		if maxAge < 0 {
+			return logical.ErrorResponse("%s: max_age must be greater than zero",
+				ErrAuthInvalidRequest), nil
+		}
+		if maxAge == 0 {
+			// TODO: solve for the potential UI loop here or make max_age=0 invalid
```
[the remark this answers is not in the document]
austingebauer: Removed the special case for `max_age=0` in 6920710. We don't need to support `max_age=0` initially. We can add this feature incrementally once the design is worked out.

vault/identity_store_oidc_provider_test.go
```diff
@@ -16,12 +16,50 @@ import (
...
+		Operation: logical.CreateOperation,
+		Data:      map[string]interface{}{},
+		Storage:   storage,
+	})
+	expectSuccess(t, resp, err)
+
+	// Create an entity
+	resp, err = c.identityStore.HandleRequest(ctx, &logical.Request{
+		Path:      "entity",
+		Operation: logical.UpdateOperation,
```
fairclothjm, line 35: nit: since we are doing a create maybe this should be `CreateOperation`?

vault/identity_store_oidc_provider_test.go
```diff
@@ -16,12 +16,50 @@ import (
...
+		},
+	})
+	expectSuccess(t, resp, err)
+	assert.NotNil(t, resp.Data["id"])
+	entityID := resp.Data["id"].(string)
+
+	// Create a group
+	resp, err = c.identityStore.HandleRequest(ctx, &logical.Request{
+		Path:      "group",
+		Operation: logical.UpdateOperation,
```
fairclothjm, line 47: nit: since we are doing a create maybe this should be `CreateOperation`?

fairclothjm (approved):

vault/identity_store_oidc_provider_test.go
austingebauer: Ya, I originally expected the same. The entity and groups endpoints don't accept a `CreateOperation` for some reason (see [identity_store_entities.go#L61](https://github.com/hashicorp/vault/blob/main/vault/identity_store_entities.go#L61)).

vault/identity_store_oidc_provider_test.go
austingebauer: See comment above.

vault/identity_store_oidc_provider.go
```diff
@@ -1384,6 +1386,17 @@ func (i *IdentityStore) pathOIDCAuthorize(ctx context.Context, req *logical.Requ
...
 	}
 
+	// We don't support the request or request_uri parameters. If they're provided,
+	// the appropriate errors must be returned. For details, see the spec at:
+	// https://openid.net/specs/openid-connect-core-1_0.html#RequestObject
+	// https://openid.net/specs/openid-connect-core-1_0.html#RequestUriParameter
+	if _, ok := d.Raw["request"]; ok {
+		return authResponse("", state, ErrAuthRequestNotSupported, "request parameter is not supported")
+	}
+	if _, ok := d.Raw["request_uri"]; ok {
```
fairclothjm, line 1396: I was looking into the spec to understand this better and I think we may need to include the `request_uri_parameter_supported=false` in the Discovery endpoint response. The spec states that the default is `true` if it is not included. https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

austingebauer: Nice catch! Added in the other PR https://github.com/hashicorp/vault/pull/12623.

calvn (approved):

Issue #12537, opened 2021-09-13
Support CORS Credentials


betandr:
We have a client to Vault which uses MTLS with a certificate to authenticate a user. We'd like to use this client with Vault although need an extra CORS header of `Access-Control-Allow-Credentials`. 

The proposal is to add this header to the `http` package in the `http/cors.go`, such as:

```
w.Header().Set("Access-Control-Allow-Credentials", credentials)
```

...where `credentials` would be true. Currently the `Access-Control-Allow-Origin` header value is true if the `Origin` request is set so maybe it could do something like this to provide the ``Access-Control-Allow-Credentials: true` header if credentials are supplied. 

How would you feel about a PR of this type? Thanks!

hsimon-hashicorp:
Hi @betandr! If you'd like to file a PR for us to review, we'd be happy to do so! Thank you so much for volunteering. :)

mickael-hc:
Thanks for opening this issue. Once the changes in https://github.com/hashicorp/vault/pull/12485 are released, it will be possible to set headers in the configuration file, including `Access-Control-Allow-Credentials` . Release is tentatively scheduled for Vault 1.9.0

PR #12544, opened 2021-09-14
enhancement: Support CORS credentials

precia-bbc:
Raised from https://github.com/hashicorp/vault/issues/12537 "Support CORS credentials"

This PR adds an additional CORS header to the ones presently used within Vault. It enables cross origin requests to be made from domains behind certificate based authentication, and for those certificates to be used as an authentication mechanism with Vaults /v1/auth/cert/login path.

The additional header is documented on MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials

Associated tests have been updated to the best of my knowledge

### DIFF
--- a/http/cors.go
+++ b/http/cors.go
@@ -50,6 +50,7 @@ func wrapCORSHandler(h http.Handler, core *vault.Core) http.Handler {
 		}
 
 		w.Header().Set("Access-Control-Allow-Origin", origin)
+		w.Header().Set("Access-Control-Allow-Credentials", "true")
 		w.Header().Set("Vary", "Origin")
 
 		// apply headers for preflight requests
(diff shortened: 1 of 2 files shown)


mickael-hc:
Hello @precia-bbc thank you for your contribution and for opening the associated issue.

If I understand this change correctly, the changes proposed will set the `Access-Control-Allow-Credentials` header for _all_ Vault requests, for all Vault instances with CORS enabled. The impact of this can vary dramatically depending on the environment in which this is deployed.

If these assumptions are correct, the impact can vary depending on the environment, and in some cases, introduce significant risks through this new default: this means that for all CORS requests, Vault will forward the Bearer token if it was set in the original request. Currently, one must also explicitly set X-Vault-Token in `allowed_headers` of the Vault CORS config in https://www.vaultproject.io/api/system/config-cors .

I think it would be more prudent to make this change opt-in. There’s a PR that is currently in review stage (https://github.com/hashicorp/vault/pull/12485) that will provide Vault operators with the ability to set custom headers. It’s tentatively scheduled for release in Vault 1.9.0. Those changes will allow you to add that header in your configuration.

I'm going to close this PR, but please don't hesitate to reopen if you think I misunderstood.

PR #12549, opened 2021-09-14
Backport cert diff algo signing #12514 to 1.7.x

sgmiller:


victorr (approved):
I verified that the changes were the same as in the original fix.
👍

PR #12563, opened 2021-09-15
Port: Premature Rotation For autorotate

HridoyRoy:
Port of: https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/28/files

### DIFF
--- a/builtin/logical/database/path_roles.go
+++ b/builtin/logical/database/path_roles.go
@@ -541,12 +541,8 @@ func (b *databaseBackend) pathStaticRoleCreateUpdate(ctx context.Context, req *l
 			return nil, err
 		}
 		item, err = b.popFromRotationQueueByKey(name)
-		if err != nil || item == nil {
-			b.Logger().Warn("expected role to exist in rotation queue but none found", "role", name, "error", err)
-			item = &queue.Item{
-				Key:      name,
-				Priority: lvr.Add(role.StaticAccount.RotationPeriod).Unix(),
-			}
+		if err != nil {
+			return nil, err
 		}
 
 	}
--- a/builtin/logical/database/path_roles.go
+++ b/builtin/logical/database/path_roles.go
@@ -543,10 +543,7 @@ func (b *databaseBackend) pathStaticRoleCreateUpdate(ctx context.Context, req *l
 		}
 		item, err = b.popFromRotationQueueByKey(name)
 		if err != nil {
-			// This update operation is actually performing a create.
-			item = &queue.Item{
-				Key: name,
-			}
+			return nil, err
 		}
 	}
 
--- a/builtin/logical/database/path_roles.go
+++ b/builtin/logical/database/path_roles.go
@@ -537,8 +537,15 @@ func (b *databaseBackend) pathStaticRoleCreateUpdate(ctx context.Context, req *l
 			return nil, err
 		}
 		item, err = b.popFromRotationQueueByKey(name)
-		if err != nil {
-			return nil, err
+		if err != nil || item == nil {
+			// in this case we want to push a new role into the rotation queue
+			if err := b.pushItem(&queue.Item{
+				Key:      name,
+				Priority: lvr.Add(role.StaticAccount.RotationPeriod).Unix(),
+			}); err != nil {
+				return nil, err
+			}
+			return nil, nil
 		}
 	}
 	item.Priority = lvr.Add(role.StaticAccount.RotationPeriod).Unix()
--- a/builtin/logical/database/path_roles.go
+++ b/builtin/logical/database/path_roles.go
@@ -507,8 +507,9 @@ func (b *databaseBackend) pathStaticRoleCreateUpdate(ctx context.Context, req *l
 	case logical.CreateOperation:
 		// setStaticAccount calls Storage.Put and saves the role to storage
 		resp, err := b.setStaticAccount(ctx, req.Storage, &setStaticAccountInput{
-			RoleName: name,
-			Role:     role,
+			RoleName:   name,
+			Role:       role,
+			CreateUser: createRole,
 		})
 		if err != nil {
 			if resp != nil && resp.WALID != "" {
--- a/builtin/logical/database/path_roles.go
+++ b/builtin/logical/database/path_roles.go
@@ -520,6 +520,10 @@ func (b *databaseBackend) pathStaticRoleCreateUpdate(ctx context.Context, req *l
 					merr = multierror.Append(merr, fmt.Errorf("failed to clean up WAL from failed role creation: %w", walDeleteErr))
 					err = merr.ErrorOrNil()
 				}
+			} else {
+				// There is no walID, but role creation failed. In this case we must simply
+				// return the error.
+				return nil, err
 			}
 		}
 		// guard against RotationTime not being set or zero-value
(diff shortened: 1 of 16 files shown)


builtin/logical/database/path_roles.go
```diff
@@ -482,19 +504,32 @@ func (b *databaseBackend) pathStaticRoleCreateUpdate(ctx context.Context, req *l
...
+				b.Logger().Debug("deleting WAL for failed role creation", "WAL ID", resp.WALID, "role", name)
+				walDeleteErr := framework.DeleteWAL(ctx, req.Storage, resp.WALID)
+				if walDeleteErr != nil {
+					b.Logger().Debug("failed to delete WAL for failed role creation", "WAL ID", resp.WALID, "error", walDeleteErr)
+					var merr *multierror.Error
+					merr = multierror.Append(merr, err)
+					merr = multierror.Append(merr, fmt.Errorf("failed to clean up WAL from failed role creation: %w", walDeleteErr))
+					err = merr.ErrorOrNil()
+				}
+			}
```
kalafut, line 526: I don't know if this applies to OpenLDAP as well, but we're swallowing an error here. For example, I had configured my connection with `allowed_roles=["my-role"]`, was trying to set up a static role `foo-role`, was getting success messages on the Write operation but "role not found" in the server logs and no role actually created. Turned out the Write was actually returning `"'foo-role' is not an allowed role"` but that was being swallowed.

HridoyRoy: Thanks! Fixed: https://github.com/hashicorp/vault/pull/12563/commits/b1c8a06e10c2b35379d16241c0f8f2c542d117cc

builtin/logical/database/rotation.go
```diff
@@ -324,37 +328,58 @@ func (b *databaseBackend) setStaticAccount(ctx context.Context, s logical.Storag
...
 
 	if output.WALID == "" {
+		newPassword, err = dbi.database.GeneratePassword(ctx, b.System(), dbConfig.PasswordPolicy)
+		if err != nil {
+			return output, err
+		}
 		output.WALID, err = framework.PutWAL(ctx, s, staticWALKey, &setCredentialsWAL{
 			RoleName:          input.RoleName,
 			Username:          config.Username,
 			NewPassword:       config.Password,
```
pcman312, line 376:
```suggestion
			NewPassword:       newPassword,
```

builtin/logical/database/path_roles.go
```diff
@@ -504,17 +540,20 @@ func (b *databaseBackend) pathStaticRoleCreateUpdate(ctx context.Context, req *l
 		if err := req.Storage.Put(ctx, entry); err != nil {
 			return nil, err
 		}
+		item, err = b.popFromRotationQueueByKey(name)
+		if err != nil || item == nil {
```
calvn, line 544: It looks like `popFromRotationQueueByKey` cannot return a nil `item` and a nil `error` at the same time, and if `item` is nil it would return a `queue.ErrEmpty` error.
```suggestion
		if err != nil {
```

builtin/logical/database/path_roles.go
```diff
@@ -504,17 +540,20 @@ func (b *databaseBackend) pathStaticRoleCreateUpdate(ctx context.Context, req *l
...
+			b.Logger().Warn("expected role to exist in rotation queue but none found", "role", name, "error", err)
+			item = &queue.Item{
+				Key:      name,
+				Priority: lvr.Add(role.StaticAccount.RotationPeriod).Unix(),
+			}
```
calvn, line 549:
```suggestion
			return nil, err
```

builtin/logical/database/rotation.go
```diff
@@ -324,37 +328,58 @@ func (b *databaseBackend) setStaticAccount(ctx context.Context, s logical.Storag
...
 	}
 
 	if output.WALID == "" {
+		newPassword, err = dbi.database.GeneratePassword(ctx, b.System(), dbConfig.PasswordPolicy)
+		if err != nil {
+			return output, err
+		}
 		output.WALID, err = framework.PutWAL(ctx, s, staticWALKey, &setCredentialsWAL{
 			RoleName:          input.RoleName,
 			Username:          config.Username,
```
calvn, line 375: You can use `input.Role.StaticAccount.Username` here and remove the `config` declaration that's on L363 since its only reference is in here.

calvn (commented):
Took a side-by-side look against https://github.com/hashicorp/vault-plugin-secrets-openldap/pull/28 (78553e326) and the implementation logic looks mostly the same aside from one code block in `builtin/logical/database/path_roles.go`.

builtin/logical/database/path_roles.go
tomhjp: Previously, the API would treat updates on static roles that don't exist mostly like creates, except it wouldn't trigger an immediate password rotation. That's a bug, because it means Vault is tracking a static account that it doesn't know the password for, but it's a separate bug and one that should be exceedingly rare, because we have an existence check on static roles that should ensure we almost always get the correct create/update operation for a static role depending on whether it already exists. I think it might be possible to have a race if there is one delete and one update request in flight at the same time, which is why I say it should be exceedingly rare to encounter this bug.

However, given the size of this PR, and the number of versions we want to backport to, I'd rather not take the risk of changing this bit of API unless we think it's part of the problem.

builtin/logical/database/rotation_test.go
```diff
@@ -1094,6 +1110,300 @@ func TestBackend_StaticRole_Rotate_Invalid_Role(t *testing.T) {
...
+			initialPassword := role.StaticAccount.Password
+
+			// Set up a WAL for our test case
+			framework.PutWAL(ctx, config.StorageView, staticWALKey, tc.wal)
+			requireWALs(t, config.StorageView, 1)
+			// Reset the rotation queue to simulate startup memory state
+			b.credRotationQueue = queue.New()
+
+			// Now finish the startup process by populating the queue, which should discard the WAL
+			b.initQueue(ctx, config, consts.ReplicationDRPrimary)
```
calvn, line 1220: @briankassouf should this be against `consts.ReplicationPerformancePrimary`? I noticed that `initQueue` doesn't exclude `consts.ReplicationDRPrimary` but I wonder whether it should.

calvn (approved):
One question on `initQueue` that came up during testing review. Otherwise looks good!

builtin/logical/database/rotation.go
```diff
@@ -324,37 +327,53 @@ func (b *databaseBackend) setStaticAccount(ctx context.Context, s logical.Storag
 		return output, err
 	}
 
-	dbi.RLock()
-	defer dbi.RUnlock()
+	dbi.Lock()
+	defer dbi.Unlock()
```
briankassouf, line 331: Out of curiosity what was the reasoning for switching to a write lock here?

builtin/logical/database/rotation_test.go
briankassouf: I think for this test's purposes you could use 0 (consts.ReplicationUnknown) here. ReplicationDRPrimary should not be excluded since a Perf Primary could also be a DR Primary. Only secondaries have read-only storage.

calvn: I changed this to `consts.ReplicationPerformancePrimary`, but realized that it should be `consts.ReplicationPerformanceDisabled` so that it's an accurate representation for OSS, though my question around the conditional within `initQueue` remains.

builtin/logical/database/rotation.go
calvn: I've reverted this in dec64cd.

briankassouf (approved):

builtin/logical/database/rotation_test.go
calvn: Ah, and if a cluster that is a DR primary is also a Perf Secondary, it would be excluded. I'll change this to ReplicationUnknown then.

Issue #12225, opened 2021-08-02
[Token Auth] Roles not returning token_no_default_policy 


tristanmorgan:
**Describe the bug**
The status of token_no_default_policy is not being returned when reading token role configuration.

**To Reproduce**
1. Run `vault auth enable token`
1. Run `vault write auth/token/roles/test token_no_default_policy=true`
2. Run `vault read auth/token/roles/test`
3. token_no_default_policy is not being returned in response

**Expected behavior**
token_no_default_policy should be displayed (if the default value had been overwritten?)

**Environment:**
* Vault Server Version: 1.8.0, 1.7.3, 1.6.5
* Vault CLI Version : 1.8.0
* Server Operating System/Architecture: Linux amd64

Vault server configuration file(s):

```hcl
-dev","-dev-listen-address=0.0.0.0:8200
```

**Additional context**
I thought it was a problem with the Terraform Vault Provider but even checking cURL showed it was Vault.

tristanmorgan:
**Additional additional context**
`vault read sys/raw/sys/token/roles/test` does show that data is present.

tristanmorgan:
Seems the code is missing from _vault/token_store.go_ `tokenStoreRoleRead()`

hsimon-hashicorp:
Hi @tristanmorgan! We will be taking a look at this soon, however, please feel free to work on a PR and submit it for review too if this is an issue you might feel comfortable tackling. Thanks! :)

PR #12565, opened 2021-09-15
[VAULT-3519] Return no_default_policy on token role read

pmmukh:
- Returns the token_no_default_policy config on token role read
- Updates tests to check the same

Fixes https://github.com/hashicorp/vault/issues/12225


vault/token_store.go
```diff
@@ -3208,6 +3208,10 @@ func (ts *TokenStore) tokenStoreRoleRead(ctx context.Context, req *logical.Reque
 		resp.Data["token_num_uses"] = role.TokenNumUses
 	}
 
+	if role.TokenNoDefaultPolicy {
+		resp.Data["token_no_default_policy"] = true
```
vishalnayak, line 3212: Is there a reason not to return `false` if it is not set?

pmmukh: not really, tbh. I wasn't sure of the criteria on which some fields are always in the response, and some only in the response if they've been set by user input. Since this is currently omitted, I erred on the side of the change that affects existing behavior least, so went with only returning this value if set and keeping the response as is otherwise.

vishalnayak: I would add it.

pmmukh: Roger, that sounds good to me too, will add it in 👍

pmmukh: 6e1ba28

vishalnayak (approved):
[Optional] You might want to update the sample response on the API docs for the read.

pmmukh:
> [Optional] You might want to update the sample response on the API docs for the read.
Ah yeah will do that, thanks for the catch!

Issue #12556, opened 2021-09-15
vault/api token renew endpoint 


findmyname666:
Hello,
I noticed nonconformity between Vault DOC and Vault API client.
This issue focuses on renew endpoint for tokens - **/auth/token/renew**.

1) The [DOC](https://www.vaultproject.io/api/auth/token#renew-a-token) says that client should do **POST** request with some increment. Increment can be of string type e.g. you can probably use value such **1h** as parsed by time.Duration.

2) The [golang Vault API lib](https://github.com/hashicorp/vault/blob/api/v1.1.1/api/auth_token.go#L139) do **PUT** request with increment which has to be of **int** type. However used URL contains API version **/v1/auth/token/renew**.

The same apply also to endpoint [renew-self](https://github.com/hashicorp/vault/blob/api/v1.1.1/api/auth_token.go#L159).
_Please note the client lib seems to work as expected so this issue doesn't affect it's functionality so far._

Side note: DOC mentions that provided increment can be ignored by Vault server. However it is not clear why and in which cases. I just noticed that it isn't necessary for periodic tokens.

```
increment (string: "") - An optional requested lease increment can be provided. This increment may be ignored.
```

pmmukh:
Hi @findmyname666 ! Thanks for filing this issue! You are correct, there is indeed a discrepancy between the docs and the API, the API accepts an integer input, like how the client library works. Also yes, after looking at the code I too think `This increment may be ignored.` applies to periodic tokens, and that the current wording is not very clear about that. I'll update the docs with both changes, thanks for pointing them out!

pmmukh:
Ah this is actually because both go's time.Duration strings, and integer representing seconds, are valid inputs here, based on this input type https://github.com/hashicorp/vault/blob/main/sdk/framework/field_type.go#L16, I'll update the docs to be clearer on that.

findmyname666:
@pmmukh thx a lot.

pmmukh:
I updated the docs to be clearer around when the increment might be ignored, and how the string input is provided. I didn't update the datatype in the docs though, cause while the field can be either string or integer, we'd like only one option in our external docs, so as long as we're saying it's a string everywhere, that should be fine.

PR #12572, opened 2021-09-16
Update token renew docs

pmmukh:
- Clarifies some details around the `increment` parameter for token renewal

Fixes https://github.com/hashicorp/vault/issues/12556

### DIFF
--- a/command/token_renew.go
+++ b/command/token_renew.go
@@ -77,10 +77,10 @@ func (c *TokenRenewCommand) Flags() *FlagSets {
 		Default:    0,
 		EnvVar:     "",
 		Completion: complete.PredictAnything,
-		Usage: "Request a specific increment for renewal. Vault will not honor this" +
-			"request for periodic tokens. If not supplied, Vault will use the default " +
-			"TTL. This is specified as a numeric string with suffix like \"30s\" " +
-			"or \"5m\".",
+		Usage: "Request a specific increment for renewal. This increment may " +
+			"not be honored, for instance in the case of periodic tokens. If not " +
+			"supplied, Vault will use the default TTL. This is specified as a " +
+			"numeric string with suffix like \"30s\" or \"5m\".",
 	})
 
 	return set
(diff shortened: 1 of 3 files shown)


command/token_renew.go
```diff
@@ -77,8 +77,8 @@ func (c *TokenRenewCommand) Flags() *FlagSets {
 		Default:    0,
 		EnvVar:     "",
 		Completion: complete.PredictAnything,
-		Usage: "Request a specific increment for renewal. Vault is not required " +
-			"to honor this request. If not supplied, Vault will use the default " +
+		Usage: "Request a specific increment for renewal. Vault will not honor this" +
+			"request for periodic tokens. If not supplied, Vault will use the default " +
```
mladlow, line 81: I have no idea about this, but if we think there might be cases other than periodic tokens where increment isn't honored, we may want to leave this more open and hybridize the statement between "is not required to honor" and restricting the limit to periodic tokens?

mladlow: Like if there are global maximums that would restrict how long the token could be renewed for?

website/content/api-docs/auth/token.mdx
```diff
@@ -337,8 +337,10 @@ if there is a lease associated with it.
 
 - `token` `(string: <required>)` - Token to renew. This can be part of the URL
   or the body.
-- `increment` `(string: "")` - An optional requested lease increment can be
-  provided. This increment may be ignored.
+- `increment` `(string: "")` - An optional requested increment duration can be
```
mladlow, line 340: I think this is much clearer!

mladlow (approved):

command/token_renew.go
pmmukh: Aha yes that is absolutely correct, just noticed this after you mentioned it https://github.com/hashicorp/vault/blob/main/sdk/framework/lease.go#L100, I'll wordsmith this a bit more, thanks for the catch!

website/content/api-docs/auth/token.mdx
pmmukh: thanks! 😀

command/token_renew.go
pmmukh: @mladlow how does "This increment may not be honored, for instance in the case of periodic tokens." sound ?

mladlow: Sounds good to me!

pmmukh: 1c8dad2

PR #12581, opened 2021-09-19
Update plugin proto to send tls.ConnectionState (Op.2)

tep:
_This is a resubmission of #9240; it still fixes #9213_

In order for external plugins to examine the client certificate(s)
presented during the original request into Vault, we need to marshal
that request's tls.ConnectionState across the gRPC boundary from the
Vault server to the external plugin process.

To that end, this change updates the protocol buffer (and associated
translation logic) used for communicating with external plugins to
include the logical.Connection's "ConnState" value -- thus making it
available to the logical.Request passed to callbacks defined by
external plugins.


hsimon-hashicorp:
Hi @tep! It looks like there's some merge conflicts. Can you resolve those? Then, the tests all look good and we can do a deeper dive on the review.

tep:
@hsimon-hashicorp  -- Conflicts resolved

PR #12582, opened 2021-09-20
Patch to support VAULT_HTTP_PROXY variable

mikeyyuen:
# What has been changed

This patch creates VAULT_HTTP_PROXY which can be used to override the default behavior of the Vault API Client.

- Without VAULT_HTTP_PROXY set the Vault API Client will continue to work as before using the default Go HTTP behavior, reading `HTTPS_PROXY` as before so the changes should likely not affect any existing users.
- With VAULT_HTTP_PROXY set, the Vault API Client creates a specific Go http transport with the proxy set via the Go API forcing it to use the specified proxy to communicate with the Vault binary.

# Why is this useful

If the Vault Client has been embedded in software (e.g. terraform provider, consul-template or other) that might need to use different proxies to talk to vault vs other upstream services such as AWS, Azure, GCP (as described in https://github.com/hashicorp/terraform-provider-vault/issues/1150), this allows the vault behavior to be specifically controlled.


hsimon-hashicorp:
Hello, and welcome to the HashiCorp Vault community! Please note that we are moving to a process where PRs need to be associated with an issue, either an existing one or a new issue, so that the community and the engineers can discuss implementation details, strategy, and where the request fits into an existing roadmap. Additionally, all PRs do need to have the CLA signed. Please feel free to create or link an existing issue to this PR. Thanks!

davidducros:
The CLA signing has now been recognised properly and an issue to cover this has been created here: https://github.com/hashicorp/vault/issues/12613

api/client.go
```diff
@@ -173,7 +180,23 @@ func DefaultConfig() *Config {
 		Backoff:      retryablehttp.LinearJitterBackoff,
 	}
 
+	if err := config.ReadEnvironment(); err != nil {
```
ncabatoff, line 183: I'm a bit nervous about moving where we call ReadEnvironment.  I haven't identified any obvious issues with it, but I haven't been able to rule out the possibility of it breaking something.  Can we instead have ReadEnvironment include the assignment to transport.Proxy?

api/client.go
```diff
@@ -42,6 +42,7 @@ const (
 	EnvVaultToken         = "VAULT_TOKEN"
 	EnvVaultMFA           = "VAULT_MFA"
 	EnvRateLimit          = "VAULT_RATE_LIMIT"
+	EnvHTTPSProxy         = "VAULT_HTTPS_PROXY"
```
ncabatoff, line 45: Since we don't enforce that this will be using scheme `https`, it feels like the more general name for the var would be VAULT_HTTP_PROXY.

api/client.go
davidducros: I also don't see any reason why moving it would matter, but it's easy enough to apply the override in `ReadEnvironment`. I've updated to do that instead.

ncabatoff:
Changes look good, could you also update https://github.com/hashicorp/vault/blob/main/website/content/docs/commands/index.mdx#environment-variables please?

ncabatoff:
Oh, and we'll also need a changelog entry, as described in https://github.com/hashicorp/vault/blob/main/CONTRIBUTING.md#changelog-entries

ncabatoff (approved):

ncabatoff:
Looks good, thanks!  You'll need to merge in the latest changes from `main` so the tests pass before I can merge this.

PR #12608, opened 2021-09-22
Fix client.go

HgAmiui9:
Modify one annotation.
`ConfigureTLS takes a set of TLS configurations and applies those to the HTTP client.`


hsimon-hashicorp (approved):

hsimon-hashicorp:
Thanks for your contribution! :)

PR #12622, opened 2021-09-23
UI/bar chart updates

hellobontempo:
To Dos:
- [x] Chart re-renders when new dates are queries
- [x] Awry tooltip
- [x] Double rendering of total values 
- [x] Double rendering of y-axis labels


arnav28 (commented):
We can get rid of dataset getter on line 67?

hellobontempo:
> We can get rid of dataset getter on line 67?

Did you test it with this? When I took it out the chart didn't render initially, but perhaps I missed something..

arnav28 (approved):

PR #12629, opened 2021-09-24
Update Azure secrets engine to use MS Graph

pcman312:
Pulls in https://github.com/hashicorp/vault-plugin-secrets-azure/pull/67 to start migrating away from the deprecated Azure Active Directory API to the Microsoft Graph API.

Also adds docs for the new feature flag: `use_microsoft_graph_api` and a warning that current users should migrate their configurations before the AAD API is removed by Microsoft.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -107,7 +107,7 @@ require (
 	github.com/hashicorp/vault-plugin-mock v0.16.1
 	github.com/hashicorp/vault-plugin-secrets-ad v0.10.0
 	github.com/hashicorp/vault-plugin-secrets-alicloud v0.9.0
-	github.com/hashicorp/vault-plugin-secrets-azure v0.10.0
+	github.com/hashicorp/vault-plugin-secrets-azure v0.6.3-0.20210924190759-58a034528e35
 	github.com/hashicorp/vault-plugin-secrets-gcp v0.10.2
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -107,7 +107,7 @@ require (
 	github.com/hashicorp/vault-plugin-mock v0.16.1
 	github.com/hashicorp/vault-plugin-secrets-ad v0.10.0
 	github.com/hashicorp/vault-plugin-secrets-alicloud v0.9.0
-	github.com/hashicorp/vault-plugin-secrets-azure v0.10.0
+	github.com/hashicorp/vault-plugin-secrets-azure v0.6.3-0.20210924190759-58a034528e35
```
pcman312, line 110: I don't know why this is going to 0.6.3, but the commit hash is correct.

jasonodonnell: I think the repo needs to be tagged to v0.11.0 and we add that here. Thoughts?

pcman312: I was going to wait on tagging the repo since I'm going to be working on root credential rotation next. I can tag it now though if you prefer?

jasonodonnell: Yeah good point, let's wait on tagging.

jasonodonnell (approved):

PR #12635, opened 2021-09-24
core: set namespace within GeneratePasswordFromPolicy

calvn:
This PR makes a change on dynamicSystemView.GeneratePasswordFromPolicy to set a namespace value on its incoming context based on the value that exists on its mount entry.

GeneratePasswordFromPolicy is accessible to plugins via framework.Backend.System(), but callback handlers on the plugin side don't carry over the namespace its context, thus leading to cases where namespace value is not set when called by plugins. The dynamic system view already has visibility into its namespace through its mountEntry so we can just set the context's namespace to the value from there.

Prior to this change, static role credentials rotations would error due to missing namespace information.

Auto-rotation failure:

```
2021-09-23T11:28:38.363-0700 [ERROR] secrets.openldap.openldap_084ff8e4.openldap.vault-plugin-secrets-openldap: unable to rotate credentials in periodic function: error="unable to generate password: rpc error: code = Internal desc = failed to generate password: failed to retrieve password policy: no namespace" timestamp=2021-09-23T11:28:38.363-0700
```

Manual rotation failure:
```
Error writing data to openldap/rotate-role/mary: Error making API request.

Namespace: test/ad1/
URL: PUT http://localhost:8200/v1/openldap/rotate-role/mary
Code: 500. Errors:

* 1 error occurred:
	* unable to generate password: rpc error: code = Internal desc = failed to generate password
```

### DIFF
--- a/builtin/logical/database/backend_test.go
+++ b/builtin/logical/database/backend_test.go
@@ -45,7 +45,7 @@ func getCluster(t *testing.T) (*vault.TestCluster, logical.SystemView) {
 
 	os.Setenv(pluginutil.PluginCACertPEMEnv, cluster.CACertPEMFile)
 
-	sys := vault.TestDynamicSystemView(cores[0].Core)
+	sys := vault.TestDynamicSystemView(cores[0].Core, nil)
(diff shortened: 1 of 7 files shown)


kalafut (approved):

tomhjp (approved):
LGTM 👍

PR #12638, opened 2021-09-26
Add Counters endpoints to api client 

julienMichaud:
Hi, 

Tried to add these endpoints https://www.vaultproject.io/api-docs/system/internal-counters to the api client.

### DIFF
--- a/api/sys_counter.go
+++ b/api/sys_counter.go
@@ -1,0 +1,240 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+func (c *Sys) ListCounters() (*Counter, error) {
+	r := c.c.NewRequest("GET", "/v1/sys/internal/counters/config")
+
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	resp, err := c.c.RawRequestWithContext(ctx, r)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	var result Counter
+	err = resp.DecodeJSON(&result)
+
+	return &result, err
+}
+
+func (c *Sys) ConfigCounters(config CountersConfig) error {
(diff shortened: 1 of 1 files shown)


hsimon-hashicorp:
Hi @julienMichaud! Is there an issue that's open for this PR? In order to make sure that the Vault engineering team understands the scope of the work being done in the PR and how it slots into the existing work and roadmap, we like to be able to discuss the problem, use cases, and implementation details in an issue. Please link the issue to the PR and we can take a look. Thanks!

julienMichaud:
Hi, 

I dont think there is any issue related to this PR. 

I wanted to develop the counter ressources for the terraform-provider-vault, but the endpoints were not available through the api client. 
So I tried to create them myself and put a PR.

HridoyRoy (changes requested):
Hi @julienMichaud , thanks for submitting this PR! 

Unfortunately, since these are internal endpoints, we don't guarantee backward compatibility for them. Thus I'm hesitant to add these methods to the api package as it encourages reliance on `sys/internal`. 

It is still possible to achieve this functionality through the api, however, if absolutely necessary, through `Logical`, e.g. ` client.Logical().Write` and so on. 

However, on the other hand if this change does help facilitate a supported/greenlit Vault workflow that I'm overlooking, then we may want to add these helpers to the api after all. To that end, I'd love to hear more about the context/use-case/workflow you had in mind when submitting this PR. 

Thanks again!

ccapurso:
Hi, @julienMichaud. Once again, thank you for the contribution! As @HridoyRoy has mentioned, we are hesitant to include internal endpoints in the `Logical` API client. These endpoints are subject to change and backward compatibility is not a guarantee. As such, I’m going to go ahead and close this PR. A suggested workaround is to use existing `Logical` API methods (e.g. `client.Logical().Read("sys/internal/counters/entities")`).

PR #12652, opened 2021-09-27
Add missing read unlock calls in transit backend code

divyapola5:
This PR is a followup for #12418 and in summary includes the following changes:
1 - Add missing read unlock calls for cacheSizeChanged variable
2 - Invoke initCache in pathCacheConfigRead routine if needed

### DIFF
--- a/builtin/logical/transit/backend.go
+++ b/builtin/logical/transit/backend.go
@@ -122,11 +122,13 @@ func (b *backend) GetPolicy(ctx context.Context, polReq keysutil.PolicyRequest, 
 		currentCacheSize := b.lm.GetCacheSize()
 		storedCacheSize, err := GetCacheSizeFromStorage(ctx, polReq.Storage)
 		if err != nil {
+			b.configMutex.RUnlock()
 			return nil, false, err
 		}
 		if currentCacheSize != storedCacheSize {
 			err = b.lm.InitCache(storedCacheSize)
 			if err != nil {
+				b.configMutex.RUnlock()
 				return nil, false, err
 			}
 		}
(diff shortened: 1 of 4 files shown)


changelog/12652.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
+Add missing read unlock calls in transit backend code.
```
victorr, line 2: I don't have experience writing changelog entries but it seems to me that as the audience are Vault users/operators, you don't need to mention detailed information such as methods called etc. Perhaps something more generic along the lines of: fixed a problem with the transit cache, and made it unnecessary to reload the backend when the configuration changes.

builtin/logical/transit/backend.go
```diff
@@ -122,11 +122,13 @@ func (b *backend) GetPolicy(ctx context.Context, polReq keysutil.PolicyRequest,
 		currentCacheSize := b.lm.GetCacheSize()
 		storedCacheSize, err := GetCacheSizeFromStorage(ctx, polReq.Storage)
 		if err != nil {
+			b.configMutex.RUnlock()
```
victorr, line 125: These code changes are 🆗, but please consider making the equivalent changes as you did in the Transform plugin (extracting a method, changing the if statement into a guard).

victorr (approved):
The changes look good to me, but please verify the changelog entry with someone (maybe Scott?)

changelog/12652.txt
sgmiller: We probably already have the changelog entry for the overall cache issue, so I think we can just remove this.

sgmiller (changes requested):

builtin/logical/transit/backend.go
sgmiller: I'd like to get this merged ASAP so as not to block the point releases.  Divya: can you follow up sometime after with a PR that aligns this to the transform changes?

changelog/12652.txt
divyapola5: Thank you. Removed the changelog entry for this

builtin/logical/transit/backend.go
divyapola5: Sure. I will open a new PR for new helper method.

sgmiller (approved):

PR #12655, opened 2021-09-27
Documentation update: Added tuner parameters to the Sensitive information section

taoism4504:
This request was based on [Asana](https://app.asana.com/0/inbox/1200328878163177/1201034068213937/1201041323679728). 

Updated **Audit Devices > Sensitive Information** section by adding the tuner parameters from the Secrets tune documentation.

:mag: [Deploy Preview](https://vault-jenqjatmn-hashicorp.vercel.app/docs/audit#sensitive-information)

### DIFF
--- a/website/content/docs/audit/index.mdx
+++ b/website/content/docs/audit/index.mdx
@@ -29,28 +29,21 @@ The audit logs contain the full request and response objects for every
 interaction with Vault. The request and response can be matched utilizing a
 unique identifier assigned to each request.
 
-With a few specific exceptions, all strings (including authentication tokens and lease information) contained within requests and
-responses are hashed with a salt using HMAC-SHA256. The purpose of the hash is
-so that secrets aren't in plaintext within your audit logs. However, you're
-still able to check the value of secrets by generating HMACs yourself; this can
-be done with the audit device's hash function and salt by using the
-`/sys/audit-hash` API endpoint (see the documentation for more details).
+Most strings contained within requests and responses are hashed with a salt using HMAC-SHA256. The purpose of the hash is so that secrets aren't in plaintext within your audit logs. However, you're still able to check the value of secrets by generating HMACs yourself; this can be done with the audit device's hash function and salt by using the `/sys/audit-hash` API endpoint (see the documentation for more details).
 
 Note that currently only strings coming from JSON or being returned in JSON are
 HMAC'd. Other data types, like integers, booleans, and so on, are passed
 through in plaintext.
 
-### Optional audit log flags
+Vault makes some exceptions, such as auth and secrets, and users can enable additional exceptions using the [secrets enable](https://www.vaultproject.io/docs/commands/secrets/enable) command, and then tune it afterward.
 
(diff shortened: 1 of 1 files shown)


website/content/docs/audit/index.mdx
```diff
@@ -40,6 +40,18 @@ Note that currently only strings coming from JSON or being returned in JSON are
 HMAC'd. Other data types, like integers, booleans, and so on, are passed
 through in plaintext.
 
+### Optional audit log flags
```
ncabatoff, line 43: These options do not apply to audit devices, but instead to secrets engines and auth methods, and are documented in the relevant commands.  It makes sense to alert the reader about the existence of these options, but they don't make sense in isolation, without the commands to which they apply.  How about linking to the auth_tune, auth_enable, secrets_tune, and secrets_enable pages where we document the options?

taoism4504: Thanks for the feedback, Nick. After discussing with Andrew and taking your feedback into account, the following changes have been made. Please re-review, and let me know if you have additional feedback.

apg (approved):
This covers the changes that I wanted to see incorporated (the tunability of it). Thanks!

website/content/docs/audit/index.mdx
```diff
@@ -29,17 +29,22 @@ The audit logs contain the full request and response objects for every
...
-still able to check the value of secrets by generating HMACs yourself; this can
-be done with the audit device's hash function and salt by using the
-`/sys/audit-hash` API endpoint (see the documentation for more details).
+Most strings contained within requests and responses are hashed with a salt using HMAC-SHA256. The purpose of the hash is so that secrets aren't in plaintext within your audit logs. However, you're still able to check the value of secrets by generating HMACs yourself; this can be done with the audit device's hash function and salt by using the `/sys/audit-hash` API endpoint (see the documentation for more details).
 
 Note that currently only strings coming from JSON or being returned in JSON are
 HMAC'd. Other data types, like integers, booleans, and so on, are passed
 through in plaintext.
 
+While most strings are hashed, Vault does make some exceptions, such as auth and secrets, and users can enable additional exceptions using the [secrets enable](https://www.vaultproject.io/docs/commands/secrets/enable) command, and then tune it afterward.
```
ncabatoff, line 38: Don't we usually use relative links?  I would think Vercel previews wouldn't work correctly otherwise.

ncabatoff (approved):

website/content/docs/audit/index.mdx
taoism4504: I've seen cases where the docs use both direct and relative links but I went ahead and changed all the links to relative. BTW, I checked Vercel preview and it actually does work with direct links as well, just fyi.

ncabatoff: It "works" but it doesn't link to the preview data.  Let's say you moved a page in your changes to a new place.  The link would work because it refers to vaultproject.io, but if we deployed the changes, the link would be broken, because of the move.  This would be discoverable in the preview with a relative link, that points to the preview itself.

taoism4504: Good to know, thanks!

PR #12673, opened 2021-09-29
KV: show delete when no read metadata

Monkeychip:
When you had the following policy, the delete button never showed because we compare the latest version to the current version. However, the current version comes from the metadata endpoint so it always returned false. Now I check for undefined which only happens when you have permissions errors and show the button instead.

To reproduce: 
1. In a KV 2 called secret, create any secret.

2. Create a user with policy like the following:
```path "secret/data/*" {capabilities = ["list", "read","create","update", "delete"]}
path "secret/metadata/*" {capabilities = ["list"]}```

3. Logged in as the user with the policy navigate to the show of the secret and you won't see the delete button. Though you have delete capabilities.

You can test this using the following CLI/ CURL:

LOGIN:

vault login -method=userpass username=<username> 

DELETE:

 curl --header "X-Vault-Token: root" --request DELETE http://127.0.0.1:4200/v1/secret/data/my-secret

### DIFF
--- a/ui/app/components/secret-delete-menu.js
+++ b/ui/app/components/secret-delete-menu.js
@@ -98,9 +98,15 @@ export default class SecretDeleteMenu extends Component {
   @alias('secretDataPath.canDelete') canDeleteSecretData;
 
   get isLatestVersion() {
+    // must have metadata access.
     let { model } = this.args;
     if (!model) return false;
     let latestVersion = model.currentVersion;
+    if (latestVersion === undefined) {
+      // it means you do not have metadata access (current version is returned on the metadata endpoint)
+      // if that's the case they should see the delete button because they can't navigate to older versions
+      return true;
+    }
     let selectedVersion = model.selectedVersion.version;
     if (latestVersion !== selectedVersion) {
       return false;
--- a/ui/app/templates/components/secret-create-or-update.hbs
+++ b/ui/app/templates/components/secret-create-or-update.hbs
@@ -137,15 +137,14 @@
 {{#if (eq @mode "edit")}}
   <form onsubmit={{action "createOrUpdateKey" "edit"}}>
     <div class="box is-sideless is-fullwidth is-marginless padding-top">
-      {{#if @model.canReadSecretData}}
-        <MessageError @model={{@modelForData}} @errorMessage={{this.error}} />
-      {{else}}
+      <MessageError @model={{@modelForData}} @errorMessage={{this.error}} />
+      {{#unless @model.canReadSecretData}}
         <AlertBanner
           @type="warning"
           @message="You do not have read permissions. If a secret exists here creating a new secret will overwrite it."
           data-test-warning-no-read-permissions
         />
-      {{/if}}
+      {{/unless}}
       <NamespaceReminder @mode="edit" @noun="secret" />
       {{#if this.isCreateNewVersionFromOldVersion}}
         <div class="form-section">

### CODE REVIEW

File: ui/app/components/secret-delete-menu.js [hunk 1]
```diff
@@ -129,9 +129,11 @@ export default class SecretDeleteMenu extends Component {
         this.args.navToNearestAncestor.perform(id);
       });
     } else {
+      // if they do not have read access on the metadata endpoint we need to pull the version from modelForData so they can perform delete and undelete operations
```
Monkeychip, line 132: this needs to be removed - i think.

ui/app/adapters/secret-v2-version.js
```diff
@@ -36,6 +36,13 @@ export default ApplicationAdapter.extend({
     });
   },
 
+  async getSecretDataVersion(backend, id) {
+    // used in secret-edit route when you don't have current version and you need it for pulling the correct secret-v2-version record
+    let url = this._url(backend, id);
+    let response = await this.ajax(this._url(backend, id), 'GET');
```
Monkeychip, line 42: make network request to https://www.vaultproject.io/api-docs/secret/kv/kv-v2#read-secret-version

ui/app/adapters/secret-v2-version.js
```diff
@@ -67,28 +74,57 @@ export default ApplicationAdapter.extend({
     return this._url(backend, path);
   },
 
-  v2DeleteOperation(store, id, deleteType = 'delete') {
```
Monkeychip, line 70: cleaned up delete operations. Problems with reloading because model.reload() is a promise and it wasn't fufilling.

ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
```diff
@@ -251,7 +251,20 @@ export default Route.extend(UnloadModelRoute, {
     if (modelType === 'secret-v2') {
       // after the the base model fetch, kv-v2 has a second associated
       // version model that contains the secret data
-      secretModel = await this.fetchV2Models(capabilities, secretModel, params);
+
+      // if no read access to metadata, return current Version from secret data endpoint's customMetadata.
+      if (!secretModel.currentVersion) {
+        let adapter = this.store.adapterFor('secret-v2-version');
+        try {
+          secretModel.currentVersion = await adapter.getSecretDataVersion(backend, secret);
```
Monkeychip, line 259: Need to assign currentVersion before performing fetchV2Models as it uses the currentVersion in the method.

Monkeychip: This fixes the error we were seeing when no read access to metadata and it was not fetching the most recent version of data or transition.

ui/app/templates/vault/cluster/secrets/backend/list.hbs
```diff
@@ -1,11 +1,26 @@
+{{log backendModel}}
 <SecretListHeader
   @isCertTab={{eq tab "certs"}}
   @model={{backendModel}}
   @baseKey={{baseKey}}
   @backendCrumb={{backendCrumb}}
   @filter={{filter}}
 />
-{{#if this.noMetadataPermissions}}
+{{#if (and this.noMetadataPermissions backendModel.isV2KV)}}
```
Monkeychip, line 9: This is what this does; shows the create button. 
image

ui/app/templates/vault/cluster/secrets/backend/metadata.hbs
```diff
@@ -1,7 +1,7 @@
 <PageHeader as |p|>
   <p.top>
     <KeyValueHeader
-      @baseKey={{hash id=model.id}}
+      @baseKey={{hash id=this.id}}
```
Monkeychip, line 4: id is defined on the route, which is pulled from the params
`this.id = secret;`

PR #12676, opened 2021-09-29
Fix changelogs

hashishaw:
Fix is not a valid release-note type


Monkeychip (approved):
Thank you! Sorry about the spelling mistake!

PR #12682, opened 2021-09-30
Adds ability to define an inline policy and internal metadata on tokens

austingebauer:
## Overview

This PR introduces the `InlinePolicy` and `InternalMeta` fields to the `TokenEntry` struct. These fields will be used to support using Vault batch tokens as access tokens for Vault OIDC providers.

The `InlinePolicy` field specifies ACL rules to be applied to the token entry. If an `InlinePolicy` is set, the token will not inherit policies from the [Entity](https://github.com/hashicorp/vault/blob/main/sdk/logical/token.go#L131) (or group membership) associated with the token. [Policies](https://github.com/hashicorp/vault/blob/main/sdk/logical/token.go#L88) set on the token will continue to work alongside a specified `InlinePolicy`.

The `InternalMeta` field provides the ability to associate metadata with a `TokenEntry` that is not audit logged or exposed via the Vault API.

Example of a `TokenEntry` with an `InlinePolicy` and `InternalMeta` set:

```go
logical.TokenEntry{
	Type:         logical.TokenTypeBatch,
	NamespaceID:  ns.ID,
	Path:         req.Path,
	TTL:          client.AccessTokenTTL,
	CreationTime: accessTokenIssuedAt.Unix(),
	EntityID:     entity.ID,
	InternalMeta: map[string]string{
		accessTokenClientIDMeta: client.ClientID,
		accessTokenScopesMeta:   strings.Join(authCodeEntry.scopes, scopesDelimiter),
	},
	InlinePolicy: fmt.Sprintf(`
		path "identity/oidc/provider/%s/userinfo" {
			capabilities = ["read", "update"]
		}
	`, name),
}
```

## Testing

I tested that the `InlinePolicy` works as expected by making a request to a resource (OIDC UserInfo Endpoint) that is authenticated using a batch token with an inline policy. I checked that the policies and capabilities for the token are as expected. I also checked Vault audit logs to ensure that the `InternalMeta` doesn't show up. The `EntityID` still shows up in the audit logs, as it's associated with the `TokenEntry`.

```sh
$ vault write sys/capabilities token="b.redacted.batch.token" path="identity/oidc/provider/my-provider/userinfo"
Key                                            Value
---                                            -----
capabilities                                   [read update]
identity/oidc/provider/my-provider/userinfo    [read update]

$ vault token lookup b.redacted.batch.token
Key                 Value
---                 -----
accessor            n/a
creation_time       1632978594
creation_ttl        1h
display_name        n/a
entity_id           71ae269a-4e58-5b34-7d86-4599aa39d9c6
expire_time         2021-09-29T23:09:54-07:00
explicit_max_ttl    0s
id                  b.redacted.batch.token
issue_time          2021-09-29T22:09:54-07:00
meta                <nil>
num_uses            0
orphan              true
path                oidc/provider/my-provider/token
policies            <nil>
renewable           false
ttl                 57m43s
type                batch
```


austingebauer:
Looking for feedback on approach based on our conversations around `InlinePolicy` and `InternalMeta`. Thanks, all!

vault/capabilities.go
```diff
@@ -58,14 +58,29 @@ func (c *Core) Capabilities(ctx context.Context, token, path string) ([]string,
 		policyCount += len(nsPolicies)
 	}
 
+	// Construct the corresponding ACL object. ACL construction should be
+	// performed on the token's namespace.
+	tokenCtx := namespace.ContextWithNamespace(ctx, tokenNS)
+	policies, err := c.policyStore.GetNamedPolicies(tokenCtx, policyNames)
```
calvn, line 64: It looks like whenever `InlinePolicies` are involved, we have to call `GetNamedPolicies`, `ParseACLPolicy`, followed by `ACLFromPolicies`. Is there a way to wrap/simplify this?

austingebauer: Good idea. Simplified the handling in 9af58c4.

sdk/logical/token.go
```diff
@@ -87,12 +87,19 @@ type TokenEntry struct {
...
+	// the token will not inherit policies from the associated Entity.
+	InlinePolicy string `json:"inline_policy" mapstructure:"inline_policy" structs:"inline_policy"`
+
 	// Used for audit trails, this is something like "auth/user/login"
 	Path string `json:"path" mapstructure:"path" structs:"path"`
 
 	// Used for auditing. This could include things like "source", "user", "ip"
 	Meta map[string]string `json:"meta" mapstructure:"meta" structs:"meta" sentinel:"meta"`
 
+	// InternalMeta is used to store internal metadata. This metadata will not be audit logged.
```
briankassouf, line 100:
```suggestion
	// InternalMeta is used to store internal metadata. This metadata will not be audit logged or returned from lookup APIs.
```

vault/policy_store.go
```diff
@@ -753,10 +753,11 @@ func (t *TemplateError) Error() string {
 }
 
 // ACL is used to return an ACL which is built using the
-// named policies.
-func (ps *PolicyStore) ACL(ctx context.Context, entity *identity.Entity, policyNames map[string][]string) (*ACL, error) {
-	var policies []*Policy
-	// Fetch the policies
+// named policies and pre-fetched policies if given.
+func (ps *PolicyStore) ACL(ctx context.Context, entity *identity.Entity, policyNames map[string][]string, policies ...*Policy) (*ACL, error) {
```
briankassouf, line 757: To disambiguate the last two parameters maybe the new param could be called `additionalPolicies` or `inlinePolicies` or something?

vault/request_handling.go
```diff
@@ -52,10 +52,10 @@ type HandlerProperties struct {
...
 // ID. If the entity is merged into a different entity object, the entity into
 // which the given entity ID is merged into will be returned. This function
-// also returns the cumulative list of policies that the entity is entitled to.
-// This list includes the policies from the entity itself and from all the
-// groups in which the given entity ID is a member of.
-func (c *Core) fetchEntityAndDerivedPolicies(ctx context.Context, tokenNS *namespace.Namespace, entityID string) (*identity.Entity, map[string][]string, error) {
+// also returns the cumulative list of policies that the entity is entitled to
+// if derivePolicies is set to true. This list includes the policies from the
+// entity itself and from all the groups in which the given entity ID is a member of.
+func (c *Core) fetchEntityAndDerivedPolicies(ctx context.Context, tokenNS *namespace.Namespace, entityID string, derivePolicies bool) (*identity.Entity, map[string][]string, error) {
```
briankassouf, line 58:
```suggestion
func (c *Core) fetchEntityAndDerivedPolicies(ctx context.Context, tokenNS *namespace.Namespace, entityID string, deriveEntityPolicies bool) (*identity.Entity, map[string][]string, error) {
```

briankassouf (approved):
Looks good, just a few small comments

ncabatoff:
The changes are:
1) new internal meta field
2) new inline policies field
3) inline policies -> no identity policies

The code looks fine, but I wanted to bring into question 3.  Not the ability to exclude identity policies from a token, but the implicit behaviour:

> If an InlinePolicy is set, the token will not inherit policies from the Entity (or group membership) associated with the token

That may be what makes sense for your current use case, but we could just as well replace (3) with (3a):

3a) new field controls whether identity policies are used

Then if someday we want inline policies for tokens which do need identity policies, we can have them.  There's also an outstanding feature request (which I'm not sure we want to do) for the ability to create tokens without identity policies applying - if we decided to implement that having this new field would be helpful.

vault/token_store.go
```diff
@@ -917,6 +924,8 @@ func (ts *TokenStore) create(ctx context.Context, entry *logical.TokenEntry) err
 			EntityID:     entry.EntityID,
 			NamespaceID:  entry.NamespaceID,
 			Type:         uint32(entry.Type),
+			InternalMeta: entry.InternalMeta,
+			InlinePolicy: entry.InlinePolicy,
```
briankassouf, line 928: Should the new skip identity policies field be added here?

briankassouf (approved):
Thanks for adding the new field, one thing but looks good otherwise

sdk/logical/token.go
```diff
@@ -128,8 +134,13 @@ type TokenEntry struct {
 	CreationTimeDeprecated   int64         `json:"CreationTime" mapstructure:"CreationTime" structs:"CreationTime" sentinel:""`
 	ExplicitMaxTTLDeprecated time.Duration `json:"ExplicitMaxTTL" mapstructure:"ExplicitMaxTTL" structs:"ExplicitMaxTTL" sentinel:""`
 
+	// EntityID is the ID of the entity associated with this token.
 	EntityID string `json:"entity_id" mapstructure:"entity_id" structs:"entity_id"`
 
+	// If SkipIdentityInheritance is true, the token will not
```
ncabatoff, line 140: I have a mild preference for a name that contains "policies", e.g. `NoIdentityPolicies` or something like that.

Issue #12157, opened 2021-07-23
Support Y10K value in `notAfter` field to be compliant with IEEE 802.1AR-2018 standard


skhilar:
**Is your feature request related to a problem? Please describe.**
As per the IEE 802.1AR-2018 standard, the `notAfter` field of the IDevID certificate should use the Generalized Time value 99991231235959Z. However, in the PKI secrets engine currently, the max ttl value that can be specified is [~290 years](https://pkg.go.dev/time#:~:text=A%20Duration%20represents%20the%20elapsed,duration%20to%20approximately%20290%20years).

**Describe the solution you'd like**
Support an additional `device_flag` in certificate roles, which, when set ignores the `ttl` field  and sets the `notAfter` value to Y10K. 

**Describe alternatives you've considered**
The  alternative to use the supported `max_ttl` value of 290 years, but that would not be inline with the standard.

**Explain any additional use-cases**
Medical equipment manufacturers need to embed device certificates  which can last for life  of the device at the time of production, to be able to securely interact with other devices and applications in a healthcare facility. 

**Additional context**
Please refer the standard at below link.
https://standards.ieee.org/standard/802_1AR-2018.html
I have done the changes in my branch. 
https://github.com/skhilar/vault/tree/vault_device_cert_feature

nkarstens:
I agree that this would be a useful enhancement.

skhilar:
I have done the changes in my branch. Please have a look.

https://github.com/skhilar/vault/tree/vault_device_cert_feature

hsimon-hashicorp:
@skhilar it's not easy to see what's been changed. Please feel free to create a pull request, and include the necessary tests, etc. Thanks!

PR #12683, opened 2021-09-30
Vault device cert feature

skhilar:


hsimon-hashicorp:
Hi @skhilar - can you provide more context around this PR, the linked issue that it addresses, and your implementation methodology? Thanks! :)

skhilar:
fixes #12157

As per the IEE 802.1AR-2018 standard, the notAfter field of the IDevID certificate should use the Generalized Time value 99991231235959Z. However, in the PKI secrets engine currently, the max ttl value that can be specified is ~290 years.

Support an additional device_flag in certificate roles, which, when set ignores the ttl field and sets the notAfter value to Y10K.

Issue #12601, opened 2021-09-21
Deadlock on HA leadership transfer when standby was actively forwarding a request


impl:
**Describe the bug**

A deadlock can occur between the `stateLock` and `requestForwardingConnectionLock` of the standby core when assuming HA leadership while actively forwarding a request. In particular, the following goroutines will hang indefinitely:

```
goroutine 109 [semacquire, 37 minutes]:
sync.runtime_SemacquireMutex(0xc0003be980, 0x0, 0x0)
	/goroot/src/runtime/sema.go:71 +0x47
sync.(*RWMutex).Lock(0xc0003be978)
	/goroot/src/sync/rwmutex.go:103 +0x85
github.com/hashicorp/vault/vault.standardUnsealStrategy.unseal(0x5ede4a0, 0xc002aee7c0, 0x5f3ab40, 0xc00111c030, 0xc0003be600, 0x0, 0x0)
	/gopath/src/github.com/hashicorp/vault/vault/core.go:1918 +0x5c
github.com/hashicorp/vault/vault.(*Core).postUnseal(0xc0003be600, 0x5ede4a0, 0xc002aee7c0, 0xc00042d330, 0x5e35000, 0x8c62d00, 0x0, 0x0)
	/gopath/src/github.com/hashicorp/vault/vault/core.go:2086 +0x336
github.com/hashicorp/vault/vault.(*Core).waitForLeadership(0xc0003be600, 0x0, 0xc00106afc0, 0xc00106b140)
	/gopath/src/github.com/hashicorp/vault/vault/ha.go:565 +0xc2e
github.com/hashicorp/vault/vault.(*Core).runStandby.func7(0x0, 0x0)
	/gopath/src/github.com/hashicorp/vault/vault/ha.go:387 +0x45
github.com/hashicorp/vault/vendor/github.com/oklog/run.(*Group).Run.func1(0xc00107d380, 0xc00111db60, 0xc00113e9a0)
	/gopath/src/github.com/hashicorp/vault/vendor/github.com/oklog/run/group.go:38 +0x27
created by github.com/hashicorp/vault/vendor/github.com/oklog/run.(*Group).Run
	/gopath/src/github.com/hashicorp/vault/vendor/github.com/oklog/run/group.go:37 +0xbb

goroutine 12292904 [semacquire, 37 minutes]:
sync.runtime_SemacquireMutex(0xc0003be6fc, 0xc00816d800, 0x0)
[...]

PR #12691, opened 2021-09-30
Fix a Deadlock on HA leadership transfer

hghaf099:
Fix a Deadlock on HA leadership transfer when standby was actively forwarding a request
fixes GH #12601

### DIFF
--- a/changelog/12691.txt
+++ b/changelog/12691.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Fix a deadlock on HA leadership transfer
+```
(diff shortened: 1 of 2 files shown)


ncabatoff (approved):

write0nly:
@hghaf099 and @ncabatoff although this fix seems only to apply to perf standbys which are not included in OSS vault, can you please clarify if this fix applies to the OSS users too? It's unclear if the code path that leads to the deadlock applies to OSS.

Thanks.

ncabatoff:
Yes, it applies to OSS.  The deadlock involves access to a mutex-protected field named perfStandby, but the field is accessed even by regular standbys.

PR #12692, opened 2021-09-30
Highlight that password policies are defined in a namespace

kalafut:
This was implied but could be clearer.

### DIFF
--- a/website/content/docs/enterprise/namespaces.mdx
+++ b/website/content/docs/enterprise/namespaces.mdx
@@ -82,7 +82,8 @@ namespace. This data includes the following:
 
 - Secret Engines
 - Auth Methods
-- Policies (including Password Policies)
+- Policies
+- Password Policies
 - Identities (Entities, Groups)
 - Tokens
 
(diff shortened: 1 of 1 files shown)


website/content/docs/enterprise/namespaces.mdx
```diff
@@ -82,7 +82,7 @@ namespace. This data includes the following:
 
 - Secret Engines
 - Auth Methods
-- Policies
+- Policies (including Password Policies)
```
pcman312, line 85: Can this be a separate line? Otherwise this could imply a relationship between policies and password policies when none exists

calvn, line 85: I wonder if it's better to break these into separate and explicit entries
```suggestion
- ACL, EGP, and RGP Policies 
- Password Policies 
```

kalafut: Yep

kalafut: Yep

calvn (approved):

pcman312 (approved):

PR #12700, opened 2021-10-01
Update Ubuntu Image used in CircleCI to `ubuntu-2004:202010-01`

vinay-gopalan:
This PR updates the Ubuntu machine image used in `build-common-layers` within CircleCI from version `ubuntu-1604:202007-01` to `ubuntu-2004:202010-01`


mladlow:
From further conversation, it seems like we might need to make this change in packagespec because of how some of the circle configs are autogenerated.

PR #12701, opened 2021-10-01
raft auto_join_scheme documentation update

rowansmithhc:
An update to https://www.vaultproject.io/docs/configuration/storage/raft to detail the available configuration values for the `auto_join_scheme` configuration parameter as these don't appear to be documented anywhere.

Reference: https://github.com/hashicorp/vault/blob/861454e0ed1390d67ddaf1a53c1798e5e291728c/http/sys_raft.go#L70-L72


ncabatoff:
Thanks!

PR #12709, opened 2021-10-01
Update builder base image

samsalisbury:


HridoyRoy (commented):
Looks good to me, but I'm relatively new to our CI files so it might be helpful to get another set of eyes on this as well. 

Is it ok for our packages to use the bullseye base image and our go tests to rely on the older buster image?

pmmukh (approved):
The config file changes look good to me 👍 However, as Roy points out, we might want to upgrade the docker image versions in our tests, in this or a separate pr.

ncabatoff (approved):

PR #12710, opened 2021-10-01
fix: hide intended elements in print

zchsh:


website/pages/print.css
```diff
@@ -1,240 +1,240 @@
...
-    [id='edit-this-page'],
-    [id='jump-to-section'],
-    [id='__next-build-watcher'] {
-      display: none !important;
-    }
+  iframe,
+  #__next > header,
+  .g-product-subnav,
+  .g-footer,
+  .g-consent-manager,
```
zchsh, line 7: apart from these changes, everything else in this file is just formatting.

Issue #12282, opened 2021-08-08
Remove 'Unpublished: true' inconsistently applied to sys/internal/* endpoints


maxb:
**Is your feature request related to a problem? Please describe.**

I am writing a tool to consume Vault's automatic OpenAPI (sys/internal/specs/openapi) for validating policies.

The Vault OpenAPI lists **almost but not all** of the operations its endpoints support. For some reason, 5 of the sys/internal/* APIs are marked 'Unpublished: true' in the code (https://github.com/hashicorp/vault/blob/main/vault/logical_system_paths.go) concealing their operation data from OpenAPI:

    sys/internal/ui/namespaces
    sys/internal/ui/resultant-acl
    sys/internal/counters/{requests,tokens,entities}

It's not just because they are internal, as other sys/internal/ endpoints are published.

Some of the sys/internal/ APIs marked as Unpublished in the code, are even part of the public documentation: https://www.vaultproject.io/api-docs/system/internal-counters

Having them in the public documentation but not the built-in Vault API Explorer (Swagger) is rather odd.

**Describe the solution you'd like**

Please delete the 5 occurrences of "Unpublished: true" from https://github.com/hashicorp/vault/blob/main/vault/logical_system_paths.go so that programs consuming the Vault OpenAPI - including but not limited to the built in Vault Swagger - can have full knowledge of the API.

**Describe alternatives you've considered**

For my own tool, I'm hardcoding extra overrides to account for the unpublished APIs - but this is inelegant, and subject to becoming out of date. For the omission of the APIs from the built-in API Explorer (Swagger), there is no workaround.

pmmukh:
Hi @maxb ! Thank you for submitting this issue! After some discussion, the rationale for not having certain endpoints in the openApi spec was due to the fact that we don't guarantee backwards compatibility on those endpoints, but we've decided that should not preclude those endpoints from being in the openApi spec, we plan on having the lack of backwards compatibility explicitly called out in the spec descriptions to solve this problem instead. So the plan currently is to move forward with this, and publish all 5 endpoints in the spec. I'm going to start work on this shortly, will keep you posted when its fixed!

pmmukh:
Sorry for the confusion, I don't think I'll actually get a chance to do this soon, but this should be a good issue for new community members, so I'll add the first issue label and see if someone picks it up.

maxb:
[...]

PR #12713, opened 2021-10-02
Fixes https://github.com/hashicorp/vault/issues/12282 Removed unpublished:true for sys/internal/* endpoints

saltperfect:
Fixes: https://github.com/hashicorp/vault/issues/12282

Removed Unpublished true to make open api spec consistent with documentation.

Useful Links
1. https://www.vaultproject.io/api-docs/system/internal-counters
2. https://github.com/hashicorp/vault/blob/main/vault/logical_system_paths.go

### DIFF
--- a/vault/logical_system_paths.go
+++ b/vault/logical_system_paths.go
@@ -928,7 +928,6 @@ func (b *SystemBackend) internalPaths() []*framework.Path {
 			Operations: map[logical.Operation]framework.OperationHandler{
 				logical.ReadOperation: &framework.PathOperation{
 					Callback:    pathInternalUINamespacesRead(b),
-					Unpublished: true,
 				},
 			},
 			HelpSynopsis:    strings.TrimSpace(sysHelp["internal-ui-namespaces"][0]),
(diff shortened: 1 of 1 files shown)


hghaf099:
@saltperfect Thanks for submitting this PR. Would you please add the changelog text file related to this change as well? Also, if you would please add some documentation on those unpublished endpoints, that would be great. We don't need thorough docs on those, just something to acknowledge their existence would suffice for now. With those changes, we can go ahead and merge this PR.

changelog/_12713.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+COMPONENT: removed unpublished true from logical_system path, making openapi spec consistent with documentation
```
hghaf099, line 2:
```suggestion
http: removed unpublished true from logical_system path, making openapi spec consistent with documentation
```
```suggestion
COMPONENT: removed unpublished true from logical_system path, making openapi spec consistent with documentation
```

hghaf099, line 2: the name of the file contains an "_", please remove it.

saltperfect:
I have edited 5 entries in logical_paths_system.go 

`internal/ui/namespaces`, `internal/ui/resultant-acl`, `internal/counters/requests`, `internal/counters/tokens`, `internal/counters/entities`. 

For now I have put in a really dumb entry for Summary. Let me know if this works.

hghaf099:
@saltperfect Thank you very much for adding the docs. Seems like the deployment check failed since the added new endpoints mdx files, are not reflected in the api docs nav data file located here:
https://github.com/hashicorp/vault/blob/main/website/data/api-docs-nav-data.json
Would you please update that file as well with the new entries?

Issue #12214, opened 2021-07-30
AWS Auth Backend rotate-root not working


danieldethloff1993:
**Describe the bug**
AWS Auth Backend rotate-root does not work, when previously configured the backend with static access_key and secret_key, the aws credentails were rotated but the new access_key and secret_key were not injected in the aws auth backend config.

**To Reproduce**
1. Run `vault auth enable aws`
2. Run `vault write /auth/aws/config/client access_key=XXXXXXXXXXXXX secret_key=yyyyyyyyyyyyyyyy`
3. Run `vault read auth/aws/config/client` and notice the previously added access_key matches
4. Run `vault write -force /auth/aws/config/rotate-root` and take a look to the access_key in the output
5. Run `vault read auth/aws/config/client` and notice the access_key from rotate-root output doesn't match

**Expected behavior**
The Output from `vault write -force /auth/aws/config/rotate-root` and the access_key from `vault read auth/aws/config/client` after the rotation should match.

**Environment:**
* Vault Server Version 1.7.2 (also tested 1.8.0)
* Vault CLI Version 1.7.2 (also tested 1.8.0)
* Server Operating System/Architecture: CentOS  7.9.2009
* Client CLI OS: Ubuntu 18.04. Windows WSL

Vault server configuration file(s):

```json
{
  "backend":
    {
     "consul":
       {
        "address":"<some-ip>:8500",
        "advertise_addr":"http://<server-ip>:8200",
        "path":"vault/"
       }
    },
  "listener":
    {
     "tcp":
      {
        "address":"0.0.0.0:8200",
        "tls_cert_file":"/vault/file/<cert.file>",
        "tls_key_file": "/vault/file/<key.file>",
      }
    },
    "ui": true,
    "telemetry": {
        "statsd_address": "<statsd_address>:<statsd-port>"
    }
}

```

danieldethloff1993:
Does nobody have an idea?

soeirosantos:
+1 I'm having the same problem here and the odd behavior is easily reproducible.

hsimon-hashicorp:
Hi folks! I've raised this with our engineering team, please hang tight while we review it. Appreciate your patience. :)

Zlaticanin:
Fixed with [#12715 ](https://github.com/hashicorp/vault/pull/12715).

PR #12715, opened 2021-10-04
Fix auth/aws so that config/rotate-root saves new key pair to vault

ludewigh:
This PR fixes an issue where when calling the endpoint config/rotate-root on auth/aws it would successfully rotate the keys in AWS but not save the updated keys back into vault.  After being called once you could no longer authenticate AWS services since the new key was unknown to Vault.  The issue was originally reported in #12214

### DIFF
--- a/builtin/credential/aws/path_config_rotate_root.go
+++ b/builtin/credential/aws/path_config_rotate_root.go
@@ -145,6 +145,10 @@ func (b *backend) pathConfigRotateRootUpdate(ctx context.Context, req *logical.R
 		}
 	}()
 
+	oldAccessKey := clientConf.AccessKey
+	clientConf.AccessKey = *createAccessKeyRes.AccessKey.AccessKeyId
+	clientConf.SecretKey = *createAccessKeyRes.AccessKey.SecretAccessKey
+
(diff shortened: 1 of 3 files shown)


kalafut:
@ludewigh Can you please sign the CLA? (Note that the email in an commits needs to match an email on your github account)

ludewigh:
@kalafut I was hoping since my Org already had a signed CLA we would be fine, but since I need to sign as well it will most likely take another day or so to get approval.  I've started our internal process and will sign as soon as I have been approved.

hsimon-hashicorp:
@ludewigh if your organization has signed a CLA, and your GitHub account is associated with that organization and it's publicly viewable that you're associated, the CLA bot should do the right thing. Please let us know if there's any hiccups. Thanks!

ludewigh:
@hsimon-hashicorp I think the publicly viewable part was the issue, I've updated it from private to public so hopefully when it runs next that fixes it.

ludewigh:
@hsimon-hashicorp @kalafut I don't think I ever got the corporate one to work, but I've been authorized to sign the CLA and I believe I have done so.  If there is still something missing from my side please let me know.

kalafut:
@ludewigh CLA looks good now. Thanks!

Issue #12336, opened 2021-08-17
Vault 1.8.1 acl templating breaks with JWT


evilmog:
**Describe the bug**

**To Reproduce**
Configure vault with JWT and then template under PKI or SSH Certificates the following allowed domain template:
```
{{identity.entity.aliases.auth_jwt_31588259.metadata.name}}
```

This matches the documentation https://learn.hashicorp.com/tutorials/vault/policy-templating

the vault ssh config is something similar to one configured by this python script:
```
import hvac
import os
import json
import requests
import time
from getpass import getpass

vault_base = "redacted"

shares = 5
threshold = 3

# Access Levels to Create
access_levels = [ "level1", "level2", "level3"]
# this is the prefix for groups to be created
group_prefix = "redacted_"
ssh_role_prefix = "ssh-jwt-"
access_policies = {}

#JWT Config Info
bound_issuer="redacted"
jwks_url="redacted"
oidc_response_mode = "form_post"
user_claim = "emailAddress"
groups_claim = "blueGroups"
preferred_username = "uid"
emailAddress = "name"
allowed_redirect_uris = []
claim_mappings = { }
claim_mappings["emailAddress"] ="name"
claim_mappings["preferred_username"] ="uid"

# OIDC Client ID from onboarding
bound_audiences = ["redacted"]

# SSH Config Into - TTL
ssh_ttl = 3600
ssh_token_period = 10

# PKI Config
# set the config options
# Root Tuning
root_pki_tune_options_dict = {}
root_pki_tune_options_dict['max_lease_ttl'] = "87601h"
root_pki_tune_options_dict['default_lease_ttl'] = "87600h"
# Intermediate Tuning
intermediate_pki_tune_options_dict = {}
intermediate_pki_tune_options_dict['max_lease_ttl'] = "43849h"
intermediate_pki_tune_options_dict['default_lease_ttl'] = "43848h"

# Root CA Config
pki_root_name = "pki_root"
root_ca_params = {}
root_ca_params['ttl'] = "87600h"
root_ca_params['max_lease_ttl'] = "87601h"
root_ca_params['cn'] = "Redacted"
root_ca_params['key_type'] = "rsa"
root_ca_params['key_bits'] = "4096"
root_ca_params['max_path_length'] = 3
root_ca_params['type'] = "internal"
root_ca_params['OU'] = "Redacted" 
root_ca_params['Organization'] = "Redacted"
root_ca_params['Country'] = "US" 
[...]

PR #12716, opened 2021-10-04
Fix 1.8 regression preventing email addresses being used as common name within pki certificates (#12336)

stevendpclark:
This PR addresses a regression introduced within 1.8 that prevents email addresses from being leveraged as common names within pki certificates. It still enforces that the email address was listed as an allowed domain within the role as well as allow_bare_domains set to true. 

Added two test cases that validate the behaviour in the bug report 
- common name as an email address with a matching email address within allowed domain in the role
- common name as an email address with a matching allowed domain name within the allowed domain in the role. 

Addresses issue reported within #12336


sgmiller (approved):
Looks good to me!

cipherboy (approved):

victorr (approved):
👍

PR #12718, opened 2021-10-04
agent: tolerate partial restore failure from persistent cache

tomhjp:


changelog/12688.txt
```diff
@@ -1,4 +1,3 @@
-+```release-note:improvement
-+auth/kubernetes: validate JWT against the provided role on alias look ahead operations
-+```
```
tomhjp, line 3: Fixing a small syntax error in the previous changelog

command/agent/cache/lease_cache_test.go
```diff
@@ -726,59 +732,81 @@ func TestLeaseCache_PersistAndRestore(t *testing.T) {
...
-	for _, ct := range cacheTests {
+	var deleteID string
+	for i, ct := range cacheTests {
 		// Send once to cache
 		sendReq := &SendRequest{
 			Token:   ct.token,
 			Request: httptest.NewRequest(ct.method, ct.urlPath, strings.NewReader(ct.body)),
 		}
+		if ct.deleteFromPersistentStore {
+			deleteID, err = computeIndexID(sendReq)
```
benashz, line 802: I wonder if we check to see if we hare already set `deleteID` since it is possible to have more than one cache test with `deleteFromPersistentStore` set to true.

We may also  want to ensure that `deleteID` is not an empty string.

command/agent/cache/lease_cache.go
```diff
@@ -1031,14 +1044,17 @@ func (c *LeaseCache) restoreLeases(leases [][]byte) error {
...
 		}
 		if err := c.db.Set(newIndex); err != nil {
-			return err
+			errors = multierror.Append(errors, err)
+			continue
 		}
 		c.logger.Trace("restored lease", "id", newIndex.ID, "path", newIndex.RequestPath)
 	}
-	return nil
+
```
benashz, line 1056: nit: extra newline seems to differ from the style of the other functions in this change

command/agent/cache/lease_cache_test.go
```diff
@@ -711,13 +712,18 @@ func setupBoltStorage(t *testing.T) (tempCacheDir string, boltStorage *cachebolt
 }
 
 func TestLeaseCache_PersistAndRestore(t *testing.T) {
-	// Emulate 4 responses from the api proxy. The first two use the auto-auth
-	// token, and the last two use another token.
+	// Emulate responses from the api proxy. The first two use the auto-auth
+	// token, and the others use another token.
```
benashz, line 716: Since we are relying on `[]cacheTests` and `[]responses` having equal lengths it might be good to document that here.

command/agent/cache/lease_cache.go
```diff
@@ -969,56 +970,68 @@ func (c *LeaseCache) Flush() error {
 // tokens first, since restoring a lease's renewal context and watcher requires
 // looking up the token in the cachememdb.
 func (c *LeaseCache) Restore(ctx context.Context, storage *cacheboltdb.BoltStorage) error {
+	var errors *multierror.Error
```
imthaghost, line 973: Could be worth looking into down the road for how we define custom errors.
https://github.com/uber-go/guide/blob/master/style.md#error-types

imthaghost (approved):

calvn (commented):
One meta-comment that I want to leave here is that this could make it trickier for us to implement VAULT-1096 (ordered restores) since we could get into a state where the dependency graph is incomplete, say if we failed to restore a parent. For instance, what happens to the lifecycle of its child tokens?

command/agent/cache/lease_cache_test.go
```diff
@@ -726,59 +732,81 @@ func TestLeaseCache_PersistAndRestore(t *testing.T) {
 	lc := testNewLeaseCacheWithPersistence(t, responses, boltStorage)
 
 	// Register an auto-auth token so that the token and lease requests are cached
-	lc.RegisterAutoAuthToken("autoauthtoken")
+	err := lc.RegisterAutoAuthToken("autoauthtoken")
+	assert.NoError(t, err)
```
tvoran, line 736: We may want to use `require` here so the test doesn't progress past here if there's an error.

command/agent/cache/lease_cache_test.go
```diff
@@ -789,24 +817,33 @@ func TestLeaseCache_PersistAndRestore(t *testing.T) {
 		}
 		respCached, err := lc.Send(context.Background(), sendCacheReq)
 		require.NoError(t, err, "failed to send request %+v", ct)
-		assert.Equal(t, respCached.Response.StatusCode, ct.wantStatusCode, "expected proxied response")
+		assert.Equal(t, responses[i].Response.StatusCode, respCached.Response.StatusCode, "expected proxied response")
 		require.NotNil(t, respCached.CacheMeta)
 		assert.True(t, respCached.CacheMeta.Hit)
 	}
 
+	err = boltStorage.Delete(deleteID)
```
tvoran, line 825: I wonder if this should be a loop over a list of deleteIDs? Perhaps built up in the cacheTests loop?

PR #12720, opened 2021-10-04
CLI request when namespace is in argument and part of the path

### LINKED ISSUE #12675: Vault CLI makes request to incorrect URL when namespace is both provided as argument and part of the path
**Describe the bug**
[Namespace documentation](https://www.vaultproject.io/docs/enterprise/namespaces#usage) states that namespace can be treated as part of the path. This works fine with HTTP API, but CLI makes request to incorrect URL when namespace is both provided as argument and part of the path.

**To Reproduce**
```
# Setup
vault namespace create dev
vault namespace create -namespace=dev myapp
vault secrets enable -version=2 -namespace=/dev/myapp -path=secrets kv
vault kv put -namespace=/dev/myapp secrets/examplePath foo=APPsecret

# Check that HTTP API works; all below requests return the same result
export VAULT_ADDR=...
export VAULT_TOKEN=...
curl -H "X-Vault-Token: $VAULT_TOKEN" -H "X-Vault-Namespace: " -X GET $VAULT_ADDR/v1/dev/myapp/secrets/data/examplePath
curl -H "X-Vault-Token: $VAULT_TOKEN" -H "X-Vault-Namespace: dev" -X GET $VAULT_ADDR/v1/myapp/secrets/data/examplePath
curl -H "X-Vault-Token: $VAULT_TOKEN" -H "X-Vault-Namespace: dev/myapp" -X GET $VAULT_ADDR/v1/secrets/data/examplePath

# Check CLI
vault kv get -field foo -namespace= dev/myapp/secrets/examplePath  # APPsecret
vault kv get -field foo -namespace=/dev myapp/secrets/examplePath  # No value found at dev/myapp/secrets/data/myapp/secrets/examplePath
vault kv get -field foo -namespace=/dev/myapp secrets/examplePath  # APPsecret
```

**Expected behavior**
Command `vault kv get -field foo -namespace=/dev myapp/secrets/examplePath` should work and return the secret. That is, CLI should call `dev/myapp/secrets/data/examplePath` instead of `dev/myapp/secrets/data/myapp/secrets/examplePath`.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.8.2+ent
* Vault CLI Version (retrieve with `vault version`): v1.8.2+ent (bf22e1eb262e59f08bb8a1374dc726ab93830178)
* Server Operating System/Architecture: Windows 10 / 5.4.72-microsoft-standard-WSL2

```hcl
(empty)
```

**Additional context**
n/a

hghaf099:
CLI makes request to incorrect URL when namespace is both provided as argument and part of the path
fixes #12675

### DIFF
--- a/command/kv_helpers.go
+++ b/command/kv_helpers.go
@@ -110,8 +110,24 @@ func addPrefixToVKVPath(p, mountPath, apiPrefix string) string {
 	case p == mountPath, p == strings.TrimSuffix(mountPath, "/"):
 		return path.Join(mountPath, apiPrefix)
 	default:
-		p = strings.TrimPrefix(p, mountPath)
-		return path.Join(mountPath, apiPrefix, p)
+		tp := strings.TrimPrefix(p, mountPath)
+		for {
+			// If the entire mountPath is included in the path, we are done
+			if tp != p {
+				break
+			}
+			// Trim the parts of the mountPath that are not included in the
+			// path, for example, in cases where the mountPath contains
+			// namespaces which are not included in the path.
+			partialMountPath := strings.SplitN(mountPath, "/", 2)
+			if len(partialMountPath) == 1 || partialMountPath[1] == ""{
+				break
+			}
+			mountPath = partialMountPath[1]
+			tp = strings.TrimPrefix(p, mountPath)
+		}
+
+		return path.Join(mountPath, apiPrefix, tp)
 	}
 }
 


command/kv_helpers.go
```diff
@@ -110,8 +110,24 @@ func addPrefixToVKVPath(p, mountPath, apiPrefix string) string {
...
+		for {
+			// If the entire mountPath is included in the path, we are done
+			if tp != p {
+				break
+			}
+			// Trim the parts of the mountPath that are not included in the
+			// path, for example, in cases where the mountPath contains
+			// namespaces which are not included in the path.
+			partialMountPath := strings.SplitN(mountPath, "/", 2)
+			if len(partialMountPath) == 1 || partialMountPath[1] == ""{
```
ccapurso, line 123: We probably should guard against the length prior to checking `partialMountPath[1]`. This could likely experience an index out of bounds error.

command/kv_helpers.go
```diff
@@ -110,8 +110,24 @@ func addPrefixToVKVPath(p, mountPath, apiPrefix string) string {
 	case p == mountPath, p == strings.TrimSuffix(mountPath, "/"):
 		return path.Join(mountPath, apiPrefix)
 	default:
-		p = strings.TrimPrefix(p, mountPath)
-		return path.Join(mountPath, apiPrefix, p)
+		tp := strings.TrimPrefix(p, mountPath)
+		for {
```
ccapurso, line 114: I know that this was already `switch` statement but it might be easier to read if it were refactored to be a single if where we check for the following with the same early return: 

```
if p == mountPath || p == strings.TrimSuffix(mountPath, "/")
```

the following loop using the trimmed prefix can then follow at the same indentation level as the `if` statement.

ccapurso (changes requested):
I left a couple comments that would improve the readability and error handling. It looks good at a high level but we should add some tests to validate that the bug no longer exists. Some test scenarios might provide a little extra context for me to analyze the logic a little further as well.

hghaf099:
Yes, I am creating some tests, however, the tests are going to be in the enterprise side, as namespaces are required to fully examine the change.

command/kv_helpers.go
ccapurso: In the case that the length of `partialMountPath` is 0, the check for `partialMountPath[1] == ""` will still be performed and thus result in an index out of bounds error. Here is an [example in action](https://play.golang.org/p/-zt_ocPgDpA).

hghaf099: Ah, It should be lessEqual than 1. Thank you!

PR #12731, opened 2021-10-04
[VAULT-3157] Move `mergeStates` utils from Agent to api module

vinay-gopalan:
This PR moves the utils `mergeStates`, `compareStates` from `api_proxy.go` in Agent to `client.go` in the API module. In addition, `ParseRequiredStates` from the `hashicorp/vault/vault` package is also moved to `client.go`. The `ParseRequiredStates` and `mergeStates` methods are also made public.

These methods need to be accessed by the Terraform Vault Provider in order to provide consistent X-Vault-Index header support.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -1152,7 +1152,7 @@ func RequireState(states ...string) RequestCallback {
 }
 
 // CompareStates returns 1 if s1 is newer or identical, -1 if s1 is older, and 0
-// if neither s1 or s2 is strictly greater.  An error is returned if s1 or s2
+// if neither s1 or s2 is strictly greater. An error is returned if s1 or s2
 // are invalid or from different clusters.
 func CompareStates(s1, s2 string) (int, error) {
 	w1, err := ParseRequiredState(s1, nil)
--- a/api/client.go
+++ b/api/client.go
@@ -2,7 +2,11 @@ package api
 
 import (
 	"context"
+	"crypto/hmac"
+	"crypto/sha256"
 	"crypto/tls"
+	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"net"
 	"net/http"
@@ -21,6 +25,8 @@ import (
 	rootcerts "github.com/hashicorp/go-rootcerts"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
 	"github.com/hashicorp/vault/sdk/helper/consts"
+	"github.com/hashicorp/vault/sdk/helper/strutil"
+	"github.com/hashicorp/vault/sdk/logical"
 	"golang.org/x/net/http2"
 	"golang.org/x/time/rate"
 )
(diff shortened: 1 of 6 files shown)


api/client.go
```diff
@@ -1145,6 +1151,103 @@ func RequireState(states ...string) RequestCallback {
 	}
 }
 
+// CompareStates returns 1 if s1 is newer or identical, -1 if s1 is older, and 0
+// if neither s1 or s2 is strictly greater. An error is returned if s1 or s2
+// are invalid or from different clusters.
+func CompareStates(s1, s2 string) (int, error) {
```
tomhjp, line 1157:
```suggestion
func CompareReplicationStates(s1, s2 string) (int, error) {
```

States seems quite generic to me, and for a public method in our `api` package, won't give developers much intuition for what these methods are to do with unless they're already very familiar with Vault's internals. Is it worth making the method names a little more specific to help? Would love a second opinion on the suggestion though.

jasonodonnell: I agree with Tom, and if you accept this please don't forget to change the documentation comment above it.

api/client.go
```diff
@@ -1145,6 +1151,103 @@ func RequireState(states ...string) RequestCallback {
...
+	if err != nil {
+		return 0, err
+	}
+	w2, err := ParseRequiredState(s2, nil)
+	if err != nil {
+		return 0, err
+	}
+
+	if w1.ClusterID != w2.ClusterID {
+		return 0, fmt.Errorf("don't know how to compare states with different ClusterIDs")
```
jasonodonnell, line 1168:
```suggestion
		return 0, fmt.Errorf("can't compare replication states with different ClusterIDs")
```

jasonodonnell (approved):
LGTM!

api/client.go
vinay-gopalan: Thanks for the suggestion! Updated the method name

api/client.go
vinay-gopalan: Updated, thanks!

api/client.go
```diff
@@ -1145,6 +1151,103 @@ func RequireState(states ...string) RequestCallback {
...
+	// We've already handled the case where both are equal above, so really we're
+	// asking here if one or both are lesser.
+	case w1.LocalIndex <= w2.LocalIndex && w1.ReplicatedIndex <= w2.ReplicatedIndex:
+		return -1, nil
+	}
+
+	return 0, nil
+}
+
+func MergeStates(old []string, new string) []string {
```
tomhjp, line 1183:
```suggestion
func MergeReplicationStates(old []string, new string) []string {
```

Please could you add a function documentation comment as well, similar to CompareReplicationStates?

api/client.go
```diff
@@ -1145,6 +1151,103 @@ func RequireState(states ...string) RequestCallback {
...
+		case -1:
+			ret = append(ret, new)
+		case 0:
+			ret = append(ret, o, new)
+		}
+	}
+	return strutil.RemoveDuplicates(ret, false)
+}
+
+func ParseRequiredState(raw string, hmacKey []byte) (*logical.WALState, error) {
```
tomhjp, line 1206:
```suggestion
func ParseReplicationState(raw string, hmacKey []byte) (*logical.WALState, error) {
```

Also, does this one need to be public? I don't see it or `MergeStates` used in your TF provider PR.

vinay-gopalan: Ah this was public when it was in Core and I didn't give a second deeper thought. We don't need `ParseReplicationState` to be public, but `MergeStates` does need to be since it is being used here https://github.com/hashicorp/terraform-provider-vault/pull/1170/files#diff-7e69de666c9006fa0a0fb8fe4e4f8c418acce7c555f69f8087d88d16b434f2c1R714

tomhjp: Oh good point, I should have said `CompareReplicationStates` instead of `MergeStates`

api/client.go
vinay-gopalan: Updated

api/client.go
vinay-gopalan: Updated and made necessary methods private

api/client.go
```diff
@@ -1145,6 +1151,106 @@ func RequireState(states ...string) RequestCallback {
...
+		}
+	}
+
+	pieces := strings.Split(state, ":")
+	if len(pieces) != 4 || pieces[0] != "v1" || pieces[1] == "" {
+		return nil, fmt.Errorf("invalid state header format")
+	}
+	localIndex, err := strconv.ParseUint(pieces[2], 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("invalid state header format")
```
calvn, line 1240:
```suggestion
		return nil, fmt.Errorf("invalid local index in state header: %w", err)
```

api/client.go
```diff
@@ -1145,6 +1151,106 @@ func RequireState(states ...string) RequestCallback {
...
+	}
+	replicatedIndex, err := strconv.ParseUint(pieces[3], 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("invalid state header format")
```
calvn, line 1244:
```suggestion
		return nil, fmt.Errorf("invalid replicated index in state header: %w", err)
```

api/client.go
```diff
@@ -1145,6 +1151,106 @@ func RequireState(states ...string) RequestCallback {
...
+func parseReplicationState(raw string, hmacKey []byte) (*logical.WALState, error) {
+	cooked, err := base64.StdEncoding.DecodeString(raw)
+	if err != nil {
+		return nil, err
+	}
+	s := string(cooked)
+
+	lastIndex := strings.LastIndexByte(s, ':')
+	if lastIndex == -1 {
+		return nil, fmt.Errorf("invalid state header format")
```
calvn, line 1218: Not sure what would be the proper message here, but it'd be good to have it be different than the message on L1236

Maybe?
```suggestion
		return nil, fmt.Errorf("invalid full state header format")
```

calvn (approved):
Few minor nits, but otherwise looks good!

api/client.go
vinay-gopalan: updated

api/client.go
vinay-gopalan: updated

api/client.go
vinay-gopalan: updated

ncabatoff:
Better make sure this works on Enterprise too.  ParseRequiredState is used in some ent-specific files.

vinay-gopalan:
> Better make sure this works on Enterprise too. ParseRequiredState is used in some ent-specific files.

@ncabatoff the last spot that needs to be updated in Vault Enterprise is here in `core_util_ent.go`: https://github.com/hashicorp/vault-enterprise/blob/main/vault/core_util_ent.go#L1271

I'll open a PR on the `vault-enterprise` side updating that line once this work is merged in and available in the enterprise repo. Does that sound reasonable?

ncabatoff:
> > Better make sure this works on Enterprise too. ParseRequiredState is used in some ent-specific files.
> 
> @ncabatoff the last spot that needs to be updated in Vault Enterprise is here in `core_util_ent.go`: https://github.com/hashicorp/vault-enterprise/blob/main/vault/core_util_ent.go#L1271
> 
> I'll open a PR on the `vault-enterprise` side by importing the `api` package in that file and updating that line once this PR is merged in and the changes available in the enterprise repo. Does that sound reasonable?

Yup, that's the usual practice.  Just don't want you to be caught unawares and scrambling to fix the build.

api/client.go
```diff
@@ -1145,6 +1151,106 @@ func RequireState(states ...string) RequestCallback {
...
+		case -1:
+			ret = append(ret, new)
+		case 0:
+			ret = append(ret, o, new)
+		}
+	}
+	return strutil.RemoveDuplicates(ret, false)
+}
+
+func parseReplicationState(raw string, hmacKey []byte) (*logical.WALState, error) {
```
ncabatoff, line 1209: This needs to be re-exported so it's available to vault ent.

PR #12736, opened 2021-10-05
build: update base image: debian:bullseye-20210927 (release/1.8.x)

samsalisbury:
Manual backport of #12709


mladlow:
Partly a note to self: I'd like to see build-common-layers succeed on this PR, but I'm ok with the test-go-remote-docker failure because it passed on the last 1.8 commit and is definitely a test that's flaky sometimes.

PR #12747, opened 2021-10-05
[VAULT-3252] Disallow alias creation if entity/accessor combination exists

pmmukh:
- This PR disallows entity alias creation if an alias for the specified entity and mount already exists, along with a test that checks the same
- It also logs a warning when entities are loaded on Vault startup, to tackle the case of such existing duplicates, in order to let the operator know this setup should be addressed ( should the log be a bit clearer about that ? )

### DIFF
--- a/vault/identity_store_aliases.go
+++ b/vault/identity_store_aliases.go
@@ -267,6 +267,12 @@ func (i *IdentityStore) handleAliasCreate(ctx context.Context, req *logical.Requ
 		}
 	}
 
+	for _, alias := range entity.Aliases {
+		if alias.MountAccessor == mountAccessor {
+			return logical.ErrorResponse("Alias already exists for requested entity and mount accessor"), nil
+		}
+	}
+
 	entity.Aliases = append(entity.Aliases, alias)
 
 	// ID creation and other validations; This is more useful for new entities
--- a/vault/identity_store_aliases_test.go
+++ b/vault/identity_store_aliases_test.go
@@ -405,6 +405,56 @@ func TestIdentityStore_AliasUpdate(t *testing.T) {
 	}
 }
 
+// Test that alias creation fails if an alias for the specified mount
+// and entity has already been created
+func TestIdentityStore_AliasCreate_DuplicateAccessor(t *testing.T) {
+	var err error
+	var resp *logical.Response
+	ctx := namespace.RootContext(nil)
+	is, githubAccessor, _ := testIdentityStoreWithGithubAuth(ctx, t)
+
+	resp, err = is.HandleRequest(ctx, &logical.Request{
+		Path:      "entity",
+		Operation: logical.UpdateOperation,
+	})
+	if err != nil || (resp != nil && resp.IsError()) {
+		t.Fatalf("bad: err:%v\nresp: %#v", err, resp)
+	}
+	entityID := resp.Data["id"].(string)
+
+	aliasData := map[string]interface{}{
+		"name":           "testaliasname",
+		"mount_accessor": githubAccessor,
+		"canonical_id":   entityID,
+	}
+
+	aliasReq := &logical.Request{
+		Operation: logical.UpdateOperation,
+		Path:      "entity-alias",
+		Data:      aliasData,
+	}
+
+	// This will create an alias and a corresponding entity
+	resp, err = is.HandleRequest(ctx, aliasReq)
+	if err != nil || (resp != nil && resp.IsError()) {
+		t.Fatalf("err:%v resp:%#v", err, resp)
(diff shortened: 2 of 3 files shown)


swayne275:
we may want to use something other than 400 for the duplicate request. perhaps 409? 400 is `The server could not understand the request due to invalid syntax.` which doesn't really seem to fit.

@chelshaw thoughts?

vault/identity_store_util.go
```diff
@@ -341,6 +341,12 @@ func (i *IdentityStore) loadEntities(ctx context.Context) error {
 					}
 				}
 
+				duplicateAccessors := getDuplicateAccessorsOnAliases(ctx, entity.Aliases)
+
+				// Log a warning if an entity has multiple aliases with the same accessor
+				if len(duplicateAccessors) > 0 {
+					i.logger.Warn("Duplicate aliases present for the same entity and accessor", "entity_name", entity.Name, "namespace_id", entity.NamespaceID, "duplicate_accessors", duplicateAccessors)
```
swayne275, line 348: are we concerned that this is leaking any privileged information?

vault/identity_store_aliases_test.go
```diff
@@ -405,6 +405,56 @@ func TestIdentityStore_AliasUpdate(t *testing.T) {
...
+
+	aliasData["name"] = "testaliasname2"
+	aliasReq = &logical.Request{
+		Operation: logical.UpdateOperation,
+		Path:      "entity-alias",
+		Data:      aliasData,
+	}
+
+	// This will try to create a new alias with the same accessor and entity
+	resp, err = is.HandleRequest(ctx, aliasReq)
```
swayne275, line 451: should we be doing something with `err`?

vault/identity_store_util.go
pmmukh: Hmm I don't think so? It logs some entity information and mount UUIDs, which afaik aren't sensitive, is there a specific field of the ones logged that concerns you more?

vault/identity_store_aliases_test.go
pmmukh: Ah yep, we should!

vault/identity_store_util.go
```diff
@@ -360,6 +366,25 @@ func (i *IdentityStore) loadEntities(ctx context.Context) error {
 	return nil
 }
 
+// getDuplicateAccessorsOnAliases returns a list of duplicate accessors present in the
+// passed in list of aliases
+func getDuplicateAccessorsOnAliases(ctx context.Context, aliases []*identity.Alias) []string {
```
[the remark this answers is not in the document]
pmmukh: Ah yeah, this can go, was just a force of habit thing I did, don't think it is required

vault/identity_store_aliases.go
```diff
@@ -267,6 +267,12 @@ func (i *IdentityStore) handleAliasCreate(ctx context.Context, req *logical.Requ
 		}
 	}
 
+	for _, alias := range entity.Aliases {
+		if alias.MountAccessor == mountAccessor {
+			return logical.ErrorResponse("Alias already exists for requested entity and mount accessor"), nil
```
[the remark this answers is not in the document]
pmmukh: Ah perfect, thanks for looking into that, I do want to change the status code to 409 as stephen also suggested, so this is helpful!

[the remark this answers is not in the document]
pmmukh: Based on our discussion today, I'll be passing on this and keeping the response status as 400 here.

pmmukh:
Looks like I've broken some existing tests with the aliasUpdate changes, taking a look at those now

pmmukh:
Just dropping a note that the updated TestCoreMetrics_EntityGauges is currently flaky, working on fixing that up now.

christophermaier:
Sorry to revive an old-ish thread, but is there a link to information for why this change was made? I previously had multiple aliases within a given accessor to the same entity, and now need to re-work how I'm representing things.

I would just like to understand what was wrong about having multiple aliases. Thanks in advance! 🙇

pmmukh:
hey @christophermaier ! Sure thing, so the issue was that the behavior could be exploited via a vulnerability, you can find the CVE noted here https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-43998, and some details around why we went with this solution to fix the cve in the 1.9 upgrade guide here https://www.vaultproject.io/docs/upgrading/upgrade-to-1.9.x#entity-alias-mapping.

christophermaier:
@pmmukh That makes sense; thanks for the links and for the very quick response! 👍

PR #12755, opened 2021-10-06
oss port of ent api lock basics

swayne275:
oss components of: https://github.com/hashicorp/vault-enterprise/pull/2171


vault/core.go
```diff
@@ -1967,6 +1968,7 @@ func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c
 	if err := c.setupMounts(ctx); err != nil {
 		return err
 	}
+	enterpriseSetupAPILock(c)
```
ncabatoff, line 1971: Should this be able to return an error?

swayne275: i was trying to keep the API lock init error free. there isn't much in the setup to error, unless we want to be really defensive and error if the logger or ent namespace manager or something don't exist (which are already set up before this is - on ent)

PR #12759, opened 2021-10-06
Add KV v1 and v2 support to Vault Agent Cache

remilapeyre:
This PR adds support for KV v1 and v2 to the Vault Agent Cache. 

For KV v1 it changes the logic that skip caching secrets to cache them based on their `lease_duration`.
For KV v2 a new parameter is added to the configuration to let the operator of the Vault Agent specify for how long they want KV v2 secrets and others non-leased secrets to be cached for.

All other secrets continue to work as before and other endpoints are still pass-through.


hghaf099:
@remilapeyre  would you please add a changelog to your PR following https://github.com/hashicorp/vault/blob/main/CHANGELOG.md

schokkal:
is this feature added to Vault Agent Cache - 12759 ?

duartecardao:
@hsimon-hashicorp has this been implemented already?

PR #12761, opened 2021-10-07
Added GMO GlobalSign getting started guide

sid-th:
@acahn - Will it be possible to add this Geeting started guide to Vault website content? TIA


schavis:
@sid-th As a general rule, the only plugins we document under the developer.hashicorp.com domain are those we develop and maintain ourselves. Documenting something on the site implies that we maintaining it and, more importantly, that we can vouch for its security and behavior. 

If would like editing help, I'm happy to do a review for you on a PR against the maintainer repo, according to the GlobalSign Atlas plugin in the registry that would be https://github.com/globalsign/atlas-hashicorp-vault.

Issue #7716, opened 2019-10-23
Feature Request: output the client source port to audit devices.


sischkg:
**Is your feature request related to a problem? Please describe.**

Vault outputs the source address of its client as "remote_address" to audit devices. 
However, since the source port of the client are not contained in Vault audit logs,
it is difficult to specify the client behind NAT(NAPT) or proxies.

[RFC 6302: Logging Recommendations for Internet-Facing Servers](https://tools.ietf.org/html/rfc6302)

**Describe the solution you'd like**

Vault outputs the source port of its client as "remote_port".

```json
{
  "time": "2019-10-23T01:47:40.061739097Z",
  "type": "request",
  "auth": {
    "token_type": "default"
  },
  "request": {
    "id": "***********************",
    "operation": "update",
    "namespace": {
      "id": "root"
    },
    "path": "auth/userpass/login/admin-user",
    "data": {
      "password": "hmac-sha256:********************"
    },
    "remote_address": "192.0.2.123",
    "remote_port": 12345
  }
}
```

PR #12790, opened 2021-10-10
Add remote_port in the audit logs when it is available

remilapeyre:
The `request.remote_port` field is now present in the audit log when it
is available:

```
{
  "time": "2021-10-10T13:53:51.760039Z",
  "type": "response",
  "auth": {
    "client_token": "hmac-sha256:1304aab0ac65747684e1b58248cc16715fa8f558f8d27e90fcbcb213220c0edf",
    "accessor": "hmac-sha256:f8cf0601dadd19aac84f205ded44c62898e3746a42108a51105a92ccc39baa43",
    "display_name": "root",
    "policies": [
      "root"
    ],
    "token_policies": [
      "root"
    ],
    "token_type": "service",
    "token_issue_time": "2021-10-10T15:53:44+02:00"
  },
  "request": {
    "id": "829c04a1-0352-2d9d-9bc9-00b928d33df5",
    "operation": "update",
    "mount_type": "system",
    "client_token": "hmac-sha256:1304aab0ac65747684e1b58248cc16715fa8f558f8d27e90fcbcb213220c0edf",
    "client_token_accessor": "hmac-sha256:f8cf0601dadd19aac84f205ded44c62898e3746a42108a51105a92ccc39baa43",
    "namespace": {
      "id": "root"
    },
    "path": "sys/audit/file",
    "data": {
      "description": "hmac-sha256:321a1d105f8c6fd62be4f34c4da4f0e6d1cdee9eb2ff4af0b59e1410950fe86b",
      "local": false,
      "options": {
        "file_path": "hmac-sha256:2421b5bf8dab1f9775b2e6e66e58d7bca99ab729f3f311782fda50717eee55b3"
      },
      "type": "hmac-sha256:30dff9607b4087e3ae6808b4a3aa395b1fc064e467748c55c25ddf0e9b150fcc"
    },
    "remote_address": "127.0.0.1",
    "remote_port": 54798
  },
  "response": {
    "mount_type": "system"
  }
}
```

Closes https://github.com/hashicorp/vault/issues/7716


remilapeyre:
> Hi @remilapeyre, thank you for the contribution! This looks like an intriguing and useful addition to audit log entries. Would you be able to add a test that verifies that the remote port is included in the audit log entry? This will help us prevent regressions. To ensure consistency, a few scenarios we might want to test are:
> 
>     * the remote host has a port and it is valid
> 
>     * the remote host has a port but it is invalid
> 
>     * the remote host does not have a port
> 
> 
> Here is a recent [example](https://github.com/hashicorp/vault/blob/aee77e3a3931bbb719e16d6d9f6216db98caad24/http/handler_test.go#L930) of a test that verifies audit log output based on a `logical.Request` that you might want to reference.

Hi @ccapurso, I updated the error handling and added a test for the happy case. I'm not sure how to test for when the port is invalid or missing since when using the test client it does not happen.

http/logical_test.go
```diff
@@ -499,3 +502,90 @@ func TestLogical_ShouldParseForm(t *testing.T) {
...
+	c := cluster.Cores[0].Client
+	vault.TestWaitActive(t, core)
+
+	if err := c.Sys().Mount("kv/", &api.MountInput{
+		Type: "kv-v2",
+	}); err != nil {
+		t.Fatalf("kv-v2 mount attempt failed - err: %#v\n", err)
+	}
+
+	auditLogFile, err := ioutil.TempFile("", "httppatch")
```
ccapurso, line 535: nit: the `"httppatch"` filename looks like it is left over from the example test that I provided. While it is only a temp file name, it may cause confusion.

ccapurso (changes requested):
I have one very minor change request but otherwise this looks good to me.

pmmukh:
Hi @remilapeyre ! Just dropping a reminder that Chris requested one more change on this PR, post which it should be good to merge !

remilapeyre:
Thanks for the ping @pmmukh, it should all be good now!

audit/format.go
```diff
@@ -346,6 +348,7 @@ type AuditRequest struct {
 	Data                          map[string]interface{} `json:"data,omitempty"`
 	PolicyOverride                bool                   `json:"policy_override,omitempty"`
 	RemoteAddr                    string                 `json:"remote_address,omitempty"`
+	RemotePort                    int                    `json:"remote_port,omitempty"`
```
HridoyRoy, line 351: Perhaps we should be consistent with int32 vs int64 here since the proto has this as int32 field.

audit/format.go
```diff
@@ -406,6 +409,14 @@ func getRemoteAddr(req *logical.Request) string {
 	return ""
 }
 
+// getRemotePort safely gets the remote port avoiding a nil pointer
+func getRemotePort(req *logical.Request) int {
```
HridoyRoy, line 413: Same here; we could specify `int32` for example, so that we don't end up with bit inconsistencies in the proto message in edge cases.

http/logical.go
```diff
@@ -491,14 +491,21 @@ WRITE_RESPONSE:
...
 func getConnection(r *http.Request) (connection *logical.Connection) {
 	var remoteAddr string
+	var remotePort int
 
-	remoteAddr, _, err := net.SplitHostPort(r.RemoteAddr)
+	remoteAddr, port, err := net.SplitHostPort(r.RemoteAddr)
 	if err != nil {
 		remoteAddr = ""
+	} else {
+		remotePort, err = strconv.Atoi(port)
```
HridoyRoy, line 500: Atoi should be ok here, but I think given that there are bit considerations we might want to use `ParseInt` instead.

http/logical_test.go
```diff
@@ -499,3 +502,90 @@ func TestLogical_ShouldParseForm(t *testing.T) {
 		}
 	}
 }
+
+func TestLogical_AuditPort(t *testing.T) {
```
HridoyRoy, line 506: This test looks great!

HridoyRoy (commented):
Hi @remilapeyre , the PR looks really good! I have some small questions primarily about int types, but otherwise I'm happy to approve.

Thanks for the contribution!

HridoyRoy (approved):
After discussing and thinking through the int type questions a little more, I think everything looks great! Lgtm!

ccapurso (approved):

Issue #10664, opened 2021-01-07
Include "num_uses" property to response when an authentication token is created


scruz-denodo:
**Is your feature request related to a problem? Please describe.**
When logging to vault you receive an authentication token which may have restrictions like a maximum TTL or a maximum number or uses. 
When that token is created, vault returns information about the maximum TTL (useful for knowing if token can be renewed), but there is not information about the number of uses. For getting that information, you have to request another endpoint _http://vault-url/v1/auth/token/lookup-self_ that returns the information.
Requesting this "lookup" endpoint consums an "use" of the token, and it is a problem when "one time tokens" are received, because you consum its unique "use" checking the information.

**Describe the solution you'd like**
As solution, return the information about the maximum number of uses when token is created.
Now, received data at token creation time is something like:

```
{
	"request_id": "02e1a9aa-081c-2c6f-5853-993552cac04f",
	"lease_id": "",
	"renewable": false,
	"lease_duration": 0,
	"data": null,
	"wrap_info": null,
	"warnings": null,
	"auth": {
		"client_token": "s.nKuuiTrUKdkAu5u5uYiYd2Js",
		"accessor": "1R0FnTq5gt1srywsFo3kXQnT",
		"policies": [
			"apps",
			"default"
		],
		"token_policies": [
			"apps",
			"default"
		],
		"metadata": {
			"role_name": "role-token-expiring"
		},
		"lease_duration": 10,
		"renewable": true,
		"entity_id": "b8b1d47c-ad59-73da-4154-db24b9c546a8",
		"token_type": "service",
		"orphan": true
	}
}
```
When "lookup" endpoint is requested for previous token, the "**num_uses**" property is returned.
```
                ...
		"issue_time": "2021-01-07T08:15:29.608571079Z",
		"meta": {
			"role_name": "role-token-expiring"
		},
		"num_uses": 1,
		"orphan": true,
		"path": "auth/approle/login"
                ...
```

Problem would be solved if "**num_uses**" property is returned with first information.

**Describe alternatives you've considered**
For detecting this situation, when "lookup" process is executed and "num_uses" value is "-1", you know that it is a "one time token" because the available usages are already fulfilled with the "lookup" request. 
But this is not the best option, because you are always consuming an "use" for knowing that information. Returning "num_uses" at fist steps would be the best option.

**Explain any additional use-cases**
Any use-cases when "one-time-tokens" are created and the application does not know about what type of restrictions will have the received tokens.

**Additional context**
[...]

PR #12791, opened 2021-10-10
Return num_uses during authentication

remilapeyre:
Closes https://github.com/hashicorp/vault/issues/10664


HridoyRoy (approved):
Thanks for this contribution! Lgtm!

PR #12804, opened 2021-10-11
adjustment of order of options

tograla:
Hey! It feels to me that listing the options from the one with the highest priority is more intuitive, also it makes the order of options listed as bullet-points aligned with how they are listed in the parenthesis.


hsimon-hashicorp:
Thank you for your contribution, @tograla! :)

PR #12835, opened 2021-10-14
Add Root Token check for TokenEntry

swayne275:
OSS parts of https://github.com/hashicorp/vault-enterprise/pull/2213


swayne275:
re-requesting reviews because i added the changelog

PR #12872, opened 2021-10-19
Restrict ECDSA/NIST P-Curve hash function sizes for cert signing

cipherboy:
In #11006, it was mentioned that #11245 updated this behavior but didn't apply the size restrictions. We make the following changes (along side #11216, which I believe is still relevant):

 - Default to `0` signature bits, to perform auto detection if not otherwise specified. Presently this only does auto detection for ECDSA keys and leaves the default 256-bit hash function for all RSA key sizes. In the future, we could consider following NIST guidelines (and using 2048->256, 3072->384, 4096->512).
 - Introduce `ValidateKeyTypeSignatureLength` for cross Key/Signature validations (and which handles the actual autodetection)

Since #11216 lacks a changelog entry, ~~we should likely update #11245's changelog entry to match the new behavior and provide a unified changelog entry for the combined change (as it appears 11245 hasn't yet landed in a release yet)~~.

**Edit**: Added a changelog entry.

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -827,7 +827,6 @@ func generateRoleSteps(t *testing.T, useCSRs bool) []logicaltest.TestStep {
 		KeyType:       "rsa",
 		KeyBits:       2048,
 		RequireCN:     true,
-		SignatureBits: -1,
 	}
 	issueVals := certutil.IssueData{}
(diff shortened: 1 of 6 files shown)


sgmiller (approved):

cipherboy:
Since the race test appears to have been broken in main, and the race in question appears to be unrelated to the present changes, going ahead and merging. Thanks all!

PR #12880, opened 2021-10-20
clarify keytab gen process

martinhristov90:
Updating the docs to have a clear statement that two keytabs are needed, one for the Vault server itself, and one for the login purposes which belongs to the user which would like to login via Kerberos auth method.


aphorise:
@yhyakuna - I think this may have missed release at the time - any chance of merging it?

yhyakuna:
I don't know why the checks don't seem to complete. I created [PR 16803](https://github.com/hashicorp/vault/pull/16803) to replicate this, so closing this..

PR #12881, opened 2021-10-20
operator generate-root -decode: allow token from stdin

davidducros:
Allow passing "-" as the value for -decode, causing the encoded token to
be read from stdin. This is intended to prevent leaking the encoded
token + otp into process logs in enterprise environments.


davidducros:
Created issue #12882 to cover this request.

PR #12892, opened 2021-10-21
CRT onboarding 1.7.x

claire-labry:


ncabatoff:
Why no CGO_ENABLED=0 in build.yml like in 1.9/main?

kpenfound:
@ncabatoff good catch, added it. It looks like packagespec had cgo_enabled=0 so this should too

ncabatoff (approved):

PR #12894, opened 2021-10-21
website: fix print styles by bumping deps

zchsh:
👀 [Preview](https://vault-git-zsfix-print-styles-hashicorp.vercel.app)

This PR bumps dependencies in order to address `@media print` style issues. Specifically:
- Bumps to latest `@hashicorp/react-consent-manager`, which hides itself in print media
- Bumps to latest `@hashicorp/react-subnav`, which hides itself in print media
- Removes now redundant project-level CSS intended to achieve the same.

Underlying component changes were made in https://github.com/hashicorp/react-components/pull/406. Also bumps to updated consent-manager with changes here: https://github.com/hashicorp/react-components/pull/407.


zchsh:
Not sure about the failing `ci/circle-ci: test-go` check - this PR only affects `website` code, so I think that might just be flake or something?

PR #12896, opened 2021-10-21
Prevent constant-refresh UI bug

hashishaw:
This alert prevents a bug where the UI seems to constantly refresh. This happens when the `VAULT_CLOUD_ADMIN_NAMESPACE=1` as env variable and vault binary is OSS.


hellobontempo (approved):
Love "developer ease of life" fixes

PR #12904, opened 2021-10-22
UI/Remove token_type field from token auth method

hellobontempo:
For the Token Auth method the token_type field should not exist, but may exist for other auth methods. Also made all auth methods linkable plus added copy for those that the UI doesn't support management of.

**With Fix**
auth-methods-token-type

**Former Behavior**
image

**All auth methods linkable**
auth-methods-list


hashishaw (approved):
Looks great, thank you!

PR #12906, opened 2021-10-22
Vertical resize on .cm-s-hashi.CodeMirror

djvs:
Allows user to resize JSON editor to be more or less fullscreen.  Essential UI fix IMO.


djvs:
By the way this resize grabber is almost invisible on the latest version - 
2022-02-28T20:33:01,813636649-05:00

Monkeychip:
@djvs looking into this a little. It appears this is the standard resize icon for codemirror.  I suspect it's a little hard to see because of the black background. Is this the size you are seeing as well? I have been playing around with this and don't see a way to make it any larger.
image

PR #12955, opened 2021-10-28
UI/OIDC namespace support

hashishaw:
Adds support for a namespace in the OIDC Authorization Flow path. As a result, the original oidc provider path was renamed, and updated with logic that can handle a namespace or no namespace

Since this is an addition to the original OIDC auth flow work (#12800), there is no changelog. 

Forthcoming PR will include tests for this workflow

When Authorization Flow includes a namespace and the user must log in, they are redirected to the login screen with the namespace field filled out and uneditable:
Screen Shot 2021-10-28 at 6 05 03 PM


arnav28 (approved):
Looks great!! 🍾

PR #12964, opened 2021-10-29
On bar chart selection replace search component with HTML

arnav28:
Maintaining state between search select component and bar chart component is tricky considering the time, so as a fallback solution, when user selects the namespace from chart, replace the search select component with custom HTML. Additionally clear out the namespace selection when new date is queried

https://user-images.githubusercontent.com/2932708/139498083-490eb793-d58f-42e9-911d-e4ab11ed7497.mov


hashishaw (approved):
Thanks for tackling this!

PR #12965, opened 2021-10-29
cockroachdb: add high-availability support

DuskEagle:
This commit adds high-availability support to the CockroachDB backend. The
locking strategy implemented is heavily influenced from the very similar
Postgres backend.

### DIFF
--- a/physical/cockroachdb/cockroachdb.go
+++ b/physical/cockroachdb/cockroachdb.go
@@ -28,18 +28,23 @@ var (
 )
 
 const (
-	defaultTableName = "vault_kv_store"
+	defaultTableName   = "vault_kv_store"
+	defaultHATableName = "vault_ha_locks"
 )
 
 // CockroachDBBackend Backend is a physical backend that stores data
 // within a CockroachDB database.
 type CockroachDBBackend struct {
-	table         string
-	client        *sql.DB
-	rawStatements map[string]string
-	statements    map[string]*sql.Stmt
-	logger        log.Logger
-	permitPool    *physical.PermitPool
+	table           string
+	haTable         string
+	client          *sql.DB
+	rawStatements   map[string]string
+	statements      map[string]*sql.Stmt
+	rawHAStatements map[string]string
+	haStatements    map[string]*sql.Stmt
+	logger          log.Logger
+	permitPool      *physical.PermitPool
+	haEnabled       bool
 }
 
 // NewCockroachDBBackend constructs a CockroachDB backend using the given
@@ -51,6 +56,8 @@ func NewCockroachDBBackend(conf map[string]string, logger log.Logger) (physical.
 		return nil, fmt.Errorf("missing connection_url")
 	}
 
+	haEnabled := conf["ha_enabled"] == "true"
+
 	dbTable := conf["table"]
 	if dbTable == "" {
 		dbTable = defaultTableName
@@ -59,6 +66,16 @@ func NewCockroachDBBackend(conf map[string]string, logger log.Logger) (physical.
 	err := validateDBTable(dbTable)
 	if err != nil {
 		return nil, fmt.Errorf("invalid table: %w", err)
+	}
+
+	dbHATable, ok := conf["ha_table"]
+	if !ok {
+		dbHATable = defaultHATableName
+	}
+
+	err = validateDBTable(dbHATable)
+	if err != nil {
+		return nil, fmt.Errorf("invalid HA table: %w", err)
 	}
 
 	maxParStr, ok := conf["max_parallel"]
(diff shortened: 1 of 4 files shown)


hsimon-hashicorp:
Hi there! Thanks for this submission. Please sign the CLA, and we can get this into review. :)
(Note: If you've signed the CLA at an organizational level, please make sure your organization is publicly viewable, so the CLA bot can pick that up.)

DuskEagle:
Hi @hsimon-hashicorp , is there anything I can help with to get the review process rolling on this PR?

hsimon-hashicorp:
Hi there @DuskEagle - I'll nudge our engineers about this again. Thanks for your patience. :)

physical/cockroachdb/cockroachdb.go
```diff
@@ -51,6 +56,11 @@ func NewCockroachDBBackend(conf map[string]string, logger log.Logger) (physical.
 		return nil, fmt.Errorf("missing connection_url")
 	}
 
+	var haEnabled bool
+	if haEnabledStr, ok := conf["ha_enabled"]; ok && haEnabledStr == "true" {
```
ncabatoff, line 60: This is fine as is, but since in Go an absent key looked up in a map results in the zero value, and this is a map[string]string, you can just say `haEnabled = conf["ha_enabled"] == "true"` if you want to be a bit more succinct.

physical/cockroachdb/cockroachdb.go
```diff
@@ -99,26 +132,45 @@ func NewCockroachDBBackend(conf map[string]string, logger log.Logger) (physical.
...
-		permitPool: physical.NewPermitPool(maxParInt),
+		rawHAStatements: map[string]string{
+			"get": "SELECT ha_value FROM " + dbHATable + " WHERE NOW() <= valid_until AND ha_key = $1",
+			"upsert": "INSERT INTO " + dbHATable + " as t (ha_identity, ha_key, ha_value, valid_until)" +
+				" VALUES ($1, $2, $3, NOW() + $4) " +
+				" ON CONFLICT (ha_key) DO " +
+				" UPDATE SET (ha_identity, ha_key, ha_value, valid_until) = ($1, $2, $3, NOW() + $4) " +
+				" WHERE (t.valid_until < NOW() AND t.ha_key = $2) OR " +
+				" (t.ha_identity = $1 AND t.ha_key = $2)  ",
+			"delete": "DELETE FROM " + dbHATable + " WHERE ha_identity = $1 AND ha_key = $2",
```
ncabatoff, line 143: Doesn't hurt to do it this way, and I know you're basing this on the PG storage which does this.  But I realized today that we don't need to include ha_identity in this DELETE, since there's a unique index on ha_key.  So if you want you could remove that part of the WHERE clause.

physical/cockroachdb/cockroachdb.go
```diff
@@ -99,26 +132,45 @@ func NewCockroachDBBackend(conf map[string]string, logger log.Logger) (physical.
...
 		},
 		statements: make(map[string]*sql.Stmt),
-		logger:     logger,
-		permitPool: physical.NewPermitPool(maxParInt),
+		rawHAStatements: map[string]string{
+			"get": "SELECT ha_value FROM " + dbHATable + " WHERE NOW() <= valid_until AND ha_key = $1",
+			"upsert": "INSERT INTO " + dbHATable + " as t (ha_identity, ha_key, ha_value, valid_until)" +
+				" VALUES ($1, $2, $3, NOW() + $4) " +
+				" ON CONFLICT (ha_key) DO " +
+				" UPDATE SET (ha_identity, ha_key, ha_value, valid_until) = ($1, $2, $3, NOW() + $4) " +
```
ncabatoff, line 140: I know little about CockroachDB, but I thought it was fully compatible with PG.  You've been diligent about adapting the PG approach here, but I noticed this one discrepancy: instead of `NOW() + $4 * INTERVAL '1 seconds'` you have simply `NOW() + $4`.  Why's that?

physical/cockroachdb/cockroachdb.go
DuskEagle: That's a good point! I've removed that part of the where clause.

physical/cockroachdb/cockroachdb.go
DuskEagle: What's happening here is that `$4` is being passed in as a string type, not an integer type. When we attempt to to `$4 * INTERVAL '1 seconds'`, we get a conversion error:

```
ERROR: unsupported binary operator: <string> * <interval> (desired <interval>)
SQLSTATE: 22023
```

We would have the option of casting `$4` to an integer, and then multiplying by an interval: `NOW() + $4::int * INTERVAL '1 seconds'`.

However, $4 is already a value in seconds, and so casting to an int just to multiply by 1 second is unnecessary. It's more succinct to just add the string value directly to `NOW()`.

DuskEagle (commented):
Apologies for the delay. I've addressed your comments now :).

physical/cockroachdb/cockroachdb_ha.go
```diff
@@ -0,0 +1,197 @@
...
+// the lock could be acquired successfully. The returned channel will be closed
+// once the lock in the CockroachDB table cannot be renewed, either due to an
+// error speaking to CockroachDB or because someone else has taken it.
+func (l *CockroachDBLock) Lock(stopCh <-chan struct{}) (<-chan struct{}, error) {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
+	var (
+		success = make(chan struct{})
+		errors  = make(chan error)
```
ncabatoff, line 83: You might want to make this channel buffered.  I saw some output from a static analysis tool recently showing that in the PG implementation, there's an issue here when tryToLock attempts to write to the error channel, but Lock has already returned due to stopCh being closed.  In addition, you could make that write to the error channel in Lock tolerate blockage by using the optional send syntax.

physical/cockroachdb/cockroachdb_test.go
```diff
@@ -33,7 +34,15 @@ func prepareCockroachDBTestContainer(t *testing.T) (func(), *Config) {
 		if tableName == "" {
 			tableName = defaultTableName
 		}
-		return func() {}, &Config{*s, "vault." + tableName}
+		haTableName := os.Getenv("CR_HA_TABLE")
```
ncabatoff, line 37: I don't think you need Getenv in this test.  I know it's present in some of the existing tests, but it's vestigial mostly.

ncabatoff (approved):
Looks good!  I have a couple of comments that you can choose to address, or not: I'm willing to merge regardless.  Let me know how you'd like to proceed.

physical/cockroachdb/cockroachdb_ha.go
DuskEagle: I've taken your suggestions and made the channel buffered and the send non-blocking.

physical/cockroachdb/cockroachdb_test.go
DuskEagle: I've removed the `os.Getenv` calls for the `CR_TABLE` and `CR_HA_TABLE` environment variables, so the tests will use the default values.

DuskEagle (commented):
Thanks for the review! I've made the changes you recommend, if they look good to you I'd like to see this merged :).

ncabatoff:
Not sure why those PG tests are failing in the remote-docker CI tests, but I'm confident it's unrelated to these changes.

ncabatoff:
Thanks @DuskEagle !

DuskEagle:
Thank you @ncabatoff !

PR #12968, opened 2021-10-29
Document allow_different_signature_algorithm param

sgmiller:


website/content/api-docs/secret/pki.mdx
```diff
@@ -1358,6 +1358,8 @@ This endpoint requires `sudo` capability.
 ### Parameters
 
 - `certificate` `(string: <required>)` – Specifies the PEM-encoded self-issued certificate.
+- `allow_different_signature_algorithm` `(bool: false)` - If true, allows the submitted
```
cipherboy, line 1361: This doesn't intuitively make sense to me. Perhaps something like:

> If true, allows the submitted certificate to be signed when the submitted certificate and the root certificate differ in type.

(I think to me, it doesn't make sense because "the signature algorithm" isn't actually specified anywhere in this request; it was incorrectly inferred based on the certificate's preferences, when in reality, it always needs to match the root)

I also wonder when you would ever want this to be false, and feel it should default to true. Most CAs that I'm aware of (RHCS and LE) really don't care too much what type of CSR you give it to sign. Cross signing, while admittedly a limited use case, really shouldn't impact this feature set, IMO.

What's your thoughts?

sgmiller: If we flip the behavior I'd want to rename the flag too, maybe `require_matching_signature_algorithm`, default false.

cipherboy: I agree with that idea, but to me, the restriction again comes from the _keys_ -- the signature algorithm is a property of the keys (specifically the parent key), not artificially selected along the way (outside of selection of hash function and bitsize of it)...

Perhaps `require_matching_key_type`? It only affects cross-signing, with the understanding that maybe in the future it'll be used to replace the existing CA cert, right?

PR #12976, opened 2021-10-29
Incorporate Ember Flight Icons

zofskeez:
Adds [Ember Flight Icons](https://flight-hashicorp.vercel.app/) to UI and updates _Icon_ component to use flight icons when available.

**Overview**
- added `ember-flight-icons` dependency
- added babel plugin to import json of icon map to determine whether to use the _FlightIcon_ component or fallback to existing implementation
- updated _Icon_ component to display icons from both sets
- updated _Icon_ component usages in application to new api and changed names to flight icon set where applicable
- added styling to flight icons to align with existing icon style
- added tests for new _Icon_ component logic
- updated icon story to display mappings to flight set
- updated storybook config to insert svg sprite into dom for flight icons

_View the Icon story to see a table of the icon mappings by running_ `yarn storybook` _and then click_ **Icon** _in the sidebar._


ui/.storybook/config.js
```diff
@@ -18,6 +19,9 @@ addParameters({
 addDecorator(storyFn => {
   const { template, context } = storyFn();
 
+  // flight icon sprite must be inserted into dom for icon lookup via use element
+  document.getElementById('root').insertAdjacentHTML('afterbegin', flightIconSprite.trim());
```
hashishaw, line 23: 👏

ui/app/models/transit-key.js
```diff
@@ -23,20 +23,28 @@ const ACTION_VALUES = {
...
     isSupported: 'supportsEncryption',
     description: 'Rewraps the ciphertext using the latest version of the named key',
-    glyph: 'refresh-default',
+    glyph: 'reload',
   },
   sign: {
     isSupported: 'supportsSigning',
     description: 'Get the cryptographic signature of the given data',
-    glyph: 'edit',
+    glyph: 'penicil-tool',
```
hashishaw, line 31: looks like a typo?

ui/lib/core/addon/templates/components/icon.hbs
```diff
@@ -1,3 +1,12 @@
-<span class="hs-icon {{sizeClass}} {{class}}" ...attributes>
-  {{svg-jar @glyph}}
-</span>
\ No newline at end of file
+{{#if this.isFlightIcon}}
+  <FlightIcon
+    @name={{@name}}
+    @size={{this.flightIconSize}}
+    class={{this.iconClass}}
```
hashishaw, line 5: Is `iconClass` still relevant to flight icons, since we pass in the size attribute?

ui/lib/core/icon-mappings.js
```diff
@@ -0,0 +1,218 @@
+export const localIconMap = {
```
hashishaw, line 1: Could we get a couple comments for our future selves about what the mappings mean? Eg. "if a value is null in `localIconMap` it means there isn't a Flight Icon equivalent"

ui/lib/core/addon/components/icon.js
```diff
@@ -2,35 +2,54 @@
...
 export default Component.extend({
   tagName: '',
   layout,
-  glyph: null,
-  size: 'm',
-  sizeClass: computed('size', function() {
-    let { size } = this;
+  name: null,
+  sizeClass: 'm', // hs-icon specific
+  size: null, // fight icon specific
```
hashishaw, line 30: Is the default size inherited from Flight Icon? Should we be prescriptive about our default for Flight icons here?

ui/tests/integration/components/icon-test.js
```diff
@@ -20,12 +20,35 @@ module('Integration | Component | icon', function(hooks) {
...
     let err = await promise;
-    assert.ok(err.message.includes('The size property of'), "errors when passed a size that's not allowed");
+    assert.ok(
+      err.message.includes('The sizeClass property of'),
+      "errors when passed a sizeClass that's not allowed"
+    );
+  });
+
+  test('it should render FlightIcon', async function(assert) {
+    assert.expect(5);
```
hashishaw, line 36: What is this?

ui/tests/integration/components/icon-test.js
```diff
@@ -20,12 +20,35 @@ module('Integration | Component | icon', function(hooks) {
...
+    render(hbs`<Icon @name="vault-logo" @sizeClass="no"/>`);
     let err = await promise;
-    assert.ok(err.message.includes('The size property of'), "errors when passed a size that's not allowed");
+    assert.ok(
+      err.message.includes('The sizeClass property of'),
+      "errors when passed a sizeClass that's not allowed"
+    );
+  });
+
+  test('it should render FlightIcon', async function(assert) {
```
hashishaw, line 35: 👏

ui/app/templates/components/console/log-error-with-html.hbs
```diff
@@ -1,5 +1,5 @@
 {{! template-lint-disable no-triple-curlies}}
 <div class="console-ui-alert has-text-danger">
-  <Icon @glyph="cancel-circle-fill" aria-hidden="true" />
+  <Icon @name="x-circle-fill" />
```
hashishaw, line 3: I think we should keep `aria-hidden` wherever we had it in the dom before -- quite a few places where this was removed

hashishaw (changes requested):
Great work so far! Excited to finally be consuming the Flight Icons.

ui/app/models/transit-key.js
zofskeez: Nice catch. Definite fat finger. Pushed a change.

ui/lib/core/addon/templates/components/icon.hbs
zofskeez: It is in order to keep the styling consistent between icon sets. Currently flight icons only support 16 and 24 for the `size` param but there are 6 different sizes for the _hs-icon_ class. I created `flight-icon.scss` to tweak the _hs-icon_ size classes to match as closely as possible which is then applied to the component.

I wrestled with this a bit but wanted to support the `size` parameter for _FlightIcon_ while still preserving the _hs-icon_ sizing that existed. If it's confusing we could make `size` private for now and rely on the _hs-icon_ size classes until _FlightIcon_ size becomes more flexible.

ui/lib/core/icon-mappings.js
zofskeez: Good call I added some in the file.

ui/lib/core/addon/components/icon.js
zofskeez: Yes the _FlightIcon_ default is 16. There is a `flightIconSize` computed property which gets passed to the component that will set the size to either 16 or 24 based on the `sizeClass` value if `size` was not provided which is the reason for the `null` default.

ui/tests/integration/components/icon-test.js
zofskeez: This ensures that an explicit number of assertions are run within the test, in this case 5. It's a force of habit for me but basically just helps to guard against unexpected assertions or unintended modifications to a test perhaps.

ui/app/templates/components/console/log-error-with-html.hbs
zofskeez: I removed these on the instances where the `name` was from the flight set since they add `aria-hidden="true"` to the underlying svg by default.
image

ui/lib/core/addon/templates/components/icon.hbs
hashishaw: I wonder if we should instead conform our sizes to the new standard 🤔 may want to wait for Ivana to weigh in on this one

zofskeez: @ivana-mcconnell what are your thoughts on this? You can see how the sizes are mapped in [this commit](https://github.com/hashicorp/vault/pull/12976/commits/a1fb58d00ca321b5c51a77257a90df9514910845#diff-74a077143fbd0008b419aa84c0809ca816c9c3496eddbcaddc7ee898a56e32cf) and here's a breakdown of the number of times the _Icon_ component is used in the app across the various sizes:

s - 7
m - 118
l - 37
xlm - 3
xl - 11
xxl - 1

ui/lib/core/addon/components/icon.js
ivana-mcconnell: 👋  @zofskeez asked me to weigh in as well. If I understand right, I'm on board with our default being prescriptive to 16px; however, it seems like right now there are icons kicking around that are `small` (i.e. 10px) in seven or so instances. We shouldn't just change it and leave them there. What we could maybe do is split those out into a separate task to clean them up/go through them one by one and decide if a 16px icon or no icon is appropriate.

ui/lib/core/addon/templates/components/icon.hbs
ivana-mcconnell: I'm on board with conforming our sizes to the new standard, but I think that it's worth a pairing session to look at the instances where icons are _not_ 16px or 24px and making adjustments for them, just so we don't end up with something incredibly disproportionate. It looks like that's a total of 22 instances (s, xlm, xl, xxl). Might be a little tedious, but probably worthwhile.

zofskeez: @hashishaw after pairing with Ivana and confirming the new size standards I updated the component and simplified things in [this commit](https://github.com/hashicorp/vault/pull/12976/commits/5aa1cd3f294d748587cdda076e088ea1defbf100).

ui/lib/core/addon/components/icon.js
zofskeez: @hashishaw the `sizeClass` property has been removed and `size` is now used exclusively with a default of 16 as of [this commit](https://github.com/hashicorp/vault/pull/12976/commits/5aa1cd3f294d748587cdda076e088ea1defbf100).

hashishaw (approved):
Great work, thank you!

PR #12977, opened 2021-10-29
Break line if words are long in flash messages

arnav28:
Screen Shot 2021-10-29 at 3 32 01 PM


hellobontempo (approved):
Yay! Thanks for fixing this

Issue #128, opened 2015-05-02
vault requires being in PATH


robfig:
Great tool, thank you! 

Minor nit, but presumably it's not necessary to require the vault CLI tool to be in PATH?  (OSX)

```
$ ./vault status
Error initializing client: Error: exit status 127

sh: vault: command not found

11:04:11 [robfig@robfig ~/Downloads]
$ PATH=$PATH:~/Downloads

11:04:18 [robfig@robfig ~/Downloads]
$ ./vault status
Sealed: false
Key Shares: 1
Key Threshold: 1
Unseal Progress: 0

High-Availability Enabled: false
```

mitchellh:
Fixed in master, 0.1.1 will be relesaed soon :)

PR #12980, opened 2021-10-31
Update documentation to remove token_reviewer_jwt, removed from UI

RemcoBuddelmeijer:
# Overview

Removing the usage of token_reviewer_jwt entry from Kubernetes's access UI. Removing from the documentation completely.

# Related Issues/Pull Requests
[Issue #12951](https://github.com/hashicorp/vault/issues/12951)
[PR #128](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/128)

# Contributor Checklist
[x] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
The docs are altered in this PR
[x] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
[x] Backwards compatible
Vault keeps working as before, only difference will be approach of the validation of the sent-in JWT token. However, this PR aims towards removing this functionality. This PR however only focusses on UI changes.

### DIFF
--- a/ui/app/models/auth-config/kubernetes.js
+++ b/ui/app/models/auth-config/kubernetes.js
@@ -17,11 +17,6 @@ export default AuthConfig.extend({
     helpText: 'PEM encoded CA cert for use by the TLS client used to talk with the Kubernetes API',
   }),
 
-  tokenReviewerJwt: attr('string', {
-    helpText:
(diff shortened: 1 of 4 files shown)


hsimon-hashicorp:
Usually, we err on the side of giving lots of notice when removing functionality, with deprecation notices and announced versions where it will take effect. As such, the Vault engineering teams will be discussing this internally - your patience is appreciated. :)

PR #12987, opened 2021-11-01
1.10.0-dev version bump

mladlow:


ncabatoff (approved):

swayne275:
Does `TestStoreMultipleVaultVersions` need to be updated? It's doing 1.9 stuff, but i'm not familiar with that test or why we do it. @HridoyRoy do you know more?

mladlow:
Ideally, it shouldn't be version-specific, right? Or at least, if the version goes up, it should still work.

ncabatoff:
Fix for broken tests is in https://github.com/hashicorp/vault/pull/13026 (which uses this branch as a base because otherwise we won't see the test failures that we're trying to fix.)

swayne275 (approved):

PR #12988, opened 2021-11-01
Flip the semantics of the matching key algorithms in sign-self-issued.

sgmiller:


builtin/logical/pki/path_root.go
```diff
@@ -109,9 +109,10 @@ func pathSignSelfIssued(b *backend) *framework.Path {
 				Type:        framework.TypeString,
 				Description: `PEM-format self-issued certificate to be signed.`,
 			},
-			"allow_different_signature_algorithm": &framework.FieldSchema{
+			"require_matching_certificate_algorithms": &framework.FieldSchema{
 				Type:        framework.TypeBool,
-				Description: `If true, allow the public key type of the signer to differ from the self issued certificate.`,
+				Default:     false,
+				Description: `If true, allow the public key algorithm of the signer to differ from the self issued certificate.`,
```
cipherboy, line 115: I don't think this description is correct? Note that we switched the order of this variable, so I think the description would be

> if true, require the public key algorithm ... to match ...

builtin/logical/pki/path_root.go
```diff
@@ -451,9 +452,12 @@ func (b *backend) pathCASignSelfIssued(ctx context.Context, req *logical.Request
 	}
 
 	if signingPubType != certPubType {
-		b, ok := data.GetOk("allow_different_signature_algorithm")
-		if ok && b.(bool) {
+		b, ok := data.GetOk("require_matching_certificate_algorithms")
+		if !ok || !b.(bool) {
```
cipherboy, line 456: Check looks correct though! :) 

 - Err: assume default of False; succeed
 - False value: Use the signature algorithm.
 - True value: Drop to the err case below.

builtin/logical/pki/path_root.go
sgmiller: Yep, whoops.

sgmiller: Weird, my local codebase doesn't read like that.

cipherboy (approved):
Change set looks good and the new behavior is better, thanks @sgmiller! 

I've tested this locally with ECDSA and RSA certs in a variety of parameter configurations and I think it'll do what users expect (and not make Consul/... add another API request parameter).

LGTM assuming CI passes. :)

PR #12993, opened 2021-11-01
[VAULT-4034] Only cache non-nil values

pmmukh:
- This PR updates the cache layer to only cache non-nil values.
- The issue fixed here, is if a nil value is persisted into the cache, it appears like there is a cache entry for the key, and so subsequent lookups will not update the cache value even if the underlying storage has been updated to have a non-nil value for that key
- The inmem backend is being used to test this change, and removing the change in cache.go will make the new test fail.
- The scripts attached in the jira ticket were also run against an ent binary with these changes ( as the scripts require perf standbys ) using this commit https://github.com/hashicorp/vault-enterprise/commit/435fd6f1e865abc17fae3c4fe7fc269d45a600e7, and the token lookup was verified to work when the perf standbys were brought back up.

### DIFF
--- a/sdk/physical/inmem/cache_test.go
+++ b/sdk/physical/inmem/cache_test.go
@@ -328,3 +328,45 @@ func TestCache_Refresh(t *testing.T) {
 		t.Fatalf("expected value baz, got %s", string(r.Value))
 	}
 }
+
+// TestCache_UpdateFromStorage fetches a nil value from cache, updates
+// the underlying storage and checks the cache value on a subsequent lookup
+func TestCache_UpdateFromStorage(t *testing.T) {
+	logger := logging.NewVaultLogger(log.Debug)
+
+	inm, err := NewInmem(nil, logger)
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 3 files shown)


ncabatoff (approved):

sdk/physical/inmem/cache_test.go
```diff
@@ -328,3 +328,45 @@ func TestCache_Refresh(t *testing.T) {
 		t.Fatalf("expected value baz, got %s", string(r.Value))
 	}
 }
+
+// TestCache_UpdateFromStorage fetches a nil value from cache, updates
+// the underlying storage and checks the cache value on a subsequent lookup
+func TestCache_UpdateFromStorage(t *testing.T) {
```
swayne275, line 334: is this supposed to be testing the change made to `cache.go`? if so, i'm not sure how it does so. i'd think we'd want to write something nil and then verify that it's not in the cache?

pmmukh: chatted with @swayne275 , verified this test is checking the change in cache.go appropriately, closing this out.

PR #13000, opened 2021-11-01
UI/kv codemirror diff

Monkeychip:
**Adding a Diff view for the KV2 secrets engine.**

See video for the diff view. It only shows if you have read access and more than 1 version of a secret. If you have deleted or destroyed a secret it will show null on page load and then not let you select that secret version. The icons in the version dropdown help you understand the secret's status.

To add this feature I added a new route and new component. I had played with extending the json-editor component that we currently use with codemirror, but that proved too messy. Additional we use a different library (jsonDiffPatch, used by another Hashicorp product) instead of codemirror because of dependency loading issues.

https://user-images.githubusercontent.com/6618863/143951114-d89c5679-b246-428b-b3c6-7fffc2070d19.mp4


ui/app/components/json-editor.js
```diff
@@ -1,4 +1,21 @@
-import Component from '@ember/component';
```
Monkeychip, line 1: only change is glimmerized this component. I was originally here thinking I'd modify this component. In the end I don't use this component, but played boy scout and "left it better than I found it."

ui/app/styles/components/diff-version-selector.scss
```diff
@@ -0,0 +1,24 @@
+.visual-diff {
+  background-color: black;
+
+  pre {
+    color: $ui-gray-010;
+  }
+}
+
```
Monkeychip, line 8: over-riding the packages colors. We wanted a specific red and green to match the reds and greens we use in the app. confirmed colors with design.

ui/app/templates/components/json-editor.hbs
```diff
@@ -1,32 +1,34 @@
-{{#if showToolbar }}
```
Monkeychip, line 1: this diff looks strange, but I think when this was last worked on the developer had their tab settings at 4. I changed the file tab settings to 2 to match the code base and this was the resulting diff. If you're nervous checkout the file in the branch and double check.

ui/tests/pages/components/control-group-success.js
```diff
@@ -1,7 +1,7 @@
 import { isPresent, fillable, clickable } from 'ember-cli-page-object';
 
 export default {
-  showsJsonViewer: isPresent('[data-test-json-viewer]'),
+  showsJsonViewer: isPresent('[data-test-component="json-editor"]'),
```
Monkeychip, line 4: because of changing the component to glimmer, this was wrapped around the component, but now that it's glimmer that doesn't work.

ui/app/templates/components/control-group-success.hbs
```diff
@@ -11,7 +11,6 @@
     <div class="control-group-success {{if unwrapData 'is-editor'}}">
       <div class="has-copy-button">
         <JsonEditor
-          data-test-json-viewer
```
Monkeychip, line 14: removing because this component has been changed to glimmer.

ui/app/adapters/secret-v2-version.js
```diff
@@ -54,6 +54,12 @@ export default ApplicationAdapter.extend({
     });
   },
 
+  querySecretDataByVersion(id) {
```
hellobontempo, line 57: Should we return the error so we have access to the secret metadata that tells us if a secret has been deleted/destroyed? Right now the promise fails if a secret version has been deleted/destroyed.

ui/app/templates/components/diff-version-selector.hbs
```diff
@@ -0,0 +1,94 @@
...
+        <nav class="box menu">
+          <ul class="menu-list">
+            {{#each (reverse @model.versions) as |leftSideSecretVersion|}}
+              <li class="action" data-test-leftSide-version={{leftSideSecretVersion.version}}>
+                <button 
+                  class="link"
+                  {{on "click" (fn this.selectVersion leftSideSecretVersion.version D.actions "left")}}
+                  >
+                  Version {{leftSideSecretVersion.version}}
+                  {{#if (and (eq leftSideSecretVersion.version (or this.leftSideVersionSelected this.model.currentVersion)) (not leftSideSecretVersion.destroyed) (not leftSideSecretVersion.deleted))}}
```
hellobontempo, line 28: It looks like `.destroyed` and `.deleted` exist on the `metadata` not `data` so we don't have access to those attributes here. I don't these icons are displaying?

ui/app/templates/components/json-editor.hbs
hellobontempo: I appreciate all of your notes! 😍

ui/app/templates/components/diff-version-selector.hbs
Monkeychip: This was resolved. Discovered it was an issue in main an is being addressed.

hellobontempo (approved):
Well done! This is awesome :)

PR #13006, opened 2021-11-01
docs: agent cache config requirements 

tvoran:
Added a note that agent cache requires at least one listener or template to be defined in the config, and a couple spelling corrections.

Updated to match the changes introduced in https://github.com/hashicorp/vault/pull/12762


website/content/docs/agent/caching/index.mdx
```diff
@@ -211,11 +211,18 @@ page.
 - `when_inconsistent` `(string: optional)` - Set to one of `"fail"`, `"retry"`,
    or `"forward"`.
 
+-> **Note:** When the `cache` block is defined, either a
+[template][agent-template] or a [listener][agent-listener] must also be defined
```
tomhjp, line 215: nit suggestion:

```suggestion
-> **Note:** When the `cache` block is defined, at least one
[template][agent-template] or [listener][agent-listener] must also be defined
```

tomhjp (approved):

benashz (approved):
LGTM!

website/content/docs/agent/caching/index.mdx
tvoran: Yep, that's better, thanks.

PR #13007, opened 2021-11-01
Docs for counting non-entity tokens in the Activity Log

HridoyRoy:


website/content/docs/concepts/client-count.mdx
```diff
@@ -108,20 +107,26 @@ From a Vault client perspective, **_option 1_** is one client, the Master CI/CD
 The number of active clients using a Vault cluster is the total of:
 
 - active entities: identity entities that create a token via a login
-- active non-entity tokens: tokens created via a method that is not associated with an entity
+- active non-entity tokens: clients associated with tokens created via a method
```
swayne275, line 110:
```suggestion
- active non-entity clients: clients associated with tokens created via a method
```

website/content/docs/concepts/client-count.mdx
```diff
@@ -108,20 +107,26 @@ From a Vault client perspective, **_option 1_** is one client, the Master CI/CD
...
 Please refer to [Vault Usage Metrics](https://learn.hashicorp.com/tutorials/vault/usage-metrics) for a
 step-by-step tutorial and description of how to use the UI.
 
 ## Measuring clients
 
-Each time a token is created, Vault checks to see whether it belongs to an identity entity
+Each time a token is used, Vault checks to see whether it belongs to an identity entity
 that has already been active in the current month. New entities are added to a log in Vault storage
-periodically. New tokens without entities are added to the "non-entity token" count.
+periodically. Tokens without entities are tracked separately and added to the "non-entity token" count.
```
swayne275, line 127:
```suggestion
periodically. Tokens without entities are tracked separately and added to the "non-entity client" count.
```

website/content/docs/concepts/client-count.mdx
```diff
@@ -188,6 +193,33 @@ To reduce the number of non-entity tokens in use, consider switching to an authe
...
+be empty for requests that are untracked by the Audit Log. Such requests are unauthenticated requests / 
+login requests, requests made with the root token, and wrapped requests. The `client_id` field can also 
+contain an Entity ID for requests that are made with tokens with entities, or a unique client ID for 
+non-entitiy tokens. 
+
+Consumers of the audit log will be able to distinguish between a non-entity `client_id` and a 
+`client_id` that is also an entity IDs by either:
+A. Comparing the value of the `client_id` in the request with the value of the `entity_id` in the `Auth`
+section of the response and checking if they are the same.
+B. Checking that the format of the `client_id` is `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`(this signifies
```
swayne275, line 219:
```suggestion
B. Checking that the format of the `client_id` is `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` (this signifies 
```

website/content/docs/concepts/client-count.mdx
```diff
@@ -188,6 +193,33 @@ To reduce the number of non-entity tokens in use, consider switching to an authe
...
+login requests, requests made with the root token, and wrapped requests. The `client_id` field can also 
+contain an Entity ID for requests that are made with tokens with entities, or a unique client ID for 
+non-entitiy tokens. 
+
+Consumers of the audit log will be able to distinguish between a non-entity `client_id` and a 
+`client_id` that is also an entity IDs by either:
+A. Comparing the value of the `client_id` in the request with the value of the `entity_id` in the `Auth`
+section of the response and checking if they are the same.
+B. Checking that the format of the `client_id` is `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`(this signifies 
+that the `client_id` is an `entity_id`, versus a `client_id` which appears to be a base64 encoding
```
swayne275, line 220:
```suggestion
that the `client_id` is an `entity_id`. A `client_id` which appears to be a base64 encoding 
```

swayne275: this suggestion is a bit strange since the stuff is in parens, but it didn't read right to me as it was

swayne275 (commented):
a few minor improvements

re clients vs tokens: i'm not totally sure how we are changing the public-facing side of calling things clients vs tokens. the parts i pointed out seemed to be changed to clients based on my memory of previous PRs, but i'm not 💯 on that. it might be good to do a quick audit of the clients vs tokens verbiage before merging?

swayne275:
should this have a 1.9 milestone?

HridoyRoy:
> a few minor improvements
> 
> re clients vs tokens: i'm not totally sure how we are changing the public-facing side of calling things clients vs tokens. the parts i pointed out seemed to be changed to clients based on my memory of previous PRs, but i'm not 💯 on that. it might be good to do a quick audit of the clients vs tokens verbiage before merging?

This is a great point, and sort of confusing. We couldn't really change the client counting API user interface, so it still reports `non_entity_tokens` in the output of the partial monthly client count, for example. In the documentation I wanted to stress that the `non_entity_tokens` output counts clients that result from these `non_entity_tokens` in 1.9, not the tokens themselves.

HridoyRoy:
> should this have a 1.9 milestone?

I think so!

website/content/docs/concepts/client-count.mdx
HridoyRoy: So here I purposefully kept this as `active non-entity tokens`, the reason for this being that the API will currently display these as `non_entity_tokens` . Thus, I wanted the documentation to convey essentially that "the active non-entity tokens count maps to clients associated with tokens, as opposed to just tokens".

website/content/docs/concepts/client-count.mdx
HridoyRoy: Same comment as above -- I believe we should keep this as is.

website/content/docs/concepts/client-count.mdx
HridoyRoy: Definitely! Removed the parens to make this more clear.

website/content/api-docs/system/internal-counters.mdx
```diff
@@ -89,9 +89,11 @@ $ curl \
 
 ## Client Count
 
-This endpoint returns the number of clients per namespace, as the sum of active entities and non-entity tokens.
+This endpoint returns the number of clients per namespace, as the sum of active entities and non-entity clients.
 An "active entity" is a distinct entity that has created one or more tokens in the given time period.
-A "non-entity token" is a token with no attached entity ID.
+A "non-entity client" is one that authenticates to vault using a "non-entity token", i.e. a token with
+no attached entity ID. For more information on how non-entity clients are counted, please visit the 
+[client count](https://www.vaultproject.io/docs/concepts/client-count) concepts page.
```
ncabatoff, line 96: The link continues to use the terminology "non-entity token".  Do we have consensus for the new "non-entity client" terminology you're introducing?

website/content/docs/concepts/client-count.mdx
```diff
@@ -188,6 +193,33 @@ To reduce the number of non-entity tokens in use, consider switching to an authe
...
+Conversely, if two non-entity tokens have a different policy set or are created in different namespaces
+they will be tracked as two separate clients. 
+
+Please note that before the release of Vault 1.9, non-entity tokens were each tracked separately. That is 
+to say, two non-entity tokens would always be counted as two separate clients.
+
+## Auditing clients
+
+As of Vault 1.9, the Vault Audit Log will contain a `client_id` field in the request. The `client_id` will
+be empty for requests that are untracked by the Audit Log. Such requests are unauthenticated requests /
```
ncabatoff, line 210: This is confusing: if the request isn't tracked by the audit log, why do we need to talk about its client_id field being blank, why should I (the reader) care?

website/content/docs/concepts/client-count.mdx
```diff
@@ -188,6 +193,33 @@ To reduce the number of non-entity tokens in use, consider switching to an authe
...
+they will be tracked as two separate clients. 
+
+Please note that before the release of Vault 1.9, non-entity tokens were each tracked separately. That is 
+to say, two non-entity tokens would always be counted as two separate clients.
+
+## Auditing clients
+
+As of Vault 1.9, the Vault Audit Log will contain a `client_id` field in the request. The `client_id` will
+be empty for requests that are untracked by the Audit Log. Such requests are unauthenticated requests / 
+login requests, requests made with the root token, and wrapped requests. The `client_id` field can also
```
ncabatoff, line 211:
```suggestion
login requests, requests made with the root token, and wrapped requests. The `client_id` field will
```

website/content/docs/concepts/client-count.mdx
```diff
@@ -188,6 +193,33 @@ To reduce the number of non-entity tokens in use, consider switching to an authe
...
+Please note that before the release of Vault 1.9, non-entity tokens were each tracked separately. That is 
+to say, two non-entity tokens would always be counted as two separate clients.
+
+## Auditing clients
+
+As of Vault 1.9, the Vault Audit Log will contain a `client_id` field in the request. The `client_id` will
+be empty for requests that are untracked by the Audit Log. Such requests are unauthenticated requests / 
+login requests, requests made with the root token, and wrapped requests. The `client_id` field can also 
+contain an Entity ID for requests that are made with tokens with entities, or a unique client ID for 
+non-entitiy tokens.
```
ncabatoff, line 213:
```suggestion
non-entity tokens. 
```

website/content/docs/concepts/client-count.mdx
```diff
@@ -188,6 +193,33 @@ To reduce the number of non-entity tokens in use, consider switching to an authe
...
+be empty for requests that are untracked by the Audit Log. Such requests are unauthenticated requests / 
+login requests, requests made with the root token, and wrapped requests. The `client_id` field can also 
+contain an Entity ID for requests that are made with tokens with entities, or a unique client ID for 
+non-entitiy tokens. 
+
+Consumers of the audit log will be able to distinguish between a non-entity `client_id` and a 
+`client_id` that is also an entity IDs by either:
+A. Comparing the value of the `client_id` in the request with the value of the `entity_id` in the `Auth`
+section of the response and checking if they are the same.
+B. Checking that the format of the `client_id` is `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`. This signifies
```
ncabatoff, line 219: Do we need option B?  I'd rather they treat client_id as an opaque value rather than becoming attached to its current form.

website/content/docs/concepts/client-count.mdx
HridoyRoy: Yeah, I should reword this -- basically, if a request uses a root token or a wrapped token, the client_id (a non-entity token which won't get a client_id and be tracked in the activity log, the client_id field in the request will be "". Initially I was thinking it might be good to let the reader know the cases in which they'll see an empty client_id field. Do you think it's worth adding something in about that?

website/content/docs/concepts/client-count.mdx
HridoyRoy: Nope, option A should definitely be enough. I'll remove option B!

website/content/api-docs/system/internal-counters.mdx
HridoyRoy: Regarding this, I've sync'ed with the PM team on terminology. For 1.9, we'll keep the terminology `non-entity token` and explain that the client counting for the `non-entity token` section of the API results have essentially been updated. I'll update the docs to reflect that!

website/content/docs/concepts/client-count.mdx
```diff
@@ -188,6 +193,31 @@ To reduce the number of non-entity tokens in use, consider switching to an authe
 [AppRole](/docs/auth/approle) instead of handing out directly-created tokens. Ensure that entities and
 [entity aliases](/api/secret/identity/entity-alias) exist for all login methods used to create batch tokens.
 
+### Tracking non-entity tokens
+
+As of Vault 1.9, non-entity tokens are tracked as unique clients based on the policies the token
+has access to and the namespace in which it was created. Non-entity tokens that are assigned 
+the same set of policies and are created in the same namespace will be tracked as the same client. 
+Conversely, if two non-entity tokens have a different policy set or are created in different namespaces
```
taoism4504, line 201:
```suggestion
Conversely, if two non-entity tokens have a different policy set or are created in different namespaces,
```

website/content/docs/concepts/client-count.mdx
```diff
@@ -188,6 +193,31 @@ To reduce the number of non-entity tokens in use, consider switching to an authe
...
+As of Vault 1.9, the Vault Audit Log will contain a `client_id` field in the request. The `client_id` field
+will contain an Entity ID for requests that are made with tokens with entities, or a unique client ID for 
+non-entity tokens. 
+
+Consumers of the audit log will be able to distinguish between these two types of client IDs by comparing
+the value in the `client_id` with the value of the `entity_id` in the `Auth` section of the response. If 
+the two values are the same, then the `client_id` is an `entity_id`. If not, then the `client_id` was 
+generated from the use of a non-entity token. 
+
+An empty `client_id` field in a request means that vault did not track a client for that
```
taoism4504, line 218:
```suggestion
An empty `client_id` field in a request means that Vault did not track a client for that 
```

taoism4504 (commented):
minor nits, but LGTM.

website/content/api-docs/system/internal-counters.mdx
darshanasivakumar: Do we want to add a line here that mentions that while there is no attached entity ID to non-entity tokens, we now have a clientID attached to them?

darshanasivakumar: Do we want to add a line here that mentions that while there is no attached entity ID to non-entity tokens, we now have a clientID attached to them?

HridoyRoy: Done! Does that look ok?

website/content/api-docs/system/internal-counters.mdx
```diff
@@ -89,9 +89,13 @@ $ curl \
...
 ## Client Count
 
-This endpoint returns the number of clients per namespace, as the sum of active entities and non-entity tokens.
+This endpoint returns the number of clients per namespace, as the sum of clients calculated from the use 
+of active entities and non-entity tokens.
 An "active entity" is a distinct entity that has created one or more tokens in the given time period.
-A "non-entity token" is a token with no attached entity ID.
+A "non-entity token" is a token with no attached entity ID. 
+Both non-entity tokens and active entities have distinct client IDs. For more information on how clients
+these client IDs, and how clients are counted, please visit the
```
pmmukh, line 97: `how clients these client IDs` --> should this be something like `how clients map to these client IDs` or similar ?

website/content/docs/concepts/client-count.mdx
```diff
@@ -108,20 +107,26 @@ From a Vault client perspective, **_option 1_** is one client, the Master CI/CD
...
 step-by-step tutorial and description of how to use the UI.
 
 ## Measuring clients
 
-Each time a token is created, Vault checks to see whether it belongs to an identity entity
+Each time a token is used, Vault checks to see whether it belongs to an identity entity
 that has already been active in the current month. New entities are added to a log in Vault storage
-periodically. New tokens without entities are added to the "non-entity token" count.
+periodically. Tokens without entities are tracked separately and added to the "non-entity token" count.
+Please see the 'Tracking non-entity tokens' subsection below for a detailed explanation of how such
```
pmmukh, line 128: comment for line above actually
`non-entity token` --> `non-entity tokens` ?

pmmukh (approved):
couple nits, lgtm!

website/content/api-docs/system/internal-counters.mdx
HridoyRoy: Yes, thank you!

website/content/docs/concepts/client-count.mdx
HridoyRoy: Ah, good call!

akshya96 (approved):

website/content/api-docs/system/internal-counters.mdx
swayne275: not sure if you care, but you can also link the relative path instead of the web (e.g. `/docs/concepts/client-count`)

website/content/docs/concepts/client-count.mdx
```diff
@@ -55,7 +54,7 @@ Using the identity system allows for Vault to make sure that entities aren’t c
...
+If you chose to use the [Token Auth Method](https://www.vaultproject.io/docs/auth/token) without an identity, a non-entity token, to avoid driving up client count, always assign each token to a role and entity alias. HashiCorp recommends creating a [Token Role](https://www.vaultproject.io/api-docs/auth/token#create-update-token-role) first, with allowable entity aliases and issuing your token with the appropriate [role and entity alias name](https://www.vaultproject.io/api-docs/auth/token#create-token). This is the name that will uniquely identify the client, no matter how many tokens are issued. See more on entity alias below.
```
swayne275, line 57: `...without an identity, a non-entity token,...` is a bit confusing for me to read (not sure why the `, a non-entity token,` bit is there. could this be rephrased to be a bit more clear?

i know this was already there, but reading it now makes me think it can be improved

website/content/docs/concepts/client-count.mdx
```diff
@@ -108,20 +107,26 @@ From a Vault client perspective, **_option 1_** is one client, the Master CI/CD
...
-- active non-entity tokens: tokens created via a method that is not associated with an entity
+- active non-entity tokens: clients associated with tokens created via a method
+that is not associated with an entity
 
 Prior to Vault 1.6, this metric could only be measured from the audit log, using the
 `vault-advisor` tool. Starting with Vault 1.6, the number of clients per month, or for
 a contiguous sequence of months, can be measured by Vault itself.
 
+As of Vault 1.9, the total client count should always be measured using Vault itself. The
+metrics shown by the Vault UI should be regarded as the source of truth for this data.
```
swayne275, line 118: `are the source of truth` rather than `should be regarded as the source of truth`?

website/content/docs/concepts/client-count.mdx
```diff
@@ -188,6 +193,31 @@ To reduce the number of non-entity tokens in use, consider switching to an authe
...
+the same set of policies and are created in the same namespace will be tracked as the same client. 
+Conversely, if two non-entity tokens have a different policy set or are created in different namespaces,
+they will be tracked as two separate clients. 
+
+Please note that before the release of Vault 1.9, non-entity tokens were each tracked separately. That is 
+to say, two non-entity tokens would always be counted as two separate clients.
+
+## Auditing clients
+
+As of Vault 1.9, the Vault Audit Log will contain a `client_id` field in the request. The `client_id` field
```
swayne275, line 209: i think you can write `contains`, rather than `will contain`. here and on the next line (and if similar language is used in other places)

swayne275 (commented):
a few suggestions, but looks good

swayne275:
oh oops already merged

HridoyRoy:
> oh oops already merged

No worries, I can make these changes in a separate PR!

PR #13013, opened 2021-11-02
[VAULT-4034] Revert back to caching nil values

pmmukh:
This reverts commit 67e1ed06c7199856f8493f416df55f06603881c6.


swayne275:
why did we need to revert?

sdk/physical/cache.go
```diff
@@ -184,10 +184,8 @@ func (c *Cache) Get(ctx context.Context, key string) (*Entry, error) {
 		return nil, err
 	}
 
-	if ent != nil {
-		// Cache the result
-		c.lru.Add(key, ent)
-	}
+	// Cache the result
```
ncabatoff, line 187:
```suggestion
	// Cache the result, even if nil
```

ncabatoff (approved):
Looks good, added comment suggestion so reader will know this is intentional.

pmmukh:
@swayne275 so removing the nil caching was fixing this issue, but nil caching also solves a problem that brian discovered after some followups with jeff / research. Essentially, the nil caching ( caching that a value is absent ) was used as a performance optimization, to reduce calls to storage for missing data. We discussed this on a zoom a few minutes ago, and the plan going forward is to keep the nil caching, and let this issue be solved with cache invalidation. That cache invalidation isn't happening atm is planned to be fixed with this jira https://hashicorp.atlassian.net/browse/VAULT-4161

swayne275:
@pmmukh thanks! how might we document that this is intended behavior?

pmmukh:
@swayne275 np! so I think the suggestion nick added, should clarify that the nil caching is intentional

briankassouf (approved):

swayne275 (approved):

PR #13014, opened 2021-11-02
dev docs: clarify internal and external token helpers

mickael-hc:


ncabatoff:
Not sure this is the right - or at least, sufficient - comment fix.  The thing is that people don't usually invoke the external/internal token helper directly, they're usually calling DefaultTokenHelper and it's choosing for them.

The broader context is that token helpers are really only intended for CLI commands.  `vault server -dev` is a special case, where it's trying to save the dev user from the trouble of having to do a `vault login`.  So I kind of feel the comment should be more about emphasizing "don't use *any* token helper unless you're a non-server CLI command".

ncabatoff (approved):

aphorise:
@mladlow @calvn @briankassouf can this be merged already since it's purely in code comments and was already approved by @ncabatoff ?

PR #13015, opened 2021-11-02
Secrets header version badge

zofskeez:
Small change to always show the version badge in the secrets engine header.

image

### DIFF
--- a/ui/app/templates/components/secret-list-header.hbs
+++ b/ui/app/templates/components/secret-list-header.hbs
@@ -16,9 +16,11 @@
       <h1 class="title is-3">
         <Icon @glyph={{or @model.engineType "secrets"}} @size="xl" class="has-text-grey-light" />
         {{@model.id}}
-        <span class="tag">
-          Version {{or @model.options.version "1"}}
-        </span>
+        {{#if this.isKV}}
+          <span class="tag" data-test-kv-version-badge>
+            Version {{or @model.options.version "1"}}
+          </span>
+        {{/if}}
       </h1>
     </p.levelLeft>
   </PageHeader>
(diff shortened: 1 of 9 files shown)


ui/app/templates/components/secret-list-header.hbs
```diff
@@ -16,11 +16,9 @@
       <h1 class="title is-3">
         <Icon @glyph={{or @model.engineType "secrets"}} @size="xl" class="has-text-grey-light" />
         {{@model.id}}
-        {{#if (eq @model.options.version 2)}}
```
hashishaw, line 19: Only KV type backends should have this badge. This component is used in every type of backend, so we still need some check here to make sure we're only rendering on mount type KV

hashishaw (changes requested):
Might be worth adding a little test too

ui/app/templates/components/secret-list-header.hbs
zofskeez: Thanks for pointing that out @hashishaw. I'll update and add a test. Quick question regarding `engineType`. I see in some places there is check for both `kv` and `generic` types. Is `generic` a pre kv type from older versions that we are now mapping to kv?

hashishaw: That is a great question. Let me check and get back to you 👀

hashishaw: Yes, `generic` was "renamed" to `kv` but we keep it for backwards compatibility 👍

zofskeez:
@hashishaw I made the change to only show the badge for `kv` and `generic` engine types and added a test. I also added a mirage factory for the `secret-engine` model as well as a test helper to return ember data models when given models generated by mirage. This could be convenient in integration tests when the component expects a model as an argument rather than stubbing out a pojo.

hashishaw (approved):
Great work, thank you!

PR #13016, opened 2021-11-02
fix 12888 release note format

swayne275:
is this the proper format


mladlow:
It is - are these publicly documented anywhere? If they are, we should also update docs. If they aren't, we should probably remove this changelog file instead of editing it.

swayne275:
@mladlow only really documented for support and developers (which have been updated).

We'd discussed if this should be a changelog or not [here](https://hashicorp.slack.com/archives/CMRSLF924/p1634762179493000) - do you still think changelog should be removed?

mladlow (approved):

PR #13019, opened 2021-11-02
Docs deprecate token issuer validation

imthaghost:
Deprecation of toke issuer validation
Relating to K8s-Auth PR [127](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/127)


benashz:
I think we will also want to update the usage/example docs to mention that configuring the `issuer` is now deprecated, so the related procedures may no longer be applicable.

website/content/api-docs/auth/kubernetes.mdx
```diff
@@ -39,7 +39,7 @@ access the Kubernetes API.
   keys.
 - `issuer` `(string: "")` - Optional JWT issuer. If no issuer is specified, then this plugin will
```
benashz, line 40: May want to document this param as being deprecated.

benashz (changes requested):
We'll want to update sample/example docs as well.

website/content/api-docs/auth/kubernetes.mdx
imthaghost: Marked issuer and disable_iss_validation as deprecated.

tvoran:
I think we should also add a note to the top of the "Discovering the service account `issuer`" section on the [kubernetes auth page](https://github.com/hashicorp/vault/blame/main/website/content/docs/auth/kubernetes.mdx#L104) that mentions as of Vault 1.9, these fields are deprecated so issuer discovery isn't required.

(I would normally suggest removing that section entirely, but I don't think we're doing versioned docs yet, so the info may still be useful to folks on previous versions.)

imthaghost:
> I think we should also add a note to the top of the "Discovering the service account `issuer`" section on the [kubernetes auth page](https://github.com/hashicorp/vault/blame/main/website/content/docs/auth/kubernetes.mdx#L104) that mentions as of Vault 1.9, these fields are deprecated so issuer discovery isn't required.
> 
> (I would normally suggest removing that section entirely, but I don't think we're doing versioned docs yet, so the info may still be useful to folks on previous versions.)

I think we can take some inspiration from the upgrade docs seen [here](https://vault-hiyhukccy-hashicorp.vercel.app/docs/upgrading/upgrade-to-1.8.0). Dedicating a section for deprecation inside the Discovering the service account `issuer` section.

imthaghost:
I was also looking at the [Vault CSI Provider docs](https://www.vaultproject.io/docs/platform/k8s/csi). There is an issuer section does this also need to be changed?

website/content/api-docs/auth/kubernetes.mdx
```diff
@@ -25,6 +25,16 @@ access the Kubernetes API.
...
 
+### Deprecations
+
+The following parameters have been deprecated:
+
+* `disable_iss_validation` `(bool: true)` Disable JWT issuer validation. Allows to skip ISS validation.
+   *TODO* instead do this.
+
+* `issuer` `(string: "")` Optional JWT issuer. If no issuer is specified, then this plugin will
+   *TODO* instead do this..
```
benashz, line 36: You may want state that the configuration will be removed in a future release of Vault, and should no longer be configured etc.

tvoran:
> I was also looking at the Vault CSI Provider docs. There is an issuer section does this also need to be changed?

Good catch! Yeah I think we can update that too, probably just the same sort of note about issuer being deprecated.

imthaghost:
> > I was also looking at the Vault CSI Provider docs. There is an issuer section does this also need to be changed?
> 
> Good catch! Yeah I think we can update that too, probably just the same sort of note about issuer being deprecated.

Added an additional note to catch that as well - 94c5cae

website/content/api-docs/auth/kubernetes.mdx
```diff
@@ -25,6 +25,16 @@ access the Kubernetes API.
...
+### Deprecations
+
+The following fields have been deprecated and will be removed in a future release:
+
+* `disable_iss_validation` `(bool: true)` Disable JWT issuer validation. Allows to skip ISS validation.
+
+
+* `issuer` `(string: "")` Optional JWT issuer. If no issuer is specified, then this plugin will
+
+
```
tvoran, line 37: I'd suggest moving this section below the `Parameters` section, and calling it `Deprecated Parameters`.

website/content/docs/platform/k8s/csi/index.mdx
```diff
@@ -69,6 +69,8 @@ long-lived default service account token, with an issuer different to the tokens
 Vault CSI Provider will create. But one Kubernetes auth mount can only be
 configured to validate a single issuer value.
 
+-> **Deprecated:** The issuer has been deprecated and will be removed in a future release.
+
```
tvoran, line 73: I think I've usually seen a deprecation notice like this at the top of a section instead of the bottom. I'd also suggest that we mention the vault version, something along the lines of "The issuer parameter has been deprecated as of Vault 1.9 and will be....".

tvoran (commented):
See my previous two comments about the deprecated notice and section placement. Also I think "[Discovering the service account issuer](https://github.com/hashicorp/vault/blame/main/website/content/docs/auth/kubernetes.mdx#L104)" still needs to be updated here.

website/content/docs/platform/k8s/csi/index.mdx
imthaghost: > See my previous two comments about the deprecated notice and section placement. Also I think "[Discovering the service account issuer](https://github.com/hashicorp/vault/blame/main/website/content/docs/auth/kubernetes.mdx#L104)" still needs to be updated here.

Can this section be removed entirely?

tvoran: I'd say since these parameters haven't been removed yet we should keep the "Discovering the service account issuer" section.

website/content/docs/platform/k8s/csi/index.mdx
```diff
@@ -5,6 +5,8 @@ description: >-
   The Vault CSI Provider allows pods to consume Vault secrets using CSI volumes.
 ---
 
+-> **Deprecated:** The issuer parameter has been deprecated as of Vault 1.9 and will and will be removed in a future release.
+
```
tvoran, line 9: Should this be at the top of the page? I'm thinking it would be good at the top of the "Setting `issuer` for Kubernetes authentication" section.

website/content/docs/platform/k8s/csi/index.mdx
```diff
@@ -69,6 +71,8 @@ long-lived default service account token, with an issuer different to the tokens
 Vault CSI Provider will create. But one Kubernetes auth mount can only be
 configured to validate a single issuer value.
 
+
+
```
tvoran, line 75: Are these extra lines needed?

website/content/api-docs/auth/kubernetes.mdx
```diff
@@ -37,11 +37,18 @@ access the Kubernetes API.
...
 - `disable_local_ca_jwt` `(bool: false)` - Disable defaulting to the local CA cert and service account JWT when running in a Kubernetes pod.
 
+### Deprecated Parameters
+
+The following fields have been deprecated and will be removed in a future release:
+
+* `disable_iss_validation` `(bool: true)` Disable JWT issuer validation. Allows to skip ISS validation.
+
+
+* `issuer` `(string: "")` Optional JWT issuer. If no issuer is specified, then this plugin will
```
tvoran, line 50: It looks like the rest of the `issuer` sentences need to be moved here as well.

benashz: I feel like it should also mention that `issuer` has been deprecated and will be removed in a future release.

website/content/docs/platform/k8s/csi/index.mdx
```diff
@@ -50,6 +50,8 @@ account provided to pods if no service account is defined.
 
 ### Setting `issuer` for Kubernetes authentication
 
+-> **Deprecated:** The issuer parameter has been deprecated as of Vault 1.9 and will and will be removed in a future release.
```
benashz, line 53:
```suggestion
-> **Deprecated:** The `issuer` parameter has been deprecated as of Vault 1.9 and will be removed in a future release.
```

tvoran (approved):
Thanks!

website/content/docs/platform/k8s/csi/index.mdx
```diff
@@ -50,6 +50,8 @@ account provided to pods if no service account is defined.
 
 ### Setting `issuer` for Kubernetes authentication
 
+-> **Deprecated:** The `issuer` parameter has been deprecated as of Vault 1.9 and will be removed in a future release.
```
benashz, line 53: In line 55 we are saying that issuer more than likely needs to be configured. Maybe add something stating that there is no need to configure the issuer if running vault-1.9+

imthaghost: 1ff8749 - States that if you are running a version prior to 1.9 then you will need to configure the issuer.

benashz (approved):
LGTM!

PR #13022, opened 2021-11-02
GCP and Azure Login methods for Go client library

digivava:
Continuing with the addition of native Login methods for each auth method to the Go client library, in order to simplify the authentication process for users who may struggle to craft the login request body on their own. As before, there is one module per auth method.


api/auth/azure/azure.go
```diff
@@ -0,0 +1,204 @@
...
+	metadataParameters := metadataEndpoint.Query()
+	metadataParameters.Add(apiVersionQueryParam, metadataAPIVersion)
+	metadataEndpoint.RawQuery = metadataParameters.Encode()
+	req, err := http.NewRequest("GET", metadataEndpoint.String(), nil)
+	if err != nil {
+		return metadataJSON{}, fmt.Errorf("error creating HTTP Request for metadata endpoint: %w", err)
+	}
+	req.Header.Add("Metadata", "true")
+
+	client := &http.Client{}
```
elsesiy, line 185: Using a default http client omits setting sensible timeouts which can lead to problems if the socket is never closed, I suggest adding a timeout of e.g. 10s for all http clients

api/auth/azure/azure.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+type computeJSON struct {
+	VMName            string `json:"name"`
+	VMScaleSetName    string `json:"vmScaleSetName"`
+	SubscriptionID    string `json:"subscriptionId"`
+	ResourceGroupName string `json:"resourceGroupName"`
+}
+
+const (
```
elsesiy, line 45: This seems to be duplicating the `go-autorest` definitions provided by azure, see [here](https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go). IMO the UX is much better if the user can just provide a struct property instead of having to fiddle around with strings to construct the resource URI etc.

api/auth/azure/azure.go
```diff
@@ -0,0 +1,204 @@
...
+	defaultMountPath     = "azure"
+	defaultResourceURL   = "https://management.azure.com"
+	metadataEndpoint     = "http://169.254.169.254"
+	metadataAPIVersion   = "2018-02-01"
```
elsesiy, line 49: nit: It should be safe to use a newer api version to avoid any issues if Microsoft starts deprecating & removing older versions. List of versions can be found [here](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/instance-metadata-service?tabs=linux#supported-api-versions).

digivava: Good point, thank you!

api/auth/azure/azure.go
digivava: Thanks, nice catch!

api/auth/azure/azure.go
digivava: This is my first time working with Azure, so thank you for making me aware of this repo!

api/auth/azure/azure.go
```diff
@@ -0,0 +1,240 @@
...
+	defer resp.Body.Close()
+
+	responseBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return metadataJSON{}, fmt.Errorf("error reading response body from metadata endpoint: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		var errResp errorJSON
+		_ = json.Unmarshal(responseBytes, &errResp)
```
calvn, line 226: Is the error in here being ignored on purpose? If so, the err check right below would be checking for the error from another function (last assignment seems to be from `ioutil.ReadAll`)

api/auth/gcp/gcp.go
```diff
@@ -0,0 +1,183 @@
...
+	req.Header.Add("Metadata-Flavor", "Google")
+	q := url.Values{}
+	q.Add("audience", fmt.Sprintf("%s/vault/%s", vaultAddress, a.roleName))
+	q.Add("format", "full")
+	req.URL.RawQuery = q.Encode()
+	resp, err := c.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("error making request to metadata service: %w", err)
+	}
+
```
calvn, line 174:
```suggestion
defer resp.Body.Close()

```

api/auth/azure/azure.go
```diff
@@ -0,0 +1,240 @@
...
+
+// WithCloud allows you to specify a different cloud environment than
+// the default of Azure Public Cloud. This should match the cloud
+// environment that an administrator configured your Vault server to use.
+//
+// See https://github.com/Azure/go-autorest/blob/master/autorest/azure/environments.go
+// for a list of valid environments.
+func WithCloud(cloud azure.Environment) LoginOption {
+	return func(a *AzureAuth) error {
+		a.resource = cloud.ResourceManagerEndpoint
```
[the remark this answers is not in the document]
digivava: cc @elsesiy I've changed this back to WithResource(url string) because as Austin explained to me, this value is just used to populate the `aud` claim so it needs to match whatever URL the auth backend was configured with. (Which may not actually be an Azure URL, it just depends on what the one who configured Vault set it as.) No Azure API access is specifically given.

Users are still welcome to use the autorest constants you mentioned to populate the string though if they know theirs is an Azure one.

`resource` is admittedly a confusing field all around. I'm keeping the default as the Azure public cloud ARM endpoint just to help those users who probably configured their Vault servers with it since that's what "resource" usually means to them.

api/auth/gcp/gcp.go
```diff
@@ -0,0 +1,184 @@
+package gcp
```
austingebauer, line 1: It would be cool to eventually replace the login helper (see [cli.go](https://github.com/hashicorp/vault-plugin-auth-gcp/blob/master/plugin/cli.go)) in the plugin with this code. It'd make it quick to add support for obtaining GCE tokens from within GCE instances 👍

api/auth/gcp/gcp.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func WithMountPath(mountPath string) LoginOption {
+	return func(a *GCPAuth) error {
+		a.mountPath = mountPath
+		return nil
+	}
+}
+
+func WithIAMAuth(serviceAccountEmail string) LoginOption {
```
austingebauer, line 104: It'd be cool to also have an option to just supply a JWT. For example, you can obtain a signed JWT using gcloud or curl ([examples](https://www.vaultproject.io/docs/auth/gcp#curl-example)). Not a blocker though, can always add this later.

austingebauer (approved):
LGTM 👍

changelog/13022.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+implements Login method in Go client libraries for GCP and Azure auth methods
```
mladlow, line 2: Just a note for next time - we'd want this file to have a submodule as a prefix to this line - in this case, it would be something like: `api: Implement Login method in...`

I'll tag you as a reviewer on an ENT PR so the changelog will format properly.

digivava: Ah, sorry about that!

PR #13024, opened 2021-11-03
VAULT-1564 report in-flight requests

hghaf099:
Adding a trace capability to the request handling or HTTP layer that shows, for each requests received but not yet answered:

- time duration since the request was received
- the request RemoteAddress
- the operation and path in the request, but not payload

This would provide a point-in-time snapshot of what user requests Vault is handling, highlighting any deadlocks or abnormally long response times.

The API reporting on this information should be called as part of the debug command. It also support unauthenticated requests if a new profiling config option is set.

Adding a new metric for the total number of in-flight requests.
Adding documentation for the metric and the new endpoint.

Here is a sample result:
`{
  "7ecdd692-d934-e668-365d-b4a6664e1548": {
    "start_time": "2021-11-03T14:54:45.774893-07:00",
    "client_remote_address": "127.0.0.2:53771",
    "request_path": "/v1/secret/data/68",
    "duration": "230750 microseconds"
  }
}`


http/handler_test.go
```diff
@@ -294,6 +295,110 @@ func TestHandler_CacheControlNoStore(t *testing.T) {
...
+
+	var actual map[string]interface{}
+	testResponseStatus(t, resp, 200)
+	testResponseBody(t, resp, &actual)
+	if actual == nil {
+		t.Fatalf("")
+	}
+}
+
+func TestHandler_InFlightRequestWithLoad(t *testing.T) {
```
hghaf099, line 328: Not sure if I should keep this test!

http/handler.go
```diff
@@ -464,6 +489,9 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
 
 		h.ServeHTTP(w, r)
 
+		// deleting the in-flight request entry
+		core.DeleteInFlightReqData(inFlightReqID)
```
ncabatoff, line 493: Why not do this in a defer instead?  That way if something should cause the preceding code to return early, we're sure not to leak map entries.

http/sys_in_flight_reqeusts_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+func TestInFlightRequestUnauthenticated(t *testing.T) {
+	conf := &vault.CoreConfig{}
+	core, _, token := vault.TestCoreUnsealedWithConfig(t, conf)
+	ln, addr := TestServer(t, core)
+	TestServerAuth(t, addr, token)
+
+	// Default: Only authenticated access
+	resp := testHttpGet(t, "", addr+"/v1/sys/in-flight-req")
+	testResponseStatus(t, resp, 400)
```
ncabatoff, line 18: Shouldn't this be a 404?

http/sys_in_flight_requests.go
```diff
@@ -0,0 +1,45 @@
...
+		currentInFlightReqMap := make(map[string]interface{})
+		syncMapRangeResult := true
+		core.RangeInFlightReqData(func(key, value interface{}) bool {
+			v, ok := value.(*vault.InFlightReqData)
+			if !ok {
+				syncMapRangeResult = false
+				return false
+			}
+			// don't report the request to the in-flight-req path
+			if v.ReqPath != "/v1/sys/in-flight-req" {
```
ncabatoff, line 29: Why should we exclude these?

http/sys_in_flight_requests.go
```diff
@@ -0,0 +1,45 @@
...
+		switch r.Method {
+		case "GET":
+		default:
+			respondError(w, http.StatusMethodNotAllowed, nil)
+			return
+		}
+		now := time.Now()
+		currentInFlightReqMap := make(map[string]interface{})
+		syncMapRangeResult := true
+		core.RangeInFlightReqData(func(key, value interface{}) bool {
```
ncabatoff, line 22: Can we replace RangeInFlightReqData with a func that does this, i.e. returns a regular map based on the sync.Map's contents?

vault/core.go
```diff
@@ -2939,3 +2944,34 @@ type LicenseState struct {
...
+
+type InFlightReqData struct {
+	StartTime        time.Time `json:"start_time"`
+	ClientRemoteAddr string    `json:"client_remote_address"`
+	ReqPath          string    `json:"request_path"`
+	Duration         string    `json:"duration"`
+}
+
+func (c *Core) StoreInFlightReqData(reqID string, inFlightReqdata *InFlightReqData) error {
+	if c.inFlightReqMap == nil {
```
ncabatoff, line 2956: It shouldn't be possible for this to ever happen, right?  A core can't be created with this nil?  So let's not test for it.

vault/logical_system.go
```diff
@@ -2955,6 +2956,37 @@ func (b *SystemBackend) handleMetrics(ctx context.Context, req *logical.Request,
 	return b.Core.metricsHelper.ResponseForFormat(format), nil
 }
 
+func (b *SystemBackend) handleInFlightRequestData(_ context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
```
ncabatoff, line 2959: Both pprof and metrics return their results in a consistent format regardless of whether we're hitting the authenticated or unauthenticated endpoints.  Shouldn't we do the same?

vault/core_metrics.go
```diff
@@ -514,3 +517,13 @@ func (c *Core) cachedGaugeMetricsEmitter() {
 
 	loopMetrics.Range(emit)
 }
+
+func (c *Core) inFlightReqGaugeMetric() {
+	totalInFlightReq := 0
+	c.inFlightReqMap.Range(func(key, value interface{}) bool {
```
ncabatoff, line 523: I wonder if it's worth having a separate atomic int to track the request count, rather than having to range across the map every second?

vault/logical_system.go
```diff
@@ -2955,6 +2956,37 @@ func (b *SystemBackend) handleMetrics(ctx context.Context, req *logical.Request,
...
+	syncMapRangeResult := true
+	b.Core.inFlightReqMap.Range(func(key, value interface{}) bool {
+		v, ok := value.(*InFlightReqData)
+		if !ok {
+			syncMapRangeResult = false
+			return false
+		}
+		// don't report the request to the in-flight-req path
+		if v.ReqPath != "/v1/sys/in-flight-req" {
+			v.Duration = fmt.Sprintf("%v microseconds", now.Sub(v.StartTime).Microseconds())
```
ncabatoff, line 2971: I'm still on the fence about whether we want to have a `duration` field, or if the starttime is good enough.  But assuming we keep duration, I would rather we not modify the struct pointed to by the sync.Map value.  For one thing, if we have two simultaneous requests to this endpoint, we could have a data race as both attempt to populate this field.  But also just on principle, it feels wrong to set that field except in this handler's response: in the sync.Map the request is still in flight, so it doesn't have a duration yet.

hghaf099: I see. I am wondering why a data race could happen considering we are using sync.Map. I think there should be some internal locking in place. Though, I do see I should not have changed the struct pointed to by the sync.Map, will fix. 
We could change the "duration" name to something more meaningful though, like "timeSinceStarted" or even the time at which we are taking the snapshot of the in-flight requests like "snapshot_time" ( I am bad with choosing field/variable names). Something like snapshot_time would be a convenient as a user won't have to consider the communication overhead time. This will come in handy for debugging specially when a lot of hops exist between the client and the server.

http/sys_in_flight_requests.go
hghaf099: Sounds good!

command/debug.go
```diff
@@ -880,6 +894,57 @@ func (c *DebugCommand) collectServerStatus(ctx context.Context) {
...
+			}
+		}
+
+		c.logger.Info("capturing in-flight request status", "count", idxCount)
+		idxCount++
+
+		req := c.cachedClient.NewRequest("GET", "/v1/sys/in-flight-req")
+		resp, err := c.cachedClient.RawRequestWithContext(ctx, req)
+		if err != nil {
+			c.captureError("inFlightReq-status", err)
```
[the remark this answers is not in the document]
ncabatoff: I think if we return an error, that will interrupt other funcs, but not if we return nil.

http/handler.go
```diff
@@ -389,6 +395,25 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
 	hostname, _ := os.Hostname()
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		// The uuid for the request is going to be generated when a logical
+		// request is generated. But, here we generate one to be able to track
+		// in-flight requests
+		inFlightReqID, err := uuid.GenerateUUID()
+		if err != nil {
+			respondError(w, http.StatusBadRequest, fmt.Errorf("failed to generate an identifier for the in-flight request"))
```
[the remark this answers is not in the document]
ncabatoff: Sure.

http/sys_in_flight_reqeusts_test.go
ncabatoff: Whoops, ignore me :)

http/sys_in_flight_requests.go
ncabatoff: I'd rather include everything.  It would be ironic if a problem caused by this endpoint couldn't be debugged because we filtered out the relevant requests.

vault/logical_system.go
ncabatoff: No, to the different response format, i.e. the use of the "data" field.

vault/logical_system.go
ncabatoff: sync.Map will protect you from races involving access to the map, but once you've got the value for an entry, any concurrent access to the record it is pointing to must be protected.

command/debug.go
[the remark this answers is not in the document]
hghaf099: Done

http/sys_in_flight_requests.go
hghaf099: Done

command/debug.go
```diff
@@ -865,10 +884,12 @@ func (c *DebugCommand) collectServerStatus(ctx context.Context) {
 		healthInfo, err := c.cachedClient.Sys().Health()
 		if err != nil {
 			c.captureError("server-status.health", err)
+			return
```
ncabatoff, line 887: Although it's true that an error on the health endpoint will probably mean one on the seal-status endpoint too, I'd rather we didn't assume that.  Can you revert this change please?

command/debug.go
```diff
@@ -865,10 +884,12 @@ func (c *DebugCommand) collectServerStatus(ctx context.Context) {
...
 		}
 		sealInfo, err := c.cachedClient.Sys().SealStatus()
 		if err != nil {
 			c.captureError("server-status.seal", err)
+			return
```
ncabatoff, line 892: Here too.

command/debug.go
```diff
@@ -880,6 +901,50 @@ func (c *DebugCommand) collectServerStatus(ctx context.Context) {
...
+			c.captureError("inFlightReq-status", err)
+			return
+		}
+
+		var data map[string]interface{}
+		if resp != nil {
+			defer resp.Body.Close()
+			err = jsonutil.DecodeJSONFromReader(resp.Body, &data)
+			if err != nil {
+				c.captureError("inFlightReq-status", err)
```
ncabatoff, line 933: It looks like the first arg to captureError is the target, which we're now calling `requests` right?

command/debug.go
```diff
@@ -880,6 +901,50 @@ func (c *DebugCommand) collectServerStatus(ctx context.Context) {
...
+				return
+			}
+
+			if data != nil && len(data) > 0 {
```
ncabatoff, line 937: I don't think we need this conditional do we?

http/handler.go
```diff
@@ -389,6 +395,27 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
...
+			respondError(w, http.StatusInternalServerError, fmt.Errorf("failed to generate an identifier for the in-flight request"))
+		}
+		core.StoreInFlightReqData(
+			inFlightReqID,
+			&vault.InFlightReqData {
+				StartTime: time.Now(),
+				ClientRemoteAddr: r.RemoteAddr,
+				ReqPath: r.URL.Path,
+			})
+		if err != nil {
```
ncabatoff, line 413: What error are we handling here?

vault/core.go
```diff
@@ -2939,3 +2947,49 @@ type LicenseState struct {
...
+	c.inFlightReqData.InFlightReqCount.Dec()
+}
+
+func (c *Core) GetInFlightReqData() (map[string]*InFlightReqData, error) {
+
+	currentInFlightReqMap := make(map[string]*InFlightReqData)
+	syncMapRangeResult := true
+	c.inFlightReqData.InFlightReqMap.Range(func(key, value interface{}) bool {
+		v, ok := value.(*InFlightReqData)
+		if !ok {
```
ncabatoff, line 2980: Is this a realistic failure scenario?  I approve of defensive programming, but given that there's only one writer to this map maybe we don't need to be so cautious here?

command/debug.go
hghaf099: yeah, there should always be at least one entry which is the /v1/sys/in-flight-req related info

http/handler.go
hghaf099: Ah, thanks for catching this!

website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -183,6 +183,8 @@ default value in the `"/sys/config/ui"` [API endpoint](/api/system/config-ui).
 
 - `unauthenticated_pprof_access` `(bool: false)` - If set to true, allows
   unauthenticated access to the `/v1/sys/pprof` endpoint.
+- `unauthenticated_in_flight_request_access` `(bool: false)` - If set to true, allows
+  unauthenticated access to the `/v1/sys/in-flight-req` endpoint.
```
mladlow, line 187: Should we add an API docs page for this endpoint?

hghaf099: Yes, I am going to add that in this PR or in a separate one

sdk/logical/response.go
```diff
@@ -202,9 +208,19 @@ type HTTPResponseWriter struct {
 // NewHTTPResponseWriter creates a new HTTPResponseWriter object that wraps the
 // provided io.Writer.
 func NewHTTPResponseWriter(w http.ResponseWriter) *HTTPResponseWriter {
-	return &HTTPResponseWriter{
-		ResponseWriter: w,
-		written:        new(uint32),
+	// FIXME: this is a band aid and not a good solution, just put it here
```
hghaf099, line 211: I have a draft PR to address this issue in a better way here
https://github.com/hashicorp/vault/pull/13200

command/server/config.go
```diff
@@ -77,6 +77,9 @@ type Config struct {
 	EnableResponseHeaderHostname    bool        `hcl:"-"`
 	EnableResponseHeaderHostnameRaw interface{} `hcl:"enable_response_header_hostname"`
 
+	EnableLogRequests    bool        `hcl:"-"`
```
ncabatoff, line 80: Instead of a boolean, what do you think about something more general?  For example, we could make this be a string argument that specifies the log level at which to log requests.  Or we could make it be something like `log_request_info = "basic"`, which would leave us room to extend this feature later without needing a new config option.

hghaf099: Sounds great! I will enable logging in all levels which will depend on this configuration. 
So, do you mean that `basic` is going to be the default value for this variable? and in which log level should `basic` be resolved?

vault/core.go
```diff
@@ -2946,3 +2964,102 @@ type LicenseState struct {
...
+		c.Logger().Trace("failed to retrieve request with ID %v", reqID)
+		return
+	}
+	reqData, _ := v.(*InFlightReqData)
+	reqData.ClientID = clientID
+	c.inFlightReqData.InFlightReqMap.Store(reqID, reqData)
+}
+
+func (c *Core) logRequests(msg string, args ...interface{}) {
+	switch c.logRequestsInfo {
```
ncabatoff, line 3027: See instead Logger.Log()

vault/request_handling.go
```diff
@@ -772,6 +777,17 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
 		return nil, nil, ctErr
 	}
 
+	// Updating in-flight request data with client/entity ID
+	inFlightReqID, ok := ctx.Value("in-flight-reqID").(string)
+	if ok {
+		switch {
+		case req.ClientID != "":
+			c.UpdateInFlightReqData(inFlightReqID, req.ClientID)
+		case req.EntityID != "":
```
ncabatoff, line 786: I think if there's an entityid it'll be present in req.ClientID too, so you shouldn't need this switch.

vault/request_handling.go
```diff
@@ -772,6 +777,17 @@ func (c *Core) handleRequest(ctx context.Context, req *logical.Request) (retResp
 		return nil, nil, ctErr
 	}
 
+	// Updating in-flight request data with client/entity ID
```
ncabatoff, line 780: What about the handleLoginRequest case?

vault/request_handling.go
```diff
@@ -439,7 +439,12 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
 		return nil, fmt.Errorf("could not parse namespace from http context: %w", err)
 	}
 
-	resp, err = c.handleCancelableRequest(namespace.ContextWithNamespace(ctx, ns), req)
+	ctx = namespace.ContextWithNamespace(ctx, ns)
+	inFlightReqID, ok := httpCtx.Value("in-flight-reqID").(string)
```
ncabatoff, line 443: There's a nice pattern I like to use for WithValue where you define an empty type to use as the key to ensure conflicts can't occur, see `ctxKeyForwardedRequestMountAccessor` for an example.

vault/logical_system.go
```diff
@@ -3010,6 +3011,29 @@ func (b *SystemBackend) handleMetrics(ctx context.Context, req *logical.Request,
...
+	resp := &logical.Response{
+		Data: map[string]interface{}{
+			logical.HTTPContentType: "text/plain",
+			logical.HTTPStatusCode:  http.StatusInternalServerError,
+		},
+	}
+
+	currentInFlightReqMap := b.Core.LoadInFlightReqData()
+
+	content, err := json.Marshal(currentInFlightReqMap)
```
ncabatoff, line 3024: Again, this looks like respondOk().

vault/core.go
```diff
@@ -2946,3 +2964,102 @@ type LicenseState struct {
...
+
+func (c *Core) ReloadLogRequestsInfo(){
+	conf := c.rawConfig.Load()
+	if conf == nil {
+		return
+	}
+	infoLevel := conf.(*server.Config).LogRequestsInfo
+	switch {
+	case log.LevelFromString(infoLevel) > 0:
+		c.logRequestsInfo = infoLevel
```
ncabatoff, line 3061: This needs synchronization to avoid a race.

ncabatoff (commented):
It's looking close!

http/sys_in_flight_reqeusts_test.go
```diff
@@ -0,0 +1,46 @@
+package http
```
ncabatoff, line 1: There's a typo in the filename ("reqeusts")

http/handler_test.go
```diff
@@ -294,6 +295,103 @@ func TestHandler_CacheControlNoStore(t *testing.T) {
...
+	}
+
+	if resp == nil {
+		t.Fatalf("nil response")
+	}
+
+	var actual map[string]interface{}
+	testResponseStatus(t, resp, 200)
+	testResponseBody(t, resp, &actual)
+	if actual == nil {
```
ncabatoff, line 323: Can we unmarshall the response into a map[string]interface{} and ensure that the map is non-empty?

http/handler_test.go
ncabatoff: I'd probably get rid of it if you address my comment above re making sure TestHandler_InFlightRequest returns a non-empty map.

website/content/api-docs/system/in-flight-req.mdx
```diff
@@ -0,0 +1,41 @@
...
+### Sample Response
+
+```json
+{
+  "9049326b-ceed-1033-c099-96c5cc97db1f": {
+    "start_time": "2021-11-19T09:13:01.34157-08:00",
+    "client_remote_address": "127.0.0.3:49816",
+    "request_path": "/v1/sys/in-flight-req",
+    "request_method": "GET",
+    "client_id": "",
```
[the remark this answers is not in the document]
hghaf099: I just checked the code, and it seems that the ClientID is only produced if several conditions are met:
1- activity log is enabled
2- if the token entry is not a wrapping token. if it is ClientID is ""
3- and if the token entry is a root token, Client ID is "" 

The function that handles creating the ClientID seems to be trying to add an entry to the logs if the above conditions are met. So, would you think that the ClientID could be created and passed into that function? or they two are tightly coupled? I am going to read that part of the code though to understand more.

http/handler.go
```diff
@@ -368,6 +376,43 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
...
+		// adding an entry to the context to enable updating in-flight
+		// data with ClientID in the logical layer
+		r = r.WithContext(context.WithValue(r.Context(), logical.CtxKeyInFlightRequestID{}, inFlightReqID))
+
+		// extracting the client address to be included in the in-flight request
+		var clientAddr string
+		headers := r.Header[textproto.CanonicalMIMEHeaderKey("X-Forwarded-For")]
+		if len(headers) == 0 {
+			clientAddr = r.RemoteAddr
+		}else {
```
ncabatoff, line 395:
```suggestion
		} else {
```

http/handler_test.go
ncabatoff: Not exactly - with the new code below, if an empty map is returned, we'll iterate zero times over it and no assertions will fail.

vault/core.go
```diff
@@ -2946,3 +2964,87 @@ type LicenseState struct {
...
+	reqData, _ := v.(*InFlightReqData)
+	reqData.ClientID = clientID
+	c.inFlightReqData.InFlightReqMap.Store(reqID, reqData)
+}
+
+// LogCompletedRequests Logs the completed request to the server logs
+func (c *Core) LogCompletedRequests(reqID string, statusCode int) {
+	v, ok := c.inFlightReqData.InFlightReqMap.Load(reqID)
+	if !ok {
+		c.logger.Log(log.LevelFromString(c.logRequestsInfo), fmt.Sprintf("failed to retrieve request with ID %v", reqID))
```
ncabatoff, line 3030: Rather than call LevelFromString for each log statement, can we parse it just once instead?

vault/core.go
ncabatoff: I think you're underestimating what can happen with a data race.  It's not just a matter of "oh, we might see an incorrect value".  As far as I understand it, concurrent access without synchronization produces *undefined behaviour*, including memory corruption.  Depending on the platform, the reader could see a partial write, e.g. only half of a longword pointer could be written when the read is done, resulting in an invalid address and a panic.  These are rare and unlikely outcomes, but they could happen.  So data races should be treated as potential panics as far as I'm concerned.

I could also see us writing a test for the reload behaviour, and then we'd have the potential for the race tests failing due to the race, which would also be bad.  So yes, please add a lock or make logRequestsInfo an atomic.

vault/core_metrics.go
```diff
@@ -534,3 +537,9 @@ func (c *Core) cachedGaugeMetricsEmitter() {
 
 	loopMetrics.Range(emit)
 }
+
+func (c *Core) inFlightReqGaugeMetric() {
+	totalInFlightReq := c.inFlightReqData.InFlightReqCount.Load()
+	// Adding a gauge metric to capture total number of inflight requests
+	c.metricSink.SetGaugeWithLabels([]string{"core", "in_flight_request"}, float32(totalInFlightReq), nil)
```
ncabatoff, line 544: Looks like you updated the telemetry docs but the not the metric itself.

vault/logical_system.go
ncabatoff: Fair.

website/content/api-docs/system/in-flight-req.mdx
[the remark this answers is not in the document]
hghaf099: Done

vault/request_handling.go
```diff
@@ -398,11 +405,53 @@ func (c *Core) checkToken(ctx context.Context, req *logical.Request, unauth bool
 
 	// If it is an authenticated ( i.e with vault token ) request, increment client count
 	if !unauth && c.activityLog != nil {
-		req.ClientID = c.activityLog.HandleTokenUsage(te)
+		c.activityLog.HandleTokenUsage(te, clientID, isTWE)
 	}
 	return auth, te, nil
 }
 
+// CreateClientID returns the client ID, and a boolean which is false if the clientID
```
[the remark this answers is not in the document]
ncabatoff: Might even want to make it a method of TokenEntry, since the only thing it's using Core for is to log something that will never happen.

vault/request_handling.go
```diff
@@ -439,7 +488,12 @@ func (c *Core) switchedLockHandleRequest(httpCtx context.Context, req *logical.R
 		return nil, fmt.Errorf("could not parse namespace from http context: %w", err)
 	}
 
-	resp, err = c.handleCancelableRequest(namespace.ContextWithNamespace(ctx, ns), req)
+	ctx = namespace.ContextWithNamespace(ctx, ns)
+	inFlightReqID, ok := httpCtx.Value(logical.CtxKeyInFlightRequestID{}).(string)
+	if ok {
```
ncabatoff, line 493: Can't we always assume this context will be there, and treat its absence as an error?

vault/core.go
```diff
@@ -386,6 +389,9 @@ type Core struct {
 	// disabled
 	physicalCache physical.ToggleablePurgemonster
 
+	// logRequestsInfo indicates at which level requests should be logged
+	logRequestsInfo uberAtomic.String
```
ncabatoff, line 393: Why a string, making us have to call LevelFromString all the time?  Why not store the int result instead?

vault/request_handling.go
ncabatoff: Ah, fair enough.

command/server/config.go
```diff
@@ -77,6 +78,9 @@ type Config struct {
 	EnableResponseHeaderHostname    bool        `hcl:"-"`
 	EnableResponseHeaderHostnameRaw interface{} `hcl:"enable_response_header_hostname"`
 
+	LogRequestsInfo    log.Level   `hcl:"-"`
```
ncabatoff, line 81: It feels like making this field log.Level might be more trouble than it's worth.  I was just thinking that the Core field (that's used all the time) should be log.Level.

ncabatoff: Also, maybe the name should be LogRequestsLevel (and the HCL field `log_requests_level`)?

hghaf099: Since the Core field would need protection for synchronization issues, how about I keep that as *uberatomic.Int32.

ncabatoff: Yeah, sorry, that's what I meant.

vault/core.go
```diff
@@ -1027,6 +1038,15 @@ func NewCore(conf *CoreConfig) (*Core, error) {
 		c.customListenerHeader.Store(([]*ListenerCustomHeaders)(nil))
 	}
 
+	logRequestsLevel := log.LevelFromString(conf.RawConfig.LogRequestsLevel)
+	c.logRequestsLevel = uberAtomic.NewInt32(0)
+	switch {
+	case logRequestsLevel > log.NoLevel && logRequestsLevel < log.Off:
+		c.logRequestsLevel.Store(int32(logRequestsLevel))
+	default:
+		c.logger.Warn("invalid log_requests_level", "level", conf.RawConfig.LogRequestsLevel)
```
ncabatoff, line 1047: I think this means that if log_requests_level is unspecified, we'll log a warning, which I'd prefer to avoid.

vault/core.go
```diff
@@ -2947,6 +2967,104 @@ type LicenseState struct {
 	Terminated bool
 }
 
+type InFlightRequests struct {
+	l                sync.RWMutex
```
ncabatoff, line 2971: I'd like us to avoid having both a mutex and sync.Map; if we're going to use a mutex to guard the map, we may as well just use a regular map, since it's way cheaper than sync.Map if concurrent access is protected otherwise.

That said, I think you could get rid of the lock if you stored `InFlightReqData` instead of `*InflightReqData`.  The race occurs when we try to update data referenced by a pointer while another goroutine is trying to load the data.  If instead updating consists of copying the data out of the map, updating that copy, then storing a copy back into the map, I don't think it'll be racy.

vault/testing.go
```diff
@@ -130,6 +130,7 @@ func TestCoreWithSeal(t testing.T, testSeal Seal, enableRaw bool) *Core {
 
 func TestCoreWithCustomResponseHeaderAndUI(t testing.T, CustomResponseHeaders map[string]map[string]string, enableUI bool) (*Core, [][]byte, string) {
 	confRaw := &server.Config{
+		LogRequestsLevel: "basic",
```
ncabatoff, line 133: This isn't a valid log level, right?

ncabatoff: But we probably shouldn't do that kind of thing outside of a test specific to the feature.  Unless it's somehow relevant to custom response header behaviour?

vault/core.go
```diff
@@ -2947,6 +2967,94 @@ type LicenseState struct {
...
+	c.inFlightReqData.InFlightReqCount.Dec()
+}
+
+// LoadInFlightReqData creates a snapshot map of the current
+// in-flight requests
+func (c *Core) LoadInFlightReqData() map[string]InFlightReqData {
+	currentInFlightReqMap := make(map[string]InFlightReqData)
+	c.inFlightReqData.InFlightReqMap.Range(func(key, value interface{}) bool {
+		// there is only one writer to this map, so skip checking for errors
+		v, _ := value.(InFlightReqData)
```
ncabatoff, line 3007: Minor nit, but: by using the two-argument form, you're deliberately ignoring errors.  Either we don't believe errors are possible, in which case you can use the single-argument form which panics on error, or we do, in which case we should handle the error.

hghaf099: I think at some point, you mentioned that there is only one writer for this and there is no need to check if the assertion worked or not. The comment I have on line 3006 talks about that. But, happy to add the error check for this.

vault/testing.go
hghaf099: hmm, make sense.

vault/core.go
ncabatoff: I'm ok with there being no error check, I'm just saying if we're not going to check for errors because we think they can't happen, why not use the single-argument form?  If we're right that errors can't happen, then they're equivalent.  If we're wrong, we'll learn of it because of panics.

hghaf099: Sounds good! I think it is highly unlikely that a panic happens. I am going to use the single-arguments form.

ncabatoff (approved):

PR #13031, opened 2021-11-03
Update README to remove IRC reference

hashicorp-peter:
The README references IRC (Freenode) as a means of communication regarding the Vault project, but it seems that:

1. Freenode has had its share of issues (https://en.wikipedia.org/wiki/Freenode#Ownership_change_and_conflict)
2. You now need a Freenode account to access their IRC server
3. The channel hasn't been very active (and to the best of my knowledge hasn't been migrated to Libera where a lot of Freenode based projects moved to)

I'd recommend just removing the reference for now, and if things change or another means of collaboration surfaces - adding that later.


mladlow (changes requested):
The change sounds good to me, but would you mind removing the changelog entry? Thank you for thinking of it but I'd prefer not to have one for README changes, and we can skip the failing build check with a label.

mladlow (approved):

Issue #12962, opened 2021-10-29
No Pagination Bar for Ldap Group Page


junchytata:
**Describe the bug**
I created more than 120 ldap groups to the LDAP Auth Method.
But I only see 100 groups in the UI page.
There are no pagination bar for the next page.

**To Reproduce**
1.Login vault
2.Click [Access] from Menu Bar
3.Click Auth Methods [ldap]
4.In the [groups] tag only 100 groups can be found.
5.No pagination bar for next page.
Page url likes this: [https://myvault.example.com/ui/vault/access/ldap/item/group]

**Expected behavior**
A pagination bar to see the next page is displayed.

**Environment**
* Vault Server 1.8.1(Docker)
Image from [https://gallery.ecr.aws/hashicorp/vault]
* Chrome 94.0.4606.54

**Additional context**
1. I used vault-cli to exec command [vault list auth/ldap/groups]
and there is no problem to list all groups.

2. Add parameter "page=2" to the end of URL
like this [https://myvault.example.com/ui/vault/access/ldap/item/group?page=2]
can display the next page.

PR #13032, opened 2021-11-03
UI/Truncate long secret names

hellobontempo:
Fixes overflow problem of long secret key values mentioned in #12962

**Previous behavior:**
image

**Current behavior:**

truncate-long-labels

### DIFF
--- a/ui/lib/core/addon/components/bar-chart.js
+++ b/ui/lib/core/addon/components/bar-chart.js
@@ -19,7 +19,7 @@
  * @param {string} title - title of the chart
  * @param {array} mapLegend - array of objects with key names 'key' and 'label' for the map legend
  * @param {object} dataset - dataset for the chart
- * @param {array} tooltipData - misc. information needed to display tooltip (i.e. total clients from query params)
+ * @param {any} tooltipData - misc. information needed to display tooltip (i.e. total clients from query params)
  * @param {string} [description] - description of the chart
  * @param {string} [labelKey=label] - labelKey is the key name in the dataset passed in that corresponds to the value labeling the y-axis
  * @param {function} [onClick] - takes function from parent and passes it to click event on data bars
--- a/ui/lib/core/addon/components/info-table-row.js
+++ b/ui/lib/core/addon/components/info-table-row.js
@@ -42,6 +42,7 @@ export default Component.extend({
   tooltipText: '',
   isTooltipCopyable: false,
   defaultShown: '',
+  hasLabelOverflow: false,
 
   valueIsBoolean: computed('value', function() {
     return typeOf(this.value) === 'boolean';
@@ -70,6 +71,7 @@ export default Component.extend({
     const labelText = this.element.querySelector('.is-label');
     if (labelText.offsetWidth > labelDiv.offsetWidth) {
       labelDiv.classList.add('label-overflow');
+      this.set('hasLabelOverflow', true);
     }
   },
 });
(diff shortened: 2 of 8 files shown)


ivana-mcconnell:
👍 The tooltip behaves as expected, but is it possible to have the pointer centered on the name of the key, rather than at the end of the name?

ui/lib/core/stories/info-table-row.stories.js
```diff
@@ -16,14 +16,18 @@ storiesOf('InfoTable/InfoTableRow', module)
...
+        @isTooltipCopyable={{isTooltipCopyable}} />
     `,
       context: {
         label: text('Label', 'TTL'),
         value: text('Value', '30m (hover to see the tooltip!)'),
         helperText: text('helperText', 'This is helperText - for a short description'),
         alwaysRender: boolean('Always render?', false),
+        defaultShown: text('Default value', 'Some default value'),
         tooltipText: text('tooltipText', 'This is tooltipText'),
+        isTooltipCopyable: boolean('Allow tooltip to be copied', true),
```
zofskeez, line 30: Thanks for adding this for me!

ui/lib/core/addon/components/info-table-row.js
```diff
@@ -63,4 +64,16 @@ export default Component.extend({
...
+
+  didInsertElement() {
+    this._super(...arguments);
+    const labelDiv = this.element.querySelector('div');
+    const labelText = this.element.querySelector('.is-label');
+    if (labelDiv && labelText) {
+      if (labelText.offsetWidth > labelDiv.offsetWidth) {
+        labelDiv.classList.add('label-overflow');
+        this.set('hasLabelOverflow', true);
+      }
```
zofskeez, line 76: Nice work! `didInsertElement` is a good place for this. I'm wondering if it is worthwhile to add a test to check if the `label-overflow` class is applied to the element and that the tooltip appears for long labels?

ui/lib/core/addon/components/bar-chart.js
```diff
@@ -19,7 +19,7 @@
  * @param {string} title - title of the chart
  * @param {array} mapLegend - array of objects with key names 'key' and 'label' for the map legend
  * @param {object} dataset - dataset for the chart
- * @param {array} tooltipData - misc. information needed to display tooltip (i.e. total clients from query params)
+ * @param {any} tooltipData - misc. information needed to display tooltip (i.e. total clients from query params)
```
Monkeychip, line 22: I'm not sure I've seen every before. Does it come in as an object, or varies? Maybe describe / give example of types in the description?

Monkeychip (approved):
Nice. At first I was wondering about a test, but then I'm not sure if it's super relevant. I'll leave it up to you.

zofskeez (commented):
Nice work and thanks again for adding the _isTooltipCopyable_ property to the InfoTableRow story that I forgot!

ui/lib/core/addon/components/bar-chart.js
hellobontempo: @Monkeychip I added it after seeing it somewhere else...but yeah. Going to open a separate PR that will address some tweaks this file needs.

ui/lib/core/addon/components/info-table-row.js
hellobontempo: Thanks @zofskeez! I'd gone back and forth on a test..but I think that you're right to suggest it- this component shows up so many places it's probably worthwhile to include. I'll add!

PR #12975, opened 2021-10-29
deps: update plugin versions for 1.9 release

calvn:
This PR updates dependencies for Vault plugins that are bundled into vault.

Note:
The `github.com/pkg/browser` indirect dependency that was initially resolving to was leading to build errors for certain osarch builds. I've gone and updated it to point to master (681adbf) which seems to have resolved the build issue locally.

```sh
$ XC_OSARCH=netbsd/386 make dev
==> Checking that build is using go version >= 1.17.2...
==> Using go version 1.17.2...
==> Removing old directory...
==> Building...
Number of parallel builds: 15

-->      netbsd/386: github.com/hashicorp/vault

1 errors occurred:
--> netbsd/386 error: exit status 2
Stderr: # github.com/pkg/browser
../../go/pkg/mod/github.com/pkg/browser@v0.0.0-20210706143420-7d21f8c997e2/browser_unsupported.go:7:2: imported and not used: "os/exec"

make: *** [dev] Error 1

$ go mod why github.com/pkg/browser
# github.com/pkg/browser
github.com/hashicorp/vault/helper/builtinplugins
github.com/hashicorp/vault-plugin-database-snowflake
github.com/snowflakedb/gosnowflake
github.com/pkg/browser

$ go get github.com/pkg/browser@master
go: downloading github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
go get: upgraded github.com/pkg/browser v0.0.0-20210706143420-7d21f8c997e2 => v0.0.0-20210911075715-681adbf594b8

$ XC_OSARCH=netbsd/386 make dev
==> Checking that build is using go version >= 1.17.2...
==> Using go version 1.17.2...
==> Removing old directory...
==> Building...
Number of parallel builds: 15

-->      netbsd/386: github.com/hashicorp/vault

==> Results:
total 323136
-rwxr-xr-x  1 calvin  staff   158M Oct 29 18:07 vault

```

PR #13034, opened 2021-11-03
secrets/azure: add changelog for rotate-root

jasonodonnell:
This adds a missing changelog in 1.9 for Azure secrets rotate root: https://github.com/hashicorp/vault-plugin-secrets-azure/pull/70.

The azure secrets plugin was updated in the PR to Vault: https://github.com/hashicorp/vault/pull/12975.


changelog/13034.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
tvoran, line 1: I wonder if the changelog file should be named for [the PR](https://github.com/hashicorp/vault/pull/12975) that updated the dep? i.e. `12975.txt`

calvn: Ah yes, it'd be good to have that reference. Would it fail the CI check though since it won't match to this PR's number?

jasonodonnell: Referenced in the PR, keeping it the same due to CI failures.

tvoran (approved):

PR #13041, opened 2021-11-04
Support configuring Vault agent TLS min/max version

evan361425:
It is useful to some environment has limited TLS version and test different TLS version performance.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -24,6 +24,7 @@ import (
 	retryablehttp "github.com/hashicorp/go-retryablehttp"
 	rootcerts "github.com/hashicorp/go-rootcerts"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
+	"github.com/hashicorp/go-secure-stdlib/tlsutil"
 	"golang.org/x/net/http2"
 	"golang.org/x/time/rate"
 
@@ -43,6 +44,7 @@ const (
 	EnvVaultSRVLookup     = "VAULT_SRV_LOOKUP"
 	EnvVaultSkipVerify    = "VAULT_SKIP_VERIFY"
 	EnvVaultNamespace     = "VAULT_NAMESPACE"
+	EnvVaultTLSMinVersion = "VAULT_TLS_MIN_VERSION"
 	EnvVaultTLSServerName = "VAULT_TLS_SERVER_NAME"
 	EnvVaultWrapTTL       = "VAULT_WRAP_TTL"
 	EnvVaultMaxRetries    = "VAULT_MAX_RETRIES"
@@ -166,6 +168,9 @@ type TLSConfig struct {
 	// ClientKey is the path to the private key for Vault communication
 	ClientKey string
 
+	// TLSMinVersion specifies the minimum supported version of TLS.
+	TLSMinVersion string
+
 	// TLSServerName, if set, is used to set the SNI host when connecting via
 	// TLS.
 	TLSServerName string
@@ -268,6 +273,10 @@ func (c *Config) ConfigureTLS(t *TLSConfig) error {
 		}
 	}
 
+	if t.TLSMinVersion != "" {
+		clientTLSConfig.MinVersion = tlsutil.TLSLookup[t.TLSMinVersion]
+	}
+
 	if t.TLSServerName != "" {
 		clientTLSConfig.ServerName = t.TLSServerName
 	}
@@ -286,6 +295,7 @@ func (c *Config) ReadEnvironment() error {
 	var envClientKey string
 	var envClientTimeout time.Duration
 	var envInsecure bool
+	var envTLSMinVersion string
 	var envTLSServerName string
 	var envMaxRetries *uint64
 	var envSRVLookup bool
@@ -355,6 +365,9 @@ func (c *Config) ReadEnvironment() error {
 		}
 	}
 
+	if v := os.Getenv(EnvVaultTLSMinVersion); v != "" {
+		envTLSMinVersion = v
+	}
 	if v := os.Getenv(EnvVaultTLSServerName); v != "" {
 		envTLSServerName = v
 	}
@@ -370,6 +383,7 @@ func (c *Config) ReadEnvironment() error {
 		ClientCert:    envClientCert,
 		ClientKey:     envClientKey,
 		TLSServerName: envTLSServerName,
+		TLSMinVersion: envTLSMinVersion,
 		Insecure:      envInsecure,
 	}
 
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -3,6 +3,7 @@ package api
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
 	"fmt"
@@ -267,18 +268,21 @@ func TestClientEnvSettings(t *testing.T) {
 	oldClientCert := os.Getenv(EnvVaultClientCert)
 	oldClientKey := os.Getenv(EnvVaultClientKey)
 	oldSkipVerify := os.Getenv(EnvVaultSkipVerify)
+	oldTLSMinVersion := os.Getenv(EnvVaultTLSMinVersion)
 	oldMaxRetries := os.Getenv(EnvVaultMaxRetries)
 	os.Setenv(EnvVaultCACert, cwd+"/test-fixtures/keys/cert.pem")
 	os.Setenv(EnvVaultCAPath, cwd+"/test-fixtures/keys")
 	os.Setenv(EnvVaultClientCert, cwd+"/test-fixtures/keys/cert.pem")
 	os.Setenv(EnvVaultClientKey, cwd+"/test-fixtures/keys/key.pem")
 	os.Setenv(EnvVaultSkipVerify, "true")
+	os.Setenv(EnvVaultTLSMinVersion, "tls13")
 	os.Setenv(EnvVaultMaxRetries, "5")
 	defer os.Setenv(EnvVaultCACert, oldCACert)
 	defer os.Setenv(EnvVaultCAPath, oldCAPath)
 	defer os.Setenv(EnvVaultClientCert, oldClientCert)
 	defer os.Setenv(EnvVaultClientKey, oldClientKey)
 	defer os.Setenv(EnvVaultSkipVerify, oldSkipVerify)
+	defer os.Setenv(EnvVaultTLSMinVersion, oldTLSMinVersion)
 	defer os.Setenv(EnvVaultMaxRetries, oldMaxRetries)
 
 	config := DefaultConfig()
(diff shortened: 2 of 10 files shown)


hsimon-hashicorp:
Hi @evan361425! Thanks for this submission. I talked with one of our lead engineers and he suggested that instead of taking environment variables and CLI options, it may be better to handle this input in the vault server listener config. Also, please don't forget to add tests for the new functionality. :) Please let me know if we can be of more help!

evan361425:
> ..., it may be better to handle this input in the vault server listener config.

Vault agent can config listener TLS min version, but I need to set the TLS version in client, not listener.

```text
 ┌──────────────────────────────────┐
 │      Confiurable TLS version in  │
 │      listener Stanza             │                 TLS min 1.2
 │          │                       │ I want to send ┌────────┐
 │ ┌─────┐  ▼             ┌───────┐ │ TLS v. 1.3     │        │
 │ │ APP ├────────────────► Agent ├─┼────────────────► Server │
 │ └─────┘  But I want it └───────┘ │                │        │
 │          to be HTTP              │                └────────┘
 │                                  │
 └──────────────────────────────────┘
```

I will add the test once this feature is accepted.

hsimon-hashicorp:
@evan361425 please feel free to open an issue to have a more in-depth discussion about the implementation details with our engineers! Unfortunately I cannot ask them to merge any PRs without sufficient tests built in. Thanks for understanding. :)

evan361425:
I've add the test, but not sure how to test Vault agent that sending HTTPS request to mock server.

Something like:

```go
srv := NewServer(httpsPort)
go srv.ListenAndServeTLS("cert.pem", "key.pem")

_, cmd := testAgentCommand(t, logger)

config := `
listener "tcp" {
  address = "127.0.0.1:8101"
  tls_disable = true
}

vault {
  address = "http://127.0.0.1:%s"
  tls_min_version = "tls13"
}`
config = fmt.Sprintf(config, httpsPort)
configPath := makeTempFile(t, "config.hcl", config)
cmd.Run([]string{"-config", configPath})

version := cmd.client.CloneConfig().HttpClient.Transport.(*http.Transport).TLSClientConfig.MinVersion
if version != tls.VersionTLS13 {
  t.Fatalf("bad: wrong version")
}
```

cipherboy:
Hi @evan361425 -- thanks for the diagram above. I'm curious what's the rationale behind this? 

I think in general our thought process is as follows:

  - Go is generally fairly opinionated about TLS, which at least in my personal view, is a good thing.
  - Go already will prefer to negotiate TLSv1.3 ahead of TLSv1.2, assuming client and server support both.
  - Any options we add to configure Go's TLS suite now, run some risk for us due to Go changing behavior and/or being ignored in the future (see for instance the PreferServerCipherSuites and/or CipherSuite changes: https://go.dev/blog/tls-cipher-suites) -- this means we assume some supportability risk in the future if Go changes its mind.
  - Servers necessarily have to work with older machines (and thus, we feel the introduction of TLSMinVersion justified from a support side -- and, due to Go's opinions, felt we had to include TLSMaxVersion for those few users who had good reason not to negotiate TLSv1.3, such as strictly needing to control cipher suites).
  - However, we feel that TLS Clients (in this case, Vault Agent acting as a client to talk to Vault Server) should be as secure as reasonably possible, within the bounds of still working with most Vault servers. The default client here will advertise both TLS 1.2 and TLS 1.3 and the cipher suites are generally considered secure. Unless the Vault server is configured with TLSMaxVersion=TLS1.1, the client will connect just fine (and, notably, this configuration is considered insecure by most).

So in general, I think we'd prefer not to merge this PR without strong justifications around why.

Could you explain more what you're trying to accomplish, and why the default settings (which will still consistently negotiate TLSv1.3) won't work? 


Thanks!

evan361425:
In my test, when Vault server set _TLSMinVersion_ as default value which is `tls12`, Vault agent will connect to server in TLS 1.2 not TLS 1.3. This is why this PR is here, I want agent to use TLS 1.3 without changing server's configuration (the reason you have mentioned).

I use Wireshark to monitor outbound networking. When I use binary by `make dev` as Vault agent and set agent's _TLSMinVersion_ to `tls13`, it successfully turn to TLS 1.3.

I've found OpenSSL `s_client` [document](https://www.openssl.org/docs/man1.1.1/man1/openssl-s_client.html) that shows client do negotiate the highest supported version of TLS which is opposite to my testing.

> -ssl3, -tls1, -tls1_1, -tls1_2, -tls1_3, -no_ssl3, -no_tls1, -no_tls1_1, -no_tls1_2, -no_tls1_3
> These options require or disable the use of the specified SSL or TLS protocols. By default s_client will negotiate the highest mutually supported protocol version. When a specific TLS version is required, only that version will be offered to and accepted from the server. Note that not all protocols and flags may be available, depending on how OpenSSL was built.

I'll show the image of monitoring and configuration when I backed to office.

evan361425:
Here is my testing:

## Using v1.8.5

### Build up agent

As the config file shows, I am using mTLS.

```sh
$ cat client/config.hcl
pid_file = "client/pid"
exit_after_auth = false

vault {
  ca_cert = "cert-generator/certs/ca.crt"
  client_cert = "cert-generator/certs/client.crt"
  client_key = "cert-generator/certs/client.key"
  // tls_min_version = "tls13"
  retry {
    num_retries = 0
  }
}

auto_auth {
  method "cert" {
    name = "localhost"
    ca_cert = "cert-generator/certs/ca.crt"
    client_cert = "cert-generator/certs/client.crt"
    client_key = "cert-generator/certs/client.key"
  }

  sink "file" {
    config {
      path = "client/token.txt"
    }
  }
}

cache {
  use_auto_auth_token = true
}

listener "tcp" {
  address = "localhost:8100"
  tls_disable = true
}

$ vault agent -config client/config.hcl -address=https://transit.example.com
WARNING! Cache is enabled and using the same certificates from the 'cert'
auto-auth method specified in the 'vault' stanza. Consider specifying
certificate information in the 'cert' auto-auth's config stanza.
==> Vault agent started! Log data will stream in below:

==> Vault agent configuration:

           Api Address 1: http://127.0.0.1:8100
                     Cgo: disabled
               Log Level: info
                 Version: Vault v1.8.5
             Version Sha: 647eccfe0bd5817bdd8628f3c3171402dfc8a8fc

2021-11-08T10:04:31.821+0800 [INFO]  sink.file: creating file sink
2021-11-08T10:04:31.822+0800 [INFO]  sink.file: file sink configured: path=client/token.txt mode=-rw-r-----
...
```

### Call the agent

```sh
$ curl -s -X POST "$VAULT_ADDR/v1/transit/encrypt/aes-key" \
  -H "x-vault-token: $token" --cacert root.crt \
  -d "{\"plaintext\":\"$encoded_data\"}" | jq '.data'
```

### Result

截圖 2021-11-08 上午10 13 45

## Using this PR binary

Add `pls_min_version` to the config.

### Build up agent

```sh
$ cat client/config.hcl
...
vault {
  ...
  tls_min_version = "tls13"
}
...

$ /Users/evan.lu/go/bin/vault agent -config client/config.hcl -address=https://transit.example.com
WARNING! Cache is enabled and using the same certificates from the 'cert'
auto-auth method specified in the 'vault' stanza. Consider specifying
certificate information in the 'cert' auto-auth's config stanza.
==> Vault agent started! Log data will stream in below:

==> Vault agent configuration:

           Api Address 1: http://127.0.0.1:8100
                     Cgo: disabled
               Log Level: info
                 Version: Vault v1.10.0-dev
             Version Sha: 5cb71155523b31bcbf70b579ddad79f3dd6c8a9e
...
```

### Call the agent

```sh
$ curl -s -X POST "$VAULT_ADDR/v1/transit/encrypt/aes-key" \
  -H "x-vault-token: $token" --cacert root.crt \
  -d "{\"plaintext\":\"$encoded_data\"}" | jq '.data'
```

### Result

截圖 2021-11-08 上午10 24 12

hsimon-hashicorp:
Hi @evan361425! Thanks for the responses. Could you please also make sure to sign the Contributor License Agreement? Thanks in advance. :)

cipherboy:
Hey @evan361425 -- do you mind sharing a pcap?

I've created a branch here: https://github.com/cipherboy/vault/commit/capture-tls-log -- which supports TLS shared secret key logging. If you export something like:

```
export VAULT_SKIP_VERIFY=true
export VAULT_INSECURE_DEBUG_TLS_KEY_LOG=/path/to/some/tls-key.log
```

You should be able to run your agent, capture a pcap, and decode the TLS stream to confirm it really is the traffic you expect. I've consistently gotten it to do TLSv1.3 with mutual auth between Agent and Vault Server, both built with Go 1.17.3 on Vault `1.10.0-dev` (the present `main` branch as of the tip of that custom branch above).

In your first screenshot, you see packet 35 with a change cipher spec and three Application Data entries. This is roughly what I'd expect to see from a TLSv1.3 mutual auth. The same is in the third screenshot, packet 20. 

Here's a before/after capture screenshot with the key log.

Screenshot from 2021-11-18 16-25-02

Screenshot from 2021-11-18 16-25-37

Now, the only thing I can think of is perhaps you're on an older version of Go, but in 1.13 it was moved to opt-out, and in 1.14, the opt-out was removed and AFAICT, TLS v1.3 was thus negotiated by default. Is there anything else in your setup? Perhaps a reverse proxy, some other client talking to the remote vault server (that might be still using TLSv1.2)? 

I've been unable to reproduce this randomness :/

evan361425:
Thanks for replied, @cipherboy.

I tried again on docker with [latest](https://hub.docker.com/layers/vault/library/vault/latest/images/sha256-5afca84d461998bfa9143942a14f36ae575862a6f566e8650dc171646c97dd81?context=explore) Vault.

The output is always TLSv1.3.

But after I using Vault on local environment (MacOS BigSur 11.6), the problem reproduced. It seams like a bug on Wireshark, I have looked into the Transport Layer Security output, but both TLSv1.2 and TLSv1.3 have the same output: `TLS 1.2 (0x0303)`.

packet 13 TLS output

packet 22 TLS output

I'll close this PR after I confirmed that the problem is come from Vault then open a new issue to show more info on that.

cipherboy:
@evan361425 Just for context, I'm using Wireshark 3.4.9 (Git commit 365e236f5efe) from the Fedora 34 distribution. 

Regarding the `TLS 1.2 (0x0303)` annotation you see in Wireshark, take a look at [TLS Annotated's 1.3 Wrapper (second to the last box on the page, just above server's application data)](https://tls13.ulfheim.net/). You'll see that in middlebox compatibility mode, TLSv1.3 Application Data is wrapped in a TLS 1.2 "record". It could be that [Go used this mode](https://github.com/golang/go/issues/28890) and earlier versions of Wireshark didn't pick up on this... Or perhaps there's other traffic between these two devices, I do not know.

Regardless, it sounds like you've figured some things out, feel free to ping me or @hsimon-hashicorp if there's any future issue with the TLS and you open a new issue. :-)

PR #13042, opened 2021-11-04
Fix errors logged on standbys when we try to write versions to storage

ncabatoff:
I haven't actually seen error logs, but we know that only the active node should be allowed to write to storage.

While I was at it, I made some methods private that didn't need to be public and removed some dead code.

### DIFF
--- a/vault/core.go
+++ b/vault/core.go
@@ -1047,17 +1047,17 @@ func NewCore(conf *CoreConfig) (*Core, error) {
 
 // HandleVersionTimeStamps stores the current version at the current time to
 // storage, and then loads all versions and upgrade timestamps out from storage.
-func (c *Core) HandleVersionTimeStamps(ctx context.Context) error {
+func (c *Core) handleVersionTimeStamps(ctx context.Context) error {
 	currentTime := time.Now()
-	isUpdated, err := c.StoreVersionTimestamp(ctx, version.Version, currentTime)
+	isUpdated, err := c.storeVersionTimestamp(ctx, version.Version, currentTime)
 	if err != nil {
-		return err
+		return fmt.Errorf("error storing vault version: %w", err)
 	}
 	if isUpdated {
 		c.logger.Info("Recorded vault version", "vault version", version.Version, "upgrade time", currentTime)
 	}
 	// Finally, load the versions into core fields
-	err = c.HandleLoadVersionTimestamps(ctx)
+	err = c.loadVersionTimestamps(ctx)
 	if err != nil {
 		return err
 	}
@@ -2000,6 +2000,9 @@ func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c
 			return err
 		}
 	}
+	if err := c.handleVersionTimeStamps(ctx); err != nil {
+		return err
+	}
 	if err := c.setupPluginCatalog(ctx); err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


hghaf099 (approved):
LGTM

vault/core.go
```diff
@@ -2000,6 +2000,9 @@ func (s standardUnsealStrategy) unseal(ctx context.Context, logger log.Logger, c
 			return err
 		}
 	}
+	if err := c.handleVersionTimeStamps(ctx); err != nil {
```
HridoyRoy, line 2003: If this executes on a perf standby when the standby unseals, will it cause the standby to error on the storage write?

vault/core.go
```diff
@@ -1047,17 +1047,17 @@ func NewCore(conf *CoreConfig) (*Core, error) {
 
 // HandleVersionTimeStamps stores the current version at the current time to
 // storage, and then loads all versions and upgrade timestamps out from storage.
-func (c *Core) HandleVersionTimeStamps(ctx context.Context) error {
+func (c *Core) handleVersionTimeStamps(ctx context.Context) error {
```
HridoyRoy, line 1050: Thank you for making these all private!

vault/core.go
```diff
@@ -2157,11 +2160,6 @@ func (c *Core) postUnseal(ctx context.Context, ctxCancelFunc context.CancelFunc,
 			c.logger.Warn("post-unseal post seal migration failed", "error", err)
 		}
 	}
-	err := c.HandleVersionTimeStamps(c.activeContext)
-	if err != nil {
-		c.logger.Warn("post-unseal version timestamp setup failed", "error", err)
```
HridoyRoy, line 2162: Is this the extra log line you were talking about in the PR description, where essentially the standby would always warn about the timestamp loading/storing not succeeding in the logs?

vault/core_util_common.go
```diff
@@ -39,27 +39,6 @@ func (c *Core) StoreVersionTimestamp(ctx context.Context, version string, curren
 	return true, nil
 }
 
-// FindMostRecentVersionTimestamp loads the current vault version and associated
-// upgrade time from storage.
-func (c *Core) FindMostRecentVersionTimestamp() (string, time.Time, error) {
```
HridoyRoy, line 44: Thanks for removing this!

HridoyRoy (commented):
Just a few questions on the PR, but looks good to me overall!

vault/core.go
ncabatoff: I moved it here because standardUnsealStrategy doesn't apply to perf standbys.

vault/core.go
ncabatoff: I didn't actually test and see the log line myself, I speculated that we'd see this line on perf standbys.

HridoyRoy (approved):
lgtm!

vault/core.go
briankassouf: Do we want to load but not store the values on perf standbys?

ncabatoff: D'oh, good point.  Will fix.

ncabatoff: Well, I guess I'll fix in ent once this is merged.

ncabatoff: Decided against loading on perf standbys since it wouldn't be reliable: https://github.com/hashicorp/vault/pull/13081

PR #13044, opened 2021-11-04
Allowing Unwrap w/ Newline files

AnPucel:
- Adding a fix in the unwrap function to handle newline characters in files so users don't get an error from newline characters

- Tested with local Vault server

### DIFF
--- a/api/logical.go
+++ b/api/logical.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/vault/sdk/helper/jsonutil"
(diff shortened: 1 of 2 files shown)


changelog/13044.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+trims newline character from wrapping token in UnWrap Go SDK
```
calvn, line 2:
```suggestion
api: Trim newline character from wrapping token in logical.Unwrap from the api package
```

api/logical.go
```diff
@@ -235,12 +236,13 @@ func (c *Logical) DeleteWithData(path string, data map[string][]string) (*Secret
 
 func (c *Logical) Unwrap(wrappingToken string) (*Secret, error) {
 	var data map[string]interface{}
+	wt := strings.TrimSuffix(string(wrappingToken), "\n")
```
calvn, line 239: Would using `strings.TrimSpace` be better in here? We don't have to wrap `wrappingToken` around `string()` since it's already of that type.

ncabatoff (approved):

PR #13045, opened 2021-11-04
Add  missing install/remove scripts to rpm/deb packages

stevendpclark:


benashz (approved):

mladlow:
I think this is the right target branch for this change, but eventually we'll need to replicate it across the rest of the release branches and everywhere else we have CRT going (main, I think). This comment is partly just a reminder for myself.

kpenfound (approved):
Nailed it, looks good!

PR #13054, opened 2021-11-04
UI/Adds pagination to auth methods list

hellobontempo:
Fix for #12962 - when lots of auth method groups or users created there is no way to navigate to view the full list.

auth-method-error

**With pagination:** 
auth-method-pagination


Monkeychip (approved):
Looks great. I added the 1.9.1 milestone, just guessing that's where it needs to go, but amend if I'm wrong.

PR #13063, opened 2021-11-05
Fix regression preventing non-docker tests from running.

ncabatoff:
Regression due to typo caused by: https://github.com/hashicorp/vault/pull/13017

I'm going to experiment with moving the `|| true`, since that helped masked the error.  These were introduced by #12630, and probably are needed, but I'd like to see how it fails without them and if there's a better way.


ncabatoff:
I decided to defer my efforts to get rid of the `|| true` bits, since it's urgent to get the tests running on main again.

ccapurso (approved):

victorr (approved):
👍

PR #13065, opened 2021-11-05
NIST SP800-108 Compliant Counter KDF

cipherboy:
This pull request implements a [NIST SP800-108](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-108.pdf) compliant Counter KDF, aligning with the [PKCS#11 v3.0](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/pkcs11-curr-v3.0.html#_Toc437440585) interface specification.

This differs from the existing KDF in two important ways:

 1. This has been tested against NIST CAVP tests. Notably, the existing implementation starts the counter at 0, rather than 1, causing non-compliance with NIST's specification. This means the existing implementation is non-compliant and not compatible with other implementations (say, in NSS or OpenSSL).
 2. Parameter width and ordering is more flexible than the fixed 4-byte counter / DKM encoding presently used. This allows for a broader range of CAVP test cases and allows for interoperability with other specifications. 

Additionally, we implement KBKDF aligning with Go's existing [HKDF](https://pkg.go.dev/golang.org/x/crypto/hkdf#Expand) interface, taking a `hash.Hash` instance and returning a `io.Reader` instance. 

Implementation is separate from the instantiation to eventually allow for a BoringCrypto/FIPS-certified implementation of this primitive, to replace the pure-Go version.


sdk/helper/kdf/counter.go
```diff
@@ -0,0 +1,63 @@
...
+	"hash"
+	"io"
+)
+
+// Creates a new SP800-108 Counter Mode KDF instance. This function takes
+// a PRF instance (preferably HMAC or CMAC) already initialized with a key,
+// an ordered list of KDF parameters to use (see params.go), and a list of
+// bit lengths of keys to derive (in order). When successfully constructed,
+// the return is an io.Reader instance, which when called with byte arrays
+// sized according to the specified originally specified keyLengths, is
```
divyapola5, line 16: nit : specified mentioned twice

divyapola5 (approved):
Good work Alex!
Today I learned that switch in go does not need a break statement ;)

sdk/helper/kdf/params.go
```diff
@@ -0,0 +1,210 @@
...
+	// the value to encode; must be divisible by 8 with a maximum value of
+	// 64.
+	Width uint8
+}
+
+var _ KBKDFParameter = &DKMLength{}
+
+// Defines a fixed byte array value (such as the info or context fields from
+// NIST SP800-108). No additional manipulation are performed on these fields.
+type ByteArray struct {
```
sgmiller, line 95: You can get away with simply "type ByteArray []byte", and implement the interface directly without the wrapper struct.  I might not call it ByteArray though, Go doesn't have arrays, it has slices (and it's very touchy about it)

sgmiller (changes requested):
Mostly Goisms

sdk/helper/kdf/counter_impl.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+type counterKDF struct {
+	// Globally provisioned values
+	prf        hash.Hash
+	parameters []KBKDFParameter
+	keyLengths []int
+
+	// Internal counter for this KDF instance.
+	counter uint64
```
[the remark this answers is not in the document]
cipherboy: `go fmt ./*.go` from `sdk/helper/kdf` results in no changes to these files (Go 1.17.2). 

Because of the comments between members, `go fmt` won't align the types. If you remove the comments, it will reformat them... Should I remove the comments?

sdk/helper/kdf/params.go
cipherboy: Sans `ChainingVariable`, the remaining types all mirror the PKCS#11v3.0 spec:  https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/pkcs11-curr-v3.0.html#_Toc30061448 

Would you still update this name? If so, to what?

cipherboy:
@sgmiller I've added the high-level interfaces we discussed -- I ended up going with two: one that setup the parameters as specified in the SP800-108 doc and one that mirrored the existing parameter set (essentially equivalent if label and context are merged together).

sdk/helper/kdf/counter_impl.go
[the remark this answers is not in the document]
sgmiller: No, that's fine.  Plenty of precedent for that.  Was just making sure that source was getting auto-formatted.

sgmiller (approved):

sdk/helper/kdf/params.go
```diff
@@ -0,0 +1,206 @@
...
+	return nil
+}
+
+func (kl DKMLength) CalculateDKMLength(prfSize int, keys []int) uint64 {
+	// Assumption: Validate has already passed, meaning there is exactly one
+	// of two values for kl.Method.
+	var result uint64
+
+	for _, keyBits := range keys {
+		if kl.Method == SumOfKeys {
```
stevendpclark, line 161: nit: Perhaps use a switch statement here? I'm always worried about additions/new value types to kl.Method?

PR #13073, opened 2021-11-07
Documentation consistency GPG keys are PGP keys.

dngray:
This one is just a minor documentation fix, which irked me when I was reading some of the help guides.

In some places documentation rightfully refers to "PGP keys", then sporadically throughout this is used interchangeably with "GPG keys".

GPG is actually referred to as GnuPG in official upstream documentation, so I fixed that. As far as referring to the keys themselves, they are still PGP keys, whether created by GnuPG [or some other implementation](https://wiki.gnupg.org/OtherFreeSoftwareOpenPGP).


hsimon-hashicorp (approved):

taoism4504 (approved):
LGTM

PR #13076, opened 2021-11-08
Backport 1.9: Fix errors logged on standbys when we try to write versions to storage (#13042)

ncabatoff:


vishalnayak (commented):
Hadn't reviewed the PR on main, but took a cursory look at the PR and 1 question stands out in my mind. You might have thought through. This might be just me trying to get a closure.

The main change I believe in this PR was to move the `handleVersionTimeStamps` from post unseal to standard unseal strategy's unseal function thereby avoiding the warnings on perf standbys. Right? The comment on the PR also mentioned that this should only be done on the active node only. What happens on regular standbys?

ncabatoff:
> The main change I believe in this PR was to move the `handleVersionTimeStamps` from post unseal to standard unseal strategy's unseal function thereby avoiding the warnings on perf standbys. Right? 

Right.

> The comment on the PR also mentioned that this should only be done on the active node only. What happens on regular standbys?

Regular standbys never call postUnseal, so they aren't implicated one way or the other.

HridoyRoy (approved):

PR #13078, opened 2021-11-08
Add a periodic test of the autoseal to detect loss of connectivity.

sgmiller:
This allows operators to spot problems with connections to their HSMs and
KMSes that would otherwise go unnoticed until the next unseal, where time is
critical.


vault/seal_autoseal.go
```diff
@@ -18,16 +21,20 @@ import (
 // applicable in the OSS side
 var barrierTypeUpgradeCheck = func(_ string, _ *SealConfig) {}
 
+const sealHeathTestInterval = 1 * time.Minute
```
stevendpclark, line 24: Didn't we want to check this hourly?

vault/seal_autoseal.go
```diff
@@ -499,3 +506,52 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+func (d *autoSeal) StartHealthCheck() {
+	d.healthCheck = time.NewTicker(sealHeathTestInterval)
+	d.healthCheckStop = make(chan struct{})
+	go func() {
+		lastTestOk := true
+		lastSeenOk := time.Now()
+		for {
+			select {
+			case <-d.healthCheckStop:
+				d.healthCheck.Stop()
```
stevendpclark, line 519: I'm probably being paranoid here but if we get multiple healthCheckStop we will get a nil deference error. Is it worth adding a quick test and return if d.healthCheck is nil?

vault/seal_autoseal.go
```diff
@@ -499,3 +506,52 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+					}
+				}
+			}
+		}
+	}()
+}
+
+func (d *autoSeal) StopHealthCheck() {
+	if d.healthCheckStop != nil {
+		d.healthCheckStop <- struct{}{}
```
ncabatoff, line 555: Wouldn't it be safer to close the channel?  That way if we're blocked waiting for a response from the upstream seal service, we won't block sealing Vault.

vault/seal_autoseal.go
stevendpclark: nvm the case statement will never fire.

stevendpclark (approved):
👍 Just review the seal test interval if we really want to run this every minute.

vault/seal_autoseal.go
sgmiller: Yeah, I'm back and forth on that.  Since it's a cheap encrypt/decrypt I coded it to be more frequent.  Open to thoughts.

vault/seal_autoseal.go
sgmiller: yes it is, thanks.

vault/seal_autoseal.go
sgmiller: Good idea.

vault/seal_autoseal.go
stevendpclark: I'm assuming this is being done on all auto-unseal implementations, not just pkcs11 implementations. If that is the case could the costs associated with using the various KMS providers within the cloud start adding up? I don't have a good sense of those costs across the various implementations.

hsimon-hashicorp: `sealHeathTestInterval` - should this be sealHea**l**thTestInterval?

vault/seal_autoseal.go
```diff
@@ -499,3 +509,67 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+					}
+				}
+			}
+		}
+	}()
+}
+
+func (d *autoSeal) StopHealthCheck() {
+	if d.healthCheckStop != nil {
+		close(d.healthCheckStop)
```
ncabatoff, line 573: Maybe assign nil to the field so that if for some reason we get called twice we don't panic?

vault/seal_autoseal.go
sgmiller: It sure should.

vault/seal_autoseal.go
sgmiller: Yep.

ncabatoff (commented):
Sorry, I somehow failed to submit these comments yesterday.

vault/seal_autoseal.go
```diff
@@ -499,3 +509,67 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
 
 	return nil
 }
+
+// StartHealthCheck starts a goroutine that tests the health of the auto-unseal backend once every 10 minutes.
+// If unhealthy, logs a warning on the condition and begins testing every one minute until healthy again.
+func (d *autoSeal) StartHealthCheck() {
+	d.healthCheck = time.NewTicker(sealHeathTestIntervalNominal)
```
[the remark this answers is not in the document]
sgmiller: I like stop better, will do.

vault/seal_autoseal.go
```diff
@@ -499,3 +509,67 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+			case <-d.healthCheckStop:
+				if d.healthCheck != nil {
+					d.healthCheck.Stop()
+				}
+				d.healthCheck = nil
+				d.healthCheckStop = nil
+				return
+			case t := <-d.healthCheck.C:
+				testVal := fmt.Sprintf("Heartbeat %d", mathrand.Intn(1000))
+				ciphertext, err := d.Wrapper.Encrypt(d.core.activeContext, []byte(testVal), nil)
```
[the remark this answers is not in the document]
sgmiller: Don't we want this logic to run on secondary nodes as well?  I thought the activeContext was related to being the active member of the cluster.

vault/seal_autoseal.go
```diff
@@ -499,3 +509,67 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
 	return nil
 }
+
+// StartHealthCheck starts a goroutine that tests the health of the auto-unseal backend once every 10 minutes.
+// If unhealthy, logs a warning on the condition and begins testing every one minute until healthy again.
+func (d *autoSeal) StartHealthCheck() {
+	d.healthCheck = time.NewTicker(sealHeathTestIntervalNominal)
+	d.healthCheckStop = make(chan struct{})
+	go func() {
+		lastTestOk := true
```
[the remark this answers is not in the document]
sgmiller: Great ideas.

vault/seal_autoseal.go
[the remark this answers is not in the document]
ncabatoff: Confusingly enough, activeContext applies to perf standbys as well.

vault/seal_autoseal.go
```diff
@@ -499,3 +512,76 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+			select {
+			case <-healthCheckStop:
+				if healthCheck != nil {
+					healthCheck.Stop()
+				}
+				healthCheckStop = nil
+				return
+			case t := <-healthCheck.C:
+				testVal := fmt.Sprintf("Heartbeat %d", mathrand.Intn(1000))
+				ciphertext, err := d.Access.Encrypt(d.core.activeContext, []byte(testVal), nil)
```
ncabatoff, line 539: I think there might still be a race here?  If we're sealing, our stop channel gets closed, but sealing could complete and Core.activeContext changed before we see the closed channel.  It's unlikely, but I think either StopHealthCheck should wait for the goroutine to exit, or we should define a local ctx var to hold activeContext before starting the for loop.

vault/seal_autoseal.go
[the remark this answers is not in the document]
sgmiller: Nm, misunderstood what you wrote.  Throwing that in now.

vault/seal_autoseal.go
sgmiller: Hmm, yeah, remote but possible.

vault/seal_autoseal.go
```diff
@@ -499,3 +513,84 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+
+					if err != nil {
+						d.logger.Warn("failed to encrypt seal health test value, seal backend may be unreachable", "error", err)
+						if lastTestOk {
+							healthCheck.Reset(sealHealthTestIntervalUnhealthy)
+						}
+						lastTestOk = false
+						d.core.MetricSink().SetGauge(autoSealUnavailableDuration, 0)
+					} else {
+						plaintext, err := d.Access.Decrypt(ctx, ciphertext, nil)
```
ncabatoff, line 556: Same timeout (shared) across both Encrypt/Decrypt, or should we have a 1m timeout for each?

vault/seal_autoseal.go
```diff
@@ -499,3 +513,84 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+							if lastTestOk {
+								healthCheck.Reset(sealHealthTestIntervalUnhealthy)
+							}
+							lastTestOk = false
+						} else {
+							d.logger.Debug("seal health test passed")
+							if !lastTestOk {
+								d.logger.Info("seal backend is now healthy again", "downtime", t.Sub(lastSeenOk).String())
+							}
+							if !lastTestOk {
```
ncabatoff, line 575: Can this be merged with previous `if`?

vault/seal_autoseal.go
```diff
@@ -499,3 +513,84 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+								healthCheck.Reset(sealHealthTestIntervalNominal)
+							}
+							lastTestOk = true
+							lastSeenOk = t
+						}
+						d.core.MetricSink().SetGauge(autoSealUnavailableDuration, float32(time.Since(lastSeenOk).Nanoseconds()))
```
ncabatoff, line 581: We tend to use millis for our time-based metrics, any reason you want nanos here?

vault/seal_autoseal.go
```diff
@@ -499,3 +513,84 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+		for {
+			select {
+			case <-healthCheckStop:
+				if healthCheck != nil {
+					healthCheck.Stop()
+				}
+				healthCheckStop = nil
+				return
+			case t := <-healthCheck.C:
+				func() {
```
ncabatoff, line 541: If you just want a lexical scope you don't need to create a func(), you can just use curlies.  Though there's enough indentation that this is hard to read (at least as a split diff), so you might well want to refactor.

vault/seal_autoseal.go
```diff
@@ -499,3 +513,81 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
+					ctx, cancel := context.WithTimeout(ctx, sealHealthTestTimeout)
+					defer cancel()
+
+					testVal := fmt.Sprintf("Heartbeat %d", mathrand.Intn(1000))
+					ciphertext, err := d.Access.Encrypt(ctx, []byte(testVal), nil)
+
+					if err != nil {
+						fail("failed to encrypt seal health test value, seal backend may be unreachable", "error", err)
```
ncabatoff, line 558: Now that this is a func, could we make this an early return instead of an else?

vault/seal_autoseal.go
```diff
@@ -499,3 +513,82 @@ func (d *autoSeal) migrateRecoveryConfig(ctx context.Context) error {
...
```
ncabatoff, line 558: Now that we're in a func we could do an early return instead of an else.

ncabatoff (approved):

heliomartins:
Hello @sgmiller,

I was wondering if you could please clarify a quick question I have around this PR (which is a great feature by the way).

From my understanding, the only way to be verify if the seal backend is health would be by checking Vault logs. Do you know if there is any API that we could call to get the status of the seal backend without having to check the logs?

PR #13079, opened 2021-11-08
Add note that monitor command may truncate logs

mladlow:


website/content/api-docs/system/monitor.mdx
```diff
@@ -8,6 +8,9 @@ description: The `/sys/monitor` endpoint is used to receive streaming logs from
 
 The `/sys/monitor` endpoint is used to receive streaming logs from the Vault server.
 
+If Vault is emitting log messages faster than a receiver can process them, the
+log messages will be truncated.
```
ncabatoff, line 12:
```suggestion
some log lines will be dropped.
```

Truncation implies (to me at least) that we send partial log lines, which isn't the case.

ncabatoff (approved):
I like my wording better but this is fine too.

website/content/docs/commands/monitor.mdx
```diff
@@ -20,6 +20,9 @@ Note that this command is designed to run indefinitely. It is similar to
 it encounters an unexpected error. As a user, you must terminate this
 process yourself to shut it down.
 
+If Vault is emitting log messages faster than a receiver can process them, the
+log messages will be truncated.
```
mladlow, line 24:
```suggestion
some log lines will be dropped.
```

PR #13080, opened 2021-11-08
Add universal default key_bits value for PKI endpoints

cipherboy:
Built on top of #12872; will be rebased once that merges.

----

PKI endpoints currently require `key_bits` field to be changed if `key_type` changes; this is because `key_bits` presently has default value 2048, which is nonsensical for `key_type=ec` keys (which support 224, 256, 384, and 521). Changing `key_bits` to a default value of 0 (a "universal default" similar to what was done with `signature_bits` in #12872) allows us to switch the actual default value of `key_bits` internally, based on what value `key_type` takes.

This retains the default value of 2048 for RSA and allows for a request with only `key_type=ec` to succeed, with 256-bit NIST P-Curve.

Also slightly refactors RSA small key detection and `pki/backend_test.go`.

### DIFF
--- a/changelog/13080.txt
+++ b/changelog/13080.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+secrets/pki: Default value for key_bits changed to 0, enabling key_type=ec key generation with default value
+```
--- a/sdk/helper/certutil/helpers.go
+++ b/sdk/helper/certutil/helpers.go
@@ -33,6 +33,14 @@ import (
 	cbasn1 "golang.org/x/crypto/cryptobyte/asn1"
 )
 
+const rsaMinimumSecureKeySize = 2048
+
+// Mapping of key types to default key lengths
+var defaultAlgorithmKeyBits = map[string]int {
+	"rsa": 2048,
+	"ec": 256,
+}
+
 // Mapping of NIST P-Curve's key length to expected signature bits.
 var expectedNISTPCurveHashBits = map[int]int{
 	224: 256,
@@ -536,7 +544,7 @@ func StringToOid(in string) (asn1.ObjectIdentifier, error) {
 // Validates that the combination of keyType, keyBits, and hashBits are
 // valid together; replaces individual calls to ValidateSignatureLength and
 // ValidateKeyTypeLength.
-func ValidateKeyTypeSignatureLength(keyType string, keyBits int, hashBits *int) error {
+func ValidateKeyTypeSignatureLength(keyType string, keyBits *int, hashBits *int) error {
 	if err := ValidateKeyTypeLength(keyType, keyBits); err != nil {
 		return err
 	}
@@ -546,10 +554,10 @@ func ValidateKeyTypeSignatureLength(keyType string, keyBits int, hashBits *int) 
 		// store policy section 5.1.2, enforce that NIST P-curves use a hash
 		// length corresponding to curve length. Note that ed25519 does not
 		// the "ec" key type.
-		expectedHashBits := expectedNISTPCurveHashBits[keyBits]
+		expectedHashBits := expectedNISTPCurveHashBits[*keyBits]
 
 		if expectedHashBits != *hashBits && *hashBits != 0 {
-			return fmt.Errorf("unsupported signature hash algorithm length (%d) for NIST P-%d", *hashBits, keyBits)
+			return fmt.Errorf("unsupported signature hash algorithm length (%d) for NIST P-%d", *hashBits, *keyBits)
 		} else if *hashBits == 0 {
 			*hashBits = expectedHashBits
 		}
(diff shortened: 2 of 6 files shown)


sdk/helper/certutil/helpers.go
```diff
@@ -586,21 +594,32 @@ func ValidateSignatureLength(hashBits int) error {
...
-func ValidateKeyTypeLength(keyType string, keyBits int) error {
+func ValidateKeyTypeLength(keyType string, keyBits *int) error {
+	if *keyBits == 0 {
+		newValue, present := defaultAlgorithmKeyBits[keyType]
+		if present {
+			*keyBits = newValue
+		}
+	}
+
+	if keyType == "rsa" && *keyBits < rsaMinimumSecureKeySize {
```
stevendpclark, line 605: nit: This could be moved into the default case below simplifying the conditional.

changelog/13080.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+secrets/pki: Default value for key_bits changed to 0 to enable default key_type=ec key generation time
```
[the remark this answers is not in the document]
sgmiller: Lol.  s/Dod/do/

sdk/helper/certutil/helpers.go
```diff
@@ -536,7 +544,7 @@ func StringToOid(in string) (asn1.ObjectIdentifier, error) {
 // Validates that the combination of keyType, keyBits, and hashBits are
 // valid together; replaces individual calls to ValidateSignatureLength and
 // ValidateKeyTypeLength.
-func ValidateKeyTypeSignatureLength(keyType string, keyBits int, hashBits *int) error {
+func ValidateKeyTypeSignatureLength(keyType string, keyBits *int, hashBits *int) error {
```
sgmiller, line 547: Bit concerned that it's surprising that a Validation function actually changes it's inputs.  Rather than use a pointer, I'd prefer a (named?) multi-return that returns the new keyBits/hashBits, and change the function name in some way to show that it's both validating and updating.

sgmiller (changes requested):
Mostly like it, but would a couple comments.

sdk/helper/certutil/helpers.go
```diff
@@ -586,21 +594,32 @@ func ValidateSignatureLength(hashBits int) error {
 	return nil
 }
 
-func ValidateKeyTypeLength(keyType string, keyBits int) error {
+func ValidateKeyTypeLength(keyType string, keyBits *int) error {
+	if *keyBits == 0 {
+		newValue, present := defaultAlgorithmKeyBits[keyType]
+		if present {
+			*keyBits = newValue
```
stevendpclark, line 601: I'll defer this to @sgmiller, but I'm not a huge fan of overwriting pointer values within functions like this. I'd rather see functions returning the proper values to use. I understand that will require more refactoring so I am not sure if it's worth the effort.

stevendpclark: And of course nvm, Scott already mentioned in while I was busy reviewing...

sdk/helper/certutil/helpers.go
cipherboy: Ah sorry, that snuck in from the earlier PR on signature bits, guess I brought in a C-ism. :dagger: 

How does multi-return work at the call site? Would we need to do:

```
var err error
...

if role.KeyBits, role.SignatureBits, err := certutil.ValidateKeyTypeSignatureLength(role.KeyType, role.KeyBits, role.SignatureBits); err != nil {
    return err
}
```

That gets really long and length, in my mind. I think the `&...` (for a C programmer) is sufficient warning to tell them to look at what the function is doing more closely.

sdk/helper/certutil/helpers.go
cipherboy: I ended up not moving it into the default case as I wanted to ensure that, even if `rsaMinimum` is updated to say 3072, we'd block 2048 (which is a good/well-known RSA key size).

changelog/13080.txt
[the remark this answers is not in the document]
cipherboy: Ah, yeah, that didn't make sense. Take a look at this?

cipherboy:
@sgmiller and @stevendpclark, any chance I could bother either of you to take a look at this early next week? Thanks!

sdk/helper/certutil/helpers.go
```diff
@@ -533,63 +541,118 @@ func StringToOid(in string) (asn1.ObjectIdentifier, error) {
...
-// Validates that the combination of keyType, keyBits, and hashBits are
-// valid together; replaces individual calls to ValidateSignatureLength and
-// ValidateKeyTypeLength.
-func ValidateKeyTypeSignatureLength(keyType string, keyBits int, hashBits *int) error {
-	if err := ValidateKeyTypeLength(keyType, keyBits); err != nil {
-		return err
-	}
+// Returns default key bits for the specified key type, or the present value
+// if keyBits is non-zero.
+func DefaultOrValueKeyBits(keyType string, keyBits int) (int, error) {
```
stevendpclark, line 546: We never return an error in any circumstance from this function, could this be simplified?

cipherboy: We don't here, but `DefaultOrValueHashBits` returns an err... does it make sense (since this is in the SDK) to keep the signature the same between the two and thus leave the potential for an error open in the future? That was my reasoning for including an (unused) error type.

If we chose to add ed25519 to `defaultAlgorithmKeyBits` we could turn `!present` into an error case, much like we've done with `expectedNISTPCurveHashBits` below. But that'd also be validating algorithm type, which I think should be done later.

Thoughts?

sgmiller (approved):

PR #13083, opened 2021-11-08
Vault 936: use core.activeContext in ActivityLog

swayne275:


vault/activity_log.go
```diff
@@ -783,30 +783,29 @@ func (a *ActivityLog) resetCurrentLog() {
 }
 
 func (a *ActivityLog) deleteLogWorker(startTimestamp int64, whenDone chan struct{}) {
-	ctx := namespace.RootContext(nil)
 	entityPath := fmt.Sprintf("%v%v/", activityEntityBasePath, startTimestamp)
 	tokenPath := fmt.Sprintf("%v%v/", activityTokenBasePath, startTimestamp)
 
 	// TODO: handle seal gracefully, if we're still working?
-	entitySegments, err := a.view.List(ctx, entityPath)
+	entitySegments, err := a.view.List(a.core.activeContext, entityPath)
```
ncabatoff, line 790: This is racey since we run without the stateLock held: someone could modify core.activeContext while we're running.

I would instead give deleteLogWorker a ctx argument.  The caller should pass in activeContext, which it reads from the core field while it has the stateLock.  In the case of SetConfig, it's called from a request, so it can assume it has the stateLock.  refreshFromStoredLogs looks like it's called with the stateLock held too.  Same applies to retentionWorker, I think, though you'll have to plumb the ctx through some intervening funcs.

swayne275: f848870fe

swayne275:
i'm not sure if there's any value in updating the context used by tests. and if i do, i might want to have a helper to grab the context under stateLock

thoughts @ncabatoff ?

vault/activity_log.go
```diff
@@ -975,7 +973,7 @@ func (a *ActivityLog) SetConfig(ctx context.Context, config activityConfig) {
 	if !a.enabled && a.currentSegment.startTimestamp != 0 {
 		a.logger.Trace("deleting current segment")
 		a.deleteDone = make(chan struct{})
-		go a.deleteLogWorker(a.currentSegment.startTimestamp, a.deleteDone)
+		go a.deleteLogWorker(ctx, a.currentSegment.startTimestamp, a.deleteDone)
```
ncabatoff, line 976: Since SetConfig is invoked from a request context, it should pass b.Core.activeContext to any workers it spawns.

ncabatoff:
> i'm not sure if there's any value in updating the context used by tests. and if i do, i might want to have a helper to grab the context under stateLock

Let's not worry about the tests here unless they become a problem.

vault/activity_log.go
swayne275: 1b5ca5c8d

ncabatoff (approved):

PR #13084, opened 2021-11-08
Remove TLS prefer server cipher suites configuration option due to deprecation in go 1.17.

schultz-is:
The `tls.Config` parameter driven by this configuration option was deprecated in go 1.17. It looks as though the vault upgrade guide for 1.9 mentions this deprecation, as do the website docs.


stevendpclark (approved):
👍

PR #13087, opened 2021-11-08
Add message while adding Oracle db connection

arnav28:
- Since UI currently doesn't have the support for custom plugin names, inform user to use the default plugin name for the oracle database

Screen Shot 2021-11-10 at 1 27 25 PM

### DIFF
--- a/ui/app/templates/components/database-connection.hbs
+++ b/ui/app/templates/components/database-connection.hbs
@@ -87,7 +87,7 @@
   {{#if (eq @model.plugin_name "vault-plugin-database-oracle")}}
       <AlertBanner @type="warning">
         Please ensure that your Oracle plugin has the default name of <b>vault-plugin-database-oracle</b>. 
-        Custom naming is not supported in the UI at this time.
+        Custom naming is not supported in the UI at this time. If the plugin is already named vault-plugin-database-oracle, disregard this warning.
       </AlertBanner>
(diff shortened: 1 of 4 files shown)


Monkeychip (approved):
Nice work. And good catch. Thank you for adding this.

ivana-mcconnell:
@arnav28 can you clarify what the message is for? In what context would the user see it? What do you mean by "custom plugin name"? I can help offer wording if I know a little more about that.

ivana-mcconnell:
Just met with Arnav and the summary of this is:

Oracle is the only plugin with a "customizable" name. However, the UI doesn't support customizable names just yet, so for now we will need the user to ensure that their Oracle plugin has the default name. In the future, we will add an additional field called "Plugin name" when Oracle is selected. Given the timing for 1.9, we could not do this without potentially introducing some regression bugs with the amount of refactoring that would be required. As a result (and because we assume that users will not change the name of the Oracle plugin), we'll show a message that tells them to ensure the plugin uses the default name. In future versions, we'll look at adding another dependent field to this form. 

@arnav28 Is that summary correct? 

Here is some suggested text: 

> Please ensure that your Oracle plugin has the default name of `vault-plugin-database-oracle`. Custom naming is not supported in the UI at this time. If you did not change the plugin name on install, disregard this warning.

PR #13088, opened 2021-11-08
adding CRT to main branch

claire-labry:
This is CRT being implemented for the main branch cc: @mdeggies

builds pass successfully: https://github.com/hashicorp/vault/actions/runs/1437011717

TODO (once this draft PR looks good/ready):

- [x]  Change the slack channel ID to vault-releases
- [x]  Change release_branches in ci.hcl to reflect this structure: 
```
 github {
    organization = "hashicorp"
    repository = "vault"
    release_branches = [
      "main",
      "release/1.6.x",
      "release/1.7.x",
      "release/1.8.x",
      "release/1.9.x"
    ]
  }
```
- [x]  Change to push to main in build.yml


.release/ci.hcl
```diff
@@ -0,0 +1,216 @@
...
+
+project "vault" {
+  team = "vault"
+  slack {
+    notification_channel = "CRF6FFKEW" // #vault-releases
+  }
+  github {
+    organization = "hashicorp"
+    repository = "vault"
+    release_branches = [
```
ncabatoff, line 11: Who uses this?

ncabatoff:
Is this missing vault.service?

.github/workflows/build.yml
```diff
@@ -0,0 +1,242 @@
...
+      # Push events on main branch
+      - main
+
+env:
+  PKG_NAME: "vault"
+  GO_TAGS: "ui"
+
+jobs:
+  get-product-version:
+    runs-on: ubuntu-latest
```
ncabatoff, line 16: Why does the ent version use `runs-on: [linux, medium]` instead?

.github/workflows/build.yml
```diff
@@ -0,0 +1,242 @@
...
+      - uses: actions/checkout@v2
+      - name: Setup go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - name: Setup node and yarn
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+          cache: 'yarn'
```
ncabatoff, line 75: There are discrepancies here with the ent PR (also no "Install Yarn" step) - is there a reason?

.github/workflows/build.yml
kpenfound: Good question! `linux, medium` are tags for our private hosted runners, which enterprise uses. OSS uses github hosted runners

.release/ci.hcl
kpenfound: This is read by the crt orchestrator, so it's the thing that decides if any workflows get kicked off beyond the build.yml.  Our general practice has been to list *all* crt enabled branches in the main branch's ci.hcl, though the only value that is actually functional here is "main". It's just listing the others so that humans know what other branches are enabled, but the other branches being enabled is dependent on their own ci.hcl. I've rambled a bit but hopefully that clears it up

ncabatoff: Yes, thanks

.github/workflows/build.yml
ncabatoff: ty

kpenfound:
@ncabatoff added vault.service, good catch

.github/workflows/build.yml
kpenfound: Similar to above, ent is using the private runners which do not have yarn installed globally, hence the difference. I should have called that out on the other PRs, but here's the changeset for ent 1.9 moving from github runners to self hosted https://github.com/hashicorp/vault-enterprise/pull/2351

ncabatoff: Gotcha

ncabatoff (approved):

PR #13090, opened 2021-11-09
UI/Custom empty state messages for transit and transform

hellobontempo:
Previous empty state message read the following for secret backends: 
Screen Shot 2021-11-08 at 9 11 08 PM

Updated copy for transform and transit: 
Screen Shot 2021-11-08 at 9 10 57 PM

Screen Shot 2021-11-08 at 9 12 18 PM

### DIFF
--- a/ui/app/helpers/options-for-backend.js
+++ b/ui/app/helpers/options-for-backend.js
@@ -87,6 +87,7 @@ const SECRET_BACKENDS = {
     displayName: 'Transformation',
     navigateTree: false,
     listItemPartial: 'secret-list/transform-list-item',
+    message: 'a transformation and a role',
     tabs: [
       {
         name: 'transformations',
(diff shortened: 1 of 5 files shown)


arnav28 (approved):

ui/app/helpers/options-for-backend.js
```diff
@@ -87,6 +87,7 @@ const SECRET_BACKENDS = {
     displayName: 'Transformation',
     navigateTree: false,
     listItemPartial: 'secret-list/transform-list-item',
+    message: 'create a transformation and a role',
```
hashishaw, line 90: small suggestion: `message` is a bit vague and couple mean anything. Maybe we call this `firstStep` instead?

hashishaw (approved):
Just one small suggestion, but looks great!

ui/app/helpers/options-for-backend.js
hellobontempo: @hashishaw - sounds good! thanks for the suggestion!

PR #13098, opened 2021-11-09
Raft peer removal bug

zofskeez:
When removing a raft storage peer from the cluster via the CLI, the UI would not reflect the change until the page was refreshed. The _findAll_ method returns everything from the store, which in this case included stale records since they were removed externally. After updating to use the _query_ method, only records from the response are returned.

**Testing**
- setup 2 vault servers using raft storage, one which will create a new cluster and the other who will join. Refer to [this document](https://docs.google.com/document/d/1mO9oWtWskUPw5ELatd8SHBf-D7mzRyT7SoDIszjRoRE/edit?usp=sharing) for steps on setting that up.
- hit /vault/storage/raft in the UI and verify both servers appear in Raft Storage list
- remove a peer from the cli `vault operator raft remove-peer name_of_raft_node`
- navigate away from route in UI and return (do not refresh since that will clear the store!)
- clicking on status in the navbar and then Raft Storage should reveal only 1 server in the list

### DIFF
--- a/changelog/13098.txt
+++ b/changelog/13098.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixes issue removing raft storage peer via cli not reflected in UI until refresh
+```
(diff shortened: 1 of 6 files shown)


hashishaw (approved):
Great work! 🚀

PR #13107, opened 2021-11-09
Raft Snapshot Restore Bug

zofskeez:
Attempting to restore a raft snapshot in the UI would result in an error thrown from _fetch_. Updating the application adapter to use _ember-fetch_ fixes the issue. 

image


hashishaw (approved):
Nice catch!

PR #13111, opened 2021-11-10
Return non-retryable errors on transit encrypt and decrypt failures

schultz-is:
This also address an issue found with transit encrypt and decrypt paths where a single batch item failure would abort the whole batch and simply return a top-level error.

### DIFF
--- a/builtin/logical/transit/path_decrypt.go
+++ b/builtin/logical/transit/path_decrypt.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/vault/sdk/framework"
-	"github.com/hashicorp/vault/sdk/helper/errutil"
 	"github.com/hashicorp/vault/sdk/helper/keysutil"
 	"github.com/hashicorp/vault/sdk/logical"
 )
@@ -142,14 +141,9 @@ func (b *backend) pathDecryptWrite(ctx context.Context, req *logical.Request, d 
 
 		plaintext, err := p.Decrypt(item.DecodedContext, item.DecodedNonce, item.Ciphertext)
 		if err != nil {
-			switch err.(type) {
-			case errutil.UserError:
-				batchResponseItems[i].Error = err.Error()
-				continue
-			default:
-				p.Unlock()
-				return nil, err
-			}
+			// Policy decryption returns InternalError in some cases which might not actually be considered internal errors.
+			batchResponseItems[i].Error = err.Error()
+			continue
 		}
 		batchResponseItems[i].Plaintext = plaintext
 	}
--- a/builtin/logical/transit/path_decrypt.go
+++ b/builtin/logical/transit/path_decrypt.go
@@ -169,6 +169,11 @@ func (b *backend) pathDecryptWrite(ctx context.Context, req *logical.Request, d 
 	} else {
 		if batchResponseItems[0].Error != "" {
 			p.Unlock()
+
+			if internalErrorInBatch {
+				return nil, errutil.InternalError{Err: batchResponseItems[0].Error}
+			}
+
 			return logical.ErrorResponse(batchResponseItems[0].Error), logical.ErrInvalidRequest
 		}
 		resp.Data = map[string]interface{}{
@@ -184,9 +189,9 @@ func (b *backend) pathDecryptWrite(ctx context.Context, req *logical.Request, d 
 	// to the user first.
 	switch {
 	case userErrorInBatch:
-		logical.RespondWithStatusCode(resp, req, http.StatusBadRequest)
+		return logical.RespondWithStatusCode(resp, req, http.StatusBadRequest)
 	case internalErrorInBatch:
-		logical.RespondWithStatusCode(resp, req, http.StatusInternalServerError)
+		return logical.RespondWithStatusCode(resp, req, http.StatusInternalServerError)
 	}
 
 	return resp, nil
--- a/builtin/logical/transit/path_decrypt.go
+++ b/builtin/logical/transit/path_decrypt.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"net/http"
 
 	"github.com/hashicorp/vault/sdk/framework"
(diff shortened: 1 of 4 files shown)


builtin/logical/transit/path_decrypt_test.go
```diff
@@ -64,74 +65,155 @@ func TestTransit_BatchDecryption(t *testing.T) {
...
+	resp, err = b.HandleRequest(context.Background(), req)
 	if err != nil || (resp != nil && resp.IsError()) {
 		t.Fatalf("err:%v resp:%#v", err, resp)
 	}
 
-	batchInput := []interface{}{
-		map[string]interface{}{"plaintext": "dGhlIHF1aWNrIGJyb3duIGZveA==", "context": "dGVzdGNvbnRleHQ="},
-		map[string]interface{}{"plaintext": "dGhlIHF1aWNrIGJyb3duIGZveA==", "context": "dGVzdGNvbnRleHQ="},
+	// Encrypt some values for use in test cases.
+	plaintextItems := []BatchRequestItem{
```
victorr, line 89: Not that it matters a whole lot, but instead of using `BatchRequestItem` couldn't you simply use an anonymous struct here, as the only fields that really matter for the test are the plaintext an the context?

builtin/logical/transit/path_decrypt_test.go
```diff
@@ -64,74 +65,155 @@ func TestTransit_BatchDecryption(t *testing.T) {
...
-	batchDecryptionResponseItems := resp.Data["batch_results"].([]DecryptBatchResponseItem)
-
-	plaintext := "dGhlIHF1aWNrIGJyb3duIGZveA=="
-	for _, item := range batchDecryptionResponseItems {
-		if item.Plaintext != plaintext {
-			t.Fatalf("bad: plaintext. Expected: %q, Actual: %q", plaintext, item.Plaintext)
-		}
+	encryptedItems := resp.Data["batch_results"].([]EncryptBatchResponseItem)
+
+	tests := []struct {
```
victorr, line 111: Great job adding good coverage for all combinations of batch results!

victorr (approved):
I added a suggestion but it looks good to me 👍

builtin/logical/transit/path_decrypt_test.go
schultz-is: Ah, so the reason I did this is mainly for convenience. In order for the encrypt request to work, the `batch_input` field has to be specifically a `[]interface{}`, which precludes map indexing without type assertions to a `map[string]interface{}`. I just made this as kind of a source of truth lookup table for easy test definitions. It lets me define test input and output without duplicating the strings inline in each test.

schultz-is:
After a discussion about partial batch error handling, there are some more changes coming with this PR before it is fully ready. Namely, we've decided to return HTTP errors on partial failures with non-retryable HTTP 400's taking precedence over retryable HTTP 500's. Some of the tests will be slightly reworked to accommodate this behavior.

schultz-is:
Alright, this should be ready for review again. Thanks y'all!

builtin/logical/transit/path_decrypt.go
```diff
@@ -170,6 +177,18 @@ func (b *backend) pathDecryptWrite(ctx context.Context, req *logical.Request, d
 	}
 
 	p.Unlock()
+
+	// Depending on the errors in the batch, different status codes should be returned. User errors
```
victorr, line 181: Shouldn't non-batch requests should benefit from this logic? The return statement in line 172 unconditionally results in a Bad Request response code, even though the error might just have been an internal one?

builtin/logical/transit/path_decrypt_test.go
victorr: I agree that using a struct is a good idea. I was suggesting using an anonymous struct rather than `BatchRequestItem` since the latter has many more fields that are not really relevant to the tests.

Something like:
```
plaintextItems := []struct{
    plaintext, context string
}{
...
```

builtin/logical/transit/path_decrypt_test.go
```diff
@@ -64,74 +66,163 @@ func TestTransit_BatchDecryption(t *testing.T) {
...
+				Path:      "decrypt/existing_key",
+				Storage:   s,
+				Data: map[string]interface{}{
+					"batch_input": tt.in,
+				},
+			}
+			resp, err = b.HandleRequest(context.Background(), req)
+
+			didErr := err != nil || (resp != nil && resp.IsError())
+			if didErr {
```
victorr, line 203: Would it be useful to test that the response have the expected status code?

victorr (changes requested):
The changes are good but I'm finishing the review with "request changes" as I think the non-batch encrypt/decrypt operations need the check for the right response code.

builtin/logical/transit/path_decrypt.go
schultz-is: Nice catch, I completely overlooked that case.

builtin/logical/transit/path_decrypt_test.go
schultz-is: It certainly would. I'll add some checks for that.

builtin/logical/transit/path_decrypt_test.go
schultz-is: Gotcha, I see what you mean now. I can certainly use an inline struct here to make the the data more directly relevant to tests. Thanks for explaining that!

builtin/logical/transit/path_decrypt_test.go
schultz-is: There is a check now that ensures returned HTTP responses contain the expected status code.

builtin/logical/transit/path_decrypt.go
schultz-is: Both encrypt and decrypt should now return an internal error if one was encountered in outside of batch input.

victorr (approved):
Looks good to me!

stevendpclark (approved):

schultz-is:
Thanks y'all!

victorr:
Closes #10842

PR #13114, opened 2021-11-10
Update `CHANGELOG.md` entry for #12621

vinay-gopalan:
Updates the Changelog entry for https://github.com/hashicorp/vault/pull/12621 which upgrades the `awsutil` package version.

PR #13116, opened 2021-11-10
Backport Changelog updates from #13114

vinay-gopalan:
Backports Changelog update from https://github.com/hashicorp/vault/pull/13114


calvn (approved):
Thanks for adding the missing entries!

calvn:
Just realized that this doesn't quite work. The entries have to come from the files within [changelog/](https://github.com/hashicorp/vault/tree/main/changelog) so as long as they're in there, the CL generation tool should take care of adding them when we're releasing GA.

PR #13132, opened 2021-11-12
OSS parts to support new kms_library configuration stanza.

stevendpclark:
- Add a new top level configuration stanza named kms_library with Vault's SharedConfig struct


victorr (approved):
👍

PR #13133, opened 2021-11-12
OIDC Auth Bug

### LINKED ISSUE #12436: OIDC UI based login not possible with MetaMask installed
**Describe the bug**
When the [MetaMask](https://metamask.io/) extension is installed and enabled for all sites in Chrome, OIDC based logins are not possible anymore.

**To Reproduce**
1. Run:
	  ```none
	  vault write auth/oidc/config \
	     oidc_client_id="your_client_id" \
	     oidc_client_secret="your_client_secret" \
	     default_role="your_default_role" \
	     oidc_discovery_url="https://login.microsoftonline.com/tenant_id/v2.0"
	  ```
2. Run:
	  ```none
	  vault write auth/oidc/role/your_default_role \
	     user_claim="email" \ 
	     allowed_redirect_uris="http://localhost:8250/oidc/callback,https://online_version_hostname:port_number/ui/vault/auth/oidc/oidc/callback"  \
	     groups_claim="groups" \
	     oidc_scopes="https://graph.microsoft.com/.default" \
	     policies=default
	  ```
3. Install [MetaMask in Chrome](https://chrome.google.com/webstore/detail/metamask/nkbihfbeogaeaoehlefnkodbefgpgknn) and enable it for all sites.
4. Head over to http://localhost:8200 and try logging in using OIDC in the UI

**Expected behavior**
One should be able to login using OIDC independent of browser extensions.

**Environment:**
  ```none
  Key             Value
  ---             -----
  Seal Type       shamir
  Initialized     true
  Sealed          false
  Total Shares    1
  Threshold       1
  Version         1.8.1
  Storage Type    inmem
  Cluster Name    vault-cluster-66939b64
  Cluster ID      f371d18c-979f-3489-d45d-a57a0e36b8ec
  HA Enabled      false
  ```
* Vault CLI Version (retrieve with `vault version`): `Vault v1.8.1 ('4b0264f28defc05454c31277cfa6ff63695a458d+CHANGES')`
* Server Operating System/Architecture: macOS Big Sur (Version 11.5.2)

Default dev server configs.
Server started using: `vault server -dev -dev-root-token-id=token`

**Additional context**

The `oidcState` variable used in [auth-jwt.js](https://github.com/hashicorp/vault/blob/main/ui/app/components/auth-jwt.js#L128) is getting minified to `t` and apparently MetaMask is overriding it.
This issue started with Vault version 1.8.0 and does not exist in earlier versions.

Screenshot 2021-08-25 at 20 56 17
Screenshot 2021-08-25 at 20 56 22

The error message in the UI:
```none
The callback from the provider did not supply all of the required parameters. Please click Sign In to try again. If the problem persists, you may want to contact your administrator.
```
Screenshot 2021-08-25 at 21 13 09

zofskeez:
Resolves #12436

When using the MetaMask chrome extension it was reported that the OIDC auth login workflow was not working. After investigating the [source code](https://github.com/MetaMask/metamask-extension/blob/29fa00a97bd505c273b51c735cd03100c319db58/app/scripts/contentscript.js#L189) it was discovered that the javascript being injected calls `window.postMessage` which was being handled inadvertently by the _auth-jwt_ component. 

To guard against unintended messages from the same origin a `source` property was added to the message data which is then verified in the event handler. Unit tests were added to verify the functionality.

### DIFF
--- a/ui/app/components/auth-jwt.js
+++ b/ui/app/components/auth-jwt.js
@@ -89,12 +89,18 @@ export default Component.extend({
     // start watching the popup window and the current one
     this.watchPopup.perform(oidcWindow);
     this.watchCurrent.perform(oidcWindow);
-    // wait for message posted from popup
-    const event = yield waitForEvent(thisWindow, 'message');
-    if (event.origin === thisWindow.origin && event.isTrusted) {
-      this.exchangeOIDC.perform(event.data, oidcWindow);
-    } else {
-      this.handleOIDCError();
+    // wait for message posted from oidc callback
+    // see issue https://github.com/hashicorp/vault/issues/12436
+    // ensure that postMessage event is from expected source
+    while (true) {
+      const event = yield waitForEvent(thisWindow, 'message');
+      if (event.origin !== thisWindow.origin || !event.isTrusted) {
+        return this.handleOIDCError();
+      }
+      if (event.data.source === 'oidc-callback') {
+        return this.exchangeOIDC.perform(event.data, oidcWindow);
+      }
+      // continue to wait for the correct message
     }
   }),
 
--- a/ui/app/routes/vault/cluster/oidc-callback.js
+++ b/ui/app/routes/vault/cluster/oidc-callback.js
@@ -9,7 +9,8 @@ export default Route.extend({
     let { auth_path: path, code, state } = this.paramsFor(this.routeName);
     let { namespaceQueryParam: namespace } = this.paramsFor('vault.cluster');
     path = window.decodeURIComponent(path);
-    let queryParams = { namespace, path, code, state };
+    const source = 'oidc-callback'; // required by event listener in auth-jwt component
+    let queryParams = { source, namespace, path, code, state };
     window.opener.postMessage(queryParams, window.origin);
   },
   renderTemplate() {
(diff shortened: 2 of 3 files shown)


zofskeez:
@chelshaw I spent some time debugging these [oidc acceptance test failures](https://app.circleci.com/pipelines/github/hashicorp/vault/23071/workflows/0f37dffb-3261-4f9e-a458-b3f1272345cb/jobs/337926) and wanted to bring them to your attention since you were working on them recently. I threw a pause in and found that a network request was failing with `Invalid client ID`.
image

After logging out the `clientId` return value from the `setupOidc` method I noticed that it was set to the following on failing runs: _Success! Data written to: identity/oidc/client/my-webapp-1636753641317_. There are 2 commands that are run to generate the `clientId` and it appears that the `lastLogOutput` value is the output from the first command:
```js
await consoleComponent.runCommands([
  `write identity/oidc/client/${webappName} redirect_uris="${redirect}" assignments="my-assignment" key="sigkey" id_token_ttl="30m" access_token_ttl="1h"`,
  `read -field=client_id identity/oidc/client/${webappName}`,
]);
```

It seems that there is a race condition in the `runCommands` method where responses are not output in the same order that they are run consistently. Waiting for a settled state before continuing the loop to run the console commands seems to have fixed it. Since this change will impact all usages of the helper I'm hoping that it doesn't break any other tests but I don't think it should :crossed_fingers:.

hashishaw (approved):
Great work, thanks for working on this!

PR #13137, opened 2021-11-12
Vault-2257: don't log token error on DR Secondary

swayne275:
- Stop gauge collector "expiration manager is nil" errors on DR Secondaries
- update log messages to be identifiable (two paths no longer yield the same message)

It looked like at least one of the two updated error messages was being hit on DR Secondaries, but after extensive testing I haven't found any evidence for that. Updating messages to identify if it does happen in the future.

### DIFF
--- a/vault/core_metrics.go
+++ b/vault/core_metrics.go
@@ -136,6 +136,11 @@ func (c *Core) metricsLoop(stopCh chan struct{}) {
 // TokenStore; there is one per method because an additional level of abstraction
 // seems confusing.
 func (c *Core) tokenGaugeCollector(ctx context.Context) ([]metricsutil.GaugeLabelValues, error) {
+	if c.IsDRSecondary() {
+		// there is no expiration manager on DR Secondaries
+		return []metricsutil.GaugeLabelValues{}, nil
+	}
+
 	// stateLock or authLock protects the tokenStore pointer
 	c.stateLock.RLock()
 	ts := c.tokenStore
@@ -147,6 +152,11 @@ func (c *Core) tokenGaugeCollector(ctx context.Context) ([]metricsutil.GaugeLabe
 }
 
 func (c *Core) tokenGaugePolicyCollector(ctx context.Context) ([]metricsutil.GaugeLabelValues, error) {
+	if c.IsDRSecondary() {
+		// there is no expiration manager on DR Secondaries
+		return []metricsutil.GaugeLabelValues{}, nil
+	}
+
 	c.stateLock.RLock()
 	ts := c.tokenStore
 	c.stateLock.RUnlock()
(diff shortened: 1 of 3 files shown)


vault/dynamic_system_view.go
```diff
@@ -65,7 +65,9 @@ func (e extendedSystemViewImpl) SudoPrivilege(ctx context.Context, path string,
 	// Resolve the token policy
 	te, err := e.core.tokenStore.Lookup(ctx, token)
 	if err != nil {
-		e.core.logger.Error("failed to lookup token", "error", err)
+		if !e.core.IsDRSecondary() {
```
HridoyRoy, line 68: Are we sure that we need this check here? `SudoPrivilege` is called by `handleCreateCommon`, which should not be called by DR secondaries, right?

vault/request_handling.go
```diff
@@ -141,7 +141,9 @@ func (c *Core) fetchACLTokenEntryAndEntity(ctx context.Context, req *logical.Req
 		var err error
 		te, err = c.tokenStore.Lookup(ctx, req.ClientToken)
 		if err != nil {
-			c.logger.Error("failed to lookup token", "error", err)
+			if !c.IsDRSecondary() {
```
HridoyRoy, line 144: Same question here -- I'm worried about using `!c.IsDRSecondary() ` too generously when we're deep in call chains. For example here, if we're sure that DR Secondaries would return an error during a lookup token, I think we should try to avoid the call to the `Lookup` call chain altogether.

vault/core_metrics.go
```diff
@@ -136,6 +136,11 @@ func (c *Core) metricsLoop(stopCh chan struct{}) {
 // TokenStore; there is one per method because an additional level of abstraction
 // seems confusing.
 func (c *Core) tokenGaugeCollector(ctx context.Context) ([]metricsutil.GaugeLabelValues, error) {
+	if c.IsDRSecondary() {
+		// there is no expiration manager on DR Secondaries
+		return []metricsutil.GaugeLabelValues{}, nil
```
HridoyRoy, line 141: This makes sense to me!

HridoyRoy (changes requested):
Suppressing the log messages looks good to me, but I had a quick follow-up question on the `IsDRSecondary` checks.

vault/dynamic_system_view.go
swayne275: i'm not sure. i don't have a stack trace, only the error which is the same in both places

in an ideal world, nothing that expects an `ExpirationManager` should be called on a DR Secondary, but we aren't there yet

if you're fairly confident i'll remove this bit! (but still update the log message)

vault/request_handling.go
swayne275: based on [this](https://hashicorp.atlassian.net/browse/VAULT-2257) it seems at least one of these two paths is being hit on the DR Secondary, resulting in the error

maybe the calls could be avoided in the first place

vault/core_metrics.go
swayne275: another option (for these 4 token gauge collectors) is to add some flag to `metricsInit` in `emitMetrics`

vault/dynamic_system_view.go
swayne275: if it's not a batch token and not the leader node this gets forwarded. presumably dr wouldn't get into code that should be handling this: c036a1a54

vault/request_handling.go
swayne275: i did a bunch of testing and didn't see anything that hit this path. the ticket error implies one of the two is being hit, but i haven't seen it so far. so at least if it happens on new versions we'll have a better idea what's going on

raskchanky (approved):

HridoyRoy (approved):
Looks good to me, thanks!

PR #13138, opened 2021-11-12
Revert "OSS parts to support new kms_library configuration stanza. (#13132)

stevendpclark:
This shouldn't be part of internalshared as it is specific to Vault.

This reverts commit 82d6662787c181b16bfdec315f96e4a81d123178.


victorr (approved):
👍

PR #13139, opened 2021-11-13
Add Rekey Example When Auto Unseal is Used

sharabinth:
Added an example to explicitly show how to perform a Rekey operation when the Vault cluster is using Auto Unseal.  This is placed as the second example. 
The existing example code combines with the PGP keys so added a simple example without the PGP keys.


taoism4504 (approved):
I've reviewed the content (not for technical accuracy) and it LGTM.

cipherboy (approved):

PR #13140, opened 2021-11-15
fix(docs-k8s-helm): changed server's podAntiAffinity labelSelector example to match helm default values

pree:
The example in the docs reffer to labels which are not set by the helm-chart. In the `values.yaml` the default is set to 
```
app.kubernetes.io/name: {{ template "vault.name" . }}
app.kubernetes.io/instance: "{{ .Release.Name }}"
component: server
```

I've updated the docs to represent that.


jasonodonnell (approved):

pree:
Is there anything thats blocking a merge?

hsimon-hashicorp:
@pree we have some flaky tests, but I've re-run those and merged it in. :)

PR #13142, opened 2021-11-15
Add extra debugging to help identify failures within mssql test

stevendpclark:
Even with the attempted fix in PR #13071, we are still occasionally seeing failures within CircleCI. This won't fix the problem
but perhaps will provide us additional insight into the issue when it does occur again.

### DIFF
--- a/sdk/database/dbplugin/v5/testing/test_helpers.go
+++ b/sdk/database/dbplugin/v5/testing/test_helpers.go
@@ -25,20 +25,22 @@ func getRequestTimeout(t *testing.T) time.Duration {
 // AssertInitializeCircleCiTest help to diagnose CircleCI failures within AssertInitialize for mssql tests failing
 // with "Failed to initialize: error verifying connection ..."
 func AssertInitializeCircleCiTest(t *testing.T, db dbplugin.Database, req dbplugin.InitializeRequest) dbplugin.InitializeResponse {
+	t.Helper()
 	maxAttempts := 5
 	var resp dbplugin.InitializeResponse
 	var err error
+
 	for i := 1; i <= maxAttempts; i++ {
 		resp, err = verifyInitialize(t, db, req)
-		if err != nil {
+		switch {
+		case err != nil:
 			t.Logf("Failed AssertInitialize attempt: %d with error:\n%+v\n", i, err)
-		} else {
-			if i > 1 {
-				t.Fatalf("AssertInitialize worked the %d time around with a 1 second sleep, but failed originally", i)
-			}
+			time.Sleep(1 * time.Second)
+		case i == 1:
 			return resp
+		default:
+			t.Fatalf("AssertInitialize worked the %d time around with a 1 second sleep, but failed originally", i)
 		}
-		time.Sleep(1 * time.Second)
 	}
 
 	t.Fatalf("Failed to initialize: %+v", err)
@@ -46,6 +48,7 @@ func AssertInitializeCircleCiTest(t *testing.T, db dbplugin.Database, req dbplug
 }
 
 func AssertInitialize(t *testing.T, db dbplugin.Database, req dbplugin.InitializeRequest) dbplugin.InitializeResponse {
+	t.Helper()
 	resp, err := verifyInitialize(t, db, req)
 	if err != nil {
 		t.Fatalf("Failed to initialize: %s", err)
@@ -54,8 +57,6 @@ func AssertInitialize(t *testing.T, db dbplugin.Database, req dbplugin.Initializ
 }
 
 func verifyInitialize(t *testing.T, db dbplugin.Database, req dbplugin.InitializeRequest) (dbplugin.InitializeResponse, error) {
-	t.Helper()
-
 	ctx, cancel := context.WithTimeout(context.Background(), getRequestTimeout(t))
 	defer cancel()
 
--- a/sdk/database/dbplugin/v5/testing/test_helpers.go
+++ b/sdk/database/dbplugin/v5/testing/test_helpers.go
@@ -23,7 +23,7 @@ func getRequestTimeout(t *testing.T) time.Duration {
 }
 
 // AssertInitializeCircleCiTest help to diagnose CircleCI failures within AssertInitialize for mssql tests failing
-// with "Failed to initialize: error verifying connection ..."
+// with "Failed to initialize: error verifying connection ...". This will now mark a test as failed instead of being fatal
 func AssertInitializeCircleCiTest(t *testing.T, db dbplugin.Database, req dbplugin.InitializeRequest) dbplugin.InitializeResponse {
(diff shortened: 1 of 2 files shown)


sdk/database/dbplugin/v5/testing/test_helpers.go
```diff
@@ -22,19 +22,46 @@ func getRequestTimeout(t *testing.T) time.Duration {
...
+// AssertInitializeCircleCiTest help to diagnose CircleCI failures within AssertInitialize for mssql tests failing
+// with "Failed to initialize: error verifying connection ..."
+func AssertInitializeCircleCiTest(t *testing.T, db dbplugin.Database, req dbplugin.InitializeRequest) dbplugin.InitializeResponse {
+	maxAttempts := 5
+	var resp dbplugin.InitializeResponse
+	var err error
+	for i := 1; i <= maxAttempts; i++ {
+		resp, err = verifyInitialize(t, db, req)
+		if err != nil {
+			t.Logf("Failed AssertInitialize attempt: %d with error:\n%+v\n", i, err)
```
ncabatoff, line 34: Slightly more idiomatic to use `continue` here, or a `switch` statement,  instead of `else`.  This is fine though.

sdk/database/dbplugin/v5/testing/test_helpers.go
```diff
@@ -22,19 +22,46 @@ func getRequestTimeout(t *testing.T) time.Duration {
...
+		} else {
+			if i > 1 {
```
stevendpclark, line 36: Crappers I lost a fix here as well, this if shouldn't be here.

ncabatoff:
Another idea to consider: before the initialize, try doing say 10s of Pings at 1s intervals and report on the results.

ncabatoff (approved):

sdk/database/dbplugin/v5/testing/test_helpers.go
```diff
@@ -22,19 +22,46 @@ func getRequestTimeout(t *testing.T) time.Duration {
...
 
-	ctx, cancel := context.WithTimeout(context.Background(), getRequestTimeout(t))
-	defer cancel()
+	t.Fatalf("Failed to initialize: %+v", err)
+	return resp
+}
 
-	resp, err := db.Initialize(ctx, req)
+func AssertInitialize(t *testing.T, db dbplugin.Database, req dbplugin.InitializeRequest) dbplugin.InitializeResponse {
+	resp, err := verifyInitialize(t, db, req)
```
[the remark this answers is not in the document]
stevendpclark: If I understand that method call properly it should be added back here, added technically to AssertInitializeCircleCiTest and removed from the internal verifyInitialize method?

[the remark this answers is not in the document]
stevendpclark: Fixed up in review feedback commit.

sdk/database/dbplugin/v5/testing/test_helpers.go
```diff
@@ -22,19 +22,47 @@ func getRequestTimeout(t *testing.T) time.Duration {
...
+	for i := 1; i <= maxAttempts; i++ {
+		resp, err = verifyInitialize(t, db, req)
+		switch {
+		case err != nil:
+			t.Logf("Failed AssertInitialize attempt: %d with error:\n%+v\n", i, err)
+			time.Sleep(1 * time.Second)
+		case i == 1:
+			return resp
+		default:
+			t.Fatalf("AssertInitialize worked the %d time around with a 1 second sleep, but failed originally", i)
```
swayne275, line 42: should this be a `t.Fatalf`? that would end the test. would we instead like to log it and return the successful response, or do we want to stop the test if it didn't work the first time?

sdk/database/dbplugin/v5/testing/test_helpers.go
```diff
@@ -22,19 +22,47 @@ func getRequestTimeout(t *testing.T) time.Duration {
...
+		}
+	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), getRequestTimeout(t))
-	defer cancel()
+	t.Fatalf("Failed to initialize: %+v", err)
```
swayne275, line 46: this will stop the test, but the next line has it returning the response. i believe that return line would not get hit

swayne275 (changes requested):
requesting changes for those comments, but disregard if those are intentional and i'm missing something

sdk/database/dbplugin/v5/testing/test_helpers.go
stevendpclark: You are absolutely correct, the return resp was just to get the code too compile. a return nil doesn't work either as the return type is a struct not a pointer to the struct. 

Perhaps a panic statement instead, not sure how to rework the code to make the intention clearer though.

sdk/database/dbplugin/v5/testing/test_helpers.go
stevendpclark: Yes the t.Fatalf is intentional. Nick and I agreed, we still wanted to fail the test to provide visibility when it would occur, with the extra debugging hopefully providing additional insights.

stevendpclark:
> Another idea to consider: before the initialize, try doing say 10s of Pings at 1s intervals and report on the results.

Spoke to Nick about this offline, for now I will not implement this bit of feedback as I'm worried it might influence the failures we are seeing within the AssertInitialize calls.

sdk/database/dbplugin/v5/testing/test_helpers.go
```diff
@@ -22,19 +22,47 @@ func getRequestTimeout(t *testing.T) time.Duration {
...
+	for i := 1; i <= maxAttempts; i++ {
+		resp, err = verifyInitialize(t, db, req)
+		if err != nil {
+			t.Errorf("Failed AssertInitialize attempt: %d with error:\n%+v\n", i, err)
+			time.Sleep(1 * time.Second)
+			continue
+		}
+
+		if i > 1 {
+			t.Logf("AssertInitialize worked the %d time around with a 1 second sleep", i)
```
swayne275, line 42:
```suggestion
			t.Errorf("AssertInitialize worked the %d time around with a 1 second sleep", i)
```

?

this is what we discussed in slack i believe

swayne275: might also be good to comment that we want the test to fail if it doesn't work the first time?

swayne275 (changes requested):

ncabatoff (approved):

sdk/database/dbplugin/v5/testing/test_helpers.go
swayne275: ohhhh oops missed the `Errorf` above!

swayne275 (approved):

PR #13154, opened 2021-11-16
identity/oidc: Adds section to 1.9 upgrade guide for ACL policy requirements

austingebauer:
This PR adds a section to the 1.9 upgrade guide to call out the ACL policy requirements for the OIDC provider feature. Existing deployments of Vault that are upgraded to 1.9 will not have the required ACL rule that was added to the default policy (see [policy_store.go#L154-L157](https://github.com/hashicorp/vault/blob/main/vault/policy_store.go#L154-L157)) for 1.9.

This functions as designed and is not a new discovery. We wanted the Authorization Endpoint to be an authenticated endpoint. This guide is necessary because changes to the default policy are not applied during upgrades.

### DIFF
--- a/website/content/docs/upgrading/upgrade-to-1.9.0.mdx
+++ b/website/content/docs/upgrading/upgrade-to-1.9.0.mdx
@@ -14,6 +14,26 @@ official guidance until the release has been completed.
 This page contains the list of deprecations and important or breaking changes
 for Vault 1.9.0 compared to 1.8. Please read it carefully.
 
+## OIDC Provider
+
+Vault 1.9.0 introduced the ability for Vault to be an OpenID Connect (OIDC) identity
+provider. To support the feature, Vault's [default policy](https://www.vaultproject.io/docs/concepts/policies#default-policy)
+was modified to include an ACL rule for its Authorization Endpoint. Due to the handling
+of Vault's default policy during upgrades, existing deployments of Vault that are upgraded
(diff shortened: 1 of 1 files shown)


vinay-gopalan (approved):

website/content/docs/upgrading/upgrade-to-1.9.0.mdx
```diff
@@ -14,6 +14,26 @@ official guidance until the release has been completed.
...
+## OIDC Provider
+
+Vault 1.9.0 introduced the ability for Vault to be an OpenID Connect (OIDC) identity
+provider. To support the feature, Vault's [default policy](https://www.vaultproject.io/docs/concepts/policies#default-policy)
+was modified to include an ACL rule for its Authorization Endpoint. Due to the handling
+of Vault's default policy during upgrades, existing deployments of Vault that are upgraded
+to 1.9.0 will not have this required ACL rule.
+
+If you're upgrading to 1.9.0 and want to use the new OIDC provider feature, the following
+ACL rule must be applied to the default policy **or** a policy associated with the Vault
```
calvn, line 26:
```suggestion
ACL rule must be added to the default policy **or** a policy associated with the Vault
```

To explicitly call out that we're adding and not replacing the policy.

calvn (approved):

taoism4504 (commented):
doc LGTM

PR #13155, opened 2021-11-16
update custom headers to mention 1.9 is required

rowansmithhc:
Per https://github.com/hashicorp/vault/blob/main/CHANGELOG.md#190-rc1 the custom response headers are a new feature introduced in 1.9, meaning we should explicitly call out this version requirement in documentation, otherwise users of earlier versions of Vault will unable to use the functionality and may consider it a bug.

### DIFF
--- a/website/content/docs/configuration/listener/tcp.mdx
+++ b/website/content/docs/configuration/listener/tcp.mdx
@@ -22,7 +22,7 @@ specify [`api_addr`][api-addr] and [`cluster_addr`][cluster-addr] so Vault will
 advertise the correct address to other nodes.
 
 ## Listener's custom response headers
-Vault supports defining custom HTTP response headers for the root path (`/`) and also on API endpoints (`/v1/*`).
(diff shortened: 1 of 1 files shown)


website/content/docs/configuration/listener/tcp.mdx
```diff
@@ -22,7 +22,7 @@ specify [`api_addr`][api-addr] and [`cluster_addr`][cluster-addr] so Vault will
 advertise the correct address to other nodes.
 
 ## Listener's custom response headers
-Vault supports defining custom HTTP response headers for the root path (`/`) and also on API endpoints (`/v1/*`).
+As of version 1.9 Vault supports defining custom HTTP response headers for the root path (`/`) and also on API endpoints (`/v1/*`).
```
taoism4504, line 25:
```suggestion
As of version 1.9, Vault supports defining custom HTTP response headers for the root path (`/`) and also on API endpoints (`/v1/*`).
```

taoism4504 (approved):
One nit, but otherwise LGTM

yhyakuna (approved):
LGTM

Issue #11152, opened 2021-03-19
Attempting to authenticate to "login" endpoint for non-existent mount path and no X-Vault-Token erroneously calls fetchACLTokenEntryAndEntity, resulting in "missing client token" instead of "invalid mount path"


lambertryanPFG:
**Describe the bug**
Vault returns `{ "error": [ "missing client token" ] } ` when you attempt to hit a login endpoint on a mount path that doesn't exist using an authentication method that doesn't require a pre-existing auth token. 
 
request_handling.go has a method handleCancelableRequest that has this logic:

```
if c.router.LoginPath(ctx, req.Path) {
   resp, auth, err = c.handleLoginRequest(ctx, req)
} else {
   resp, auth, err = c.handleRequest(ctx, req)
}
```

`router.LoginPath` takes the endpoint path and uses the go-radix library to determine if it's a valid login path (as defined in the unit test as an existing mount path appending with one of the following).

```
n := &NoopBackend{
   Login: []string{
      "login",
      "oauth/*",
   },
}
```

If the mount path doesn't exist, the call to `LongestPrefix` fails to figure out if the path contains "login" or "oauth/*" and just returns false by default, causing `LoginPath` to return false. Here is the relevant code for `LoginPath`:

```
func (r *Router) LoginPath(ctx context.Context, path string) bool {
   ns, err := namespace.FromContext(ctx)
   if err != nil {
      return false
   }
   adjustedPath := ns.Path + path
   r.l.RLock()
   mount, raw, ok := r.root.LongestPrefix(adjustedPath)
   r.l.RUnlock()
   if !ok {
      return false
   }
```

The problem is that `LoginPath` returning false causes the above "if" clause in `handleCancelableRequest` to send the request to `handleRequest` instead of `handleLoginRequest`. Both methods call a `checkToken` method but with different parameters. `handleLoginRequest` calls it with a boolean flag "unauth" set to true, while `handleRequest` calls it with that flag set to false.

`auth, te, ctErr := c.checkToken(ctx, req, false) // The call from handleRequest`

Inside of checkToken is the following logic:

```
// Even if unauth, if a token is provided, there's little reason not to
// gather as much info as possible for the audit log and to e.g. control
// trace mode for EGPs.
if !unauth || (unauth && req.ClientToken != "") {
   acl, te, entity, identityPolicies, err = c.fetchACLTokenEntryAndEntity(ctx, req)
   // In the unauth case we don't want to fail the command, since it's
[...]

PR #13162, opened 2021-11-16
Authenticate to "login" endpoint for non-existent mount path bug

akshya96:
Fixes issue https://github.com/hashicorp/vault/issues/11152

Testing:
+ curl --request POST --cacert cert.pem --cert vault-cert-certificate.pem --key vault-cert-private-key.pem --data '{"name": "vault-cert"}' https://127.0.0.1:8200/v1/auth/akshya/login
{"errors":["permission denied"]}
+ curl -w '%{http_code}\n' -H Accept:application/json -H Content-Type:application/json --cacert cert.pem --cert vault-cert-certificate.pem --key vault-cert-private-key.pem --data '{"name": "vault-cert"}' https://127.0.0.1:8200/v1/auth/akshya/login -s -o /dev/null
403


vault/request_handling.go
```diff
@@ -126,7 +126,7 @@ func (c *Core) fetchACLTokenEntryAndEntity(ctx context.Context, req *logical.Req
 
 	// Ensure there is a client token
 	if req.ClientToken == "" {
-		return nil, nil, nil, nil, &logical.StatusBadRequest{Err: "missing client token"}
+		return nil, nil, nil, nil, logical.ErrPermissionDenied
```
HridoyRoy, line 129: I'm still of the opinion that we don't want to return the generic `PermissionDenied` error is actually less descriptive than a `missing client token` message, which more accurately captures what is going on here. I'm happy to let this go in if others think differently, but I'm of the opinion that we should more towards more descriptive error messages in each clause as opposed to consistently throwing a message like `logical.ErrPermissionDenied` no matter which error clause of a function the error occurs in.

pmmukh: Hmm I think I'm also in favor of keeping the more descriptive `missing client token` error here. Also, I don't think this really addresses the issue reported, imo we would either want to change https://github.com/hashicorp/vault-enterprise/blob/main/vault/router.go#L810 to return true for a login attempt on a non-existent mount, or close the issue out as a won't do.

vault/router_test.go
```diff
@@ -373,6 +373,7 @@ func TestRouter_LoginPath(t *testing.T) {
 		{"auth/foo/bar", false},
 		{"auth/foo/login", true},
 		{"auth/foo/login/", false},
+		{"auth/invalid/login/", false},
```
pmmukh, line 376: should this be `auth/invalid/login` , since `auth/foo/login/` is not a valid path, and so I'm guessing the trailing slash makes the login path invalid even if `invalid` were to be actually valid like `foo` ?

vault/request_handling.go
pmmukh: ( retracting my comment around the PR not addressing the issue, after going through slack conversation )

pmmukh (approved):
one test nit, lgtm!

kalafut (changes requested):
I think if we're changing an API response code (in this case from 400 to 403) this should have a CHANGES entry in the changelog. Also, what is the target milestone? I'd suggest 1.10. WDYT @ncabatoff

http/help_test.go
```diff
@@ -14,8 +15,9 @@ func TestHelp(t *testing.T) {
 	TestServerAuth(t, addr, token)
 
 	resp := testHttpGet(t, "", addr+"/v1/sys/mounts?help=1")
-	if resp.StatusCode != http.StatusBadRequest {
-		t.Fatal("expected bad request with no token")
+	fmt.Println(resp.StatusCode)
```
hghaf099, line 18: Is this needed?

hghaf099 (approved):

http/help_test.go
akshya96: Will remove the print statement. Was using it to debug

ccapurso (approved):
Others have already commented on things that I was planning to focus on. With that said, looks good to me.

ncabatoff:
> I think if we're changing an API response code (in this case from 400 to 403) this should have a CHANGES entry in the changelog. Also, what is the target milestone? I'd suggest 1.10. WDYT @ncabatoff

@kalafut That works for me.

akshya96:
> I think if we're changing an API response code (in this case from 400 to 403) this should have a CHANGES entry in the changelog. Also, what is the target milestone? I'd suggest 1.10. WDYT @ncabatoff

Added entry to changelog

kalafut (approved):

PR #13166, opened 2021-11-16
KV automatic delete state issue in UI

zofskeez:
When an automatic deletion time for a secret passes during a UI session the state was not properly updating to reflect the change. Instead of displaying the placeholder to indicate that the version has been deleted the secret is displayed with missing data.
image

The _secret-edit_ route is fetching the version data but since the `deletionTime` does not change the `deleted` computed property on the _secret-v2-version_ model was returning the cached value. After converting the model to a native class and using a getter the deleted value is calculated each time it is accessed which fixes the state issue. 

image

To test:
- in a terminal run `vault kv put secret/testversions key="value" && vault kv metadata put -delete-version-after="30s" secret/testversions`
- before 30 seconds view the secret in the UI
- after 30 seconds go back to the secrets list (do not refresh) and then back to the secret again
- the version deleted placeholder copy should display.

I started working on an acceptance test but since we are dealing with time elapsing I became concerned that it might end up flaky. If we think it's important to cover this case I'm happy to work on it more.

### DIFF
--- a/ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
+++ b/ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
@@ -155,12 +155,9 @@ export default Route.extend(UnloadModelRoute, {
     try {
       if (secretModel.failedServerRead) {
         // we couldn't read metadata, so we want to directly fetch the version
-
-        versionModel =
-          this.store.peekRecord('secret-v2-version', JSON.stringify(versionId)) ||
-          (await this.store.findRecord('secret-v2-version', JSON.stringify(versionId), {
-            reload: true,
-          }));
+        versionModel = await this.store.findRecord('secret-v2-version', JSON.stringify(versionId), {
+          reload: true,
+        });
       } else {
         // we may have previously errored, so roll it back here
         version.rollbackAttributes();
--- a/changelog/13166.txt
+++ b/changelog/13166.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixes issue with placeholder not displaying for automatically deleted secrets when deletion time has passed
+```
--- a/ui/app/models/secret-v2-version.js
+++ b/ui/app/models/secret-v2-version.js
@@ -1,20 +1,21 @@
 import { belongsTo, attr } from '@ember-data/model';
-import Secret from './secret';
-import { computed } from '@ember/object';
+import SecretModel from './secret';
 
-export default Secret.extend({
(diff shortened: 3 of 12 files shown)


ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
```diff
@@ -155,12 +155,9 @@ export default Route.extend(UnloadModelRoute, {
...
         // we couldn't read metadata, so we want to directly fetch the version
-
-        versionModel =
-          this.store.peekRecord('secret-v2-version', JSON.stringify(versionId)) ||
-          (await this.store.findRecord('secret-v2-version', JSON.stringify(versionId), {
-            reload: true,
-          }));
+        versionModel = await this.store.findRecord('secret-v2-version', JSON.stringify(versionId), {
+          reload: true,
+        });
```
zofskeez, line 160: @Monkeychip I made this change to always fetch the version record to avoid state issues when updating via the CLI which broke this [acceptance test](https://app.circleci.com/pipelines/github/hashicorp/vault/23178/workflows/c453af19-0ccd-4681-bd35-e3a3f0fc1036/jobs/340582) and I wanted to get your take on it.

In the test this policy does not have _read_ permission for the metadata endpoint.
```js
let V2_POLICY_NO_LIST = `
  path "${enginePath}/metadata/*" {
    capabilities = ["update","delete"]
  }
  path "${enginePath}/data/*" {
    capabilities = ["create", "read", "update", "delete"]
  }
`;
```
A new version is created successfully but the subsequent fetch fails. I'm wondering if I need to revert my change, add _read_ to the policy in the test or update the assertion to confirm that the new version does not display after being created?

zofskeez: After discussions with Angel and Ivana it was decided to disable the _Create new version_ action when the user does not have read access to the secret metadata which was addressed [in this commit](https://github.com/hashicorp/vault/pull/13166/commits/3aa225ca28d398207e3c750acb9ba76b893e578e). A tooltip is displayed on hover to indicate why the action is not available. I updated the failing acceptance test to assert that the action is now disabled.
Screenshot from 2021-11-22 14-25-55

Monkeychip (approved):
Well done 👏 Thanks for taking all the extra steps here too! Oh, don't forget the labels too. UI and not sure if this is getting backported or not. Sounds like something that should be backported to 1.9 which is where the KV work was done. If you backport farther than that you might hit a lot of changes that make this merge difficult.

PR #13168, opened 2021-11-16
Revert more downgrades from #12975.

ncabatoff:


calvn (approved):
👍 , might want to run `go mod tidy` to clean up go.sum

PR #13177, opened 2021-11-17
Form field component ttl picker not initially enabling

zofskeez:
The _TtlPicker2_ instance in the _FormField_ component was missing the `initialEnabled` argument which was causing the toggle to be unchecked and the value to be hidden initially. Toggling the input would reveal the set value.
image
After adding the argument the initial state is properly reflected. 
image

### DIFF
--- a/changelog/13177.txt
+++ b/changelog/13177.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixes issue with automate secret deletion value not displaying initially if set in secret metadata edit view
+```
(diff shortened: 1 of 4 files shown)


Monkeychip (approved):
Nice work. Remember the labels (UI) and then a milestone label of 1.9.1

zofskeez:
Thanks @Monkeychip. After looking into your comment on Slack I noticed that the toggle was enabled for the default `0s` value which I fixed and added a test.

PR #13181, opened 2021-11-17
fixing the injector.webhookAnnotations annotation

martinhristov90:
Fixing the `injector.webhookAnnotations` annotation in Helm install command which should contain the name of the `Certificate` resource, instead of the name of the underlying K8S secret which keeps the TLS cert.

[More info](https://cert-manager.io/docs/concepts/ca-injector/#injecting-ca-data-from-a-certificate-resource)


tvoran (approved):
Thanks!

PR #13189, opened 2021-11-17
Use new auth modules + Login method in Go client docs

digivava:
We've updated the [vault-examples](https://github.com/hashicorp/vault-examples) code snippets repo to use the new Login pattern introduced to the Go client library. This is me updating the snippets on the website to match. 

I also shortened many of the comment lines to make the code samples look tidier.


taoism4504 (approved):
Val's got the code reviewed already covered, so I'm simply approving this so she can get this into the documentation.

PR #13191, opened 2021-11-17
use GetOkErr in patch handler so type coercion errors result in error response

ccapurso:
Using `GetOkErr` in `HandlePatchOperation`, allows request handlers to return errors when a request a field's type does not match (i.e. parse errors). `GetOkErr` will error if a field does not exist in the provided `FieldSchema`. `HandlePatchOperation` will check the `FieldSchema` first to determine if the field needs to be skipped to prevent unwanted errors. While this check is redundant, it seems less worrisome than attempting to duplicate the type checking that `GetOkErr` performs.

Related test in `vault-plugin-secrets-kv` PR [#57](https://github.com/hashicorp/vault-plugin-secrets-kv/pull/57): https://github.com/hashicorp/vault-plugin-secrets-kv/blob/c920d7840602f8342d3f31606442f2e9e1333304/path_metadata_test.go#L445-L475


swayne275:
why do we want to do this for unknown keys?

ccapurso:
> why do we want to do this for unknown keys?

@swayne275, you're right. I don't think we do. In most cases, Vault's request handlers silently ignore unknown fields. `GetOkErr` is useful in that it has type handling. The first thing it does, however, is return an error if the key is not present.

I will refactor that PATCH handler logic a bit to first check for the field in `FieldData.Schema`. This is redundant in that `GetOkErr` also does this but it seems less concerning to duplicate that check than the type checking logic that occurs afterward.

swayne275:
@ccapurso could you perhaps write a similar helper function to `GetOkErr` that doesn't error for unknown fields (but does the usual stuff like type and !ok if not present)? do you think that would be generally useful, or is this sort of a one-off case?

ccapurso:
> do you think that would be generally useful, or is this sort of a one-off case?

There is a function that does that but specifically for validation purposes, see [field_data.go:32](https://github.com/hashicorp/vault/blob/e27671a740122fbd8d8a5e42a4760d69c6d93c9a/sdk/framework/field_data.go#L32). I think that the logic in this PR might be a one-off case in that it partly does what `Validate` does but also acts as a filter.

HridoyRoy (commented):
I'm not sure I understand how this differs from the implementation of `GetOk` . `GetOk` does not panic if a field is not in the schema. It only panics if `GetOkErr` returns an error. 

Perhaps you could include a small unit test to demonstrate the use-case in which the `GetOk` method doesn't do what we want it to do?

ccapurso:
@HridoyRoy, this is meant to avoid the panic so that we can use the error in the response. KVv2 is the only logical backend that supports the `PatchOperation` thus far and testing for this concept is done there. I linked a test that specifically validates this functionality in the PR description above.

HridoyRoy (approved):
Thanks for linking the test in the PR description! Lgtm!

PR #4783, opened 2018-06-18
Alibaba Object Storage support

bonifaido:
Hi, here at @banzaicloud we are trying to run Vault on Alibaba Cloud, and thus we would like to use the cloud provider offered object storage backend. It seemed easy to implement, so we have added support for Alibaba Cloud's Object Storage Service: https://www.alibabacloud.com/product/oss
FYI: This is our first non-trivial Vault contribution :)
Please feel free to comment, request changes etc.

PR #13198, opened 2021-11-18
Unhide or remove docs sidebar elements

kalafut:
A few sidebar elements are hidden for unknown reasons. If we have a
reason to keep them hidden (vs deleting the element and associated docs),
maybe we could add `"_comment":"Hidden because <reason>"` to them.

A few other elements were definitely obsolete so I've removed them.


website/data/docs-nav-data.json
```diff
@@ -283,8 +283,7 @@
           },
           {
             "title": "Alicloud OSS",
-            "path": "configuration/storage/alicloudoss",
-            "hidden": true
+            "path": "configuration/storage/alicloudoss"
```
mladlow, line 286: I think this might be hidden because we're worried about even half supporting it, but if the code exists I think we should probably unhide it.

kalafut: It looks like we missed updating the sidebar originally and never noticed until now.  https://github.com/hashicorp/vault/pull/4783

mladlow (approved):
Maybe give @HridoyRoy a day or so to chime in but I think I'm ok with this.

HridoyRoy (approved):

PR #13200, opened 2021-11-18
Unify HTTPResponseWriter and StatusHeaderResponseWriter

hghaf099:
There are a number of interfaces that implement http.ResponseWriter. For that, wrapping http.ResponseWriter needs a lot of care. In this PR, I tried to unify HTTPResponseWriter and StatusHeaderResponseWriter. The former is used in the logical layer in order to enable streaming. The latter is used to set custom response header. 
StatusHeaderResponseWriter is instantiated in the http layer which wraps http.ResponseWriter. Then, HTTPResponseWriter wraps the already wrapped ResponseWriter. This causes issues as for example in the `sys/monitor`, the `http.Flusher` cannot be asserted. 
This PR aims to unify the two wrappers and combining the functionality of the two to fix the above issue.

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -906,8 +906,11 @@ func request(core *vault.Core, w http.ResponseWriter, rawReq *http.Request, r *l
 	// additional output. Headers have already been sent. If the response writer
 	// is set but has not been written to it likely means there was some kind of
 	// error
-	if r.ResponseWriter != nil && r.ResponseWriter.Written() {
-		return nil, true, false
+	if r.ResponseWriter != nil {
+		w, ok := (*r.ResponseWriter).(logical.WrappingResponseWriter)
+		if ok && w.Written() {
+			return nil, true, false
+		}
 	}
 
 	if respondErrorCommon(w, r, resp, err) {
--- a/http/handler.go
+++ b/http/handler.go
@@ -17,14 +17,12 @@ import (
 	"net/url"
 	"os"
 	"regexp"
-	"strconv"
 	"strings"
 	"time"
 
 	"github.com/NYTimes/gziphandler"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-cleanhttp"
-	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-secure-stdlib/parseutil"
 	"github.com/hashicorp/go-sockaddr"
 	"github.com/hashicorp/vault/helper/namespace"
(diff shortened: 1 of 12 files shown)


http/handler.go
```diff
@@ -389,25 +304,21 @@ func wrapGenericHandler(core *vault.Core, h http.Handler, props *vault.HandlerPr
...
 			listenerCustomHeaders := core.GetListenerCustomResponseHeaders(la)
 			if listenerCustomHeaders != nil {
-				w = &statusHeaderResponseWriter{
-					wrapped:     w,
-					logger:      core.Logger(),
-					wroteHeader: false,
-					statusCode:  200,
-					headers:     listenerCustomHeaders.StatusCodeHeaderMap,
-				}
+				nw.SetHeaders(listenerCustomHeaders.StatusCodeHeaderMap)
```
ncabatoff, line 315: Why introduce a method for this instead of just making it an argument to the constructor?

http/logical.go
```diff
@@ -199,7 +199,7 @@ func buildLogicalRequestNoAuth(perfStandby bool, w http.ResponseWriter, r *http.
 		req.HTTPRequest = r
 	}
 	if responseWriter != nil {
-		req.ResponseWriter = logical.NewHTTPResponseWriter(responseWriter)
+		req.ResponseWriter = logical.NewStatusHeaderResponseWriter(responseWriter)
```
ncabatoff, line 202: Why do we need to re-wrap?  Why can't we just assign responseWriter directly?

sdk/logical/response.go
```diff
@@ -192,30 +193,127 @@ func RespondWithStatusCode(resp *Response, req *Request, code int) (*Response, e
...
-// HTTPResponseWriter is optionally added to a request object and can be used to
-// write directly to the HTTP response writer.
-type HTTPResponseWriter struct {
+type WrappingResponseWriter interface {
 	http.ResponseWriter
-	written *uint32
+	Wrapped() http.ResponseWriter
+	SetHeaders(h map[string][]*CustomHeader)
+	GetHeaders() map[string][]*CustomHeader
+	StatusCode() int
```
ncabatoff, line 201: Is this actually used?

http/logical.go
hghaf099: So, responseWriter is of type http.ResponseWriter. if I assign that directly, then, where ever we would need extra functionality like `ResponseWriter.Written()`, we would need to assert the `WrappingResponseWriter` interface first. Although there are not many places right now for such functionality, it is easy to miss fixing those. It is not just the `Written()` function as well.

sdk/logical/response.go
hghaf099: I found the issue in this PR while I was working at the logging requests feature. So, this is not needed now, but it is needed then.

http/logical.go
ncabatoff: Is it though?  I thought everything was wrapped by wrapGenericHandler, and it's always passing down a response writer of the new type.

hghaf099: hmm, I think so? The function signature of this function defines `w http.ResponseWriter` which is assigned to `responseWriter`. But, another thing is that of which type should the `req.ResponseWriter` be defined? should it be `http.ResponseWriter`, or `logical.StatusHeaderResponseWriter`?

sdk/logical/response.go
hghaf099: You are right, this is not needed. Fixing it in both PRs. Thanks!

http/logical.go
ncabatoff: I wasn't speaking in terms of function signature, I'm saying that since wrapGenericHandler is wrapping all request handlers, we shouldn't need to wrap the response handler again.  And if we don't need to handle re-wrapping I think some of the code will be simplified.

I think we should probably stick with http.ResponseWriter, but that's mostly because I expect trying to switch it would cause problems.  If it's possible to use StatusHeaderResponseWriter without it being really disruptive, that could make sense.

hghaf099: Sounds good! I will stick with http.ResponseWriter.

sdk/logical/request.go
```diff
@@ -203,7 +203,7 @@ type Request struct {
 
 	// ResponseWriter if set can be used to stream a response value to the http
 	// request that generated this logical.Request object.
-	ResponseWriter *HTTPResponseWriter `json:"-" sentinel:""`
+	ResponseWriter *http.ResponseWriter `json:"-" sentinel:""`
```
ncabatoff, line 206: I'd rather use the explicit type here than the interface, since there aren't any cases where we'll be using anything other than that type.

vault/custom_response_headers_test.go
```diff
@@ -127,8 +128,9 @@ func TestCustomResponseHeadersConfigInteractUiConfig(t *testing.T) {
 		t.Fatalf("custom header config should be configured in core")
 	}
 
-	w := httptest.NewRecorder()
-	hw := logical.NewHTTPResponseWriter(w)
+	hw := func(w http.ResponseWriter) *http.ResponseWriter {
```
ncabatoff, line 131: This is a rather complicated way to avoid defining a variable.

http/handler.go
ncabatoff: Sorry, I'm not following.  What would break if we got rid of SetHeaders and instead assigned listenerCustomHeaders.StatusCodeHeaderMap to a local var here?  Then passed that var as a 2nd arg to logical.NewStatusHeaderResponseWriter?

My primary goal here is to minimize mutable state in the type.

sdk/logical/response.go
```diff
@@ -192,30 +193,106 @@ func RespondWithStatusCode(resp *Response, req *Request, code int) (*Response, e
...
-		written:        new(uint32),
+type StatusHeaderResponseWriter struct {
+	wrapped     http.ResponseWriter
+	wroteHeader bool
+	statusCode  int
+	headers     map[string][]*CustomHeader
+	written     *uint32
+}
+
+func NewStatusHeaderResponseWriter (w http.ResponseWriter) *StatusHeaderResponseWriter {
```
ncabatoff, line 210:
```suggestion
func NewStatusHeaderResponseWriter(w http.ResponseWriter) *StatusHeaderResponseWriter {
```

ncabatoff (commented):
Looking very close!

ncabatoff (approved):

PR #13201, opened 2021-11-18
docs: move deprecation notices down the navbar

jweissig:
Moved from one of the first items in the navbar down to one of the last. They are not high priority information and should be grouped with upgrade and release notes.


taoism4504 (approved):
Looks good, and new location makes sense.

PR #13212, opened 2021-11-18
Backport 1.9.x: Raft Peer Removal Bug

zofskeez:
Backport for #13098


ui/mirage/factories/configuration.js
```diff
@@ -0,0 +1,27 @@
+import { Factory, trait } from 'ember-cli-mirage';
+import faker from 'faker';
```
hashishaw, line 2: I don't see faker in package.json for 1.9 🤔 sort of an interesting scenario, because we typically don't touch dependencies in patch versions. Let's talk about this next week

hashishaw (commented):
One thing, which applies to the rest of the backports too. Let's hold off on merging this for now

ui/mirage/factories/configuration.js
zofskeez: Ok sounds good. I just noticed that only the tests for 1.7.x are failing when trying to import faker but not on the other branches which is strange.

zofskeez: Cherry picking the faker removal from #13778

hashishaw (approved):
🙏

PR #13038, opened 2021-11-03
PGP key list input fix 

zofskeez:
Originally reported https://github.com/hashicorp/vault/issues/12056

This fix addresses an issue where the number of PGP Key inputs was not matching the key shares number in the initialization form on change.

image

PR #13213, opened 2021-11-18
Backport 1.8.x: Raft Peer Removal Bug

zofskeez:
Backport for #13038


hashishaw (approved):

hashishaw (changes requested):
Just realized faker wasn't in package.json until a change marked for 1.10, so the remove faker change should be added here too

hashishaw (approved):
🙌

PR #13215, opened 2021-11-18
Add HTTP PATCH support for KV key metadata

ccapurso:
PR [#57](https://github.com/hashicorp/vault-plugin-secrets-kv/pull/57) for `vault-plugin-secrets-kv` introduces a `PatchOperation` handler for the KV `<mount>/metadata/:path` endpoint. This PR introduces the CLI command `kv metadata patch`. The command will only include fields in the request data that have been provided as flags to prevent unexpectedly clearing them via the patch operation. This was actually an issue that was discovered for the `kv metadata put` command and has been fixed as part of this PR.


website/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -561,6 +561,60 @@ $ curl \
...
+
+| Method  | Path                     |
+| :------ | :----------------------- |
+| `PATCH` | `/secret/metadata/:path` |
+
+### Parameters
+
+- `max_versions` `(int: 0)` – The number of versions to keep per key. If not
+  set, the backend’s configured max version is used. Once a key has more than
+  the configured allowed versions the oldest version will be permanently
```
taoism4504, line 579:
```suggestion
  the configured allowed versions, the oldest version will be permanently
```

website/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -561,6 +561,60 @@ $ curl \
...
+  deleted.
+
+- `cas_required` `(bool: false)` – If true the key will require the cas
```
taoism4504, line 582:
```suggestion
- `cas_required` `(bool: false)` – If true, the key will require the cas
```

taoism4504 (commented):
just made some minor nits to the text.

go.mod
```diff
@@ -108,15 +108,15 @@ require (
 	github.com/hashicorp/vault-plugin-secrets-azure v0.11.2
 	github.com/hashicorp/vault-plugin-secrets-gcp v0.11.0
 	github.com/hashicorp/vault-plugin-secrets-gcpkms v0.10.0
-	github.com/hashicorp/vault-plugin-secrets-kv v0.5.7-0.20211123171606-16933c88368a
+	github.com/hashicorp/vault-plugin-secrets-kv v0.5.7-0.20211206210934-74fe79f60b74
```
HridoyRoy, line 111: Were these upgrades intentional? If you wanted to include the upgrades in the other 2 PRs open for review, I think you'd need to re-import after those PRs are merged.

command/kv_metadata_patch.go
```diff
@@ -0,0 +1,193 @@
...
+func (c *KVMetadataPatchCommand) Flags() *FlagSets {
+	set := c.flagSet(FlagSetHTTP | FlagSetOutputFormat)
+
+	// Common Options
+	f := set.NewFlagSet("Common Options")
+
+	f.IntVar(&IntVar{
+		Name:    "max-versions",
+		Target:  &c.flagMaxVersions,
+		Default: -1,
```
HridoyRoy, line 75: Why are the defaults here -1 instead of 0?

command/kv_metadata_put.go
```diff
@@ -71,14 +71,13 @@ func (c *KVMetadataPutCommand) Flags() *FlagSets {
 	f.IntVar(&IntVar{
 		Name:    "max-versions",
 		Target:  &c.flagMaxVersions,
-		Default: 0,
+		Default: -1,
 		Usage:   `The number of versions to keep. If not set, the backend’s configured max version is used.`,
 	})
 
-	f.BoolVar(&BoolVar{
+	f.BoolPtrVar(&BoolPtrVar{
```
HridoyRoy, line 78: I don't quite understand why `BoolPtrVar` is being used here over `BoolVar`, but I also don't really understand the purpose of `BoolVar` versus just using the bool. Could you clarify what the purpose of these are?

command/kv_metadata_put.go
```diff
@@ -71,14 +71,13 @@ func (c *KVMetadataPutCommand) Flags() *FlagSets {
 	f.IntVar(&IntVar{
 		Name:    "max-versions",
 		Target:  &c.flagMaxVersions,
-		Default: 0,
+		Default: -1,
 		Usage:   `The number of versions to keep. If not set, the backend’s configured max version is used.`,
 	})
 
-	f.BoolVar(&BoolVar{
```
HridoyRoy, line 78: It seems like `BoolVar` is no longer being used anywhere. Could we clean up those struct definitions?

website/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -561,6 +561,60 @@ $ curl \
...
+using a `Content-Type` header value of `application/merge-patch+json`. It does
+not create a new version.
+
+| Method  | Path                     |
+| :------ | :----------------------- |
+| `PATCH` | `/secret/metadata/:path` |
+
+### Parameters
+
+- `max_versions` `(int: 0)` – The number of versions to keep per key. If not
```
HridoyRoy, line 577: Currently I think the defaults are set to be `-1`. If we do decide to keep them at `-1`, we should update them here as well.

HridoyRoy (changes requested):
A few small questions/comments, but looks great overall!

command/kv_metadata_patch.go
ccapurso: We do not want to overwrite the field if the user did not provide the flag. Providing 0 is a valid value for the `max-versions` flag. There is a check further on in the command handler that will check for >= 0 to determine whether or not to provide the field in the request body.

command/kv_metadata_put.go
ccapurso: Using `BoolPtrVar` allows us to determine whether or not the flag was actually specified.

website/content/api-docs/secret/kv/kv-v2.mdx
ccapurso: The default of -1 above is mainly a way to determine whether or not to include the user-provided value. The secret engine itself will [default to 0](https://github.com/hashicorp/vault-plugin-secrets-kv/blob/vault-4290-patch-metadata/path_metadata.go#L252) when the field is not provided given that `GetOk` will result in the zero-value.

command/kv_metadata_put.go
ccapurso: I don't see any reference to `BoolVar` in [command/kv_metadata_put.go](https://github.com/hashicorp/vault/blob/c31ccf7fcf55fd81c48e497452d30f808f9a3038/command/kv_metadata_put.go).

go.mod
ccapurso: Yes, the intention is to include the upgrade for `vault-plugin-secrets-kv` for review. As you mentioned, I plan to re-import after those PRs are merged.

command/kv_metadata_put.go
HridoyRoy: Right, but the `BoolVar` struct definition still exists. I think it can just be removed entirely, right?

website/content/api-docs/secret/kv/kv-v2.mdx
HridoyRoy: Ahh that makes sense!  Thanks!

HridoyRoy (approved):
Looks good to me, though as a minor nit I'd still prefer for us to clean up the `BoolVar` struct if we can, since it doesn't seem to be used. Overall it looks good to go!

command/kv_metadata_put.go
HridoyRoy: Nevermind, I think I was hunting for references in the wrong repo!

PR #13227, opened 2021-11-19
[Homepage] connect homepage to dato content

alexcarpenter:


website/components/io-home-card/product-logos.ts
```diff
@@ -1,10 +1,34 @@
...
+  },
+  vault: {
+    color: require('@hashicorp/mktg-logos/product/vault/logomark/color.svg?include'),
+    neutral: require('@hashicorp/mktg-logos/product/vault/logomark/white.svg?include'),
+  },
+  waypoint: {
+    color: require('@hashicorp/mktg-logos/product/waypoint/logomark/color.svg?include'),
+    neutral: require('@hashicorp/mktg-logos/product/waypoint/logomark/white.svg?include'),
+  },
 }
```
BrandonRomano, line 34: Not a big downside if it's the case, but I believe this causes every logo to be loaded in-browser on every page this is imported?

Just wanted to tap on that; don't change it if that's the case, cause the files will be so small.

website/components/io-home-case-studies/index.tsx
```diff
@@ -46,7 +51,11 @@ export default function IoHomeCaseStudies({
               <a className={s.link} href={item.link}>
                 <span className={s.linkInner}>
                   <h3 className={s.linkHeading}>{item.heading}</h3>
-                  <IconArrowRight16 />
+                  {isInternalLink(item.link) ? (
+                    <IconArrowRight16 />
+                  ) : (
+                    <IconExternalLink16 />
```
BrandonRomano, line 57: 🛫

BrandonRomano: Glad to see this getting some usage.

website/lib/utils.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (
+    link.startsWith('/') ||
+    link.startsWith('#') ||
+    link.startsWith('https://vaultproject.io') ||
+    link.startsWith('https://www.vaultproject.io')
+  ) {
+    return true
+  }
+  return false
+}
```
BrandonRomano, line 11: Love this, reduction in author decisions FTW 👨‍🍳 😗

BrandonRomano (approved):

website/components/io-home-card/product-logos.ts
alexcarpenter: Acknowledged, idea is this gets replaced by Flight icons once products are included https://github.com/hashicorp/flight/issues/330

Issue #13225, opened 2021-11-19
Panic when token is null for /v1/auth/token/lookup, /v1/auth/token/renew, /v1/auth/token/revoke, /v1/auth/token/revoke-orphan


ludovicianul:
**Describe the bug**
Making a `POST` at any of the 4 URLs with `token = null` results in no HTTP response and the following stacktrace:

```shell
 invalid syntax"
2021-11-19T14:25:57.498+0200 [INFO]  http: panic serving 127.0.0.1:52114: interface conversion: interface {} is nil, not string
goroutine 227193 [running]:
net/http.(*conn).serve.func1()
        /Users/runner/hostedtoolcache/go/1.17.2/x64/src/net/http/server.go:1801 +0xb9
panic({0x50a5d80, 0xc002395590})
        /Users/runner/hostedtoolcache/go/1.17.2/x64/src/runtime/panic.go:1047 +0x266
github.com/hashicorp/vault/vault.(*Core).handleCancelableRequest(0xc00084cc00, {0x6a43270, 0xc002395110}, 0xc0006ae180)
        /Users/runner/work/vault/vault/vault/request_handling.go:546 +0x1d94
github.com/hashicorp/vault/vault.(*Core).switchedLockHandleRequest(0xc00084cc00, {0x6a43270, 0xc002394e70}, 0xc0006ae180, 0x0)
        /Users/runner/work/vault/vault/vault/request_handling.go:442 +0x4a5
github.com/hashicorp/vault/vault.(*Core).HandleRequest(...)
        /Users/runner/work/vault/vault/vault/request_handling.go:408
github.com/hashicorp/vault/http.request(0x54f6a00, {0x69f4e70, 0xc0014a4b80}, 0xc00134a500, 0xc0006ae180)
        /Users/runner/work/vault/vault/http/handler.go:953 +0x86
github.com/hashicorp/vault/http.handleLogicalInternal.func1({0x69f4e70, 0xc0014a4b80}, 0xc00134a500)
        /Users/runner/work/vault/vault/http/logical.go:341 +0xb6
net/http.HandlerFunc.ServeHTTP(0xc001a5bd15, {0x69f4e70, 0xc0014a4b80}, 0xc001172000)
        /Users/runner/hostedtoolcache/go/1.17.2/x64/src/net/http/server.go:2046 +0x2f
github.com/hashicorp/vault/http.handleRequestForwarding.func1({0x69f4e70, 0xc0014a4b80}, 0xc00134a500)
[...]

PR #13233, opened 2021-11-20
Fixed null token panic from 'v1/auth/token/' endpoints and returned p…

ryowright:
Fixes null token panic when hitting **/v1/auth/token/lookup, /v1/auth/token/renew, /v1/auth/token/revoke, /v1/auth/token/revoke-orphan** endpoints and returns proper error response [(Issue #13225)](https://github.com/hashicorp/vault/issues/13225).


hsimon-hashicorp:
And thanks for this one too! Similar story here, I'll get a review going soon.

vault/request_handling.go
```diff
@@ -543,6 +543,9 @@ func (c *Core) handleCancelableRequest(ctx context.Context, req *logical.Request
 				}
 				break
 			}
+			if token == nil {
+				return logical.ErrorResponse("bad token"), logical.ErrPermissionDenied
```
pmmukh, line 547: maybe `invalid token` might be better? cause browsing through current error responses, invalid seems to be the word mostly used for erroneous input.

pmmukh (approved):
one nit, lgtm! And thanks for submitting this PR!

vault/request_handling.go
ryowright: Hi, I just committed a change to the error message, changing it to invalid token. I accidentally wrote this as 'null token' in the commit message, but in the code it is 'invalid token' as requested.

pmmukh:
Thanks for submitting that change, looks good! Could you please rebase this PR off main once ? There seems to be some conflicts, and a couple failing tests that I believe should be fixed on main now.

pmmukh:
@ryowright I see you commited some changes in the vault-data folder, which I assume is a file backend you were using for testing, could you please revert those changes ?

ryowright:
I apologize, I missed removing the vault-data folder, which I usually do. Will revert it right now.

pmmukh:
Thanks for removing that folder, the pr looks good to me now! Merging it in, and thanks for the submission!

Issue #13232, opened 2021-11-20
Panic when token_type is null for /auth/token/roles/{role_name}


en-milie:
**Describe the bug**
Making a POST at  `/auth/token/roles/{role_name}` with token_type = null results in no HTTP response and the following stacktrace:

```shell
2021-11-20T22:41:52.487+0200 [INFO]  http: panic serving 127.0.0.1:51337: interface conversion: interface {} is nil, not string
goroutine 58 [running]:
net/http.(*conn).serve.func1()
        /Users/runner/hostedtoolcache/go/1.17.2/x64/src/net/http/server.go:1801 +0xb9
panic({0x50a5d80, 0xc0016008d0})
        /Users/runner/hostedtoolcache/go/1.17.2/x64/src/runtime/panic.go:1047 +0x266
github.com/hashicorp/vault/vault.(*TokenStore).tokenStoreRoleCreateUpdate(0xc000ad8c40, {0x6a43270, 0xc0015f1c20}, 0xc00104cd80, 0xc0015fa700)
        /Users/runner/work/vault/vault/vault/token_store.go:3351 +0x1fa5
github.com/hashicorp/vault/sdk/framework.(*Backend).HandleRequest(0xc000596d00, {0x6a43270, 0xc0015f1c20}, 0xc00104cd80)
        /Users/runner/work/vault/vault/sdk/framework/backend.go:278 +0x7ed
github.com/hashicorp/vault/vault.(*Router).routeCommon(0xc00098c2d0, {0x6a43270, 0xc0015f1c20}, 0xc00104cd80, 0x0)
        /Users/runner/work/vault/vault/vault/router.go:708 +0x15ec
github.com/hashicorp/vault/vault.(*Router).Route(...)
        /Users/runner/work/vault/vault/vault/router.go:505
github.com/hashicorp/vault/vault.(*Core).doRouting(0xc00081db00, {0x6a43270, 0xc0015f1c20}, 0xc000497570)
        /Users/runner/work/vault/vault/vault/request_handling.go:741 +0x2c
github.com/hashicorp/vault/vault.(*Core).handleRequest(0xc000ade600, {0x6a43270, 0xc0015f1c20}, 0xc00104cd80)
        /Users/runner/work/vault/vault/vault/request_handling.go:917 +0x1192
github.com/hashicorp/vault/vault.(*Core).handleCancelableRequest(0xc000ade600, {0x6a43270, 0xc0015f1860}, 0xc00104cd80)
[...]

PR #13236, opened 2021-11-22
Fix null token type bug

ryowright:
Similar fix to [Issue #13225](https://github.com/hashicorp/vault/issues/13225). Fixes panic resulting from `token_type` field being set to `null` in a request body to **/v1/auth/token/roles/{role_name}** to return a proper error response instead. This a fix for [Issue #13232 ](https://github.com/hashicorp/vault/issues/13232).


hsimon-hashicorp:
Thanks for this submission! I'll get a review going as soon as I can. :)

akshya96 (approved):

hsimon-hashicorp:
Hi @ryowright - can you add a changelog entry please? Thanks! :)

ryowright:
Hi @hsimon-hashicorp, just added a changelog entry.

Issue #13219, opened 2021-11-18
Web-UI errors with EC certs on PKI backend


tristanmorgan:
**Describe the bug**
The new PKI UI fails with an error when trying to view certificates generated with an EC key

**To Reproduce**
1. log into the Vault UI
2. generate a certificate with EC keys configured
3. See error

<img width="849" alt="Issue ECC cert error" src="https://user-images.githubusercontent.com/6336384/142508599-a08d1acb-190e-46e3-9568-16addf95b8c4.png">

**Expected behavior**
if the JS library can't decode the certificate it should fail gracefully.

**Environment:**
* Vault Server Version: 1.9.0
* Vault CLI Version: 1.9.0 
* Server Operating System/Architecture: macOS

Vault server configuration file(s):

```hcl
seal "awskms" {
  region     = "ap-southeast-2"
  kms_key_id = "alias/vault"
  endpoint   = "http://10.10.10.115:8080"
}

storage "raft" {
  path    = "raft/"
  node_id = "vault_1"
}

service_registration "consul" {
  scheme       = "http"
  service_tags = "urlprefix-vault.service.consul/"
}

listener "tcp" {
  address     = "0.0.0.0:8200"
  tls_disable = true

  telemetry {
    unauthenticated_metrics_access = true
  }
}

telemetry {
  prometheus_retention_time = "30s"
  disable_hostname          = true
}

default_lease_ttl    = "1h"
max_lease_ttl        = "12h"
ui                   = true
raw_storage_endpoint = true
api_addr             = "http://{{GetPrivateIP}}:8200"
cluster_addr         = "https://{{GetPrivateIP}}:8201"
```

**Additional context**
Seems this stems from an upstream issue [ECC Support #116](https://github.com/digitalbazaar/forge/issues/116)

hsimon-hashicorp:
Hi @tristanmorgan! Can you show the error you're seeing? The screenshot looks sort of empty. (Or, is that the error?) Let me know, and we'll get our engineering teams on it. Thank you! :)

tristanmorgan:
Hi @hsimon-hashicorp, Yes, the error is the missing certificate details.

Related to this error is what happens when looking at an existing certificate, this error is shown:
<img width="849" alt="existing ecc certificat error" src="https://user-images.githubusercontent.com/6336384/142514533-6add46e1-bd18-4264-bdb4-d75af86a7567.png">

hsimon-hashicorp:
[...]

PR #13238, opened 2021-11-22
UI/Fix node-forge EC error

hellobontempo:
Fix for #13219

Adds a catch for when `node-forge` returns an error converting or parsing a certificate, and returns unparsed certificate (so metadata will not display).

Addresses error returned when trying to read EC certs.  `Error: Cannot read public key. OID is not RSA.`

When unable to parse certificate metadata, UI reverts to original behavior and just displays serial number and certificate: 

Screen Shot 2021-11-22 at 4 19 04 PM

### DIFF
--- a/changelog/13238.txt
+++ b/changelog/13238.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixes node-forge error when parsing EC (elliptical curve) certs
+```
--- a/ui/app/helpers/parse-pki-cert.js
+++ b/ui/app/helpers/parse-pki-cert.js
@@ -8,7 +8,7 @@ export function parsePkiCert([model]) {
   }
   let cert;
   // node-forge cannot parse EC (elliptical curve) certs
-  // so just return model (original response) if pki parse returns an error
+  // return original response if unable to convert a Forge cert from PEM
   try {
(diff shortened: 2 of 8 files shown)


cipherboy:
Just to mention, we tested this via [setting up](https://www.vaultproject.io/docs/secrets/pki#setup) two secrets engines:

```
$ source devvault

$ vault secrets enable pki
$ vault secrets tune -max-lease-ttl=8760h pki 
$ vault write pki/root/generate/internal common_name=my-website.com ttl=8760h key_type=ec key_bits=256

$ vault secrets enable  -path=pki2/ pki
$ vault secrets tune -max-lease-ttl=8760h pki2 
$ vault write pki2/root/generate/internal common_name=my-website.com ttl=8760h key_type=rsa
```

One CA was with RSA and the other was with EC. Since the root CA certificate for the mount is visible from the secrets page, we were able to confirm the original issue and that this fixed it.

cipherboy (approved):

hashishaw (approved):

Monkeychip (approved):
Looks great. Don't forget the bug and ui label and the milestone. (unsure if this is being backported to 1.9, guessing it is?)

PR #13239, opened 2021-11-22
docs: added minor recommendation for k8s agent annotations 

nicolaka:
I updated the k8s agent annotation docs to include a recommendation on using the `vault.hashicorp.com/agent-pre-populate-only` .


tvoran (approved):
Thanks!

PR #13241, opened 2021-11-22
Respect WithWrappingToken for all secret ID's in approle auth

averche:
### Background

When authenticating with vault via AppRole, the secret ID can be specified in one of 3 ways:

- `FromFile`
- `FromEnv`
- `FromString`

If the secret ID happens to be a wrapping token, the option `auth.WithWrappingToken()` will let the client know to treat it as such. However, the current implementation will only respect `auth.WithWrappingToken()` if the secret ID is specified from a file (`FromFile`) and silently fails otherwise, which could be surprising to callers. 

The rationale behind this was that the wrapping token should be provided through a [trusted orchestrator](https://learn.hashicorp.com/tutorials/vault/secure-introduction#trusted-orchestrator) and is typically written to a file. However, it is conceivable that the trusted orchestrator would write it to an environment variable instead.

### Solution

This PR will respect the `auth.WithWrappingToken()` for all secret ID's (whether specified via `FromFile`, `FromEnv` or `FromString`)

### DIFF
--- a/api/auth/approle/approle.go
+++ b/api/auth/approle/approle.go
@@ -105,31 +105,34 @@ func (a *AppRoleAuth) Login(ctx context.Context, client *api.Client) (*api.Secre
 		"role_id": a.roleID,
 	}
 
-	if a.secretIDFile != "" {
-		secretIDValue, err := a.readSecretIDFromFile()
+	var secretIDValue string
+
+	switch {
+	case a.secretIDFile != "":
+		s, err := a.readSecretIDFromFile()
 		if err != nil {
 			return nil, fmt.Errorf("error reading secret ID: %w", err)
 		}
+		secretIDValue = s
+	case a.secretIDEnv != "":
+		s := os.Getenv(a.secretIDEnv)
+		if s == "" {
+			return nil, fmt.Errorf("secret ID was specified with an environment variable %q with an empty value", a.secretIDEnv)
+		}
+		secretIDValue = s
+	default:
+		secretIDValue = a.secretID
+	}
 
-		// if it was indicated that the value in the file was actually a wrapping
-		// token, unwrap it first
-		if a.unwrap {
-			unwrappedToken, err := client.Logical().Unwrap(secretIDValue)
-			if err != nil {
-				return nil, fmt.Errorf("unable to unwrap token: %w. If the AppRoleAuth struct was initialized with the WithWrappingToken LoginOption, then the secret ID's filepath should be a path to a response-wrapping token", err)
-			}
-			loginData["secret_id"] = unwrappedToken.Data["secret_id"]
-		} else {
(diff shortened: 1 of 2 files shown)


api/auth/approle/approle.go
```diff
@@ -22,13 +22,12 @@ type AppRoleAuth struct {
 var _ api.AuthMethod = (*AppRoleAuth)(nil)
 
 // SecretID is a struct that allows you to specify where your application is
-// storing the secret ID required for login to the AppRole auth method.
+// storing the secret ID required for login to the AppRole auth method. The
+// recommended secure pattern is to use response-wrapping tokens rather than
```
digivava, line 26: Good idea moving the recommendation here where it's more easily discoverable.

api/auth/approle/approle.go
```diff
@@ -105,31 +104,34 @@ func (a *AppRoleAuth) Login(ctx context.Context, client *api.Client) (*api.Secre
...
+		secretIDValue = s
+	default:
+		secretIDValue = a.secretID
+	}
+
+	// if the caller indicated that the value was actually a wrapping token, unwrap it first
+	if a.unwrap {
+		unwrappedToken, err := client.Logical().Unwrap(secretIDValue)
+		if err != nil {
+			return nil, fmt.Errorf("unable to unwrap response wrapping token: %w", err)
```
digivava, line 130: Do you think it's worth still including a little extra clarifying message in this like the original error had? Originally it was "If the AppRoleAuth struct was initialized with the WithWrappingToken LoginOption, then the secret ID's filepath should be a path to a response-wrapping token", but we could change it to something more generic.

If you think it's redundant we can remove it, I was just thinking a lot of people using this might be unfamiliar with what a "response-wrapping token" is. Perhaps they just glanced at the docs quickly, knew they needed it but didn't read carefully, and so tried to just "enable response-wrapping" by tacking on the WithWrappingToken option without understanding that they needed to get the wrapping token separately first.

I'm open to removing it though if it feels like it doesn't really help with that problem anyway.

digivava (approved):
Awesome, thank you! Just one comment with something optional, but looks good to go otherwise.

api/auth/approle/approle.go
averche: I figured the recommendation now applies to all 3 options so it makes sense to move it up.

api/auth/approle/approle.go
averche: My thinking here was that the error can only happen if the user explicitly specified the non-default `WithWrappingToken()` option, so the error just needs to inform/remind the caller that we are dealing with a wrapping token, not something else.

However if this is a very foreign concept to Vault, I'm open to adding more context here.

digivava: Sounds good. Ship it!

PR #13247, opened 2021-11-23
http: fix dropped test error

alrs:
Can someone label this `no-changelog`?

This fixes a dropped test error in the `http` package.


calvn (approved):
Thanks for catching this!

PR #13252, opened 2021-11-23
Don't claim that Vault obfuscates the environment variable for sensitive values

sgmiller:
This doesn't work at least in recent version of Go, as Go makes a
copy of the environment, so we're only modifying that one, and not the one
visible to the rest of the system.


kalafut:
5000 files changed… I’m assuming bc you need to merge to main, not master.

sgmiller:
LOL, yes.  My CLI github tool is old and doesn't properly figure out the main branch unless I tell it to.

stevendpclark (approved):
👍

PR #13255, opened 2021-11-23
CLI changes for new mount tune config parameter allowed_managed_keys

divyapola5:
This tunable grants access to a managed key in the current or any parent namespaces, if "any_mount" was not set on the managed key. (VAULT-3880)

### DIFF
--- a/command/auth_enable.go
+++ b/command/auth_enable.go
@@ -37,6 +37,7 @@ type AuthEnableCommand struct {
 	flagExternalEntropyAccess     bool
 	flagTokenType                 string
 	flagVersion                   int
+	flagAllowedManagedKeys        []string
 }
 
 func (c *AuthEnableCommand) Synopsis() string {
@@ -199,6 +200,13 @@ func (c *AuthEnableCommand) Flags() *FlagSets {
 		Usage:   "Select the version of the auth method to run. Not supported by all auth methods.",
 	})
 
+	f.StringSliceVar(&StringSliceVar{
+		Name:   flagNameAllowedManagedKeys,
+		Target: &c.flagAllowedManagedKeys,
+		Usage: "Comma-separated string or list of keys that will not be HMAC'd by audit " +
+			"devices in the request data object.",
+	})
+
 	return set
 }
 
@@ -297,6 +305,10 @@ func (c *AuthEnableCommand) Run(args []string) int {
 
 		if fl.Name == flagNameTokenType {
 			authOpts.Config.TokenType = c.flagTokenType
+		}
+
+		if fl.Name == flagNameAllowedManagedKeys {
+			authOpts.Config.AllowedManagedKeys = c.flagAllowedManagedKeys
 		}
 	})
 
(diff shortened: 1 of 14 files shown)


command/auth_enable.go
```diff
@@ -199,6 +200,13 @@ func (c *AuthEnableCommand) Flags() *FlagSets {
 		Usage:   "Select the version of the auth method to run. Not supported by all auth methods.",
 	})
 
+	f.StringSliceVar(&StringSliceVar{
+		Name:   flagNameAllowedManagedKeys,
+		Target: &c.flagAllowedManagedKeys,
+		Usage: "Comma-separated string or list of managed key registry entry names" +
```
stevendpclark, line 206: I believe this documentation and the test below are conflicting?

The test below is taking the comma separated value as a single list element, while my reading of this indicates there should have been two elements within the list.

stevendpclark (changes requested):
Overall looks good @divyapola5!

Beyond the comment about the comma separated value parsing below, I discussed with @sgmiller and for now could we limit the new argument to just the secrets commands, removing it from the auth related ones? We don't have a use case at the moment for using the managed keys within auth mounts, we can add them later if something pops up.

command/auth_enable.go
divyapola5: Updated the description for allowed-managed-keys. Will make changes for other flags in a different PR

stevendpclark (approved):
👍 LGTM

command/secrets_enable.go
```diff
@@ -209,6 +210,14 @@ func (c *SecretsEnableCommand) Flags() *FlagSets {
 		Usage:   "Select the version of the engine to run. Not supported by all engines.",
 	})
 
+	f.StringSliceVar(&StringSliceVar{
+		Name:   flagNameAllowedManagedKeys,
+		Target: &c.flagAllowedManagedKeys,
+		Usage: "Managed key registry entry name that the mount in question is allowed" +
```
sgmiller, line 216: The registry isn't something users need to be aware of.  May be change this to "Managed key name(s) that the..."

sgmiller: Also, can string slices just be comma separated with a single flag?  Can't remember, but if so that might be better guidance in the next part.

command/secrets_tune.go
```diff
@@ -137,6 +138,14 @@ func (c *SecretsTuneCommand) Flags() *FlagSets {
 		Usage:   "Select the version of the engine to run. Not supported by all engines.",
 	})
 
+	f.StringSliceVar(&StringSliceVar{
+		Name:   flagNameAllowedManagedKeys,
+		Target: &c.flagAllowedManagedKeys,
+		Usage: "Managed key registry entry name that the mount in question is allowed" +
+			"to access. Note that multiple keys may be specified by providing this option" +
+			"multiple times, each time with 1 key.",
```
sgmiller, line 146: Ditto from above.

sgmiller (changes requested):
Looks good with the small exception of some CLI help languages changes.

taoism4504 (commented):
Reviewed secrets and secrets tune text. LG.

command/secrets_enable.go
divyapola5: Done. Corrected the description. Thanks.

sgmiller (approved):

PR #13259, opened 2021-11-23
Fix case typo in docs

mike-code:
The k8s helm chart docs specifies `statefulset` as one of the keys for chart values when[ in fact it's `statefulSet`](https://github.com/hashicorp/vault-helm/blob/main/values.yaml#L690). Using lowercase won't work.


hsimon-hashicorp (approved):

hsimon-hashicorp:
Thanks for this catch!

PR #13260, opened 2021-11-23
Backport 1.9.x: Fixing ResponseWriter to assert to http.Flusher in /sys/monitor endpoint (#13200)

hghaf099:
There are a number of interfaces that implement `http.ResponseWriter`. For that, wrapping `http.ResponseWriter` needs a lot of care.
`StatusHeaderResponseWriter` is instantiated in the http layer which wraps http.ResponseWriter. Then, `HTTPResponseWriter` wraps the already wrapped ResponseWriter. This causes issues as for example in the `sys/monitor`, the `http.Flusher` cannot be asserted and calling the endpoint returns stream not supported


changelog/13200.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+http:Unify HTTPResponseWriter and StatusHeaderResponseWriter
```
hghaf099, line 2:
```suggestion
http:Fixing ResponseWriter to assert to http.Flusher in /sys/monitor endpoint
```

changelog/13200.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+http:Fixing ResponseWriter to assert to http.Flusher in /sys/monitor endpoint
```
ncabatoff, line 2: Think about it from the user's perspective who's reading the CL.   Most of the above is meaningless to them.  Can you rewrite in a way that a user could understand it, without reference to internals?

hghaf099: Sounds good! How about the following?
"Fixing `/sys/monitor` endpoint returning `streaming not supported`"

ncabatoff: Sounds good, though I suggest "Fix" rather than "Fixing", just to be consistent with how we usually phrase the CLs.

hghaf099, line 2:
```suggestion
http:Fix /sys/monitor endpoint returning streaming not supported
```

ncabatoff (approved):

PR #13266, opened 2021-11-23
Component test fixes

zofskeez:
A couple of component tests were failing from a failed request to `GET /v1/sys/replication/dr/secondary/generate-operation-token/attempt`. Since it is not needed by the current tests an interceptor was added.


hashishaw (approved):
Thanks for the quick turnaround!

PR #13272, opened 2021-11-24
Add ability to capture container logs, and have mssql test helper use…

ncabatoff:
… it when a test fails.

### DIFF
--- a/helper/testhelpers/docker/testhelpers.go
+++ b/helper/testhelpers/docker/testhelpers.go
@@ -40,6 +40,7 @@ type RunOptions struct {
 	DoNotAutoRemove bool
 	AuthUsername    string
 	AuthPassword    string
+	LogConsumer     func(string)
 }
 
 func NewServiceRunner(opts RunOptions) (*Runner, error) {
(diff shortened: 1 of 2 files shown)


helper/testhelpers/docker/testhelpers.go
```diff
@@ -135,6 +136,21 @@ func (d *Runner) StartService(ctx context.Context, connect ServiceAdapter) (*Ser
...
+				ShowStderr: true,
+				Timestamps: true,
+				Details:    true,
+			})
+			if err == nil {
+				b, err := ioutil.ReadAll(rc)
+				if err == nil {
+					d.RunOptions.LogConsumer(string(b))
+				}
+			}
```
stevendpclark, line 151: should we send the err string to the LogConsumer if err != nil?

stevendpclark (approved):
Small comment but 👍

helper/testhelpers/docker/testhelpers.go
ncabatoff: Good call.

stevendpclark: Outputting the err from DockerAPI.ContainerLogs might be useful as well, say the container id completely disappeared or we had issues with docker daemon itself, it might contain further clues?

ncabatoff: That'll have to wait for a future PR, I've reached my timebox limit for this effort.

PR #13275, opened 2021-11-24
Docs to clarify k8s auth options with short-lived tokens

tomhjp:
Since short-lived tokens became the default in 1.21, we've seen a lot of people asking questions around how to use them with k8s auth. This attempts to lay out the options we have today. Hopefully we'll have another option soon that looks something like https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/122. I would also like to add an example of how to use Vault injector with JWT auth, but will save that for another PR.

### DIFF
--- a/website/content/docs/auth/jwt/oidc_providers.mdx
+++ b/website/content/docs/auth/jwt/oidc_providers.mdx
@@ -262,3 +262,148 @@ vault write auth/oidc/role/your_default_role \
 1. Save client ID and secret.
 
 Note your policy will need `oidc_scopes` to include `profile` to get a full profile ("[Fat Token](https://support.okta.com/help/s/article/Okta-Groups-or-Attribute-Missing-from-Id-Token)"). You will also need to configure bound audience along the lines of `"bound_audiences": ["api://default", "0a4........."]` if you are using the default authorization server.
+
+## Kubernetes
+
+Kubernetes can function as an OIDC provider such that Vault can validate its
+service account tokens using JWT/OIDC auth.
+
+-> **Note:** The JWT auth engine does **not** use Kubernetes' `TokenReview` API
+during authentication, and instead uses public key cryptography to verify the
+contents of JWTs. This means tokens that have been revoked by Kubernetes will
+still be considered valid by Vault until their expiry time. To mitigate this
+risk, use short TTLs for service account tokens or use
+[Kubernetes auth][/docs/auth/kubernetes] which _does_ use the `TokenReview` API.
+
+### Using service account issuer discovery
+
+Kubernetes cluster requirements:
+
+* `ServiceAccountIssuerDiscovery` feature gate enabled.
+   * Present from 1.18, defaults to true from 1.20.
+* Must use short-lived service account tokens when logging in.
+   * Tokens mounted into pods default to short-lived from 1.21.
+* kube-apiserver's `--service-account-issuer` flag is set to a URL that is
+  reachable from Vault. Public by default for most managed Kubernetes solutions.
+
+1. Ensure OIDC discovery URLs do not require authentication, as detailed
+   [here](k8s-sa-issuer-discovery):
+
+   ```bash
+   kubectl create clusterrolebinding oidc-reviewer  \
+      --clusterrole=system:service-account-issuer-discovery \
+      --group=system:unauthenticated
+   ```
+
+1. Enable and configure JWT auth in Vault.
+
+   1. If Vault is running in Kubernetes:
+
+      ```bash
+      kubectl exec vault-0 -- vault auth enable jwt
+      kubectl exec vault-0 -- vault write auth/jwt/config \
+         oidc_discovery_url=https://kubernetes.default.svc.cluster.local \
+         oidc_discovery_ca_pem=@/run/secrets/kubernetes.io/serviceaccount/ca.crt
+      ```
+
+   1. Alternatively, if Vault is _not_ running in Kubernetes:
+
+      -> **Note:** When Vault is outside the cluster, the `$ISSUER` below may
+         or may not be reachable. If not, you can configure JWT auth using
+         [`jwt_validation_pubkeys`](#using-jwt-validation-public-keys) instead.
+
+      ```bash
+      # 1. Find the issuer URL of the cluster.
+      kubectl proxy &
+      ISSUER="$(curl --silent 127.0.0.1:8001/.well-known/openid-configuration | jq -r '.issuer')"
+
+      # 2. Enable and configure the JWT auth mount.
+      vault auth enable jwt
+      vault write auth/jwt/config oidc_discovery_url="${ISSUER}"
+      ```
+
+1. Configure a role and log in as detailed below.
+
+[k8s-sa-issuer-discovery]: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery
+[k8s-token-audience]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#serviceaccounttokenprojection-v1-core
+
+### Using JWT validation public keys
+
+Kubernetes cluster requirements:
+
+* `ServiceAccountIssuerDiscovery` feature gate enabled.
+   * Present from 1.18, defaults to true from 1.20.
+   * This requirement can be avoided if you can access the Kubernetes master
+      nodes to read the public signing key directly from disk at
+      `/etc/kubernetes/pki/sa.pub`. In this case, you can also skip the step to
+      convert the key as it will already be in PEM format.
+* Must use short-lived service account tokens when logging in.
+   * Tokens mounted into pods default to short-lived from 1.21.
+
+This method can be useful if Kubernetes' API is not reachable from Vault or if
+you would like a single JWT auth mount to service multiple Kubernetes clusters
+by chaining their public signing keys.
+
+1. Fetch the service account signing public key from your cluster's JWKS URI.
+
+   ```bash
+   # 1. Find the issuer URL of the cluster.
+   kubectl proxy &
+   ISSUER="$(curl --silent 127.0.0.1:8001/.well-known/openid-configuration | jq -r '.issuer')"
+
+   # 2. Query the jwks_uri specified in /.well-known/openid-configuration
+   # NB: You may need to run this from a pod within the cluster if the $ISSUER
+   # URL is not available outside the cluster.
+   curl "$(curl --silent "${ISSUER}/.well-known/openid-configuration" | jq -r '.jwks_uri')"
+   ```
+
+1. Convert the keys from JWK format to PEM. You can use a CLI tool or an online
+   converter such as [this one][jwk-to-pem].
+
+1. Configure the JWT auth mount with those public keys.
+
+   ```bash
+   vault write auth/jwt/config \
+      jwt_validation_pubkeys="-----BEGIN PUBLIC KEY-----
+   MIIBIjANBgkqhkiG9...
+   -----END PUBLIC KEY-----","-----BEGIN PUBLIC KEY-----
+   MIIBIjANBgkqhkiG9...
+   -----END PUBLIC KEY-----"
+   ```
+
+1. Configure a role and log in as detailed below.
+
+[jwk-to-pem]: https://8gwifi.org/jwkconvertfunctions.jsp
+
+### Configuring a role and logging in
+
+Once your JWT auth mount is configured, you're ready to configure a role and
+log in.
+
+1. Create a role for JWT auth that the `default` service account from the
+   `default` namespace can use. The audience of tokens defaults to the same as
+   the issuer, but it is [configurable][k8s-token-audience].
+
+   ```bash
+   vault write auth/jwt/role/my-role \
+      role_type="jwt" \
+      bound_audiences="${ISSUER}" \
+      user_claim="sub" \
+      bound_subject="system:serviceaccount:default:default" \
+      policies="default" \
+      ttl="1h"
+   ```
+
+1. Pods or other clients with access to a service account JWT can then log in.
+
+   ```bash
+   vault write auth/jwt/login \
+      role=my-role \
+      jwt=@/var/run/secrets/kubernetes.io/serviceaccount/token
+   # OR equivalent to:
+   curl \
+      --request PUT \
+      --header "X-Vault-Request: true" \
+      --data '{"jwt":"<JWT-TOKEN-HERE>","role":"my-role"}' \
+      "${VAULT_ADDR}/v1/auth/jwt/login"
+   ```
--- a/website/content/docs/auth/jwt/oidc_providers.mdx
+++ b/website/content/docs/auth/jwt/oidc_providers.mdx
@@ -253,16 +253,6 @@ vault write auth/oidc/role/your_default_role \
 1. Save.
 1. Visit Credentials. Select Client ID and Secret and note the generated secret.
 
-## Okta
-
-1. Make sure an Authorization Server has been created. The "Issuer" field shown on the Setting page
-   will be used as the `oidc_discovery_url`.
-1. Visit Applications > Add Application (Web).
-1. Configure Login redirect URIs. Save.
-1. Save client ID and secret.
-
-Note your policy will need `oidc_scopes` to include `profile` to get a full profile ("[Fat Token](https://support.okta.com/help/s/article/Okta-Groups-or-Attribute-Missing-from-Id-Token)"). You will also need to configure bound audience along the lines of `"bound_audiences": ["api://default", "0a4........."]` if you are using the default authorization server.
-
 ## Kubernetes
 
 Kubernetes can function as an OIDC provider such that Vault can validate its
@@ -290,6 +280,8 @@ Kubernetes cluster requirements:
   reachable from Vault. Public by default for most managed Kubernetes solutions.
 * Must use short-lived service account tokens when logging in.
    * Tokens mounted into pods default to short-lived from 1.21.
+
+Configuration steps:
 
 1. Ensure OIDC discovery URLs do not require authentication, as detailed
    [here][k8s-sa-issuer-discovery]:
@@ -349,6 +341,8 @@ Kubernetes cluster requirements:
       retrieve and then convert the key as it will already be in PEM format.
 * Must use short-lived service account tokens when logging in.
    * Tokens mounted into pods default to short-lived from 1.21.
+
+Configuration steps:
 
 1. Fetch the service account signing public key from your cluster's JWKS URI.
 
@@ -461,3 +455,13 @@ spec:
 ```
 
 [k8s-extended-tokens]: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/#options
+
+## Okta
+
+1. Make sure an Authorization Server has been created. The "Issuer" field shown on the Setting page
+   will be used as the `oidc_discovery_url`.
+1. Visit Applications > Add Application (Web).
+1. Configure Login redirect URIs. Save.
+1. Save client ID and secret.
+
+Note your policy will need `oidc_scopes` to include `profile` to get a full profile ("[Fat Token](https://support.okta.com/help/s/article/Okta-Groups-or-Attribute-Missing-from-Id-Token)"). You will also need to configure bound audience along the lines of `"bound_audiences": ["api://default", "0a4........."]` if you are using the default authorization server.
--- a/website/content/docs/auth/jwt/oidc_providers.mdx
+++ b/website/content/docs/auth/jwt/oidc_providers.mdx
@@ -295,7 +295,7 @@ Configuration steps:
 1. Find the issuer URL of the cluster.
 
    ```bash
-   kubectl proxy &
+   timeout 30s kubectl proxy &
    ISSUER="$(curl --fail --silent --show-error 127.0.0.1:8001/.well-known/openid-configuration | jq -r '.issuer')"
    ```
 
@@ -348,7 +348,7 @@ Configuration steps:
 
    ```bash
    # 1. Find the issuer URL of the cluster.
-   kubectl proxy &
+   timeout 30s kubectl proxy &
    ISSUER="$(curl --fail --silent --show-error 127.0.0.1:8001/.well-known/openid-configuration | jq -r '.issuer')"
 
    # 2. Query the jwks_uri specified in /.well-known/openid-configuration
--- a/website/content/docs/auth/jwt/oidc_providers.mdx
+++ b/website/content/docs/auth/jwt/oidc_providers.mdx
@@ -308,7 +308,7 @@ Kubernetes cluster requirements:
 
    1. Alternatively, if Vault is _not_ running in Kubernetes:
 
-      -> **Note:** When Vault is outside the cluster, the `$ISSUER` below may
+      -> **Note:** When Vault is outside the cluster, the `$ISSUER` endpoint below may
          or may not be reachable. If not, you can configure JWT auth using
          [`jwt_validation_pubkeys`](#using-jwt-validation-public-keys) instead.
 
--- a/website/content/docs/auth/jwt/oidc_providers.mdx
+++ b/website/content/docs/auth/jwt/oidc_providers.mdx
@@ -432,6 +432,11 @@ kind: Pod
 metadata:
   name: nginx
 spec:
+  # automountServiceAccountToken is redundant in this example because the
+  # mountPath overlapping with the default path below will already stop the
+  # default admission injected token from being created. Use this option if you
+  # choose a different mount path.
+  automountServiceAccountToken: false
   containers:
     - name: nginx
       image: nginx
--- a/website/content/docs/auth/jwt/oidc_providers.mdx
+++ b/website/content/docs/auth/jwt/oidc_providers.mdx
@@ -273,26 +273,38 @@ during authentication, and instead uses public key cryptography to verify the
 contents of JWTs. This means tokens that have been revoked by Kubernetes will
 still be considered valid by Vault until their expiry time. To mitigate this
 risk, use short TTLs for service account tokens or use
-[Kubernetes auth][/docs/auth/kubernetes] which _does_ use the `TokenReview` API.
+[Kubernetes auth](/docs/auth/kubernetes) which _does_ use the `TokenReview` API.
 
 ### Using service account issuer discovery
 
+When using service account issuer discovery, you only need to provide the JWT
+auth mount with an OIDC discovery URL, and sometimes a TLS certificate authority
+to trust. This makes it the most straightforward method to configure if your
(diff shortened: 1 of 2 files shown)


website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,148 @@ vault write auth/oidc/role/your_default_role \
...
+
+   ```bash
+   kubectl create clusterrolebinding oidc-reviewer  \
+      --clusterrole=system:service-account-issuer-discovery \
+      --group=system:unauthenticated
+   ```
+
+1. Enable and configure JWT auth in Vault.
+
+   1. If Vault is running in Kubernetes:
```
RemcoBuddelmeijer, line 300: I am missing the `ISSUER` variable here that is defined in the alternative step and used in the role configuration step, can this be added or a link to the [discovering the service account issuer](https://www.vaultproject.io/docs/auth/kubernetes#discovering-the-service-account-issuer) block?

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,148 @@ vault write auth/oidc/role/your_default_role \
...
+      ```
+
+   1. Alternatively, if Vault is _not_ running in Kubernetes:
+
+      -> **Note:** When Vault is outside the cluster, the `$ISSUER` below may
+         or may not be reachable. If not, you can configure JWT auth using
+         [`jwt_validation_pubkeys`](#using-jwt-validation-public-keys) instead.
+
+      ```bash
+      # 1. Find the issuer URL of the cluster.
```
RemcoBuddelmeijer, line 316: This approach speculates that every cluster can obtain the information through this endpoint. This is different from the current "[discovering the service account issuer](https://www.vaultproject.io/docs/auth/kubernetes#discovering-the-service-account-issuer)" block. Can this be changed?

website/content/docs/auth/kubernetes.mdx
```diff
@@ -101,9 +106,72 @@ management tool.
...
+used as the `token_reviewer_jwt` parameter when configuring Kubernetes auth,
+because Vault stores that value in its own storage and does not automatically
+refresh it. There are a few different ways to configure Kubernetes auth in this
+scenario, each with their own tradeoffs.
+
+[k8s-1.21-changelog]: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md#api-change-2
+
+#### Use the Vault client's JWT as the reviewer JWT
+
+When configuring Kubernetes auth, you can omit the `token_reviewer_jwt`, and Vault
```
RemcoBuddelmeijer, line 124: If you just omit the `token_reviewer_jwt` and not use the `disable_local_ca_jwt` option, it will simply load in the token from the file system of the Vault server and use this one instead. Can this part be added?

RemcoBuddelmeijer (changes requested):

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,148 @@ vault write auth/oidc/role/your_default_role \
...
+      kubectl proxy &
+      ISSUER="$(curl --silent 127.0.0.1:8001/.well-known/openid-configuration | jq -r '.issuer')"
```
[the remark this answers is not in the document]
benashz: You could also go with passing the short option equivalents to curl. Personally I like using long options in docs.

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,148 @@ vault write auth/oidc/role/your_default_role \
...
+
+1. Pods or other clients with access to a service account JWT can then log in.
+
+   ```bash
+   vault write auth/jwt/login \
+      role=my-role \
+      jwt=@/var/run/secrets/kubernetes.io/serviceaccount/token
+   # OR equivalent to:
+   curl \
+      --request PUT \
```
benashz, line 405:
```suggestion
      --fail \
      --request PUT \
```

website/content/docs/auth/kubernetes.mdx
```diff
@@ -101,9 +106,72 @@ management tool.
...
+Option | All tokens are short-lived | Can revoke tokens early | Other considerations
+------------------------------------------------------------------------
+Use client JWT as reviewer JWT       | Yes | Yes | Operational overhead
+Use long-lived token as reviewer JWT | No  | Yes |
+Use JWT auth                         | Yes | No  |
+
 ### Discovering the service account `issuer`
 
--> **Deprecated:** The `issuer` parameter has been deprecated as of Vault 1.9 and will be removed in a future release.
+-> **Note:** From Vault 1.9, `disable_iss_validation` and `issuer` are deprecated
```
benashz, line 170: 👍

benashz (approved):
Looks great to me. Only a few minor suggestions. Approving it.

website/content/docs/auth/jwt/oidc_providers.mdx
RemcoBuddelmeijer: @benashz My apologies for not being clearer. With "the block" I was talking about the following text in the "discovering the service account issuer" section in the Kubernetes Auth docs:
```
Most clusters will also have that information available at the .well-known/openid-configuration endpoint:
```

Here it states that _most_ clusters will also have that information available rather than always having this available. 

I don't particularly see this difference as a downside, perhaps even a bit picky, which is also the reason why I didn't put this under "Request change". Rather this might be a nice to have as the documentation in that case is identical.

website/content/docs/auth/jwt/oidc_providers.mdx
RemcoBuddelmeijer: @benashz I did notice that the standard issuer is used here when writing to the configuration of the JWT auth.

However, if you would only follow the steps that you are supposed to follow for when Vault is running in the cluster: you end up not setting the "env" variable for `ISSUER`.
This variable is later on needed to supply a bound audience for the created role and if not setting this would mean that this audience will be empty thus creating a role that would be invalid and any login with the Kubernetes issuer would fail.

While this is up to the Vault administrator to have enough knowledge about the used audience, it would be nice to have this in both steps. As this could be mistaken very easily as this variable is set in the alternative step, but not in this step; which might come over as "done on purpose". 😃

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,148 @@ vault write auth/oidc/role/your_default_role \
...
+This method can be useful if Kubernetes' API is not reachable from Vault or if
+you would like a single JWT auth mount to service multiple Kubernetes clusters
+by chaining their public signing keys.
+
+1. Fetch the service account signing public key from your cluster's JWKS URI.
+
+   ```bash
+   # 1. Find the issuer URL of the cluster.
+   kubectl proxy &
+   ISSUER="$(curl --silent 127.0.0.1:8001/.well-known/openid-configuration | jq -r '.issuer')"
```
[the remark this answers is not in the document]
tomhjp: Nice! TIL

website/content/docs/auth/kubernetes.mdx
tomhjp: Good spot, it does indeed need setting for this scenario when Vault is running in Kubernetes. Fixed.

website/content/docs/auth/jwt/oidc_providers.mdx
tomhjp: If `ServiceAccountIssuerDiscovery` is enabled, then the cluster does have the openid-configuration endpoint available: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery.

Given we have that as a requirement at the beginning of the section, I think we're covered here. If users have problems using the issuer URL themselves then I think they're likely to run into problems later when Vault tries to use it too, so I'm not too inclined to link to an alternative method either. I've cleared up the language below to make clear you don't necessarily need the issuer URL for public keys.

website/content/docs/auth/jwt/oidc_providers.mdx
tomhjp: I've pulled getting the `ISSUER` out into a separate step to avoid confusion. Thanks for the fast feedback.

RemcoBuddelmeijer (approved):

website/content/docs/auth/kubernetes.mdx
RemcoBuddelmeijer: Thank you for fixing this! I have now changed my review to approved (for the sake of it). 😄

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,202 @@ vault write auth/oidc/role/your_default_role \
...
+   ```bash
+   kubectl create clusterrolebinding oidc-reviewer  \
+      --clusterrole=system:service-account-issuer-discovery \
+      --group=system:unauthenticated
+   ```
+
+1. Find the issuer URL of the cluster.
+
+   ```bash
+   kubectl proxy &
```
calvn, line 306: Is it okay to leave the process in the background like so?

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,202 @@ vault write auth/oidc/role/your_default_role \
...
+1. Ensure OIDC discovery URLs do not require authentication, as detailed
+   [here][k8s-sa-issuer-discovery]:
+
+   ```bash
+   kubectl create clusterrolebinding oidc-reviewer  \
+      --clusterrole=system:service-account-issuer-discovery \
+      --group=system:unauthenticated
+   ```
+
+1. Find the issuer URL of the cluster.
```
calvn, line 303: I looks like this step is only needed if Vault is running outside of k8s, is that right?

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,202 @@ vault write auth/oidc/role/your_default_role \
...
+      jwt_validation_pubkeys="-----BEGIN PUBLIC KEY-----
+   MIIBIjANBgkqhkiG9...
+   -----END PUBLIC KEY-----","-----BEGIN PUBLIC KEY-----
+   MIIBIjANBgkqhkiG9...
+   -----END PUBLIC KEY-----"
+   ```
+
+1. Configure a role and log in as detailed [below](#creating-a-role-and-logging-in).
+
+[jwk-to-pem]: https://8gwifi.org/jwkconvertfunctions.jsp
```
calvn, line 382: If the JWK includes bits for constructing the private key, I wonder if it's better to point users to a local CLI tool (https://www.npmjs.com/package/jwk-to-pem) rather than an online version that they'd need to upload the blob to.

website/content/docs/auth/jwt/oidc_providers.mdx
tomhjp: It's needed further down when creating the role too, for `bound_audiences`

website/content/docs/auth/jwt/oidc_providers.mdx
tomhjp: It only includes the public key - I searched a lot for a better solution here, and unfortunately came up short.

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,202 @@ vault write auth/oidc/role/your_default_role \
...
+to trust. This makes it the most straightforward method to configure if your
+Kubernetes cluster meets the requirements.
+
+Kubernetes cluster requirements:
+
+* [`ServiceAccountIssuerDiscovery`][k8s-sa-issuer-discovery] feature enabled.
+   * Present from 1.18, defaults to enabled from 1.20.
+* kube-apiserver's `--service-account-issuer` flag is set to a URL that is
+  reachable from Vault. Public by default for most managed Kubernetes solutions.
+* Must use short-lived service account tokens when logging in.
```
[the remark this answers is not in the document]
tomhjp: It's a little complicated - nominally 3607 seconds, but as mentioned elsewhere there's currently an extension introduced by Kubernetes to ease the transition to short-lived tokens, so the default pod tokens actually last for a year (unless the associated pod or service account is deleted first).

website/content/docs/auth/jwt/oidc_providers.mdx
tomhjp: It's not a great idea to leave it running forever after these instructions are done. I didn't want to complicate the instructions further by including extra steps for killing it, as the process to get here is already pretty complex.

website/content/docs/auth/kubernetes.mdx
```diff
@@ -101,9 +106,91 @@ management tool.
...
+This solution allows you to use short-lived tokens for all clients and removes
+the need for a reviewer JWT. However, the client tokens cannot be revoked before
+their TTL expires, so it is recommended to keep the TTL short with that
+limitation in mind.
+
 ### Discovering the service account `issuer`
 
--> **Deprecated:** The `issuer` parameter has been deprecated as of Vault 1.9 and will be removed in a future release.
+-> **Note:** From Vault 1.9, `disable_iss_validation` and `issuer` are deprecated
+and the default for `disable_iss_validation` has changed to `true`. The following
```
calvn, line 190: This only applies on new mounts/configs. Should we call out the need to update the config to set this to true if they update their kubernetes clusters to 1.21 and above?

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,202 @@ vault write auth/oidc/role/your_default_role \
...
+* [`ServiceAccountIssuerDiscovery`][k8s-sa-issuer-discovery] feature enabled.
+   * Present from 1.18, defaults to enabled from 1.20.
+   * This requirement can be avoided if you can access the Kubernetes master
+      nodes to read the public signing key directly from disk at
+      `/etc/kubernetes/pki/sa.pub`. In this case, you can skip the steps to
+      retrieve and then convert the key as it will already be in PEM format.
+* Must use short-lived service account tokens when logging in.
+   * Tokens mounted into pods default to short-lived from 1.21.
+
+1. Fetch the service account signing public key from your cluster's JWKS URI.
```
tvoran, line 353: Similar question about another break here, something like "Configuration steps:"

website/content/docs/auth/kubernetes.mdx
```diff
@@ -101,9 +106,91 @@ management tool.
...
+section only applies if you set `disable_iss_validation=false`. The Kubernetes API
+does the same validation when reviewing tokens, so enabling issuer validation is
+duplicated work.
```
tvoran, line 193:
```suggestion
does the same validation when reviewing tokens, so enabling issuer validation on
the Vault side is duplicated work.
```

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -262,3 +262,202 @@ vault write auth/oidc/role/your_default_role \
...
+### Using JWT validation public keys
+
+This method can be useful if Kubernetes' API is not reachable from Vault or if
+you would like a single JWT auth mount to service multiple Kubernetes clusters
+by chaining their public signing keys.
+
+Kubernetes cluster requirements:
+
+* [`ServiceAccountIssuerDiscovery`][k8s-sa-issuer-discovery] feature enabled.
+   * Present from 1.18, defaults to enabled from 1.20.
```
tvoran, line 345: Is it defaulting to enabled on 1.20? Or is it 1.21?

Just curious since the 1.20 docs seem to indicate feature flags still need to be enabled: https://v1-20.docs.kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery

website/content/docs/auth/kubernetes.mdx
```diff
@@ -101,9 +106,91 @@ management tool.
...
+find the secret, run:
+
+```bash
+kubectl get secret "$(kubectl get serviceaccount default -o jsonpath='{.secrets[0].name}')"
+```
+
+Using this maintains previous workflows but does not fully take advantage of the
+new default short-lived tokens.
+
+#### Use JWT auth
```
gavinelder, line 173: I am having issues with the relevance in this section given that this is for the JWT plugin vs the k8s auth as such it may be better to provider  an "Important note" section as an introductory for this this documentation and redirect users quickly to that method / reading versus potentially discovering this half way down the page.

website/content/docs/auth/kubernetes.mdx
```diff
@@ -101,9 +106,91 @@ management tool.
     For the complete list of configuration options, please see the [API
     documentation](/api/auth/kubernetes).
 
+### How to work with short-lived Kubernetes tokens
```
gavinelder, line 109: I think this section would have better value to the reader at the top of the page where the flow is as follows. 

- Kuberneetes Auth Method
    - Note on JWT auth
- Explanation of different Auth methods & choosing whats right for you. 
- Configuration examples

gavinelder (commented):
This looks like a significant improvement and addresses the concerns of https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/106 I appreciate that Kuberneetes & the relevant plugin has a significant amount of flexibility allowing it to be configured in many different ways hence the need for broad guidance. 

I have left some comments however I feel that along side there should be the "Hashicorp best practice" for setting up k8s auth.

website/content/docs/auth/jwt/oidc_providers.mdx
calvn: `timeout 30s kubectl proxy &` might work.

website/content/docs/auth/kubernetes.mdx
tomhjp: I want to keep the pattern of having quick reference setup commands highly visible at the top. However, I agree it would be good to draw attention to this earlier, so I've added a note to the opening paragraph and pulled all the documentation for Kubernetes 1.21 changes under one section.

website/content/docs/auth/kubernetes.mdx
tomhjp: Agreed, I've added a note at the top and updated some of the language to contextualise it better.

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -295,7 +295,8 @@ Configuration steps:
 1. Find the issuer URL of the cluster.
 
    ```bash
-   timeout 30s kubectl proxy &
+   # When you're done, kill this background process, e.g. using: kill %1
```
benashz, line 298: 👍

website/content/docs/auth/jwt/oidc_providers.mdx
tomhjp: Reverted and added a comment instead after Ben's comment about `timeout` not being available sometimes.

website/content/docs/auth/kubernetes.mdx
tomhjp: Good point. I've added some comments about this.

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -253,6 +253,209 @@ vault write auth/oidc/role/your_default_role \
...
+   kubectl create clusterrolebinding oidc-reviewer  \
+      --clusterrole=system:service-account-issuer-discovery \
+      --group=system:unauthenticated
+   ```
+
+1. Find the issuer URL of the cluster.
+
+   ```bash
+   timeout 30s kubectl proxy &
+   ISSUER="$(curl --fail --silent --show-error 127.0.0.1:8001/.well-known/openid-configuration | jq -r '.issuer')"
```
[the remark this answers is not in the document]
tomhjp: After discussing offline, agreed to add a comment instead to try and keep the steps focussed on the main goal of the instructions.

website/content/docs/auth/jwt/oidc_providers.mdx
benashz: Actually I think you can make that `kill %%` which will kill the last job started in the background.

tomhjp: Oh good point. I finally actually looked up jobspecs and `kill %kubectl\ proxy` also works but I prefer `kill %%` for the simplicity.

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -253,6 +253,211 @@ vault write auth/oidc/role/your_default_role \
...
+
+1. Enable and configure JWT auth in Vault.
+
+   1. If Vault is running in Kubernetes:
+
+      ```bash
+      kubectl exec vault-0 -- vault auth enable jwt
+      kubectl exec vault-0 -- vault write auth/jwt/config \
+         oidc_discovery_url=https://kubernetes.default.svc.cluster.local \
+         oidc_discovery_ca_pem=@/run/secrets/kubernetes.io/serviceaccount/ca.crt
```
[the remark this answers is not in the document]
tomhjp: Oops, this was a mistake. Thanks for the catch.

website/content/docs/auth/kubernetes.mdx
```diff
@@ -136,7 +258,6 @@ This value is then used when configuring Kubernetes auth, e.g.:
 
 ```bash
 vault write auth/kubernetes/config \
-  token_reviewer_jwt="$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
   kubernetes_host="https://$KUBERNETES_PORT_443_TCP_ADDR:443" \
   kubernetes_ca_cert=@/var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
```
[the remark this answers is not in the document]
tomhjp: Good point, thanks

website/content/docs/auth/kubernetes.mdx
```diff
@@ -12,6 +12,16 @@ The `kubernetes` auth method can be used to authenticate with Vault using a
...
 
+You can also use a Kubernetes Service Account Token to [log in via JWT auth][k8s-jwt-auth].
+See the section on [How to work with short-lived Kubernetes tokens][short-lived-tokens]
+for a summary of why you might want to use JWT auth instead and how it compares to
+Kubernetes auth.
+
+-> **Note:** If you are upgrading to Kubernetes v1.21+, ensure the config option
+`disable_iss_validation` is set to true. Assuming the default mount path, you
+can check with `vault read -field disable_iss_validation auth/kubernetes/config`.
+See [Kubernetes 1.21][#kubernetes-1.21] below for more details.
```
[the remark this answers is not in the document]
tomhjp: So far my only drawback to doing the short-name links is I keep making mistakes like this 😅

website/content/docs/auth/kubernetes.mdx
```diff
@@ -68,7 +78,10 @@ management tool.
     $ vault auth enable kubernetes
     ```
 
-1.  Use the `/config` endpoint to configure Vault to talk to Kubernetes. Use `kubectl cluster-info` to validate the Kubernetes host address and TCP port. Kubernetes 1.21+ clusters may require setting the service account `issuer`, [as described here](/docs/auth/kubernetes#discovering-the-service-account-issuer). For the list of available configuration options, please see the [API documentation](/api/auth/kubernetes).
+1.  Use the `/config` endpoint to configure Vault to talk to Kubernetes. Use
+`kubectl cluster-info` to validate the Kubernetes host address and TCP port.
+For the list of available configuration options, please see the
+[API documentation](/api/auth/kubernetes).
 
     ```text
```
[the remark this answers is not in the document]
tomhjp: Good spot. This was annoying to figure out, but the fix was to indent the first paragraph of text for this bullet so that the note gets properly parsed as a paragraph within the bullet. I'm still confused how I broke it in the first place though.

website/content/docs/auth/jwt/oidc_providers.mdx
```diff
@@ -253,6 +253,220 @@ vault write auth/oidc/role/your_default_role \
...
+   # 2. Query the jwks_uri specified in /.well-known/openid-configuration
+   # NB: You may need to run this from a pod within the cluster if the $ISSUER
+   # URL is not available outside the cluster.
+   curl "$(curl --fail --silent --show-error "${ISSUER}/.well-known/openid-configuration" | jq -r '.jwks_uri')"
+
+   # Kill the background proxy process when you're done
+   kill %%
+   ```
+
+1. Convert the keys from JWK format to PEM. You can use a CLI tool or an online
```
austingebauer, line 366: I wonder if it'd be possible to suggest configuring the JWT auth method using [jwks_url](https://www.vaultproject.io/api-docs/auth/jwt#jwks_url) instead of [jwt_validation_pubkeys](https://www.vaultproject.io/api-docs/auth/jwt#jwt_validation_pubkeys). That is a simpler setup and would eliminate the step of converting JWKS -> PEM here. That configuration also accounts for the scenario where a public key in the JWKS is rotated, where `jwt_validation_pubkeys` is static (i.e., only uses the public keys you give it). I'm not sure if k8s rotates the keys they use to sign/verify JWTs.

It looks like the JWKS is published publicly from the k8s docs. I don't have much context from the k8s side of things, but it _seems_ possible to configure JWT auth with just the `jwks_url`. 
> When enabled, the Kubernetes API server provides an OpenID Provider Configuration document at /.well-known/openid-configuration and the associated JSON Web Key Set (JWKS) at /openid/v1/jwks.

austingebauer (commented):
One comment on JWT auth configuration. Otherwise, this looks great! Thanks @tomhjp.

taoism4504 (commented):
reviewed content. LG

website/content/docs/auth/jwt/oidc_providers.mdx
tomhjp: My thinking was that `jwt_validation_pubkeys` is definitely second choice to using the OIDC discovery URL for the same benefits you state for `jwks_url` (hence discovery coming first here), but it has some nice benefits too like the fact you can configure it to trust multiple clusters and can be permanently network-partitioned from the cluster.

As a layman, using `jwks_url` seemed basically equivalent to `oidc_discovery_url` in terms of the benefits and drawbacks. Does `jwks_url` have any advantages over `oidc_discovery_url`? You're right that we definitely can configure JWT auth with `jwks_url` for Kubernetes, but I don't think there are any (common?) configurations where Kubernetes could support `jwks_url` but couldn't support `oidc_discovery_url`. If there are distinct advantages it has though, then I'm definitely keen to add `jwks_url` as a third option.

austingebauer: You're right that `jwks_url` doesn't have any distinct advantages over `oidc_discovery_url` other than the issuer check limitation that we discussed outside of this PR. That's the only reason I brought it up. Looks like you have some good guidance in the documentation for `oidc_discovery_url`, so keeping `jwt_validation_pubkeys` documented here sounds good to me 👍

austingebauer (approved):

PR #13257, opened 2021-11-23
recognize ed25519 key type and return PKCS8 format

navaneethrameshan:
Addresses #13256

convertRespToPKCS8 does not recognize the ed25519 key. Changes
to recognize ed25519 key and return its PKCS8 format

Issue #13256, opened 2021-11-23
PKI ed25519 key type response in PKCS8 format fails


navaneethrameshan:
**Describe the bug**
Generating an ed25519 key with the PKI plugin and expecting the key in PKCS8 format fails with error: 
```
* unknown private key type "ed25519"
```

**To Reproduce**
1. Run 
``` 
vault write pki/root/generate/exported \ 
common_name="example.com" \
key_type="ed25519" \
private_key_format="pkcs8"
 ```
2. See error: 
``` 
Error writing data to pki/root/generate/exported: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/pki/root/generate/exported
Code: 500. Errors:

* 1 error occurred:
	* unknown private key type "ed25519"

```

**Expected behavior**
Expect root CA certificate to be generated and the key generation to succeed

**Environment:**
* Vault Server Version (retrieve with `vault status`):
```
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    1
Threshold       1
Version         1.10.0-dev
Storage Type    inmem
Cluster Name    vault-cluster-483668ff
Cluster ID      cca902b2-049e-67a1-6009-647327ba5daf
HA Enabled      false

```
* Vault CLI Version (retrieve with `vault version`):
```
Vault v1.10.0-dev ('7800c45bb7a968757fb18264bee4ef5f2517e62c+CHANGES')

```

stevendpclark:
Closing issue, thanks to @navaneethrameshan for the PR contribution!

The fix was also backported to 1.9 within #13276 and should be available with the next 1.9 release.

PR #13276, opened 2021-11-24
Backport 1.9: recognize ed25519 key type and return PKCS8 format (#13257)

stevendpclark:
Backport of #13257 to address #13256 

* return pkcs8 format for ed25519 curve

convertRespToPKCS8 does not recognize the ed25519 key. Changes
to recognize ed25519 key and return its PKCS8 format


victorr (approved):
👍

Issue #13281, opened 2021-11-25
raft crashes with error="key too large" (DoS)


Slm0n87:
**Describe the bug**
Internal raft storage crashes when user put secrets into deep nested paths.

**To Reproduce**
0. Having a running cluster (or single node) with raft storage backend with a secret-engine "secret"
1. Run `vault login ...`
3. Run `for count in {550..600}; do echo COUNT:$count; vault kv put secret/$(for i in {1..$count}; do echo -n "1/"; done) foo=bar ; done`
3. For me at a count of 564 directories the raft cluster crashed.

**Expected behavior**
vault should prevent raft cluster from crashing

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.9.0
* Vault CLI Version (retrieve with `vault version`): Vault v1.8.4 ('925bc650ad1d997e84fbb832f302a6bfe0105bbb+CHANGES')
* Server Operating System/Architecture: Ubuntu 20.4 LTS

Vault server configuration file(s):

```hcl
# Ansible managed

storage "raft" {
  path    = "/space/raft-storage/xxxxxxxx-cnxyaqpi"
  node_id = "xxxxxxxx"
}

listener "tcp" {
  address = "0.0.0.0:8200"
  cluster_address = "x.x.x.x:8201"
  tls_disable = "true"
  # trust x-forwarded-for header of HA-Proxies
  x_forwarded_for_authorized_addrs = "x.x.x.x/32,x.x.x.x/32"
  x_forwarded_for_reject_not_present = "false"
}

seal "transit" {
  address = "http://x.x.x.x:8200"
  # token is read from VAULT_TOKEN env
  disable_renewal = "true"
  // Key configuration
  key_name           = "unseal_key"
  mount_path         = "transit/"
}

cluster_addr = "http://x.x.x.x:8201"
api_addr = "http://vault-test.app.xxxxxxxx:8200"
cluster_name = "vault"
raw_storage_endpoint = "true"
ui = "true"
```

**Vault systemd logs:**
```
Nov 25 17:42:13 xxxxxxxx vault[24393]: 2021-11-25T17:42:13.678Z [ERROR] storage.raft.fsm: failed to store data: error="key too large"
Nov 25 17:42:13 xxxxxxxx vault[24393]: panic: failed to store data
Nov 25 17:42:13 xxxxxxxx vault[24393]: goroutine 422 [running]:
[...]

PR #13282, opened 2021-11-25
Return an error when trying to store a too-large key with Raft 

ncabatoff:
…instead of panicing.  Fixes https://github.com/hashicorp/vault/issues/13281.


stevendpclark (approved):
👍

PR #13287, opened 2021-11-25
secrets/azure: Fixes use_microsoft_graph_api parameter description in API docs

austingebauer:
This PR fixes the `use_microsoft_graph_api` parameter description in the Azure secrets API docs. See that a portion of the description comes after the [root_password_ttl](https://www.vaultproject.io/api-docs/secret/azure#root_password_ttl) parameter.


benashz (approved):
LGTM!

PR #13308, opened 2021-11-30
Add an option to `vault debug` so that it can query multiple servers

ncabatoff:


command/debug.go
```diff
@@ -255,26 +345,58 @@ func (c *DebugCommand) Run(args []string) int {
...
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			errs[i] = collectors[i].capturePollingTargets()
+		}(i)
 	}
+	wg.Wait()
+	for i, err := range errs {
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error capturing static information for address %s: %s", c.flagAddresses[i], err))
```
swayne275, line 380: polling info rather than static? error message is the same as above

command/debug.go
```diff
@@ -441,29 +553,29 @@ func (c *DebugCommand) preflight(rawArgs []string) (string, error) {
...
 	default:
-		return "", fmt.Errorf("output directory already exists: %s", c.flagOutput)
+		return "", nil, fmt.Errorf("output directory already exists: %s", c.flagOutput)
 	}
 
 	// Populate initial index fields
-	c.debugIndex = &debugIndex{
-		VaultAddress:           client.Address(),
+	debugIndex := &debugIndex{
+		// VaultAddress:           client.Address(),
```
swayne275, line 578: why are we removing this

swayne275, line 578: why are we removing this

command/debug.go
swayne275, line 380: polling info rather than static? error message is the same as above

command/debug_test.go
```diff
@@ -699,3 +689,83 @@ func TestDebugCommand_PartialPermissions(t *testing.T) {
...
+
+	var addrArgs []string
+	for _, core := range cluster.Cores {
+		addrArgs = append(addrArgs, "-addresses", core.Client.Address())
+	}
+	cases := []struct {
+		name string
+		args []string
+	}{
+		{
```
swayne275, line 710: not that critical, but you could add a test that does both addresses and cluster to validate that it errors

command/debug.go
ncabatoff: Fixed, ty

command/debug.go
ncabatoff: This is the "base" debugIndex which is common to all nodes.  Later we use it as the basis for node-specific debugIndexes which include the address.

command/debug_test.go
ncabatoff: Done

PR #13316, opened 2021-11-30
Add iam:GetUser permission to the example.

cjyar:
Without `iam:GetUser` permission, I wasn't able to get Vault to rotate its own credentials.


hsimon-hashicorp:
Thanks for this submission! When the CLA is signed, we'll get this reviewed. :)

hsimon-hashicorp (approved):

PR #13318, opened 2021-12-01
Fix possible nil pointer dereference

reugn:


changelog/13318.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+vault: Fix possible nil pointer dereference.
```
kalafut, line 2:
```suggestion
identity: Fix possible nil pointer dereference.
```

ncabatoff (approved):

ncabatoff:
Looking at this more closely, I don't see us ever deleting StoragePacker buckets, so I don't think this will ever panic.  I'm still happy to have it fixed, if for no other reason than to avoid people looking at it and worrying that it might panic, so thanks @reugn.

PR #13324, opened 2021-12-01
Rename master key to root key

kalafut:
Change internal references from "master key" to "root key". Any storage items (e.g. a key name in an internal data structure)  have not been changed due to the complexity/risk around upgrading or downgrading.


vault/barrier.go
```diff
@@ -142,7 +142,6 @@ type SecurityBarrier interface {
 	// SetRotationConfig updates the auto-rotation config for the barrier key
 	SetRotationConfig(ctx context.Context, config KeyRotationConfig) error
 
-	// Rekey is used to change the master key used to protect the keyring
```
ncabatoff, line 145: Did you mean to remove this?

vault/rekey.go
```diff
@@ -674,23 +674,23 @@ func (c *Core) RecoveryRekeyUpdate(ctx context.Context, key []byte, nonce string
 		return nil, logical.CodedError(http.StatusBadRequest, fmt.Errorf("recovery key verification failed: %w", err).Error())
 	}
 
-	// Generate a new master key
-	newMasterKey, err := c.barrier.GenerateKey(c.secureRandomReader)
+	// Generate a new root key
```
ncabatoff, line 677: I think many of the master/root key names in this little section should probably be unseal instead, since I see that we're splitting with shamir.

ncabatoff: My mistake, not unseal, recovery.

ncabatoff (approved):
A few minor tweaks, overall looks good

vault/barrier.go
kalafut: Nope. Restored.

calvn (commented):
Should we also update the references on the doc pages to `root key` as well? On a related note, I think that there was concern at one point of root key being confused with [root token](https://www.vaultproject.io/docs/concepts/tokens#root-tokens) and having those two being used interchangeably by mistake.

kalafut:
@calvn Yes, there are a lot of docs, ui and learn content changes. We have them broken out as separate tasks in the epic. Re confusion with "root token", I don't recall that specifically but there was a lot of discussion over the replacement term when this all began and "root key" was selected.

calvn (approved):

hsimon-hashicorp:
> @calvn Yes, there are a lot of docs, ui and learn content changes. We have them broken out as separate tasks in the epic. Re confusion with "root token", I don't recall that specifically but there was a lot of discussion over the replacement term when this all began and "root key" was selected.

Additionally I've talked with @taoism4504 about this specifically, within the past couple of weeks. We'll be ready. :)

PR #13325, opened 2021-12-01
[Homepage] populate use case dropdown from use case pages

alexcarpenter:
Following the same pattern we use on dotcom.


BrandonRomano (approved):

jmfury (approved):
💯  - Thanks for the tag.

website/pages/_app.js
```diff
@@ -59,11 +60,25 @@ export default function App({ Component, pageProps }) {
...
+        <Component {...pageProps} />
+      </Layout>
       <ConsentManager className="g-consent-manager" />
     </ErrorBoundary>
   )
 }
+
+App.getInitialProps = async ({ Component, ctx }) => {
+  const layoutQuery = Component.layout
+    ? Component.layout?.rivetParams ?? null
```
jmfury, line 73: They're not perfect, and it's not a blocker by any means, but here's what I put in the global types for `RivetParams`

```ts
interface RivetParams {
  query: string
  dependencies?: any[]
  variables?: Record<string, any>
  fragments?: any[]
}
```

alexcarpenter: Will keep in mind, currently app is not a TS file.

PR #13328, opened 2021-12-02
Finish removing help page

mladlow:
I'm not sure this is correct, and we may want to put this help page back in the side nav, but the docs were broken for me locally.

Related to https://github.com/hashicorp/vault/pull/13198


mladlow:
Superseded by https://github.com/hashicorp/vault/pull/13329

PR #13336, opened 2021-12-03
GetCredentials Bug fix

Monkeychip:
A regression bug caused by a required property added to the GetCredentialsCard component that was not added to a previous instance of the component.

Will backport to 1.9.x


hashishaw (approved):
And a test! 😍

Issue #13347, opened 2021-12-06
Add certificate extensions as metadata


peterverraedt:
**Is your feature request related to a problem? Please describe.**
I'm running a configuration management tool (puppet) on ~1000 nodes in a decentralized manner, and each node needs access to vault, limited to a number of scopes depending on the node properties (role such as webserver/compute node/..., hostname, ...). The number of nodes and the set of possible "roles" changes regularly, and I want rather static vault policies based on identity metadata.

This is currently partially possible by using the puppet host certificates combined with the cert authentication method: the certificate name, common name and serial number are already available as identity metadata and can be used in policy templates. This means that I can already create a policy that restricts read access to paths in vault containing the hostname (= certificate common name) as a variable. However, I would need more key value metadata retrieved from the certificate to work efficiently.

**Describe the solution you'd like**
All certificate ASN1 extensions should be added as key/value pairs in the metadata.

**Describe alternatives you've considered**
Creating a single approle role and different secret_id's for each host, with metadata, but I ran into #11798. Creating a role_id per host is too dynamic and would require manually creating policies and assignments, when my set of host "roles" grows. 

**Explain any additional use-cases**
Any situation where certificates carry embedded metadata as ASN1 extensions, would benifit from the possibility to use it in vault policy templating.

pmmukh:
Hey @peterverraedt ! Thanks for submitting this feature request and the linked PR, the ask makes sense and the solution looks to be on the right path! :) The only major concern we have with the approach, is that it might lead to large amounts of metadata getting added to entities and thus eating up disk space, leading to performance problems. One solution we've thought of for that problem, is to make this functionality opt-in, and also to be more explicit about which extensions to add into the entity_metadata. 
We think that adding a field to https://www.vaultproject.io/api-docs/auth/cert#create-ca-certificate-role specifying the valid extension Ids to add , and checking the cert extensions against this list when logging in with the certificate, adding only those extensions that were explicitly requested, would make sense. The list would be an optional argument, so for anyone not specifying it when creating the cert role, the entity metadata would be populated as it is today. Please let me know how that addition to the approach sounds to you!

peterverraedt:
Thanks for the feedback, @pmmukh . I've updated the PR to include a `metadata_extensions` field in the role configuration.

voiprodrigo:
[...]

Issue #10503, opened 2020-12-05
Support custom x509v3 extension key/value pairs of ASN1.OID format


gallak87:
**Is your feature request related to a problem? Please describe.**
I am running a hyper ledger fabric blockchain network and identities are managed with x509 certificates. A required custom x509v3 extension uses a key of `1.2.3.4.5.6.7.8.1`, with some json blob as a value. I cannot figure out how to get vault CLI to set this custom extension attribute. I tried using `allowed_other_sans` feature, and set `other_sans` in my issue call to the key/value with json blob but the resulting output was incorrect. I suppose it doesn't qualify as a "san" but a custom extension attribute.

**Describe the solution you'd like**
I would like vault to support setting custom x509v3 extension attributes such that my output certificate looks like the following (omitted irrelevant cert info):

```
Certificate:
  ...
    Subject Public Key Info:
     ...
        ASN1 OID: prime256v1
        NIST CURVE: P-256
    X509v3 extensions:
    ...
      X509v3 Subject Alternative Name: 
        DNS:1b80cd8b1cbe
      1.2.3.4.5.6.7.8.1: 
        {"attrs":{"company.ADDR":"XZPYGWWht3fRa2rgMVA3dxpfCW4dcUDXUdsaUFwh17FLvJUDm","company.Affiliation":"Company1","hf.Affiliation":"company1","hf.EnrollmentID":"company1.domain.com","hf.Type":"orderer"}}
  Signature Algorithm: ecdsa-with-SHA256
...
```

Specifically the last bit here with the key/value pair: 
```
      1.2.3.4.5.6.7.8.1: 
        {"attrs":{"company.ADDR":"XZPYGWWht3fRa2rgMVA3dxpfCW4dcUDXUdsaUFwh17FLvJUDm","company.Affiliation":"Company1","hf.Affiliation":"company1","hf.EnrollmentID":"company1.domain.com","hf.Type":"orderer"}}
```

This is a custom x509v3 extension attribute that I suppose is unnamed with a key of `1.2.3.4.5.6.7.8.1`

**Describe alternatives you've considered**
I want vault to be my CA for my hyper ledger fabric identity management. These are long-lived identities (2 years max) and I want to centralize my platform identity management around vault. 

**Explain any additional use-cases**
n/a

**Additional context**
[...]

PR #13348, opened 2021-12-06
auth/cert: Add certificate extensions as metadata

peterverraedt:
This adds certificate extensions as metadata to be used in vault policy templates. We parse the ASN1 objectidentifiers to a string separated by dashes instead of dots, to ease the use in golang templating language.

Fixes: #13347

Refs: #10503 

Signed-off-by: Peter Verraedt <peter.verraedt@kuleuven.be>

### DIFF
--- a/builtin/credential/cert/backend_test.go
+++ b/builtin/credential/cert/backend_test.go
@@ -1107,6 +1107,10 @@ func TestBackend_ext_singleCert(t *testing.T) {
 			testAccStepLoginInvalid(t, connState),
 			testAccStepCert(t, "web", ca, "foo", allowed{names: "invalid", ext: "2.1.1.1:*,2.1.1.2:The Wrong Value"}, false),
 			testAccStepLoginInvalid(t, connState),
+			testAccStepCert(t, "web", ca, "foo", allowed{metadata_ext: "2.1.1.1,1.2.3.45"}, false),
+			testAccStepLoginWithMetadata(t, connState, "web", map[string]string{"2-1-1-1": "A UTF8String Extension"}),
+			testAccStepCert(t, "web", ca, "foo", allowed{metadata_ext: "1.2.3.45"}, false),
+			testAccStepLoginWithMetadata(t, connState, "web", map[string]string{}),
 		},
 	})
 }
@@ -1556,6 +1560,40 @@ func testAccStepLoginDefaultLease(t *testing.T, connState tls.ConnectionState) l
 	}
 }
 
+func testAccStepLoginWithMetadata(t *testing.T, connState tls.ConnectionState, certName string, metadata map[string]string) logicaltest.TestStep {
+	return logicaltest.TestStep{
+		Operation:       logical.UpdateOperation,
+		Path:            "login",
+		Unauthenticated: true,
+		ConnState:       &connState,
+		Check: func(resp *logical.Response) error {
+			// Check for fixed metadata too
+			metadata["cert_name"] = certName
+			metadata["common_name"] = connState.PeerCertificates[0].Subject.CommonName
+			metadata["serial_number"] = connState.PeerCertificates[0].SerialNumber.String()
+			metadata["subject_key_id"] = certutil.GetHexFormatted(connState.PeerCertificates[0].SubjectKeyId, ":")
+			metadata["authority_key_id"] = certutil.GetHexFormatted(connState.PeerCertificates[0].AuthorityKeyId, ":")
+
+			for key, expected := range metadata {
+				value, ok := resp.Auth.Metadata[key]
+				if !ok {
+					t.Fatalf("missing metadata key: %s", key)
+				}
+
+				if value != expected {
+					t.Fatalf("expected metadata key %s to equal %s, but got: %s", key, expected, value)
+				}
+			}
+
+			fn := logicaltest.TestCheckAuth([]string{"default", "foo"})
+			return fn(resp)
+		},
+		Data: map[string]interface{}{
+			"metadata": metadata,
+		},
+	}
+}
+
 func testAccStepLoginInvalid(t *testing.T, connState tls.ConnectionState) logicaltest.TestStep {
 	return testAccStepLoginWithNameInvalid(t, connState, "")
 }
@@ -1633,6 +1671,7 @@ type allowed struct {
 	uris                 string // allowed uris in SAN extension of the certificate
 	organizational_units string // allowed OUs in the certificate
 	ext                  string // required extensions in the certificate
+	metadata_ext         string // allowed metadata extensions to add to identity alias
 }
 
 func testAccStepCert(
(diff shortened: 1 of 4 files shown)


pmmukh:
Hey @peterverraedt ! Sorry for the delay here, I've been caught up with other stuff but I'll be sure to review this PR this week, and thanks for the prompt response to the feedback we gave!

builtin/credential/cert/backend_test.go
```diff
@@ -1652,6 +1691,7 @@ func testAccStepCert(
 			"allowed_uri_sans":             testData.uris,
 			"allowed_organizational_units": testData.organizational_units,
 			"required_extensions":          testData.ext,
+			"metadata_extensions":          testData.metadata_ext,
```
pmmukh, line 1694: Would it make sense to rename this to something like `allowed_metadata_extensions` ? I think having the field name reflect the purpose, that extensions with these oids are allowed to be added into identity metadata, like allowed_uri_sans / allowed_organizational_units, might be clearer ?

builtin/credential/cert/path_certs.go
```diff
@@ -114,6 +114,15 @@ formatted as "oid:value". Expects the extension value to be some type of ASN1 en
 All values much match. Supports globbing on "value".`,
 			},
 
+			"metadata_extensions": {
+				Type: framework.TypeCommaStringSlice,
+				Description: `A comma-separated string or array of extensions
+formatted as "oid" or "oid:alias". Upon successfull authentication, these extensions will
```
pmmukh, line 120: Couple points here.
1. Do we need the 2 forms, oid and oid:alias ? As in, if we just have a comma separate list of oids, and always use the santiized ( dots to dashes ) version of it as the metadata key, would there be an issue ? I'm guessing redabale metadata keys is the argument there, but wanted to check.
2. If we do want the oid:alias form, I'd recommend using a different word and not `alias`, as that is a loaded word, being a key concept in the identity system.

builtin/credential/cert/path_login.go
```diff
@@ -409,6 +419,47 @@ func (b *backend) matchesCertificateExtensions(clientCert *x509.Certificate, con
...
+	// If no metadata extensions are configured, return an empty map
+	if len(config.Entry.MetadataExtensions) == 0 {
+		return map[string]string{}
+	}
+
+	// Build a map of the extensions that should be set as metadata
+	oidMap := map[string]string{}
+	for _, metadataExt := range config.Entry.MetadataExtensions {
+		parts := strings.SplitN(metadataExt, ":", 2)
+		if len(parts) == 1 {
```
pmmukh, line 434: A comment here explaining the conditional might be nice imo

builtin/credential/cert/path_login.go
```diff
@@ -409,6 +419,47 @@ func (b *backend) matchesCertificateExtensions(clientCert *x509.Certificate, con
...
+			// Avoid dots in metadata keys - hard to work with in policy templates.
+			oidMap[metadataExt] = strings.ReplaceAll(metadataExt, ".", "-")
+
```
pmmukh, line 437:
```suggestion
```

builtin/credential/cert/path_login.go
```diff
@@ -409,6 +419,47 @@ func (b *backend) matchesCertificateExtensions(clientCert *x509.Certificate, con
...
+			continue
+		}
+
+		oidMap[parts[0]] = parts[1]
+	}
+
+	metadata := make(map[string]string, len(config.Entry.MetadataExtensions))
+	for _, ext := range clientCert.Extensions {
+		for oidString, alias := range oidMap {
```
pmmukh, line 446: Do we need to loop through the map in here with the string checking, or could we do something like
`if alias, ok:= oidMap[ext.IdString()]; ok {` ?
The suggestion is less for a time complexity improvement, which I think doesn't matter tbh, and more for readability.

builtin/credential/cert/path_login.go
```diff
@@ -89,19 +89,29 @@ func (b *backend) pathLogin(ctx context.Context, req *logical.Request, data *fra
...
+	metadata := map[string]string{
+		"cert_name":        matched.Entry.Name,
+		"common_name":      clientCerts[0].Subject.CommonName,
+		"serial_number":    clientCerts[0].SerialNumber.String(),
+		"subject_key_id":   certutil.GetHexFormatted(clientCerts[0].SubjectKeyId, ":"),
+		"authority_key_id": certutil.GetHexFormatted(clientCerts[0].AuthorityKeyId, ":"),
+	}
+
+	for k, v := range b.certificateExtensionsMetadata(clientCerts[0], matched) {
+		if _, ok := metadata[k]; ok {
```
pmmukh, line 101: Could you add a comment explaining this conditional too please ?

pmmukh (commented):
Some comments / questions, overall the PR looks pretty great! :)

builtin/credential/cert/backend_test.go
peterverraedt: I wanted to initially avoid `allowed_metadata_extensions` as the other `allowed_*` keys restrict access to the authentication endpoint for certificates lacking those properties. But I follow you, and it indeed makes sense to put `allowed_metadata_extensions`, as it is a filter on the oids that we'll accept to set as metadata.

builtin/credential/cert/path_certs.go
peterverraedt: We don't need aliases apart from readability wherever the metadata is used, and actually putting the dashes-separated oid as metadata key will make it clear throughout that the metadata originated as certificate extensions. As we're renaming the configuration value to `allowed_metadata_extensions`, it also makes more sense to just accept a list of oids. I'll remove the aliasing.

changelog/13348.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
pmmukh, line 1: I think this should be `improvement` not `feature`, as `feature` usually refers to large changes in a major release.

builtin/credential/cert/path_certs.go
```diff
@@ -114,6 +114,14 @@ formatted as "oid:value". Expects the extension value to be some type of ASN1 en
 All values much match. Supports globbing on "value".`,
 			},
 
+			"allowed_metadata_extensions": {
+				Type: framework.TypeCommaStringSlice,
+				Description: `A comma-separated string or array of oid extensions.
```
pmmukh, line 119: Does a comma separated string work here ? I didn't see a split by comma way of parsing the input in the code, hence figured i'd check.

pmmukh (approved):
couple minor nits, olgtm!

pmmukh:
Also, could you please rebase this branch off main ? There's some test failures that should be fixed in main now.

builtin/credential/cert/path_certs.go
pmmukh: ugh i missed the `TypeCommaStringSlice` there, please ignore!

PR #13361, opened 2021-12-07
UI/ horizontal bar chart component

hellobontempo:
Adjust CSS to accommodate side by side charts
Screen Shot 2021-12-09 at 3 09 09 PM


ui/app/components/clients/horizontal-bar-chart.js
```diff
@@ -0,0 +1,200 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { assert } from '@ember/debug';
+import { stack } from 'd3-shape';
+// eslint-disable-next-line no-unused-vars
```
Monkeychip, line 5: what's the unused var?

ui/app/components/clients/horizontal-bar-chart.js
```diff
@@ -0,0 +1,200 @@
...
+ * @param {array} chartLegend - array of objects with key names 'key' and 'label' for the map legend
+ * @param {string} [labelKey=label] - labelKey is the key name in the dataset passed in that corresponds to the value labeling the y-axis (i.e. 'namespace_path')
+ * @param {string} [param1=defaultValue] - param1 is...
+ */
+
+// TODO: delete original bar chart component
+// TODO: Move constants to helper
+
+// SIZING CONSTANTS
+const CHART_MARGIN = { top: 10, left: 137 }; // makes space for y-axis legend
```
Monkeychip, line 29: nice comments

ui/tests/integration/components/horizontal-bar-chart-test.js
```diff
@@ -0,0 +1,26 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | horizontal-bar-chart', function(hooks) {
```
Monkeychip, line 6: I deleted my default test file, feel free to keep it but maybe make a note to add or amend to it.

Monkeychip (approved):
nice. unsure about the overflow. I set the chart-container height, but I would think the height of the svg was a percent so it would fit, but looks like with the space between the bars it doesn't. We should check with ivana if she want's consistent heights or if they can be different.

ui/app/components/clients/horizontal-bar-chart.js
hellobontempo: for some reason eslint underlines `selectAll` even though it _is_ used - so the comment just tells linter to ignore 
Screen Shot 2021-12-07 at 4 31 43 PM

ui/tests/integration/components/horizontal-bar-chart-test.js
hellobontempo: Ah yeah - I was wondering what to do with it... I'll add a note to fix

PR #13366, opened 2021-12-08
Warn user supplying nonce values in FIPS mode for transit encryption requests

stevendpclark:
- Send back a warning within the response if an end-user supplies nonce values that we use within the various transit encrypt apis.
 - We do not send a warning if an end-user supplies a nonce value but we don't use it.
 - Affected api methods are encrypt, rewrap and datakey
 - The warning is only sent when we are operating in FIPS mode.

### DIFF
--- a/builtin/logical/transit/path_datakey.go
+++ b/builtin/logical/transit/path_datakey.go
@@ -5,8 +5,8 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"fmt"
-
 	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/helper/consts"
 	"github.com/hashicorp/vault/sdk/helper/errutil"
 	"github.com/hashicorp/vault/sdk/helper/keysutil"
 	"github.com/hashicorp/vault/sdk/logical"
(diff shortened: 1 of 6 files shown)


schultz-is (approved):
lgtm

builtin/logical/transit/path_datakey.go
```diff
@@ -159,6 +159,10 @@ func (b *backend) pathDatakeyWrite(ctx context.Context, req *logical.Request, d
 		},
 	}
 
+	if consts.IsFIPS() && shouldWarnAboutNonceUsage(p, nonce) {
+		resp.AddWarning("A provided nonce value was used within FIPS mode, this violates FIPS standards.")
```
sgmiller, line 163: "violates FIPS 140 compliance."

sdk/helper/consts/fips.go
```diff
@@ -0,0 +1,9 @@
+//go:build !fips_140_3
+// +build !fips_140_3
```
sgmiller, line 2: Only need the second of these, on the very first line.

sgmiller (changes requested):
Couple small things.

sdk/helper/consts/fips.go
stevendpclark: 😞 I need to figure out the GoLand setting that keeps adding these back...

builtin/logical/transit/path_datakey.go
stevendpclark: Fixed.

sgmiller (approved):

PR #13370, opened 2021-12-08
[VAULT-3252] Add entity-alias behavior change to docs

pmmukh:
- Adds the change to the docs indicating that an entity can only have 1 alias per auth mount. This had been previously added to a now removed page `website/content/docs/secrets/identity.mdx` in this PR https://github.com/hashicorp/vault/pull/12747/files, but it appears to have gotten lost in the shuffle, hence re-adding it back in this PR
- Also documents the change as an upgrade note for 1.7, 1.8 and 1.9
- Renames the 1.7-1.9 upgrade pages to be for 1.7/8/9.x instead of 1.7/8/9.0, as well as similar changes inside the docs

### DIFF
--- a/website/content/partials/entity-alias-mapping.mdx
+++ b/website/content/partials/entity-alias-mapping.mdx
@@ -1,6 +1,6 @@
 ## Entity Alias mapping
 
 Previously, an entity in Vault could be mapped to multiple entity aliases on the same authentication backend. This
-led to a potential security vulnerability, as ACL policies templated with alias information, would match the first
+led to a potential security vulnerability (CVE-2021-43998), as ACL policies templated with alias information would match the first
 alias created. As a result, the mapping behavior was updated such that an entity can only have one alias per authentication
 backend. This change exists in Vault 1.9.0+, 1.8.5+ and 1.7.6+.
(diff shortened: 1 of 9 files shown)


hghaf099 (approved):
LGTM!
Should doc/cherrypick be added in the labels?

pmmukh:
> LGTM! Should doc/cherrypick be added in the labels?

It should, thanks for the catch!

website/content/partials/entity-alias-mapping.mdx
```diff
@@ -0,0 +1,6 @@
+## Entity Alias mapping
+
+Previously, an entity in Vault could be mapped to multiple entity aliases on the same authentication backend. This
+led to a potential security vulnerability, as ACL policies templated with alias information, would match the first
```
mladlow, line 4:
```suggestion
led to a potential security vulnerability (CVE-2021-43998), as ACL policies templated with alias information would match the first
```

website/content/partials/entity-alias-mapping.mdx
```diff
@@ -0,0 +1,6 @@
...
+alias created. As a result, the mapping behavior was updated such that an entity can only have one alias per authentication
```
mladlow, line 5: This doesn't quite cover the impact of ACL policy templating matching the first alias created - could you please add a statement about the impact before "As a result"?

pmmukh: added a note about that here 21edb57, lemme know how it looks!

mladlow (approved):
Thanks for the extra changes!

PR #13385, opened 2021-12-09
Some changelog tidying for 1.10 preview

mladlow:
@raskchanky - storage/raft seemed more used than just "raft" as a subsystem in the changelog, but I wasn't 100% sure whether we used both prefixes to mean different things.

@divyapola5 - I wasn't sure what subsystem to use for the mount tune change. Maybe `config`?

Most of this is just FYI notes about feature formatting vs improvements. If any of these should be called out as new features in the FEATURES section though, could you make a change to this branch with what the description of the new feature should be?

### DIFF
--- a/changelog/12792.txt
+++ b/changelog/12792.txt
@@ -1,3 +1,3 @@
-```release-note: feature
-core: reading `sys/mounts/:path` now returns the configuration for the secret engine at the given path
+```release-note:improvement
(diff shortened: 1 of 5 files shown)


digivava (approved):

raskchanky (approved):

cipherboy (approved):
Thanks @mladlow and sorry for not understanding the difference during initial review! :-)

PR #13390, opened 2021-12-10
Remove another use gopsutil/host.

ncabatoff:


ccapurso:
It looks like we also need to remove this https://github.com/hashicorp/vault/blob/fix-openbsd-build-take2/helper/hostutil/hostinfo.go#L47. `host.Info` simply calls into `host.InfoWithContext` with a background context.

ccapurso (approved):

PR #13391, opened 2021-12-10
Upgrade to golang.org/x/net/http2 v0.0.0-20211209124913-491a49abca63 …

ncabatoff:
…to address CVE-2021-44716 and Go issue go.dev/issue/50058.


command/server.go
```diff
@@ -668,6 +671,12 @@ func (c *ServerCommand) runRecoveryMode() int {
 			IdleTimeout:       5 * time.Minute,
 			ErrorLog:          c.logger.StandardLogger(nil),
 		}
+		conf := &http2.Server{
+			NewWriteScheduler: func() http2.WriteScheduler {
+				return http2.NewPriorityWriteScheduler(nil)
+			},
+		}
+		http2.ConfigureServer(server, conf)
```
ncabatoff, line 679: TODO handle this error, but where?

tsaarni: I wonder why is this needed, isn't this same as what [gets set by default](https://github.com/golang/go/blob/1afa432ab93aa9adb2e0f04b6c15eb654762d652/src/net/http/server.go#L3299-L3311)?

tsaarni: OK now I think I understand: by setting `NewWriterScheduler` like this, the server implementation will use (at least relevant parts of) `x/net/http` instead of `net/http` and go version upgrade is avoided?  This is bit tricky to understand, so maybe worth of comment?

ncabatoff: We prefer not to upgrade Go major versions on a Vault release branch, correct.

ncabatoff: Oh, and this was an experiment that didn't pan out, FYI.

tsaarni: @ncabatoff ok thanks for the information.  I wonder do you have mitigation plan for CVE-2021-44716? I ended up here by searching for that.

ncabatoff: I'm now leaning towards breaking with precedent and upgrading to Go 1.16.12 on release/1.7.x.  But that's not settled yet.

ncabatoff: We've decided to go ahead with 1.16.12 on 1.7.x, see https://github.com/hashicorp/vault/pull/13417.

PR #13396, opened 2021-12-10
UI/fix client count partial

hashishaw:
Fixes bug where the partial client counts don't show up unless there is data in the monthly tab. 

With data collection off:
image

With data collection on, but no data present:
Screen Shot 2021-12-10 at 2 06 42 PM


Monkeychip (approved):
Nice work!

Monkeychip (approved):

Issue #13410, opened 2021-12-13
Standalone-TLS document is outdated


phozzy:
The new k8s certificates API version makes the document "Standalone Server with TLS" outdated.
PR is coming.

PR #13411, opened 2021-12-13
Update docs 4 standalone tls

phozzy:
Solves #13410
Updates the document to meet the changes to k8s certificates API.


phozzy:
I have signed the agreement, but the status is still pending.

hsimon-hashicorp:
@phozzy Usually when we see that the CLA has been signed but isn't showing as signed in the PR, it's because the CLA was signed with a different email address than your GitHub user. :)

PR #13414, opened 2021-12-14
Add parameter to disable escaping the username and password for database connections

robmonte:
This feature adds the ability to disable Vault escaping any special characters in the username and password fields when connecting to a database. When writing the database config, supplying `disable_escaping="true"` will activate this feature. It is off by default/not specified.

### DIFF
--- a/sdk/database/helper/connutil/sql.go
+++ b/sdk/database/helper/connutil/sql.go
@@ -58,11 +58,11 @@ func (c *SQLConnectionProducer) Init(ctx context.Context, conf map[string]interf
 
 	// Do not allow the username or password template pattern to be used as
 	// part of the user-supplied username or password
-	if c.DisableEscaping &&
-		(strings.Contains(c.Username, "{{username}}") ||
-			strings.Contains(c.Username, "{{password}}") ||
-			strings.Contains(c.Password, "{{username}}") ||
-			strings.Contains(c.Password, "{{password}}")) {
+	if strings.Contains(c.Username, "{{username}}") ||
+		strings.Contains(c.Username, "{{password}}") ||
+		strings.Contains(c.Password, "{{username}}") ||
+		strings.Contains(c.Password, "{{password}}") {
+
 		return nil, fmt.Errorf("username and/or password cannot contain the template variables")
 	}
 
--- a/sdk/database/helper/connutil/sql.go
+++ b/sdk/database/helper/connutil/sql.go
@@ -59,10 +59,10 @@ func (c *SQLConnectionProducer) Init(ctx context.Context, conf map[string]interf
 	// Do not allow the username or password template pattern to be used as
 	// part of the user-supplied username or password
 	if c.DisableEscaping &&
-		strings.Contains(c.Username, "{{username}}") ||
-		strings.Contains(c.Username, "{{password}}") ||
-		strings.Contains(c.Password, "{{username}}") ||
-		strings.Contains(c.Password, "{{password}}") {
+		(strings.Contains(c.Username, "{{username}}") ||
+			strings.Contains(c.Username, "{{password}}") ||
+			strings.Contains(c.Password, "{{username}}") ||
+			strings.Contains(c.Password, "{{password}}")) {
 		return nil, fmt.Errorf("username and/or password cannot contain the template variables")
 	}
 
--- a/sdk/database/helper/connutil/sql.go
+++ b/sdk/database/helper/connutil/sql.go
@@ -26,6 +26,7 @@ type SQLConnectionProducer struct {
 	MaxConnectionLifetimeRaw interface{} `json:"max_connection_lifetime" mapstructure:"max_connection_lifetime" structs:"max_connection_lifetime"`
 	Username                 string      `json:"username" mapstructure:"username" structs:"username"`
 	Password                 string      `json:"password" mapstructure:"password" structs:"password"`
+	DisableEscaping          bool        `json:"disable_escaping" mapstructure:"disable_escaping" structs:"disable_escaping"`
 
 	Type                  string
 	RawConfig             map[string]interface{}
@@ -55,16 +56,32 @@ func (c *SQLConnectionProducer) Init(ctx context.Context, conf map[string]interf
 		return nil, fmt.Errorf("connection_url cannot be empty")
 	}
 
+	// Do not allow the username or password template pattern to be used as
+	// part of the user-supplied username or password
+	if c.DisableEscaping &&
+		strings.Contains(c.Username, "{{username}}") ||
+		strings.Contains(c.Username, "{{password}}") ||
+		strings.Contains(c.Password, "{{username}}") ||
+		strings.Contains(c.Password, "{{password}}") {
+		return nil, fmt.Errorf("username and/or password cannot contain the template variables")
+	}
+
 	// Don't escape special characters for MySQL password
+	// Also don't escape special characters for the username and password if
+	// the disable_escaping parameter is set to true
+	username := c.Username
 	password := c.Password
-	if c.Type != "mysql" {
+	if !c.DisableEscaping {
+		username = url.PathEscape(c.Username)
+	}
+	if (c.Type != "mysql") && !c.DisableEscaping {
 		password = url.PathEscape(c.Password)
 	}
 
 	// QueryHelper doesn't do any SQL escaping, but if it starts to do so
 	// then maybe we won't be able to use it to do URL substitution any more.
 	c.ConnectionURL = dbutil.QueryHelper(c.ConnectionURL, map[string]string{
-		"username": url.PathEscape(c.Username),
+		"username": username,
 		"password": password,
 	})
 
--- a/changelog/13414.txt
+++ b/changelog/13414.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+secrets/database: Add database configuration parameter 'disable_escaping' for username and password when connecting to a database.
+```
--- a/sdk/database/helper/connutil/sql_test.go
+++ b/sdk/database/helper/connutil/sql_test.go
@@ -111,24 +111,28 @@ func TestSQLDisallowTemplates(t *testing.T) {
 		{"abc{{username}}xyz", "123{{password}}789"},
 		{"abc{{{username}}}xyz", "123{{{password}}}789"},
 	}
-	for _, tc := range testCases {
-		t.Logf("username %q password %q", tc.Username, tc.Password)
+	for _, disableEscaping := range []bool{true, false} {
+		for _, tc := range testCases {
+			t.Logf("username %q password %q disable_escaping %t", tc.Username, tc.Password, disableEscaping)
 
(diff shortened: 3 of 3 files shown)


hsimon-hashicorp:
Thanks for this submission! Can you please add a changelog entry? I'll get a reviewer for you as soon as I can. Thanks!

builtin/logical/database/path_config_connection.go
```diff
@@ -31,6 +31,8 @@ type DatabaseConfig struct {
 	RootCredentialsRotateStatements []string `json:"root_credentials_rotate_statements" structs:"root_credentials_rotate_statements" mapstructure:"root_credentials_rotate_statements"`
 
 	PasswordPolicy string `json:"password_policy" structs:"password_policy" mapstructure:"password_policy"`
+
+	DisableEscaping bool `json:"disable_escaping" structs:"disable_escaping" mapstructure:"disable_escaping"`
```
austingebauer, line 35: From what I've observed, it looks like all configuration parameters that are consumed on the plugin side (via [sql.go#L43](https://github.com/hashicorp/vault/blob/main/sdk/database/helper/connutil/sql.go#L43)) don't get added to this list of fields. Examples of those fields are `connection_url`, `username`, `password`, etc. (defined in [sql.go#L23-L28](https://github.com/hashicorp/vault/blob/main/sdk/database/helper/connutil/sql.go#L23-L28)). I'm thinking we could treat `disable_escaping` similarly and remove this?

If we do remove this from `DatabaseConfig`, then we can also remove the `FieldSchema` entry.

builtin/logical/database/path_config_connection.go
```diff
@@ -292,6 +303,7 @@ func (b *databaseBackend) connectionWriteHandler() framework.OperationFunc {
 		delete(data.Raw, "verify_connection")
 		delete(data.Raw, "root_rotation_statements")
 		delete(data.Raw, "password_policy")
+		delete(data.Raw, "disable_escaping")
```
austingebauer, line 306: I don't think we need to delete `disable_escaping` here, since we end up adding it right back to [ConnectionDetails](https://github.com/hashicorp/vault/pull/13414/files#diff-25b7b056201175f5a21ca0e438bebb95f83648322ec77dd6bde35598e95eb511R326) below. I'd suggest to leave it in `data.Raw` and let [this logic](https://github.com/hashicorp/vault/blob/main/builtin/logical/database/path_config_connection.go#L304-L313) add it to `ConnectionDetails`. After that, you can remove the line on 326, since it will already be added.

sdk/database/helper/connutil/sql.go
```diff
@@ -56,15 +56,22 @@ func (c *SQLConnectionProducer) Init(ctx context.Context, conf map[string]interf
 	}
 
 	// Don't escape special characters for MySQL password
+	// Also don't escape special characters for the username and password if
+	// the disable_escaping parameter is set to true
+	disableEscaping := c.RawConfig["disable_escaping"].(bool)
```
austingebauer, line 61: I wonder if we want to add `disable_escaping` to [SQLConnectionProducer](https://github.com/hashicorp/vault/blob/main/sdk/database/helper/connutil/sql.go#L22-L28) and let mapstructure do the [decoding](https://github.com/hashicorp/vault/blob/main/sdk/database/helper/connutil/sql.go#L49) for us? Right now, we're pulling the configuration value out of `RawConfig`, which may be okay. It appears that everything else is being decoded by mapstructure though.

austingebauer (commented):
Nice job, @robmonte! Have a few questions/suggestions around where the configuration parameter should live. +1 @hsimon-hashicorp on changelog entry needed. Would also be nice to add some level of test coverage for this.

I was able to successfully test this using an ADO style connection string and MSSQL. One observation about the ADO style connection string is that we no longer [warn](https://github.com/hashicorp/vault/blob/main/builtin/logical/database/path_config_connection.go#L351-L359) for a password in `connection_url` or [redact](https://github.com/hashicorp/vault/blob/main/builtin/logical/database/path_config_connection.go#L198-L206) when reading the config. This is only true is the username/password is supplied directly in the `connection_url`, but I thought it'd be worth mentioning.

robmonte:
Still working on updating sql_test.go

sdk/database/helper/connutil/sql.go
```diff
@@ -55,16 +56,32 @@ func (c *SQLConnectionProducer) Init(ctx context.Context, conf map[string]interf
 		return nil, fmt.Errorf("connection_url cannot be empty")
 	}
 
+	// Do not allow the username or password template pattern to be used as
+	// part of the user-supplied username or password
+	if c.DisableEscaping &&
+		strings.Contains(c.Username, "{{username}}") ||
+		strings.Contains(c.Username, "{{password}}") ||
+		strings.Contains(c.Password, "{{username}}") ||
+		strings.Contains(c.Password, "{{password}}") {
```
robmonte, line 65: Do these OR conditions need to be encased in parentheses? Or would maybe a helper function be better here? I wasn't sure if a helper function would be overkill for a single if statement.

robmonte: Addressed with https://github.com/hashicorp/vault/commit/945fa1eeaf847a019df047ac20a67ca3c4bd5910

sdk/database/helper/connutil/sql.go
```diff
@@ -55,16 +56,32 @@ func (c *SQLConnectionProducer) Init(ctx context.Context, conf map[string]interf
 		return nil, fmt.Errorf("connection_url cannot be empty")
 	}
 
+	// Do not allow the username or password template pattern to be used as
+	// part of the user-supplied username or password
+	if c.DisableEscaping &&
```
calvn, line 61: I wonder if we should always disallow this, whether disable_escaping is set or not. I believe that the template can't be properly interpolated if escaping is enabled due to curly brackets, but the error could surface as a potential typo to the user vs accepting that as the input. I'm fine with leaving this as-is too though.

calvn (approved):
One small note, otherwise looks good!

sdk/database/helper/connutil/sql.go
robmonte: I think I agree with you. My only reservation would be causing a problem in the very-unlikely-but-not-impossible case someone actually configured their database already with these values in the username or password. @austingebauer When you get a chance to review as well, do you have an opinion on this matter?

austingebauer: Hmm, I see why we'd want to disallow this in general. I also wonder if that behavior change would break some users. Seems unlikely but possible as you've mentioned. I'm okay leaving this as-is.

austingebauer (approved):
LGTM!

sdk/database/helper/connutil/sql.go
robmonte: Decided to go with always disallowing it, changed in commit https://github.com/hashicorp/vault/pull/13414/commits/927d4600bba9e8ff2bdeb44586f1fed26d8b4c98

maxb:
What problem is this code trying to solve?

```
	// Do not allow the username or password template pattern to be used as
	// part of the user-supplied username or password
	if strings.Contains(c.Username, "{{username}}") ||
		strings.Contains(c.Username, "{{password}}") ||
		strings.Contains(c.Password, "{{username}}") ||
		strings.Contains(c.Password, "{{password}}") {

		return nil, fmt.Errorf("username and/or password cannot contain the template variables")
	}
```

I just came across it in the code, and it is unclear to me why this logic was added.

robmonte:
> What problem is this code trying to solve?
> 
> ```
> 	// Do not allow the username or password template pattern to be used as
> 	// part of the user-supplied username or password
> 	if strings.Contains(c.Username, "{{username}}") ||
> 		strings.Contains(c.Username, "{{password}}") ||
> 		strings.Contains(c.Password, "{{username}}") ||
> 		strings.Contains(c.Password, "{{password}}") {
> 
> 		return nil, fmt.Errorf("username and/or password cannot contain the template variables")
> 	}
> ```
> 
> I just came across it in the code, and it is unclear to me why this logic was added.

Hi Max, when working on this feature I kept getting infrequent test failures in between successful runs, like a race condition. At the time, I had narrowed it down to the ConnectionURL sometimes being a different value despite no changes. 

When building the ConnectionURL, it calls the function `QueryHelper` which then calls `strings.ReplaceAll`. It was at this point that it seemed to randomly fail replacing the the [old](https://cs.opensource.google/go/go/+/refs/tags/go1.20.7:src/strings/strings.go;l=1049) value with a new value containing the old value itself.

At the time I couldn't figure out why this was happening. I started going down this rabbit hole again because of your question here, and I think I figured out the issue this time.

To quickly test this, I first commented out the block of code restricting the use of the template values. Next, I used the `TestSQLDisableEscaping` test case in sql_test.go. I commented out all of the table tests and used `{"{{password}}mssql{0}", "password{0}", true}` as the only test value. Here is the git patch file if you'd like to run the same test: 
[gh.patch](https://github.com/hashicorp/vault/files/12264161/gh.patch)

I was overlooking the fact that this `QueryHelper` function was calling `strings.ReplaceAll` in a loop from a map. Since map order isn't guaranteed, it changes the result.

Replacing the username template first (the common case):
1. Starting value: `server=localhost;port=1433;user id={{username}};password={{password}};database=mydb;`
2. Replace username: `server=localhost;port=1433;user id={{password}}mssql{0};password={{password}};database=mydb;`
3. Replace password: `server=localhost;port=1433;user id=password{0}mssql{0};password=password{0};database=mydb;`

...but sometimes, it does the password template first:
1. Starting value: `server=localhost;port=1433;user id={{username}};password={{password}};database=mydb;`
2. Replace password: `server=localhost;port=1433;user id={{username}};password=password{0};database=mydb;`
3. Replace username: `server=localhost;port=1433;user id={{password}}mssql{0};password=password{0};database=mydb;`

The opposite case would also happen from using {{username}} in the password value.

At the time, the quickest solution was to simply disallow the use of the template variables as part of the username or password themselves. I also don't see any good coming from allowing the use of them. It just seems like it's asking for trouble.

maxb:
Aha! Thanks for the elucidation! I had observed that all template expressions were being replaced in one call to `dbutil.QueryHelper`, but hadn't looked inside it to see that it is performing a separate iteration over the string for each one, rendering it liable to the kinds of effects you saw.

Disallowing the specific template expressions as replacements is a reasonable workaroud - I dislike it because of the potential of surprising someone at some point - it wouldn't be that crazy for `{{password}}` to appear in a test password - but I admit it is pretty unlikely to happen in practice.

maxb:
> I wonder if we should add a comment with a brief reason for this behavior?

I propose instead we fix the underlying templating routine so the workaround is no longer required: #22224 

> Specifically might be worth mentioning that it is only required to prevent race conditions in the tests.

It would be possible to run into this problem in production too if usernames or passwords contained any sequences like `{{name}}`, `{{username}}`, `{{password}}`, `{{expiration}}`, as whilst this workaround guarded one particular use of `dbutil.QueryHelper`, there are others.

PR #13419, opened 2021-12-14
Add ManagedKeyRegistry to keep track of keys managed by a KMS.

victorr:


victorr:
Closing the PR without merging, as it needs to be re-worked.

PR #13425, opened 2021-12-14
crt main fix for ecr tag

kpenfound:


mladlow (approved):
I confirmed that public.ecr.aws is a thing that redirected me somewhere, so I'm ok merging this, but I also have no concrete knowledge backing up this approval :)

PR #13428, opened 2021-12-14
Fix for `main`: Add vault revision to --version cmd

mdeggies:
The output of `./vault --version` changed since Vault onboarded to CRT/GitHub Actions for building artifacts.

```
elle: ~/Users/elle/Downloads $ ./vault version
Vault v1.9.1
elle: ~/Users/elle/Downloads $ ./vault version
Vault v1.8.1 (4b0264f28defc05454c31277cfa6ff63695a458d)
```

This change adds back in the revision string to the version command. Output from the artifacts built on this branch:

```
elle: ~/Users/elle/Downloads $ ./vault --version
Vault v1.10.0-dev (9e2f21337b47d8ef59f057b6d192d6fde2c5fb19)
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,7 @@ jobs:
         run: |
           make version
           echo "::set-output name=product-version::$(make version)"
+
   generate-metadata-file:
     needs: get-product-version
     runs-on: ubuntu-latest
@@ -88,7 +89,7 @@ jobs:
           CGO_ENABLED: 0
         run: |
           mkdir dist out
-          VAULT_VERSION=${{ needs.get-product-version.outputs.product-version }} VAULT_COMMIT=${GITHUB_SHA} GO_TAGS="${{ env.GO_TAGS }}" make build
+          GO_TAGS="${{ env.GO_TAGS }}" VAULT_VERSION=${{ needs.get-product-version.outputs.product-version }} VAULT_REVISION="$(git rev-parse HEAD)" make build
           zip -r -j out/${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip dist/
       - uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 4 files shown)


kpenfound (approved):
Makes sense!

mdeggies:
Thanks for the review! Duplicating this PR to target the various release branches.

.github/workflows/build.yml
```diff
@@ -88,7 +89,7 @@ jobs:
           CGO_ENABLED: 0
         run: |
           mkdir dist out
-          VAULT_VERSION=${{ needs.get-product-version.outputs.product-version }} VAULT_COMMIT=${GITHUB_SHA} GO_TAGS="${{ env.GO_TAGS }}" make build
+          GO_TAGS="${{ env.GO_TAGS }}" VAULT_VERSION=${{ needs.get-product-version.outputs.product-version }} VAULT_REVISION="$(git rev-parse HEAD)" make build
```
mladlow, line 92: Is `rev-parse HEAD` definitely right? This will not get the head of the release branch, but will get the specific SHA tied to this commit?

Makefile
```diff
@@ -259,7 +259,7 @@ ci-verify:
 # This is used for release builds by .github/workflows/build.yml
 build:
 	@echo "--> Building Vault $(VAULT_VERSION)"
-	@go build -v -tags "$(GO_TAGS)" -ldflags " -X $VERSION_PKG_PATH.Version=$(VAULT_VERSION) -X $VERSION_PKG_PATH.GitCommit=$(VAULT_COMMIT)" -o dist/
+	@go build -v -tags "$(GO_TAGS)" -ldflags " -X github.com/hashicorp/vault/sdk/version.Version=$(VAULT_VERSION) -X github.com/hashicorp/vault/sdk/version.GitCommit=$(VAULT_REVISION)" -o dist/
```
mladlow, line 262: It's fairly likely I don't understand what's going on here, but based on `version.go` should this be?
```suggestion
	@go build -v -tags "$(GO_TAGS)" -ldflags " -X github.com/hashicorp/vault/sdk/version.Version=$(VAULT_VERSION) -X github.com/hashicorp/vault/sdk/version.Revision=$(VAULT_REVISION)" -o dist/
```

mladlow (commented):
Is there a way to test the makefile before committing the changes? Probably you've already done that?

Makefile
mdeggies: I followed consul's example here: https://github.com/hashicorp/consul/blob/main/command/version/version.go, https://github.com/hashicorp/consul/blob/main/command/version/version.go. I _think_ either would work, but I haven't tested the alternative

.github/workflows/build.yml
mdeggies: Yep, that's right! This is the full commit from the branch that is building the artifacts. In the test example below: 

```
elle: ~/Users/elle/Downloads $ ./vault --version
Vault v1.10.0-dev (9e2f21337b47d8ef59f057b6d192d6fde2c5fb19)
```

[this commit](https://github.com/hashicorp/vault/commit/9e2f21337b47d8ef59f057b6d192d6fde2c5fb19` shows my changes that triggered the builds. When this is merged into a release branch, the artifacts built at that time will show the SHA from the merge commit. At release time, the artifacts will show the SHA from the final commit to the release branch. I think that's what we want

mladlow (approved):

.github/workflows/build.yml
mdeggies: This does that, yes!

mdeggies: Today (before the staging releases tmr), I'll confirm the SHA's with you

PR #13430, opened 2021-12-14
Styling and legend component

Monkeychip:


ui/app/components/clients/total-client-usage.js
```diff
@@ -72,7 +72,7 @@ export default class TotalClientUsage extends Component {
     chartSvg.attr('viewBox', `0 5 725 305`); // set aspect ratio
 
     let groups = chartSvg
-      .selectAll('g')
+      .selectAll('g.rect')
```
hellobontempo, line 75: we should probably remove this - I don't think it made a difference and might mess with stuff later

ui/app/components/clients/total-client-usage.js
```diff
@@ -110,6 +110,7 @@ export default class TotalClientUsage extends Component {
         .append('g')
         .attr('transform', `translate(${TRANSLATE.right})`)
         .attr('class', 'axes-lines')
+      // .style('mix-blend-mode', 'darken')
```
hellobontempo, line 113: can delete since we moved this to the CSS file

ui/app/styles/core/charts.scss
```diff
@@ -74,13 +74,6 @@
   }
 }
 
-.chart {
```
hellobontempo, line 77: Just realize I'm going to need to keep this for my chart. But `.axes-lines` CSS can be used for the line chart, too

PR #13439, opened 2021-12-15
update okta-sdk-golang to v2.9.1

ccapurso:
It is possible for Vault to get into a deadlock with the current version of the Okta SDK based on how it handles backoff for 429 responses. The SDK trusts the backoff time from the Okta API without any means of short-circuiting with a maximum backoff limit.

We have witnessed long lived Okta login requests resulting in the k8s service registry incorrectly labeling two nodes as `vault-active=true`. In this scenario, Okta login request goroutines were sleeping in a function called `backoffPause`. The following is the call chain from Vault’s Okta backend:

- `Login` - [builtin/credential/okta/backend.go:291](https://github.com/hashicorp/vault/blob/release/1.7.x/builtin/credential/okta/backend.go#L291)
- `getOktaGroups` - [builtin/credential/okta/backend.go:340](https://github.com/hashicorp/vault/blob/release/1.7.x/builtin/credential/okta/backend.go#L340)
- `ListUserGroups` -  [okta-sdk-golang/v2/okta/user.go:519](https://github.com/hashicorp/vault/blob/ff185c0adc6301337f0a553187c5dfd0bfd61654/vendor/github.com/okta/okta-sdk-golang/v2/okta/user.go#L519)
- `Do` - [okta-sdk-golang/v2/okta/requestExecutor.go:237](https://github.com/hashicorp/vault/blob/ff185c0adc6301337f0a553187c5dfd0bfd61654/vendor/github.com/okta/okta-sdk-golang/v2/okta/requestExecutor.go#L237)
- `doWithRetries` - [okta-sdk-golang/v2/okta/requestExecutor.go:284](https://github.com/hashicorp/vault/blob/ff185c0adc6301337f0a553187c5dfd0bfd61654/vendor/github.com/okta/okta-sdk-golang/v2/okta/requestExecutor.go#L284)
- `backoffPause` - [okta-sdk-golang/v2/okta/requestExecutor.go:314](https://github.com/hashicorp/vault/blob/ff185c0adc6301337f0a553187c5dfd0bfd61654/vendor/github.com/okta/okta-sdk-golang/v2/okta/requestExecutor.go#L314)

In order to enter `backoffPause`, Vault needed to receive a `429 Too Many Requests` response from the Okta API. The implementation of the `backoffPause` function in version 2.0.0 of the SDK trusts the `X-Rate-Limit-Reset` header value from the Okta API which is used to calculate the sleep time. My theory is that the SDK calculated an extremely large backoff time given that the default max retries is set to 2. The backoff logic has since changed in newer versions of the SDK by introducing a config parameter `Okta.Client.RateLimit.MaxBackoff` which defaults to 30 seconds when using `NewClient` (see [builtin/credential/okta/path_config.go:326](https://github.com/hashicorp/vault/blob/ff185c0adc6301337f0a553187c5dfd0bfd61654/builtin/credential/okta/path_config.go#L326)). The important change is that if the SDK calculates a backoff time using the response headers that exceeds the one configured on the Okta client, the SDK will default to using the one configured on the client, see [okta-sdk-golang/requestExecutor.go at v2.9.1](https://github.com/okta/okta-sdk-golang/blob/v2.9.1/okta/requestExecutor.go#L342-L343).

This PR updates `okta-sdk-golang` to v2.9.1 in order to consume refactored request handling logic which introduces this maximum backoff limit. In addition to the existing unit tests, I have also manually tested the Okta auth method with MFA to ensure that there have not been any regressions.


austingebauer (approved):
LGTM!

PR #13441, opened 2021-12-15
website: Update Nomad-related text

EnMod:
This PR updates some text in the Vault docs related to Nomad.


mwickett (approved):
Let's hold off on merging this until the AWS outage is cleared up. The backporter is hung up because of it.

EnMod:
Seems the AWS outage is [cleared up](https://status.aws.amazon.com/) :+1:

PR #13443, opened 2021-12-15
Ember Upgrade to 3.24

zofskeez:
The primary goal here was to upgrade Ember to the latest LTS version (3.24) and as many of the other dependencies as necessary to get the app to build. Here is a brief overview of the changes and things to be aware of.

### Linting
Some new lint rules were added which led to cleaning up all of the new and existing `js` and `hbs` lint errors. Since some rules are Octane focused, individual rules have either been globally ignored until we can fully glimmerize everything or ignored/configured to warn on a per file basis. The `.eslintrc.js` and `.template-lintrc.js` contain all the rule overrides. We should work towards fully adopting the recommended lint rules over time.

While template linting was available in the project prior to the upgrade it was not being utilized. Part of the issue (for me at least) was that the [Ember Language Server VSCode extension](https://marketplace.visualstudio.com/items?itemName=EmberTooling.vscode-ember) was not working with ember-template-lint. It may no longer be maintained since there hasn't been an update since May 2020 and perhaps that is why it is not working in the latest VSCode with the newest version of ember-template-lint.

It is highly recommended to use the [Unstable Ember Language Server extension](https://marketplace.visualstudio.com/items?itemName=lifeart.vscode-ember-unstable) which is a fork that is actively maintained and works with ember-template-lint in the Vault project. Since some folks may not prefer live linting in the editor I added the `lint:hbs:quiet` task to the test run. Both `js` and `hbs` linting must find no errors before the Ember tests will run. Note that the lint tasks that contain `:quiet` will not fail if they contain warnings. We should strive to resolve all warnings over time and eventually flip them back to error.  

### Deprecations
[ember-cli-deprecation-workflow](https://github.com/mixonic/ember-cli-deprecation-workflow) was enabled to handle deprecation warnings. It's not always trivial to refactor to get around deprecations and they may also be reported from addons which may involve updating, or in the case of no longer maintained addons there may need to be a plan to migrate away from usage. With that in mind I created an initializer to filter out deprecation warnings that will remain until version 4.0.0. This gives us time to address any immediate issues and plan a path forward before our next major upgrade.

### Test Failures
After touching many files fixing lint errors it was expected that there would be failing tests. I would like to briefly highlight some async issues that reoccur throughout the acceptance tests. It seems that after the upgrade the issue has increased and many acceptance tests that wait on an async action before asserting that a route transition has taken place fail intermittently.
```js
await click('[data-test-save]');
assert.equal(currentURL(), '/some/route', 'route changes after save');
```
There is a new lint rule that errors when placing `settled();` after a `click` citing that the click handler uses settled internally. To get around this for the time being and attempt to fix the flakiness I refactored any failures caught by CircleCI to use `waitUntil`:
```js
await click('[data-test-save']);
assert.ok(
  await waitUntil(() => currentURL() === '/some/route'),
  'route changes after save'
);
```
We should attempt to address this issue globally and try and get to a point where we can reduce the usage of `settled` and `waitUntil` which will both clean up the readability of the tests and reduce intermittent failures.


ui/app/adapters/application.js
```diff
@@ -41,15 +41,15 @@ export default RESTAdapter.extend({
       headers['X-Vault-Wrap-TTL'] = options.wrapTTL;
     }
     let namespace = typeof options.namespace === 'undefined' ? this.namespaceService.path : options.namespace;
-    if (namespace && !NAMESPACE_ROOT_URLS.some(str => url.includes(str))) {
+    if (namespace && !NAMESPACE_ROOT_URLS.some((str) => url.includes(str))) {
```
Monkeychip, line 44: assuming this is a new linting rule to use parens in a single line arrow function?

ui/app/adapters/pki-config.js
```diff
@@ -69,22 +69,24 @@ export default ApplicationAdapter.extend({
     return hash({
       backend: backendPath,
       id: this.id(backendPath),
-      der: this.rawRequest(derURL, 'GET', { unauthenticated: true }).then(response => response.blob()),
-      pem: this.rawRequest(pemURL, 'GET', { unauthenticated: true }).then(response => response.text()),
-      ca_chain: this.rawRequest(chainURL, 'GET', { unauthenticated: true }).then(response => response.text()),
+      der: this.rawRequest(derURL, 'GET', { unauthenticated: true }).then((response) => response.blob()),
+      pem: this.rawRequest(pemURL, 'GET', { unauthenticated: true }).then((response) => response.text()),
+      ca_chain: this.rawRequest(chainURL, 'GET', { unauthenticated: true }).then((response) =>
+        response.text()
```
Monkeychip, line 75: just curious but did your editor force this to be on one line because of length?

ui/app/adapters/secret-engine.js
```diff
@@ -135,11 +135,7 @@ export default ApplicationAdapter.extend({
 
   saveZeroAddressConfig(store, type, snapshot) {
     const path = encodePath(snapshot.id);
-    const roles = store
-      .peekAll('role-ssh')
-      .filterBy('zeroAddress')
-      .mapBy('id')
-      .join(',');
+    const roles = store.peekAll('role-ssh').filterBy('zeroAddress').mapBy('id').join(',');
```
Monkeychip, line 138: again, is this a linting thing? What's the rule change here?

ui/app/components/b64-toggle.js
```diff
@@ -99,6 +99,7 @@ export default Component.extend({
   },
 
   didReceiveAttrs() {
+    this._super();
```
Monkeychip, line 102: is super required for hooks?

ui/app/adapters/secret-engine.js
zofskeez: I checked and the `printWidth` did not change between versions so I'm not sure why this changed. I'm going to double check that the `prettier-eslint` cli is using the project config.

ui/app/components/b64-toggle.js
zofskeez: Yes it should be called. That looks like it was auto fixed by `eslint`. I usually do `this._super(...arguments)` but for this hook there may not be arguments passed in.

ui/app/components/tool-actions-form.js
```diff
@@ -4,11 +4,12 @@ import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import { setProperties, computed, set } from '@ember/object';
 import { addSeconds, parseISO } from 'date-fns';
+import { A } from '@ember/array';
 
 const DEFAULTS = {
   token: null,
   rewrap_token: null,
-  errors: [],
+  errors: A(),
```
Monkeychip, line 12: is this standard now for all default arrays?

ui/app/initializers/deprecation-filter.js
```diff
@@ -0,0 +1,15 @@
+import { registerDeprecationHandler } from '@ember/debug';
```
Monkeychip, line 1: nice!

ui/app/components/tool-actions-form.js
zofskeez: It is one approach when initializing a property to an array. Another is to use the `init` hook. Here's the [link to the eslint rule](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/avoid-leaking-state-in-ember-objects.md).

ui/app/templates/components/json-editor.hbs
```diff
@@ -20,17 +20,17 @@
...
 	{{/if}}
 
-	{{ivy-codemirror
-		data-test-component="json-editor" 
-		value=@value
-		options=this.options
-		valueUpdated=(action "updateValue")
-		onFocusOut=(action "onFocus")    
-	}}
+	<IvyCodemirror
```
Monkeychip: Not something for you to fix here, but the indentation looks off on this file. I think I've dealt with it too, and I'm not sure what's up with this particular file.

Monkeychip (approved):
Amazing work. Make sure to add changelog and milestone and labels.

ui/app/templates/components/json-editor.hbs
zofskeez: Weird it looks good in my editor. I think it would be a good idea to use the prettier plugin for ember-template-lint as well to catch these things. I'll bring it up in our discussion.
https://github.com/ember-template-lint/ember-template-lint-plugin-prettier

zofskeez: @Monkeychip there is a _stylistic_ rule set for ember-template-lint which I enabled and it found this identation issue among many others. I think it's a good idea to enforce it.

zofskeez:
@Monkeychip would you mind giving the recent changes a once over when you have a minute? Based on our conversation, the latest batch of commits address the following:
- added an hbs specific rule to prettier to use double quotes in templates
- updated the `lint-staged` task (which is triggered on pre commit) to format (via prettier) and lint `.js` and `.hbs` files. Prettier is also run on `.scss` files which remains unchanged.
- set the prettier cli config arg to use .prettierrc.js rather than passing in individual rule flags to align with editor extension formatting
- ran prettier again on `.hbs` files and manually fixed anything that couldn't be auto-fixed. There is an opportunity to leverage custom fixers in ember-template-lint in the future but it wasn't something I was ready to tackle last night
- fixed enterprise tests

zofskeez:
@Monkeychip I fixed the issue with the hanging acceptance test. I traced it back to the `ember-concurrency-test-waiters` addon which is deprecated and they suggested to use `@ember/test-waiters` so I refactored any instances of the old method to the new.

PR #13454, opened 2021-12-16
Update docs with new images

pmmukh:
- Adds new images created in https://app.asana.com/0/1179616438052993/1201489771509180 to the Identity concepts page


website/content/docs/concepts/identity.mdx
```diff
@@ -25,11 +25,19 @@ a particular authentication backend.
 
 For example, a user with accounts in both GitHub and LDAP can be mapped to a
 single entity in Vault with two aliases, one of type GitHub and one of type
-LDAP. Note however, if both aliases are created on the same auth mount, say
-a Github mount, both aliases cannot be mapped to the same entity.
+LDAP.
 
 ![Entity  overview](/img/vault-identity-doc-1.png)
 
+However, if both aliases are created on the same auth mount, say
```
taoism4504, line 32:
```suggestion
However, if both aliases are created on the same auth mount, such as
```

website/content/docs/concepts/identity.mdx
```diff
@@ -25,11 +25,19 @@ a particular authentication backend.
...
+a Github mount, both aliases cannot be mapped to the same entity. The aliases can
+have the same auth type, say Github, as long as the auth mounts are different, and
```
taoism4504, line 34: Not sure if you need to say Github here again since it's obvious we're using Github as an example. I'll let that up to you.

taoism4504 (approved):
one suggestion and one comment, but otherwise LGTM.

website/content/docs/concepts/identity.mdx
pmmukh: oh yeah fair point, removed 78921c4

PR #13463, opened 2021-12-16
Move KmsLibrary code to the ent side of config parsing

sgmiller:


command/server/config.go
```diff
@@ -569,16 +541,7 @@ func ParseConfig(d, source string) (*Config, error) {
...
-	if o := list.Filter("kms_library"); len(o.Items) > 0 {
-		delete(result.UnusedKeys, "kms_library")
-		if err := parseKmsLibraries(result, o); err != nil {
-			return nil, fmt.Errorf("error parsing 'kms_library': %w", err)
-		}
-	}
-
-	entConfig := &(result.entConfig)
-	if err := entConfig.parseConfig(list); err != nil {
+	if err := result.parseConfig(list); err != nil {
```
stevendpclark, line 544: I think you removed a little too much here should have kept the 

```
	entConfig := &(result.entConfig)
	if err := entConfig.parseConfig(list); err != nil {
```

stevendpclark (approved):

Issue #13340, opened 2021-12-03
Token roles don't respect 'allowed_entity_aliases' parameter


kauedg:
**Describe the bug**
When a token role is created using the `allowed_entity_aliases` parameter it is still possible to create a token using the role when no entity alias is passed.

**To Reproduce**
1) Create a dev server with the needed resources:
```shell
$ vault server -dev &
$ export VAULT_ADDR='http://127.0.0.1:8200'
$ vault login [root token]

# Create a KV with a secret
$ vault secrets enable -path=TEST -version=1 kv
$ vault write TEST/somesecret password=passw0rd

# Create a policy that allows reading the above secret and the role, restricting the aliases allowed to use it
$ echo '{ "path": [ { "TEST/*": { "capabilities": ["read", "list"] } } ] }' | vault policy write read_secret -
$ vault write auth/token/roles/read_test_role allowed_policies=read_secret allowed_entity_aliases=privileged_alias

# Enable userpass, create a user and a policy that allows it to create tokens using the role
$ vault auth enable userpass
$ echo '{ "path": [ { "auth/token/create/read_test_role": { "capabilities": ["update"] } } ] }' | vault policy write someuser_policy -
$ vault write auth/userpass/users/someuser password='12344321' token_policies=someuser_policy
```

2) Check that the user can't read the secret and can't create a token using his alias
```
$ vault login -method=userpass username=someuser password=12344321
$ vault read TEST/somesecret
Error reading TEST/somesecret: Error making API request.

URL: GET http://127.0.0.1:8200/v1/TEST/somesecret
Code: 403. Errors:

* 1 error occurred:
        * permission denied

$ export VAULT_TOKEN=$(vault token create -role=read_test_role -entity-alias=someuser -field token)
Error creating token: Error making API request.

URL: POST http://127.0.0.1:8200/v1/auth/token/create/read_test_role
Code: 400. Errors:

* invalid 'entity_alias' value

$ unset VAULT_TOKEN
```

3) But the user is able to get a token and read the secret, without passing the `entity_alias` parameter or with a valid alias:
```
$ export VAULT_TOKEN=$(vault token create -role=read_test_role -field token)
$ vault read TEST/somesecret
Key                 Value
---                 -----
refresh_interval    768h
password            passw0rd

$ unset VAULT_TOKEN
$ export VAULT_TOKEN=$(vault token create -role=read_test_role -entity-alias=privileged_alias -field token)
$ vault read TEST/somesecret
Key                 Value
[...]

PR #13474, opened 2021-12-20
Add error handling for empty entity_alias field

ryowright:
Fix for [issue #13340](https://github.com/hashicorp/vault/issues/13340).


HridoyRoy (changes requested):
Hi, thanks for submitting a PR to vault! 

As I understand it, the `entity_alias` field does not need to be set in a call to token create, so we do not want to throw an error in this case.Looking at the Github issue relating to this (https://github.com/hashicorp/vault/issues/13340), it seems like the request is to not allow an empty `entity_alias` when `allowed_entity_aliases` is specifically set. So, I think if we do want to go ahead with this feature, we should update the logic here: https://github.com/hashicorp/vault/blob/b45b9d8e92ef7e1d0a96f276daaa60b9bf617ae3/vault/token_store.go#L2403 . 

This request seems good to me from an architectural perspective, but as @ncabatoff pointed out, it does break backward compatibility so we'd like to understand the use-cases this would solve a little more before going ahead with this change. 
Thanks!

aphorise:
hey @ryowright - can you kindly provide a [response to the earlier request](https://github.com/hashicorp/vault/pull/13474#pullrequestreview-861307776).

I'm curious @HridoyRoy if this is likely to still be considered long term or subject to no response?

HridoyRoy:
Hi @ryowright , I'm closing this PR due to my original comment [here](https://github.com/hashicorp/vault/pull/13474#pullrequestreview-861307776). Please let me know if there's anything else you'd like to follow up on.
Thanks!

PR #13477, opened 2021-12-20
UI/ Client counts range (running total component)

hellobontempo:
In this PR:
- constructs line chart
- renames components/files
- adds tooltips to line plot and horizontal bar chart
- makes util folder for shared constants 

line-plot-tooltips

TO DO:
- [ ] fix styling for tooltips (different for each chart)


Monkeychip (approved):
Nice work!!!

PR #13487, opened 2021-12-21
[Vault-4628] OpenAPI endpoint not expanding root alternations

VinnyHC:
Improve OpenAPI endpoint `/sys/internal/specs/openapi` to generate proper specs for path patterns that use an alternation as the root.
Example:
```
(creds|sts)/(?P<name>\w(([\w-.@]+)?\w)?)
```
should generate two paths in OpenAPI
```
creds/{name}
sts/{name}
```

### DIFF
--- a/sdk/framework/openapi.go
+++ b/sdk/framework/openapi.go
@@ -191,16 +191,16 @@ var OASStdRespNoContent = &OASResponse{
 var optRe = regexp.MustCompile(`(?U)\([^(]*\)\?|\(/\(\?P<[^(]*\)\)\?`)
 
 var (
-	reqdRe           = regexp.MustCompile(`\(?\?P<(\w+)>[^)]*\)?`)   // Capture required parameters, e.g. "(?P<name>regex)"
-	altRe            = regexp.MustCompile(`\((.*)\|(.*)\)`)          // Capture alternation elements, e.g. "(raw/?$|raw/(?P<path>.+))"
-	pathFieldsRe     = regexp.MustCompile(`{(\w+)}`)                 // Capture OpenAPI-style named parameters, e.g. "lookup/{urltoken}",
-	cleanCharsRe     = regexp.MustCompile("[()^$?]")                 // Set of regex characters that will be stripped during cleaning
-	cleanSuffixRe    = regexp.MustCompile(`/\?\$?$`)                 // Path suffix patterns that will be stripped during cleaning
-	wsRe             = regexp.MustCompile(`\s+`)                     // Match whitespace, to be compressed during cleaning
-	altFieldsGroupRe = regexp.MustCompile(`\(\?P<\w+>\w+(\|\w+)+\)`) // Match named groups that limit options, e.g. "(?<foo>a|b|c)"
-	altFieldsRe      = regexp.MustCompile(`\w+(\|\w+)+`)             // Match an options set, e.g. "a|b|c"
-	nonWordRe        = regexp.MustCompile(`[^\w]+`)                  // Match a sequence of non-word characters
-	altRootsRe       = regexp.MustCompile(`^\([\w-_|]+\)`)           // Pattern starting with alts, e.g. "(root1|root2)/(?P<name>regex)"
+	reqdRe           = regexp.MustCompile(`\(?\?P<(\w+)>[^)]*\)?`)                // Capture required parameters, e.g. "(?P<name>regex)"
+	altRe            = regexp.MustCompile(`\((.*)\|(.*)\)`)                       // Capture alternation elements, e.g. "(raw/?$|raw/(?P<path>.+))"
+	pathFieldsRe     = regexp.MustCompile(`{(\w+)}`)                              // Capture OpenAPI-style named parameters, e.g. "lookup/{urltoken}",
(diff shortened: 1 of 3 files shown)


VinnyHC:
This was tested comparing the output from `scripts/gen_openapi.sh -p` when run against `main` and `vinnyhc/VAULT-4628-OpenAPI`. The only difference was the proper spec for the `aws/creds/{name}` entry.

sdk/framework/openapi.go
```diff
@@ -200,6 +200,7 @@ var (
 	altFieldsGroupRe = regexp.MustCompile(`\(\?P<\w+>\w+(\|\w+)+\)`) // Match named groups that limit options, e.g. "(?<foo>a|b|c)"
 	altFieldsRe      = regexp.MustCompile(`\w+(\|\w+)+`)             // Match an options set, e.g. "a|b|c"
 	nonWordRe        = regexp.MustCompile(`[^\w]+`)                  // Match a sequence of non-word characters
+	altRootsRe       = regexp.MustCompile(`\(((\w*)[\||\)])+\/`)     // Paths that start with alts, e.g. "(creds|sts)/(?P<name>regex)"
```
averche, line 203: It looks like this will only match on the prefix (e.g. "(creds|sts)/"). Should that be in the example comment?

I believe this part `[\||\)]` can be simplified to `[|\)]`

The regex will also match something like `(creds)sts)/` or `(||)|))/`  (not sure if that's an issue)

sdk/framework/openapi.go
```diff
@@ -464,6 +465,21 @@ func specialPathMatch(path string, specialPaths []string) bool {
 func expandPattern(pattern string) []string {
 	var paths []string
 
+	// recursively operate on each pattern after determining a set of unique roots is present,
+	// e.g. "(rootOne|rootTwo)/(?P<name>regex)" should yield "rootOne/{name}", "rootTwo/{name}"
```
averche, line 469: It may be a good idea to extract this logic into a separate function and add a simple unit test to verify it

sdk/framework/openapi.go
```diff
@@ -464,6 +465,21 @@ func specialPathMatch(path string, specialPaths []string) bool {
...
+	rootMatch := altRootsRe.FindAllString(pattern, -1)
+	if len(rootMatch) > 0 {
+		m := strings.Replace(rootMatch[0][1:], ")/", "", -1)
+		if m != "" {
+			var expandedRoots []string
+			suffix := strings.Replace(pattern, rootMatch[0], "", -1)
```
averche, line 475: There are a number of of [replace utilities](https://pkg.go.dev/regexp#Regexp.ReplaceAllString) for in regexp package, not sure if they would be useful here.

VinnyHC: Since `strings` was already used in the package I thought it was safe here. Are you worried about performance?

sdk/framework/openapi.go
VinnyHC: Since this only operates on `Pattern` s that are used in `Path`s I'm assuming that only valid regex patterns would land here. Your concerns are definitely valid though, I'll add some more tests and make sure this doesn't introduce something weird.

sdk/framework/openapi.go
averche: I was more wondering if the code could be simplified a bit into something easy to verify at a glance. One thing that could help add clarity here is capturing each root expression (`rootOne`, `rootTwo`, etc.) into its own subgroup and then iterating over the matches to form expanded roots. This would remove the need to split the string manually as well as the `string.Replace` calls.

sdk/framework/openapi.go
```diff
@@ -191,15 +191,16 @@ var OASStdRespNoContent = &OASResponse{
...
+	pathFieldsRe     = regexp.MustCompile(`{(\w+)}`)                              // Capture OpenAPI-style named parameters, e.g. "lookup/{urltoken}",
+	cleanCharsRe     = regexp.MustCompile("[()^$?]")                              // Set of regex characters that will be stripped during cleaning
+	cleanSuffixRe    = regexp.MustCompile(`/\?\$?$`)                              // Path suffix patterns that will be stripped during cleaning
+	wsRe             = regexp.MustCompile(`\s+`)                                  // Match whitespace, to be compressed during cleaning
+	altFieldsGroupRe = regexp.MustCompile(`\(\?P<\w+>\w+(\|\w+)+\)`)              // Match named groups that limit options, e.g. "(?<foo>a|b|c)"
+	altFieldsRe      = regexp.MustCompile(`\w+(\|\w+)+`)                          // Match an options set, e.g. "a|b|c"
+	nonWordRe        = regexp.MustCompile(`[^\w]+`)                               // Match a sequence of non-word characters
+	altRootsRe       = regexp.MustCompile(`^\(([\w\-_]+(?:\|[\w\-_]+)+)\)(/.*)$`) // Pattern starting with alts, e.g. "(root1|root2)/(?P<name>regex)"
```
averche, line 203: Maybe move this one up to be next to the other `alt*` fields, alternatively sorting these alphabetically would be nice too 😄

averche (approved):
LGTM 👍

averche (approved):

PR #13484, opened 2021-12-21
Backport of add tencentCloud  Secrets&Auth Plugins into stable-website

hc-github-team-secure-vault-core:
## Backport

This PR is auto-generated from #13415 to be assessed for backporting due to the inclusion of the label backport/website.

The below text is copied from the body of the original PR.

---

_Original PR had no description content._

PR #13415, opened 2021-12-14
add tencentCloud  Secrets&Auth Plugins

firingLi:

PR #13488, opened 2021-12-21
fix(backport-docs): always run 2nd step

thiskevinwang:
# Description

On merge of PR's with the `backport/website`, 2 workflow steps should run.

This updates the workflow so that if the 1st (backport+automerge to `stable-website`) fails, the 2nd (backport PR to `release/<latest>`) still runs

### Notes

- an original PR https://github.com/hashicorp/vault/pull/13415 w/ `backport/website`
  - workflow with `405` https://github.com/hashicorp/vault/actions/runs/1604682279
  - Spawned backport PR https://github.com/hashicorp/vault/pull/13484

### DIFF
--- a/.github/workflows/backport-docs.yml
+++ b/.github/workflows/backport-docs.yml
@@ -21,6 +21,7 @@ jobs:
           BACKPORT_TARGET_TEMPLATE: "stable-{{.target}}"
           GITHUB_TOKEN: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
       - name: Backport changes labeled website to latest release branch
+        if: always()
         run: |
           resp=$(curl -f -s "https://api.github.com/repos/$GITHUB_REPOSITORY/releases?per_page=100")
           ret="$?"
@@ -37,7 +38,7 @@ jobs:
           target="${latest_version#v}"
           # replace patch version with "x"
           target="${target%.*}.x"
-          
+
           export BACKPORT_TARGET_TEMPLATE="release/$target"
           backport-assistant backport
         env:


BRKalow (approved):
Makes sense to me

.github/workflows/backport-docs.yml
```diff
@@ -21,6 +21,7 @@ jobs:
           BACKPORT_TARGET_TEMPLATE: "stable-{{.target}}"
           GITHUB_TOKEN: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
       - name: Backport changes labeled website to latest release branch
+        if: always()
```
mladlow, line 24: In the docs (https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-not-requiring-dependent-jobs-to-be-successful) this is surrounded with `{{` - do you happen to know whether those are necessary and if not why?

mladlow (approved):
I have that one question but thank you for opening this!

.github/workflows/backport-docs.yml
thiskevinwang: @mladlow I had to google this... Found the answer [here](https://github.community/t/multiple-if-conditional-with-always-is-always-true/128840/3), which ultimately pointed right back to the docs 😂 
- https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsif

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression. For more information, see "[Expressions](https://docs.github.com/en/actions/learn-github-actions/expressions)."

mladlow: Thank you!

PR #13490, opened 2021-12-21
Fix some go vet issue

harsimranmaan:
There are more issues involving trivial composites and the
non trivial copylocks. The copylocks is interesting as inner mutexes
are sometime passed as value. One of fixing those is to use ref on the
to pass the enclosing struct around but that might need to subtle bugs like
nil pointer exception as the zero vault for these types would need to be
handled gracefully.

Run 'make vet' for  the results.

The vet command should be added to the CI when all issues are resolved
go vet -copylocks=false -composites=false ./... passes after this change

### DIFF
--- a/command/agent/cache/cache_test.go
+++ b/command/agent/cache/cache_test.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/go-test/deep"
 	hclog "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-multierror"
 	kv "github.com/hashicorp/vault-plugin-secrets-kv"
 	"github.com/hashicorp/vault/api"
 	"github.com/hashicorp/vault/builtin/credential/userpass"
@@ -359,6 +360,7 @@ func TestCache_ConcurrentRequests(t *testing.T) {
 	}
 
 	wg := &sync.WaitGroup{}
+	var asyncErr *multierror.Error
 	for i := 0; i < 100; i++ {
 		wg.Add(1)
 		go func(i int) {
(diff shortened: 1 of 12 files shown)


physical/postgresql/postgresql_test.go
```diff
@@ -180,13 +180,6 @@ const maxTries = 3
 func testPostgresSQLLockTTL(t *testing.T, ha physical.HABackend) {
 	t.Log("Skipping testPostgresSQLLockTTL portion of test.")
 	return
-
-	for tries := 1; tries <= maxTries; tries++ {
```
harsimranmaan, line 184: Dead code

command/agent/cache/cache_test.go
```diff
@@ -368,19 +370,24 @@ func TestCache_ConcurrentRequests(t *testing.T) {
 				"key": key,
 			})
 			if err != nil {
-				t.Fatal(err)
```
harsimranmaan, line 371: t.Fatal called in  non-test goroutine

plugins/database/mongodb/mongodb_test.go
```diff
@@ -406,7 +406,8 @@ func assertDeepEqual(t *testing.T, a, b *options.ClientOptions) {
 func createDBUser(t testing.TB, connURL, db, username, password string) {
 	t.Helper()
 
-	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cf := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cf()
```
harsimranmaan, line 410: Missing call to cancelfunc. Potential context leak

Issue #13498, opened 2021-12-22
Ensure non-zero and unique co-efficients for generating polynomials


supriya-premkumar:
[Recommendations for ensuring non-zero and unique co-efficients for generating polynomials](https://www.zkdocs.com/docs/zkdocs/protocol-primitives/shamir/#what-can-go-wrong)

**Zero Share Problem**
```
Any co-efficient if it is set to zero will compromise the secret because the shard holder will not have to evaluate the polynomial (0, S) which reveals the secret to the ith share holder.
```

**Non Unique Shares**
```
This happens when we interpolate lagrangian. In case if the two coefficients are the same their modulo inverse will be 0 and will crash 
```

cipherboy:
Hi @supriya-premkumar -- please see my response on the pull request. Thank you!

hsimon-hashicorp:
Hi @supriya-premkumar - wanted to chime in for the New Year and see if you had any questions or comments for Alex. Thanks!

supriya-premkumar:
Hi @hsimon-hashicorp, Yes. Just responded to @cipherboy.
Happy New Year to you both!

PR #13500, opened 2021-12-22
Ensure coefficients are non zero and unique for makePolynomial()

supriya-premkumar:
This mitigates secret leak to the i-th share holder and prevents evaluating (0, S)
Having unique coefficients will also ensure that we don't run into calculating lagrangian
interpolation for the modular inverse of zero i.e Xi/(Xi - Xj) as it doesn't exist.

Issue: https://github.com/hashicorp/vault/issues/13498#issue-1086336740

Ran the tests in a loop for 100 times.
```
$ time go test -timeout 1h -count 100  github.com/hashicorp/vault/shamir
ok  	github.com/hashicorp/vault/shamir	640.335s

real	10m40.641s
user	0m0.422s
sys	0m0.463s
```

### DIFF
--- a/shamir/shamir.go
+++ b/shamir/shamir.go
@@ -3,6 +3,7 @@ package shamir
 import (
 	"crypto/rand"
 	"crypto/subtle"
+	"errors"
 	"fmt"
 	mathrand "math/rand"
 	"time"
@@ -15,6 +16,12 @@ const (
 	ShareOverhead = 1
 )
 
+// ErrZeroShare is returned when rand.Read returns 0
+var ErrZeroShare = errors.New("coefficients are zero")
+
+// ErrNonUniqueShare is returned when there are duplicate coefficients
+var ErrNonUniqueShare = errors.New("coefficients are not unique")
+
 // polynomial represents a polynomial of arbitrary degree
 type polynomial struct {
 	coefficients []uint8
@@ -23,6 +30,9 @@ type polynomial struct {
 // makePolynomial constructs a random polynomial of the given
 // degree but with the provided intercept value.
 func makePolynomial(intercept, degree uint8) (polynomial, error) {
+	// Choosing the degree and not degree+1 since we need to enforce uniqueness only to the coefficients and not
+	// the intercept
+	coefficientSet := make(map[byte]bool, degree)
 	// Create a wrapper
 	p := polynomial{
 		coefficients: make([]byte, degree+1),
@@ -34,6 +44,16 @@ func makePolynomial(intercept, degree uint8) (polynomial, error) {
 	// Assign random co-efficients to the polynomial
 	if _, err := rand.Read(p.coefficients[1:]); err != nil {
 		return p, err
+	}
+	// Safety check to ensure coefficients are non zero and unique
+	for i := 1; i < len(p.coefficients); i++ {
+		if p.coefficients[i] == 0 {
+			return p, ErrZeroShare
+		}
+		if ok := coefficientSet[p.coefficients[i]]; ok {
+			return p, ErrNonUniqueShare
+		}
+		coefficientSet[p.coefficients[i]] = true
 	}
 
 	return p, nil
@@ -162,7 +182,7 @@ func Split(secret []byte, parts, threshold int) ([][]byte, error) {
 	for idx, val := range secret {
 		p, err := makePolynomial(val, uint8(threshold-1))
 		if err != nil {
-			return nil, fmt.Errorf("failed to generate polynomial: %w", err)
+			return nil, err
 		}
 
 		// Generate a `parts` number of (x,y) pairs
--- a/shamir/shamir_test.go
+++ b/shamir/shamir_test.go
@@ -2,29 +2,34 @@ package shamir
 
 import (
 	"bytes"
+	"fmt"
 	"testing"
+	"time"
 )
+
+const MaxAttempts = 10
+const MaxWaitInBetweenRetries = time.Second * 2
 
 func TestSplit_invalid(t *testing.T) {
 	secret := []byte("test")
 
-	if _, err := Split(secret, 0, 0); err == nil {
+	if _, err := retrySplit(MaxAttempts, MaxWaitInBetweenRetries, secret, 0, 0); err == nil {
 		t.Fatalf("expect error")
 	}
 
-	if _, err := Split(secret, 2, 3); err == nil {
+	if _, err := retrySplit(MaxAttempts, MaxWaitInBetweenRetries, secret, 2, 3); err == nil {
 		t.Fatalf("expect error")
 	}
 
-	if _, err := Split(secret, 1000, 3); err == nil {
+	if _, err := retrySplit(MaxAttempts, MaxWaitInBetweenRetries, secret, 1000, 3); err == nil {
 		t.Fatalf("expect error")
 	}
 
-	if _, err := Split(secret, 10, 1); err == nil {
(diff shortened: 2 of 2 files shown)


cipherboy:
Greetings @supriya-premkumar! Thank you for the pull request.

My understanding is that this is inspired by the [Trail of Bits](https://blog.trailofbits.com/2021/12/21/disclosing-shamirs-secret-sharing-vulnerabilities-and-announcing-zkdocs/) security audit of Verified Secret Sharing algorithms (from [Feldman’s paper](https://www.cs.umd.edu/~gasarch/TOPICS/secretsharing/feldmanVSS.pdf)). 

This implementation in Vault isn’t vulnerable to either of the two vulnerabilities in the TSS implementation for the following reasons:

1. We are not vulnerable to zero->hero due to careful selection of X values. From [`Split`](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L122-L127), we can see that [X coordinates](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L147) are generated in the range `[0, 254]` ([inclusive](https://pkg.go.dev/math/rand#Perm)). These are then [incremented by one](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L155) to then become in the range `[1, 255]` (inclusive). This ensures we never leak the secret directly, due to a zero-value coordinate.
2. We are additionally not vulnerable to zero->crash. This is a little more subtle: due to [earlier mentioned permutation](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L147), we ensure X coordinates are unique at creation time. When combining, we [validate each share has unique X coordinate](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L208-L218) and err if any duplicate value is detected. This ensures we never [subtract from the same identifier](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L71-L72) and [thus avoid the panic in div](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L83-L87). 

Note that our implementation uses randomized X values (not randomized Y values like [some others](https://dsprenkels.github.io/sss/#comparison-of-secret-sharing-libraries)) for evaluation. Uniqueness of coefficients is not required to prevent any divide-by-zero errors, and coefficients on the Shamir’s polynomial aren’t actually used in the denominator. Note that coefficients are effectively secret/privileged information, and while used for evaluation and generation of the shares, aren't known during combining until after polynomial interpolation is complete.

Additionally, as we do allow up to 255 parts, using the suggestion in this PR would result in `255 choose <shares>` possible coefficients (which is less than the theoretical `256^<shares>`) possible under this scheme. In practice, this perhaps does not matter much, but using unique permutations for both X and coefficients seems excessive and not immediately obvious what security benefit the latter would provide over uniform random selection. Each byte of secret receives a unique polynomial; as long as the crash and zero-evaluation is avoided, the odds of fully compromising a secret become astronomically small (given the length of said secret; an AES key).

Thus, I’m not inclined to merge this pull request, though I definitely thank you for the contribution!

Let me know if you disagree on anything and we can discuss more. :-)

supriya-premkumar:
Thank you for the response @cipherboy. Appreciate the detailed explanation.

This makes sense and  I agree that the permutation and incrementing by 1 avoids zero share  and non unique share problem for X coefficients.

However the problem still persits when our y coordinates are zero. 
Cases where this can happen.

1. [out == 0](https://github.com/hashicorp/vault/blob/3688d63348b319e0fb6f9bc0411d2500970fc905/shamir/shamir.go#L51) when coeffiecients[degree] == 0 a.k.a the last element
2. any of the [coeffiecients[i] == 0](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L53) In this case we will have our y coefficient to be zero as well.

 Imo this PR ensures our y coefficients are evaulated over a non zero random polynomial.

Let me know your thoughts. Thanks again!

cipherboy:
Hello @supriya-premkumar,

Sorry for the delay, was on break. :-)

> However the problem still persits when our y coordinates are zero.

Sorry, could you rephrase what problem persists when our y-coordinates are zero?

Operating under the assumption it is a similar risk as the original Trail Of Bits post...

After validating our shares have unique x-coordinates, we [pass the coordinates](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L227-L228) (reshaped from `[(x, y), ...]` into `[x, ...], [y, ...]`) into  [`interpolatePolynomial`](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L59-L79). Note that the [`y_sample`](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L75) is only used in a multiplication. At worst, when `y=0`, this step of the lagrangian interpolation is zero and is then [added to the result](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L76). If all steps are zero, `result` simply stays as zero.

There's no risk:

 1. Of server crash, as no division occurs over either `y_samples` elements or `result`. (Zero->Crash)
 2. Of information exposure, as, if a bunch of invalid shares are given (where `y` value is artificially set at zero), the server has no knowledge of the actual secret in the interpolation step and thus cannot leak any secret. (Zero->Hero)

Indeed, it is perfectly valid for a share to have a zero `y` value. When? Exactly when the specified `x` value is an x-intercept of the specified polynomial. :-)

> Cases where this can happen.
> 1. out == 0 when coeffiecients[degree] == 0 a.k.a the last element

I'm not sure I follow. If `coefficients[degree] == 0` but at least one other coefficient is non-zero, then `out==0` iff the requested share's `x` value is a x-intercept of the specified (reduced-degree polynomial), which is perfectly fine and natural. Other (non x-intercept) values for `x` would result in `out != 0`. 

If `out == 0` for all `x` values, this would imply _all_ coefficients are zero, including the constant term (the byte of secret we're splitting). As mentioned in the [ZKDocs -- last bullet point](https://www.zkdocs.com/docs/zkdocs/protocol-primitives/shamir/#how-things-go-wrong), the real concern isn't when this occurs _randomly_, but when this is more indicative of a RNG failure.

Such a RNG failure, while definitely concerning to Vault, would [come](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L4) as a [result](https://github.com/hashicorp/vault/blob/main/shamir/shamir.go#L34-L35) of bugs in Go's `crypto/rand` package, and thus should really be addressed there as they affect far more than simply Vault's Shamir's implementation (or even more than Vault itself). Additionally, specifications such as FIPS 140-3 are trying to make online health checks of RNGs (and their underlying entropy sources) more common place to avoid these types of errors.

Indeed, as my colleague @kitography pointed out recently, in general it isn't a problem for coefficients of the polynomial (even the leading one!) to be zero, _as long as it isn't indicative of a RNG failure or bias in the RNG_. Their argument follows from the same principles of a one-time pad: the all zero-key _must_ occur at some point in the one-time pad key stream (otherwise it violates its randomness requirements), but as long as all _other_ keys are equally likely (and random), you can't distinguish e.g., `Hello Bob` from `CryptoTeam` or `My Doggos` or `Donut Now`. 

Being able to detect whether a given (insufficient subset of) share(s) results in revelaing any given byte of the secret implies you have an oracle for `byte_is_part_of_secret(candidate, index)`, which leads to much more efficient compromise of the key (by simply iterate over all possible bytes at each index and add the one which is valid).

> 2. any of the coeffiecients[i] == 0 In this case we will have our y coefficient to be zero as well.

Again, not sure how this follows :-) Only when the specified `x` value for evaluation is the exact x-intercept of the present non-zero polynomial will the output be zero.

cipherboy:
I'm going to go ahead and close this, for reasons stated above.

PR #13509, opened 2021-12-22
update vault and auth submodules to api v1.3.1

digivava:
Updates the go.mod files to point to the latest tag of the api module.


api/auth/azure/go.mod
```diff
@@ -2,7 +2,4 @@ module github.com/hashicorp/vault/api/auth/azure
 
 go 1.16
 
-require (
-	github.com/Azure/go-autorest/autorest v0.11.21
```
averche, line 6: This got removed by `go mod tidy`, not sure if it's expected

digivava: Looks like that was hanging on from before when the AzureAuth's `WithResource` method took in parameters from go-autorest, but it no longer does so we're good.

averche (approved):

PR #13514, opened 2021-12-22
Backport of website: Upgrade <Subnav /> & <ProductDownloadsPage /> into release/1.9.x

hc-github-team-secure-vault-core:
## Backport

This PR is auto-generated from #13470 to be assessed for backporting due to the inclusion of the label backport/website.

---

[:mag: Preview Link](https://vault-git-nqweb-upgrade-react-subnav-hashicorp.vercel.app/downloads)

---

This PR upgrades the `<Subnav />` and `<ProductDownloadsPage />` components on the Vault website to their latest versions.


EnMod (changes requested):
Unsure of why this cherry-pick would have failed 🤔

mladlow:
We're still working out some rough edges with backport-assistant.

EnMod:
Thanks for the context @mladlow 👍

EnMod:
Got it @mladlow , can do. Thanks for the heads-up 👍

EnMod:
@mladlow For future PRs (on this and our other product site repos), do you advise against using the backport assistant until the issues are solved?

mladlow:
I think we should keep using it. It works some of the time, and if it doesn't work, presumably one would have needed to do the manual cherry pick anyway, since the plan was to backport? That is my reasoning anyway.

The problem that we're having is when backport PRs aren't merged to 1.9 quickly the automatic cherry picking is more likely to fail, so we get progressively more of the "no-op commit due to failed cherry-picking" PRs (at least that is the theory, unclear why this one failed to cherry pick).

EnMod:
Ahhh ok, thanks for the context here. I think your reasoning is sound 👍

PR #13515, opened 2021-12-22
Add ability to optionally clone a Client's token

benashz:
In the Vault Provider we have three separate instances where we clone the parent's token. This PR should make it simpler for us to handle this use case, and provide a more consistent experience in the provider.

Relates to https://github.com/hashicorp/terraform-provider-vault/pull/1270

### DIFF
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -420,6 +420,7 @@ func TestClone(t *testing.T) {
 		name    string
 		config  *Config
 		headers *http.Header
+		token   string
 	}{
 		{
 			name:   "default",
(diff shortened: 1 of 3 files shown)


api/client_test.go
```diff
@@ -1052,3 +1081,45 @@ func TestClient_SetReadYourWrites(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &Client{
+				config: &Config{},
+			}
+
+			var expected bool
+			for _, v := range tt.calls {
+				actual := c.CloneToken()
+				if expected != c.CloneToken() {
```
vinay-gopalan, line 1112: nit: should we compare to `actual` here since we already have the value captured?

benashz: Good call! Fixed in https://github.com/hashicorp/vault/pull/13515/commits/816fc3083787a6e6df795a0689d23b11177eda11

vinay-gopalan (approved):

PR #13516, opened 2021-12-22
fix: algolia-index CircleCI test

EnMod:
This PR fixes the `algolia-index` test by installing the latest `npm` at test time.


BRKalow (approved):
good catch!

PR #13529, opened 2021-12-28
Fix documentation of "replication_performance_mode" in health API

stklcode:
The field `replication_perf_mode` was renamed in 0.9.2 (a109e2a11ea46c78f235b20632f9320d8813258b), but the docs have never been updated.

Update the documentation to present the correct name `replication_performance_mode` as used in the actual implementation:
https://github.com/hashicorp/vault/blob/a620c2dc8f82b91372521a149ce7e77b97587f9c/api/sys_health.go#L34

### DIFF
--- a/website/content/api-docs/system/health.mdx
+++ b/website/content/api-docs/system/health.mdx
@@ -69,7 +69,7 @@ $ curl \
 
 This response is only returned for a `GET` request.
 
-Note: `replication_perf_mode` and `replication_dr_mode` reflect the state of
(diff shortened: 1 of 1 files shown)


hsimon-hashicorp:
Welcome to the HashiCorp Vault repo, @stklcode, and thanks for this catch! I'll have @taoism4504 review it, and we'll go from there. Thanks!

taoism4504 (approved):

Issue #13536, opened 2021-12-31
sys/raw endpoint does not allow writing values that cannot be encoded in json


shwuandwing:
**Describe the bug**
The sys/raw/* write endpoint does allow writing values that cannot be encoded in json.

For example, the mount table optimizes for space by [gzip](https://www.vaultproject.io/docs/internals/limits#mount-point-limits) compressing the entry.  In a recovery scenario where we need to modify the mount table, it is not possible to write in the compressed format (since the gzip data likely contains invalid utf-8 byte sequences and cannot be encoded in json).   In that scenario, the mount limits goes from ~7000 to ~1000 before you couldn't use the raw endpoint to modify the mount table without the table entry being too large to fit in storage.

It is also an issue for fixing issues in the KV mount, since most entries in the KV are protobuf messages, and often cannot be encoded in json.

**To Reproduce**
Pre-req -- setup vault to use consul
1. Run vault with in recovery mode
2. Read out the entry for sys/raw/core/mounts
3. Write back table mount entry into sys/raw/core/mounts
4. Notice the mount table entry in consul is significantly bigger then before
5. Then try to use gzip the entry like vault's compress util library; and then notice writing it out can't work because the gziped data cannot be encoded in json (invalid utf-8 sequences).

**Expected behavior**
sys/raw support writing values that cannot be encoded in json.  i.e.  sys/raw takes an alternative "value_base64" argument that is used if "value" is empty.  "value_base64" is the base64 encoded value of intended entry (allowing data that cannot be encoded in json).

**Environment:**
* Vault 1.9

aphorise:
hey @shwuandwing I'm curious:

1. How are you reading `sys/raw/core/mounts`? is it with: `vault read -format=json sys/raw/core/mounts | jq '.data'`?
2. Any chance you can provide the JSON `.data` of your read?
3. What is the change you'd like to make to the `mounts` document before re-writting it? (if any)
4. How are you performing the write back? (exactly CLI).

PR #13537, opened 2021-12-31
Enhance sys/raw to read and write values that cannot be encoded in json

### LINKED ISSUE #12746: Compression Handler Falsely Assumes All Raw Data is JSON
**Describe the bug**
The [decompression helper](https://github.com/hashicorp/vault/blob/main/sdk/helper/compressutil/compress.go) in the Vault SDK improperly assumes all keys in raw storage are in JSON format. For keys like `sys/token/salt` that is just a random UUID saved as a flat string, this can cause Vault to try and decompress an already decompressed value if it starts with one of the magic bytes.

**To Reproduce**
1. Run Vault with /sys/raw enabled.
1. Find a value in Vault's raw storage that is just a string (not JSON) and starts with `G`, `L`, `S`, or `4`.
   - I would assume you could just write a value starting with the above characters to raw storage as well to test this. 
1. Run `vault read sys/raw/{path}`
1. See error
```
Error reading sys/raw/sys/token/salt: Error making API request.

URL: GET http://localhost:8200/v1/sys/raw/sys/token/salt
Code: 400. Errors:

* lz4: bad magic number
```

**Expected behavior**
Vault should fall back to returning the uncompressed string if the decompression attempt fails. Maybe return the decompression error as a warning.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.8.3 (bug is present in main branch now)
* Vault CLI Version (retrieve with `vault version`): 1.8.3 (bug is present in main branch now)
* Server Operating System/Architecture: All

```json
{
  "raw_storage_endpoint": true
}
```

**Additional context**
The error included above is returned by the logical raw handler here: https://github.com/hashicorp/vault/blob/main/vault/logical_raw.go#L79-L82

shwuandwing:
This enhances sys/raw so it can return / write values that cannot be encoded in json (like gzip data for core/mounts, or protobuf in KV v2 mounts)

Addresses / fixes #12746, #13536


ncabatoff:
Hi @shwuandwing,

I'm excited to see you working on this, as this limitation is something I'd love to see addressed.

I don't think your current solution actually solves #12746; I would expect to still see the same error as was reported in that issue after your changes, since we're still looking at the error from compressutil.Decompress.

What do you think about adding  an optional `compressed` boolean option?  handleRawRead would maintain its current behaviour when the option isn't provided or when compressed=true; if compressed=false is given, it would not attempt to call Decompress.  handleRawWrite would compress the provided value before storing it if compressed=true was provided.  I prefer that we not rely on the user to do the compression of the value they want written, since that seems error prone - I'm sure plenty of users would try to compress the data without using our compressutil.

This doesn't obviate the need for base64.  I don't *think* it's strictly required - I just did some tests with uncompressed binary storage entries and I was able to read/write the value just fine.  But it's not ideal to have binary unencoded data go to the user's terminal.  From that perspective, your solution of returning both the original `value` plus a `value_base64` isn't ideal, in that it still allows for that to happen.  How about instead we just retain the single `value` (no `value_base64`), and add an `encoding=base64` option for both reads and writes?

vault/logical_raw.go
```diff
@@ -72,23 +82,31 @@ func (b *RawBackend) handleRawRead(ctx context.Context, req *logical.Request, da
...
+		if valueBytes == nil {
+			valueBytes = entry.Value
+		}
 	}
 
-	// `outputBytes` is nil if the input is uncompressed. In that case set it to the original input.
-	if outputBytes == nil {
-		outputBytes = entry.Value
+	var value interface{} = string(valueBytes)
+	if encoding == "base64" {
```
ncabatoff, line 103: Should we return an error if encoding is anything other than `base64` or empty string?

vault/logical_raw.go
```diff
@@ -97,6 +115,10 @@ func (b *RawBackend) handleRawRead(ctx context.Context, req *logical.Request, da
 // handleRawWrite is used to write directly to the barrier
 func (b *RawBackend) handleRawWrite(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
 	path := data.Get("path").(string)
+	compressed := data.Get("compressed").(bool)
+	compressionType := data.Get("compression_type").(string)
```
ncabatoff, line 119: Hmm, I understand why you added `compression_type`: if we want to allow arbitrary writes, including writes to create storage entries that don't presently exist, it makes sense to provide this option.

However, similar to my original concern that users could get the compression wrong if we make them do it themselves, I have the same concern here: most users aren't going to know what compression type is appropriate to a given storage path, and I'd rather they not have to study the code to find out, if we can avoid it.  For the case where the write is going to update an existing value - which I presume is going be by far the most common case - can we use compressutil to work out what the compression type of the current stored value is, and thereby make the param optional for updates?

shwuandwing: There is no function in compressutil to return the type of compression used.   I could write one (but my understanding is it would take 3 diffs (one for compressutil in the SDK, one to update the go.mod to reference the new sdk, and then this diff).

Auto-detection behavior could be confusing (only works for updates, but not new values).  And I'm not sure specifying compression_type is that bad.  Because you can read the existing value (without decompression), a user can find the canary byte to know what an existing compression was.

But I'm up for doing multiple diffs if that's what's required to push this diff through.

ncabatoff: > my understanding is it would take 3 diffs (one for compressutil in the SDK, one to update the go.mod to reference the new sdk, and then this diff

We have a `replace => ./sdk` directive in go.mod to use the sdk within the repo, the actual version of the sdk in Vault's go.mod isn't really used.  So you should be able to include the change in this one PR.

> Auto-detection behavior could be confusing (only works for updates, but not new values).

Yeah, I'm not thrilled about that either, but I've never seen sys/raw used to create new values, only delete, read, and update.  I think "creates" should be possible but I'm okay with them being harder (and with this inconsistency) if it makes things better for the much more common update case. 

> And I'm not sure specifying compression_type is that bad. Because you can read the existing value (without decompression), a user can find the canary byte to know what an existing compression was.

It's not that bad, agreed, but please bear in mind that the primary use case we have for sys/raw is: a customer is down, Vault won't start, and this is the only way to get them running again.  In such a high-stress urgent context, anything we can do to make things easier is desirable.

shwuandwing: Thanks for the pointers.  I've updated write so it will infer compression type from the existing value (if it exists and was compressed) -- error out if it cannot infer.  Users can still specify it for new values.

vault/logical_raw.go
shwuandwing: Added a check.

vault/logical_raw.go
shwuandwing: @ncabatoff -- Do you have any other feedback on this change?

vault/logical_raw.go
```diff
@@ -46,6 +48,17 @@ func NewRawBackend(core *Core) *RawBackend {
...
 
+	// Preserve pre-existing behavior to decompress if `compressed` is missing
+	compressed := true
+	if d, ok := data.GetOk("compressed"); ok {
+		compressed = d.(bool)
+	}
+
+	encoding := data.Get("encoding").(string)
+	if encoding != "" && encoding != "base64" {
+		return logical.ErrorResponse("invalid encoding '%s'", encoding), logical.ErrInvalidRequest
```
ncabatoff, line 59: This is fine, but FYI more commonly we omit the second return (`error`) value when we're using logical.ErrorResponse.  logical.ErrInvalidRequest translates to http.StatusBadRequest, which is the default for when we return an error with a non-nil response, as is the case for logical.ErrorResponse.

vault/logical_raw.go
```diff
@@ -72,23 +85,31 @@ func (b *RawBackend) handleRawRead(ctx context.Context, req *logical.Request, da
...
+			valueBytes = entry.Value
+		}
 	}
 
-	// `outputBytes` is nil if the input is uncompressed. In that case set it to the original input.
-	if outputBytes == nil {
-		outputBytes = entry.Value
+	var value interface{} = string(valueBytes)
+	if encoding == "base64" {
+		value = valueBytes
```
ncabatoff, line 107: Shouldn't this be base64.StdEncoding.EncodeToString(valueBytes)?

vault/logical_raw.go
ncabatoff: Yes, but not in this thread.

vault/logical_raw.go
```diff
@@ -110,11 +138,76 @@ func (b *RawBackend) handleRawWrite(ctx context.Context, req *logical.Request, d
...
+	value := []byte(v)
+	if encoding == "base64" {
+		var err error
+		value, err = base64.StdEncoding.DecodeString(v)
+		if err != nil {
+			return logical.ErrorResponse(err.Error()), logical.ErrInvalidRequest
+		}
+	}
+
+	if compressed {
```
ncabatoff, line 151: If the user doesn't provide compressed=true, we're not making use of the information we have in the existing storage entry, and we'll go ahead and replace an existing compressed entry with an uncompressed one.  Can you reorder things so that if this is an update (instead of a create), we use the information in the existing storage entry to determine whether and how to compress?  And in that case maybe we should return an error if they did specify those options and they don't match the existing storage entry?

sdk/helper/compressutil/compress.go
```diff
@@ -141,10 +141,21 @@ func Compress(data []byte, config *CompressionConfig) ([]byte, error) {
...
+	bytes, _, notCompressed, err := DecompressWithCanary(data)
+	return bytes, notCompressed, err
+}
+
+// DecompressWithCanary checks if the first byte in the input matches the canary byte.
+// If the first byte is a canary byte, then the input past the canary byte
+// will be decompressed using the method specified in the given configuration. The type of compression used is also
+// returned. If the first byte isn't a canary byte, then the utility returns a boolean
+// value indicating that the input was not compressed.
+func DecompressWithCanary(data []byte) ([]byte, string, bool, error) {
```
ncabatoff, line 153: I'd prefer not to overload this function which already has a large number of return values, esp since currently callers only need one of these two things (the compression type, or the decompressed data).  How about instead adding a func like

```
func CompressionType(data []byte) string {
	if data == nil || len(data) < 2 {
		return ""
	}

	canary := data[0]

	switch canary {
	case CompressionCanaryGzip:
		return CompressionTypeGzip
	case CompressionCanaryLZW:
		return CompressionTypeLZW
	case CompressionCanarySnappy:
		return CompressionTypeSnappy
	case CompressionCanaryLZ4:
		return CompressionTypeLZ4
	}
	return ""
}
```
?

shwuandwing: I considered this approach before the last revision.

The problem is this approach would recreate a similar bug to #12746.  Any write to non-compressed content that happened to start with 'G', '4' like sys/token/salt would accidentally cause the write to be compressed.

The only way to avoid #12746 is during a write, verify the existing data is actually compressed data by attempting decompression.  If one is going to do that, then it makes sense to find out get the compression type at the same time.

vault/logical_raw.go
shwuandwing: From golang docs.   (https://pkg.go.dev/encoding/json#Marshal)  

"Array and slice values encode as JSON arrays, except **that []byte encodes as a base64-encoded string**, and a nil slice encodes as the null JSON value."

I could change it to value = base64.StdEncoding.EncodeToString(valueBytes) --- but its an extraneous no-op.

vault/logical_raw.go
shwuandwing: Just clarifying.   To differentiate between update vs create, backends generally need to implement "ExistenceCheck" -- https://github.com/hashicorp/vault/blob/main/sdk/framework/path.go#L92.

Are you saying you want this diff to implement "ExistenceCheck" for sys/raw and use that in "handleRawWrite", or skip ExistenceCheck and have "handleRawWrite" do something like

			
entry, err := b.barrier.Get(ctx, path)
_, compressionType, _, _ := compressutil.DecompressWithCanary(entry.Value)
if compressionType != "" && !compressed {
    return err // this is compressed data but compression was not specified
}

One problem with this approach is what if users **want** to replace an existing compressed entry with an uncompressed one (that's actually what happens today if you use sys/raw to update any json-compressed entries like core/audit or core/mounts).

shwuandwing: Okay, I implemented an existence check -- and updates now will use the existing compression type (or no compression) -- and check + error out if the new compression type is different the existing compression type.

I did make one small exception, users can force an entry to be non-compressed even if it was compressed before.  This ensures a users can workaround unknown issues with compressutil .

sdk/helper/compressutil/compress.go
ncabatoff: Ah, good point.

vault/logical_raw.go
ncabatoff: Ah, fair enough.  Mind adding a comment in that case, since it's non-obvious enough to have confused at least one developer (me)?

sdk/helper/compressutil/compress_test.go
```diff
@@ -86,6 +86,15 @@ func TestCompressUtil_CompressDecompress(t *testing.T) {
...
 			t.Fatalf("bad (%s): decompressed value;\nexpected: %q\nactual: %q", test.compressionType, string(inputJSONBytes), string(decompressedJSONBytes))
 		}
+
+		decompressedJSONBytes, compressionType, wasNotCompressed, err := DecompressWithCanary(compressedJSONBytes)
+		if err != nil {
+			t.Fatalf("decompress error (%s): %s", test.compressionType, err)
+		}
+
+		if compressionType != test.compressionConfig.Type {
+			t.Fatalf("bad compressiontType value;\nexpected: %q\nactional: %q", test.compressionConfig.Type, compressionType)
```
ncabatoff, line 96:
```suggestion
			t.Fatalf("bad compressionType value;\nexpected: %q\nactual: %q", test.compressionConfig.Type, compressionType)
```

vault/logical_raw.go
```diff
@@ -110,11 +141,83 @@ func (b *RawBackend) handleRawWrite(ctx context.Context, req *logical.Request, d
...
+	value := []byte(v)
+	if encoding == "base64" {
+		var err error
+		value, err = base64.StdEncoding.DecodeString(v)
+		if err != nil {
+			return logical.ErrorResponse(err.Error()), logical.ErrInvalidRequest
+		}
+	}
+
+	if req.Operation == UpdateCapability {
```
ncabatoff, line 154:
```suggestion
	if req.Operation == logical.UpdateOperation {
```

ncabatoff: I'm not sure this test actually works.  All PUT/POSTs are "update"s by default, and we only distinguish create/update if the backend defines an ExistenceCheck, which I don't think we have for sys/raw.

shwuandwing: I defined an existence check.   lines 282-290 of the latest revision.

ncabatoff: My bad, should've kept reading.

vault/logical_raw.go
```diff
@@ -110,11 +141,83 @@ func (b *RawBackend) handleRawWrite(ctx context.Context, req *logical.Request, d
...
+		// Check if this is an existing value with compression applied, if so, use the same compression (or no compression)
+		entry, err := b.barrier.Get(ctx, path)
+		if err != nil {
+			return handleErrorNoReadOnlyForward(err)
+		}
+		if entry == nil {
+			err := fmt.Sprintf("cannot figure out compression type because entry does not exist")
```
ncabatoff, line 161: Do we want to use compression_type if provided in this case?

vault/logical_system_test.go
```diff
@@ -1940,16 +1941,140 @@ func TestSystemBackend_disableAudit(t *testing.T) {
...
-	}
-	if !strings.HasPrefix(resp.Data["value"].(string), "{\"type\":\"mounts\"") {
-		t.Fatalf("bad: %v", resp)
-	}
+		req := logical.TestRequest(t, logical.ReadOperation, "raw/core/mounts")
+		resp, err := b.HandleRequest(namespace.RootContext(nil), req)
+		if err != nil {
+			t.Fatalf("err: %v", err)
+		}
+		if !strings.HasPrefix(resp.Data["value"].(string), "{\"type\":\"mounts\"") {
```
ncabatoff, line 1952:
```suggestion
		if !strings.HasPrefix(resp.Data["value"].(string), `{"type":"mounts"`) {
```

vault/logical_system_test.go
```diff
@@ -1940,16 +1941,140 @@ func TestSystemBackend_disableAudit(t *testing.T) {
...
+		req := logical.TestRequest(t, logical.ReadOperation, "raw/core/mounts")
+		req.Data = map[string]interface{}{
+			"encoding": "base64",
+		}
+		resp, err := b.HandleRequest(namespace.RootContext(nil), req)
+		if err != nil {
+			t.Fatalf("err: %v", err)
+		}
+
+		if fmt.Sprintf("%T", resp.Data["value"]) != "[]uint8" {
```
ncabatoff, line 1969: This is a bit unconventional, why not something like

```
if _, ok := resp.Data["value"].([]byte); !ok {
```

?

vault/logical_raw.go
ncabatoff: Not resolving this conversation yet because I still want to see us use the correct symbol to compare against (even if they both have the value "update" so it doesn't have any code impact, just a readability thing.)

ncabatoff:
Looking really good, ty for the comprehensive tests.

vault/logical_raw.go
shwuandwing: added comment

vault/logical_raw.go
shwuandwing: In this case, I think error is better.  This is an update operation and somehow the unexpected occurred (there's no data to update).

vault/logical_system_test.go
shwuandwing: Good catch, resolved.

vault/logical_raw.go
shwuandwing: using the correct symbol

vault/logical_raw.go
ncabatoff: Right, well there is a race inherent, since we don't lock between when the ExistenceCheck runs and this code runs.  But I'm ok with leaving it as-is.

ncabatoff (approved):

ncabatoff:
Nice work, thanks @shwuandwing !

Issue #9462, opened 2020-07-13
Vault Proxy Protocol support is broken on Google Cloud


upodroid:
**Describe the bug**

Vault fails to pass Google Health Checks that have proxy protocol enabled. Google Support investigated this and concluded with the following:

The product team found out that vault is not prepared to handle PROXY1V_UNKNOWN string. That is why the healthchecks were failing.

I will share the findings below:

https://github.com/hashicorp/vault/blob/3831cbfac99222866c0f4eb3b9f563f37e2852aa/command/server/listener.go#L43 - vault proxy protocol setup

https://github.com/hashicorp/vault/blob/3831cbfac99222866c0f4eb3b9f563f37e2852aa/helper/proxyutil/proxyutil.go#L9 - proxy protocol implemented in first link - which can handle the `unknown` string - https://github.com/armon/go-proxyproto/blob/f0b8253b1507a570415b16910691d11751a0685c/protocol.go#L52 , but doesn't, as can be seen in proxyutil.go

Given that the healthcheck failed due to vault not being prepared to receive the "UNKNOWN" string, we will not pursue this as a fixable issue.

**To Reproduce**
1. Deploy Google Cloud LB with a Google HC that has the following config.
```
checkIntervalSec: 1
creationTimestamp: '2020-06-28T13:37:23.358-07:00'
description: ''
healthyThreshold: 2
http2HealthCheck:
  host: vault.upo.one
  port: 8200
  proxyHeader: PROXY_V1
  requestPath: /v1/sys/health
id: '471244148877640044'
kind: compute#healthCheck
logConfig:
  enable: true
name: vault-hc
selfLink: https://www.googleapis.com/compute/v1/projects/REDACTED/global/healthChecks/vault-hc
timeoutSec: 1
type: HTTP2
unhealthyThreshold: 2
```

**Expected behavior**
It should work but SSL failures happen.

App Log:

```
{"@level":"info","@message":"http: TLS handshake error from 130.211.0.205:45346: read tcp 10.100.0.20:8200-\u003e130.211.0.205:45346: use of closed network connection","@timestamp":"2020-06-29T09:34:20.480293Z"}
2020-06-29 09:34:20.742588 I | [ERR] Failed to read proxy prefix: Invalid header line: PROXY UNKNOWN
[...]

Issue #3807, opened 2018-01-17
PROXY Protocol v2 logging support


jantman:
**Feature Request:**

Vault currently supports using a client IP from a PROXY protocol header via the work done in #3098 and modified in #3195.

We're looking at leveraging a new AWS solution, [PrivateLink](https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Introduction.html#what-is-privatelink), a SDN construct that allows securely exposing services behind [Network Load Balancers (NLBs)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) in a "service provider" account, to multiple clients in separate AWS accounts - without network-level connections (i.e. it provides a virtual endpoint in client accounts, that routes traffic directly to a load balancer in a service provider account).

Traffic from such connections is seen as coming from a single source IP (the load balancer node), and the only information about which client it's coming from is encoded in a custom Type-Length-Value vector in a PROXY v2 header, per [these docs from AWS](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#proxy-protocol).

I'll admit that I doubt many people outside of AWS have need for this, and I'm not entirely sure what would be involved in parsing this out or how that would be configured. But support for logging PROXYv2 fields (maybe in a way similar to how request headers are logged, via #2321) would be extremely helpful in this situation.

JayH5:
Just to add, NLBs are not only used for PrivateLink and can be used instead of Elastic/Classic Load Balancers (ELB). One big advantage of NLBs over ELBs is that a single NLB will have fixed private IPs, whereas ELB private IPs can change over time. Why this is important is that it means that NLB private IPs can be used for the [`proxy_protocol_authorized_addresses`](https://www.vaultproject.io/docs/configuration/listener/tcp.html#proxy_protocol_authorized_addrs) setting, but ELB ones cannot.

ELB supports only PROXY protocol v1, while NLB supports only v2.

To add v2 support to Vault it would first need to be added to the upstream library, [`go-proxyproto`](https://github.com/armon/go-proxyproto).

thenoid:
I would like see this implemented as well, currently in a similar boat as the OP

craigday:
I would like to see this as well. Just ran into this when adding IP constraints to an AppRole and was not seeing the client's IP addresses. We use NLB end expose Vault across multiple VPCs using endpoint services.

thorix:
[...]

PR #13540, opened 2022-01-02
Add support for PROXY protocol v2 in TCP listener

remilapeyre:
I did not find tests for this so I added one trying to cover different
configurations to make sure I did not break something. As far as I know,
the behavior should be exactly the same as before except for one thing
when proxy_protocol_behavior is set to "deny_unauthorized", unauthorized
requests were previously silently reject because of https://github.com/armon/go-proxyproto/blob/7e956b284f0a/protocol.go#L81-L84
but it will now be logged.

Also fixes https://github.com/hashicorp/vault/issues/9462 by adding
support for `PROXY UNKNOWN` for PROXY protocol v1.

Closes https://github.com/hashicorp/vault/issues/3807


hsimon-hashicorp:
Hi @remilapeyre - please don't forget a changelog entry. :)

hghaf099:
@remilapeyre Thank you for submitting this PR. We would like to review and possibly merge it. Would you please rebase with main and resolve the conflicts?

hsimon-hashicorp:
Hi @remilapeyre - we're super excited to get this reviewed, so could you take a look at the merge conflicts and we can move forward? Thanks! :)

remilapeyre:
> Hi @remilapeyre - we're super excited to get this reviewed, so could you take a look at the merge conflicts and we can move forward? Thanks! :)

Hi @hsimon-hashicorp, all should be good now :)

Issue #11697, opened 2021-05-25
Add Redis Secret backend


josegonzalez:
**Is your feature request related to a problem? Please describe.**

It would be nice to be able to create/remove redis users automatically via a Vault Secret Backend. This would allow operators to setup rotating credentials via Vault in Redis 6+.

**Describe the solution you'd like**

We currently have Redis clusters that have hard-coded passwords. Rotating the passwords is painful. We would like to be able to do this automatically with Vault.

**Describe alternatives you've considered**

We could alternatively:

- not rotate passwords (unsafe)
- manually rotate them (painful, easy to forget to do on)
- run a nomad job to handle rotation (would still tie us to 1 credential per job)

**Explain any additional use-cases**

- This would allow us to provision credential-per-container for any Nomad allocations talking to Redis.
- This would enable tracking user access to Redis.

**Additional context**

Page describing  Redis ACL rules: https://redis.io/commands/acl-setuser

vishalnayak:
Thanks for submitting this! For those others who are also interested in this, please hit the thumbs up emoji on the issue.

remilapeyre:
This has been added at https://developer.hashicorp.com/vault/docs/secrets/databases/redis. @hsimon-hashicorp I think this issue can be closed.

PR #13541, opened 2022-01-02
Add the Redis secrets engine

remilapeyre:
Closes https://github.com/hashicorp/vault/issues/11697


hsimon-hashicorp:
Hi again @remilapeyre - this one needs a changelog too, if you would please. :)

remilapeyre:
Thanks fr the ping @hsimon-hashicorp

website/content/api-docs/secret/redis.mdx
```diff
@@ -0,0 +1,276 @@
...
+layout: api
+page_title: Redis - Secrets Engines - HTTP API
+description: This is the API documentation for the Vault Redis secrets engine.
+---
+
+# Redis Secrets Engine (API)
+
+This is the API documentation for the Vault Redis secrets engine. For general
+information about the usage and operation of the Redis secrets engine, please
+see the [Redis secret engine documentation](/docs/secrets/redis).
```
taoism4504, line 11:
```suggestion
see the [Redis secrets engine documentation](/docs/secrets/redis).
```

website/content/api-docs/secret/redis.mdx
```diff
@@ -0,0 +1,276 @@
...
+
+### Parameters
+
+- `address` `(string: <required>)` - Specifies the address of the Redis server.
+
+- `username` `(string: <required>)` - Specifies the Redis user to connect with.
+
+- `password` `(string)` - Specifies the password to use.
+
+- `rotate` `(bool)` - Whether to rotate the password immediately. This ensures
```
taoism4504, line 33:
```suggestion
- `rotate` `(bool)` - Specifies whether to rotate the password immediately. This ensures
```

website/content/api-docs/secret/redis.mdx
```diff
@@ -0,0 +1,276 @@
...
+  that only Vault knows the password of the administrative user, it is
```
taoism4504, line 34:
```suggestion
  that only Vault knows the password of the administrative user. It is
```

website/content/api-docs/secret/redis.mdx
```diff
@@ -0,0 +1,276 @@
...
+    password="password" \
+    rotate=true
+```
+
+</Tab>
+</Tabs>
+
+## Rotate Root Credentials
+
+This endpoint rotates the password used by Vault. Once this method is called
```
taoism4504, line 77:
```suggestion
This endpoint rotates the password used by Vault. Once this method is called,
```

hsimon-hashicorp:
Hi @remilapeyre - I wanted to come back to this and let you know that we are discussing this internally. This is a little bit more involved, hence the delay. We're aware of the level of effort you've spent on this PR, and I wanted to make sure we communicated that it's appreciated, and our engineering team is talking about it! :)

remilapeyre:
Thanks for letting me know. To give more context this is a secret engine that was asked to us by a Vault Enterprise client. Since it is also one of the most asked for by the community I think it makes more sense to integrate it in Vault rather than distributing it as a separate plugin: this makes it easier as Redis is a very common infrastructure piece and avoid having to audit this and update a separate binary. I think this could be useful to many Vault users. 

I know that an additional secret engine also adds maintenance so not all secret engines can be a part of Vault core but the Redis protocol and feature set is very stable so I think it would be very manageable. I also took the liberty to add myself to the CODEOWNERS file since I don’t mind helping if there is any issue to fix or updates to do. Let me know if you think this is an issue and I should remove myself from there.

Zlaticanin:
Hello Remi, thank you for your contribution. We decided not to proceed with this solution because we want to integrate Redis as a DB plugin, as opposed to the general secrets plugin. We preferred Redis to be a DB plugin because it conforms to a common Database interface, making it easier to maintain and adding less cognitive load to users who are already familiar with the Database secrets engine. During discovery, we found a community repository that supports any kind of Redis OSS installation and we decided to proceed with that one instead. We talked to the contributor and we asked if we could take ownership of his repository, to which he agreed. The plugin is now part of HashiCorp, and has been tested to be fully functional. We are still working on some small modifications and creating the docs, but the plugin will be available on the next major Vault release. Thank you!

Zlaticanin:
https://github.com/hashicorp/vault-plugin-database-redis

PR #13544, opened 2022-01-03
Ignore retry_join elements that refer to ourself.

ncabatoff:
Potential problem with this change I just thought of: what if the retry_join address is a load balancer?  In that case I would expect it to be the only retry_join address in the config, and should match redirect_addr.  I think I'd need something that would compare against listener addrs rather redirect_addr.


ncabatoff:
Closing in favour of #13606

PR #13552, opened 2022-01-03
Fix status icon color after moving to flight icons

hashishaw:
After updating to Flight icons in #12976, the status indicator no longer had the corresponding color (green, yellow, red) to the status of Vault. This PR fixes the regression by adding a span outside the icon, which has a different color property depending on the Vault status.


ui/app/styles/components/status-menu.scss
```diff
@@ -1,31 +1,19 @@
 .status-indicator-button {
   &[data-status='good'] {
-    .status-indicator-icon-dot {
```
hashishaw, line 3: we no longer have control over individual paths on the SVG, so keep just the dot color for each

Monkeychip (approved):
Thanks for tackling this!

PR #13561, opened 2022-01-04
Add just the basic interfaces for managed keys to the OSS side.

sgmiller:
This will at least allow code to test whether the system view returned by
System() can be cast to ManagedKeySystemView, in which case you have access
to the managed key registry.

### DIFF
--- a/vault/managed_key.go
+++ b/vault/managed_key.go
@@ -1,0 +1,11 @@
+package vault
+
+import "context"
+
+type ManagedKey interface {
+	Name() string
+}
+
+type ManagedKeySystemView interface {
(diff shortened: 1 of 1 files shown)


vault/managed_key.go
```diff
@@ -0,0 +1,11 @@
+package vault
```
stevendpclark, line 1: Should this have a negative enterprise build flag? Or will we remove these definitions from the enterprise side?

sgmiller: The latter

stevendpclark (approved):
👍

PR #13568, opened 2022-01-05
UI/Fix SearchSelect focus bug

hellobontempo:
Fixes `SearchSelect` component by downgrading relative components - since ember upgrade the dropdown would not stay focused, to type in the component you had to hold the mouse down. (Bug only presented in Chrome, component worked in Firefox)

Problematic behavior:
search_select


hellobontempo:
Fixed by #13590

PR #13574, opened 2022-01-05
UI/client count 1.10

Monkeychip:
This is hidden work for the Client Count. 1.10 feature. See video on the code you uncomment to show the work. Every change has been peer reviewed by Claire or myself. However, it would be nice to get a review outside us two for this first chunk of the feature. 

We want to merge this into main to avoid falling behind on changes. We are also certain (as you can be) that this work will be complete and done for 1.10. 

All the API endpoints are mocked at this point using mirage.

Confirmed no changelog at this point until the final PR.

https://user-images.githubusercontent.com/6618863/148270259-cb91c9f3-ff54-438e-a16e-9a9f5a49ff6e.mp4


ui/lib/core/addon/components/bar-chart.js
```diff
@@ -47,11 +47,10 @@ const CHAR_LIMIT = 18; // character count limit for y-axis labels to trigger tru
...
 
 // COLOR THEME:
-const BAR_COLOR_DEFAULT = ['#BFD4FF', '#8AB1FF'];
+const LIGHT_AND_DARK_BLUE = ['#BFD4FF', '#8AB1FF'];
 const BAR_COLOR_HOVER = ['#1563FF', '#0F4FD1'];
-const BACKGROUND_BAR_COLOR = '#EBEEF2';
 const TOOLTIP_BACKGROUND = '#525761';
-
+const GREY = '#EBEEF2';
 class BarChartComponent extends Component {
```
hellobontempo, line 54: This component will ultimately be deleted

ui/package.json
```diff
@@ -124,6 +125,7 @@
     "ember-load-initializers": "^2.1.2",
     "ember-maybe-import-regenerator": "^0.1.6",
     "ember-maybe-in-element": "^2.0.3",
+    "ember-modal-dialog": "^3.0.3",
```
hellobontempo, line 128: This is for the tooltip the charts use

ui/package.json
```diff
@@ -138,6 +139,7 @@
     "ember-svg-jar": "^2.1.0",
     "ember-template-lint": "^3.14.0",
     "ember-test-selectors": "^2.1.0",
+    "ember-tether": "^2.0.1",
```
hellobontempo, line 142: Required for `ember-modal-dialog`

ui/app/utils/chart-helpers.js
```diff
@@ -0,0 +1,16 @@
+import { format } from 'd3-format';
```
hashishaw, line 1: 👏

ui/mirage/config.js
```diff
@@ -22,11 +22,280 @@ export default function () {
     };
   });
 
-  this.get('/sys/internal/ui/feature-flags', (db) => {
-    const featuresResponse = db.features.first();
+  this.get('/sys/internal/ui/feature-flags', () => {
```
hashishaw, line 25: I know this is just a mirage file, but this doesn't look like the right path for this response?

ui/app/templates/vault/cluster/clients/index.hbs
```diff
@@ -62,5 +63,6 @@
   </Toolbar>
   <Clients::Config @model={{this.model.config}} @isLoading={{this.currentlyLoading}} />
 {{else}}
+  {{!-- <Clients::Dashboard @tab={{this.tab}} @model={{this.model}} @isLoading={{this.currentlyLoading}} /> --}}
```
hashishaw, line 66: Is this still needed?

ui/app/templates/vault/cluster/clients/index.hbs
```diff
@@ -62,5 +63,6 @@
   </Toolbar>
   <Clients::Config @model={{this.model.config}} @isLoading={{this.currentlyLoading}} />
 {{else}}
```
hashishaw, line 65: It might be cleaner to check for the tab here and render a different component rather than keeping it all in the Clients::Dashboard component. I know that the layouts are very similar, but that component is getting hard to follow and I'd prefer a little duplication over not being able to parse what the template should look like in a given state.

ui/app/templates/vault/cluster/clients/index.hbs
```diff
@@ -7,6 +7,7 @@
 </PageHeader>
 
 <div class="tabs-container box is-bottomless is-marginless is-fullwidth is-paddingless" data-test-pricing-metrics>
+  {{! template-lint-configure no-unknown-arguments-for-builtin-components "warn" }}
```
hashishaw, line 10: Why are we changing this to warn?

ui/app/templates/components/clients/horizontal-bar-chart.hbs
```diff
@@ -0,0 +1,25 @@
+<svg
+  class="chart is-horizontal"
+  {{on "mouseleave" this.removeTooltip}}
+  {{did-insert this.renderChart @dataset}}
+  {{did-update this.renderChart @dataset}}
```
hashishaw, line 5: Are the rest of the charts also going to get `did-update` so they're responsive?

ui/app/styles/utils/_spacing.scss
```diff
@@ -4,6 +4,7 @@ $spacing-xxs: 4px;
 $spacing-xs: 8px;
 $spacing-s: 12px;
 $spacing-m: 16px;
+$spacing-xm: 18px;
```
hashishaw, line 7: Since this value is only used in charts.scss I'd rather define it there and use it within that file's scope only, instead of mucking up our already-in-flux util spacing. Or maybe define here but as `$spacing-chart`

ui/app/components/clients/line-chart.js
```diff
@@ -0,0 +1,121 @@
...
+ * LineChart components are used to display data in a line plot with accompanying tooltip
+ *
+ * @example
+ * ```js
+ * <LineChart @dataset={dataset} />
+ * ```
+ * @param {array} dataset - dataset is an array of objects
+ */
+
+export default class LineChart extends Component {
```
hashishaw, line 23: If the charts share some core things, I think you could make a base `Chart` component and then import & extend that so that if shared interactions/styles update, you only have to update it on the base component

hashishaw (commented):
This is a great start! Some non-blocking comments, but it might be best to address them before we merge and forget they exist :)

ui/app/utils/chart-helpers.js
hashishaw: (clapping for this util file not the line 😂 )

ui/app/templates/vault/cluster/clients/index.hbs
Monkeychip: Yeah, this is the hidden code. It's how users can access the 1.9 client counts while we work on 1.10 which has no data coming in.

ui/app/templates/vault/cluster/clients/index.hbs
Monkeychip: Warn as kind of a reminder. We likely can't do anything about it soon, but it's not something we want to forget. A discussion Jordan and I had.

ui/app/templates/components/clients/horizontal-bar-chart.hbs
Monkeychip: Maybe. It depends on the data flow. Right now the data is going to come into the dashboard component and be sent to components from there. But this is a good reminder. I'll check-in with Claire on this.

ui/app/templates/vault/cluster/clients/index.hbs
Monkeychip: Good idea. I'll put a comment in the code as right now changing this will break the layout for client count view 1.9.

ui/app/templates/components/clients/horizontal-bar-chart.hbs
hellobontempo: Yes - thank you for pointing out! It will be added to the other charts once we have the calendar wired up, which will update the dataset passed in. (Since this code was taken from the bar chart in 1.9 it already has `did-update`. I didn't want to blindly include the helper in the other charts until we had a way to mock the dataset changing and test the functionality)

ui/app/components/clients/line-chart.js
hellobontempo: Can we sync on this to clarify? I think I follow, but want to make sure I'm understanding!

ui/mirage/config.js
hellobontempo: Good catch! This was a merge conflict that had been resolved incorrectly when main was last merged into this branch

hashishaw (approved):

PR #13585, opened 2022-01-06
UI/fix kmip role form

hashishaw:
Before this fix, there was a bug where creating a KMIP role did not save any non-operator keys. There was also an issue where TLS info was not displayed on the show details page, due to an issue in InfoTableRow. As part of this work, I glimmerized InfoTableRow Component and added tests for various scenarios.

**Before (key bits value does not save)**
bad-role

**After**
kmip-role-after

### DIFF
--- a/ui/app/templates/components/identity/item-alias/alias-details.hbs
+++ b/ui/app/templates/components/identity/item-alias/alias-details.hbs
@@ -1,4 +1,4 @@
-<InfoTableRow @label="Name" @value={{@model.name}} @data-test-alias-name={{true}} />
+<InfoTableRow @label="Name" @value={{@model.name}} data-test-alias-name={{true}} />
 <InfoTableRow @label="ID" @value={{@model.id}} />
 <InfoTableRow @label={{if (eq @model.identityType "entity-alias") "Entity ID" "Group ID"}} @value={{this.model.canonicalId}}>
   <LinkTo
--- a/ui/tests/integration/components/info-table-row-test.js
+++ b/ui/tests/integration/components/info-table-row-test.js
@@ -76,9 +76,9 @@ module('Integration | Component | InfoTableRow', function (hooks) {
     this.set('isCopyable', false);
 
     await render(hbs`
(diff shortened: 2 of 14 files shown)


ui/app/templates/components/identity/item-alias/alias-details.hbs
```diff
@@ -1,4 +1,4 @@
-<InfoTableRow @label="Name" @value={{@model.name}} @data-test-alias-name={{true}} />
+<InfoTableRow @label="Name" @value={{@model.name}} data-test-alias-name={{true}} />
```
Monkeychip, line 1: Curious how that @data.... got in there!

Monkeychip: Probably an upgrade maybe even back to the big one in 1.7

hashishaw: I think it was the latest ember upgrade. I considered find-replacing all the `@data-test` instances, but we have a few cases where it looks like they're actually supposed to work that way (eg. ToolbarSecretLink 😱)

ui/tests/integration/components/info-table-row-test.js
```diff
@@ -156,10 +156,96 @@ module('Integration | Component | InfoTableRow', function (hooks) {
...
       @alwaysRender={{true}}>
-      Block content is here 
+      Block content is here
       </InfoTableRow>`);
 
     let block = document.querySelector('[data-test-value-div]').textContent.trim();
     assert.equal(block, 'Block content is here', 'renders block passed through');
   });
+
+  test('Row renders when block content even if alwaysRender = false', async function (assert) {
```
Monkeychip, line 166: Thank you for adding test coverage!

Monkeychip (approved):
Nice work!

zofskeez:
I was wondering how the layout thing was going to work when glimmerizing addon components. Looks like the template has to be moved along side the component file and it just works?

PR #13590, opened 2022-01-06
Search Select Input Fix

zofskeez:
There is an issue in the _PowerSelect_ component where triggering the search input with the _mousedown_ event will cause the input/content to immediately close, forcing the user to hold the mouse button down to interact. Changing to `@eventType="click"` fixes the issue but unfortunately the _PowerSelectWithCreate_ component does not proxy that value to _PowerSelect_.

This PR addresses the issue by removing the _ember-power-select-with-create_ and incorporating the necessary functionality into the _SearchSelect_ component which now uses _PowerSelect_ directly.

### DIFF
--- a/ui/lib/core/addon/components/search-select.js
+++ b/ui/lib/core/addon/components/search-select.js
@@ -1,34 +1,36 @@
 import Component from '@ember/component';
 import { inject as service } from '@ember/service';
 import { task } from 'ember-concurrency';
-import { computed } from '@ember/object';
+import { computed, get } from '@ember/object';
 import { singularize } from 'ember-inflector';
+import { resolve } from 'rsvp';
+import { filterOptions, defaultMatcher } from 'ember-power-select/utils/group-utils';
 import layout from '../templates/components/search-select';
 
 /**
  * @module SearchSelect
- * The `SearchSelect` is an implementation of the [ember-power-select-with-create](https://github.com/poteto/ember-cli-flash) used for form elements where options come dynamically from the API.
+ * The `SearchSelect` is an implementation of the [ember-power-select](https://github.com/cibernox/ember-power-select) used for form elements where options come dynamically from the API.
  * @example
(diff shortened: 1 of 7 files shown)


zofskeez:
I looked at adding test coverage but there are already 16 tests for this component, some of which cover the search and create functionality ported from _ember-power-select-with-create_ which are passing for me after the last commit.

ui/lib/core/addon/components/search-select.js
```diff
@@ -3,32 +3,34 @@ import { inject as service } from '@ember/service';
...
+ * @param {string} [subText] - Text to be displayed below the label
+ * @param {string} [subLabel] - a smaller label below the main Label
+ * @param {string} [wildcardLabel] - when you want the searchSelect component to return a count on the model for options returned when using a wildcard you must provide a label of the count e.g. role.  Should be singular.
  *
- * @param options {Array} - *Advanced usage* - `options` can be passed directly from the outside to the
+ * @param {Array} options - *Advanced usage* - `options` can be passed directly from the outside to the
  * power-select component. If doing this, `models` should not also be passed as that will overwrite the
  * passed value.
- * @param search {Func} - *Advanced usage* - Customizes how the power-select component searches for matches -
+ * @param {function} search - *Advanced usage* - Customizes how the power-select component searches for matches -
```
Monkeychip, line 33: thank you for cleaning up the documentation!

ui/lib/core/addon/templates/components/search-select.hbs
```diff
@@ -22,17 +22,14 @@
   {{/if}}
   {{! template-lint-configure simple-unless "warn" }}
   {{#unless (gte this.selectedOptions.length this.selectLimit)}}
-    <PowerSelectWithCreate
+    <PowerSelect
```
Monkeychip, line 25: much cleaner, I like it!

Monkeychip (approved):
Great problem solving!

PR #13601, opened 2022-01-07
UI/fix icon colors

hashishaw:
After updating to use [Hashicorp Flight Icons 🎉](https://github.com/hashicorp/flight) there were a few instances where we need to use a specific color version of the icon. Fixes a couple other small post-upgrade style things as well.

Screen Shot 2022-01-07 at 1 16 15 PM
Screen Shot 2022-01-07 at 1 24 00 PM
Screen Shot 2022-01-07 at 1 24 05 PM

### DIFF
--- a/ui/app/templates/components/alert-popup.hbs
+++ b/ui/app/templates/components/alert-popup.hbs
@@ -11,12 +11,7 @@
         {{this.type.text}}
       </div>
       {{#if this.message}}
-        {{! template-lint-disable block-indentation}}
-        <p
-          class="message-body {{if @isPreformatted "pre"}}"
-          data-test-flash-message-body="true"
-          test="another attribute"
-        >{{this.message}}</p>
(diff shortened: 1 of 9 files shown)


ui/app/templates/components/alert-popup.hbs
```diff
@@ -11,9 +11,12 @@
...
       </div>
       {{#if this.message}}
-        <p class="message-body {{if @isPreformatted "pre"}}" data-test-flash-message-body="true" test="another attribute">
-          {{this.message}}
-        </p>
+        {{! template-lint-disable block-indentation}}
+        <p
+          class="message-body {{if @isPreformatted "pre"}}"
+          data-test-flash-message-body="true"
+          test="another attribute"
```
hellobontempo, line 18: it looks like this `test` attribute may have been left over from the upgrade?

hashishaw: perhaps, good catch! I'll remove

hellobontempo (approved):
Thanks for tackling!

PR #13604, opened 2022-01-08
UI/Fixes secrets list breadcrumb 

### LINKED ISSUE #13581: Navigation bar is missing
**Describe the bug**
I upgraded from 1.2.2 to 1.9.2 and the very first thing I noticed is that the navigation UI is missing. In 1.2.2 the UI had a small breadcrumbs-like navigator at the top that allowed me to easily navigate to any part of the path (e.g. `foo > bar > 123 > xyz > abc`, I could click on `bar` and I'd end up listing the path `foo > bar`).

In 1.9.2 I'm just getting an input that get's updated with the path that I'm currently listing. I get to see the old "breadcrumb-like navigator" only when I navigate to a secret, but not when I'm listing paths.

Vault

This means that instead of being able to go back to a "previous" path with a single click, now I must use both my mouse and keyboard in order to focus the input field and then delete the part of the path that I'm not interested in listing. This is plain terrible UX.

Why was the breadcrumbs navigator removed? Is that a bug or was it decided to be removed?

**To Reproduce**
1. Open Vault UI
2. Navigate to any path
3. See error

**Expected behavior**
I'd like to be able to navigate paths as I used to in 1.2.2.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.9.2
* Vault CLI Version (retrieve with `vault version`): N/A
* Server Operating System/Architecture: Debian Buster

```hcl
storage "mysql" {
  address = "++++++++"
  username = "++++++++"
  database = "++++++++++"
  password = "++++++++++++"
}

listener "tcp" {
  address     = "127.0.0.1:8200"
  tls_disable = 1
}

ui = true
log_level = "warning"
```

**Additional context**
N/A

hellobontempo:
Breadcrumbs weren't updating as file tree was navigated. Fixes #13581
 
**Bug**
secrets_breadcrumb

**With fix**

breadcrumb_bug

### DIFF
--- a/ui/app/templates/components/secret-list-header.hbs
+++ b/ui/app/templates/components/secret-list-header.hbs
@@ -1,7 +1,7 @@
 {{#with (options-for-backend @model.engineType) as |options|}}
   <PageHeader as |p|>
     <p.top>
-      <KeyValueHeader @baseKey={{this.baseKey}} @path="vault.cluster.secrets.backend.list" @root={{@backendCrumb}}>
+      <KeyValueHeader @baseKey={{@baseKey}} @path="vault.cluster.secrets.backend.list" @root={{@backendCrumb}}>
         <li>
           <span class="sep">
             /
--- a/changelog/13604.txt
+++ b/changelog/13604.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixes breadcrumb bug for secrets navigation
+```


hellobontempo:
Will add milestone label once backporting is confirmed!

ui/app/templates/components/secret-list-header.hbs
```diff
@@ -1,7 +1,7 @@
 {{#with (options-for-backend @model.engineType) as |options|}}
   <PageHeader as |p|>
     <p.top>
-      <KeyValueHeader @baseKey={{this.baseKey}} @path="vault.cluster.secrets.backend.list" @root={{@backendCrumb}}>
+      <KeyValueHeader @baseKey={{@baseKey}} @path="vault.cluster.secrets.backend.list" @root={{@backendCrumb}}>
```
Monkeychip, line 4: Ah, probably a backport to 1.9 because I believe in 1.9 I glimmerized this component and this would be a glimmer issue. Nice catch!

Monkeychip (approved):

ui/app/templates/components/secret-list-header.hbs
hellobontempo: @Monkeychip - from what I could tell this behavior has been happening since before 1.7? But I'll reconfirm next week

PR #13612, opened 2022-01-10
Skip tests on backported docs branches.

ncabatoff:
Example backport to which this should apply: https://github.com/hashicorp/vault/pull/13610

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,7 +173,7 @@ jobs:
     - run:
         command: |
           # If the branch being tested starts with ui/ or docs/ we want to exit the job without failing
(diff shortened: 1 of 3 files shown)


mladlow:
Should these edits be in https://github.com/hashicorp/vault/blob/main/.circleci/config/workflows/ci.yml instead? There's a comment at the top of `.circleci/config.yml` saying that it's autogenerated and not to edit it manually, and there's a section in the `ci.yml` file like:

```
  - test-go-race:
      requires:
        - pre-flight-checks
      filters:
        branches:
          # UI and Docs-only branches should skip go tests
          ignore:
            - /^docs\/.*/
            - /^ui\/.*/
```

mladlow (approved):
Great, thanks for cleaning up the old stuff too!

PR #13613, opened 2022-01-10
Client Count Calendar Widget for 1.10 

Monkeychip:
_This is hidden under the Dashboard component, which is commented out._

This PR adds the calendar widget that helps users select an endDate, single Month, or the billing period. Outside of the widget, on the Dashboard component, this PR adds another modal that allows the user to pick the startMonth. This layout/set-up is very specific to the Client Counts data and billing periods for 1.10. Though not wired up in this PR, the billing startDate will init with the startDate under the license endpoint.

Once a start and end date is selected, a network request will be fired off (when the API is ready) to return the data.

Here is a video of the modal and calendar-widget. Note, no data is sent, though the structure is partially in place. You'll also notice that you cannot select an endDate from the calendar-widget that comes before the startDate. 

https://user-images.githubusercontent.com/6618863/148823104-636079b8-bdb4-4785-8a4e-2859d7afe878.mp4

### DIFF
--- a/ui/app/components/calendar-widget.js
+++ b/ui/app/components/calendar-widget.js
@@ -1,0 +1,177 @@
+import Component from '@glimmer/component';
+import layout from '../templates/components/calendar-widget';
+import { setComponentTemplate } from '@ember/component';
+import { format, formatRFC3339 } from 'date-fns';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+// ARG TODO documentation takes start and end for handQuery
+/**
+ * @module CalendarWidget
+ * CalendarWidget components are used in the client counts metrics. It helps users understand the ranges they can select.
+ *
+ * @example
+ * ```js
+ * <CalendarWidget
+ * @param {function} handleQuery - calls the parent pricing-metrics-dates handleQueryFromCalendar method which sends the data for the network request.
+ * @param {object} startDate - The start date is calculated from the parent. This component is only responsible for modifying the end Date. ANd effecting single month.
+ * />
+ *
+ * ```
+ */
+
+const MONTH_MAPPING = [
+  'January',
+  'February',
+  'March',
+  'April',
+  'May',
+  'June',
+  'July',
+  'August',
+  'September',
+  'October',
+  'November',
+  'December',
+];
+
+class CalendarWidget extends Component {
+  currentDate = new Date();
+  currentYear = this.currentDate.getFullYear(); // integer
+  currentMonth = parseInt(format(this.currentDate, 'M')); // integer
+
+  @tracked showCalendar = false;
+  @tracked showSingleMonth = false;
+
+  @tracked displayYear = this.currentYear; // init to currentYear and then changes as a user clicks on the chevrons
+  @tracked disablePastYear = this.isObsoleteYear(); // if obsolete year, disable left chevron
+  @tracked disableFutureYear = this.isCurrentYear(); // if current year, disable right chevron
+  @tracked preselectRangeOfMonths = null; // the number of months selected by the quickSelect options (e.g. last month, last 3 months, etc.)
+  @tracked allMonthsNodeList = [];
+  @tracked areMonthsSelected = false;
+  @tracked mouseClickCount = 0;
+  @tracked clickRange = []; // the range of months by individually selecting them
+  @tracked endDateDisplay;
+  @tracked firstClick;
+  @tracked secondClick;
+
+  constructor() {
+    super(...arguments);
+    // ARG TODO we need to return the config's duration if not default to 12 months to calculate, now it's 12 months
+    let date = new Date();
+    date.setMonth(date.getMonth() - 1); // by default you calculate the end month as the month prior to the current month.
+    this.endDateDisplay = format(date, 'MMMM yyyy');
+  }
+
+  // HELPER FUNCTIONS (alphabetically) //
+
+  addClass(element, classString) {
+    element.classList.add(classString);
+  }
+
+  isCurrentYear() {
+    return this.currentYear === this.displayYear;
+  }
+
+  isObsoleteYear() {
+    // do not allow them to choose a end year before the this.args.startDate
+    let startYear = this.args.startDate.split(' ')[1];
+    return this.displayYear.toString() === startYear; // if on startYear then don't let them click back to the year prior
+  }
+
+  removeClass(element, classString) {
+    element.classList.remove(classString);
+  }
+
+  // ACTIONS //
+
+  @action
+  addYear() {
+    this.displayYear = this.displayYear + 1;
+    this.disableMonths();
+    this.disableFutureYear = this.isCurrentYear();
+    this.disablePastYear = this.isObsoleteYear();
+  }
+
+  @action
+  disableMonths() {
+    this.allMonthsNodeList = document.querySelectorAll('.is-month-list');
+    this.allMonthsNodeList.forEach((e) => {
+      // clear all is-readOnly classes and start over.
+      this.removeClass(e, 'is-readOnly');
+
+      let elementMonthId = parseInt(e.id.split('-')[0]); // dependent on the shape of the element id
+      // for current year
+      if (this.currentMonth <= elementMonthId) {
+        // only disable months when current year is selected
+        if (this.isCurrentYear()) {
+          e.classList.add('is-readOnly');
+        }
+      }
+      // compare for startYear view
+      if (this.displayYear.toString() === this.args.startDate.split(' ')[1]) {
+        // if they are on the view where the start year equals the display year, check which months should not show.
+        let startMonth = this.args.startDate.split(' ')[0]; // returns month name e.g. January
+        // return the index of the startMonth
+        let startMonthIndex = MONTH_MAPPING.indexOf(startMonth);
+        // then add readOnly class to any month less than the startMonth index.
+        if (startMonthIndex > elementMonthId) {
+          e.classList.add('is-readOnly');
+        }
+      }
+    });
+  }
+
+  // action to parent Dashboard
+  @action
+  selectEndMonth(month, year, e) {
+    // select month
+    this.addClass(e.target, 'is-selected');
+    // API requires start and end time as EPOCH or RFC3339 timestamp
+    let endMonthSelected = formatRFC3339(new Date(year, month));
+    this.args.handleEndMonth(endMonthSelected);
+    let monthName = MONTH_MAPPING.find((e, index) => {
+      if (index === month) {
+        return e;
+      }
+    });
+    this.endDateDisplay = `${monthName} ${year}`;
+    this.toggleShowCalendar();
+  }
+
+  @action
+  selectSingleMonth(month, year, e) {
+    // select month
+    this.addClass(e.target, 'is-selected');
+    // API requires start and end time as EPOCH or RFC3339 timestamp
+    // let singleMonthSelected = formatRFC3339(new Date(year, month));
+    // ARG TODO unsure on what we're going to do yet with date. Depends on API.
+    // this.args.handleEndMonth(singleMonthSelected);
+    // let monthName = MONTH_MAPPING.find((e, index) => {
+    //   if (index === month) {
+    //     return e;
+    //   }
+    // });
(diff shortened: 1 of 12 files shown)


ui/app/components/clients/dashboard.js
```diff
@@ -9,9 +9,28 @@ export default class Dashboard extends Component {
     { key: 'distinct_entities', label: 'unique entities' },
     { key: 'non_entity_tokens', label: 'non-entity tokens' },
   ];
-  @tracked selectedNamespace = null;
```
Monkeychip, line 12: this is now on row 25, alphabetically organized with all the other tracked properties.

ui/app/components/calendar-widget.js
```diff
@@ -0,0 +1,176 @@
+import Component from '@glimmer/component';
+import layout from '../templates/components/calendar-widget';
+import { setComponentTemplate } from '@ember/component';
+import { format, formatRFC3339 } from 'date-fns';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+// ARG TODO documentation takes start and end for handQuery
```
Monkeychip, line 8: this component will eventually have functions passed down from the parent which is why I left the TODO comment there.

ui/app/components/calendar-widget.js
```diff
@@ -0,0 +1,176 @@
...
+  @tracked disablePastYear = this.isObsoleteYear(); // if obsolete year, disable left chevron
+  @tracked disableFutureYear = this.isCurrentYear(); // if current year, disable right chevron
+  @tracked endDateDisplay;
+  @tracked showCalendar = false;
+  @tracked showSingleMonth = false;
+
+  constructor() {
+    super(...arguments);
+    // ARG TODO we need to return the config's duration if not default to 12 months to calculate, now it's 12 months
+    let date = new Date();
```
Monkeychip, line 54: right now I'm calculating the startDate based on today, but the next PR will pull the startDate from the license endpoint and send it to the component as a param. This component is only in charge of endDates or singleMonth or saying, just go back to the default billing period.

ui/app/components/clients/dashboard.js
```diff
@@ -175,8 +217,12 @@ export default class Dashboard extends Component {
   }
 
   @action
-  resetData() {
-    this.barChartSelection = false;
-    this.selectedNamespace = null;
+  selectStartMonth(month) {
```
Monkeychip, line 220: these are fired off from the modal that sets the startMonth. Right now, they do nothing but set a param, when it's wired up they'll fire off a network request.

ui/app/styles/components/calendar-widget.scss
```diff
@@ -0,0 +1,113 @@
...
+      text-align: center;
+      cursor: pointer;
+    }
+    &.is-readOnly {
+      background-color: $ui-gray-100;
+      color: lighten($dark-gray, 30%);
+      pointer-events: none;
+    }
+    &.is-selected {
+      background-color: $blue-500;
```
Monkeychip, line 83: this never shows, even though we do add this class. Thoughts on if I should remove it or it keep it in case there is a lag?

ui/app/styles/core/buttons.scss
```diff
@@ -129,6 +129,14 @@ $button-box-shadow-standard: 0 3px 1px 0 rgba($black, 0.12);
     text-shadow: 0 1px 1px rgba($black, 0.25);
   }
 
+  &.is-link {
```
Monkeychip, line 132: We need the button we use to look like a link because we were told it should be a button even if it doesn't look like one for accessibility reasons.

ui/app/templates/components/calendar-widget.hbs
```diff
@@ -0,0 +1,285 @@
...
+            class="is-month-list"
+            id={{concat "11-" this.displayYear}}
+          >
+            December
+          </button>
+        </div>
+      </div>
+    {{/if}}
+
+    {{#if this.showSingleMonth}}
```
Monkeychip, line 168: This is essentially a copy of the above calendar chunk with a different instance and id names. The reason I did this, even though it's a copy is because the widget needs to know what calendar we're looking at. They behave differently even if they look identical.

ui/app/components/calendar-widget.js
```diff
@@ -0,0 +1,176 @@
...
+  'September',
+  'October',
+  'November',
+  'December',
+];
+
+class CalendarWidget extends Component {
+  currentDate = new Date();
+  currentYear = this.currentDate.getFullYear(); // integer
+  currentMonth = parseInt(format(this.currentDate, 'M')) - 1; // integer and zero index
```
hellobontempo, line 41: It may be cleaner to use `this.currentDate.getMonth()` here? Since all we want is the integer

ui/app/components/calendar-widget.js
```diff
@@ -0,0 +1,176 @@
...
+
+  @tracked allMonthsNodeList = [];
+  @tracked displayYear = this.currentYear; // init to currentYear and then changes as a user clicks on the chevrons
```
hellobontempo, line 44: Thank you for these comments! 😍

ui/app/components/calendar-widget.js
```diff
@@ -0,0 +1,176 @@
...
+  @tracked endDateDisplay;
+  @tracked showCalendar = false;
+  @tracked showSingleMonth = false;
+
+  constructor() {
+    super(...arguments);
+    // ARG TODO we need to return the config's duration if not default to 12 months to calculate, now it's 12 months
+    let date = new Date();
+    date.setMonth(date.getMonth() - 1); // by default you calculate the end month as the month prior to the current month.
+    this.endDateDisplay = format(date, 'MMMM yyyy');
```
hellobontempo, line 56: I think naming here is a little confusing (yay dates!) but this will be the `startDateDisplay` as it's the billing start date and, as you mentioned above, will be grabbed from the license endpoint? 

Once all wired up, the constructor may look more like this - right? 
```
    let startDate = this.args.billingStart
    this.startDateDisplay = format(startDate, 'MMMM yyyy') 
```

ui/app/components/calendar-widget.js
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  // HELPER FUNCTIONS (alphabetically) //
```
hellobontempo, line 59: alphabetically 👏 👏 👏 😻

ui/app/components/calendar-widget.js
Monkeychip: Good question. I think my comment needs to change. Because it's the endDate that is displayed. 
image
The Decemeber 2021 in this image.

ui/app/components/calendar-widget.js
```diff
@@ -0,0 +1,176 @@
...
+
+  @action
+  selectCurrentBillingPeriod() {
+    // ARG TOOD send to dashboard the select current billing period. The parent may know this it's just a boolean.
+    // Turn the calendars off if they are showing.
+    this.showCalendar = false;
+    this.showSingleMonth = false;
+  }
+  @action
+  selectEndMonth(month, year, e) {
```
hashishaw, line 124: what is `e` supposed to be? A more descriptive param name would be helpful :)

ui/app/components/clients/dashboard.js
```diff
@@ -161,6 +184,25 @@ export default class Dashboard extends Component {
     });
   }
 
+  @action
+  handleEditStartMonth(month, year) {
+    if (!month && !year) {
+      // reset and do nothing. They pressed cancel
+      this.startMonth = this.startYear = null;
```
hashishaw, line 191: Are you intending to set `startMonth` to a boolean here?

ui/app/styles/components/calendar-widget.scss
hashishaw: I think there's no harm in keeping it, especially if there is a lag

ui/app/styles/core/title.scss
```diff
@@ -11,6 +11,10 @@
     color: $black;
     text-decoration: none;
   }
+
+  &.title-has-bottom-margin-xs {
```
hashishaw, line 15: Could you instead add the `has-bottom-margin-xs` class to the title? Looks like it has the same style definition

ui/app/templates/components/calendar-widget.hbs
```diff
@@ -0,0 +1,285 @@
...
+            type="button"
+            class="button is-transparent padding-right"
+            disabled={{this.disableFutureYear}}
+            {{on "click" this.addYear}}
+          >
+            <Chevron @direction="right" @size="s" class="has-text-grey" />
+          </button>
+        </div>
+        <div {{did-insert this.disableMonths}} class="calendar-widget-grid calendar-widget">
+          <button
```
hashishaw, line 68: It would be lovely to make this template a little DRYer by iterating over the months and rendering each option -- less potential for error when editing this code in the future, too

hashishaw (changes requested):
So excited for the slick calendar widget 😻

ui/app/components/calendar-widget.js
Monkeychip: Good call, it also seems to return as zero index. Didn't know that was a Date FNS method, now I do :)

ui/app/components/calendar-widget.js
hellobontempo: Ah! I see - yes that will change/be clearer with real data

ui/app/components/calendar-widget.js
```diff
@@ -0,0 +1,176 @@
...
+    date.setMonth(date.getMonth() - 1); // by default you calculate the end month as the month prior to the current month.
+    this.endDateDisplay = format(date, 'MMMM yyyy');
+  }
+
+  // HELPER FUNCTIONS (alphabetically) //
+  addClass(element, classString) {
+    element.classList.add(classString);
+  }
+
+  isCurrentYear() {
```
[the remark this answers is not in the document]
Monkeychip: @chelshaw likely a dumb question. What is the benefit of making these getters?

ui/app/components/calendar-widget.js
Monkeychip: good call. it's the element, so the specific month selected. In this particular case we need to filter through the months using the index and return that element when they match.

modified e to element to clarify.

ui/app/components/clients/dashboard.js
Monkeychip: At the beginning, via a tracked property startMonth is null, and in this situation, I was trying to prevent them from saving a new startMonth if they hadn't selected one. However, I solved for this by disabling save unless they have a startMonth and startYear. So I can remove this. Thanks for the question!

ui/app/styles/components/calendar-widget.scss
Monkeychip: thanks!

ui/app/styles/core/title.scss
Monkeychip: Good catch! Thank you.

ui/app/templates/components/calendar-widget.hbs
Monkeychip: Good call. I thought about this and promptly forgot over the weekend. Will amend now.

ui/app/components/clients/dashboard.js
```diff
@@ -9,9 +9,28 @@ export default class Dashboard extends Component {
     { key: 'distinct_entities', label: 'unique entities' },
     { key: 'non_entity_tokens', label: 'non-entity tokens' },
   ];
-  @tracked selectedNamespace = null;
+  // For startDate Modal
+  months = Array.from({ length: 12 }, (item, i) => {
```
hellobontempo, line 13: We could use the const `DATE_MAPPING` here if it was in a util file - might make sense if there are other shared functions between files

ui/app/components/calendar-widget.js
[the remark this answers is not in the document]
hashishaw: Pretty sure it does a similar thing as `computed` did, where it will be updated each time and stay up to date. No such guarantees if it's not a getter, but if you're calling it explicitly each time it might not matter 🤷

hashishaw (approved):
🚀

PR #13615, opened 2022-01-10
If we get a 405 doing an HTTP PATCH, assume the server is pre-1.9 and…

### LINKED ISSUE #13502: Vault cli kv patch performs call to api v1
**Describe the bug**
When executing `vault kv patch` it shows error with 405 http response with path `https://vault_domain/v1/secret/data/my_secrets`
Rollback to version 1.7 resolved problem

**To Reproduce**
1. Run `vault kv patch`
2. See error

**Expected behavior**
Command should without any problems make patch request to api v2, which supports patch request 

**Environment:**
* vault version 1.9.1:
* Fedora 34

ncabatoff:
… fall back to old readThenWrite approach.  Fixes #13502

### DIFF
--- a/changelog/13615.txt
+++ b/changelog/13615.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Fix using kv patch with older server versions that don't support HTTP PATCH.
+```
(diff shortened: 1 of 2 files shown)


ccapurso (approved):
Looks good!

PR #13620, opened 2022-01-10
Mirage Dev Workflow

zofskeez:
_These changes were pulled out of the MFA branch so they can be widely used immediately._ 

The aim of this work is to improve the experience when opting to use mirage during feature development. Running the command `yarn start:mirage handlerName` will start ember with mirage enabled and bypass the default config in favor of specific handlers. This can be useful when new endpoints are being mocked and the rest of the requests should be passed through or when intercepting individual requests and modifying the response for added functionality.

**Steps to add a dev handler:**
- create js file in `mirage/handlers` directory
- export default function with server as the lone argument
- use the server instance to define handlers eg. `server.get('/some/endpoint', callback);`
- import handler in `mirage/handlers/index.js` file and add to export object
- start ember server eg. `yarn start:mirage handlerName`

When starting ember with mirage and a handler is not specified or the specified handler is not found the `base` handlers will be used. This is also the case when using mirage in the test environment or when starting mirage without running `yarn start:mirage`.


ui/mirage/config.js
```diff
@@ -1,581 +1,317 @@
...
-            _comment: 'by_namespace will remain as it is',
-          },
-        ],
-        months: [
-          {
-            'jan/2022': {
+      };
+    });
+
+    this.get('/sys/internal/counters/activity/monthly', function () {
```
hellobontempo, line 47: This is technically the response from the `/activity` endpoint - I created a `activity/monthly` response to get data to charts for that specific tab in the UI. The tab is no longer relevant in 1.10, but can be left as is for now.

Just a note @Monkeychip - we will need to update these when wiring up data with our client count work

ui/mirage/handlers/example.js
```diff
@@ -0,0 +1,9 @@
+// this can be removed at any time
```
hellobontempo, line 1: thank you for including this example! definitely clarifies the steps

changelog/13620.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
hellobontempo, line 1: We don't need to add a changelog for developer happiness additions (and can just mention that in the PR to explain why there isn't one)

hellobontempo (approved):
Thank you, Jordan! This is super helpful. Good to go after removing the changelog  🚀

changelog/13620.txt
zofskeez: Ok thanks that's good to know I'll remove it.

PR #13627, opened 2022-01-11
add security-scan for CRT

claire-labry:
What this PR does is it enables security scanning (owned by the Security team) for this repository. This is a part of the Common Release Tooling (CRT) effort which `enables a consistent workflow for security scanning that can be used across products and services. It provides a single abstraction point to allow us integrate our own scanning capabilities, as well as leverage external tools and APIs from vendors, open source, or free-to-use projects.` Learn more about HashiCorp's security-scanner [here](https://github.com/hashicorp/security-scanner/).

The security-scan tests are passing successfully: 

[security-scan-binaries](https://github.com/hashicorp/crt-workflows-common/actions/runs/1683527590)
[security-scan-containers](https://github.com/hashicorp/crt-workflows-common/actions/runs/1683537405)

You'll also see that I had to upgrade the Dockerfile to use alpine:3  as the older version caused the security-scanner to pick up some vulnerabilities, CVEs to be specific. See error outputs [here](https://github.com/hashicorp/crt-workflows-common/runs/4777697057?check_suite_focus=true) that arose. We worked with Security and they scanned different versions of the alpine base image and concluded that version 3 picks up the most stable + clean version of alpine and suggested to upgrade to that version to avoid any friction and potential vulnerabilities in the future and simply tidy up the container.

I ran into some merge conflicts, but resolved those and will be happy to squash and merge once this PR is approved. 

If anyone needs any further clarification or if nothing makes sense, feel free to comment or reach out to me! ❤️


mladlow:
I'm requesting review of the Docker aspects from @jasonodonnell.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.14 as default
+FROM alpine:3 as default
```
jasonodonnell, line 1: This looks fine as it will always pull in the latest version of 3.x, but my only concern is it's difficult to tell what version of Alpine was used in the build. As long as we can keep a receipt of that, and we're okay with always rolling forward, then this should be fine.

sarahethompson (approved):
LGTM

mladlow (approved):
Thanks @claire-labry! Do you need someone on the Vault team to merge this?

PR #13630, opened 2022-01-11
Update s3.mdx

frbayart:
fix IAM requirements to use KMS key


hsimon-hashicorp:
Looks okay to me, @taoism4504. Thanks for your contribution, @frbayart! :)

taoism4504 (approved):

taoism4504:
@frbayart -the reason why this is blocked is because the repo was forked, which, as I understand it, disallows any merging of PRs. Please clone the repo and resubmit your changes, and we can get it merged. Thanks.

frbayart:
I am not allowed to create a PR based on a clone on your repository; it needs to come from the fork. Your automation is probably keeping safe your secret when it comes from a fork and doesn't pass the checks I guess.

In similar situations in different projects solutions are:
- manually run the check so it will be allowed to use secrets (safe one if your workflow allows it)
- add contributor in approval run-list
- recreate the PR yourself and take ownership on changes

PR #13635, opened 2022-01-12
UI/Clients single month template

hellobontempo:
Creates single month view for client count data 
Screen Shot 2022-01-12 at 8 40 18 AM

_no changelog because part of feature for 1.10 client count work_

### DIFF
--- a/ui/app/styles/components/stat-text.scss
+++ b/ui/app/styles/components/stat-text.scss
@@ -1,6 +1,6 @@
 .stat-text-container {
   line-height: normal;
-  height: 100%;
+  max-height: 100%;
   display: flex;
   flex-direction: column;
 
@@ -17,7 +17,6 @@
       font-weight: $font-weight-normal;
       color: $ui-gray-700;
       line-height: inherit;
-      flex-grow: 1;
     }
     .stat-value {
       font-size: $size-3;
@@ -103,4 +102,16 @@
       line-height: inherit;
     }
   }
+}
+
+.has-gap {
+  gap: $spacing-xxl;
+}
+
+.stat-text-wrapper {
+  display: flex;
+  justify-content: flex-start;
+  flex-wrap: wrap;
+  column-gap: 70px;
+  row-gap: 20px;
 }
--- a/ui/app/styles/components/stat-text.scss
+++ b/ui/app/styles/components/stat-text.scss
@@ -108,7 +108,7 @@
   gap: $spacing-xxl;
 }
 
-.stat-text-wrapper {
+.stat-text-flex-wrapper {
   display: flex;
   justify-content: flex-start;
   flex-wrap: wrap;
--- a/ui/app/styles/core/charts.scss
+++ b/ui/app/styles/core/charts.scss
@@ -32,7 +32,10 @@
   grid-column-end: span col4-end;
   grid-row-start: 1;
   box-shadow: inset 0 -1px 0 $vault-gray-200;
-  margin-bottom: $spacing-xl;
+
+  > h2 {
+    padding-bottom: $spacing-xs;
+  }
 }
 
 .has-export {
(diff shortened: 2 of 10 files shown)


ui/app/styles/core/charts.scss
```diff
@@ -32,7 +32,10 @@
   grid-column-end: span col4-end;
   grid-row-start: 1;
   box-shadow: inset 0 -1px 0 $vault-gray-200;
-  margin-bottom: $spacing-xl;
```
hellobontempo, line 35: Removing because components don't all have the same bottom margin size

ui/app/templates/components/clients/attribution.hbs
```diff
@@ -1,5 +1,5 @@
 <div class="chart-wrapper dual-chart-grid">
-  <div class="chart-header has-export">
+  <div class="chart-header has-export has-bottom-margin-m">
```
hellobontempo, line 2: using sizing specific margin classes (from `helpers.scss`) so each template header has the correct margin size

ui/lib/core/addon/helpers/format-number.js
```diff
@@ -1,11 +1,11 @@
 import { helper } from '@ember/component/helper';
```
hellobontempo, line 2: Originally was going to add `parseInt` to properly format a string value passed in, but the _only_ reason to do so would be if a string was passed in that wasn't formatted properly (i.e. "2141" instead of "2,141"). It ended up being more trouble than it was worth, for an unlikely edge case. SO, I updated the variable name and went on my merry way.

ui/app/styles/components/stat-text.scss
```diff
@@ -104,3 +103,15 @@
     }
   }
 }
+
+.has-gap {
+  gap: $spacing-xxl;
+}
+
+.stat-text-flex-wrapper {
```
hellobontempo, line 111: all this flex business feels untidy, open to clean-up ideas but was also planning to revisit once the feature was fully built rather than get hung up on css styling now

ui/lib/core/addon/components/stat-text.js
```diff
@@ -4,10 +4,10 @@
  *
  * @example
  * ```js
- * <StatText @label="Active Clients" @stat="4,198" @size="l" @subText="These are the active client counts"/>
+ * <StatText @label="Active Clients" @value="4,198" @size="l" @subText="These are the active client counts"/>
  * ```
  * @param {string} label=null - The label for the statistic
- * @param {string} value=null - Value passed in, usually a number or statistic
+ * @param {integer} value=null - Value passed in, usually a number or statistic
```
zofskeez, line 10: I think the standard jsdoc param type is `number`. I think it's fine if we want to use `integer` but wanted to point that out.

ui/app/styles/components/stat-text.scss
zofskeez: If the column and row gap values may become common across the app you could create a reusable class, something like:
```css
.is-flex-wrap-with-gap {
  flex-wrap: wrap;
  column-gap: 70px;
  row-gap: 20px;
}
```
Something like that could be used along with `.is-flex-start`. I think it's fine the way you approached it though, especially if those gap values are specific to the stat text.

zofskeez (approved):
This looks good to me. Nice work!

ui/lib/core/addon/components/stat-text.js
hellobontempo: Ohh, I'll fix that! Thank you!

ui/app/styles/components/stat-text.scss
hellobontempo: I think this is good to keep in mind, but right now it seems like styling is unique to this component

PR #13638, opened 2022-01-12
KMSE: Key Model / Adapter / Serializer setup

hashishaw:
This PR is large (sorry) but handles the setup for Key Management Secrets Engine (KMSE) Key actions and representation in the app. None of the changes made are available in the app, but if you'd like to pull down this PR and play around with it these are the steps:

1. Enable KMSE in the cli by running `vault secrets enable keymgmt`
2. Run the UI with `yarn start:mirage kms`
3. Add `'keymgmt'` to the `SUPPORTED_SECRET_BACKENDS` array in [supported-secret-backends.js](https://github.com/hashicorp/vault/blob/main/ui/app/helpers/supported-secret-backends.js)


ui/app/adapters/keymgmt/key.js
```diff
@@ -0,0 +1,121 @@
...
+    return {
+      data: {
+        ...data,
+        id: name,
+        backend,
+      },
+    };
+  }
+
+  async updateRecord(store, type, snapshot) {
```
Monkeychip, line 70: this is async, but I don't see an await anywhere, is that intentional?

Monkeychip:
This is a question more for design, but I know in the new Secrets config Ivana had me replace falsey values with things like "Not set." I know TTLs inherit so maybe it's not applicable and it certainly isn't something we consistently do, but might be worth asking her what we should do going for with TTLs that are not set?  
image

ui/tests/integration/components/keymgmt/key-edit-test.js
```diff
@@ -0,0 +1,76 @@
...
+          creation_time: now,
+        },
+      ],
+    });
+    this.model = model;
+    this.tab = '';
+  });
+
+  test('it renders show view as default', async function (assert) {
+    // this.set('model', model);
```
Monkeychip, line 32: should this comment be kept or is it a leftover?

Monkeychip (approved):
Nice work!!! The setup sometimes feels like the biggest hurdle.

ui/app/adapters/keymgmt/key.js
hashishaw: Good catch, I'll remove!

ui/app/templates/components/keymgmt/key-edit.hbs
```diff
@@ -0,0 +1,179 @@
...
+
+{{#if (eq this.mode "show")}}
+  <div class="tabs-container box is-sideless is-fullwidth is-paddingless is-marginless" data-test-keymgmt-key-toolbar>
+    <nav class="tabs">
+      <ul>
+        <LinkTo
+          @route="vault.cluster.secrets.backend.show"
+          @model={{@model.id}}
+          {{! template-lint-disable no-unknown-arguments-for-builtin-components }}
+          @tagName="li"
```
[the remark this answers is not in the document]
hashishaw: Yeah, probably good to do in a separate ticket so we capture all the instances (I copied this code from another secret-edit component)

ui/tests/integration/components/keymgmt/key-edit-test.js
hashishaw: Right, it's not needed

hashishaw:
> This is a question more for design, but I know in the new Secrets config Ivana had me replace falsey values with things like "Not set." I know TTLs inherit so maybe it's not applicable and it certainly isn't something we consistently do, but might be worth asking her what we should do going for with TTLs that are not set?

I think this is a bigger discussion about what constitutes as "not set" -- sometimes zero is a value we want to show, sometimes we want it to be read as "no value". Maybe there's a setting in InfoTableRow (which is already bloated, tbh) where if the value matches the defaultValue, we show some other custom value? Either way, this is not something specific to KMSE so we can talk about it offline.

ui/app/templates/components/keymgmt/key-edit.hbs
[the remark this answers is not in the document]
Monkeychip: Will do!

PR #13642, opened 2022-01-13
UI/Attribution component updates

hellobontempo:
Adds csv export button and single chart view for when a date range is provided for client count data. Because _all_ clients are new in a given billing period having two separate charts (one for total client counts and one for only new counts) doesn't make sense when there is a date range provided. This breakdown is only available when the user had drilled down to a single month.      

Export button (still need designs for what the CSV file should look like once downloaded)
Screen Shot 2022-01-12 at 3 26 12 PM

### **Conditionals for single vs double charts**

_Sample data does not have auth methods included so differences are in chart view and copy. Some copy has not been finalized screen shots are to display accurate use of "Auth method" and "Namespace" and "month" vs "date range"_

**Namespace breakdown and date range**
`isDateRange=true`
`isAllNamespaces=true`
Screen Shot 2022-01-12 at 4 10 26 PM

**Namespace breakdown and single month**
`isDateRange=false` 
`isAllNamespaces=true`
Screen Shot 2022-01-12 at 4 26 59 PM

**Auth method breakdown and date range**
`isDateRange=true`
`isAllNamespaces=false` (auth method breakdown)
Screen Shot 2022-01-12 at 4 20 27 PM

**Auth method breakdown and single month**
`isDateRange=false`
`isAllNamespaces=false`
Screen Shot 2022-01-12 at 4 24 53 PM

### DIFF
--- a/ui/app/components/clients/attribution.js
+++ b/ui/app/components/clients/attribution.js
@@ -40,7 +40,7 @@ export default class Attribution extends Component {
   }
 
   get clientCountBreakdown() {
-    return this.isAllNamespaces ? 'Namespace path' : 'Auth method mount';
+    return this.isAllNamespaces ? 'Namespace' : 'Auth method';
   }
 
   get chartText() {
@@ -57,7 +57,7 @@ export default class Attribution extends Component {
     } else if (!this.isAllNamespaces) {
       return {
         description:
-          'This data shows the top ten authentication methods by client count within this namespace, and can be used to understand where new clients and total clients are originating. Authentication methods are organized by path.',
+          'This data shows the top ten authentication methods by client count within this namespace, and can be used to understand where clients are originating. Authentication methods are organized by path.',
         newCopy: `The new clients used by the auth method for this ${dateText}. This aids in understanding which auth methods create and use new clients 
         ${dateText === 'date range' ? ' over time.' : '.'}`,
         totalCopy: `The total clients used by the auth method for this ${dateText}. This number is useful for identifying overall usage volume. `,
--- a/ui/app/components/clients/attribution.js
+++ b/ui/app/components/clients/attribution.js
@@ -1,45 +1,105 @@
 import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
 
-// TODO: fill out below!!
 /**
  * @module Attribution
- * Attribution components are used to...
+ * Attribution components display the top 10 total client counts for namespaces or auth methods (mounts) during a billing period.
+ * If view is filtered for a single month, two graphs display and show a breakdown of new and total client counts by namespace or auth method, respectively 
  *
  * @example
  * ```js
- * <Attribution @requiredParam={requiredParam} @optionalParam={optionalParam} @param1={{param1}}>
- * Pass in export button
- * </Attribution>
+ * <Clients::Attribution
+ *    @newClientsData={{this.newClientData}}
+ *    @totalClientsData={{this.totalClientData}}
+ *    @chartLegend={{this.chartLegend}}
+ *    @isDateRange={{this.isDateRange}}
+ *    @isAllNamespaces={{this.isAllNamespaces}}
+    >
+      <button type="button"> Export attribution data </button>
+    </Clients::Attribution>
  * ```
- * @param {object} requiredParam - requiredParam is...
- * @param {string} [optionalParam] - optionalParam is...
- * @param {string} [param1=defaultValue] - param1 is...
+
+ * @param {array} newClientsData - (passed to child chart) must be an array of flattened objects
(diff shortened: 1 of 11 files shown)


ui/app/components/clients/attribution.js
```diff
@@ -1,45 +1,105 @@
...
+    </Clients::Attribution>
  * ```
- * @param {object} requiredParam - requiredParam is...
- * @param {string} [optionalParam] - optionalParam is...
- * @param {string} [param1=defaultValue] - param1 is...
+
+ * @param {array} newClientsData - (passed to child chart) must be an array of flattened objects
+ * @param {array} totalClientsData - (passed to child chart) must be an array of flattened objects
+ * @param {array} chartLegend - (passed to child) array of objects with key names 'key' and 'label' so data can be stacked
+ * @param {boolean} isDateRange - discerns if dataset from API is a date range or single month
```
Monkeychip, line 25: I like the label name. I'll keep note of this because I have the event handler wired up to handle when they click single month and this is the param I need to modify when they do.

ui/app/components/clients/attribution.js
```diff
@@ -1,45 +1,105 @@
...
+ * @param {string} activityDateRange - for modal and csv download to display correct date range
  */
 
 export default class Attribution extends Component {
-  get dateRange() {
-    // some conditional that returns "date range" or "month" depending on what the params are
-    return 'date range';
+  @tracked showCSVDownloadModal = false;
+
+  // TODO should this be a getter? It will be updated from parent component's filter bar
```
Monkeychip, line 33: NIT, any chance with TODO you can add your initials in front of them (e.g. CBS TODO or ARG TODO). There are some super old TODOs left over from previous developers, and this way we can show who should be following up with the TODO.

ui/app/components/clients/attribution.js
```diff
@@ -1,45 +1,105 @@
...
+      fields = [`${this.clientCountBreakdown}`, 'Active clients', 'Unique entities', 'Non-entity tokens'];
+    } else {
+      data = this.args.newClientsData;
+      // will need add a column for new clients
+      fields = [`${this.clientCountBreakdown}`, 'Active clients', 'Unique entities', 'Non-entity tokens'];
+    }
+
+    results = fields.join(',') + '\n';
+    data.forEach(function (item) {
+      // debugger
```
Monkeychip, line 89: just left over? maybe remove unless you need it later

ui/app/components/clients/attribution.js
hellobontempo: Oh my gosh!!! Absolutely!  This whole time I thought your were writing "arg" like "argument" so I left it out because in my vscode makes it nice and orange if I have just have `// TODO` 😆 😆 😆 

These todos will all be cleared up by the time the feature is released, but absolutely will do moving forward!

ui/app/components/clients/dashboard.js
```diff
@@ -24,6 +24,11 @@ export default class Dashboard extends Component {
   @tracked startYear = null;
   @tracked selectedNamespace = null;
 
+  // conditionals for filtering HBS file
+  // TODO update values depending on data returned from API
+  @tracked isDateRange = true; // false when viewing single month (true on init b/c defaults to billing period)
```
Monkeychip, line 29: we hope 🤞 . If it turns out we cannot reliable get the billing period from an endpoint then I assume (🤷‍♀️) this could be false? Basically, what should this value be if we don't have a billing period start date at all, and therefor no data, just an empty state telling them they need to select a start date.

ui/app/components/clients/attribution.js
Monkeychip: Ha! Someone once thought I had a pirate fetish thing or something. My initials aren't a great example :)

ui/app/components/clients/attribution.js
hellobontempo: I can remove! I left behind since this function will change quite a bit once data format is finalized

ui/mirage/handlers/activity.js
```diff
@@ -0,0 +1,686 @@
+export default function (server) {
```
Monkeychip, line 1: thanks for adding!

Monkeychip (approved):
Nice work! Tackled this really fast 🤙

ui/app/components/clients/dashboard.js
hellobontempo: Yeah - I wondered about this. None of the child components will show if there is no data / no date. (They will all have empty state elses if data isn't coming in properly). Based on your comment, I think it'd be better to just leave both of these tracked properties with no default and have them set based on what is inputed into the date selector.

PR #13643, opened 2022-01-13
Distinguish LIST-only paths in OpenAPI

### LINKED ISSUE #12283: Vault OpenAPI does not allow you to tell whether an endpoint supports only capability 'list', or both 'list' and 'read'

I am writing a tool to consume Vault's automatic OpenAPI (sys/internal/specs/openapi) for validating policies.

I have found that whilst the OpenAPI lets you know **almost** what capabilities each Vault path implements, there is an omission:

Support of the 'list' capability can be inferred when Vault renders this into a 'get' operation:
```
                "parameters": [
                    {
                        "name": "list",
                        "description": "Return a list if `true`",
                        "in": "query",
                        "schema": {
                            "type": "string"
                        }
                    }
                ],
```

However, there is no way to reliably tell whether a path supporting the 'list' capability also supports 'read'. Some do and some do not.

Represent whether a path supports only 'list', or 'list' and 'read' in the OpenAPI document. This could be done by, for list-only endpoints, marking the above-shown list parameter as '"required": true' and '"enum": ["true"]' - i.e. you must pass this parameter and the only valid value to pass is "true".

It would also be appropriate to change the description of the list parameter from "Return a list if `true`" to "Must be set to `true`" for list-only operations.

A smaller, but less machine-parseable change would be to just change the description as mentioned above.

digivava:
Fixes #12283 

Differentiates between paths that only support LIST (including GET w/ list=true query params), and those that support both LIST and plain GET.

According to my testing, at this point in time there are 77 List-only endpoints and 10 that support both.

### DIFF
--- a/changelog/13643.txt
+++ b/changelog/13643.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+sdk: Fixes OpenAPI to distinguish between paths that can do only List, or both List and Read.
+```
--- a/sdk/framework/openapi.go
+++ b/sdk/framework/openapi.go
@@ -128,13 +128,12 @@ type OASOperation struct {
 }
 
 type OASParameter struct {
-	Name        string        `json:"name"`
-	Description string        `json:"description,omitempty"`
-	In          string        `json:"in"`
-	Schema      *OASSchema    `json:"schema,omitempty"`
-	Required    bool          `json:"required,omitempty"`
-	Enum        []interface{} `json:"enum,omitempty"`
-	Deprecated  bool          `json:"deprecated,omitempty"`
+	Name        string     `json:"name"`
+	Description string     `json:"description,omitempty"`
+	In          string     `json:"in"`
+	Schema      *OASSchema `json:"schema,omitempty"`
+	Required    bool       `json:"required,omitempty"`
+	Deprecated  bool       `json:"deprecated,omitempty"`
 }
 
 type OASRequestBody struct {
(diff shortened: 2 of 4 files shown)


kalafut (commented):
LGTM, but I'll let one of the devs approve. You might want to generate our full openapi spec using this branch and pop it into a validator (e.g. I used to use https://editor.swagger.io/) just to make sure we're spec-compliant.

digivava:
> You might want to generate our full openapi spec using this branch and pop it into a validator (e.g. I used to use https://editor.swagger.io/) just to make sure we're spec-compliant.

Good catch there. It looked fine to me since it was valid json but now that I paste it in there, I see that it's giving me this for every List-only path: 
```
Structural error at paths./ad/library.get.parameters.0
should NOT have additional properties
additionalProperty: enum
```

The output json looks like this:
```
      parameters:
        - name: list
          description: Must be set to `true`
          in: query
          schema:
            type: string
          required: true
          enum:
            - 'true'
```

Should I perhaps remove the newly added enum field and just keep the required: true, since that's still programmatically useful enough to differentiate from a list+read endpoint?

kalafut:
Are you validating as OpenAPI 3 (I don't recall how that is set in the swagger tool)? Because I _think_ the enum you added looks ok, per https://swagger.io/docs/specification/data-models/enums/ ?

ncabatoff (approved):

PR #13650, opened 2022-01-13
Update archiver to v3.5.1

mickael-hc:
Addresses CVE-2020-16845 for its dependency, github.com/ulikunitz/xz

In this case, Vault is not be affected by this vulnerability, as Vault only _writes_ archives with sanitized data. However, we should update this dependency to appease vulnerability scanners.


mickael-hc:
closed via https://github.com/hashicorp/vault/pull/11584

Issue #13376, opened 2021-12-09
vault cli `-output-curl-string` does not include client cert info


bmendric:
**Environment:**
* Vault Version: 1.8.4, 1.8.5, 1.9.0
* Operating System/Architecture: Ubuntu 18.04 x86_64

**Expected Behavior:**

Performing the following command:
  `vault login -method=cert -address=https://localhost:1234 -client-cert=client.crt -client-key=client.key -output-curl-string`
should result in the curl command:
  `curl -X PUT -H "X-Vault-Request: true" -d '{"name":""}' --cert ./client.crt --key ./client.key https://localhost:1234/v1/auth/cert/login`

**Actual Behavior:**

Performing the following command:
  `vault login -method=cert -address=https://localhost:1234 -client-cert=client.crt -client-key=client.key -output-curl-string`
should result in the curl command:
  `curl -X PUT -H "X-Vault-Request: true" -d '{"name":""}' https://localhost:1234/v1/auth/cert/login`

**Steps to Reproduce:**

See above commands; only need validly PEM encoded crt and key files to test

**References:**

- https://www.vaultproject.io/api-docs/auth/cert#login-with-tls-certificate-method
- https://www.vaultproject.io/docs/auth/cert#via-the-cli
- https://www.vaultproject.io/docs/auth/cert#via-the-api

hsimon-hashicorp:
@bmendric This should be good to start, we'll let you know if we have questions. :)

remilapeyre:
Thanks for reporting this @bmendric, this was also affecting other commands like `vault read`, `vault write` and `vault delete`.

PR #13660, opened 2022-01-13
Add support for client certificates to -output-curl-string

remilapeyre:
I did not write tests for this feature as -output-curl-string was not
already tested and this is a simple change. Because the name of the
certificates would be lost once loaded I added fields to Config to keep
track of them. I did not add a public method for the user to set them
explicitely as I don't think anyone would need this functionnality
outside of the Vault CLI.

Closes https://github.com/hashicorp/vault/issues/13376

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -132,6 +132,12 @@ type Config struct {
 	// with the same client. Cloning a client will not clone this value.
 	OutputCurlString bool
 
+	// curlCACert, curlCAPath, curlClientCert and curlClientKey are used to keep
+	// track of the name of the TLS certs and keys when OutputCurlString is set.
+	// Cloning a client will also not clone those values.
+	curlCACert, curlCAPath        string
+	curlClientCert, curlClientKey string
+
 	// SRVLookup enables the client to lookup the host through DNS SRV lookup
 	SRVLookup bool
 
@@ -244,11 +250,15 @@ func (c *Config) ConfigureTLS(t *TLSConfig) error {
 			return err
 		}
 		foundClientCert = true
+		c.curlClientCert = t.ClientCert
+		c.curlClientKey = t.ClientKey
 	case t.ClientCert != "" || t.ClientKey != "":
 		return fmt.Errorf("both client cert and client key must be provided")
 	}
 
 	if t.CACert != "" || t.CAPath != "" {
+		c.curlCACert = t.CACert
+		c.curlCAPath = t.CAPath
 		rootConfig := &rootcerts.Config{
 			CAFile: t.CACert,
 			CAPath: t.CAPath,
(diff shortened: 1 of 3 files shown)


api/client.go
```diff
@@ -244,11 +250,15 @@ func (c *Config) ConfigureTLS(t *TLSConfig) error {
 			return err
 		}
 		foundClientCert = true
```
ncabatoff, line 252: Do we need to grab the modifyLock?

api/output_string.go
```diff
@@ -46,6 +48,22 @@ func (d *OutputStringError) parseRequest() {
 	if d.Request.Method != "GET" {
 		d.parsedCurlString = fmt.Sprintf("%s-X %s ", d.parsedCurlString, d.Request.Method)
 	}
+	if d.ClientCACert != "" {
+		clientCACert := strings.Replace(d.ClientCACert, "'", "'\"'\"'", -1)
```
ncabatoff, line 52:
```suggestion
		clientCACert := strings.Replace(d.ClientCACert, "'", `'"'"'`, -1)
```

I'm still not sure I understand it even with that change.

api/client.go
remilapeyre: Yes, I think this is needed indeed. Since `ConfigureTLS` is already called with the lock held in `ReadEnvironment` I made a private function `configureTLS` that can be called when the lock is already held. `ConfigureTLS` now just takes the lock and calls it.

api/output_string.go
remilapeyre: This is so that if the filename contains a special character:

```
$ ls
clie'nt.crt client.key
```
then the `'` is escaped in the output of `-output-curl-string`:
```
$ ../bin/vault read -client-cert clie\'nt.crt -client-key client.key -output-curl-string foo
curl --cert 'clie'"'"'nt.crt' --key 'client.key' -H "X-Vault-Request: true" -H "X-Vault-Token: $(vault print token)" https://127.0.0.1:8200/v1/foo
```

The `--cert` option is the concatenation of three strings `'clie'`, `"'"` and `nt.crt` so it gives `"clie'nt.crt"`.

For what it's worth this is very naive, it does not escape `$(...)` or ` `` ` which would also interpreted by the shell. I don't think this is a solvable problem thought as different shells will interpret different constructs. I used this method to escape `'` because it is already used for the body.

HridoyRoy (approved):
Lgtm! 

I've done some basic testing to verify that the cert and keys are included now in the output-curl-string. I also looked through the callers of ConfigureTLS to double-check if grabbing the modifyLock in the function could cause issues, but it all looked good to me. 

Thanks for submitting this PR!

Issue #9636, opened 2020-07-30
Token revocation via accessor is not idempotent operation


VladRassokhin:
**Describe the bug**
Second subsequent token revocation using accessor fails, while when multiple attempts to revoke token itself succeeds with message "Success! Revoked token (if it existed)"

**To Reproduce**
1. Start dev server (`vault server -dev`), navigate to second shell with (`export VAULT_ADDR='http://127.0.0.1:8200'`)
1. Run `vault token create`, remember token
1. Run `vault token revoke $token`, observe `"Success! Revoked token (if it existed)"`
1. Run once more `vault token revoke $token`, observe `"Success! Revoked token (if it existed)"` again, while token is already revoked (confirm in server log) (that's expected).
1. Run `valut token create`, remember accessor
1. Run `vault token revoke -accessor $accessor`, observe `"Success! Revoked token (if it existed)"`
1. Run once more `vault token revoke -accessor $accessor`, observe error:
>Error revoking token: Error making API request.
>
>URL: POST http://127.0.0.1:8200/v1/auth/token/revoke-accessor
>Code: 400. Errors:
>
>* 1 error occurred:
>	* invalid accessor

**Expected behavior**
Expected `"Success! Revoked token (if it existed)"` for last request and any subsequent ones

**Environment:**
* Vault Server Version: 1.4.3
* Vault CLI Version: 1.4.3
* Server Operating System/Architecture: Mac OS X 10.14.6

Vault server configuration file(s):
None, dev server

**Additional context:**
I've distributed system (TeamCity CI) which revokes token from different places: as `token remoke $token` on agents side and `token revoke -accessor $accessor` on server side (just to make sure it's revoked). 
Server does not possess token (using cubbyhole). 
Offtopic is that to revoke first token I've to login again and obtain second token which is used purely to revoke first token via accessor. I'd like to see /auth/token/revoke-accessor to be authless (if someone knows accessor it should be enough to revoke it), like in gpg key revocation.

PR #13661, opened 2022-01-13
Make auth/token/revoke-accessor idempotent

remilapeyre:
The auth/token/revoke will not error out if the token does not exists, it
always tries to revoke the token and return success to the client whether
or not the token exists. This makes the behavior of
auth/token/revoke-accessor coherent with this and remove the need to
check whether the token still exists.

The error had been added by https://github.com/hashicorp/vault/pull/2391 
which added a lot of error reporting across the code so this might just
have been overlooked.

Closes https://github.com/hashicorp/vault/issues/9636

### DIFF
--- a/vault/token_store.go
+++ b/vault/token_store.go
@@ -747,6 +747,10 @@ func (ts *TokenStore) tokenStoreAccessorList(ctx context.Context, req *logical.R
 			resp.AddWarning(fmt.Sprintf("Found an accessor entry that could not be successfully decoded; associated error is %q", err.Error()))
 			continue
 		}
+		if aEntry == nil {
+			resp.AddWarning("Found an accessor entry that could not be successfully decoded: invalid accessor")
+			continue
+		}
 
 		if aEntry.TokenID == "" {
 			resp.AddWarning(fmt.Sprintf("Found an accessor entry missing a token: %v", aEntry.AccessorID))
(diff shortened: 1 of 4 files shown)


hsimon-hashicorp:
Hi @remilapeyre! Thanks a lot for this contribution - please don't forget to include a changelog entry. :)

vault/token_store.go
```diff
@@ -2190,6 +2204,9 @@ func (ts *TokenStore) handleUpdateRevokeAccessor(ctx context.Context, req *logic
 	if err != nil {
 		return nil, err
 	}
+	if aEntry == nil {
```
ncabatoff, line 2207: Worth adding a warning to the response in this case, since we're changing the behaviour?

vault/token_store.go
```diff
@@ -747,6 +747,10 @@ func (ts *TokenStore) tokenStoreAccessorList(ctx context.Context, req *logical.R
 			resp.AddWarning(fmt.Sprintf("Found an accessor entry that could not be successfully decoded; associated error is %q", err.Error()))
 			continue
 		}
+		if aEntry == nil {
+			resp.AddWarning("Found an accessor entry that could not be successfully decoded: invalid accessor")
```
ncabatoff, line 751: I don't think we need this warning.  There's already the one below that occurs when "entry missing a token".  The warning about "can't successfully decode", which you're trying to maintain here, was I think inappropriate for the case you're changing, where there's not actually an accessorEntry in storage.  In between our list and our lookup the token may have been deleted, which is perfectly normal and doesn't warrant a warning IMO.

remilapeyre:
Hi @ncabatoff, thanks for the review! Everything should be fine now.

ncabatoff (approved):

ncabatoff:
Thanks @remilapeyre !

PR #13690, opened 2022-01-18
Add validation for nonce size when we aren't in convergent encryption mode within transit backend

divyapola5:
The backend will panic when provided with a nonce value that is an invalid size when the key type is one of AES256_GCM96, AES128_GCM96, ChaCha20_Poly1305 within transit when the key does not have convergent encryption enabled. 

Within sdk/helper/keysutil/policy.go func SymmetricEncryptRaw we are missing a test for the nonce size when we aren't in convergent encryption mode. We only check the nonce size is 0 to generate a value.

See VAULT-4533 for more details.

### DIFF
--- a/builtin/logical/transit/path_encrypt_test.go
+++ b/builtin/logical/transit/path_encrypt_test.go
@@ -578,6 +578,31 @@ func TestTransit_BatchEncryptionCase12(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


stevendpclark (approved):
LGTM!

sgmiller (approved):

PR #13691, opened 2022-01-18
Time-based transit key autorotation

schultz-is:
# Summary
This introduces automatic, time-based key rotation to the transit secrets engine. A new field is added to the `keysutil.Policy` that keeps track of an interval for key rotation. The value can be specified at key creation time or via a key config update. The value has a minimum of 1 hour and keys are checked for rotation about once an hour.

# Testing
There are three unit test suites included with this functionality:
- `TestTransit_CreateKeyWithAutorotation` in `builtin/logical/transit/path_keys_test.go` which tests the creation of a key with the new `auto_rotate_interval` request field.
- `TestTransit_UpdateKeyConfigWithAutorotation` in `builtin/logical/transit/path_config_test.go` which tests the update of an existing key with the new `auto_rotate_interval` request field.
- `TestTransit_AutoRotateKeys` in `builtin/logical/transit/backend_test.go` which tests the actual automatic rotation of keys.

### DIFF
--- a/builtin/logical/transit/backend.go
+++ b/builtin/logical/transit/backend.go
@@ -101,6 +101,7 @@ type backend struct {
 	configMutex          sync.RWMutex
 	cacheSizeChanged     bool
 	checkAutoRotateAfter time.Time
+	autoRotateOnce       sync.Once
 }
 
 func GetCacheSizeFromStorage(ctx context.Context, s logical.Storage) (int, error) {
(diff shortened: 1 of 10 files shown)


sgmiller (commented):
Looks good to me but I'd like one more person's eyes on it.

sgmiller (approved):

builtin/logical/transit/backend.go
```diff
@@ -162,3 +168,79 @@ func (b *backend) invalidate(ctx context.Context, key string) {
...
+		if err != nil {
+			errs = multierror.Append(errs, err)
+			continue
+		}
+
+		// If the policy is nil, move onto the next one.
+		if p == nil {
+			continue
+		}
+
```
stevendpclark, line 224: We are missing the policy locking here that does exist within the path_rotate.go::pathRotateWrite . Feels like we should be grabbing it here as well.

schultz-is: Excellent catch! I'll add that locking in for nodes with caching enabled.

schultz-is: Added cache checks a la `path_rotate.go`.

stevendpclark (approved):
Looks good!

sgmiller (approved):

schultz-is:
Thanks y’all!

builtin/logical/transit/backend.go
```diff
@@ -162,3 +169,91 @@ func (b *backend) invalidate(ctx context.Context, key string) {
...
+		latestKey := p.Keys[strconv.Itoa(p.LatestVersion)]
+		if time.Now().After(latestKey.CreationTime.Add(p.AutoRotateInterval)) {
+			if b.Logger().IsDebug() {
+				b.Logger().Debug("automatically rotating key", "key", key)
+			}
+			if !b.System().CachingDisabled() {
+				p.Lock(true)
+			}
+			err = p.Rotate(ctx, req.Storage, b.GetRandomReader())
+			p.Unlock()
```
briankassouf, line 250: Would this panic if caching is disabled? Unlocking a not locked mutex, i believe, panics

PR #13693, opened 2022-01-18
Depend explicitly on go-msgpack v1.1.5

ncabatoff:


calvn (approved):

physical/raft/msgpack.go
```diff
@@ -0,0 +1,11 @@
+package raft
+
+import (
+	"github.com/hashicorp/go-msgpack/codec"
```
calvn, line 4: Does it complain if you change it to something like this without referencing anything from the package?
```suggestion
	_ "github.com/hashicorp/go-msgpack/codec"
```

ncabatoff: Nope, good idea.

PR #13698, opened 2022-01-19
Backport of auth/kubernetes: support for dynamically reloading short-lived tokens into release/1.9.x

### LINKED ISSUE #12855: Kubernetes Auth Method documentation not compatible with Kubernetes v1.21+
## Describe the bug

Kubernetes v1.21 introduced new bound service account tokens with following new properties:

1. Tokens are bound to a pod and they get revoked when pod is deleted.
2. Tokens will expire.
3. Tokens are audience-limited.

The Kubernetes Auth Method documentation does not cover the use of the new tokens. This results in failed authentication.

## To Reproduce

Pre-condition:

Deploy Vault inside Kubernetes. Configure Kubernetes Auth Method according to examples at https://www.vaultproject.io/docs/auth/kubernetes. Set JWT value to `token_reviewer_jwt="$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)"`.

**(1) Authentication fails because of invalidated token**

1. Delete the pod that runs Vault and wait for it to be recreated.
2. Try to authenticate using Kubernetes Auth Method
3. Vault returns `403 Forbidden` with `{ "errors": [ "permission denied" ] }` in response body.

Vault prints following error:

```
2021-10-18T12:09:01.075Z [ERROR] auth.kubernetes.auth_kubernetes_726dbb64: login unauthorized due to: lookup failed: service account unauthorized; this could mean it has been deleted or recreated with a new token
```

Kubernetes API server prints following error

```
E1018 12:09:01.074341       1 authentication.go:63] "Unable to authenticate the request" err="[invalid bearer token, Token has been invalidated]"
```

**(2) Authentication fails because of expired token**

1. Configure Kubernetes cluster with a short expiration time by setting API server parameter `service-account-max-token-expiration` to `1h`. Wait for 1 hour for the token to expire .
2. Try to authenticate using Kubernetes Auth Method. 
3. Vault returns `403 Forbidden` with `{ "errors": [ "permission denied" ] }` in response body.

Vault prints following error:

```
2021-10-18T11:47:23.756Z [ERROR] auth.kubernetes.auth_kubernetes_97f02f17: login unauthorized due to: lookup failed: service account unauthorized; this could mean it has been deleted or recreated with a new token
```

Kubernetes API server prints following error

```
E1018 11:47:23.755730       1 authentication.go:63] "Unable to authenticate the request" err="[invalid bearer token, Token has expired.]"
```

## Proposed fix

Include following PR
* https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/122

Update documentation, taking following points into account:

* It is recommended NOT to store Kubernetes service account token persistently into config when running Vault inside Kubernetes. Instead, the recommendation is to use the local token and CA files. 
* When running Vault outside of Kubernetes, user needs to use legacy token which is not bound to any pod and does not expire. User should be instructed how to access legacy token via Secret directly. User should be WARNED that the legacy tokens could be removed in later Kubernetes release completely.

I also think following point would be beneficial

* Recommend `disable_iss_validation=True`. The benefit is that user does not need to find the correct `iss` from their specific cluster (or change it if it changes later). Kubernetes Token Review API will validate the token, so I do not see negative side in recommending that.

hc-github-team-secure-vault-core:
## Backport

This PR is auto-generated from #13595 to be assessed for backporting due to the inclusion of the label backport/1.9.x.

---

This change is continuation to https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/122 where support was added for dynamically reloading Kubernetes short-lived tokens for better Kubernetes 1.21+ compatibility.  This change takes the feature into use in Vault.

* Uplift new version of Kubernetes auth plugin that does not store the  service account token persistently to Vault storage.
* Update the documentation to recommend local token again when running  Vault inside cluster.

Fixes #12855


tvoran:
Just waiting on [backporting to the kubernetes plugin's release branch](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/131), then I'll update this one.

tvoran:
Ok, I think this one is ready to go.

tomhjp (approved):

sudermanjr:
Is there a way to use this currently via publicly-available images? I seem to see it should be available when 1.9.3 is released, but I got myself into a bit of a bind and could definitely use the fix sooner if there's an image available. I would also be very happy to test out a release candidate if one becomes available.

tvoran:
Hi @sudermanjr, unfortunately I don't think we're publishing any pre-release images yet. And yes, this change will be included in the 1.9.3 release, which should come out soon.

PR #13699, opened 2022-01-19
OSS side of the managed key interfaces

sgmiller:


victorr (approved):
👍

managed_key/config_test.go
```diff
@@ -0,0 +1,47 @@
+//go:build hsm
```
stevendpclark, line 1: Not sure why this build tag exists within OSS? With this the test will never run within this build so maybe the test doesn't belong within OSS or we need to remove this build tag?

PR #13702, opened 2022-01-19
vault: fix dropped barrier test error

alrs:
This fixes a dropped `err` variable in the `vault` package.

Could a Hashicorp employee add the `no-changelog` label?


HridoyRoy (commented):
Hi @alrs , thanks for the contribution! The PR looks good to me, but it seems like there was a CI glitch that caused some tests to fail. The failed tests seem to be passing for me locally. 

Could you please re-run the CI job by pushing an empty commit? I'd happy to approve once the tests pass.

PR #13695, opened 2022-01-18
Backport of Update http requests API link to versioned docs into release/1.9.x

hc-github-team-secure-vault-core:
## Backport

This PR is auto-generated from #13692 to be assessed for backporting due to the inclusion of the label backport/website.

**WARNING** automatic cherry-pick of commits failed. Commits will require human attention.

The below text is copied from the body of the original PR.

---

In the "latest" version of the API docs, the HTTP requests endpoint has been removed. Updating to reference the 1.8.x docs.

PR #13713, opened 2022-01-19
Backport of docs: add known issues section to 1.9.x upgrade guide into release/1.9.x

hc-github-team-secure-vault-core:
## Backport

This PR is auto-generated from #13662 to be assessed for backporting due to the inclusion of the label backport/website.

---

Adds a "Known Issues" section to our 1.9.x upgrade guide, and includes a note about a panic that can be encountered on Vault installations that are using the ID token backend and have named keys created.


calvn:
@mladlow the upgrade guide file got renamed, leading to failures on any backports that involve this file.

Should also close https://github.com/hashicorp/vault/pull/13695

mladlow:
@calvn I think you'll need to update the docs-nav-data.json file too? But I'm worried about the rest of the upgrade guides on this branch now, which I think are all the .0 versions.

calvn:
Actually this is not quite right -- I would also need to update the references on the navbar and remove the 1.9.0 upgrade guide page. @mladlow do you have any advice on how to properly backport that entire file, include the changes from https://github.com/hashicorp/vault/pull/13370?

mladlow:
Manually? I think? We might not be able to cherry pick them but my git-skills are also definitely not top-notch. I would do it manually because that would be the fastest one-time solution for me, but it might be worth trying to find out if there is a better way.

website/content/docs/concepts/identity.mdx
```diff
@@ -20,11 +20,13 @@ Each user may have multiple accounts with various identity providers, and Vault
...
 tie authentications from various auth methods to a single representation. This representation of a consolidated identity is called an **Entity** and their
 corresponding accounts with authentication providers can be mapped as
-**Aliases**. In essence, each entity is made up of zero or more aliases. 
+**Aliases**. In essence, each entity is made up of zero or more aliases. An entity cannot have more than one alias for
+a particular authentication backend.
 
 For example, a user with accounts in both GitHub and LDAP can be mapped to a
 single entity in Vault with two aliases, one of type GitHub and one of type
-LDAP.
+LDAP. Note however, if both aliases are created on the same auth mount, say
```
mladlow, line 28: This file has different content in `main` (more words and extra images) - is it worth getting those into the release branch?

website/content/docs/upgrading/upgrade-to-1.9.x.mdx
```diff
@@ -0,0 +1,90 @@
...
+  This page contains the list of deprecations and important or breaking changes
+  for Vault 1.9.x. Please read it carefully.
+---
+
+# Overview
+
+This page contains the list of deprecations and important or breaking changes
+for Vault 1.9.x compared to 1.8. Please read it carefully.
+
+## Identity Tokens
```
mladlow, line 14: This file also has different content in `main`.

mladlow (commented):
I'm not sure whether you want the newer content in identity.mdx and the 1.9 upgrade guide from main?

website/content/docs/upgrading/upgrade-to-1.9.x.mdx
calvn: Thanks for catching that, I might have messed up during merge conflict... I'll fix this up.

website/content/docs/concepts/identity.mdx
calvn: I'm not sure if those changes should be backported or whether they are already inflight as backport PRs elsewhere. I've kept the changes from other files strictly the same as what's in https://github.com/hashicorp/vault/pull/13370/files.

website/content/docs/upgrading/upgrade-to-1.9.x.mdx
calvn: Updated, should be good now!

mladlow (approved):
Thanks for the extra digging!

So one problem is that I'm still not sure how to get the CLA to realize that this account has signed it...

calvn:
I'm pretty sure it's due to how backport-assistant makes that no-op commit. Linked you the source offline.

website/data/docs-nav-data.json
```diff
@@ -1437,17 +1437,17 @@
         "path": "upgrading/plugins"
       },
       {
-        "title": "Upgrade to 1.9.0",
-        "path": "upgrading/upgrade-to-1.9.0",
+        "title": "Upgrade to 1.9.x",
+        "path": "upgrading/upgrade-to-1.9.x",
         "hidden": true
```
calvn, line 1442: @mladlow I wonder if we should remove this attribute. Otherwise, the docs might not show when 1.9.x is versioned.

calvn: I might leave it out of this PR since I want to keep the result from the diff the same from the resulting 3 cherry-picks that I did.

PR #13714, opened 2022-01-19
Revert #13679 

ncabatoff:
Reason: TestCluster_ListenForRequest now fails intermittently in alarming ways, e.g. https://app.circleci.com/pipelines/github/hashicorp/vault/25138/workflows/58ce7db1-3f32-4455-8258-fe2ba19e2e1b/jobs/368209

### DIFF
--- a/vault/cluster/cluster.go
+++ b/vault/cluster/cluster.go
@@ -6,6 +6,8 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/vault/sdk/helper/certutil"
+	"github.com/hashicorp/vault/sdk/helper/tlsutil"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


vault/cluster/cluster.go
```diff
@@ -6,6 +6,8 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/vault/sdk/helper/certutil"
```
hghaf099, line 9: would you want to sort the order of the imports? otherwise, LGTM

hghaf099 (approved):

vault/cluster/cluster.go
ncabatoff: It was a straight revert, i.e. this was how it was before.  I plan to come back to this fix and try again, can address the imports then.

Issue #11803, opened 2021-06-09
request.tokenentry.meta should be accessible in role.template during oidc token creation


dlamotte:
**Is your feature request related to a problem? Please describe.**

Multiple secret-id's per approle could be more useful if a secret-id's meta was exposed so that JWTs could encode details from the secret-id meta.

**Describe the solution you'd like**

I'd like to hand a service an role-id+secret-id that was minted with a specific metadata field such that future requests to vault with the token could have secret-id specific authorization.

 that could be enforced by applications based on the contents of the JWT.  ie: JWT with field `scope: /objects/subset1` which ensures clients with the JWT cannot modify all `/objects` but only objects below `subset1` (the rules here are opaque and application dependent).

There are two aspects here:
- secret-id.meta in policy (a specific approle/secret-id combo gets a unique policy to it which could limit access within vault)
- secret-id.meta in JWTs via an oidc role template accessing secret-id.meta (services receiving the JWT could authorize behavior based on the secret-id.meta)

The initial attempt here utilized querying into entity aliases in templates like `identity.entity.aliases[$accessor].metadata[$secret-id-field]` but this was found to work only intermittently (see https://github.com/hashicorp/vault/issues/11798).

**Describe alternatives you've considered**

It seems the alternative is creating an approle per scope with a policy/role.template associated specifically with it.

The issue here is that approle/secret-id scale 1:1 with each other here vs the ideal solution is a single approle with a secret-id per variation.

**Explain any additional use-cases**

Any use cases where applications authenticating to vault with secret-id's that have some limited scope within an approle.

**Additional context**

Terraform to setup a `vault server -dev`:

```
resource "vault_identity_oidc" "server" {
  issuer = "http://localhost:8200"
}

resource "vault_identity_oidc_key" "key" {
  name      = "key"
  algorithm = "ES256"
}

resource "vault_identity_oidc_role" "role" {
  name = "role"
  key  = vault_identity_oidc_key.key.name
  ttl  = 86400

  template = <<EOF
{
  "groups": {{identity.entity.groups.names}},
  "nbf": {{time.now}},
  "alias.metadata.hello": {{identity.entity.aliases.${vault_auth_backend.app.accessor}.metadata.hello}},
  "entity_data": {
    "id": {{identity.entity.id}},
    "name": {{identity.entity.name}}
  }
}
EOF
}

resource "vault_identity_oidc_key_allowed_client_id" "role" {
  key_name          = vault_identity_oidc_key.key.name
[...]

PR #13715, opened 2022-01-19
Support token.metadata.<key> in ACL templates

remilapeyre:
We have a Vault Enterprise client that is using Sentinel for two
applications. What they need to do with Sentinel is quite simple and
could be done much more easily by using ACL templates if they supported
injecting a token's metadata. That would be more readable than a large
Sentinel script and make it much easier to onboard operators that need
to make changes to those applications.

### DIFF
--- a/changelog/13715.txt
+++ b/changelog/13715.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+acl: The ACL templating system now supports referencing a token metadata attribute
+with `{{ token.metadata.<attribute key> }}`.
+```
--- a/sdk/helper/identitytpl/templating.go
+++ b/sdk/helper/identitytpl/templating.go
@@ -29,6 +29,7 @@ type PopulateStringInput struct {
 	ValidityCheckOnly bool
 	Entity            *logical.Entity
 	Groups            []*logical.Group
+	TokenEntry        *logical.TokenEntry
 	NamespaceID       string
 	Mode              int       // processing mode, ACLTemplate or JSONTemplating
 	Now               time.Time // optional, defaults to current time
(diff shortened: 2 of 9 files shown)


ncabatoff:
Hi Remi,

We've had to say "no" to template-token-metadata requests many times I'm afraid, most recently in #10682.  Excerpted:

> We view this change as dangerous, since a user can create child tokens with whatever arbitrary metadata they like. This means policy templating using token metadata is vulnerable to any attacker who's aware of what policies are in use. I expect that you could come up with safeguards in your environment to protect against this (e.g. not granting access to the create token endpoints), but we as vault devs have to be vigilant against introducing features that allow for insecure by default behaviour.

Can you use identity metadata instead for your use case?

remilapeyre:
> We view this change as dangerous, since a user can create child tokens with whatever arbitrary metadata they like.

Wow, indeed I completely missed this. I suppose implementing this feature like this is a no go then.

> Can you use identity metadata instead for your use case?

In our case we can't use entities because of the way Vault Enterprise is priced, doing this would add thousands of entities.

Perhaps their could be some way to do this by adding a new field to `TokenEntry` that would not suffer the same issue and be inherited when creating a child token, let's call it `attributes` as an example.

Just like you need to be root to set `orphan` on `auth/token/create`, you would need to be `root` to set this new attribute.
This new attribute would always be copied verbatim to child tokens, only `root` could overwrite it.

In `auth/token/roles/:role_name` there could be a new `allow_custom_attributes` that would make it possible for a non-root application to generate tokens with the data it needs.

This could make this feature work, `root` would be able to create tokens with whatever `attributes` it wants, normal users would not be able to add or remove entries in `attributes` by creating child token as it would always be copied and applications needing special rights to generate token with custom `attributes` could do it by using a token role.

It seems to me that this could be secure, after all an application that can make changes to `auth/token/roles/:role_name` can already get full access to Vault by choosing the permission they get and generate root tokens.

If you think that it's worth it to try something like that I can propose a better patch to explore things out and fix the shortcomings of this one, otherwise we can close this PR.

UXabre:
Sorry for replying on an older PR but I was browsing to see if others came up with new insights in the subject.

I'll go ahead and link this [topic](https://github.com/hashicorp/vault/issues/11803#issuecomment-1006921960) as I think it offers a similar solution as what you describe. TL;DR just as you propose, we need to "prove" the data to be authorative and not "arbitrary".

mladlow:
Hey @remilapeyre ,

We discussed this a little more internally. Identity metadata should allow users to template ACL policies while using a unique identity for each application, service, or user that needs to authenticate to a cluster. At this time we don't want to add a new mechanism for ACL templating.

celesteking:
> Wow, indeed I completely missed this. I suppose implementing this feature like this is a no go then.

It's not your fault, it's Hashicorp's and they don't want to admit the mistake.

PR #13766, opened 2022-01-24
Add sys/version-history endpoint and associated command

ccapurso:
Tracking version history was added in Vault 1.9.0. Version entries are stored in entries keyed by `core/versions/<version-string>` (e.g. `/core/versions/1.9.0`). The entries themselves contain the installation timestamp of the version. Originally, the timestamps stored were created using local time (i.e. `time.Now()`). UTC is preferred and thus the logic has been modified to store new entries using UTC (i.e. `time.Now().UTC()`). Consistency is achieved by adding self-heal logic upon the initial read to cache existing entries in `Core.versionTimestamps`. If an existing timestamp's `time.Location` is not `time.UTC` then it is modified and rewritten.

The `Core.storeVersionTimestamp` method originally checked for the existence of an entry for the provided `version`. If one existed, a write to storage would not occur. The has been modified to take a `force` argument in order to facilitate the self-heal process.

Note: all of the `core/versions` logic used to reside in `core_util_common.go`. This PR has moved it to `version_store.go`.

Users currently only have the ability to determine the current version of the running Vault. This PR introduces a new authenticated `LIST` endpoint, `sys/version-history`, that will return the version history of a Vault cluster. The `keys` array will contain the versions in chronological order. `key_info` will include entries indexed by the version strings found in `keys` with additional information including the `previous_version` and `timestamp_installed`.

Example response:

```
{
  "keys": [
    "1.9.0",
    "1.9.1",
    "1.9.2"
  ],
  "key_info": {
    "1.9.0": {
      "previous_version": null,
      "timestamp_installed": "2021-11-18T10:23:16Z"
    },
    "1.9.1": {
      "previous_version": "1.9.0",
      "timestamp_installed": "2021-12-13T11:09:52Z"
    },
    "1.9.2": {
      "previous_version": "1.9.1",
      "timestamp_installed": "2021-12-23T10:56:37Z"
    }
  }
}
```

An associated `vault version-history` CLI command has been added. By default, it will print table output as follows. The raw JSON response can be printed using `-format=json`.
```
Note: Version tracking was added in 1.9.0. Earlier versions have not been tracked.

Version   Installation Time
-------   -----------------
1.9.0     2021-11-18T10:23:16Z
1.9.1     2022-12-13T11:09:52Z
1.9.2     2021-12-23T10:56:37Z
```


command/version_history.go
```diff
@@ -0,0 +1,127 @@
...
+	*BaseCommand
+}
+
+func (c *VersionHistoryCommand) Synopsis() string {
+	return "Prints the version history of the target Vault server"
+}
+
+func (c *VersionHistoryCommand) Help() string {
+	helpText := `
+Usage: vault version
```
swayne275, line 27:
```suggestion
Usage: vault version-history
```

is it supposed to be version-history?

ccapurso: Ah, yes! Thank you for catching that.

command/version_history.go
```diff
@@ -0,0 +1,127 @@
...
+
+	client, err := c.Client()
+	if err != nil {
+		c.UI.Error(err.Error())
+		return 2
+	}
+
+	resp, err := client.Logical().List("sys/version-history")
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("Error reading version history: %s", err))
```
ncabatoff, line 68: Should we also mention that it was only added in 1.10 here?   We could also use the seal-status endpoint to detect the server version and not bother with the version-history call if the server is <1.10, but I'm not sure I want to set that precedent.

http/sys_version_history_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/hashicorp/vault/sdk/version"
+	"github.com/hashicorp/vault/vault"
+)
+
+// TestSysVersionHistory_List tests the sys/version-history endpoint. Requests
+// to the endpoint must be authenticated. Without synthetically altering the
+// underlying core/versions storage entries, a single version entry should
+// exist.
+func TestSysVersionHistory_List(t *testing.T) {
```
ncabatoff, line 14: Typically we put only tests in the `http` package when they're exercising code that lives in that package, i.e. "non-logical" endpoints.

website/content/api-docs/system/version-history.mdx
```diff
@@ -0,0 +1,54 @@
+---
+layout: api
+page_title: /sys/version-history - HTTP API
+description: The `/sys/version-history` endpoint is used to retrieve the version history of a Vault.
+---
+
+# `/sys/version-history`
+
+~> **NOTE**: Tracking Vault version history was added in Vault version 1.9.0. Entries for versions installed prior to version 1.9.0 will not be available via this API.
```
ncabatoff, line 9:
```suggestion
~> **NOTE**: Tracking Vault version history was added in Vault version 1.9.0. Entries for versions installed prior to version 1.9.0 will not be available via this API.
This API was added in 1.10.0.
```

Although we don't normally say what version an endpoint was introduced in, I fear that if we don't in this case, people will read the above as "sys/version-history appeared in 1.9.0".

ncabatoff (approved):

vault/version_store_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/vault/sdk/version"
+)
+
+// TestVersionStore_StoreMultipleVaultVersions writes multiple versions of 1.9.0 and verifies that only
```
swayne275, line 11: love the descriptions!

swayne275 (approved):
a few comments but looks good!

vault/version_store.go
```diff
@@ -11,45 +11,68 @@ import (
...
-// storeVersionTimestamp will store the version and timestamp pair to storage only if no entry
-// for that version already exists in storage.
-func (c *Core) storeVersionTimestamp(ctx context.Context, version string, currentTime time.Time) (bool, error) {
-	timeStamp, err := c.barrier.Get(ctx, vaultVersionPath+version)
+// storeVersionTimestamp will store the version and timestamp pair to storage
+// only if no entry for that version already exists in storage. Version
+// timestamps were initially stored in local time. UTC should be used. Existing
+// entries can be overwritten via the force flag. A bool will be returned
+// denoting whether the entry was updated
+func (c *Core) storeVersionTimestamp(ctx context.Context, version string, currentTime time.Time, force bool) (bool, error) {
```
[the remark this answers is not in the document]
ccapurso: Yup, that's a great point. I'll keep the calls to `UTC()` in the existing calling code as it doesn't hurt and provides a the example that `time.Now().UTC()` is preferred over `time.Now()`.

vault/version_store.go
```diff
@@ -83,7 +107,29 @@ func (c *Core) loadVersionTimestamps(ctx context.Context) (retErr error) {
...
+
+		timestampInstalled := vaultVersion.TimestampInstalled
+
+		// self-heal entries that were not stored in UTC
+		if timestampInstalled.Location() != time.UTC {
+			timestampInstalled = timestampInstalled.UTC()
+			isUpdated, err := c.storeVersionTimestamp(ctx, vaultVersion.Version, timestampInstalled, true)
+
+			if err != nil {
+				c.logger.Warn("error rewriting vault version timestamp as UTC: %w", err)
```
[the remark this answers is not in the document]
ccapurso: Good catch, no it does not!

vault/version_store_test.go
ccapurso: Thanks :)

vault/version_store.go
```diff
@@ -83,7 +107,29 @@ func (c *Core) loadVersionTimestamps(ctx context.Context) (retErr error) {
...
+			}
+
+			if isUpdated {
+				c.logger.Info("Self-healed pre-existing vault version in UTC",
+					"vault version", vaultVersion.Version, "UTC time", timestampInstalled)
+			}
+
+			if err != nil {
```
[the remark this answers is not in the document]
ccapurso: Ah, no I don't that is not possible. This check should be removed. Thanks for catching that.

http/sys_version_history_test.go
ccapurso: I was trying to follow the example of testing other sys endpoints:

```
❯ find http -type f -name sys*_test.go
http/sys_rekey_test.go
http/sys_init_test.go
http/sys_version_history_test.go
http/sys_rotate_test.go
http/sys_generate_root_test.go
http/sys_in_flight_requests_test.go
http/sys_lease_test.go
http/sys_policy_test.go
http/sys_monitor_test.go
http/sys_auth_test.go
http/sys_mounts_test.go
http/sys_config_cors_test.go
http/sys_hostinfo_test.go
http/sys_health_test.go
http/sys_metrics_test.go
http/sys_mount_test.go
http/sys_internal_test.go
http/sys_seal_test.go
http/sys_audit_test.go
http/sys_config_state_test.go
http/sys_leader_test.go
http/sys_wrapping_test.go
```

command/version_history.go
ccapurso: I don't think that I like falling back to the seal-status endpoint either. I think mentioning that the version-history endpoint was added in 1.10 as well makes sense.

http/sys_version_history_test.go
ncabatoff: Some of these are indeed non-logical endpoints, e.g. health, metrics, pprof, monitor.  Most of the rest are just ancient tests that we probably wouldn't write this way nowadays.  If you look for tests starting with "TestSystemBackend" you'll find a ton of other examples elsewhere that are more conventional.

website/content/api-docs/system/version-history.mdx
ccapurso: I went ahead and added this. There are a handful of examples in our docs where we mention what version an endpoint was added. Like you said, it makes a lot of sense here given the callout of version tracking being introduced in 1.9.0.

command/version_history.go
ccapurso: I didn't add this warning directly to the error handling logic but instead to the existing warning at the top of the CLI output that mentions the earliest version that has been tracked.

See this commit https://github.com/hashicorp/vault/pull/13766/commits/7417fbb168f7bf1acccf19b8844accf2805eda85. What do you think?

http/sys_version_history_test.go
ccapurso: That's great context, thank you! I'll take a look and modify accordingly.

taoism4504 (approved):
Text LGTM

PR #13767, opened 2022-01-24
Client count 1.10 reshuffle

Monkeychip:
Based on various conversations we are restructuring the charting work for the 1.10 client counts. This PR is a first of maybe one or two more that gets us to a visual layout that matches the new designs. The last screenshot is the borked view for users (should be fixed in future PR).

**Here are sceenshots of the view without data:**

**Here are the sceenshots of the view with 1.10 activity data (mirage)**. Note, configuration tab hides because we don't mock for permissions on this endpoint.

Here are screenshots of the view with 1.9 mirage data.

This the scene that is going to be borked for users.


ui/app/routes/vault/cluster/clients/index.js
```diff
@@ -11,6 +11,7 @@ const getActivityParams = ({ tab, start, end }) => {
   if (tab === 'current') {
     params.tab = tab;
   } else if (tab === 'history') {
```
hellobontempo, line 13: Just to clarify - this whole `tab === 'history'` block will be replaced by the `dashboard` block

ui/app/templates/components/clients/current.hbs
```diff
@@ -0,0 +1,70 @@
...
+          Go to configuration
+        </LinkTo>
+      {{/if}}
+    </EmptyState>
+  {{/if}}
+
+  {{#if @isLoading}}
+    <LayoutLoading />
+  {{else}}
+    <div class="card has-bottom-margin-m">
```
hellobontempo, line 21: I think we might be able to use the <Clients::SingleMonth> template here (can be added in future PR once components are wired up correctly)

ui/app/routes/vault/cluster/clients/index.js
Monkeychip: correct. this can actually be removed, but I'm keeping it here so I can remember the remove the query param work as well. However, I don't want to do that until we've gotten the dashboard component working with the 1.9 payload. so TLDR; it will be removed, soon just not in this pr.

ui/app/templates/components/clients/current.hbs
Monkeychip: Or partial month, however it's referred to in the API. but agree

ui/lib/open-api-explorer/addon/templates/components/swagger-ui.hbs
```diff
@@ -35,4 +35,4 @@
     for examples.
   </NamespaceReminder>
   <div id="{{this.elementId}}-swagger" class="swagger-ember"></div>
-</div>
+</div>
```
hellobontempo, line 38: Was this a linting thing?

hellobontempo (approved):
Looks good to me! Thanks for cleaning up - just a couple comments/notes for reference

ui/lib/open-api-explorer/addon/templates/components/swagger-ui.hbs
Monkeychip: yeah, so strange. It caused my tests to fail first round.

PR #13769, opened 2022-01-24
[#13750] Allow policies to control sub-fields of requests

### LINKED ISSUE #13750: Allow policies to control sub-fields of requests
The current policy system works well for controlling the allowed values of "shallow" fields: those at the root of a request. However, for fields that expect a map of values, only a few, restrictive options are currently available:
- Existence checking the map, allowing / denying any values within
- Perform exact string matches for the entire map against its JSON equivalents

For a more concrete example, suppose you wanted to ensure that all `auth/token/create` requests for a given policy, have a populated `metadata` map containing a key (i.e. `creator_id`), set to any value i.e.:
```
POST auth/token/create
{
  "meta": {
    "creator_id": "creator_1",
    "some_other": "meta_key" 
  }
}
```
As far as I'm aware, this cannot be handled by the current policy scheme. The current policy structure would require either: 
- Existence checking the map, allowing / denying any values within:
  ```hcl
  path "auth/token/create" {
    allowed_parameters = {
      "meta" = []
    }
  } 
  ```
- Performing exact string matches for the entire map against its JSON equivalents:
  ```hcl
  path "auth/token/create" {
    allowed_parameters = {
      "meta" = [{"creator_id" = "creator_1"},
        {"creator_id" = "creator_2"}, 
        {"creator_id" = "creator_1", "some_other" = "meta_key"},
        ...]
    }
  } 
  ```

, neither of which feasibly solves the problem.

The policy handler should support "nested parameters": constraints on the contents of some field of a request. For example:
```hcl
path "auth/token/create" {
  capabilities = ["create", "update"]
  required_parameters = ["meta.creator_id"]
  allowed_parameters = {
    "meta.creator_id" = ["creator_*"]
    "meta.*" = []
  }
}
```

This should ideally support an arbitrary depth.

Covered in the problem statement.

Any endpoint that accepts any map could benefit from this.

**Additional context**
N/A

maneac:
**Summary**
Introduce a dot-delimited ('.') syntax to policy parameters to allow the application of parameter restrictions to the _contents_ of request maps. These can be applied to `allowed`, `denied`, and `required` parameters.

**Description**
The unit tests for the ACL define this behaviour in more detail, but for an overview:
- Arbitrary nesting depths are supported
  - e.g. `map.top.middle.bottom`
- Globs for the final section are supported
  - e.g. `map.top.*` will match `map.top.foo` and `map.top.bar.baz`.
- Denies override allows
  - e.g. a denied parameter of `map.*` will overrule an allow for `map.foo`.
- Behaviour of the values for nested parameters match that of "flat" parameters
  - e.g. `map.foo = ["bar", "baz*"]` will allow `bar, baz, bazza, ...`

The policy parser contains validation for the new nested parameter format.

**Approach**
Nested parameter _values_ was considered as an approach. However, using a flat, delimited key format over nested values provides more flexibility when defining constraints. Their behaviour also proved more intuitive, in particular when combining `denied_parameters` with `allowed_parameters`.

**Drawbacks**
Any existing, "flat" parameter in a policy which contains a '.' will no longer be valid. As far as I can tell, these do not exist.

Resolves #13750.

### DIFF
--- a/vault/acl.go
+++ b/vault/acl.go
@@ -447,7 +447,23 @@ CHECK:
 	// parameters.
 	if op == logical.ReadOperation || op == logical.UpdateOperation || op == logical.CreateOperation || op == logical.PatchOperation {
 		for _, parameter := range permissions.RequiredParameters {
-			if _, ok := req.Data[strings.ToLower(parameter)]; !ok {
+			if !strings.Contains(parameter, ".") {
+				if _, ok := req.Data[strings.ToLower(parameter)]; !ok {
+					return
+				}
+				continue
+			}
+
+			parameterParts := strings.Split(strings.ToLower(parameter), ".")
+			data := req.Data
+			for _, part := range parameterParts[:len(parameterParts)-1] {
+				rawData, ok := data[part]
+				if !ok || reflect.TypeOf(rawData).String() != "map[string]interface {}" {
+					return
+				}
+				data = rawData.(map[string]interface{})
+			}
+			if _, ok := data[parameterParts[len(parameterParts)-1]]; !ok {
 				return
 			}
 		}
@@ -713,6 +729,7 @@ func valueInParameterList(v interface{}, list []interface{}) bool {
 }
 
 func valueInSlice(v interface{}, list []interface{}) bool {
+	var nestedGlobs []NestedParameter
 	for _, el := range list {
 		if el == nil || v == nil {
 			// It doesn't seem possible to set up a nil entry in the list, but it is possible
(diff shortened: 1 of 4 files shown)


changelog/13769.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
HridoyRoy, line 1: Could we change this to `improvement`?

HridoyRoy (commented):
Hi @maneac , thanks for the contribution! The code changes in the PR looks good to me, but I'd like to get another engineer's review of it as well ( cc @hghaf099 ). 

I especially would like to think through some cases to make sure these changes are backward compatible. 

Thanks!

PR #13776, opened 2022-01-25
docs: fix limits docs to reflect listener variable name

mickael-hc:
`max_request_size` is appears to be the correct variable name, per https://github.com/hashicorp/vault/blob/1f20ad96dc5251e9599332e769b100335a6b7260/command/server.go#L961


mladlow (approved):
🎉  Thank you!

PR #13778, opened 2022-01-25
Remove Faker

zofskeez:
While the Fakerjs project was temporarily shut down it was decided to remove _faker_ as a dependency in our application and refactor it's usage in mirage factories to static data. It looks like the project has been picked up by a new group of maintainers so we can always add it back in the future if we feel that it would be useful.

https://fakerjs.dev/update.html
https://github.com/faker-js/faker


ui/app/components/file-to-array-buffer.js
```diff
@@ -29,7 +29,9 @@ export default Component.extend({
 
   readFile(file) {
     const reader = new FileReader();
-    reader.onload = () => this.send('onChange', reader.result, file);
+    if (!this.isDestroyed && !this.isDestroying) {
```
hashishaw, line 32: Can we get a little info on why this was added, for posterity?

hashishaw (approved):
Just one comment. Thanks for the updates!

ui/app/components/file-to-array-buffer.js
zofskeez: Sure thing I added a couple of comments.

PR #13788, opened 2022-01-26
UI/Current month view

hellobontempo:
Builds the "Current Month" view by creating a`UsageStat` template 
Screen Shot 2022-01-25 at 5 28 25 PM


ui/app/serializers/clients/activity.js
```diff
@@ -109,9 +110,11 @@ export default ApplicationSerializer.extend({
   },
 
   normalizeResponse(store, primaryModelClass, payload, id, requestType) {
+    let response_timestamp = formatISO(new Date());
     let transformedPayload = {
       ...payload,
       // TODO CMB should these be nested under "data" to go to model correctly?)
+      response_timestamp,
       by_namespace: this.flattenDataset(payload.data.by_namespace),
```
hellobontempo, line 118: Is there a better way to add a timestamp to responses?

ui/app/styles/components/stat-text.scss
```diff
@@ -103,15 +103,3 @@
     }
   }
 }
-
```
hellobontempo, line 106: Kept previous use of `columns` and `column` instead for the `<Clients::UsageStat>` template - unless we want to move away from using those?

ui/app/templates/components/clients/current.hbs
```diff
@@ -18,64 +18,16 @@
...
-              @label="Non-entity tokens"
-              @value={{or @model.monthly.nonEntityTokens "0"}}
-              @size="l"
-              @subText="Tokens created via a method that is not associated with an entity."
-            />
-          </div>
-        </div>
-      </div>
-    </div>
+    <Clients::UsageStats @title={{date-format this.responseTimestamp "MMMM"}} @runningTotals={{this.runningTotals}} />
```
hellobontempo, line 21: This view of 3 `StatText` boxes in a row will also appear in the `dashboard.js` so pulled into a separate template

ui/app/components/clients/current.js
```diff
@@ -0,0 +1,31 @@
...
+    { key: 'entity_clients', label: 'entity clients' },
+    { key: 'non_entity_clients', label: 'non-entity clients' },
+  ];
+
+  // data for horizontal bar chart in attribution component
+  get topTenNamespaces() {
+    if (!this.args.model.monthly || !this.args.model.monthly.byNamespace) {
+      return null;
+    }
+    return this.args.model.monthly.byNamespace;
```
hashishaw, line 14: I _think_ you can ditch the whole if statement block if you do a ternary operator like `return this.args.model.monthly?.byNamespace`

ui/app/serializers/clients/activity.js
```diff
@@ -109,9 +110,11 @@ export default ApplicationSerializer.extend({
   },
 
   normalizeResponse(store, primaryModelClass, payload, id, requestType) {
+    let response_timestamp = formatISO(new Date());
     let transformedPayload = {
       ...payload,
       // TODO CMB should these be nested under "data" to go to model correctly?)
```
hashishaw, line 116: Can probably remove the comment :)

ui/app/styles/components/stat-text.scss
hashishaw: We'll likely replace it with something else once we do move from bulma so that seems fine 👍

hashishaw (approved):
Looks great! Would love to see tests, but I understand if things are still too in flux for that

ui/app/serializers/clients/activity.js
hashishaw: This was supposed to be a response to your columns question

ui/app/components/clients/current.js
hellobontempo: Awesome! I'll play with that - would definitely like to clean it up. I pulled this over from 1.9 client count work

ui/app/serializers/clients/monthly.js
```diff
@@ -1,11 +1,11 @@
-import { formatISO } from 'date-fns';
 import ApplicationSerializer from '../application';
+import { formatISO } from 'date-fns';
 
 export default ApplicationSerializer.extend({
   flattenDataset(payload) {
-    let topTen = payload.slice(0, 10);
+    let topTen = payload?.slice(0, 10);
```
hellobontempo, line 6: @hashishaw added the try `?` here and on line 8 - thoughts? the serializer may change a bit as we're still not sure of whether or not mounts will be included and displayed on the frontend. that's TBD, but wanted a catch in case there was an issue with the `by_namespace` payload. thought it should always be an array (just an empty one..)

hashishaw: I would instead do something like this:
```
let topTen = payload ? payload.slice(0, 10); [];
```
Because as you have it, it will return null rather than the type which you expect, an array. Then you can remove the `?` from line 8 because it will always be an array. 
There's also a nit about styling, since the ternary `?.` operator is typically only used to check for keys _before_ you use the key for something. Hope that makes sense!

PR #13793, opened 2022-01-26
OSS integration of the PKI plugin with managed key infrastructure

stevendpclark:
- The OSS side of things to leverage managed keys from the PKI secrets engine


victorr (approved):
Looks good to me!

PR #13795, opened 2022-01-26
Fix a transit deadlock

schultz-is:
# Summary
In introducing transit time-based key autorotation, I managed to introduce a policy deadlock scenario when transit is run with caching disabled. Calls to `LockManager.GetPolicy` will return a locked policy when caching is disabled, which is why calls to `Policy.Unlock()` won't panic when caching is enabled. During autorotation, keys are retrieved with `GetPolicy` but locking and unlocking is only performed when the key is actually rotated. This means that keys without an autorotate interval defined and keys which haven't reached their autorotation deadline at the time of checking will be permanently locked. The fix in this PR was to elevate the locking and unlocking to cover the whole loop iteration rather than only when the key is due for rotation.

### DIFF
--- a/builtin/logical/transit/backend.go
+++ b/builtin/logical/transit/backend.go
@@ -234,7 +234,6 @@ func (b *backend) autoRotateKeys(ctx context.Context, req *logical.Request) erro
 		err = b.rotateIfRequired(ctx, req, key, p)
 		if err != nil {
 			errs = multierror.Append(errs, err)
-			continue
 		}
 	}
 
--- a/builtin/logical/transit/backend.go
+++ b/builtin/logical/transit/backend.go
@@ -231,32 +231,37 @@ func (b *backend) autoRotateKeys(ctx context.Context, req *logical.Request) erro
 			continue
 		}
 
-		if !b.System().CachingDisabled() {
-			p.Lock(true)
-		}
-
-		// If the policy's automatic rotation interval is 0, it should not
-		// automatically rotate.
-		if p.AutoRotateInterval == 0 {
-			p.Unlock()
+		err = b.rotateIfRequired(ctx, req, key, p)
+		if err != nil {
+			errs = multierror.Append(errs, err)
 			continue
 		}
-
-		// Retrieve the latest version of the policy and determine if it is time to rotate.
-		latestKey := p.Keys[strconv.Itoa(p.LatestVersion)]
-		if time.Now().After(latestKey.CreationTime.Add(p.AutoRotateInterval)) {
-			if b.Logger().IsDebug() {
-				b.Logger().Debug("automatically rotating key", "key", key)
-			}
-			err = p.Rotate(ctx, req.Storage, b.GetRandomReader())
-			if err != nil {
-				errs = multierror.Append(errs, err)
-				p.Unlock()
-				continue
-			}
-		}
-		p.Unlock()
 	}
 
 	return errs.ErrorOrNil()
+}
+
+// rotateIfRequired rotates a key if it is due for autorotation.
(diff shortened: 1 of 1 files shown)


sgmiller (approved):

builtin/logical/transit/backend.go
```diff
@@ -225,15 +225,20 @@ func (b *backend) autoRotateKeys(ctx context.Context, req *logical.Request) erro
 			errs = multierror.Append(errs, err)
 			continue
 		}
+		if !b.System().CachingDisabled() {
+			p.Lock(true)
+		}
 
 		// If the policy is nil, move onto the next one.
 		if p == nil {
+			p.Unlock()
```
victorr, line 234: Won't this and the other unlock calls throw if caching is enabled?

victorr: Ah, I see now that the behaviour is as in the previous code.

schultz-is: Well that one in particular will panic because p is nil. Other unlock calls shouldn't panic because when caching is enabled we manually lock the policy and when caching is disabled GetPolicy returns the policy already locked.

builtin/logical/transit/backend.go
```diff
@@ -231,29 +231,37 @@ func (b *backend) autoRotateKeys(ctx context.Context, req *logical.Request) erro
 			continue
 		}
 
-		// If the policy's automatic rotation interval is 0, it should not
-		// automatically rotate.
-		if p.AutoRotateInterval == 0 {
+		err = b.rotateIfRequired(ctx, req, key, p)
+		if err != nil {
+			errs = multierror.Append(errs, err)
 			continue
```
stevendpclark, line 237: trivial nit: You don't need the continue here.

stevendpclark (approved):
👍

builtin/logical/transit/backend.go
schultz-is: Good eye; removed!

builtin/logical/transit/backend.go
```diff
@@ -231,29 +231,36 @@ func (b *backend) autoRotateKeys(ctx context.Context, req *logical.Request) erro
...
 
 	return errs.ErrorOrNil()
 }
+
+// rotateIfRequired rotates a key if it is due for autorotation.
+func (b *backend) rotateIfRequired(ctx context.Context, req *logical.Request, key string, p *keysutil.Policy) error {
+	if !b.System().CachingDisabled() {
+		p.Lock(true)
+	}
+	defer p.Unlock()
```
briankassouf, line 248: Is it safe to unlock if we didn't just lock?

briankassouf: Oh, i see:

> Calls to LockManager.GetPolicy will return a locked policy when caching is disabled, which is why calls to Policy.Unlock() won't panic when caching is enabled

briankassouf (approved):

builtin/logical/transit/backend.go
stevendpclark: So you are correct Brian, it is not safe to call Unlock if you don't have a previous lock but in this case we do have the lock as Matt mentioned in the summary.

It's an odd pattern and I had to prove it to myself, the lock that is grabbed within the lock manager at line [293](https://github.com/hashicorp/vault/blob/main/sdk/helper/keysutil/lock_manager.go#L293) is not released if caching is disabled (look at the deferred cleanup function a bit lower). 

So we need to perform this pattern or else we will be grabbing a double lock if caching is disabled and hang, but if caching was enabled, the lock within the lock manager was released so we need to grab it again...

PR #13797, opened 2022-01-26
KMS Providers

zofskeez:
This PR adds Providers to the _keymgmt_ secrets backend.

- added model, adapter and serializer
- added `provider-edit` component for show/create/edit views
- added `pagination-controls` component for more flexible pagination to be used in nested views, specifically the _Keys_ tab of the provider-edit show mode.

### DIFF
--- a/ui/app/adapters/keymgmt/key.js
+++ b/ui/app/adapters/keymgmt/key.js
@@ -103,8 +103,13 @@ export default class KeymgmtKeyAdapter extends ApplicationAdapter {
   }
 
   async query(store, type, query) {
-    const { backend } = query;
-    return this.ajax(this.url(backend), 'GET', {
+    const { backend, provider } = query;
+    const providerAdapter = store.adapterFor('keymgmt/provider');
+    const url = provider
+      ? `${providerAdapter.buildURL('keymgmt/provider')}/${provider}/key`
+      : this.url(backend);
+
+    return this.ajax(url, 'GET', {
       data: {
         list: true,
       },
--- a/ui/app/adapters/keymgmt/key.js
+++ b/ui/app/adapters/keymgmt/key.js
@@ -105,9 +105,7 @@ export default class KeymgmtKeyAdapter extends ApplicationAdapter {
   async query(store, type, query) {
     const { backend, provider } = query;
     const providerAdapter = store.adapterFor('keymgmt/provider');
-    const url = provider
-      ? `${providerAdapter.buildURL('keymgmt/provider')}/${provider}/key`
-      : this.url(backend);
+    const url = provider ? providerAdapter.buildKeysURL(query) : this.url(backend);
 
     return this.ajax(url, 'GET', {
       data: {
--- a/ui/app/adapters/keymgmt/provider.js
+++ b/ui/app/adapters/keymgmt/provider.js
@@ -18,6 +18,10 @@ export default class KeymgmtKeyAdapter extends ApplicationAdapter {
     }
     return url;
   }
+  buildKeysURL(query) {
+    const url = this.buildURL('keymgmt/provider', null, null, 'query', query);
+    return `${url}/${query.provider}/key`;
+  }
   async createRecord(store, { modelName }, snapshot) {
     // create uses PUT instead of POST
     const data = store.serializerFor(modelName).serialize(snapshot);
--- a/ui/app/adapters/keymgmt/provider.js
+++ b/ui/app/adapters/keymgmt/provider.js
@@ -1,0 +1,35 @@
+import ApplicationAdapter from '../application';
+import { all } from 'rsvp';
+
+export default class KeymgmtKeyAdapter extends ApplicationAdapter {
+  namespace = 'v1';
+  listPayload = { data: { list: true } };
+
+  pathForType() {
+    return 'keymgmt/kms';
+  }
+  async createRecord(store, type, snapshot) {
+    // create uses PUT instead of POST
+    const data = store.serializerFor(type.modelName).serialize(snapshot);
+    const url = `${this.buildURL(type.modelName)}/${snapshot.attr('name')}`;
+    return this.ajax(url, 'PUT', { data }).then(() => data);
+  }
+  findRecord(store, type, name) {
+    return super.findRecord(...arguments).then((resp) => {
+      resp.data = { ...resp.data, name };
+      return resp;
+    });
+  }
+  async query(store, type) {
+    return this.ajax(this.buildURL(type.modelName), 'GET', this.listPayload).then(async (resp) => {
+      // additional data is needed to fullfil the list view requirements
+      // pull in full record for listed items
+      const records = await all(resp.data.keys.map((name) => this.findRecord(store, type, name)));
+      resp.data.keys = records.map((record) => record.data);
+      return resp;
+    });
+  }
+  async queryRecord(store, type, query) {
+    return this.findRecord(store, type, query.id);
+  }
+}
--- a/ui/app/adapters/keymgmt/provider.js
+++ b/ui/app/adapters/keymgmt/provider.js
@@ -6,12 +6,22 @@ export default class KeymgmtKeyAdapter extends ApplicationAdapter {
   listPayload = { data: { list: true } };
 
   pathForType() {
-    return 'keymgmt/kms';
+    // backend name prepended in buildURL method
+    return 'kms';
   }
-  async createRecord(store, type, snapshot) {
+  buildURL(modelName, id, snapshot, requestType, query) {
+    let url = super.buildURL(...arguments);
+    if (snapshot) {
+      url = url.replace('kms', `${snapshot.attr('backend')}/kms`);
+    } else if (query) {
+      url = url.replace('kms', `${query.backend}/kms`);
+    }
+    return url;
+  }
+  async createRecord(store, { modelName }, snapshot) {
     // create uses PUT instead of POST
-    const data = store.serializerFor(type.modelName).serialize(snapshot);
-    const url = `${this.buildURL(type.modelName)}/${snapshot.attr('name')}`;
+    const data = store.serializerFor(modelName).serialize(snapshot);
+    const url = this.buildURL(modelName, snapshot.attr('name'), snapshot, 'updateRecord');
     return this.ajax(url, 'PUT', { data }).then(() => data);
   }
   findRecord(store, type, name) {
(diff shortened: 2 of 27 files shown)


ui/app/adapters/keymgmt/provider.js
```diff
@@ -0,0 +1,35 @@
+import ApplicationAdapter from '../application';
+import { all } from 'rsvp';
+
+export default class KeymgmtKeyAdapter extends ApplicationAdapter {
+  namespace = 'v1';
+  listPayload = { data: { list: true } };
+
+  pathForType() {
+    return 'keymgmt/kms';
```
hashishaw, line 9: `keymgmt` is actually a dynamic variable, which is the :backend param on the path

ui/app/components/keymgmt/provider-edit.js
```diff
@@ -0,0 +1,108 @@
...
+
+  @action
+  async onSave(event) {
+    event.preventDefault();
+    const { validations } = await this.args.model.validate();
+    if (validations.isValid) {
+      this.saveTask.perform();
+    } else {
+      // FormField expects validationMessages in a shape not output by ember-cp-validations
+      this.modelValidations = validations.errors.reduce(
```
hashishaw, line 76: Since we use FormField all over the place, I wonder if this is worth putting into a helper for use in other areas later?

ui/app/helpers/supported-secret-backends.js
```diff
@@ -11,6 +11,7 @@ const SUPPORTED_SECRET_BACKENDS = [
   'transit',
   'kmip',
   'transform',
+  'keymgmt',
```
hashishaw, line 14: We don't want to commit this until phase 2

ui/app/models/keymgmt/provider.js
```diff
@@ -0,0 +1,119 @@
...
+
+const CRED_PROPS = {
+  azurekeyvault: ['client_id', 'client_secret', 'tenant_id'],
+  awskms: ['access_key', 'secret_key', 'session_token', 'endpoint'],
+  gcpckms: ['service_account_file'],
+};
+const OPTIONAL_CRED_PROPS = ['session_token', 'endpoint'];
+// since we have dynamic credential attributes based on provider we need a dynamic presence validator
+// add validators for all cred props and return true for value if not associated with selected provider
+const credValidators = Object.keys(CRED_PROPS).reduce((obj, providerKey) => {
```
hashishaw, line 14: 👏

ui/app/models/keymgmt/provider.js
```diff
@@ -0,0 +1,119 @@
...
+      azurekeyvault: 'Azure Key Vault',
+      awskms: 'AWS Key Management Service',
+      gcpckms: 'Google Cloud Key Management Service',
+    }[this.provider];
+  }
+  get showFields() {
+    const attrs = expandAttributeMeta(this, ['name', 'created', 'keyCollection']);
+    attrs.splice(1, 0, { hasBlock: true, label: 'Type', value: this.typeName, icon: this.icon });
+    const l = this.keys.length;
+    const value = l ? `${l} ${l > 1 ? 'keys' : 'key'}` : 'None';
```
hashishaw, line 81: We actually have a `type=array` option on InfoTableRow which may allow you to not to this fussing for the keys, at least. See `transform-role-edit.hbs` for one example of how that works.

ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
```diff
@@ -213,10 +212,10 @@ export default Route.extend(UnloadModelRoute, {
     return secretModel;
   },
 
-  async model(params) {
+  async model(params, transition) {
     let secret = this.secretParam();
     let backend = this.enginePathParam();
-    let modelType = this.modelType(backend, secret);
+    let modelType = this.modelType(backend, secret, transition);
```
hashishaw, line 218: I don't love needing to pass around the transition for the type. Could we instead get the `itemType` param here in the model hook (and the other spots where the type is passed in) and then pass it to modelType as a string?

ui/app/templates/components/secret-list/item.hbs
```diff
@@ -4,20 +4,20 @@
...
-  queryParams=(secret-query-params @backendModel.type)
+  queryParams=(secret-query-params @backendModel.type @item.type)
 }}
   <div class="columns is-mobile">
     <div class="column is-10">
       <SecretLink
         @mode={{if @item.isFolder "list" "show"}}
         @secret={{@item.id}}
-        @queryParams={{if (eq @backendModel.type "transit") (query-params tab="actions") ""}}
+        @queryParams={{secret-query-params @backendModel.type @item.type asQueryParams=true}}
```
hashishaw, line 14: If this impacts any of your decisions, you can override this component in `options-for-backend` so we have custom stuff specific to KMSE

ui/tests/integration/components/keymgmt/provider-edit-test.js
```diff
@@ -0,0 +1,203 @@
...
+
+const ts = 'data-test-kms-provider';
+const root = {
+  path: 'vault.cluster.secrets.backend.list-root',
+  model: 'keymgmt',
+  label: 'keymgmt',
+  text: 'keymgmt',
+};
+
+module('Integration | Component | keymgmt/provider-edit', function (hooks) {
```
hashishaw, line 17: 👏

hashishaw (changes requested):
Great work on the pagination controls! There are still a few secrets-engine specific things to fix for providers, but looks like a great start

ui/app/adapters/keymgmt/key.js
```diff
@@ -103,8 +103,13 @@ export default class KeymgmtKeyAdapter extends ApplicationAdapter {
   }
 
   async query(store, type, query) {
-    const { backend } = query;
-    return this.ajax(this.url(backend), 'GET', {
+    const { backend, provider } = query;
+    const providerAdapter = store.adapterFor('keymgmt/provider');
```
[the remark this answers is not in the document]
zofskeez: I went back and forth on this. I need to populate the provider show route keys tab list with keys models. If I do this on the provider adapter I could do something like `store.push` and then `store.peekRecord` but that didn't seem ideal to me considering the query method on the keys adapter already returns keys models. Essentially we are querying keys by a provider.

ui/app/adapters/keymgmt/provider.js
zofskeez: Ahhhh ok I didn't catch that and overlooked the fact that `keymgmt` is autopopulated when you enable the secrets backend but you can actually name it whatever. Thanks!

ui/app/adapters/keymgmt/key.js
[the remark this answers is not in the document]
hashishaw: This isn't always true though, since the keys tab on the backend itself will have keys from any number of providers. This is an interesting point though, which is if the store is populated with *all* keys as they move through the UI, how do we filter it down for the keys tab on the provider view 🤔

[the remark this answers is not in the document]
zofskeez: So on line 108 I do a check for `provider` in the query and then use a different url, `/:backend/:provider/key` which should only return keys for the given provider. Otherwise the `/:backend/key` endpoint is used to fetch all.

ui/app/adapters/keymgmt/provider.js
zofskeez: I refactored the adapter and decided to override the `buildURL` method to pull the backend value from either the snapshot or query in [this commit](https://github.com/hashicorp/vault/pull/13797/commits/014fefa3e6d41f2a1f3ed8840f196c3e0b2860fa).

ui/app/components/keymgmt/provider-edit.js
zofskeez: Yes I think that's a good idea for the time being. I will move it out to a util that can be imported. It might be best in the future to look at refactoring the `FormField` component to accept this shape for validation errors if they are always coming from models.

zofskeez: [I created a helper and refactored it a bit](https://github.com/hashicorp/vault/pull/13797/commits/f9c090d72d3f578894ab42a83c7c011fdcb57ea9) since I had specific logic to build the object based on the nested credentials object. Now it will build the error object for any level of nested attributes.

ui/app/helpers/supported-secret-backends.js
zofskeez: Gotcha I removed it for now.

ui/app/models/keymgmt/provider.js
zofskeez: Cool that's good to know! The `InfoTableItemArray` component won't work in this case since a number based on the array length needs to be displayed such as _6 keys_ rather than the actual array elements.

ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
zofskeez: I agree it's not my favorite either. On line 71 the `buildModel` method is called in the `beforeModel` hook which then calls `modelType`. Since only the transition is available in `beforeModel` that's what I chose to pass along to keep it consistent.

I guess another reason I elected to pass along the transition was if we pull out `itemType` then we have an arg to `modelType` that is specific to keymgmt whereas with the transition there is access to the route params and query params if a case like this comes up again in the future for a different backend.

ui/app/templates/components/secret-list/item.hbs
zofskeez: Gotcha would that be the `listItemPartial` prop then? In this case I noticed that there was already the `secret-query-params` helper created for transit and database backends so I was able to leverage that with minor changes.

ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
hashishaw: If we need the param in a future secret engine, I would suggest we reuse the "itemType" param for that secret engine as well. I think the core issue is that in principle, `modelType` method shouldn't need transition, but maybe we pass in an object "options" that can be extended in the future. That way we can pass in all the queryparams from the transition in the caller of `modelType`.

zofskeez: Ok that sounds good to me I will update it to pass an `options` object that for now will have the `queryParams` from the transition.

ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
```diff
@@ -102,7 +102,7 @@ export default Route.extend(UnloadModelRoute, {
       pki: secret && secret.startsWith('cert/') ? 'pki-certificate' : 'role-pki',
       cubbyhole: 'secret',
       kv: backendModel.get('modelTypeForKV'),
-      keymgmt: `keymgmt/${transition.to.queryParams.itemType || 'key'}`,
+      keymgmt: `keymgmt/${options.queryParams.itemType || 'key'}`,
```
hashishaw, line 105: Should we throw in a ternary here so it doesn't fail when the keys list has no query params?

hashishaw (approved):
One last comment, thanks for all your work on this!

ui/app/routes/vault/cluster/secrets/backend/secret-edit.js
zofskeez: Sure I can but I'm pretty sure `queryParams` is always an object when coming from the `transition`.

zofskeez: Ok I added some additional checks to ensure there isn't a type error when accessing `itemType` on `queryParams`.

PR #13799, opened 2022-01-26
Remove fmt strings and replace with inline queries

imthaghost:
Removed fmt strings and replaced them with inline queries and named parameters.


tomhjp (approved):

calvn (commented):
Can we apply the proper milestones and labels for the PR? It'd also be good to add a changelog entry for this.

@mickael-hc how far do you think we should be backporting this to?

calvn (approved):

changelog/13799.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:security
+database/mssql: Removed fmt strings and replaced them with inline queries using named parameters.
```
calvn, line 2: Slight suggestion:
```suggestion
database/mssql: Removed string interpolation on internal queries and replaced them with inline queries using named parameters.
```

imthaghost: Changed in eca76b4

PR #13803, opened 2022-01-27
Add make fmt CI check

tomhjp:
Following on from #13794, this adds a preflight CI check for formatting to help enforce our formatting rules. It should fail until that PR lands and I merge main into this branch.

I flip-flopped between making this its own step and including it in pre-flight checks, but landed on having its own step to ensure it doesn't slow down CI - lmk what you think in terms of structure.


ncabatoff (approved):
I think it might be better UX to include the diff output.  Otherwise if for some reason the PR author runs `make fmt` and doesn't see any local changes, it'll be confusing and frustrating.

tomhjp:
Good point! I've removed the `--no-patch` option.

PR #13811, opened 2022-01-27
Remove deprecated call to BuildNameToCertificate

cipherboy:
This function call was previously used to generate mappings from
potential subjects (or SANs) to certificates within the TLS client
object. However, newer Go versions have deprecated this method, instead
building the mapping automatically based on present certificates at
request time. Because the corresponding client configuration field is
not used in Vault (`NameToCertificate`), it is safe to remove this call
and leave it `nil`.

See also: https://github.com/golang/go/commit/67d894ee652a3c6fd0a883a33b86686371b96a0e
See also: https://pkg.go.dev/crypto/tls#Config.BuildNameToCertificate


sgmiller (approved):

cipherboy:
Thanks @sgmiller!

PR #13815, opened 2022-01-27
Add notes on the PKI cert generation forwarding regression

sgmiller:


website/content/partials/pki-forwarding-bug.mdx
```diff
@@ -0,0 +1,10 @@
+## PKI Certificate Generation Forwarding Regression
+
+A bug introduced in Vault 1.8 causes certificates issue by the PKI secrets engine made on a performance
```
ncabatoff, line 3:
```suggestion
A bug introduced in Vault 1.8 causes certificates generation requests via the PKI secrets engine made on a performance
```

website/content/partials/pki-forwarding-bug.mdx
```diff
@@ -0,0 +1,10 @@
...
+secondary node to be forwarded to a Vault cluster's primary node.  This affects horizontal scalability, but
```
ncabatoff, line 4: I don't expect the horizontal scalability is the real issue most users, it's more about the certs/CRLs being in the wrong place.  If we do want to retain this bit, I might expand it a little - not everyone's going to follow what we mean by horizontal scalability.

website/content/partials/pki-forwarding-bug.mdx
mladlow: I think "causes certificate generation requests via" (get rid of the pluralization of "certificates")?

website/content/partials/pki-forwarding-bug.mdx
sgmiller: Okay, I'll just drop that part.

mladlow:
@sgmiller I changed the label to "backport/website" which will open and automerge a PR to stable website and open a PR against release/1.9 for versioned docs - could you please make sure to merge the release/1.9 PR? Those don't automerge.

mladlow (approved):
By my read this resolves @ncabatoff's comments but I don't necessarily want to speak for him. Anyway, I didn't understand this issue before reading this note and I now understand it much better, so from my perspective this is good.

PR #13465, opened 2021-12-17
Add LDAP authentication to api

RickyGrassmuck:
This pull request adds LDAP to the Vault API package.

First PR here so Roast away and let me know if there is anything else needed from my end!

PR #13841, opened 2022-01-29
[API] Add LDAP auth method

RickyGrassmuck:
This commit adds LDAP to the available Vault API auth methods.

Recreated to replace #13465


api/auth/ldap/ldap.go
```diff
@@ -0,0 +1,83 @@
...
+
+var _ api.AuthMethod = (*LDAPAuth)(nil)
+
+type LoginOption func(a *LDAPAuth) error
+
+const (
+	defaultMountPath = "ldap"
+)
+
+// NewLDAPAuth initializes a new Userpass auth method interface to be
```
digivava, line 29: Copy-paste typo here, I think you probably meant "a new LDAP auth method".

api/auth/ldap/ldap.go
```diff
@@ -0,0 +1,83 @@
...
+type LDAPAuth struct {
+	mountPath string
+	username  string
+	password  string
+}
+
+type Password struct {
+	// The password as a plaintext string value.
+	FromString string
+}
```
digivava, line 19: With the Userpass and Approle auth modules, we provide the ability to pass in the password FromString, FromEnv, or FromFile, just to allow for different use cases (and to keep consistent). 

Would you be able to provide the same thing here, or is there a reason you only chose to use FromString?

digivava (changes requested):
Thank you for the contribution to this! I've just mentioned one thing you might want to add and then otherwise it seems good to go.

api/auth/ldap/ldap.go
RickyGrassmuck: I think, at the time, my reasoning was likely centered around this being used in more interactive scenarios given the nature of LDAP authentication.

After reading this, It reminded me that I actually have some projects I plan to use this in where having `FromFile` and `FromEnv` would be necessary so I'm positive others out there would benefit from it. 

Got those two methods added in as well as their test coverage, let me know if there's anything else!

api/auth/ldap/ldap.go
```diff
@@ -81,3 +116,46 @@ func WithMountPath(mountPath string) LoginOption {
...
+	}
+
+	passwordValue := strings.TrimSuffix(string(passwordBytes), "\n")
+
+	return passwordValue, nil
+}
+
+func (password *Password) validate() error {
+	if password.FromFile == "" && password.FromEnv == "" && password.FromString == "" {
+		return fmt.Errorf("password for Userpass auth must be provided with a source file, environment variable, or plaintext string")
```
digivava, line 140: Appreciate it! Just one more Userpass -> LDAP change that needs to happen here and then we’re good.

RickyGrassmuck: All set!

digivava (approved):
Great! Thanks again!

PR #13850, opened 2022-01-31
Add support for consul namespaces and admin partitions

robmonte:
Consul has a few new features we'd like to add support for in the Consul secrets engine. This PR is to add support for both Namespaces and Admin Partitioning.

### DIFF
--- a/builtin/logical/consul/backend_test.go
+++ b/builtin/logical/consul/backend_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"log"
+	"os"
 	"reflect"
 	"testing"
 	"time"
@@ -38,7 +39,7 @@ func testBackendConfigAccess(t *testing.T, version string) {
 		t.Fatal(err)
 	}
 
-	cleanup, consulConfig := consul.PrepareTestContainer(t, version)
+	cleanup, consulConfig := consul.PrepareTestContainer(t, "consul", version)
 	defer cleanup()
 
 	connData := map[string]interface{}{
@@ -87,7 +88,7 @@ func TestBackend_Renew_Revoke(t *testing.T) {
 			t.Parallel()
 			t.Run("legacy", func(t *testing.T) {
 				t.Parallel()
-				testBackendRenewRevoke(t, "")
+				testBackendRenewRevoke(t, "1.4.4")
 			})
 
 			testBackendRenewRevoke14(t, "")
@@ -103,7 +104,7 @@ func testBackendRenewRevoke(t *testing.T, version string) {
 		t.Fatal(err)
 	}
 
-	cleanup, consulConfig := consul.PrepareTestContainer(t, version)
+	cleanup, consulConfig := consul.PrepareTestContainer(t, "consul", version)
 	defer cleanup()
 
 	connData := map[string]interface{}{
@@ -196,7 +197,7 @@ func testBackendRenewRevoke(t *testing.T, version string) {
 		Value: []byte("bar"),
 	}, nil)
 	if err == nil {
-		t.Fatal("expected error")
+		t.Fatal("err: expected error")
 	}
 }
 
@@ -208,7 +209,7 @@ func testBackendRenewRevoke14(t *testing.T, version string) {
 		t.Fatal(err)
 	}
 
-	cleanup, consulConfig := consul.PrepareTestContainer(t, version)
+	cleanup, consulConfig := consul.PrepareTestContainer(t, "consul", version)
 	defer cleanup()
 
 	connData := map[string]interface{}{
@@ -320,7 +321,7 @@ func TestBackend_LocalToken(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	cleanup, consulConfig := consul.PrepareTestContainer(t, "")
+	cleanup, consulConfig := consul.PrepareTestContainer(t, "consul", "")
 	defer cleanup()
 
 	connData := map[string]interface{}{
@@ -450,7 +451,7 @@ func TestBackend_Management(t *testing.T) {
 		})
 		t.Run("post-1.4.0", func(t *testing.T) {
 			t.Parallel()
-			testBackendManagement(t, "")
+			testBackendManagement(t, "1.4.4")
 		})
 	})
 }
@@ -463,7 +464,7 @@ func testBackendManagement(t *testing.T, version string) {
 		t.Fatal(err)
 	}
 
-	cleanup, consulConfig := consul.PrepareTestContainer(t, version)
+	cleanup, consulConfig := consul.PrepareTestContainer(t, "consul", version)
 	defer cleanup()
 
 	connData := map[string]interface{}{
@@ -492,10 +493,9 @@ func TestBackend_Basic(t *testing.T) {
 			t.Parallel()
 			t.Run("legacy", func(t *testing.T) {
 				t.Parallel()
-				testBackendRenewRevoke(t, "")
+				testBackendRenewRevoke(t, "1.4.4")
 			})
-
-			testBackendBasic(t, "")
+			testBackendBasic(t, "1.4.4")
 		})
 	})
 }
@@ -508,7 +508,7 @@ func testBackendBasic(t *testing.T, version string) {
 		t.Fatal(err)
 	}
 
-	cleanup, consulConfig := consul.PrepareTestContainer(t, version)
+	cleanup, consulConfig := consul.PrepareTestContainer(t, "consul", version)
 	defer cleanup()
 
 	connData := map[string]interface{}{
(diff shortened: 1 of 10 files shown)


builtin/logical/consul/path_roles.go
```diff
@@ -71,6 +71,20 @@ Defaults to 'client'.`,
 				Description: "Use ttl instead.",
 				Deprecated:  true,
 			},
+
+			"consul_namespace": {
```
kalafut, line 75: Do we need "consul_" here?

builtin/logical/consul/path_token.go
```diff
@@ -90,26 +108,43 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
...
+	var policyLink []*api.ACLTokenPolicyLink
+	if len(policies) > 0 {
+		policyLink = getPolicies(policies)
+	} else {
+		policyLink = getPolicies(result.Policies)
+	}
+
+	// If a namespace was supplied here, then overwrite the namespace
+	// that was given when the role was written
+	if len(namespace) == 0 {
```
kalafut, line 122: Checking for `namespace == ""` is perhaps cleaner and confirms the type to the reader.

builtin/logical/consul/path_token.go
```diff
@@ -90,26 +108,43 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
...
+		namespace = result.Namespace
+	}
+	// If a partition was supplied here, then overwrite the partition
+	// that was given when the role was written
+	if len(partition) == 0 {
```
kalafut, line 127: Same as namespace comment.

builtin/logical/consul/path_roles.go
```diff
@@ -71,6 +71,20 @@ Defaults to 'client'.`,
...
+				Type:    framework.TypeString,
+				Default: "default",
```
kalafut, line 77: How is `"default"` used? Is it a special term that consul understands, different from `""`?

builtin/logical/consul/path_roles.go
robmonte: My thought was to be verbose to avoid potential confusion when using namespaces in the CLI command. I considered that
`vault read -namespace="ns1" consul/creds/my-role namespace="ns2"`
could be mistakenly read and cause someone to mix up which namespace was vault's and which namespace was consul's. So instead, it would be
`vault read -namespace="ns1" consul/creds/my-role consul_namespace="ns2"`

I'm interested in any other opinions or ideas that might be a better solution.

builtin/logical/consul/path_token.go
robmonte: Agreed

builtin/logical/consul/path_roles.go
robmonte: I believe that default is the literal name of the default namespace and partition both, and so consul does understand it. However, removing the default definition here still results in using the "default" in consul, so perhaps it is unnecessary.

builtin/logical/consul/path_token.go
robmonte: Changed https://github.com/hashicorp/vault/pull/13850/commits/45a4d65bcbbef3d2adb5892ed048f5fbf262ad72

builtin/logical/consul/path_roles.go
robmonte: I removed it because it is redundant https://github.com/hashicorp/vault/pull/13850/commits/45a4d65bcbbef3d2adb5892ed048f5fbf262ad72

builtin/logical/consul/path_token.go
```diff
@@ -90,26 +108,43 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
...
 
 	// Create an ACLToken for Consul 1.4 and above
-	policyLink := []*api.ACLTokenPolicyLink{}
-	for _, policyName := range result.Policies {
-		policyLink = append(policyLink, &api.ACLTokenPolicyLink{
-			Name: policyName,
-		})
+	// If policies were supplied here, then overwrite the policies
+	// that were given when the role was written
+	var policyLink []*api.ACLTokenPolicyLink
```
tomhjp, line 113: A bit of an uber nit, and a pre-existing name, but IME this sort of nit helps readability a bit :) it's easier to remember/infer further down that the type is a slice if it's name is plural

```suggestion
	var policyLinks []*api.ACLTokenPolicyLink
```

builtin/logical/consul/path_token.go
```diff
@@ -32,6 +47,9 @@ func pathToken(b *backend) *framework.Path {
 
 func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
 	role := d.Get("role").(string)
+	policies := d.Get("policies").([]string)
+	namespace := d.Get("consul_namespace").(string)
+	partition := d.Get("partition").(string)
```
tomhjp, line 52: I can see why it feels tidier to group these statements together, but I would recommend declaring each right before it's used instead. Separating declaration and usage makes it more difficult to read and also increases the surface area for bugs, especially for mutable variables.

builtin/logical/consul/path_token.go
```diff
@@ -90,26 +108,43 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
...
-		policyLink = append(policyLink, &api.ACLTokenPolicyLink{
-			Name: policyName,
-		})
+	// If policies were supplied here, then overwrite the policies
+	// that were given when the role was written
+	var policyLink []*api.ACLTokenPolicyLink
+	if len(policies) > 0 {
+		policyLink = getPolicies(policies)
+	} else {
+		policyLink = getPolicies(result.Policies)
```
tomhjp, line 117: Not your name, but given the size of the function, it may be a good idea to give `result` a more descriptive name.

tomhjp (approved):
LGTM - just a few small suggestions

builtin/logical/consul/path_token.go
robmonte: Yeah I reused the name already present without thinking too much about it, but I'm glad you pointed it out because I agree with you.

builtin/logical/consul/path_token.go
robmonte: I think you're right and that it's an easy choice for me to move them to where they are used in this case. What is your opinion of the same principle applied to a similar section in path_roles instead: https://github.com/hashicorp/vault/blob/1a9e1c8677dfb543ce606fad0fda442caff4d27f/builtin/logical/consul/path_roles.go#L149-L156
The first 5 were already present and I added just the last 2. Should I place the two I added with their code sections and leave the other five at the top? Or would dispersing the others to their appropriate sections of code start approaching on the "too much refactoring" territory in your opinion?

(At second glace, it seems only a couple of the others would move in the context of the function, but let's assume all 5 of the originals were to move for the sake of argument. I'm not sure which choice I'd prefer.)

builtin/logical/consul/path_token.go
robmonte: Refactored in https://github.com/hashicorp/vault/pull/13850/commits/dfec2cfed14973eb8d5cfe3c92f24bbc8e08c72f

builtin/logical/consul/path_token.go
robmonte: Refactored in https://github.com/hashicorp/vault/pull/13850/commits/dfec2cfed14973eb8d5cfe3c92f24bbc8e08c72f

I think it looks better with them moved after trying it out. Let me know what you think.

tomhjp: On the set of 5, good question. It's in judgement call territory, so I wouldn't argue against either decision personally, but I like the way you've done it now.

tomhjp (approved):

builtin/logical/consul/path_roles.go
```diff
@@ -207,4 +225,6 @@ type roleConfig struct {
 	MaxTTL    time.Duration `json:"max_ttl"`
 	TokenType string        `json:"token_type"`
 	Local     bool          `json:"local"`
+	Namespace string        `json:"consul_namespace"`
```
calvn, line 228:
```suggestion
	ConsulNamespace string        `json:"consul_namespace"`
```

We should be consistent on the storage field and the JSON tag so we should either rename it to `ConsulNamespace` or change the marshaling to `namespace`.

builtin/logical/consul/path_roles.go
```diff
@@ -207,4 +225,6 @@ type roleConfig struct {
...
+	Partition string        `json:partition"`
```
austingebauer, line 229:
```suggestion
	Partition string        `json:"partition"`
```

builtin/logical/consul/path_token.go
```diff
@@ -84,39 +98,71 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
...
+	// that was given when the role was written
+	namespace := d.Get("consul_namespace").(string)
+	if namespace == "" {
+		namespace = roleConfigData.Namespace
+	}
+	// If a partition was supplied here, then overwrite the partition
+	// that was given when the role was written
+	partition := d.Get("partition").(string)
+	if partition == "" {
+		partition = roleConfigData.Partition
```
calvn, line 127: I don't think that we should allow these parameters to be specified during creds generation. Though exposing `policies`, `consul_namespace`, and `partition` on credentials (Consul token) generation can provide flexibility by allowing roles to be re-used, this same flexibility makes it really difficult for Vault to guard how a token should be generated based on values specified in the role.

The pattern for most, if not all, secrets engines is that the credentials generation endpoint takes nothing but the role name. This allows operators to be explicit on the attributes that are tied to the creds to always be consistent and always originate from role parameters. It also makes it easier to protect via Vault's ACL those attributes are tied to the role endpoint which is separate from the credentials generation endpoint.

builtin/logical/consul/path_roles.go
kalafut: sgtm

builtin/logical/consul/path_roles.go
robmonte: Thanks for catching that!

builtin/logical/consul/path_roles.go
robmonte: Makes sense, updated https://github.com/hashicorp/vault/pull/13850/commits/443e3e68fa3e880e3af23bbfee71692734df1395

builtin/logical/consul/path_token.go
robmonte: Removed the ability to do this in https://github.com/hashicorp/vault/pull/13850/commits/443e3e68fa3e880e3af23bbfee71692734df1395

robmonte:
To test the enterprise features, you must set and export the CONSUL_LICENSE variable with a Consul enterprise license key. It was simpler to use the license variable directly and pass it into the test container rather than use a license file and and have to mount it on a docker volume. However, if that's a better implementation then it can be changed.

builtin/logical/consul/backend_test.go
```diff
@@ -702,6 +702,255 @@ func testAccStepDeletePolicy(t *testing.T, name string) logicaltest.TestStep {
...
+
+	var d struct {
+		Token           string `mapstructure:"token"`
+		Accessor        string `mapstructure:"accessor"`
+		ConsulNamespace string `mapstructure:"consul_namespace"`
+	}
+	if err := mapstructure.Decode(resp.Data, &d); err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("Generated token: %s with accessor %s", d.Token, d.Accessor)
```
calvn, line 785: After we decode, can we check that `d.ConsulNamespace == "ns1"`? The same thing for the partition test as well.

calvn (commented):
Looking good, left a few small comments!

helper/testhelpers/consul/consulhelper.go
```diff
@@ -41,21 +41,35 @@ func PrepareTestContainer(t *testing.T, version string) (func(), *Config) {
...
+	name := "consul"
+	isEnterprise := false
+	var envVars []string
+
+	// If running the enterprise container, set the appropriate values below.
+	if strings.HasSuffix(repo, "enterprise") {
+		version += "-ent"
+		name += "-enterprise"
+		isEnterprise = true
+		license, _ := os.LookupEnv("CONSUL_LICENSE")
```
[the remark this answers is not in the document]
robmonte: It does not work, I tried out the image manually and when there's no license provided it immediately stops running with the output:
```
2022-02-08T18:48:23.081Z [ERROR] agent: Error starting agent: error="license is missing. To add a license, configure "license_path" in your configuration file, use the CONSUL_LICENSE environment variable, or use the CONSUL_LICENSE_PATH environment variable. For a trial license of Consul Enterprise, visit https://consul.io/trial."
2022-02-08T18:48:23.081Z [INFO]  agent: Exit code: code=1
```

However I'm not sure where you'd catch this in the code. I think because the container technically starts correctly it doesn't recognize it as an error. What ends up failing is it timeouts on the endpoint to bootstrap the ACL. It actually tripped me up for a bit when I was passing in the license incorrectly at first. I couldn't find a reference to any issue with the license when running in a debugger.

helper/testhelpers/consul/consulhelper.go
```diff
@@ -41,21 +41,35 @@ func PrepareTestContainer(t *testing.T, version string) (func(), *Config) {
...
 
-	repo := os.Getenv("CONSUL_DOCKER_REPO")
-	if repo == "" {
-		repo = "consul"
+	name := "consul"
+	isEnterprise := false
+	var envVars []string
+
+	// If running the enterprise container, set the appropriate values below.
+	if strings.HasSuffix(repo, "enterprise") {
```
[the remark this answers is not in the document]
robmonte: I like that idea, I didn't think of doing it that way. I'll refactor it.

builtin/logical/consul/backend_test.go
robmonte: Yeah makes sense to specifically test for this, added in https://github.com/hashicorp/vault/pull/13850/commits/5cb00f1a8d207a3677df703e176d4d414d684388

helper/testhelpers/consul/consulhelper.go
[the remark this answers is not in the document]
robmonte: Changed in https://github.com/hashicorp/vault/pull/13850/commits/5cb00f1a8d207a3677df703e176d4d414d684388

helper/testhelpers/consul/consulhelper.go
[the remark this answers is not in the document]
robmonte: Austin helped clarify my question about this. Instead of trying to see when the docker container stops, I changed it to match what I did in backend_test. If LookupEnv returns with a false as the second value, then it will throw a t.Fatal with the error message "Failed to find enterprise license".

helper/testhelpers/consul/consulhelper.go
```diff
@@ -26,7 +26,7 @@ func (c *Config) APIConfig() *consulapi.Config {
 // the Consul version used will be given by the environment variable
 // CONSUL_DOCKER_VERSION, or if that's empty, whatever we've hardcoded as the
 // the latest Consul version.
-func PrepareTestContainer(t *testing.T, version string) (func(), *Config) {
+func PrepareTestContainer(t *testing.T, version string, isEnterprise bool) (func(), *Config) {
```
calvn, line 29: Nit: Can we add [t.Helper](https://pkg.go.dev/testing#T.Helper) call in here so that the line error emits from unit tests where this gets call if we ever error out of this function? 

```suggestion
func PrepareTestContainer(t *testing.T, version string, isEnterprise bool) (func(), *Config) {
	t.Helper()
```

calvn (approved):
One last comment, but otherwise looks great!

austingebauer (approved):
LGTM 👍

helper/testhelpers/consul/consulhelper.go
robmonte: Added

PR #13851, opened 2022-01-31
Backport 1.8.x: Fix pkcs7 parsing in some cases (#12519)

jasonodonnell:
* Fix pkcs7 parsing in some cases

brings in https://github.com/mozilla-services/pkcs7/pull/61 from upstream

In some cases but not all, aws includes a certificate in the pkcs7 response,
and currently vault fails to parse those certificates:
```
URL: PUT https://vault.example.com/v1/auth/aws/login
Code: 500. Errors
* failed to parse the BER encoded PKCS#7 signature: ber2der: Invalid BER format
```

This fixes logins on those instances.  Note we could not readily ascertain why
some instances have those certificates and others don't.

* Add changelog entry

* Correct missed line

### DIFF
--- a/builtin/credential/aws/pkcs7/ber.go
+++ b/builtin/credential/aws/pkcs7/ber.go
@@ -175,7 +175,7 @@ func readObject(ber []byte, offset int) (asn1Object, int, error) {
 	if offset > berLen {
 		return nil, 0, errors.New("ber2der: cannot move offset forward, end of ber data reached")
 	}
-	hack := 0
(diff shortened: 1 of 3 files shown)


calvn:
Should we also make a backport for release/1.9.x?

kalafut:
@calvn This landed in `main` back in Sept, so it’s already in the 1.9 releases.

harsimranmaan:
This also seems to be an issue on 1.7.9

PR #13852, opened 2022-01-31
Update k8s auth long-lived token instructions

### LINKED ISSUE #13844: Instructions for obtaining Kubernetes service account token manually are incorrect
**Describe the bug**

https://github.com/hashicorp/vault/blob/main/website/content/docs/auth/kubernetes.mdx describes how to obtain a Kubernetes service account token manually as:

```
kubectl get secret "$(kubectl get serviceaccount default -o jsonpath='{.secrets[0].name}')"
```

This is scraping an auto-generated token, rather than requesting one for external use. The auto-generated tokens are only intended for mounting into pods, and with the transition away from secret-based tokens in Kubernetes 1.22+, new secret-based tokens will no longer be auto-generated in future Kubernetes releases (planned starting in Kubernetes 1.24).

While use of secret-based tokens is discouraged due to their weaker security properties, if one is still needed, https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-service-account-api-token has instructions for manually creating a secret to contain a long-lived token.

https://github.com/hashicorp/vault/blob/main/website/content/docs/auth/kubernetes.mdx should update to align with or reference those instructions.

tomhjp:
Fixes #13844 
As explained in the linked issue, we should advise creating a dedicated token for this purpose.

I've also updated the required version for auto-reloading, and slightly strengthened the wording about the downside of using long-lived tokens, especially now that we have released the auto-reloading feature.

### DIFF
--- a/website/content/docs/auth/kubernetes.mdx
+++ b/website/content/docs/auth/kubernetes.mdx
@@ -154,7 +154,7 @@ table summarizes the options, each of which is explained in more detail below.
 
 | Option                               | All tokens are short-lived | Can revoke tokens early | Other considerations |
 | ------------------------------------ | -------------------------- | ----------------------- | -------------------- |
-| Use local token as reviewer JWT      | Yes                        | Yes                     | Requires Vault (1.10+) to be deployed on the Kubernetes cluster |
+| Use local token as reviewer JWT      | Yes                        | Yes                     | Requires Vault (1.9.3+) to be deployed on the Kubernetes cluster |
 | Use client JWT as reviewer JWT       | Yes                        | Yes                     | Operational overhead |
 | Use long-lived token as reviewer JWT | No                         | Yes                     |                      |
 | Use JWT auth instead                 | Yes                        | No                      |                      |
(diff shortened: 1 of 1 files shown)


tvoran (approved):

tvoran:
Now that https://github.com/hashicorp/vault/pull/13853 is in stable-website, the backport of this one should go cleanly 🤞

asenyaev:
Hi, @tomhjp!

As I understand, we can configure vault to use service-account`s token for an authentication, right?

I tried to follow the [instruction](https://www.vaultproject.io/docs/auth/kubernetes#continue-using-long-lived-tokens) on hashicorp website where I have to create a token as a secret of vault service-account, give `system:auth-delegator` ClusterRole for vault service-account and define `token_reviewer_jwt` to use this secret. However, I cannot apply k8s config in vault with defined `token_reviewer_jwt`. What am I doing wrong?

The command for k8s config inside Vault:
```
vault write auth/kubernetes/config \
    token_reviewer_jwt="vault-k8s-auth-secret" \
    kubernetes_host=https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT \
    kubernetes_ca_cert=@/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
```

The output:
```
Error writing data to auth/kubernetes/config: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/auth/kubernetes/config
Code: 500. Errors:

* 1 error occurred:
        * not a compact JWS
```

If I add a token for `token_reviewer_jwt`, it applies. But I think it's not a good way when we have to find a token, decode and use.

PR #13863, opened 2022-02-01
add security-scan for CRT (#13627)

claire-labry:
* add security-scan

* updating the alpine version

* clean up

* update the alpine version to be more prescriptive


mladlow (approved):
Approved the backports - I think I have a pretty clear understanding of how this works except for the security-scan.hcl file, would you mind dropping a description of that in the comment chain on maybe the main PR or the ENT PR?

PR #13864, opened 2022-02-01
add security-scan for CRT (#13627)

claire-labry:
* add security-scan

* updating the alpine version

* clean up

* update the alpine version to be more prescriptive

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ Vagrantfile
 # Configs
 *.hcl
 !.release/ci.hcl
+!.release/security-scan.hcl
 !.release/linux/package/etc/vault.d/vault.hcl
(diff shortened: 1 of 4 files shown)


mladlow (approved):

claire-labry:
the `security-scan.hcl` is a declarative file where the security-scanner (owned by Security) will attempt to find in your repo. In each of the two stanzas `container` and `binary` there are options that are set to true/false to allow the scanner to read and detect any vulns/CVEs that may arise. If any of those vulns arise, then the workflows will stop and output the vulnerability.
here’s a link to more in depth explanation for that config file: https://github.com/hashicorp/security-scanner/blob/main/CONFIG.md

PR #13872, opened 2022-02-02
Fix kv secret access bug

arnav28:
- Set permissions state when call is successful


Monkeychip (approved):
Nice! This will need to be backported to 1.9, nothing earlier as the KV metadata work was not complete.

PR #13876, opened 2022-02-02
Make backport assistant work with ENT backport tags

mladlow:
I want the regexp to include word characters and the plus sign on
enterprise. I'm confused about the levels of escaping required, because
this is a shell env var that gets passed to a go program as a regular
expression. I didn't escape the square brackets because the parens
weren't escaped. I matched what was done previously with the '.'s for
the '+' because I also want that literal, but it was unclear from
playing around with https://regex101.com/ and go regexps whether the
extra backslashes were required for the '+' char.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -17,6 +17,6 @@ jobs:
         run: |
           backport-assistant backport
         env:
-          BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.[\\w\\+]+)"
+          BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.[+\\w]+)"
           BACKPORT_TARGET_TEMPLATE: "release/{{.target}}"
           GITHUB_TOKEN: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -17,6 +17,6 @@ jobs:
         run: |
           backport-assistant backport
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -17,6 +17,6 @@ jobs:
         run: |
           backport-assistant backport
         env:
-          BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.\\w+)"
+          BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.[\\w\\+]+)"
```
ncabatoff, line 20: I think this would include `\` as a valid char.  Maybe instead 

```suggestion
          BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.[+\\w]+)"
```

(putting the `+` first just to avoid confusing people into thinking it applies to the `\\w`)

.github/workflows/backport.yml
```diff
@@ -17,6 +17,6 @@ jobs:
         run: |
           backport-assistant backport
         env:
-          BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.\\w+)"
+          BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+\\.[+\\w]+)"
```
ncabatoff, line 20: So the goal is to match things like `backport/1.3.+abc`?  Seems a bit weird to have a `+` right after `.`.

ncabatoff (approved):

.github/workflows/backport.yml
mladlow: I think we chatted on Slack but to summarize here, that would be weird but I'm not too worried given the limited usecase to release branches, and having a limited number of release tags.

PR #13879, opened 2022-02-02
UI/Fix activity and monthly serializer

hellobontempo:
Fix serializer to return object if response _does_ contain updated key names `entity_clients` and `non_entity_clients`

Also move truncating data from serializer to component


ui/app/components/clients/attribution.js
```diff
@@ -43,13 +43,15 @@ export default class Attribution extends Component {
     return !!this.args.selectedNamespace;
   }
 
-  get totalClientsData() {
-    return this.args.totalClientsData;
+  // truncate data before sending to chart component
+  // move truncating to serializer when we add separate request to fetch and export ALL namespace data
+  get barChartTotalClients() {
+    return this.args.totalClientsData.slice(0, 10);
```
hellobontempo, line 49: Move truncating back to component and out of serializer

ui/app/components/clients/dashboard.js
```diff
@@ -74,25 +74,23 @@ export default class Dashboard extends Component {
     );
   }
 
-  // top level TOTAL client counts from response for given date range
+  // API client count data by namespace for date range
```
hellobontempo, line 77: Reordered the functions in this component so that `dashboard.js` matches `current.js` when viewing side by side since both have a lot of duplicate functions

ui/app/serializers/clients/activity.js
```diff
@@ -43,6 +45,7 @@ export default class ActivitySerializer extends ApplicationSerializer {
         non_entity_clients,
       };
     }
+    return object;
```
hellobontempo, line 48: Right now this returned object will look like this (based on RFC draft) - TODO is if we want to delete unnecessary keys

```     
         { label:"ns1/",
            distinct_entities:5,
            non_entity_tokens:5,
            clients:10,
            entity_clients:5,
            non_entity_clients:5
         }

Monkeychip (approved):
Nice work!

PR #13885, opened 2022-02-03
UI/Update Routing and Handle response from API

hellobontempo:
Previously all metrics views/tabs were nested under the `clients/index` route. This PR's _optimistic_ goal is to separate the tabs of "Current" and "History" into separate routes and requests to the following:
- `clients/` is the "Current month" tab and hits the `/monthly` endpoint
- `clients/history` is the "History" tab and hits the `/activity` endpoint 
- `clients/config` is the show page for the configuration
- `clients/edit` edits the config

**routing demo**
routing-demo

<hr>

**This PR also updates the UI when different dates are queried and return the respective data. The CSV export also contains **all** namespace and client count info for the respective date range (we were previously truncating this in the serializer)**

query-updates

TO DO: 

- [x] Update subsequent `<LinkTo>` (in templates and empty state messages)
- [x] Test routing actually works
- [x] Fix tests (oh my gosh this is going to break so many)
- [x] Merge Angel's filtering PR
- [x] Figure out this error (UI doesn't error but why am I getting a TypeError - is model not cleaning up?)
route-error


ui/app/components/clients/dashboard.js
```diff
@@ -35,6 +35,7 @@ export default class Dashboard extends Component {
 
   @service store;
 
+  @tracked queriedActivityResponse = null;
```
Monkeychip, line 38: maybe put these in alphabetical order? I think we started that way but then it got all messed up.

ui/app/components/clients/dashboard.js
```diff
@@ -92,12 +93,12 @@ export default class Dashboard extends Component {
...
       return filteredNamespace.mounts ? this.filterByNamespace(this.selectedNamespace).mounts : null;
     } else {
-      return this.byNamespaceActivity;
+      return this.getActivityResponse.byNamespace;
     }
   }
 
   get responseTimestamp() {
-    return this.args.model.activity?.responseTimestamp;
+    return this.getActivityResponse.responseTimestamp;
```
Monkeychip, line 101: on init, does getActivityResponse have any data? I thought on init, the data would be on this.args.model.activity and then after a queried response either through filter or the date pickers, we would return this.activityResponse. Or maybe I'm reading this wrong.

ui/app/components/clients/dashboard.js
hellobontempo: YES - I was waiting to alphabetize because I wanted to clean-up and remove tracked properties we were no longer using (to avoid changing the order and creating unnecessary git conflicts). I'll add a comment TODO to make sure it gets done

ui/app/components/clients/dashboard.js
hellobontempo: Great question! Everywhere this component relied on `this.args.model.activity` I changed to `getActivityResponse` 
Which returns either the passed in arg `model.activity` (the initial response on init, using license start_date). Then subsequently updates as dates are queried. Happy to walk through if you have questions (and make sure this is a good way to approach it). Here's the getter for reference:

```  // on init API response uses license start_date, getter updates when user queries dates
  get getActivityResponse() {
    return this.queriedActivityResponse || this.args.model.activity;
  }
`

Comment (Monkeychip) [reply ui/app/components/clients/dashboard.js:102]: ah got it. Thanks for the explanation!

File: ui/app/components/clients/history-old.js [hunk 3]
```diff
@@ -0,0 +1,124 @@
+import Component from '@glimmer/component';
```
hellobontempo, line 1: This file will be deleted - holding onto now for reference.

ui/app/components/clients/history.js
```diff
@@ -1,124 +1,191 @@
 import Component from '@glimmer/component';
```
hellobontempo, line 1: This file was previously named `dashboard.js` from 1.11 designs, reverting to previous name of `history.js` so correlates to route

ui/app/controllers/vault/cluster/clients/config.js
```diff
@@ -0,0 +1,3 @@
+import Controller from '@ember/controller';
```
hellobontempo, line 1: Need a controller to use the `loading` action in the corresponding route

ui/app/controllers/vault/cluster/clients/index.js
```diff
@@ -1,8 +1,3 @@
 import Controller from '@ember/controller';
 
-export default class ClientsController extends Controller {
```
hellobontempo, line 3: goodbye params!

ui/app/router.js
```diff
@@ -18,7 +18,8 @@ Router.map(function () {
       this.mount('open-api-explorer', { path: '/api-explorer' });
       this.route('license');
       this.route('clients', function () {
-        this.route('index', { path: '/' });
```
hellobontempo, line 21: Ember knows that `clients` is the `index` route so this is redundant (thanks @zofskeez for the ember lesson 🤓 )

ui/app/routes/vault/cluster/clients/index.js
```diff
@@ -1,85 +1,40 @@
 import Route from '@ember/routing/route';
-import ClusterRoute from 'vault/mixins/cluster-route';
```
hellobontempo, line 2: Okay when the heck do we need `import ClusterRoute from 'vault/mixins/cluster-route'` vs `import Route`

ui/app/routes/vault/cluster/clients/index.js
```diff
@@ -1,85 +1,40 @@
...
-  async model() {
-    let config = this.store.queryRecord('clients/config', {}).catch((e) => {
-      console.debug(e);
-      // swallowing error so activity can show if no config permissions
-      return {};
+import RSVP from 'rsvp';
+import { action } from '@ember/object';
+
+export default class ClientsRoute extends Route {
+  // async getMonthly() {
```
hellobontempo, line 6: [RSVP.hash](https://guides.emberjs.com/release/routing/specifying-a-routes-model/#toc_multiple-models) handles multiple API requests - so I don't see the need to separate these functions into separate `async` calls, but perhaps I'm missing something?

ui/app/styles/components/tabs.scss
```diff
@@ -44,3 +44,29 @@
     outline: none;
   }
 }
+
+.nav-tab-link {
```
hellobontempo, line 48: Creating/using this class allowed me to remove the weird nesting doll `<LinkTo>` situation.

ui/app/templates/components/clients/dashboard.hbs
```diff
@@ -1,205 +1,25 @@
...
-        {{/if}}
-      </EmptyState>
-    {{/if}}
-  {{else}}
-    {{#if (eq @model.config.enabled "Off")}}
-      <AlertBanner data-test-tracking-disabled @type="warning" @title="Tracking is disabled">
-        Tracking is currently disabled and data is not being collected. Historical data can be searched, but you will need to
-        <LinkTo @route="vault.cluster.clients.edit">
-          edit the configuration
+<PageHeader as |p|>
```
hellobontempo, line 1: the former `dashboard.hbs` is now `history.hbs` (which is now `old-history.hbs`)

This new _and improved_ 🙃 `dashboard.hbs` contains the tabs and header for client counts. I had trouble using `{{outlet}}` (perhaps it had something to do with importing Route instead of ClusterRoute?!). Though it did work if the parent template was moved _outside_ of the `clients` directory. Opted for creating a separate component instead (per Jordan's recommendation, thanks again) mostly for simplicity and the config view wasn't exactly the same as history and current month

ui/app/templates/components/clients/history-old.hbs
```diff
@@ -0,0 +1,213 @@
+{{#if (and (eq @tab "history") (eq @model.config.queriesAvailable false))}}
```
hellobontempo, line 1: to be deleted..

ui/app/templates/components/clients/history.hbs
```diff
@@ -1,213 +1,205 @@
...
+        @message="There is no data in the monthly history yet. We collect it at the end of each month, so your data will be available on the first of next month."
+      />
+    {{else}}
+      <EmptyState
+        @title="Data tracking is disabled"
+        @message="Tracking is disabled, and no data is being collected. To turn it on, edit the configuration."
+      >
+        {{#if @model.config.configPath.canUpdate}}
+          <p>
+            <LinkTo @route="vault.cluster.clients.index" @query={{hash tab="config"}}>
```
hellobontempo, line 33: Need to update this `<LinkTo>` (haven't updated templates and empty state messages yet)

ui/app/templates/components/clients/history.hbs
```diff
@@ -1,213 +1,205 @@
...
-        tab.
-      </p>
-      {{#if (eq @model.config.enabled "Off")}}
-        <EmptyState
-          @title="Tracking is disabled"
-          @message="Tracking is disabled and data is not being collected. To turn it on edit the configuration."
-        >
-          {{#if @model.config.configPath.canUpdate}}
-            <LinkTo @route="vault.cluster.clients.edit">
+<div class="box is-sideless is-fullwidth is-marginless is-bottomless">
```
hellobontempo, line 1: Will want to move all static info to the the history route, and only have dynamic stuff reliant on API response and getters in component - ya?

ui/app/routes/vault/cluster/clients/index.js
hashishaw: When it's outside of `{ }` we're importing the default export, and so the name here is just whatever we want to call it

zofskeez (approved):
Nice work! I like the change to distinct routes here. I left some comments for consideration. Test coverage would be nice as well but I understand if you are deferring that until closer to the end.

hashishaw (approved):

ui/app/routes/vault/cluster/clients/index.js
hashishaw: But if it was breaking things then I have no idea 🤔

hashishaw:
We just need to fix the merge conflicts and make sure the tests pass (they're all ember-wormhole issues, which should just involve adding an `#ember-wormhole` div in the test render)

ui/app/controllers/vault/cluster/clients/config.js
hellobontempo: When I remove it I get this error - perhaps I'm missing something?

Screen Shot 2022-02-07 at 10 28 09 AM

ui/app/routes/vault/cluster/clients/index.js
hellobontempo: Yeah `config` is required to check if tracking is enabled, and without `monthly` there is no data to display. So if I understand correctly. We would want to trigger a more general error state of neither of those promises were fulfilled.

ui/app/controllers/vault/cluster/clients/config.js
zofskeez: Hmm that's interesting. There's a comment in the docs to do a check for the controller instance in case it hasn't been instantiated yet.

https://api.emberjs.com/ember/4.1/classes/Route/events/loading?anchor=loading

hellobontempo: Oh! I'll look into this - thanks for sending

ui/app/routes/vault/cluster/clients/history.js
```diff
@@ -0,0 +1,54 @@
...
+  rfc33395ToMonthYear(timestamp) {
+    // return ['2021', 2] (e.g. 2021 March, make 0-indexed)
+    return timestamp
+      ? [timestamp.split('-')[0], Number(timestamp.split('-')[1].replace(/^0+/, '')) - 1]
+      : null;
+  }
+
+  async model() {
+    let license = await this.getLicense(); // get default start_time
+    let activity = await this.getActivity(license.startTime); // returns client counts using license start_time.
```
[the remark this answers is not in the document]
hellobontempo: I like this idea... then when fiddling with it (and chatting with Angel) we discussed that we only want to query with the license `start_time` on init, subsequently all the activity data returned is based on what the user queries. So maybe this happening, when initially loading the route does make sense?

I cleaned this file up a bit. But want your thoughts!

.release/security-scan.hcl
```diff
@@ -5,8 +5,8 @@ container {
 }
 
 binary {
-	secrets      = true
-	go_modules   = true
+	secrets      = false
```
hashishaw, line 8: Was this on purpose? 🤔

hashishaw (approved):
Nice! 🚀

.release/security-scan.hcl
hashishaw: I saw a bunch of non-ui changes when I clicked "changes from last review" but I didn't see them on the overall changes so not sure what happened here but just ignore probably

hellobontempo: Could have been when I merged main in? Otherwise I don't know 😬

PR #13888, opened 2022-02-03
turning go modules to false until jwt cve issue fixed

hghaf099:


claire-labry (approved):
LGTM. Thanks for applying this change, much appreciated!

PR #13889, opened 2022-02-03
Add duration/count metrics to PKI issue and revoke flows

sgmiller:


builtin/logical/pki/backend.go
```diff
@@ -136,3 +141,48 @@ The PKI backend dynamically generates X509 server and client certificates.
 After mounting this backend, configure the CA using the "pem_bundle" endpoint within
 the "config/" path.
 `
+
+func metricsKey(req *logical.Request, extra ...string) ([]string, error) {
```
schultz-is, line 145: Looks like this signature could be simplified by removing the error from the return values.

sgmiller: legacy to earlier work, thanks.

schultz-is (approved):
lgtm!

builtin/logical/pki/path_issue_sign.go
```diff
@@ -106,53 +106,24 @@ this value to an empty list.`,
...
-		return logical.ErrorResponse(fmt.Sprintf("unknown role: %s", roleName)), nil
-	}
-
+func (b *backend) pathIssue(ctx context.Context, req *logical.Request, data *framework.FieldData, role *roleEntry) (*logical.Response, error) {
 	if role.KeyType == "any" {
 		return logical.ErrorResponse("role key type \"any\" not allowed for issuing certificates, only signing"), nil
 	}
 
-	return b.pathIssueSignCert(ctx, req, data, role, false, false)
+	resp, err := b.pathIssueSignCert(ctx, req, data, role, false, false)
```
stevendpclark, line 114: nit: Simple return here and within pathSignVerbatim.

stevendpclark (approved):
👍 Just need to address that make fmt error CI is reporting.

builtin/logical/pki/path_revoke.go
```diff
@@ -57,11 +61,28 @@ func (b *backend) pathRevokeWrite(ctx context.Context, req *logical.Request, dat
...
+	}
 
 	b.revokeStorageLock.Lock()
 	defer b.revokeStorageLock.Unlock()
 
-	return revokeCert(ctx, b, req, serial, false)
+	labels := []metrics.Label{
+		metricsutil.NamespaceLabel(ns),
+	}
+	key := metricsKey(req, "revoke")
```
kitography, line 75: Is there a reason that this can't use the metrics wrap that the other calls use?

kitography (commented):
This looks pretty good to me.  Would be cool if there was a test against the metrics here (I'm not sure how to test metrics).

builtin/logical/pki/path_revoke.go
sgmiller: Yes, the metrics wrap assumes "role" is a parameter and grabs it.  revoke just operates on serial numbers.

sgmiller:
> This looks pretty good to me. Would be cool if there was a test against the metrics here (I'm not sure how to test metrics).

I'll have a look, I don't think we test metrics often outside the metrics system tests, but I'll make sure.

builtin/logical/pki/path_issue_sign.go
sgmiller: :nod:  This is what happens when you go through several iterations and then don't check your work.

kitography (approved):
Looks Good to Me :)

builtin/logical/pki/backend.go
```diff
@@ -145,31 +151,49 @@ the "config/" path.
...
+			roleName = data.Get("role").(string)
+		case roleOptional:
+			r, ok := data.GetOk("role")
+			if ok {
+				roleName = r.(string)
+			}
 		}
-		if role == nil {
-			return logical.ErrorResponse(fmt.Sprintf("unknown role: %s", roleName)), nil
+		if roleMode > noRole {
```
kitography, line 180: I think this would be clearer with a != ;
Would also be cool to throw an error if the roleMode is not one of the possible values.

kitography (approved):
LGTM, one readability nit / possible error, but it's internally used so up to you.

Issue #13680, opened 2022-01-17
PKI Engine: Error parsing metadata when creating ed25519 or ec root cert


nwerker:
**Describe the bug**
When generating a new root certificate of the key_type ed25519 or ec I get an error regarding that the metadata couldn't be parsed and therefore I can't see the common_name or expiration dates. This problem does not happen when I use rsa as key_type. In the Vault UI, I get following Info Message:

![grafik](https://user-images.githubusercontent.com/45071484/149792987-d05f438d-2c4a-49dc-af14-e840dc59c4e0.png)

The generated certificate is valid for the set amount, but not being able to see the expiration date via API or CLI makes it hard to monitor. Are there any limitations I am not aware of? 

**To Reproduce**
1. Run `vault secrets enable pki`
2. Run `vault secrets tune --max-lease-ttl=87600h pki`
3. Run `vault write -field=certificate pki/root/generate/internal common_name="root" ttl=87600h key_type=ed25519`
4. See error / screenshot above

I also tried this with lower TTls or with and without common_name resulting in the same error.

**Expected behavior**
Root Cert supports ed25519 with all metadata so it can be queried by API or CLI!

**Environment:**
* Vault Server Version: 1.9.1
* Vault CLI Version: 1.9.1
* Server Operating System/Architecture: RHEL 8.5 (Server and Client)

Vault server configuration file(s):

Can be provided, if needed. The cluster is using raft as storage backend.

cipherboy:
> Are there any limitations I am not aware of? 

Our web UI uses the `node-forge` library which has an [outstanding PR](https://github.com/digitalbazaar/forge/pull/925) for ECDSA support (and lacks support for `ed25519`). This error message [was added recently](https://github.com/hashicorp/vault/pull/13238) to still display _some_ information about the certificate, rather than erring out completely. 

As mentioned by @pcuttriss, perhaps [fetching the cert](https://www.vaultproject.io/api/secret/pki#read-ca-certificate) (which can be done unauthenticated) and parsed by `openssl` suffices? This can be made into a nice one-liner:

```
$ openssl x509 -text -noout <<< "$(curl -sSL http://localhost:8200/v1/pki/ca/pem)"
$ openssl x509 -startdate -enddate -noout <<< "$(curl -sSL http://localhost:8200/v1/pki/ca/pem)"
```

I agree the `vault` CLI lacks support for reading or displaying attributes, this does seem like a worthwhile improvement:

```
$ vault read pki/ca/pem
[...]

PR #13894, opened 2022-02-03
Switch from node-forge to PKI.js

cipherboy:
Definitely needing some review from the UI team here :-D I feel I shouldn't have so many changes to `package.json`, but maybe I can get some thoughts on that?

After breaking a few things, I think I ended up doing the equivalent of:

```
git clean -xdf # to reset node_modules to a clean checkout...
# some command to install all the missing deps again that might've also updated stuff?
npm install --save-dev asn1js pvutils pkijs
```

to get the dependencies to install and build fine with `make static-dist`. 

I've roughly followed the following procedure after building Vault with the built-in UI:

```
source devvault
vault secrets enable pki
vault write pki/root/generate/internal common_name=example.com key_type=ed25519
# (test this change in web UI)
vault delete pki/root
vault write pki/root/generate/internal common_name=example.com key_type=ec key_bits=256
# (test this change in web UI)
vault delete pki/root
vault write pki/root/generate/internal common_name=example.com key_type=ec key_bits=384
# (test this change in web UI)
vault delete pki/root
vault write pki/root/generate/internal common_name=example.com key_type=ec key_bits=521
# (test this change in web UI)
vault delete pki/root
vault write pki/root/generate/internal common_name=example.com key_type=rsa
# (test this change in web UI)
```

I left the `console.log` in case someone has issues parsing certificates in the future, we'll at least be able to point them to a log statement they can share along with the original cert. Let me know if that should be removed. 

Resolves: #13680 (if merged).

/cc @hellobontempo @hashishaw


cipherboy:
Thanks @hashishaw for helping to clean up that `package.json` file. :D

I should probably add a changelog now? Are there any tests or documentation that need to be updated?

hashishaw:
> Thanks @hashishaw for helping to clean up that `package.json` file. :D
> 
> I should probably add a changelog now? Are there any tests or documentation that need to be updated?

Changelog would be excellent, and it looks like our acceptance tests are still passing which means you successfully switch out the libraries while maintaining expected functionality 🎉 

Thanks again for tackling this!

hashishaw (approved):
Great inline comments. Add a changelog and we'll be good to go! 🚀

ui/app/helpers/parse-pki-cert.js
```diff
@@ -1,24 +1,53 @@
...
+  // and where a type is an OID. The OID for CN can be found here:
+  //
+  //    http://oid-info.com/get/2.5.4.3
+  //    https://datatracker.ietf.org/doc/html/rfc5280#page-112
+  //
+  // Each value is then encoded as another ASN.1 object; in the case of a
+  // CommonName field, this is usually a PrintableString, BMPString, or a
+  // UTF8String. Regardless of encoding, it should be present in the
+  // valueBlock's value field if it is renderable.
+  const commonNameOID = '2.5.4.3';
```
cipherboy, line 35: Linting hadn't run the first time, but it did now :-D

changelog/13894.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
hellobontempo, line 1: 🎉

ui/app/helpers/parse-pki-cert.js
hellobontempo: thank you for breaking this down and making is so clear to follow. 😍

ui/app/helpers/parse-pki-cert.js
```diff
@@ -1,24 +1,53 @@
...
+  const commonNames = cert?.subject?.typesAndValues
+    .filter((rdn) => rdn?.type === commonNameOID)
+    .map((rdn) => rdn?.value?.valueBlock?.value);
```
cipherboy, line 38: In case anyone was wondering, this does appear to be the correct way to parse the RDN's value: https://github.com/PeculiarVentures/PKI.js/blob/master/src/AttributeTypeAndValue.js#L213-L214

cipherboy:
Thanks @hashishaw and @hellobontempo!

PR #13895, opened 2022-02-03
contained_db DeleteUser unit test

imthaghost:
This PR adds unit test `TestDeleteUserContainedDB` which allows us to test the inline SQL delete query. Two helper functions ( `assertContainedDBCredsDoNotExist` | `testContainedDBCredsExist` ) were added to check if the credentials exist after the delete query is executed.

### DIFF
--- a/plugins/database/mssql/mssql_test.go
+++ b/plugins/database/mssql/mssql_test.go
@@ -10,8 +10,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/hashicorp/vault/sdk/database/dbplugin"
-
 	mssqlhelper "github.com/hashicorp/vault/helper/testhelpers/mssql"
 	"github.com/hashicorp/vault/sdk/database/dbplugin/v5"
 	dbtesting "github.com/hashicorp/vault/sdk/database/dbplugin/v5/testing"
--- a/plugins/database/mssql/mssql_test.go
+++ b/plugins/database/mssql/mssql_test.go
@@ -4,12 +4,13 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"github.com/hashicorp/vault/sdk/database/dbplugin"
 	"reflect"
 	"regexp"
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/hashicorp/vault/sdk/database/dbplugin"
 
 	mssqlhelper "github.com/hashicorp/vault/helper/testhelpers/mssql"
 	"github.com/hashicorp/vault/sdk/database/dbplugin/v5"
--- a/plugins/database/mssql/mssql_test.go
+++ b/plugins/database/mssql/mssql_test.go
@@ -404,6 +404,44 @@ func TestDeleteUserContainedDB(t *testing.T) {
 	assertContainedDBCredsDoNotExist(t, connURL, dbUser)
 }
 
+func TestContainedDBSQLSanitization(t *testing.T) {
+	cleanup, connURL := mssqlhelper.PrepareMSSQLTestContainer(t)
+	defer cleanup()
+
+	injectionString := "vaultuser]"
+	dbUser := "vaultuser"
+	initPassword := "p4$sw0rd"
+
+	initReq := dbplugin.InitializeRequest{
+		Config: map[string]interface{}{
+			"connection_url": connURL,
+		},
+		VerifyConnection: true,
+	}
+
+	db := new()
+
+	dbtesting.AssertInitializeCircleCiTest(t, db, initReq)
+	defer dbtesting.AssertClose(t, db)
+
+	err := createTestMSSQLUser(connURL, dbUser, initPassword, testMSSQLContainedLogin)
+	if err != nil {
+		t.Fatalf("Failed to create user: %s", err)
+	}
+
+	assertCredsExist(t, connURL, dbUser, initPassword)
+
+	deleteReq := dbplugin.DeleteUserRequest{
+		Username: injectionString,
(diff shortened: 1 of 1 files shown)


plugins/database/mssql/mssql_test.go
```diff
@@ -357,6 +357,53 @@ func TestDeleteUser(t *testing.T) {
 	assertCredsDoNotExist(t, connURL, dbUser, initPassword)
 }
 
+func TestDeleteUserContainedDB(t *testing.T) {
+	cleanup, connURL := mssqlhelper.PrepareMSSQLTestContainer(t)
```
imthaghost, line 361: @tvoran We talked about adding a cleanup helper function to contained db, however doesn't the `PrepareMSSQLTestContainer` function provide us a clean container each time we run this test so it shouldn't interfere with the happy path as well.

tvoran: Yep, I think that's correct. In the previous commit it looks like a single test container was being used for all the test cases, so that's why it was running into side effect issues.

plugins/database/mssql/mssql_test.go
```diff
@@ -411,6 +458,30 @@ func assertCredsDoNotExist(t testing.TB, connURL, username, password string) {
 	}
 }
 
+func assertContainedDBCredsExist(t testing.TB, connURL, username string) {
+	t.Helper()
+	err := testContainedDBCredsExist(connURL, username)
+	if err == nil {
+		t.Fatalf("Able to drop user when it shouldn't")
+	}
```
tvoran, line 466: I think we may want to do some checking on the actual content of this error, too, since it could be any number of things. (For instance right now it's actually `sql: expected 0 arguments, got 1`.)

plugins/database/mssql/mssql_test.go
```diff
@@ -411,6 +458,30 @@ func assertCredsDoNotExist(t testing.TB, connURL, username, password string) {
...
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	userQuery := `DECLARE @stmt nvarchar(max);
+			SET @stmt = 'DROP USER ' + QuoteName(@username);
+			EXEC(@stmt);`
+
+	_, err = db.Exec(userQuery, username)
```
tvoran, line 481: I think this style of sql statement requires a prepare or something, right? We should either do that, or use a non-prepared query like `DROP USER [%s]`.

imthaghost: Added non prepared query in f782d30

plugins/database/mssql/mssql_test.go
imthaghost: f782d30 Adds a check of the error message, this also made me change the function name since we are trying to ensure that there is an error - `assertContainedDBCredsDoNotExist`

plugins/database/mssql/mssql_test.go
imthaghost: Sounds good!

kalafut:
@imthaghost I think one of your commits is using a different email which is causing the CLA check to fail.

imthaghost:
> @imthaghost I think one of your commits is using a different email which is causing the CLA check to fail.

fcfccf9 signed :)

plugins/database/mssql/mssql_test.go
```diff
@@ -357,6 +357,53 @@ func TestDeleteUser(t *testing.T) {
...
+		},
+		VerifyConnection: true,
+	}
+
+	db := new()
+
+	dbtesting.AssertInitializeCircleCiTest(t, db, initReq)
+	defer dbtesting.AssertClose(t, db)
+
+	err := createTestMSSQLUser(connURL, dbUser, initPassword, testMSSQLLogin)
```
tvoran, line 380: I was experimenting a bit with this test, and it looks like `assertContainedDBCredsDoNotExist()` will always think that `vaultuser` doesn't exist, whether `db.DeleteUser` was called or not. So I think as part of "pretending" this is a contained db, we'll want to also create a user to go along with the login, so that there's something to be dropped. i.e.

```suggestion
	err := createTestMSSQLUser(connURL, dbUser, initPassword, testMSSQLContainedLogin)
```
where `testMSSQLContainedLogin` is something like:
```go
const testMSSQLContainedLogin = `
CREATE LOGIN [{{name}}] WITH PASSWORD = '{{password}}';
CREATE USER [{{name}}] FOR LOGIN [{{name}}];
`
```

plugins/database/mssql/mssql_test.go
```diff
@@ -411,6 +458,28 @@ func assertCredsDoNotExist(t testing.TB, connURL, username, password string) {
...
+	assert.EqualError(t, err, "mssql: Cannot drop the user 'vaultuser', because it does not exist or you do not have permission.")
+}
+
+func testContainedDBCredsExist(connURL, username string) error {
+	ctx := context.Background()
+	// Log in
+	db, err := sql.Open("mssql", connURL)
+	if err != nil {
+		return err
+	}
```
tvoran, line 473: Don't forget to close the connection:

```suggestion
	db, err := sql.Open("mssql", connURL)
	if err != nil {
		return err
	}
	defer db.Close()
```

imthaghost: Updated in cf19eef

plugins/database/mssql/mssql_test.go
imthaghost: Updated in cf19eef

tvoran (approved):

PR #13907, opened 2022-02-04
Client counts version-history endpoint

Monkeychip:
This helps setup the error message you should receive if the version date you upgrade comes after the startTime returned from the response. 

**Here is a video walk through:**
_* note for the current tab I compare to the first day of the current month, and for the history tab I look at the startTime returned from the response._

https://user-images.githubusercontent.com/6618863/152590556-f8221f2c-4716-430b-9da6-4e3728b032c9.mp4

**Here is an image of the two warning merged together as requested by design.**
image


ui/app/routes/vault/cluster/clients/index.js
```diff
@@ -15,18 +15,27 @@ export default Route.extend(ClusterRoute, {
     },
   },
 
-  async getLicense() {
+  async getActivity(start_time) {
```
Monkeychip, line 18: alphabetized some.

ui/app/components/clients/current.js
```diff
@@ -11,9 +12,20 @@ export default class Current extends Component {
   });
   @tracked selectedNamespace = null;
 
-  // TODO CMB get from model
   get upgradeDate() {
-    return this.args.upgradeDate || null;
+    let keyInfoObject = this.args.model.versionHistory.key_info;
```
hellobontempo, line 16: i think this will work if `key_info` is camel cased? (here and in the model)

hellobontempo: (Putting here for my brain)
sample response: 
```{
  "keys": [
    "1.9.0",
    "1.9.1",
    "1.9.2"
  ],
  "key_info": {
    "1.9.0": {
      "previous_version": null,
      "timestamp_installed": "2021-11-18T10:23:16Z"
    },
    "1.9.1": {
      "previous_version": "1.9.0",
      "timestamp_installed": "2021-12-13T11:09:52Z"
    },
    "1.9.2": {
      "previous_version": "1.9.1",
      "timestamp_installed": "2021-12-23T10:56:37Z"
    }
  }
}

File: ui/app/components/clients/current.js [hunk 3]
```diff
@@ -11,9 +12,20 @@ export default class Current extends Component {
...
+    if (!keyInfoObject) {
+      return false;
+    }
+    let firstKey = Object.keys(keyInfoObject); // key name here will change, but we want the first one as this indicates when they upgraded to 1.9 or above
```
hellobontempo, line 20: same thing regarding naming here, and since we're calling it the "first" I think grabbing the zero index here would make sense 
`let earliestUpgradeVersion = Object.keys(keyInfoObject)[0]`

ui/app/components/clients/dashboard.js
```diff
@@ -2,7 +2,7 @@ import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
-import { isSameMonth } from 'date-fns';
+import { isSameMonth, compareAsc } from 'date-fns';
```
hellobontempo, line 5: would it make sense to pull these out into a util? so both dashboard and current can use them? And if we're adding to charts in 1.11 then each chart could access the function too?

ui/app/components/clients/current.js
```diff
@@ -11,9 +12,20 @@ export default class Current extends Component {
...
+    let versionDate = new Date(keyInfoObject[firstKey].timestamp_installed);
+    // compare against this month and this year to show message or not.
+    let compare = compareAsc(versionDate, startOfMonth(new Date()));
```
hellobontempo, line 23: is there a reason you chose `compareAsc` over `isAfter`? The latter feels more readable, and would turn it into a one liner

`return isAfter(versionDate, startOfMonth(newDate())`

ui/app/components/clients/current.js
Monkeychip: ah, so camel case in the model and then camelCase here?

ui/app/components/clients/current.js
Monkeychip: good call.

ui/app/components/clients/current.js
Monkeychip: nope, just didn't know that one existed. Searched the docs and the compareAsc came first. I like isAfter. I'll amend.

ui/app/components/clients/dashboard.js
Monkeychip: I think with the change to the isAfter I'm now using the method on date-fns and it's pretty clean. Thoughts?

ui/app/components/clients/current.js
```diff
@@ -11,9 +12,15 @@ export default class Current extends Component {
...
   get upgradeDate() {
-    return this.args.upgradeDate || null;
+    let keyInfoObject = this.args.model.versionHistory.keyInfo;
+    if (!keyInfoObject) {
+      return false;
+    }
+    let earliestUpgradeVersion = Object.keys(keyInfoObject)[0]; // key name here will change, but we want the first one as this indicates when they upgraded to 1.9 or above
+    let versionDate = new Date(keyInfoObject[earliestUpgradeVersion].timestamp_installed);
+    // compare against this month and this year to show message or not.
+    return isAfter(versionDate, startOfMonth(new Date())) ? versionDate : false;
```
hashishaw, line 23: Based on the method name I would expect this to return a date, not a boolean. I wonder if the naming could be updated to be more clear about that, or always return the version date and do the `isAfter` check in the consuming method

ui/app/adapters/clients/version-history.js
```diff
@@ -0,0 +1,13 @@
+import Application from '../application';
+
+export default Application.extend({
+  queryRecord() {
```
hashishaw, line 4: I believe we get free caching from ember-data if we instead use the `findAll` method, which we can use here since we're not actually using any query params it's just a static API call.

ui/app/models/clients/version-history.js
```diff
@@ -0,0 +1,5 @@
+import Model, { attr } from '@ember-data/model';
+export default class VersionHistoryModel extends Model {
+  @attr('array') keys;
+  @attr('object') keyInfo;
```
hashishaw, line 4: I wonder if it would be valuable to store each version upgrade data as a new record 🤔 especially for the future where we may incorporate these values into the charts

ui/app/routes/vault/cluster/clients/index.js
```diff
@@ -15,18 +15,27 @@ export default Route.extend(ClusterRoute, {
...
   },
 
-  async getActivity(start_time) {
+  async getVersionHistory() {
     try {
-      return await this.store.queryRecord('clients/activity', { start_time });
+      return await this.store.queryRecord('clients/version-history', {});
+    } catch (e) {
+      // ARG TODO handle
+      return e;
```
hashishaw, line 32: Failing this call should allow the UI to fallback gracefully so I'd go ahead and return null rather than the error. Would likely be a permissions issue anyway

ui/app/components/clients/dashboard.js
```diff
@@ -101,6 +101,17 @@ export default class Dashboard extends Component {
   get responseTimestamp() {
     return this.args.model.activity?.responseTimestamp;
   }
+
+  get upgradeDate() {
+    let keyInfoObject = this.args.model.versionHistory.keyInfo;
```
hashishaw, line 106: If the version history is not defined we should add a check here or we'll get a `versionHistory is undefined` error
```suggestion
    let keyInfoObject = this.args.model.versionHistory?.keyInfo;
```

hashishaw (commented):
Nice work! I made a few suggestions that I think will make the code a bit cleaner and easier to expand on in the future. Also, tests?

hashishaw (approved):
Beautiful, very nice work!

PR #13913, opened 2022-02-04
UI/Fix parsing CRL in PKI engine

hellobontempo:
Changes the `console.log` to a `console.debug` and adds additional check so `parsePkiCert` skips CRLs. Also removes tests because new PKI parsing library _can_ parse EC certs! 🎉 
(thank you @cipherboy! 😄 )
<hr>
Screen Shot 2022-02-04 at 1 23 13 PM

May want to consider adding test coverage using Pretender (had trouble getting this to work with all the intricate configs necessary to mount and view a PKI cert)


Monkeychip (approved):
Thanks for jumping on this!

PR #13920, opened 2022-02-07
Turning off binary secret scanning

sarahethompson:
Turning off binary secret scanning to allow builds to be processed until https://github.com/hashicorp/security-scanner/issues/166 is addressed.


mladlow (approved):
Thanks @sarahethompson !

PR #13927, opened 2022-02-07
Use application/pem-certificate-chain for PEMs

cipherboy:
As mentioned in #10948, it appears we're incorrectly using the
`application/pkix-cert` media type for PEM blobs, when
`application/x-pem-file` is more appropriate. Per RFC 5280 Section
[4.2.1.13](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.13), `application/pkix-crl` is only appropriate when the CRL is in
DER form. Likewise, Section [4.2.2.1](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.2.1) states that `application/pkix-cert`
is only applicable when a single DER certificate is used.

Per recommendation in RFC 8555 ("ACME"), Section [7.4.2](https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2) and [9.1](https://datatracker.ietf.org/doc/html/rfc8555#section-9.1), we use
the newer `application/pem-certificate-chain` media type for
certificates. However, this is not applicable for CRLs, so we use fall
back to `application/x-pem-file` for these. Notably, no official IETF
source is present for the latter. On the OpenSSL PKI tutorial
(https://pki-tutorial.readthedocs.io/en/latest/mime.html), this type is
cited as coming from S/MIME's predecessor, PEM, but neither of the main
PEM RFCs (RFC 934, 1421, 1422, 1423, or 1424) mention this type.

---

While https://github.com/ietf-wg-acme/acme/issues/120#issuecomment-218182901 seems to suggest `application/x-pem-file` is appropriate for PEM blobs in general (regardless of type), they ended up removing this in the final draft and introducing a new type.

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -1768,7 +1768,7 @@ func TestBackend_PathFetchValidRaw(t *testing.T) {
 	if bytes.Compare(resp.Data[logical.HTTPRawBody].([]byte), pemCert) != 0 {
 		t.Fatalf("failed to get pem cert")
 	}
-	if resp.Data[logical.HTTPContentType] != "application/pkix-cert" {
+	if resp.Data[logical.HTTPContentType] != "application/pem-certificate-chain" {
 		t.Fatalf("failed to get raw cert content-type")
 	}
 }
--- a/builtin/logical/pki/ca_test.go
+++ b/builtin/logical/pki/ca_test.go
@@ -320,7 +320,7 @@ func runSteps(t *testing.T, rootB, intB *backend, client *api.Client, rootName, 
 			if diff := deep.Equal(resp.Data["http_raw_body"].([]byte), []byte(caCert)); diff != nil {
 				t.Fatal(diff)
 			}
-			if resp.Data["http_content_type"].(string) != "application/pkix-cert" {
+			if resp.Data["http_content_type"].(string) != "application/pem-certificate-chain" {
 				t.Fatal("wrong content type")
 			}
 		}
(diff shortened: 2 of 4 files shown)


cipherboy:
@abriening what's your thoughts on using `application/pem-certificate-chain` instead of `application/x-pem-file`? 

For what its worth, Dogtag produces `application/x-pem-file`: https://github.com/dogtagpki/pki/blob/master/base/common/src/main/java/com/netscape/certsrv/authority/AuthorityResource.java#L42-L45 but respects the ACME standard `application/pem-certificate-chain`: https://github.com/dogtagpki/pki/blob/master/base/acme/src/main/java/org/dogtagpki/acme/server/ACMECertificateService.java#L35-L39

kitography (approved):
This makes sense to me :)

(I'm following the slack chat, but I think it's very unlikely that someone is relying on an incorrect contentType)

abriening:
I've used the `application/pkix-*` content-types in PKI testing. I only added the `raw/pem` endpoint to have parity with the existing `raw(pem)` endpoint.

Looks like ACME draft was `application/x-pem-file` then created `application/pem-certificate-chain` specifically:
https://github.com/ietf-wg-acme/acme/commit/709a654e37a7ce714c0c17d483900ce9fa47e700

Seems like `application/pem-certificate-chain` is appropriate, just hope no one sends `Accept: application/pem-certificate-chain` expecting the full chain.

cipherboy:
@abriening Thanks! For what it is worth, it appears that Vault doesn't respect (or check) the Accept header either way, so in that regard we'll at least be consistent with other APIs.

That said, we have approvals and consensus internally that we're fine with this change (but not backporting it), so I think it is OK to merge. 

Thanks all for the reviews!

PR #13931, opened 2022-02-07
Correct return value explanation in docstring

### LINKED ISSUE #13636: docstring for api.DefaultConfig() explains error case wrongly
The current docstring is https://github.com/hashicorp/vault/blob/33d61b5ca647f308c173e29506727e3d643e51aa/api/client.go#L180-L187

However, the function never actually returns nil. If there is an error, a non-nil Config instance is returned and `config.Error` will be filled accordingly. I was going to do a PR, but I won't bother with the CLA stuff for a one-line patch.

digivava:
Fixes #13636 

Updated the docstring to reflect what actually happens when an error is encountered in DefaultConfig.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -189,7 +189,7 @@ type TLSConfig struct {
 // The default Address is https://127.0.0.1:8200, but this can be overridden by
 // setting the `VAULT_ADDR` environment variable.
 //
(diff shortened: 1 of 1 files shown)


VinnyHC (approved):
👍

Issue #13489, opened 2021-12-21
pki: ca_chain endpoints do not return the chain when populated by root/generate


dnephin:
**Background**

Vault has 3 endpoints for reading certificates from a PKI secrets engine, one of which has multiple options for the response:
1. [pki/ca/pem](https://www.vaultproject.io/api/secret/pki#read-ca-certificate) - returns the single certificate
2. [pki/ca_chain](https://www.vaultproject.io/api/secret/pki#read-ca-certificate-chain) - returns the full certificate chain
3. [pki/cert/{ca,ca_chain}](https://www.vaultproject.io/api/secret/pki#read-certificate) - returns a JSON response with either the single certificate, or the full certificate chain

**Describe the bug**

`pki/ca_chain` and `pki/cert/ca_chain` return an empty response when the value was created with [/pki/root/generate/internal](https://www.vaultproject.io/api/secret/pki#generate-root). They correctly return the full chain if the value was populated using [/pki/intermediate/set-signed](https://www.vaultproject.io/api/secret/pki#submit-ca-information).

**To Reproduce**
```
$ vault secrets enable pki
$ vault write -field=certificate pki/root/generate/internal common_name="dc1.consul"

$ curl -v "${VAULT_ADDR}/v1/pki/ca_chain"
# 204 No Content , with empty response body

$ curl -v "${VAULT_ADDR}/v1/pki/cert/ca_chain"
{"data":{"certificate":"","revocation_time":0}}
```

**Expected behavior**

The docs say the following about these endpoints:

> This endpoint retrieves the CA certificate chain, **including the CA**

And, when a chain is saved using `/pki/intermediate/set-signed` they do properly return the full chain (including the root cert, if it was included in the PEM passed to set-signed).

So my expectations are:

The first `curl` should have a response body with the certificate PEM, not an empty response. A chain of 1 root cert is still a valid chain.

The second `curl` should have a non-empty value for `data.certificate`. The value should be the certificate PEM of the root cert.

dnephin:
Related to https://github.com/hashicorp/consul/issues/11598

dnephin:
I understand fixing this may be a breaking change for anyone expecting an empty result. But it seems like `pki/cert/ca_chain` can be changed in a backwards compatible way. A new `ca_chain` field could be added to the JSON response.

PR #13935, opened 2022-02-07
Add full CA Chain to /pki/cert/ca_chain response

cipherboy:
As mentioned in #13489, calls to read `/pki/cert/ca_chain` would return a partial chain, excluding the root authority. By adding a new field, `ca_chain` to the response, we can include the entire chain without breaking backwards compatibility.

We test three main scenarios:
    
 1. A root-only CA's `/cert/ca_chain`'s `.data.ca_chain` field should contain only the root,
 2. An intermediate CA (with root provide) should contain both the root and the intermediate.
 3. An external (e.g., `/config/ca`-provided) CA with both root and intermediate should contain both certs.

n.b.: I think the existing note around `ca_chain` in the documentation is wrong, hence its removal. 

Additional note: this is only the full chain as _known to Vault_. If the entire (from a TLS client's PoV) chain wasn't provided during either `/intermediate/set-signed` or `/config/ca`, Vault wouldn't know the entire chain and thus couldn't provided. That'd be an enhancement for the future.

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -3724,6 +3724,159 @@ func TestBackend_RevokePlusTidy_Intermediate(t *testing.T) {
 	}
 }
 
+func TestBackend_Root_FullCAChain(t *testing.T) {
+	coreConfig := &vault.CoreConfig{
+		LogicalBackends: map[string]logical.Factory{
+			"pki": Factory,
+		},
+	}
+	cluster := vault.NewTestCluster(t, coreConfig, &vault.TestClusterOptions{
+		HandlerFunc: vaulthttp.Handler,
+	})
+	cluster.Start()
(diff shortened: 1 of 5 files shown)


kitography (commented):
This looks good to me.

kitography (approved):
Actually approving this time.  Looks good to me :)

divyapola5 (approved):

stevendpclark (commented):
Code looks great 👍  

I would just like to see an update to the "Read CA Certificate Chain" api docs to mention if they want the full CA Certificate chain it is available within the "Read Certificate" api with the ca_chain parameter...

cipherboy:
Thanks @stevendpclark, I've added a note -- let me know if that seems right?

stevendpclark (approved):

PR #13937, opened 2022-02-07
Calendar widget tooltip

Monkeychip:
**This PR does the following:**

* Adds a tooltip to explain why previous years are disabled in the calendar-widget.
* When the warnings are combined (e.g. an upgrade to 1.9 warning and you selected data before you have data) then we bulletize those into one warning
* Fixes a bug on the calendar widget on disabling current year months.

https://user-images.githubusercontent.com/6618863/153245349-5c8eb153-f994-4525-a85a-4b5dfd9b7828.mp4


ui/app/components/clients/history.js
```diff
@@ -171,9 +154,13 @@ export default class History extends Component {
...
-      // this only gets triggered if the data was returned, which does not happen if the user selects a startTime after for which we have data. That's an adapter error and is captured differently.
-      if (!this.areArraysTheSame(this.startTimeFromResponse, this.startTimeRequested)) {
+      // compare if the response startTime comes after the requested startTime. If true throw a warning.
+      if (
+        isAfter(
+          new Date(this.getActivityResponse.startTime),
+          new Date(this.startTimeRequested[0], this.startTimeRequested[1])
+        )
+      ) {
         this.responseRangeDiffMessage = `You requested data from ${month} ${year}. We only have data from ${this.startTimeDisplay}, and that is what is being shown here.`;
```
hellobontempo, line 164: In the screenshot I selected August as a new **end date** from the widget and it displayed as `7 2021` (the warning shouldn't even be displaying the end date, right?). Figure it's mirage weirdness - when I tested with the "real" mock data all seemed to update and query correctly! Just pointing out in case it hints to a larger issue I'm not aware of
Screen Shot 2022-02-09 at 4 29 10 PM

hellobontempo (commented):
I think having someone else formally review would be best, but left a couple comments. 

For some reason the version info in the warning banner didn't display in Firefox? Not sure if it's some weird local issue or what...

ui/app/components/clients/history.js
Monkeychip: Good catch. I needed another conditional to only trigger comparison when they've selected a startTime. Amended.

hashishaw (approved):
LGTM 🚀

PR #13950, opened 2022-02-08
UI: Add check for renewal time before triggering renew-self

hashishaw:
This PR addresses a complaint where the background token renewal within the app was silently triggering a 3rd party MFA check after a few minutes of inactivity. The behavior in the UI is updated so that the token is only renewed if 3 minutes of inactivity pass *and* at least half of the token's TTL is passed from authentication time. 

Thank you @austingebauer for your deep contributions to this PR!


Monkeychip (approved):
Nice work on tracking down the fix.

Issue #13157, opened 2021-11-16
An error if specify other_sans In the sign-intermediate endpoint


hiyosi:
**Describe the bug**

The sign-intermediate endpoint returns an error if specify `other_sans` in a request.
We can not use any value for `other_sans`.

**To Reproduce**

```
$ curl \
   --header "X-Vault-Token: $VAULT_TOKEN"  \
   --data '{"csr": "$CSR", "other_sans": "1.3.6.1.4.1.311.20.2.3;utf8:devops@example.com"}'
  $VAULT_ADDR/v1/pki/root/sign-intermediate

{"errors":["other SAN OID 1.3.6.1.4.1.311.20.2.3 not allowed by this role"]}
```

cf. https://www.vaultproject.io/api/secret/pki#other_sans-3

**Expected behavior**

We can specify any values for `other_sans` in sign-intermediate endpoint.
Or remove the parameter if we can not use it.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.8.4
* Vault CLI Version (retrieve with `vault version`): v1.8.4
* Server Operating System/Architecture:  running on k8s, installed by helm chart

Vault server configuration file(s):

```hcl
disable_mlock = true
ui = true

listener "tcp" {
  tls_disable = 1
  address = "[::]:8200"
  cluster_address = "[::]:8201"
}
storage "file" {
  path = "/vault/data"
}

# Example configuration for using auto-unseal, using Google Cloud KMS. The
# GKMS keys must already exist, and the cluster must have a service account
# that is authorized to access GCP KMS.
#seal "gcpckms" {
#   project     = "vault-helm-dev"
#   region      = "global"
#   key_ring    = "vault-helm-unseal-kr"
#   crypto_key  = "vault-helm-unseal-key"
#}
```

**Additional context**

There is no problem even if specify other_sans in generate API like `/pki/root/generate/:type` and `pki/intermediate/generate/:type`.

https://github.com/hashicorp/vault/blob/main/builtin/logical/pki/ca_util.go#L42

In sign-intermediate, AllowedOtherSANs parameter is not set.

https://github.com/hashicorp/vault/blob/main/builtin/logical/pki/path_root.go#L269-L287

cipherboy:
Hello @hiyosi, I agree with your analysis. Are you interested in opening a pull request addressing it? If so, I'd be happy to review.

PR #13958, opened 2022-02-08
Allow all other_sans in sign-intermediate and sign-verbatim

cipherboy:
`/sign-verbatim` and `/sign-intermediate` are more dangerous endpoints in
that they (usually) do not have an associated role. In this case, a
permissive role is constructed during execution of these tests. However,
the `AllowedOtherSANs` field was missing from this, prohibiting its use
when issuing certificates.

Resolves: #13157

--

`/sign-verbatim` actually takes an optional name parameter, which then uses the role's restrictions. This doesn't impact that case at all.


stevendpclark (approved):
LGTM!

PR #13964, opened 2022-02-09
UI/Client counts view if no license

hellobontempo:
Shows appropriate empty state message and adds a dropdown so a user can select a start date if one is not returned from the license or if a license doesn't exist.

no-billing-start


ui/app/components/clients/history.js
```diff
@@ -60,7 +60,6 @@ export default class History extends Component {
 
   get startTimeDisplay() {
     if (!this.startTimeFromResponse) {
-      // otherwise will return date of new Date(null)
```
hellobontempo, line 63: I think this comment is no longer relevant?

ui/app/routes/vault/cluster/clients/history.js
```diff
@@ -5,16 +5,19 @@ import { action } from '@ember/object';
 export default class HistoryRoute extends Route {
   async getActivity(start_time) {
```
hellobontempo, line 6: Thought it made sense to _only_ make the request if we had a valid `start_time` from the license, but thoughts on why this could cause problems?

ui/app/components/date-dropdown.js
```diff
@@ -0,0 +1,38 @@
+import Component from '@glimmer/component';
```
hellobontempo, line 1: Make `add-on` component?

hashishaw: I don't think it's quite generic enough yet. Keeping it as its own component is 👍 though

ui/app/components/clients/history.js
```diff
@@ -179,7 +180,7 @@ export default class History extends Component {
       }
       this.queriedActivityResponse = response;
     } catch (e) {
-      // ARG TODO handle error
+      return e;
```
hashishaw, line 183: `return e` will return an error-shaped object which you have to check for, while `throw e` will allow you to wrap the consumer in a try/catch. If you do want to throw e you actually don't need to wrap this method in a try/catch because it will throw automatically. Just some notes when you think through how you want to handle errors

hashishaw (approved):
A few very minor comments. Nice work!

ui/app/templates/components/date-dropdown.hbs
```diff
@@ -48,10 +48,7 @@
   type="button"
   class="button is-primary"
   disabled={{if (and this.startMonth this.startYear) false true}}
-  {{on
-    "click"
-    (queue (fn @handleDateSelection this.startMonth this.startYear @dateType) (action (mut this.isEditStartMonthOpen) false))
-  }}
+  {{on "click" (fn @handleDateSelection this.startMonth this.startYear @name)}}
```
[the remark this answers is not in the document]
hellobontempo: Good question! 
This dropdown just shows when there is no start date, as soon a start_date is selected the DOM changes and the date renders instead with an 'edit' button next to it, which opens the start date modal 
Screen Shot 2022-02-10 at 9 25 10 AM

The gif in the PR description may help clarify

PR #10315, opened 2020-11-03
Documentation for client count / activity log API.

mgritter:

PR #13971, opened 2022-02-09
Adding sys/internal/counters/activity endpoint

tonyp-hc:
The sys/internal/counters/activity endpoint was introduced in 1.6 but is only present in versioned docs from 1.7 onward. This edit ports the details back into our versioned docs for 1.6.

### DIFF
--- a/website/pages/api-docs/system/internal-counters.mdx
+++ b/website/pages/api-docs/system/internal-counters.mdx
@@ -130,3 +130,109 @@ $ curl \
   "auth": null
 }
 ```
+
+## Client Count
+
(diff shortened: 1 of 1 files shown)


tomhjp (commented):
Should this also add the client count concepts page: https://www.vaultproject.io/docs/v1.7.x/concepts/client-count? And maybe also the telemetry update that was included in #10315?

tonyp-hc:
Thanks @tomhjp for catching that. I included those updates and modified the docs nav to include the additional client count page.

tomhjp (approved):
Thanks! 👍

PR #13991, opened 2022-02-09
Allow specifying multiple allowed SSH key lengths

cipherboy:
In the ssh secrets engine, only a single allowed key length was allowed
for each algorithm type. However, many algorithms have multiple safe
values (such as RSA and ECDSA); allowing a single role to have multiple
values for a single algorithm is thus helpful.

On creation or update, roles can now specify multiple types using a list
or comma separated string of allowed values:

    allowed_user_key_lengths: map[string][]int{"rsa": []int{2048, 4096}}

---

As mentioned by @stevendpclark, there was some discussion of `map[string][]string` to support named curves, but with the exception of `x448` (which behaves like `x25519` and takes no length), support in cryptography libraries for other named curves besides the 4 we support (+ P-224 which we don't) isn't broad. There's little value in adding them and most all should be happy with the three NIST P-curves and x25519 (+/- x448 if/when we add it).

Since the PKI engine already takes curves-as-bits, I'm fine using this behavior. My 2c.

### DIFF
--- a/builtin/logical/ssh/path_sign.go
+++ b/builtin/logical/ssh/path_sign.go
@@ -447,7 +447,7 @@ func (b *backend) calculateTTL(data *framework.FieldData, role *sshRole) (time.D
 }
 
 func (b *backend) validateSignedKeyRequirements(publickey ssh.PublicKey, role *sshRole) error {
-	if len(role.AllowedUserKeyLengths) != 0 {
+	if len(role.AllowedUserKeyTypesLengths) != 0 {
 		var kstr string
 		var kbits int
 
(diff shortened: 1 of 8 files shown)


stevendpclark (approved):

sgmiller (changes requested):

website/content/api-docs/secret/ssh.mdx
```diff
@@ -210,8 +210,10 @@ This endpoint creates or updates a named role.
   '{{public_key_hash}}' - A SHA256 checksum of the public key that is being signed.
   e.g. "custom-keyid-{{token_display_name}}"
 
-- `allowed_user_key_lengths` `(map<string|int>: "")` – Specifies a map of ssh key types
+- `allowed_user_key_lengths` `(map<string|int|[]int>: "")` – Specifies a map of ssh key types
```
cipherboy, line 213: Should this type signature be `map<string|(int|[]int)>` instead perhaps? Not really sure how to indicate an optional list of int-like types.

builtin/logical/ssh/path_sign.go
```diff
@@ -473,31 +473,36 @@ func (b *backend) validateSignedKeyRequirements(publickey ssh.PublicKey, role *s
...
-				}
-			case "dsa":
-				if kbits == value {
-					pass = true
-				}
-			case "ecdsa":
-				if kbits == value {
+			for _, value := range allowed_values {
+				switch kstr {
+				case "rsa":
```
stevendpclark, line 480: Technically correct, but we are now incongruent on the key names we accept for "key types" from PR#14008 and what the names/values we allow for this key. Not the end of the world just pointing it out.

stevendpclark (approved):

cipherboy:
Thanks all!

Issue #11608, opened 2021-05-13
SSH CA: rsa-sha1 should not be default signing algorithm  


gargath:
**Describe the bug**
Modern-ish versions of sshd refuse SSH CA certificates signed with sha1. https://github.com/hashicorp/vault/pull/9096 added support for providing a `algorithm_signer` attribute to override this, but sha1 is still the default if this is not specified.

This means that following the documentation at https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates#signing-key-role-configuration will result in certificates that will be rejected by most hosts.

The documentation should outline the use of the `algorithm_signer` attribute and include it in the sample role definitions provided.
In addition, maybe `sha1` should no longer be the default?

**To Reproduce**
1. With a fresh, unsealed vault, follow documentation at https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates#signing-key-role-configuration
2. Note that Step 4 in the guide does not mention the `algorithm_signer`
3. Configure a recent ubuntu machine to trust the SSH CA
4. Sign a key with ault write ssh-client-signer/sign/<some_fole> public_key=@$HOME/.ssh/id_rsa.pub
5. Try to log into the ubuntu machine with the signed key and cert
6. Observe sshd logs to see error: `sshd[60694]: userauth_pubkey: certificate signature algorithm ssh-rsa: signature algorithm not supported [preauth]`

**Expected behavior**
The certificate should be signed with an algorithm that is supported by modern sshd implementations, e.g. `rsa-sha2-512`.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.7.0
* Vault CLI Version (retrieve with `vault version`):  v1.7.1 ('917142287996a005cb1ed9d96d00d06a0590e44e+CHANGES')
* Server Operating System/Architecture: client: MacOS Cataline (10.15.7), server: Ubuntu 20.04.2 LTS

**Additional context**
Using this role instead of the one in the documentation worked as expected:
```
$ vault write ssh-client-signer/roles/my-role -<<"EOH"
{
  "allow_user_certificates": true,
  "allowed_users": "*",
  "allowed_extensions": "permit-pty,permit-port-forwarding",
  "default_extensions": [
    {
      "permit-pty": ""
    }
  ],
  "key_type": "ca",
  "default_user": "ubuntu",
  "algorithm_signer": "rsa-sha2-512",
  "ttl": "30m0s"
}
EOH
```

[...]

PR #14006, opened 2022-02-10
Switch to secure signing algorithm for SSH secrets engine

cipherboy:
~Built on top of #13991; will be rebased once that merges.~

---

The default (empty) value for RSA keys, `ssh-rsa`, is an insecure default because it uses SHA-1 signing. This has lead to its removal from OpenSSH >= 8.2's default algorithm list. However, the more secure, modern signing algorithms for SSH keys, `rsa-sha2-256` and `rsa-sha2-512`, aren't available on platforms with OpenSSH < 7.2.

This PR takes an approach of migrating existing SSH roles using the default (empty) value to an explicit `ssh-rsa` value and changing the default to `default`, which can change over time as secure defaults change. This explicit `default` value presently uses `rsa-sha2-256`. Roles can of course still explicitly specify a value if desired.

Documentation has been clarified to point out that the legacy value is insecure, what the secure value presently is, and what will occur as a result of migration.

This is motivated by https://github.com/hashicorp/vault/issues/11608. 

Resolves: #11608

### DIFF
--- a/website/data/docs-nav-data.json
+++ b/website/data/docs-nav-data.json
@@ -1495,6 +1495,10 @@
         "path": "upgrading/plugins"
       },
       {
+        "title": "Upgrade to 1.10.x",
+        "path": "upgrading/upgrade-to-1.10.x"
+      },
+      {
         "title": "Upgrade to 1.9.x",
         "path": "upgrading/upgrade-to-1.9.x"
       },


website/content/api-docs/secret/ssh.mdx
```diff
@@ -210,14 +210,22 @@ This endpoint creates or updates a named role.
   '{{public_key_hash}}' - A SHA256 checksum of the public key that is being signed.
   e.g. "custom-keyid-{{token_display_name}}"
 
-- `allowed_user_key_lengths` `(map<string|int>: "")` – Specifies a map of ssh key types
+- `allowed_user_key_lengths` `(map<string|int|[]int>: "")` – Specifies a map of ssh key types
   and their expected sizes which are allowed to be signed by the CA type.
+  To specify multiple sizes, either use a comma-separated list or an array
+  of allowed key widths.
 
 - `algorithm_signer` `(string: "")` - Algorithm to sign keys with. Valid
```
cipherboy, line 218: should change to `default`

builtin/logical/ssh/path_roles.go
```diff
@@ -496,7 +499,7 @@ func (b *backend) pathRoleWrite(ctx context.Context, req *logical.Request, d *fr
 			Version:         roleEntryVersion,
 		}
 	} else if keyType == KeyTypeCA {
-		algorithmSigner := ""
+		algorithmSigner := DefaultAlgorithmSigner
```
stevendpclark, line 502: I might be mis-reading this code but at line 508 the case "": statement I would have assumed that we would have reset algorithmSigner back to DefaultAlgorithmSigner? Because we aren't I *think* we are writing out an empty value within the role's AlgorithmSigner with a V2 version

cipherboy: Ah, you're right that we to allow users to make this explicitly default. 

I'm not sure I follow though -- we set algorithmSigner to whatever was passed in the request _if something was passed_... Which might be `default` or it might be the empty string. Technically there's a subtle behavior difference between the empty string (library default) and the literal `default` (our Vault default -- preferring SHA2-256 in the RSA case). So it is valid to pass an empty string via the API IMO.

stevendpclark (approved):

builtin/logical/ssh/path_roles.go
cipherboy: I see, explicit is better than implicit. Let's change it here to explicitly be default and have consistent behavior between the two.

Issue #11035, opened 2021-03-03
SSH CA: generate_signing_key to support ed25519 keys


candlerb:
**Is your feature request related to a problem? Please describe.**
Vault's SSH CA is able to sign certificates using elliptic curve ciphers like ed25519, but cannot generate the private key.  If you set `generate_signing_key` to `true` (which is the default), you always get an RSA key.

Therefore, to use these key types, you have to generate the private key externally and import it.

RSA keys are problematic for SSH CAs:
* Modern versions of OpenSSH (8.2+) disable the `ssh-rsa` algorithm by default, as it's based on weak SHA1
* Only relatively recent versions (7.2+) support `rsa-sha2-256` and `rsa-sha2-512`. This excludes RHEL ≤ 6 and Debian ≤ 8, which are still widely deployed
* However, `ssh-ed25519` has been supported since 6.5, and `ecdsa-sha2-nistp256/384/521` since 5.7

This makes the elliptic curve ciphers much more widely applicable for a CA covering a mixed estate.

**Describe the solution you'd like**
A new parameter to [ssh/config/ca](https://www.vaultproject.io/api-docs/secret/ssh#submit-ca-information) to specify the key type when `generate_signing_key` is true.

For backwards compatibility, this can default to RSA.

(Arguably this would be a good place to specify other key parameters, particularly key size)

**Describe alternatives you've considered**
As mentioned, it is already possible to generate an elliptic curve key externally, and import it using the `private_key` and `public_key` fields.

However this exposes the CA private key on the machine where it is generated. Special care is required to ensure it does not persist on disk - e.g. by booting an ISO image with ramdisk only.

**Explain any additional use-cases**
N/A

**Additional context**
Relates to #10067, where a user tried to set `algorithm_signer` when signing an rsa user key with an ed25519 CA key. (Of course an RSA signing algorithm makes no sense when signing with an ed25519 CA key - even if the user key being signed is RSA)

Documentation is poor in this area, it's not even clear that you *can* use non-RSA keys for the CA.  `algorithm_signer` accepts only RSA-based signature types, enforced in code:

```
                        algorithmSigner = algorithmSignerRaw.(string)
                        switch algorithmSigner {
                        case ssh.SigAlgoRSA, ssh.SigAlgoRSASHA2256, ssh.SigAlgoRSASHA2512:
                        case "":
                                // This case is valid, and the sign operation will use the signer's
                                // default algorithm.
                        default:
[...]

PR #14008, opened 2022-02-10
Allow generation of other types of SSH CA keys

cipherboy:
Curious about general sentiment here; if it is positive, I'll add docs and a changelog and we can decide what to do about the UI (leave it as-is or drop it for future work).

See https://github.com/hashicorp/vault/issues/11035 for context. We derive types from the `x/crypto/ssh` types (which match OpenSSH's constants), thus making key bits only relevant for RSA keys. We're notably short exporting the private key (for parity with PKI secrets engine).

Resolves: #11035 

---

This probably needs some additional UI work, but it works :laughing:

@ivana-mcconnell or @hellobontempo do either you have any thoughts here? It seems like the UI of SSH CA generation isn't at all like the PKI functionality; it looks like there's no dynamic fields (the hbs template needed to be updated manually) and lacks a lot of the `actions` for groups or changes &c. I couldn't figure out how to use a `<Select />` component either; it appears to require an `onChange` action (which we appear to lack completely) and even without it, the `options` didn't populate (either manually or from the element's chosen values). Similarly failed via `<FormFieldFromModel />` -- that didn't even display at all. :/

**Edit**: Screenshot

Screenshot from 2022-02-14 10-09-32

### DIFF
--- a/builtin/logical/ssh/path_config_ca.go
+++ b/builtin/logical/ssh/path_config_ca.go
@@ -2,6 +2,10 @@ package ssh
 
 import (
 	"context"
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
@@ -44,6 +48,16 @@ func pathConfigCA(b *backend) *framework.Path {
 				Type:        framework.TypeBool,
 				Description: `Generate SSH key pair internally rather than use the private_key and public_key fields.`,
 				Default:     true,
+			},
+			"key_type": {
+				Type:        framework.TypeString,
+				Description: `Specifies the desired key type when generating; could be a OpenSSH key type identifier (ssh-rsa, ecdsa-sha2-nistp256, ecdsa-sha2-nistp384, ecdsa-sha2-nistp521, or ssh-ed25519) or an algorithm (rsa, ec, ed25519).`,
+				Default:     "ssh-rsa",
+			},
+			"key_bits": {
+				Type:        framework.TypeInt,
+				Description: `Specifies the desired key bits when generating variable-length keys (such as when key_type="ssh-rsa") or which NIST P-curve to use when key_type="ec" (256, 384, or 521).`,
+				Default:     0,
 			},
 		},
 
@@ -191,7 +205,10 @@ func (b *backend) pathConfigCAUpdate(ctx context.Context, req *logical.Request, 
 	}
 
 	if generateSigningKey {
-		publicKey, privateKey, err = generateSSHKeyPair(b.Backend.GetRandomReader())
+		keyType := data.Get("key_type").(string)
+		keyBits := data.Get("key_bits").(int)
+
+		publicKey, privateKey, err = generateSSHKeyPair(b.Backend.GetRandomReader(), keyType, keyBits)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


builtin/logical/ssh/path_config_ca_test.go
```diff
@@ -167,4 +168,132 @@ func TestSSH_ConfigCAUpdateDelete(t *testing.T) {
...
 	}
+
+	// Delete the configured keys
+	caReq.Operation = logical.DeleteOperation
+	resp, err = b.HandleRequest(context.Background(), caReq)
+	if err != nil || (resp != nil && resp.IsError()) {
+		t.Fatalf("bad: err: %v, resp:%v", err, resp)
+	}
+
+	// Successfully create a new RSA/2048 one
```
stevendpclark, line 179: Could you convert this to a table driven test to cut down on the overall code and make it clearer in the future if a test fails which algo is causing us issues?

builtin/logical/ssh/path_config_ca.go
```diff
@@ -265,22 +281,89 @@ func (b *backend) pathConfigCAUpdate(ctx context.Context, req *logical.Request,
...
-	privateSeed, err := rsa.GenerateKey(randomSource, 4096)
-	if err != nil {
-		return "", "", err
-	}
 
-	privateBlock := &pem.Block{
-		Type:    "RSA PRIVATE KEY",
-		Headers: nil,
-		Bytes:   x509.MarshalPKCS1PrivateKey(privateSeed),
+	var publicKey interface{}
```
stevendpclark, line 289:
```suggestion
	var publicKey crypto.PublicKey
```

builtin/logical/ssh/path_config_ca.go
```diff
@@ -265,22 +281,89 @@ func (b *backend) pathConfigCAUpdate(ctx context.Context, req *logical.Request,
...
+			return "", "", err
+		}
+
+		privateBlock = &pem.Block{
+			Type:    "RSA PRIVATE KEY",
+			Headers: nil,
+			Bytes:   x509.MarshalPKCS1PrivateKey(privateSeed),
+		}
+
+		publicKey = &privateSeed.PublicKey
```
stevendpclark, line 313: I'd rather us use the Public() method on the crypto algos as they behave a little differently across the various implementations :(

```suggestion
		publicKey = privateSeed.Public()
```

builtin/logical/ssh/path_config_ca.go
```diff
@@ -265,22 +281,89 @@ func (b *backend) pathConfigCAUpdate(ctx context.Context, req *logical.Request,
...
+			return "", "", err
+		}
+
+		privateBlock = &pem.Block{
+			Type:    "EC PRIVATE KEY",
+			Headers: nil,
+			Bytes:   marshalled,
+		}
+
+		publicKey = &privateSeed.PublicKey
```
stevendpclark, line 343:
```suggestion
		publicKey = privateSeed.Public()
```

stevendpclark (commented):
Overall +1 from me, a few little nits on the specific PR and I'll let others discuss/decide about the UI parts of this PR, hence the lack of approval from my part.

sgmiller (changes requested):
Looks good to me.  Needs docs and changelog though.

ivana-mcconnell:
@cipherboy Thanks! Just wanting to make sure I understand the desired behaviour here: we want to enable a user to select the key type first (from ssh-rsa, ecdsa-sha2-nistp256, ecdsa-sha2-nistp384, ecdsa-sha2-nistp521, or ssh-ed25519) and then, based on that, enter the key bits. Is that correct? 

Should key bits be an open input, or are there only specific values that are allowed for each key type?

cipherboy:
Hey @ivana-mcconnell!

I'd actually love to see something similar to those boxes in the PKI UI RFC doc where the user first selects if they're going to import a SSH CA (first two fields on that form) OR if they're going to generate a key. (Or put them under folding menus, you'd know better than me!).

In the former case, they just need the top two fields (private/public key).

In the latter case, they just need the two new fields. Yes, ideally a drop-down would be best for the key type; I tried getting that working with a select but failed. Presently key bits is only necessary when `keyType == ssh-rsa` so yes, could definitely be hidden otherwise :-) For now, its the standard RSA key sizes (which would either be an int or if you wanted some sane limits, values of 2048, 3072, 4096, and _maybe_ something larger but unlikely). Lemme know if you think there's some easy fixes I could do (and some existing code that does it right) or what :-)

hellobontempo:
> Hey @ivana-mcconnell!
> 
> I'd actually love to see something similar to those boxes in the PKI UI RFC doc where the user first selects if they're going to import a SSH CA (first two fields on that form) OR if they're going to generate a key. (Or put them under folding menus, you'd know better than me!).
> 
> In the former case, they just need the top two fields (private/public key).
> 
> In the latter case, they just need the two new fields. Yes, ideally a drop-down would be best for the key type; I tried getting that working with a select but failed. Presently key bits is only necessary when `keyType == ssh-rsa` so yes, could definitely be hidden otherwise :-) For now, its the standard RSA key sizes (which would either be an int or if you wanted some sane limits, values of 2048, 3072, 4096, and _maybe_ something larger but unlikely). Lemme know if you think there's some easy fixes I could do (and some existing code that does it right) or what :-)

Once design (@ivana-mcconnell) has put together designs for this I'm happy to review your PR, or pair through implementing in the UI

builtin/logical/ssh/path_config_ca.go
```diff
@@ -45,6 +49,16 @@ func pathConfigCA(b *backend) *framework.Path {
 				Description: `Generate SSH key pair internally rather than use the private_key and public_key fields.`,
 				Default:     true,
 			},
+			"key_type": {
+				Type:        framework.TypeString,
+				Description: `Specifies the desired key type when generating; must be ssh-rsa, ecdsa-sha2-nistp256, ecdsa-sha2-nistp384, ecdsa-sha2-nistp521, or ssh-ed25519.`,
```
cipherboy, line 54: Should accept all forms (rsa, ec, ssh-rsa, &c).

cipherboy:
fmt errors all in unrelated files.

Per discussion with @ivana-mcconnell, removed the fledgling UI :-)

stevendpclark (approved):

sgmiller (approved):

Issue #14005, opened 2022-02-10
OIDC Provider: "assignments" do not honor inherited groups


candlerb:
**Describe the bug**
Inherited groups are not honored in the groups list for an OIDC Provider client assignment.

**To Reproduce**
This requires an OIDC-capable application, in my case Grafana.

1. Create groups "staff" and "grafana-admin"
2. Make the "grafana-admin" group have "staff" as a member group
3. Create an OIDC provider assignment called "grafana" which lists the group_id for "grafana-admin"
4. Connect the application to the OIDC provider, using this assignment
5. Pick a user who is a member of the "staff" group, and try to login as that user.

```
$ vault read identity/group/name/grafana-admin
Key                  Value
---                  -----
...
id                   91326332-7b6c-ed59-f3f4-84607a410ce9
...

$ vault write identity/oidc/assignment/grafana \
    group_ids="91326332-7b6c-ed59-f3f4-84607a410ce9"

$ vault write identity/oidc/client/grafana \
    redirect_uris="https://xxxxxxxx:443/login/generic_oauth" \
    assignments="grafana" \
    key="oidc-key" \
    id_token_ttl="12h" \
    access_token_ttl="12h"
```

**Expected behavior**
Login should be permitted.

**Actual behavior**
It only works when the entity is a *direct* member of a group listed in the assignment (i.e. the "grafana-admin" group)

A user in the "staff" group cannot login, even though the "grafana-admin" group includes the "staff" group as a member, and the entity shows grafana-admin as an inherited group.

However, if I then add the entity as a direct member of "grafana-admin", I can login.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.9.3
* Vault CLI Version (retrieve with `vault version`): n/a
* Server Operating System/Architecture: Ubuntu 20.04 x86_64

Vault server configuration file(s): n/a

**Additional context**
n/a

candlerb:
Thank you for the very quick turnaround!

I have been running a proof-of-concept project for Vault as an IDP.  I am planning to write an experience report which highlights some areas where Vault could be improved for this application.  Are you interested in seeing this, and if so, how should I submit it?

austingebauer:
@candlerb - Happy to help! Absolutely, I'm very interested. We have some general UX improvements in the works, so I'd be super helpful to gather your feedback / discuss. 

Feel free to reach out to me directly: agebauer@hashicorp.com

PR #14013, opened 2022-02-11
identity/oidc: use inherited group membership for client assignments

austingebauer:
### Overview

This PR uses inherited group membership when evaluating client [assignments](https://www.vaultproject.io/api-docs/secret/identity/oidc-provider#create-or-update-an-assignment) during an OIDC flow.

Given the following group membership tree:
```
Group A
└── Group B
    └── Entity A
```

If Entity A authenticates through a client with an assignment that has Group A in its [group_ids](https://www.vaultproject.io/api-docs/secret/identity/oidc-provider#group_ids), the assignment membership check will be satisfied by group inheritance.

Fixes https://github.com/hashicorp/vault/issues/14005

### Testing

I added a test to this PR which sets up a similar scenario to the one explained above.

### DIFF
--- a/changelog/14013.txt
+++ b/changelog/14013.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+identity/oidc: Fixes inherited group membership when evaluating client assignments
+```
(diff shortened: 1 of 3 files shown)


fairclothjm (approved):

calvn (approved):
Looks good! Can we include a changelog entry for this bug fix?

Issue #10752, opened 2021-01-24
Consul Secrets Engine - Support Consul ACL Roles


BrandonIngalls:
**Is your feature request related to a problem? Please describe.**

The current [Consul Secrets Engine](https://www.vaultproject.io/api/secret/consul#create-update-role) `role` does not support Consul's [ACL Roles](https://www.consul.io/api-docs/acl/roles).

I can logically assign Consul policies to a Consul role, but right now I have to replicate that logic in Vault to mimic the Consul role I already created.

See hashicorp/terraform-provider-vault#951

**Describe the solution you'd like**

Allow creation of a token that is assigned roles and policies, not just policies.

**Describe alternatives you've considered**

* Make an API call to get a list of policies currently applied to the desired Consul role
    * **NOTE:** This would likely get out-of-sync pretty fast
    * The tokens vault creates will not get changes Consul admins make to Consul roles

**Explain any additional use-cases**

Consul supports ACL roles to help reduce duplication of work, and to make management of tokens much easier.

**Additional context**

n/a

dekimsey:
It's worth noting this would also help Consul Connect integration. Currently Service Identity (integrated, templated policies managed by Consul) aren't actual policies exposed to the user. They are only assigned internally to roles. This means when trying to configure Vault for Consul Connect services, one *cannot* use Service Identities to assign privileges. So now the operator is in the position of having to manually replicate these auto-managed policies.

Native role assignment would be desirable.

![image](https://user-images.githubusercontent.com/90741/110525971-df267f00-80da-11eb-8596-83d4cac94419.png)


```
./vault read consul/creds/magic
Error reading consul/creds/magic: Error making API request.

URL: GET https://vault-cluster.private.vault.1337.aws.hashicorp.cloud:8200/v1/consul/creds/magic
Code: 400. Errors:

* Unexpected response code: 500 (rpc error making call: rpc error making call: No such ACL policy with name "magic")
```

PR #11025, opened 2021-02-28
Consul acl role support

BrandonIngalls:
## About

Attempt to support Consul [ACL Roles](https://www.consul.io/api/acl/roles).

## Related

hashicorp/vault#10752

PR #14014, opened 2022-02-11
Consul acl role support

robmonte:
This is a refresh of the PR submitted by @BrandonIngalls here: https://github.com/hashicorp/vault/pull/11025 which adds support for Consul ACL roles. This simplifies management of Vault tokens that use multiple Consul policies. Previously, you needed to replicate the work of assigning multiple Consul policies to a role in both Vault and Consul.

For example, let's say you work on assigning 20 different ACL policies into a single Consul role named "admin-access". With current Vault, you'd need to assign all 20 polices again to a Vault role like this:
```
❯ vault write consul/roles/my-role policies="policy1,policy2,policy3,...,policy20"
```

With Consul role support, it changes to this:
```
❯ vault write consul/roles/my-role roles="admin-access"
```

Additionally, if you want to assign multiple Consul roles to a Vault role, you can:
```
❯ vault write consul/roles/my-role roles="admin-access,role2,role3,...,etc"
```

&nbsp;

The original author's commits are retained to preserve credit for the work generously provided, save for merge conflicts, and the CLA was previously signed here: https://github.com/hashicorp/vault/pull/11025#issuecomment-787511814

Closes https://github.com/hashicorp/vault/issues/10752, https://github.com/hashicorp/vault/pull/11025

### DIFF
--- a/builtin/logical/consul/path_roles.go
+++ b/builtin/logical/consul/path_roles.go
@@ -37,27 +37,29 @@ for 'client' tokens. Required for Consul pre-1.4.`,
 
 			"policies": {
 				Type: framework.TypeCommaStringSlice,
-				Description: `List of policies to attach to the token. Either policies or 
-roles are required for Consul 1.5 and above, or just policies if using Consul 1.4.`,
+				Description: `List of policies to attach to the token. Either "policies"
+or "consul_roles" are required for Consul 1.5 and above, or just "policies" if
+using Consul 1.4.`,
 			},
 
-			"roles": {
+			"consul_roles": {
 				Type: framework.TypeCommaStringSlice,
-				Description: `List of roles to attach to the token. Either policies or
-roles are required for Consul 1.5 and above.`,
+				Description: `List of Consul roles to attach to the token. Either "policies"
+or "consul_roles" are required for Consul 1.5 and above.`,
 			},
 
 			"local": {
 				Type: framework.TypeBool,
 				Description: `Indicates that the token should not be replicated globally 
-and instead be local to the current datacenter.  Available in Consul 1.4 and above.`,
+and instead be local to the current datacenter. Available in Consul 1.4 and above.`,
 			},
 
 			"token_type": {
 				Type:    framework.TypeString,
 				Default: "client",
 				Description: `Which type of token to create: 'client' or 'management'. If
-a 'management' token, the "policy" parameter is not required. Defaults to 'client'.`,
+a 'management' token, the "policy", "policies", and "consul_roles" parameters are not
+required. Defaults to 'client'.`,
 			},
 
 			"ttl": {
@@ -72,7 +74,7 @@ a 'management' token, the "policy" parameter is not required. Defaults to 'clien
 
 			"lease": {
 				Type:        framework.TypeDurationSecond,
-				Description: "Use ttl instead.",
+				Description: `Use "ttl" instead.`,
 				Deprecated:  true,
 			},
 
@@ -144,8 +146,8 @@ func (b *backend) pathRolesRead(ctx context.Context, req *logical.Request, d *fr
 	if len(roleConfigData.Policies) > 0 {
 		resp.Data["policies"] = roleConfigData.Policies
 	}
-	if len(roleConfigData.Roles) > 0 {
-		resp.Data["roles"] = roleConfigData.Roles
+	if len(roleConfigData.ConsulRoles) > 0 {
+		resp.Data["consul_roles"] = roleConfigData.ConsulRoles
 	}
 	return resp, nil
 }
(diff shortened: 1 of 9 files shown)


website/content/api-docs/secret/consul.mdx
```diff
@@ -81,13 +81,15 @@ updated attributes.
...
 
-- `policy` `(string: <policy or policies>)` – Specifies the base64 encoded ACL policy. The
-  ACL format can be found in the [Consul ACL
-  documentation](https://www.consul.io/docs/internals/acl). This is
-  required unless the `token_type` is `management`.
+- `policy` `(string: <policy>)` – Specifies the base64-encoded ACL policy. This is
+  required unless the `token_type` is `"management"`. [Deprecated as of Consul 1.4 and 
+  removed as of Consul 1.11.](https://www.consul.io/api/acl/legacy)
+
+- `policies` `(list: <policy or policies>)` – The list of policies to assign to the
```
calvn, line 88: Should we make the mention of "Either policies or roles are required for Consul 1.5 and above." in here as well?

builtin/logical/consul/path_roles.go
```diff
@@ -50,10 +56,8 @@ and instead be local to the current datacenter.  Available in Consul 1.4 and abo
 			"token_type": {
 				Type:    framework.TypeString,
 				Default: "client",
-				Description: `Which type of token to create: 'client'
-or 'management'. If a 'management' token,
-the "policy" parameter is not required.
-Defaults to 'client'.`,
+				Description: `Which type of token to create: 'client' or 'management'. If
+a 'management' token, the "policy" parameter is not required. Defaults to 'client'.`,
```
calvn, line 60:
```suggestion
a 'management' token, the "policy" and "role" parameters are not required. Defaults to 'client'.`,
```

builtin/logical/consul/path_roles.go
```diff
@@ -37,8 +37,14 @@ for 'client' tokens. Required for Consul pre-1.4.`,
 
 			"policies": {
 				Type: framework.TypeCommaStringSlice,
-				Description: `List of policies to attach to the token. Required
-for Consul 1.4 or above.`,
+				Description: `List of policies to attach to the token. Either policies or 
+roles are required for Consul 1.5 and above, or just policies if using Consul 1.4.`,
+			},
+
+			"roles": {
```
calvn, line 44: Similar to `consul_namespace`, I wonder if we should name this `consul_roles` so that the command feels a bit more explicit when specifying this.

```sh
$ vault write consul/roles/my-role roles=api-server
```

vs 

```sh
$ vault write consul/roles/my-role consul_roles=api-server
```

robmonte: Good catch. I meant to include a comment myself (but forgot...) to start this discussion.
I initially considered consul_roles, but used just roles to begin because there's no -role flag like there is with -namespace. That makes it a bit more separate than namespaces were between the two. However when looking at these side-by-side, it likely is better to use consul_roles to be certain there's no ambiguity down the line.

austingebauer (commented):
One comment, otherwise LGTM along with Calvin's suggestions. Great level of detail with the description and documentation 👍

builtin/logical/consul/path_roles.go
```diff
@@ -113,47 +117,51 @@ func (b *backend) pathRolesRead(ctx context.Context, req *logical.Request, d *fr
...
 }
 
 func (b *backend) pathRolesWrite(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
 	tokenType := d.Get("token_type").(string)
 	policy := d.Get("policy").(string)
 	policies := d.Get("policies").([]string)
-	if len(policies) == 0 {
+	roles := d.Get("roles").([]string)
+
+	if len(policies) == 0 && len(roles) == 0 {
```
austingebauer, line 159: I think we should bring this validation within the switch case for the "client" token type? It's possible to create a role with an invalid token type with this current validation. 

For example, something like this will be successful:
```
vault write consul/roles/consul-server-role roles="r1" token_type="invalid"
```

robmonte: Great catch, I think that this may be a preexisting bug that's being allowed for policies as well. I'll get it fixed.

builtin/logical/consul/path_roles.go
robmonte: Reworded it slightly more to mention both policy doc and the newer policies. Also added the same statement to the web docs, similar to your comment above. https://github.com/hashicorp/vault/pull/14014/commits/aa1901b0cb5e6f64ad77cd8f807b2b51195ac534

website/content/api-docs/secret/consul.mdx
robmonte: Sorry for the delay on getting this committed, I agree it should be mentioned here as well. Added it in the last commits. https://github.com/hashicorp/vault/pull/14014/commits/aa1901b0cb5e6f64ad77cd8f807b2b51195ac534

builtin/logical/consul/path_roles.go
robmonte: Changed here https://github.com/hashicorp/vault/pull/14014/commits/aa1901b0cb5e6f64ad77cd8f807b2b51195ac534

builtin/logical/consul/path_roles.go
robmonte: Fixed here: https://github.com/hashicorp/vault/pull/14014/commits/aa1901b0cb5e6f64ad77cd8f807b2b51195ac534#diff-b5a41064e6772dda37492b33baab6c5d795f30dcc60f9ad8e93adc324b312876

calvn (approved):
Looks great!

austingebauer (approved):
LGTM 👍

builtin/logical/consul/path_roles.go
benashz: @calvn @robmonte  Question, is the parameter naming scheme we are using here meant to disambiguate consul roles from vault roles? If so that seems to be inconsistent with other parameters like: [policies](https://www.vaultproject.io/api-docs/secret/consul#policies). I wonder if we should rename `policies` etc. to match what we are doing for `roles`.

robmonte: Yes, that was the goal. Namespaces look particularly confusing if we just used the word "namespace" like so:
`vault read -namespace="ns1" consul/creds/my-role namespace="ns2"`
`vault read -namespace="ns1" consul/creds/my-role consul_namespace="ns2"`
While Consul's roles aren't quite as confusing as that, I think it still helps avoid confusion to prefix "consul_" to it as well.

As for policies, I'm not sure if it would be better to change that after it's been in-use for such a long time.

benashz: Thanks @robmonte !

PR #14022, opened 2022-02-11
Some folder restructuring to help with route loading states

Monkeychip:
There was some issues we encountered with the routes not having access to the controller property set on the loading action. The only way around this appeared to be a reshuffle of folder/routing structure. Should make it easier to debug in the future.

* Index was more or less replaced by clients as the PARENT route. It's at vault.cluster instead of vault.cluster.clients. It uses {{outlet}} to render the child routes which are current, history, config and edit. 
* Made a route for current that only calls the monthly endpoint. 
* The children routes get access to the model information on the parent clients route by modelFor requesting data from that parent and adding it to their model. 

Here is the loading with a 3 second delay to history added to mirage.

https://user-images.githubusercontent.com/6618863/153654508-b65a59ce-7082-4c95-a70a-2cbf6c94a41e.mp4

Here are the delays without mirage (no flashing between routes).

https://user-images.githubusercontent.com/6618863/153654773-9a960f5a-5fc9-435e-bab0-f04ad3e02fa2.mp4

### DIFF
--- a/ui/app/components/clients/history.js
+++ b/ui/app/components/clients/history.js
@@ -141,7 +141,7 @@ export default class History extends Component {
       this.startTimeRequested = this.args.model.startTimeFromLicense;
       this.endTimeRequested = null;
     }
-    // clicked "Edit" Billing start month in Clients which opens a modal.
+    // clicked "Edit" Billing start month in History which opens a modal.
     if (dateType === 'startTime') {
       let monthIndex = this.arrayOfMonths.indexOf(month);
       this.startTimeRequested = [year.toString(), monthIndex]; // ['2021', 0] (e.g. January 2021)
--- a/ui/app/components/clients/history.js
+++ b/ui/app/components/clients/history.js
@@ -140,7 +140,7 @@ export default class History extends Component {
       this.startTimeRequested = this.args.model.startTimeFromLicense;
       this.endTimeRequested = null;
     }
-    // clicked "Edit" Billing start month in Dashboard which opens a modal.
+    // clicked "Edit" Billing start month in Clients which opens a modal.
     if (dateType === 'startTime') {
       let monthIndex = this.arrayOfMonths.indexOf(month);
       this.startTimeRequested = [year.toString(), monthIndex]; // ['2021', 0] (e.g. January 2021)
--- a/ui/app/components/clients/history.js
+++ b/ui/app/components/clients/history.js
@@ -60,6 +60,7 @@ export default class History extends Component {
   // TEMPLATE MESSAGING
   @tracked noActivityDate = '';
   @tracked responseRangeDiffMessage = null;
+  @tracked isLoadingQuery = false;
 
   // on init API response uses license start_date, getter updates when user queries dates
   get getActivityResponse() {
@@ -154,6 +155,7 @@ export default class History extends Component {
     }
 
     try {
+      this.isLoadingQuery = true;
       let response = await this.store.queryRecord('clients/activity', {
         start_time: this.startTimeRequested,
         end_time: this.endTimeRequested,
@@ -183,6 +185,8 @@ export default class History extends Component {
       this.queriedActivityResponse = response;
     } catch (e) {
       return e;
+    } finally {
+      this.isLoadingQuery = false;
     }
   }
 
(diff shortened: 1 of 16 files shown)


ui/app/routes/vault/cluster/clients/history.js
```diff
@@ -26,24 +25,6 @@ export default class HistoryRoute extends Route {
     }
   }
 
-  async getVersionHistory() {
```
Monkeychip, line 29: this all comes now from the clients.js route, and is added to the routes model by modelFor

ui/app/templates/components/clients/history.hbs
```diff
@@ -212,8 +212,8 @@
         type="button"
         class="button is-primary"
         onclick={{queue
-          (action "handleClientActivityQuery" this.startMonth this.startYear "startTime")
           (action (mut this.isEditStartMonthOpen) false)
+          (action "handleClientActivityQuery" this.startMonth this.startYear "startTime")
```
Monkeychip, line 216: reverse so that the modal closes first then the try happens and the loader is added other wise loader will be showing and the modal will be open.

ui/app/components/clients/history.js
```diff
@@ -140,7 +141,7 @@ export default class History extends Component {
       this.startTimeRequested = this.args.model.startTimeFromLicense;
       this.endTimeRequested = null;
     }
-    // clicked "Edit" Billing start month in Dashboard which opens a modal.
+    // clicked "Edit" Billing start month in Clients which opens a modal.
```
hellobontempo, line 144: should this be in "History"?

ui/app/templates/vault/cluster/clients/config.hbs
```diff
@@ -1,6 +1,4 @@
 {{! TODO CMB CHECK THIS ENTIRE VIEW WITH DESIGN }}
```
hellobontempo, line 1: We can remove this comment! :)

ui/app/templates/vault/cluster/clients.hbs
```diff
@@ -10,7 +9,7 @@
 <div class="tabs-container box is-bottomless is-marginless is-fullwidth is-paddingless" data-test-pricing-metrics>
   <nav class="tabs">
     <ul>
-      <LinkTo @route="vault.cluster.clients.index" class="nav-tab-link">
+      <LinkTo @route="vault.cluster.clients.current" class="nav-tab-link">
```
hellobontempo, line 12: Since we deleted the `clients.index` route we need to update the Status dropdown menu link in `cluster-info.hbs`

Monkeychip: Yep, just got that myself!

ui/app/routes/vault/cluster/clients/current.js
```diff
@@ -0,0 +1,14 @@
+import Route from '@ember/routing/route';
+import RSVP from 'rsvp';
+
+export default class CurrentRoute extends Route {
+  async model() {
+    let parentModel = this.modelFor('vault.cluster.clients');
```
zofskeez, line 6: :tada:

ui/app/routes/vault/cluster/clients/history.js
zofskeez: At first I was wondering about this but now that I see the data is shared between both child routes so nice work loading it in the parent and accessing it this way!

ui/app/templates/components/clients/history.hbs
zofskeez: Not a request for change but a suggestion that rather than using the `queue` helper here it might be cleaner to mutate the property value in the `handleClientActivityQuery` instead.

zofskeez (approved):
Nice work on this! I like the restructuring of the routes and glad you get the loading state issues resolved :rocket:

ui/app/controllers/vault/cluster/clients/current.js
```diff
@@ -0,0 +1,3 @@
+import Controller from '@ember/controller';
```
hellobontempo, line 1: Since we removed all the `loading` actions from the child route (right?) Can we remove empty controllers for the children?

I _think_ I was getting errors when I had tried doing this because I was calling `controllerFor` in the loading function...but i'm not sure

zofskeez: Good call @hellobontempo these should be good to remove now.

PR #14024, opened 2022-02-11
Fix max measurements gauge test

ccapurso:
We have a theory that the amount of metrics emitted in `TestGauge_MaximumMeasurements` is causing the test to timeout in CircleCI. It seems like we can exercise the same functionality with a fraction of the original metric count. A constant for the batch size has also been added for clarification when debugging the test and associated underlying  `streamGaugesToSink` function.

### DIFF
--- a/helper/metricsutil/gauge_process.go
+++ b/helper/metricsutil/gauge_process.go
@@ -197,16 +197,18 @@ func (p *GaugeCollectionProcess) collectAndFilterGauges() {
 	p.streamGaugesToSink(values)
 }
 
+// batchSize is the number of metrics to be sent per tick duration.
+const batchSize = 25
+
 func (p *GaugeCollectionProcess) streamGaugesToSink(values []GaugeLabelValues) {
 	// Dumping 500 metrics in one big chunk is somewhat unfriendly to UDP-based
 	// transport, and to the rest of the metrics trying to get through.
 	// Let's smooth things out over the course of a second.
-	// 1 second / 500 = 2 ms each, so we can send 25 per 50 milliseconds.
+	// 1 second / 500 = 2 ms each, so we can send 25 (batchSize) per 50 milliseconds.
 	// That should be one or two packets.
 	sendTick := p.clock.NewTicker(50 * time.Millisecond)
 	defer sendTick.Stop()
 
-	batchSize := 25
 	for i, lv := range values {
 		if i > 0 && i%batchSize == 0 {
 			select {
--- a/helper/metricsutil/gauge_process_test.go
+++ b/helper/metricsutil/gauge_process_test.go
@@ -433,11 +433,11 @@ func TestGauge_MaximumMeasurements(t *testing.T) {
 		2000000*time.Hour)
 
 	sink := NewClusterMetricSink("test", inmemSink)
-	sink.MaxGaugeCardinality = 500
+	sink.MaxGaugeCardinality = 100
 	sink.GaugeInterval = 2 * time.Hour
 
 	// Create a report larger than the default limit
-	excessGauges := 100
+	excessGauges := 20
 	values := makeLabels(sink.MaxGaugeCardinality + excessGauges)
 	rand.Shuffle(len(values), func(i, j int) {
 		values[i], values[j] = values[j], values[i]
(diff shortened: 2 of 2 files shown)


helper/metricsutil/gauge_process.go
```diff
@@ -197,16 +197,18 @@ func (p *GaugeCollectionProcess) collectAndFilterGauges() {
 	p.streamGaugesToSink(values)
 }
 
+// batchSize is the number of metrics to be sent per tick duration.
+const batchSize = 25
```
swayne275, line 201: lovely

helper/metricsutil/gauge_process.go
```diff
@@ -197,16 +197,18 @@ func (p *GaugeCollectionProcess) collectAndFilterGauges() {
...
+
 func (p *GaugeCollectionProcess) streamGaugesToSink(values []GaugeLabelValues) {
 	// Dumping 500 metrics in one big chunk is somewhat unfriendly to UDP-based
```
swayne275, line 204: i know this was already here, but where is the 500 coming from?

swayne275 (approved):
nice cleanup! big question about a comment that was already there though (and seems to be the basis of the 25 items per tick

helper/metricsutil/gauge_process.go
hghaf099: Yeah, I have the same question, cause that essentially affects the batchsize

hghaf099: seems to be from here:
https://github.com/hashicorp/vault/blob/5f590d4d505dcb7f0021893c23ce2ab49f204d11/internalshared/configutil/telemetry.go#L29

helper/metricsutil/gauge_process.go
hghaf099: I think this should depend on the default value here:
https://github.com/hashicorp/vault/blob/5f590d4d505dcb7f0021893c23ce2ab49f204d11/internalshared/configutil/telemetry.go#L29

ccapurso: I just lifted the local var (`batchSize := 25`) from `streamGuagesToSink` and made it a package-level const.

helper/metricsutil/gauge_process.go
ccapurso: 500 was the original `MaxGaugeCardinality` for the sink in the test. We reduced that to 100.

See:
https://github.com/hashicorp/vault/blob/7fd37fe778eb9ff187beb9af023d5fa1e5f50360/helper/metricsutil/gauge_process_test.go#L436

hghaf099: I think we may need to think what would happen if `MaximumGaugeCardinalityDefault` configured differently that the default value.

ccapurso: I believe that `MaximumGaugeCardinalityDefault` comes into play if the telemetry configuration doesn't specify `MaximumGaugeCardinality`. It isn't configurable.

https://github.com/hashicorp/vault/blob/5f590d4d505dcb7f0021893c23ce2ab49f204d11/internalshared/configutil/telemetry.go#L212-L214

hghaf099: Yes, sorry I am not talking about the test though. The test is perfectly fine. 
I am saying it seems the calculations in this function seem to be all based on the assumption that `MaximumGaugeCardinality` is always 500 which is the default value. So, if that is configured differently, do we want to keep the constant values in this function as is? `MaximumGaugeCardinality` could be configured to be 1000 for example.

ccapurso: Oh that's an interesting point! Attempting to fix this test together has brought up some interesting discussion. It looks like `MaximumGaugeCardinality ` is used in the calling function `collectAndFilterGauges`. Having a constant batch size that is not calculated based on the `MaxGaugeCardinality` means that transmission rate does not scale as the max increases. There obviously needs to be a max batch size to address the concerns discussed in the comment but your point is starting to make me wonder if this hard-coded batch size is warranted or if any flexibility can be introduced. I'm going to dig into the logic a bit more closely now that you mentioned this.

ccapurso: Yup, I agree. I think we should do that as a separate PR though. I'd rather allow the necessary time to investigate it more thoroughly and provide appropriate test coverage. I can make a ticket with context from our discussion here.

PR #14027, opened 2022-02-11
Remove --privileged recommendation from entrypoint script message

mickael-hc:
The use of the `--privileged` [flag](https://docs.docker.com/engine/reference/commandline/run/#options) is not recommended, as it does not adhere to the principle of least privilege.


calvn:
Sounds reasonable, but wanted to get @jasonodonnell 's input here to see if there was any particular reason for having this flag set in the past.

jasonodonnell (approved):

PR #14042, opened 2022-02-14
Address slow CI causing failures in TestRateLimitQuota_Allow_WithBlock

stevendpclark:
- An attempt to fix CI runs that are extremely slow and the for loop
   runs across two BlockIntervals within the rate limit window of operation.
 - Increasing BlockInterval was looked at but the normal test times would
   be increased due to us also validating that we are releasing clients post
   BlockInterval.

```
=== RUN   TestRateLimitQuota_Allow_WithBlock
    quotas_rate_limit_test.go:194: 
        	Error Trace:	quotas_rate_limit_test.go:194
        	Error:      	Not equal: 
        	            	expected: 17
        	            	actual  : 33
        	Test:       	TestRateLimitQuota_Allow_WithBlock
--- FAIL: TestRateLimitQuota_Allow_WithBlock (5.08s)
```

### DIFF
--- a/vault/quotas/quotas_rate_limit_test.go
+++ b/vault/quotas/quotas_rate_limit_test.go
@@ -146,11 +146,16 @@ func TestRateLimitQuota_Allow_WithBlock(t *testing.T) {
 
 	var wg sync.WaitGroup
 
-	reqFunc := func(addr string, atomicNumAllow, atomicNumFail *atomic.Int32) {
+	reqFunc := func(addr string, atomicNumAllow, atomicNumFail *atomic.Int32, start time.Time) {
 		defer wg.Done()
 
 		resp, err := rlq.allow(context.Background(), &Request{ClientAddress: addr})
 		if err != nil {
+			return
+		}
+
+		// Avoid slow CI passing the BlockInterval of the rate limiter.
+		if time.Now().After(start.Add(rlq.BlockInterval)) {
 			return
 		}
 
@@ -177,7 +182,7 @@ func TestRateLimitQuota_Allow_WithBlock(t *testing.T) {
 				cr = results[addr]
 			}
 
-			go reqFunc(addr, cr.atomicNumAllow, cr.atomicNumFail)
+			go reqFunc(addr, cr.atomicNumAllow, cr.atomicNumFail, start)
 
 			time.Sleep(2 * time.Millisecond)
 		}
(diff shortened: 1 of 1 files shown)


swayne275 (approved):

vault/quotas/quotas_rate_limit_test.go
```diff
@@ -146,14 +146,19 @@ func TestRateLimitQuota_Allow_WithBlock(t *testing.T) {
...
+	reqFunc := func(addr string, atomicNumAllow, atomicNumFail *atomic.Int32, start time.Time) {
 		defer wg.Done()
 
 		resp, err := rlq.allow(context.Background(), &Request{ClientAddress: addr})
 		if err != nil {
 			return
 		}
 
+		// Avoid slow CI passing the BlockInterval of the rate limiter.
+		if time.Now().After(start.Add(rlq.BlockInterval)) {
```
vishalnayak, line 158: I think this condition will never be met.

`start.Add(rlq.BlockInterval)` will be "now+10seconds" whereas the for loop which triggers this function runs for "now+5seconds".

This also makes me wonder if the test fix in this PR is passing due to sheer chance?

vishalnayak: Please let me know if I misread the fix.

vault/quotas/quotas_rate_limit_test.go
```diff
@@ -191,7 +192,7 @@ func TestRateLimitQuota_Allow_WithBlock(t *testing.T) {
 
 		// Since blocking is enabled, each client should only have 'rate' successful
```
ncabatoff, line 193: Could I ask you to make another change while you're here?  My read of this test is: over a 5s period, every 2ms, it spawns 5 goroutines, each with a different (fake) ClientAddress.  In other words, a total of 5 different IPs are seen as generating one request every 2ms.  We want to ensure that those IPs will only get 17 successful requests each during the interval.

My concern is that this amounts to a total of 500 goroutines per IP per second, or 12500 goroutines.  Of course normally we won't have anywhere near that many, because they should complete quickly.  Still, I feel we could protect ourselves against having too many easily enough by making the waitgroup local to the inner for loop, so we never have more than 5 goroutines in-flight at any time.

vishalnayak: You mean, "run 5, wait for 5" every 2ms? I guess that should work.

stevendpclark: Done.

stevendpclark:
Rebased the changesets onto the tip of main as a lot has happened since this PR was originally created. No issues/conflicts with the rebase.

vishalnayak (approved):

ncabatoff (approved):

PR #14049, opened 2022-02-14
MFA

zofskeez:
This feature adds multi-factor authentication support to the UI. If the user is required to take further action to satisfy the an mfa requirement they will be presented a form, otherwise for push notifications the standard login form will wait until the mfa requirements are fulfilled and display a message to the user.

image
image


hashishaw (approved):
Nice work! I pulled down the branch and was playing with the various logins, but I didn't see the error timer on a failed passcode. I might have set it up wrong though. Either way, this is a great start and we can follow on with updates!

zofskeez:
> Nice work! I pulled down the branch and was playing with the various logins, but I didn't see the error timer on a failed passcode. I might have set it up wrong though. Either way, this is a great start and we can follow on with updates!

For now I commented out that code and show an error message until I can confirm if the backend is returning a specific error for that case.

ui/app/services/auth.js
```diff
@@ -321,16 +325,98 @@ export default Service.extend({
...
     const [options] = arguments;
     const adapter = this.clusterAdapter();
+    let resp;
+
+    try {
+      resp = await adapter.authenticate(options);
+    } catch (e) {
+      // check for totp not configured mfa error before throwing
+      const errors = this.handleError(e);
+      // stubbing error - verify once API is finalized
```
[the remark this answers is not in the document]
zofskeez: Good call I updated the comment

ui/mirage/handlers/mfa.js
```diff
@@ -0,0 +1,146 @@
+import { Response } from 'miragejs';
```
Monkeychip, line 1: @zofskeez do we need the mirage handler now that the API is done, or will this be useful later for any testing or debugging?

ui/tests/acceptance/mfa-test.js
```diff
@@ -0,0 +1,135 @@
...
+    await click('[data-test-mfa-validate]');
+    didLogin(assert);
+  });
+
+  test('it should handle multiple mfa constraints with 1 passcode method each', async function (assert) {
+    await login('mfa-f');
+    assert
+      .dom('[data-test-mfa-description]')
+      .includesText(
+        '2 methods are required for successful authentication.',
```
Monkeychip, line 95: not sure where this is in the code, but if possible it seems like this should be a written two instead of 2. Any numbers under 10 are generally written out.

Monkeychip (approved):
A couple of comments/questions but nothing blocking. Nice work!

ui/tests/acceptance/mfa-test.js
hashishaw: I think a helper for this would be useful! Angel is right, usually numbers under 10 are wordy

PR #14050, opened 2022-02-14
Client count: remove default period and extra controllers

Monkeychip:
A small PR that removes the default display on the config page. This PR also removes extra controllers that were left over from the nested folder work in my last [PR](https://github.com/hashicorp/vault/pull/14022).

Before:
image

After:
image

### DIFF
--- a/ui/app/templates/components/clients/config.hbs
+++ b/ui/app/templates/components/clients/config.hbs
@@ -1,16 +1,47 @@
-{{#if @isLoading}}
-  <LayoutLoading />
-{{else}}
-  {{#if (eq @mode "edit")}}
-    <form onsubmit={{action "onSaveChanges"}} data-test-pricing-metrics-config-form>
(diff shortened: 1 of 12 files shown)


ui/app/templates/components/clients/config.hbs
```diff
@@ -1,16 +1,47 @@
-{{#if @isLoading}}
```
Monkeychip, line 1: removed @isLoading

ui/app/templates/components/clients/current.hbs
```diff
@@ -45,25 +45,21 @@
         </DocLink>
       </AlertBanner>
     {{/if}}
-    {{#if @isLoading}}
```
Monkeychip, line 48: removed isLoading

hashishaw (approved):
Nice work! "Hide whitespace" saved me here 😅

PR #14054, opened 2022-02-14
UI: Switch usage of localStorage to sessionStorage

hashishaw:
Swap out use of localStorage in favor of sessionStorage across the webapp.

### DIFF
--- a/ui/app/lib/local-storage.js
+++ b/ui/app/lib/local-storage.js
@@ -1,18 +1,18 @@
 export default {
   getItem(key) {
-    var item = window.localStorage.getItem(key);
+    var item = window.sessionStorage.getItem(key);
     return item && JSON.parse(item);
   },
 
   setItem(key, val) {
-    window.localStorage.setItem(key, JSON.stringify(val));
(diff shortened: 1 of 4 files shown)


zofskeez (approved):
Looks good thanks!

zerkms:
Why was it implemented? It causes this bug now: https://github.com/hashicorp/vault/issues/15311

danmasta:
I think this has an unintended side affect of breaking all OAuth/ Single Sign On type workflows because once you close the tab you are effectively signed out. If you generate a token on sign in with a 30 day refresh for example it should stay logged in even if you close the tab or use other tabs.

We have to now generate lots of extra tokens which is not how OAuth should be used. Also if you use MFA with yubikeys or something it's really frustrating to have to do that flow so many extra times.

zerkms:
> If you generate a token on sign in with a 30 day refresh for example it should stay logged in even if you close the tab or use other tabs.

The token would be still "legit", but UI would behave the same: it still would use the session storage not local storage :shrug:

PR #14061, opened 2022-02-14
Fix format issues on current `main`

cipherboy:

---

Looks like a few `make fmt` issues snuck into main :-) 

See also: #14008


ccapurso:
Oops, sorry about that and it looks like we duplicated the work to fix them. See https://github.com/hashicorp/vault/pull/14062.

cipherboy:
All good @ccapurso :)

PR #14064, opened 2022-02-14
Tool Tip Styling Issue

zofskeez:
This PR addresses classes not being applied to the _BasicDropdownContent_ component which was noticed in various tooltips. The _Content_ component yielded by _BasicDropdown_ does not use splattributes but exposes a [defaultClass arg which is applied to the BasicDropdownOptionalTag wrapper component.](https://github.com/cibernox/ember-basic-dropdown/blob/6e529ca0766eb40f23ae2828387eb7a12ce3aa7b/addon/templates/components/basic-dropdown-content.hbs#L17) A codemod was previously created to apply some transformations to _BasicDropdown_ components which was updated and run against hbs files to change the `class` attribute to `@defaultClass` arg.

### DIFF
--- a/changelog/14064.txt
+++ b/changelog/14064.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixes tooltip styling issue
+```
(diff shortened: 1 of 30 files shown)


zofskeez:
Good call thanks for bringing that up @Monkeychip. I double checked on 1.9.3 just to be sure and the tool tips are styled appropriately so no backport needed and I will remove the changelog entry.

PR #14066, opened 2022-02-15
Ui/ Clients error handling

hellobontempo:
In addition to adding an `error.hbs` for routing errors, this PR tackle some specific cases: 
- When a user attempts to query an invalid start date (since this is historical data, start date cannot be in the future, or the present month)
no-billing-error

- When a user doesn't have permission to view the `/activity`, `/activity/monthly` or `/license` endpoints
Screen Shot 2022-02-15 at 4 43 31 PM
Screen Shot 2022-02-15 at 4 44 49 PM


ui/app/components/clients/history.js
```diff
@@ -38,10 +38,15 @@ export default class History extends Component {
...
     return new Date().getFullYear() - i;
   });
+  currentDate = new Date();
+  currentYear = this.currentDate.getFullYear(); // integer of year
+  currentMonth = this.currentDate.getMonth(); // index of month
 
   @tracked isEditStartMonthOpen = false;
   @tracked startMonth = null;
   @tracked startYear = null;
+  @tracked allowedMonthMax = 12;
```
Monkeychip, line 48: Curious about this. The name would imply they can maybe only return 12 months of data. However, some licenses go two years. Not sure if that's how it's being used, but thought it worth mentioning.

ui/app/templates/components/date-dropdown.hbs
```diff
@@ -36,7 +37,12 @@
     <nav class="box menu">
       <ul class="menu-list">
         {{#each this.years as |year|}}
-          <button type="button" class="link" {{on "click" (queue (fn this.selectStartYear year) (action D.actions.close))}}>
+          <button
+            type="button"
+            class="button link"
+            disabled={{if (eq year this.disabledYear) true false}}
+            {{on "click" (queue (fn this.selectStartYear year) (action D.actions.close))}}
```
Monkeychip, line 44: if you don't want to use queue, you can pass D.actions or just D into this.selectStartYear as a param and then call it in the function. e.g. function(D.actions) { code code ... D.actions.close}

ui/app/templates/vault/cluster/clients/error.hbs
```diff
@@ -0,0 +1,29 @@
...
+{{else}}
+  <EmptyState
+    @title={{if (eq @model.httpStatus 403) "You are not authorized" "Error"}}
+    @subTitle={{concat "Error " @model.httpStatus}}
+    @icon="skip"
+  >
+    {{#if (eq @model.httpStatus 403)}}
+      <p>
+        You must be granted permissions to view this page. Ask your administrator if you think you should have access to the
+        <code>{{@model.path}}</code>
```
Monkeychip, line 12: nice!

ui/app/components/clients/history.js
```diff
@@ -231,11 +237,23 @@ export default class History extends Component {
   @action
   selectStartMonth(month) {
     this.startMonth = month;
+    // disables months if in the future
+    this.months.indexOf(month) >= this.currentMonth
```
zofskeez, line 241: It might be more clear to rewrite this so the ternary is on the right hand side of the statement to determine the conditional value.
`this.disabledYear = this.months.indexOf(month) >= this.currentMonth ? this.currentYear : null;`

ui/app/components/date-dropdown.js
```diff
@@ -28,11 +34,22 @@ export default class DateDropdown extends Component {
   @action
   selectStartMonth(month) {
     this.startMonth = month;
+    // disables months if in the future
+    this.months.indexOf(month) >= this.currentMonth
```
zofskeez, line 38: Same as previous suggestion :slightly_smiling_face:

ui/app/templates/components/date-dropdown.hbs
zofskeez: :point_up:. I think we should avoid using queue when possible.

zofskeez (approved):
Nice work! There are a couple of suggestions to look at but this looks good.

ui/app/components/clients/history.js
hellobontempo: good call! thank you

PR #14067, opened 2022-02-15
api/client: forward and inconsistent header const 

tvoran:
Adds constants for `X-Vault-Forward` and `X-Vault-Inconsistent` headers to api/client.go, so we don't need to redefine them when using the client outside of vault. 

Related to https://github.com/hashicorp/vault-lambda-extension/pull/58


tomhjp (approved):

digivava (approved):
Makes sense to me!

PR #14068, opened 2022-02-15
UI support for Consul secret engine

mr-miles:
This PR provides support for the consul secret engine so that users can configure roles and generate credentials via the UI, by following the necessary bits of the aws engine code.

It does not support configuring the engine from scratch - the api does not return enough detail to make it straightforward.


hsimon-hashicorp:
Hello! Is this in relation to an open issue?

mr-miles:
Sorry, I got sidetracked by the two contributors issue and haven't written enough useful information.

I've linked an issue

hsimon-hashicorp:
Thanks! Please don't forget the changelog entry and any tests, as necessary. :)

mr-miles:
ok i have sorted out the contributor names so now it matches the CLA.  What's the test policy for the UI - it seemed quite variable?  I'm a novice at ember so if you can point me to a part of the codebase to copy that'd be helpful.  Otherwise I've run it against our uat vault instance and got the desired effects (add/edit/delete/list roles + generate credentials for a given consul 
secret engine)

mr-miles:
Hi - could you give me some direction on what's left to get this merged pls?

hashishaw:
Thank you for your patience! Taking another look this week 👀

hsimon-hashicorp:
Hi there! We recognize the opportunity for improvement here, and we appreciate you filing this PR to help fill a needed option. This really highlighted a gap for our UI. Our engineering teams have decided on a more interdisciplinary approach, with our frontend and backend engineers working with product on a robust solution for the Consul secret engine in the UI. I don't have any guarantees on when it would be released, but do "watch this space".

Our engineers also had the following feedback:
- The form to create a new role was broken if the type was "manager" because it hid the required policies field
- The form to create a new role doesn't include all of the options in our API docs
- We have a pattern for generating credentials, but your approach was very creative and we got a kick out of the help text!
- Consul's custom config endpoint is unique, so the required fields in our regular secret engine configure page does not actually configure it correctly. We will need to handle this in a new way from other secret engines.

As I said, we really appreciate the amount of time and effort that went into this PR. While we will be tackling this on a larger scale internally, I do hope this doesn't discourage you from making other contributions to the repo in the future. :)

PR #14069, opened 2022-02-15
Update documentation to reduce confusion about default_extensions.

### LINKED ISSUE #10958: [Documentation] ssh ca default_extensions: map, or list of maps?
**Describe the issue**
The documentation for [api/secret/ssh](https://www.vaultproject.io/api/secret/ssh#parameters-2) says that `default_extensions` for an ssh role is a `map<string|string>` with default None.  There is an example later in the page:

```
  "default_extensions": {},
```

However, the documentation at [secrets/ssh/signed-ssh-certificates](https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates#signing-key-role-configuration) shows it as a *list of maps*:

```
  "default_extensions": [
    {
      "permit-pty": ""
    }
  ],
```

Put together, this implies there could be three different ways to set multiple extensions:

*Case 1*

```
  "default_extensions": {
    "permit-pty": "",
    "permit-agent-forwarding": ""
  },

```

*Case 2*

```
  "default_extensions": [
    {
      "permit-pty": "",
      "permit-agent-forwarding": ""
    }
  ],
```

*Case 3*

```
  "default_extensions": [
    {
      "permit-pty": ""
    },
    {
      "permit-agent-forwarding": ""
    }
  ],
```

**Empirical answer (tested with vault v1.6.2)**

All three appear to work.  In all three cases, retrieving the role with `vault read` shows the value as a simple map:

```
$ vault read ssh-user-ca/roles/ssh-test
Key                         Value
---                         -----
...
default_extensions          map[permit-agent-forwarding: permit-pty:]
...
```

*Aside:* as expected, a plain list in the form `"permit-pty,permit-agent-forwarding"` or `["permit-pty","permit-agent-forwarding"]` is rejected:

```
	* error converting input permit-pty,permit-agent-forwarding for field "default_extensions": '' expected a map, got 'string'
	* error converting input [permit-pty permit-agent-forwarding] for field "default_extensions": '[0]' expected a map, got 'string'
```

**Code analysis**

There are test cases for both forms: e.g. [here](https://github.com/hashicorp/vault/blob/master/builtin/logical/ssh/backend_test.go#L1100) and [here](https://github.com/hashicorp/vault/blob/master/builtin/logical/ssh/backend_test.go#L804) (although with only one extension, so they don't distinguish case 2 and case 3).

I couldn't find the code which merges the list of maps into a single map.  I note that Vault uses [github.com/mitchellh/mapstructure](https://github.com/mitchellh/mapstructure) but I wasn't able to replicate this behaviour - [1](https://play.golang.org/p/LGAiho1nelc), [2](https://play.golang.org/p/R9OX8K4DLHf).

**Suggestion**

I don't see any particular reason to wrap the options in a list, so my suggestion is to remove the "list of maps" form from the documentation - *or* to document in the API that both forms are acceptable when creating or updating an ssh role.

spacex:
Fixes #10958

### DIFF
--- a/website/content/docs/secrets/ssh/signed-ssh-certificates.mdx
+++ b/website/content/docs/secrets/ssh/signed-ssh-certificates.mdx
@@ -118,11 +118,9 @@ team, or configuration management tooling.
       "allow_user_certificates": true,
       "allowed_users": "*",
       "allowed_extensions": "permit-pty,permit-port-forwarding",
-      "default_extensions": [
-        {
-          "permit-pty": ""
-        }
-      ],
+      "default_extensions": {
+        "permit-pty": ""
+      },
       "key_type": "ca",
       "default_user": "ubuntu",
       "ttl": "30m0s"
@@ -401,11 +399,9 @@ this extension to the signed certificate.
   ```text
   $ vault write ssh-client-signer/roles/my-role -<<"EOH"
   {
-    "default_extensions": [
-      {
-        "permit-pty": ""
-      }
-    ]
+    "default_extensions": {
+      "permit-pty": ""
+    }
     // ...
   }
   EOH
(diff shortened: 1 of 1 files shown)


hsimon-hashicorp:
Looks good to me, also requesting a review from @taoism4504 as our docs lead!

cipherboy (approved):

cipherboy:
Looks good, thanks @spacex :D

PR #14078, opened 2022-02-15
Return integral seconds for transit auto_rotate_interval instead of a time string.

schultz-is:
# Overview
This PR changes the type of transit's `auto_rotate_interval` from a time string (e.g. `"1h2m3s") to integral seconds as used elsewhere in vault.

# Testing
- `vault secrets enable transit`
- `vault write -f transit/keys/my-key auto_rotate_interval=1h2m3s`
- `vault read transit/keys/my-key`
- Observe the type of the returned `"auto_rotate_interval"` field. Output should look something like this:
```json
{
  "request_id": "7724661b-7f62-7c7b-3d9d-d4e719a5a3b0",
  "lease_id": "",
  "renewable": false,
  "lease_duration": 0,
  "data": {
    "allow_plaintext_backup": false,
    "auto_rotate_interval": 3723,
...
```

### DIFF
--- a/builtin/logical/transit/path_keys.go
+++ b/builtin/logical/transit/path_keys.go
@@ -238,7 +238,7 @@ func (b *backend) pathPolicyRead(ctx context.Context, req *logical.Request, d *f
 			"supports_decryption":    p.Type.DecryptionSupported(),
 			"supports_signing":       p.Type.SigningSupported(),
 			"supports_derivation":    p.Type.DerivationSupported(),
-			"auto_rotate_interval":   p.AutoRotateInterval.String(),
+			"auto_rotate_interval":   int64(p.AutoRotateInterval.Seconds()),
 		},
 	}
 
--- a/builtin/logical/transit/path_config_test.go
+++ b/builtin/logical/transit/path_config_test.go
@@ -3,6 +3,7 @@ package transit
 import (
 	"context"
 	"encoding/hex"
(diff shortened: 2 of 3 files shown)


builtin/logical/transit/path_keys.go
```diff
@@ -238,7 +238,7 @@ func (b *backend) pathPolicyRead(ctx context.Context, req *logical.Request, d *f
 			"supports_decryption":    p.Type.DecryptionSupported(),
 			"supports_signing":       p.Type.SigningSupported(),
 			"supports_derivation":    p.Type.DerivationSupported(),
-			"auto_rotate_interval":   p.AutoRotateInterval.String(),
+			"auto_rotate_interval":   int64(p.AutoRotateInterval.Seconds()),
```
kalafut, line 241: Are we set on "_interval"? If we use "_period" then the CLI will format the output into a human-friendly timestamp.

schultz-is: Nope, no attachments from me! That suffix will likely help a decent amount in this case by removing a couple orders of magnitude; most of the values for this will likely be on the order of days, weeks, or months.

schultz-is: I'll submit a separate PR with the renaming of this field to keep this changeset small.

stevendpclark (approved):
lgtm!

PR #14093, opened 2022-02-15
UI/client count auth filter

hashishaw:
This PR adds an auth mount selector to the filter bar for both current and history tabs of client counts

current-month-auth-filter
history-auth-mount-filter


ui/mirage/handlers/activity.js
```diff
@@ -20,10 +20,15 @@ export default function (server) {
     };
   });
 
-  server.get('sys/internal/counters/config', function (db) {
+  server.get('sys/internal/counters/config', function () {
     return {
       request_id: '00001',
-      data: db['clients/configs'].first(),
+      data: {
+        default_report_months: 12,
```
Monkeychip, line 27: you can remove the default_report_months as we don't display this anymore (I removed it in a recent PR)

Monkeychip (approved):
Nice work. Looks really clean.

ui/mirage/handlers/activity.js
hashishaw: It does still come back from the API, so I'm inclined to leave it for as close parity to the real API as we can get. But good to know we don't display, another thing to add to the tests! :D

PR #14098, opened 2022-02-16
validate cipher length before decrypting

swayne275:
This addresses TOB-016.

I'm not sure if erroring when the `len(ciphertext) < 4` is revealing any important information to a potential attacker, so please let me know. As is, it would panic (which is less directly revealing the same information).

This will be backported back to 1.7.x

### DIFF
--- a/vault/barrier_aes_gcm.go
+++ b/vault/barrier_aes_gcm.go
@@ -1014,6 +1014,9 @@ func termLabel(term uint32) []metrics.Label {
 
 // decrypt is used to decrypt a value using the keyring
 func (b *AESGCMBarrier) decrypt(path string, gcm cipher.AEAD, cipher []byte) ([]byte, error) {
+	if len(cipher) < 5+gcm.NonceSize() {
+		return nil, fmt.Errorf("invalid cipher length")
+	}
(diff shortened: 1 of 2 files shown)


cipherboy (commented):
Do we also need `b.decrypt` to check length? It kinda looks like it doesn't:

https://github.com/hashicorp/vault/blob/fe5c8db54cc4c62b0187e4b3840f1f84a99fccf7/vault/barrier_aes_gcm.go#L1015-L1035

swayne275:
@cipherboy it looks like it should as well

ccapurso (approved):

mladlow:
@swayne275 if you'd like you can use backport-assistant to automatically create the backport PRs for stuff like this in the future. There's a google doc on how to do this, which I'll eventually move to the wiki.

swayne275:
@mladlow oh awesome! last i'd heard my understanding was that was only for docs changes. i had been using https://docs.google.com/document/d/1uQDcq2gd-rO4PHeK9fEEZ5qauGJVsEUrG57il3mOcQw/edit#heading=h.1n8tmqhs7xas

PR #14101, opened 2022-02-16
Fix ed25519 generated SSH key marshalling

cipherboy:
This updates the generation test to ensure that, for each generated CA key type, we can sign certificates with this key pair. This ensures that they're stored in a format understood by `golang.org/x/crypto/ssh`.  

Notably, this caught a bug in the Ed25519 key handling: OpenSSH (and `golang.org/x/crypto/ssh`) do not understand the PKCS8 PrivateKey structure we were using. As mentioned in an internal RFC (TF 461), and with investigation from [a `terraform-provider-tls` PR author](https://github.com/hashicorp/terraform-provider-tls/pull/59#issuecomment-568442395), it appears that [using `edkey`](https://github.com/hashicorp/terraform-provider-tls/pull/59/files#diff-8f3863add37ed2fd994775bd64132dedfdb47902bb54edcef278bcec98c12cf0R109) is the best approach. 

As this feature hasn't been released, we shouldn't need a changelog entry for it.

### DIFF
--- a/builtin/logical/ssh/path_config_ca.go
+++ b/builtin/logical/ssh/path_config_ca.go
@@ -10,6 +10,7 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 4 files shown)


schultz-is (approved):
lgtm

cipherboy:
Thanks @schultz-is!

PR #14102, opened 2022-02-16
Cleanup bool checks

swayne275:
cleanup (bool) == true and (bool) == false checks (where it didn't seem to harm readability)

this is the nittiest of nits, so not bummed if we think this isn't useful.

### DIFF
--- a/plugins/database/influxdb/connection_producer.go
+++ b/plugins/database/influxdb/connection_producer.go
@@ -257,7 +257,7 @@ func isUserAdmin(cli influx.Client, user string) (bool, error) {
 	for _, res := range response.Results {
 		for _, serie := range res.Series {
 			for _, val := range serie.Values {
-				if val[0].(string) == user && val[1].(bool) == true {
+				if val[0].(string) == user && val[1].(bool) {
 					return true, nil
 				}
 			}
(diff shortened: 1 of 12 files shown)


plugins/database/influxdb/connection_producer.go
```diff
@@ -257,7 +257,7 @@ func isUserAdmin(cli influx.Client, user string) (bool, error) {
 	for _, res := range response.Results {
 		for _, serie := range res.Series {
 			for _, val := range serie.Values {
-				if val[0].(string) == user && val[1].(bool) == true {
+				if val[0].(string) == user && val[1].(bool) {
```
ccapurso, line 260: Obviously not the intention of your PR, but couldn't the type assertions here panic?

ccapurso (commented):
Looks good to me. Just one note on a line that was changed that might already be panic-prone.

plugins/database/influxdb/connection_producer.go
ccapurso: Eh, it's probably fine. As in many DB drivers, the columns and values are split across separate fields. Eeach entry of `Row.Values` is `[]interface{}` thus the type assertions. I have not used influxdb but it looks like a `SHOW USERS` query would output something like:

```
SHOW USERS
user    admin
----    -----
admin   true
```

Feel free to ignore me :)

ccapurso (approved):

PR #14103, opened 2022-02-16
Rename transit's auto_rotate_interval to auto_rotate_period

schultz-is:
# Overview
This renames the `auto_rotate_interval` field on transit keys to `auto_rotate_period` for consistency’s sake and to get CLI pretty printing for free. Since these values will likely be on the order of days, weeks, or months, changing this will knock a couple orders of magnitude off of the CLI output (e.g. `3600` becomes `“1h”`). A separate UI fix is included to account for this field recently changing types from a time string to integral seconds.

Issue: VAULT-5128

# Testing
## CLI
- `vault secrets enable transit`
- `vault write -f transit/keys/test-key auto_rotate_period=1h`
- `vault read transit/keys/test-key`
- Observe the `auto_rotate_interval` field is now `auto_rotate_period` and formatted as a time string on the CLI (`"1h"`)
- `curl -H 'X-Vault-Token: $VAULT_TOKEN' $VAULT_ADDR/v1/transit/keys/test-key`
- Observe `auto_rotate_interval` field is now `auto_rotate_period` and formatted as integral seconds (`3600`)
## UI
- Enable transit secrets engine
- Create a transit key with an auto rotate period
- View the key and observe its auto rotate period is named and displayed correctly
- Create a transit key without an auto rotate period
- View the key and observe its auto rotate period is named and displayed correctly

### DIFF
--- a/builtin/logical/transit/path_config.go
+++ b/builtin/logical/transit/path_config.go
@@ -49,7 +49,7 @@ the latest version of the key is allowed.`,
 				Description: `Enables taking a backup of the named key in plaintext format. Once set, this cannot be disabled.`,
 			},
 
-			"auto_rotate_interval": {
+			"auto_rotate_period": {
 				Type: framework.TypeDurationSecond,
 				Description: `Amount of time the key should live before
 being automatically rotated. A value of 0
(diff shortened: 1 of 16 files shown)


builtin/logical/transit/path_config.go
```diff
@@ -193,19 +193,19 @@ func (b *backend) pathConfigWrite(ctx context.Context, req *logical.Request, d *
 		}
 	}
 
-	autoRotateIntervalRaw, ok, err := d.GetOkErr("auto_rotate_interval")
+	autoRotatePeriodRaw, ok, err := d.GetOkErr("auto_rotate_period")
```
stevendpclark, line 196: Possibly outside the scope of this PR, but should we handle the case where err == nil and ok is false? See https://github.com/hashicorp/vault/blob/main/sdk/framework/field_data.go#L218

Or are we okay with just ignoring a provided nil value of "auto_rotate_period" and keep on going?

schultz-is: In this case we're ok continuing with a nil value as the field is not required for a config update. We only need to act when a valid value is provided by updating and persisting the new, provided value.

stevendpclark (approved):
👍

hashishaw (approved):
Looks great! Thanks for making these changes

PR #14118, opened 2022-02-17
UI/format-duration error fix

schultz-is:
# Overview
When the output type of a transit key's `auto_rotate_period` changed, its template formatting helper also changed. Switching from `format-ttl` to `format-duration` navigating to the key details tab and then to the top-level transit backend page resulted in a value of `undefined` being passed to the template helper. This resulted in a parsing error and would prevent further navigation. This PR updates the return value of `format-duration` to mimic that of `format-ttl` when an invalid input is provided.

## Old Behavior
old

## New Behavior
new

# Testing
## UI (Manual)
- Enable the transit secrets engine
- Create a transit key
- Navigate to the newly created transit key's "Details" tab
- Attempt to navigate to the transit secrets engine in the breadcrumb bar

## UI (Automated)
- `cd ui && yarn run test -s -f="transit backend: "`

### DIFF
--- a/ui/app/helpers/format-duration.js
+++ b/ui/app/helpers/format-duration.js
@@ -1,5 +1,4 @@
 import { helper } from '@ember/component/helper';
-import { assert } from '@ember/debug';
 import { formatDuration, intervalToDuration } from 'date-fns';
 
 export function duration([time], { removeZero = false }) {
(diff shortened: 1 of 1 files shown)


hashishaw (approved):
Thank you for the quick turnaround!

PR #14121, opened 2022-02-17
Add methods with context propagation to api

### LINKED ISSUE #14070: Support context propagation in http clients
In the case of long requests, it is impossible to terminate them via context cancellation.

Add methods with context propagation, e.g. 
- `ReadContext(ctx context.Context, path string) (*Secret, error)`
- `ListContext(ctx context.Context, path string) (*Secret, error)`
- `WriteContext(ctx context.Context, path string, data map[string]interface{}) (*Secret, error)`
- etc.

No alternatives in the context of Golang best practices.

It can take a long time for a certificate to be generated in the Vault. It may be necessary to abort the request and terminate the job correctly.

**Additional context**
No change of existing methods/signatures is required. New methods will be added. The only difference is using the context passed to the method as a parameter instead of `context.Background()`.

ilyakaznacheev:
A fistful of methods with context was added to API. This will allow canceling long requests when the context is canceled, e.g. by a timeout or by the graceful shutdown.

Closes #14070

### DIFF
--- a/api/auth_token.go
+++ b/api/auth_token.go
@@ -15,13 +15,17 @@ func (a *Auth) Token() *TokenAuth {
 }
 
 func (c *TokenAuth) Create(opts *TokenCreateRequest) (*Secret, error) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	return c.CreateContext(ctx, opts)
+}
+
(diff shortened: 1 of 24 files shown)


ilyakaznacheev:
Oh shi, I branched from `master`...

ilyakaznacheev:
#14151 created instead

PR #14136, opened 2022-02-17
WIP Azure Auth Against Multiple Identities

imthaghost:
In this PR I added a small helper function that will instead only attempt login with a given client and object ID. 

We initially make a call to IDMS to get the access token 
``` bash
curl -H Metadata:true "http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fmanagement.azure.com/&client_id=<UAMI CLIENT ID>"
```

We can then use only the **ClientID** and **ObjectID** to specify a particular identity. 
https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview

### DIFF
--- a/command/agent/auth/azure/azure.go
+++ b/command/agent/auth/azure/azure.go
@@ -7,9 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 
-	cleanhttp "github.com/hashicorp/go-cleanhttp"
 	hclog "github.com/hashicorp/go-hclog"
-	"github.com/hashicorp/vault/api"
 	"github.com/hashicorp/vault/command/agent/auth"
 	"github.com/hashicorp/vault/sdk/helper/jsonutil"
 	"github.com/hashicorp/vault/sdk/helper/useragent"
@@ -27,9 +25,10 @@ const (
 type azureMethod struct {
 	logger    hclog.Logger
 	mountPath string
-
-	role     string
-	resource string
+	clientID  string
+	objectID  string
+	role      string
+	resource  string
 }
 
 func NewAzureAuthMethod(conf *auth.AuthConfig) (auth.AuthMethod, error) {
@@ -73,7 +72,7 @@ func NewAzureAuthMethod(conf *auth.AuthConfig) (auth.AuthMethod, error) {
 	return a, nil
 }
 
-func (a *azureMethod) Authenticate(ctx context.Context, client *api.Client) (retPath string, header http.Header, retData map[string]interface{}, retErr error) {
+func (a *azureMethod) Authenticate(ctx context.Context) (retPath string, header http.Header, retData map[string]interface{}, retErr error) {
 	a.logger.Trace("beginning authentication")
 
 	// Fetch instance data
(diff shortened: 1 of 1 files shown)


command/agent/auth/azure/azure.go
```diff
@@ -73,7 +72,7 @@ func NewAzureAuthMethod(conf *auth.AuthConfig) (auth.AuthMethod, error) {
 	return a, nil
 }
 
-func (a *azureMethod) Authenticate(ctx context.Context, client *api.Client) (retPath string, header http.Header, retData map[string]interface{}, retErr error) {
+func (a *azureMethod) Authenticate(ctx context.Context) (retPath string, header http.Header, retData map[string]interface{}, retErr error) {
```
tvoran, line 75: I think this needs the client parameter to satisfy the [AuthMethod interface](https://github.com/hashicorp/vault/blob/74a613545e2e1baef8517e30ab5595e5ab4da1fc/command/agent/auth/auth.go#L22-L31), which is returned from NewAzureAuthMethod() above.

But since it's not used in this function, maybe just use a blank identifier?
```suggestion
func (a *azureMethod) Authenticate(ctx context.Context, _ *api.Client) (retPath string, header http.Header, retData map[string]interface{}, retErr error) {
```

command/agent/auth/azure/azure.go
```diff
@@ -176,3 +175,33 @@ func getMetadataInfo(ctx context.Context, endpoint, resource string) ([]byte, er
 
 	return body, nil
 }
+
+func verifyIdentity(ctx context.Context) (retPath string, header http.Header, retData map[string]interface{}, retErr error) {
```
calvn, line 179: This function doesn't seem to be called anywhere, where should this be called? Does the return values need to be named here?

command/agent/auth/azure/azure.go
```diff
@@ -176,3 +175,33 @@ func getMetadataInfo(ctx context.Context, endpoint, resource string) ([]byte, er
...
+		}
+	}
+
+	body, err := getMetadataInfo(ctx, instanceEndpoint, "")
+	if err != nil {
+		retErr = err
+		return
+	}
+
+	err = jsonutil.DecodeJSON(body, &instance)
```
calvn, line 200: It looks like we decode the JSON body into instance, but the object is scoped to this function, should that be returned instead? I also see that `retData` remains empty and not assigned, should this be set somewhere?

tvoran:
If I'm reading https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http right, it looks like we just need to set the `client_id` and/or `object_id` query parameter when `getMetadataInfo()` is called for the identity endpoint. So we could pass those through to `getMetadataInfo()` and set them if they're not blank (similar to how the resource parameter is handled now).

(Since that would make the args list a bit long, we may want to pass resource, client_id, and object_id to `getMetadataInfo()` in a simple args struct too.)

austingebauer:
+1 on @tvoran suggestion. We only need to do this for the `/metadata/identity/oauth2/token` API and not for the `metadata/instance` API.

imthaghost:
This was completed in PR https://github.com/hashicorp/vault/pull/14214

PR #14139, opened 2022-02-17
Calendar widget test and small styling fix on wide screens

Monkeychip:
This PR is the component test for the calendar widget. It also fixes an issue of display on wider screens.

### DIFF
--- a/ui/tests/integration/components/calendar-widget-test.js
+++ b/ui/tests/integration/components/calendar-widget-test.js
@@ -1,6 +1,6 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render, click, fillIn } from '@ember/test-helpers';
+import { render, click } from '@ember/test-helpers';
 import sinon from 'sinon';
 import hbs from 'htmlbars-inline-precompile';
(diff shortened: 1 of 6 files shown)


ui/tests/integration/components/calendar-widget-test.js
```diff
@@ -0,0 +1,123 @@
...
+        @handleCurrentBillingPeriod={{handleCurrentBillingPeriod}}
+        @startTimeDisplay={{"February 2021"}}
+      />
+    `);
+
+    await calendarDropdown.openCalendar();
+    assert.ok(calendarDropdown.showsCalendar, 'renders the calendar component');
+  });
+
+  test('it does not allow a user to click to a future year but does allow a user to click to previous year', async function (assert) {
```
hashishaw, line 49: It's a little unclear on this one which value passed in it's basing on "previous" vs "future". I think having all the values be different dates might help make that apparent

hashishaw (approved):
👏

ui/tests/integration/components/calendar-widget-test.js
Monkeychip: Good idea. I'll amend.

PR #14143, opened 2022-02-17
Allow OpenSSH-style key type identifiers

cipherboy:
To bring better parity with the changes of #14008, wherein we allowed
OpenSSH-style key identifiers during generation. When specifying a list
of allowed keys, validate against both OpenSSH-style key identifiers
and the usual simplified names as well ("rsa" or "ecdsa"). Notably, the
PKI secrets engine prefers "ec" over "ecdsa", so we permit both as well.


stevendpclark (approved):

cipherboy:
Thanks all!

PR #14147, opened 2022-02-17
MFA Error Handling

zofskeez:
Changes to handle mfa validation error scenarios.

General passcode validation failure
image

Passcode has already been used
image

Push method failure
image


Monkeychip (approved):
Nice work. Looks a lot cleaner with this structure.

Issue #4973, opened 2018-07-23
Empty api.Response.Body returned when Content-Length is non-zero


ebilling:
**Describe the bug**
In an integration test, I started noticing about a month ago that repeated requests would occasionally return an error.  After much digging, I found that the api.Response.Body contained no data, while the Content-Length was correctly set.  

After much more digging, I discovered the problem.  In function:

func (c *Client) RawRequest(r *Request) (*Response, error) {

.........

	var result *Response
	resp, err := client.Do(req)
	if cancelFunc != nil {
		cancelFunc()
	}
	if resp != nil {
		result = &Response{Response: resp}
	}

should be:

	if cancelFunc != nil {
		defer cancelFunc()
	}
	var result *Response
	resp, err := client.Do(req)
	if resp != nil {
		result = &Response{Response: resp}
	}

see documentation for WithTimeout:

// WithTimeout returns WithDeadline(parent, time.Now().Add(timeout)).
//
// Canceling this context releases resources associated with it, so code should
// call cancel as soon as the operations running in this Context complete:
//
// 	func slowOperationWithTimeout(ctx context.Context) (Result, error) {
// 		ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
// 		**defer cancel()**  // releases resources if slowOperation completes before timeout elapses
// 		return slowOperation(ctx)
// 	}

**To Reproduce**
1.) Using the go API, set up a loop that reads the same secret repeatedly with a low timeout (I used 1 second) on the same system that's running the instance of vault you are testing against.
2.) Track whether the secret is returned.

**Expected behavior**
If the data has been returned to the client, it should be returned.

**Environment:**
* Vault Server Version (retrieve with `vault status`): Version         0.10.3
* Vault CLI Version (retrieve with `vault version`): Vault v0.10.3 (cgo)
* Server Operating System/Architecture: MacOS and Linux

jefferai:
The problem with deferring is that we support retries so these could pile up. However, it's good to know that canceling a context will actually wipe the data. The documentation is frustratingly vague for what "releases sesources associated with it means" since that's going to depend strongly on individual package behavior.

jefferai:
Can you check out the `cancelshift` branch and see if that fixes things for you?

ebilling:
I’ll try it, but I don’t think it will work.  The cancelFunc call needs to be deferred.  Otherwise, you have race conditions for when it is actually called, leading to unpredictable behavior, especially under load.

[...]

PR #14152, opened 2022-02-18
Add methods with context propagation to api

### LINKED ISSUE #14070: Support context propagation in http clients

ilyakaznacheev:
A fistful of methods with context was added to API. This will allow canceling long requests when the context is canceled, e.g. by a timeout or by the graceful shutdown.

Closes #14070

### DIFF
--- a/api/auth_token.go
+++ b/api/auth_token.go
@@ -27,7 +27,7 @@ func (c *TokenAuth) CreateWithContext(ctx context.Context, opts *TokenCreateRequ
 		return nil, err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -49,7 +49,7 @@ func (c *TokenAuth) CreateOrphanWithContext(ctx context.Context, opts *TokenCrea
 		return nil, err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -71,7 +71,7 @@ func (c *TokenAuth) CreateWithRoleWithContext(ctx context.Context, opts *TokenCr
 		return nil, err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -95,7 +95,7 @@ func (c *TokenAuth) LookupWithContext(ctx context.Context, token string) (*Secre
 		return nil, err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -119,7 +119,7 @@ func (c *TokenAuth) LookupAccessorWithContext(ctx context.Context, accessor stri
 		return nil, err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -138,7 +138,7 @@ func (c *TokenAuth) LookupSelfWithContext(ctx context.Context) (*Secret, error) 
 
 	r := c.c.NewRequest("GET", "/v1/auth/token/lookup-self")
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -163,7 +163,7 @@ func (c *TokenAuth) RenewAccessorWithContext(ctx context.Context, accessor strin
 		return nil, err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -188,7 +188,7 @@ func (c *TokenAuth) RenewWithContext(ctx context.Context, token string, incremen
 		return nil, err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -212,7 +212,7 @@ func (c *TokenAuth) RenewSelfWithContext(ctx context.Context, increment int) (*S
 		return nil, err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -240,7 +240,7 @@ func (c *TokenAuth) RenewTokenAsSelfWithContext(ctx context.Context, token strin
 		return nil, err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -267,7 +267,7 @@ func (c *TokenAuth) RevokeAccessorWithContext(ctx context.Context, accessor stri
 		return err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return err
 	}
@@ -294,7 +294,7 @@ func (c *TokenAuth) RevokeOrphanWithContext(ctx context.Context, token string) e
 		return err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return err
 	}
@@ -317,7 +317,7 @@ func (c *TokenAuth) RevokeSelfWithContext(ctx context.Context, token string) err
 
 	r := c.c.NewRequest("PUT", "/v1/auth/token/revoke-self")
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return err
 	}
@@ -345,7 +345,7 @@ func (c *TokenAuth) RevokeTreeWithContext(ctx context.Context, token string) err
 		return err
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
 	if err != nil {
 		return err
 	}
--- a/api/auth_token.go
+++ b/api/auth_token.go
@@ -15,13 +15,17 @@ func (a *Auth) Token() *TokenAuth {
 }
 
 func (c *TokenAuth) Create(opts *TokenCreateRequest) (*Secret, error) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	return c.CreateContext(ctx, opts)
+}
+
+func (c *TokenAuth) CreateContext(ctx context.Context, opts *TokenCreateRequest) (*Secret, error) {
 	r := c.c.NewRequest("POST", "/v1/auth/token/create")
 	if err := r.SetJSONBody(opts); err != nil {
 		return nil, err
 	}
 
-	ctx, cancelFunc := context.WithCancel(context.Background())
-	defer cancelFunc()
 	resp, err := c.c.RawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
@@ -32,13 +36,17 @@ func (c *TokenAuth) Create(opts *TokenCreateRequest) (*Secret, error) {
 }
 
 func (c *TokenAuth) CreateOrphan(opts *TokenCreateRequest) (*Secret, error) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	return c.CreateOrphanContext(ctx, opts)
+}
+
+func (c *TokenAuth) CreateOrphanContext(ctx context.Context, opts *TokenCreateRequest) (*Secret, error) {
 	r := c.c.NewRequest("POST", "/v1/auth/token/create-orphan")
 	if err := r.SetJSONBody(opts); err != nil {
 		return nil, err
 	}
 
-	ctx, cancelFunc := context.WithCancel(context.Background())
-	defer cancelFunc()
 	resp, err := c.c.RawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
@@ -49,13 +57,17 @@ func (c *TokenAuth) CreateOrphan(opts *TokenCreateRequest) (*Secret, error) {
 }
 
 func (c *TokenAuth) CreateWithRole(opts *TokenCreateRequest, roleName string) (*Secret, error) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	return c.CreateWithRoleContext(ctx, opts, roleName)
+}
+
+func (c *TokenAuth) CreateWithRoleContext(ctx context.Context, opts *TokenCreateRequest, roleName string) (*Secret, error) {
 	r := c.c.NewRequest("POST", "/v1/auth/token/create/"+roleName)
 	if err := r.SetJSONBody(opts); err != nil {
 		return nil, err
 	}
 
-	ctx, cancelFunc := context.WithCancel(context.Background())
-	defer cancelFunc()
 	resp, err := c.c.RawRequestWithContext(ctx, r)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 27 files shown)


ilyakaznacheev:
Created instead of #14121.

api/auth_token.go
```diff
@@ -15,13 +15,17 @@ func (a *Auth) Token() *TokenAuth {
 }
 
 func (c *TokenAuth) Create(opts *TokenCreateRequest) (*Secret, error) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	return c.CreateContext(ctx, opts)
+}
+
+func (c *TokenAuth) CreateContext(ctx context.Context, opts *TokenCreateRequest) (*Secret, error) {
```
VinnyHC, line 23: A tiny nit here, can you update these to `<method>WithContext(...)` to match the existing pattern  `RawRequestWithContext`.

VinnyHC (changes requested):
Thank you for the submission; this looks good! Since this increases our API surface by a bit, I'll need some input from other team members which I hope to have done shortly. In the meantime, I have a tiny nit to keep these consistent with the other patterns in our codebase.
We've settled on using `WithContext` as opposed to `Context`. I know both patterns are used throughout Go's std lib but since these are calling `RawRequestWithContext` we'd like to keep it consistent.
Thanks again, let us know if there's any questions!

api/auth_token.go
ilyakaznacheev: No probled

ilyakaznacheev: Done

api/auth_token.go
```diff
@@ -15,13 +15,17 @@ func (a *Auth) Token() *TokenAuth {
...
+
+func (c *TokenAuth) CreateWithContext(ctx context.Context, opts *TokenCreateRequest) (*Secret, error) {
 	r := c.c.NewRequest("POST", "/v1/auth/token/create")
 	if err := r.SetJSONBody(opts); err != nil {
 		return nil, err
 	}
 
-	ctx, cancelFunc := context.WithCancel(context.Background())
-	defer cancelFunc()
 	resp, err := c.c.RawRequestWithContext(ctx, r)
```
averche, line 29: I believe the pattern of calling `context.WithCancel`  was a bit of a hack to prevent context leaks introduced in `RawRequestWithContext`:

https://github.com/hashicorp/vault/blob/8b1e041c61ba7ec61c85584fa5fcaa6f6cc0704f/api/client.go#L1160-L1166

`RawRequestWithContext()` currently requires a cancellable context, which our current code does by passing it `context.WithCancel(context.Background())`. Unfortunately, moving these lines into the wrapper will make `CreateWithContext` (and other `__WithContext` methods) prone to context leaks:

```go
// client code: OK
ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
defer cancel()
c.CreateWithContext(ctx, opts)
```
```go
// client code: POSSIBLE CONTEXT LEAK
c.CreateWithContext(context.Background(), opts)
```
In the second example, if a user has a timeout set at the client level but the timeout is never hit, the context will never be cancelled and the resources will be leaked.

One way I can think of to solve this issue is to move the timeout logic out of `RawRequestWithContext` and into each individual `___WithContext`  function:
```go
func (c *TokenAuth) Create(opts *TokenCreateRequest) (*Secret, error) {
	return c.CreateWithContext(context.Background(), opts)
}

func (c *TokenAuth) CreateWithContext(ctx context.Context, opts *TokenCreateRequest) (*Secret, error) {
    // ...
    
    if c.c.config.Timeout != 0 {
        var cancelFunc context.CancelFunc
        ctx, cancelFunc = context.WithTimeout(context.Background(), c.c.config.Timeout)
        defer cancelFunc()
    }

    resp, err := c.c.rawRequestWithContext(ctx, r)
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()

    // ...
}
```

Additionally, since there are some [non-api callers](https://github.com/hashicorp/vault/blob/7e74bebea84b098d9d2de0eed2490f352481717a/vault/raft.go#L1135) of `RawRequestWithContext`, it should be refactored as well:

```go
func (c* Client) RawRequestWithContext(ctx context.Context, r *Request) (*Response, error) {
     	if c.config.Timeout != 0 {
		// Note: we purposefully do not call cancel manually. The reason is
		// when canceled, the request.Body will EOF when reading due to the way
		// it streams data in. Cancel will still be run when the timeout is
		// hit, so this doesn't really harm anything.
		ctx, _ = context.WithTimeout(ctx, c.config.Timeout)
	}
	
	return rawRequestWithContext(ctx, r)
}

func (c* Client) rawRequestWithContext(ctx context.Context, r *Request) (*Response, error) {
    // the rest of the RawRequestWithContext implementation
}
```

Admittedly, this adds a few more lines to each `_WithContext` function and changes may not be an ideal solution. However, it ensures that:

1. the context is only cancelled after resp.Body.Close()
2. `_WithContext()` functions can be safely called with `context.Background()` without any potential context leaks

I’d be happy to discuss any other potential fixes to the context leak issue or help with the implementation if needed. Please let me know what you think :slightly_smiling_face:

api/auth_token.go
```diff
@@ -179,6 +215,13 @@ func (c *TokenAuth) RenewSelf(increment int) (*Secret, error) {
 // RenewTokenAsSelf behaves like renew-self, but authenticates using a provided
 // token instead of the token attached to the client.
 func (c *TokenAuth) RenewTokenAsSelf(token string, increment int) (*Secret, error) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	return c.RenewTokenAsSelfWithContext(ctx, token, increment)
+}
+
+// RenewTokenAsSelfWithContext the same as RenewTokenAsSelf, but with a custom context.
```
averche, line 223: This is a bit of a nit but would it be possible to move the documentation for each pair of methods into the `__WithContext(...)` one, since it will be the primary way of calling the API moving forward.

The wrapper function could have a simple one-liner like the following:
```
// RenewTokenAsSelf wraps RenewTokenAsSelfWithContext using context.Background.
```

This is consistent with the [net/http NewRequest + NewRequestWithContext](https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/net/http/request.go;l=829) documentation.

api/auth_token.go
ilyakaznacheev: Good idea. I suppose to add the function like this

```go
func (c *Client) defaultContext() (context.Context, context.CancelFunc) {
	if c.config.Timeout > 0 {
		return context.WithTimeout(context.Background(), c.config.Timeout)
	}
	return context.WithCancel(context.Background())
}
```

And use it to obtain context

```go
func (c *TokenAuth) Create(opts *TokenCreateRequest) (*Secret, error) {
	ctx, cancelFunc := c.c.defaultContext()
	defer cancelFunc()
	return c.CreateWithContext(ctx, opts)
}
```

What do you think?

api/auth_token.go
ilyakaznacheev: Done

api/auth_token.go
averche: I don't think this would solve the issue, unfortunately.

I apologize if my explanation was not very clear. The `Create` function codepath is actually OK with your current implementation. The issue is with `CreateWithContext` function (as well as all the other `_WithContext` functions. Since the following two lines are removed:

```diff
-	ctx, cancelFunc := context.WithCancel(context.Background())
-	defer cancelFunc()
```

... the following code will have a context leak:

```go
	config := vault.DefaultConfig()
	config.Timeout = 30 * time.Second  // set the request timeout

	client, _ := vault.NewClient(config)
       
	ctx := context.Background()
	s, err := client.Auth().CreateWithContext(ctx, opts) // the context will never be cancelled here
```

ilyakaznacheev: I think I get your idea. You want to wrap the context into a context with a timeout inside each `_WithContext` method?

ilyakaznacheev: Like the method

```go
func (c *Client) wrapWithTimeout(ctx context.Context) (context.Context, context.CancelFunc) {
	if c.config.Timeout > 0 {
		return context.WithTimeout(ctx, c.config.Timeout)
	}
	return ctx, func() {}
}
```

And then call it like this
```go
func (c *TokenAuth) CreateWithContext(ctx context.Context, opts *TokenCreateRequest) (*Secret, error) {
	ctx, cancelFunc := c.c.wrapWithTimeout(ctx)
	defer cancelFunc()
```

averche: Yes, I think that would work 👍 

I would maybe call this helper function `withConfiguredTimeout` to be more explicit about its purpose. It would also follow the `WithXYZ` standard pattern.

api/auth_token.go
averche: Thanks for making the change! It looks like a few extra `// //` sneaked into the comments:

https://github.com/hashicorp/vault/blob/af651acc743aa378bfe53397797a5b38c26f77c6/api/auth_token.go#L223

ilyakaznacheev: fixed

api/auth_token.go
ilyakaznacheev: Done.

I have an idea. Since in all API methods that use context we call the `withConfiguredTimeout` method, we can simplify methods as follows:

```go
func (c *Client) withConfiguredTimeout(ctx context.Context) (context.Context, context.CancelFunc) {
	if c.config.Timeout > 0 {
		return context.WithTimeout(ctx, c.config.Timeout)
	}
	return context.WithCancel(ctx)
}

func (c *Client) Help(path string) (*Help, error) {
	return c.HelpWithContext(context.Background(), path)
}

// HelpWithContext reads the help information for the given path.
func (c *Client) HelpWithContext(ctx context.Context, path string) (*Help, error) {
	ctx, cancelFunc := c.withConfiguredTimeout(ctx)
	defer cancelFunc()
	// ...
}
```

What do you think?

averche: I don't it's necessary to change `withConfiguredTimeout`. Now that the context with timeout is always cancelled, we can simply remove the redundant `WithCancel(ctx)` calls:

```go
// unchanged
func (c *Client) withConfiguredTimeout(ctx context.Context) (context.Context, context.CancelFunc) {
	if c.config.Timeout > 0 {
		return context.WithTimeout(ctx, c.config.Timeout)
	}
	return ctx, func() {}
}

// removed context.WithCancel
func (c *Client) Help(path string) (*Help, error) {
	return c.HelpWithContext(context.Background(), path)
}
```

ilyakaznacheev: Perhaps, yes. I'll change that.

ilyakaznacheev: Done

averche: Looks great so far 👍 

One last thing here, the `RawRequestWithContext` function is invoked in [several places](https://github.com/hashicorp/vault/blob/7e74bebea84b098d9d2de0eed2490f352481717a/vault/raft.go#L1135) outside of the vault/api. Therefore we still need to keep the timeout hack inside of it. One way to solve this is to refactor the `RawRequestWithContext` implementation into unexported `rawRequestWithContext` (which will be called by all other vault/api functions) and add back the timeout hack (I have an example in my first comment).

Somewhat related to this, the `RawRequest` func should probably call the unexported `rawRequestWithContext` with a `withConfiguredTimeout(context.Background())` context.

api/client.go
```diff
@@ -1097,6 +1097,12 @@ func (c *Client) RawRequest(r *Request) (*Response, error) {
 // a Vault server not configured with this client. This is an advanced operation
 // that generally won't need to be called externally.
 func (c *Client) RawRequestWithContext(ctx context.Context, r *Request) (*Response, error) {
+	ctx, cancelFunc := c.withConfiguredTimeout(ctx)
+	defer cancelFunc()
```
averche, line 1101: Unfortunately, we can't cancel the context here because it will cause the `request.Body` to be invalidated (EOF) before the calling function is done with it. This was the reason for the original hack, which was trying to address #4973 issue. So the function(s) that call `RawRequestWithContext` should be responsible for cancelling the context. I think it would be best to not cancel here and add back the comment explaining the hack.

In my opinion, `RawRequestWithContext` should not have been exported -- this would prevent the issues we are dealing with here.

api/auth_token.go
averche: I replied inline with the change since this thread is getting too large.

api/client.go
ilyakaznacheev: I think we can't unexport `RawRequestWithContext` at the moment anyway, so I assume to have my changes but remove the cancel function call:

```go
func (c *Client) RawRequestWithContext(ctx context.Context, r *Request) (*Response, error) {
	// some meaningful description
	ctx, _ := c.withConfiguredTimeout(ctx)
	return c.rawRequestWithContext(ctx, r)
}
```

IMO it should work the same way as if the context was inside the `rawRequestWithContext` method but just looks cleaner to me. Maybe I missing some part of leakage magic.

ilyakaznacheev: Done

api/client.go
```diff
@@ -1089,14 +1089,29 @@ func (c *Client) NewRequest(method, requestPath string) *Request {
...
 func (c *Client) RawRequest(r *Request) (*Response, error) {
 	return c.RawRequestWithContext(context.Background(), r)
 }
 
 // RawRequestWithContext performs the raw request given. This request may be against
 // a Vault server not configured with this client. This is an advanced operation
 // that generally won't need to be called externally.
+//
+// Deprecated: method is left for backwards compatibility.
+// It is expected to be used internally. Use higher level methods instead.
```
averche, line 1104:
```suggestion
// Deprecated: This method should not be used directly. Use higher level
// methods instead.
```

ilyakaznacheev: Fixed

api/sys_health.go
```diff
@@ -3,6 +3,10 @@ package api
 import "context"
 
 func (c *Sys) Health() (*HealthResponse, error) {
+	return c.HealthWithContext(context.Background())
+}
+
+func (c *Sys) HealthWithContext(ctx context.Context) (*HealthResponse, error) {
```
averche, line 9: This function is missing withConfiguredTimeout call

api/sys_monitor.go
```diff
@@ -17,7 +17,7 @@ func (c *Sys) Monitor(ctx context.Context, logLevel string) (chan string, error)
 		r.Params.Add("log_level", logLevel)
 	}
 
-	resp, err := c.c.RawRequestWithContext(ctx, r)
+	resp, err := c.c.rawRequestWithContext(ctx, r)
```
averche, line 20: This one is also missing withConfiguredTimeout call

api/sys_raft.go
```diff
@@ -143,6 +157,8 @@ func (c *Sys) RaftSnapshot(snapWriter io.Writer) error {
 		return err
 	}
 
+	req = req.WithContext(ctx)
```
averche, line 160: I think it would be better to just use c.c.NewRequestWithContext above.

api/sys_raft.go
```diff
@@ -132,9 +138,17 @@ func (c *Sys) RaftJoin(opts *RaftJoinRequest) (*RaftJoinResponse, error) {
...
+// RaftSnapshot wraps RaftSnapshotWithContext using context.Background.
 func (c *Sys) RaftSnapshot(snapWriter io.Writer) error {
+	return c.RaftSnapshotWithContext(context.Background(), snapWriter)
+}
+
+// RaftSnapshotWithContext invokes the API that takes the snapshot of the raft cluster and
+// writes it to the supplied io.Writer.
+func (c *Sys) RaftSnapshotWithContext(ctx context.Context, snapWriter io.Writer) error {
+	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
+	defer cancelFunc()
```
averche, line 150: We may not want to impose the default timeout here since this is likely a long-running operation. @VinnyHC, what do you think?

Similar question for `RaftSnapshotRestoreWithContext`

api/sys_seal.go
```diff
@@ -3,57 +3,91 @@ package api
 import "context"
 
 func (c *Sys) SealStatus() (*SealStatusResponse, error) {
+	return c.SealStatusWithContext(context.Background())
+}
+
+func (c *Sys) SealStatusWithContext(ctx context.Context) (*SealStatusResponse, error) {
+	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
+	defer cancelFunc()
+
```
averche, line 12:
```suggestion
```
This is redundant since we already call withConfiguredTimeout in sealStatusRequestWithContext helper

This applies to all other functions that call `sealStatusRequestWithContext`

averche:
Almost there! 👍  I found a few small issues above.

VinnyHC (changes requested):
A few comments around our Snapshot API and a general observation on accessing configuration values.

api/sys_raft.go
```diff
@@ -271,9 +287,17 @@ func (c *Sys) RaftSnapshot(snapWriter io.Writer) error {
...
 func (c *Sys) RaftSnapshotRestore(snapReader io.Reader, force bool) error {
+	return c.RaftSnapshotRestoreWithContext(context.Background(), snapReader, force)
+}
+
+// RaftSnapshotRestoreWithContext reads the snapshot from the io.Reader and installs that
+// snapshot, returning the cluster to the state defined by it.
+func (c *Sys) RaftSnapshotRestoreWithContext(ctx context.Context, snapReader io.Reader, force bool) error {
+	ctx, cancelFunc := c.c.withConfiguredTimeout(ctx)
+	defer cancelFunc()
+
```
VinnyHC, line 300: There is work in progress to revamp the `RaftSnapshotRestore` removing its dependency on `RawRequestWithContext` and functioning more similarly to `RaftSnapshot`, so let's remove the timeout here as well.

averche:
Hi IIya, just wanted to loop you in that we're in the final phases of getting these contributions committed to Vault. Due to the impact of this change we've decided to take over the PR which will help us ferry it through some of our internal approval and review processes. We've kept all of your commits intact to properly reflect your contribution. The new PR can can be found here: #14388. Thanks again for your submission!

ilyakaznacheev:
Thank you for that. Unfortunately due to moving to another country, I didn't have enough time to make changes to the PR, but I hope I was able to help improve the API. I look forward to the new release.

PR #14160, opened 2022-02-18
OSS Port: Activity log breakdowns

vishalnayak:


HridoyRoy (approved):
Lgtm -- From my code inspection it achieves parity with ent.

ccapurso:
Ignore my comments as the goal is to be in parity with ent.

PR #14168, opened 2022-02-18
add missed test

pmmukh:
missed a test in my previous merge https://github.com/hashicorp/vault/pull/14141/files, adding that in now

### DIFF
--- a/vault/logical_system_test.go
+++ b/vault/logical_system_test.go
@@ -875,6 +875,63 @@ func TestSystemBackend_remount(t *testing.T) {
 	})
 }
 
+func TestSystemBackend_remount_destinationInUse(t *testing.T) {
+	c, b, _ := testCoreSystemBackend(t)
+
+	me := &MountEntry{
+		Table: mountTableType,
+		Path:  "foo/",
+		Type:  "generic",
+	}
+	err := c.mount(namespace.RootContext(nil), me)
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+
+	req := logical.TestRequest(t, logical.UpdateOperation, "remount")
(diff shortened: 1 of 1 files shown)


hghaf099 (approved):
LGTM!

vault/logical_system_test.go
```diff
@@ -875,6 +875,63 @@ func TestSystemBackend_remount(t *testing.T) {
...
+	req := logical.TestRequest(t, logical.UpdateOperation, "remount")
+	req.Data["from"] = "secret"
+	req.Data["to"] = "foo"
+	req.Data["config"] = structs.Map(MountConfig{})
+	resp, err := b.HandleRequest(namespace.RootContext(nil), req)
+
+	if err != logical.ErrInvalidRequest {
+		t.Fatalf("err: %v", err)
+	}
+	if !strings.Contains(resp.Data["error"].(string), "path already in use at \"foo/\"") {
```
ccapurso, line 900: You can use `resp.Error()` instead of pulling `"error"` out of `resp.Data`. In order to get the string you would call `resp.Error().Error()`. Keep in mind that `resp.Error()` can return nil so you likely want to guard with `resp.IsError()`.

There are numerous examples where you could change but I'll just leave this one comment.

ccapurso (approved):
Looks good, just one minor comment about a potential change for error handling.

vault/logical_system_test.go
pmmukh: oh neat, that is def cleaner, and I'm going to keep that in mind for future tests! for now though, I'm going to merge as is, cause i'm rushing to get the ent pr in by eod and i kinda don't want to kick off another checks cycle 😅

PR #14180, opened 2022-02-21
Add documentation for managed key test sign API

stevendpclark:
Add documentation for the new managed key api that allows operators to test the managed key configuration by going through a sign/verify workflow with some randomly generated data.

See VAULT-4989

### DIFF
--- a/website/content/api-docs/system/managed-keys.mdx
+++ b/website/content/api-docs/system/managed-keys.mdx
@@ -177,6 +177,34 @@ $ curl \
 }
 ```
 
+## Test Sign with a managed key
+
+This endpoint allows an operator to validate that a managed key configuration works
+by signing and verifying some randomly generated data. No values can be provided to sign,
+nor are signed or verified data returned to the caller. If the call returns a successful HTTP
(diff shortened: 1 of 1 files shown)


website/content/api-docs/system/managed-keys.mdx
```diff
@@ -177,6 +177,34 @@ $ curl \
...
 ```
 
+## Test Sign with a managed key
+
+This endpoint allows an operator to validate that a managed key configuration works
+by signing and verifying some randomly generated data. No values can be provided to sign,
+nor are signed or verified data returned to the caller. If the call returns a successful HTTP
+status code, the configuration can be considered valid.
+
+~> **Note**: if key generation is allowed this call will generate a key within the KMS if missing.
```
victorr, line 187: Can I suggest ".. if key generation is allowed and no existing key can be found, this call will generate the key" to make the phrase a bit closer to the description of allow_generate_key.

victorr (approved):
Approved with an optional suggestion.

kitography (approved):
That looks super readable.

PR #14181, opened 2022-02-21
Fix a simple typo

wazery:
Fixes a simple typo in the [system backend - monitor endpoint documentation](https://www.vaultproject.io/api-docs/system/monitor).


kalafut (approved):
Thanks!

PR #14192, opened 2022-02-22
Change address lookup behaviour if VAULT_SRV_LOOKUP is set.

gounselor:
I think the behaviour of VAULT_SRV_LOOKUP is set is strange:
It should not expect http (what about a prod https setup?)
And actually the VAULT_ADDR should not contain a scheme in this case.

I made a minimal change to make it work for us. 
We are using SRV records like _vault._tcp.example.domain which return
a list of hosts with the port appended.

The impure thing is that it overrides the scheme to be https, because i expect to use a SRV record approach in prod only which should https.

I am not 100% sure if this is the correct way, maybe we should discuss howto work with SRV records.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -1042,10 +1042,16 @@ func (c *Client) NewRequest(method, requestPath string) *Request {
 	// if SRV records exist (see https://tools.ietf.org/html/draft-andrews-http-srv-02), lookup the SRV
 	// record and take the highest match; this is not designed for high-availability, just discovery
 	// Internet Draft specifies that the SRV record is ignored if a port is given
+	// According to that draft, if there is not port it should follow
+	// https://datatracker.ietf.org/doc/html/rfc2782
+	// We expect an addr which looks for example like _vault._tcp.example.com
+	// DNS does randomization of the records if there are multiple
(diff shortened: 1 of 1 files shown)


hsimon-hashicorp:
Hello @gounselor - is this in relationship to an open issue? We ask that PRs be linked to an existing issue, so discussions about the desired solution can be hashed out there. Thanks!

gounselor:
Hi,

no, it’s not but i could open one if this is required.

Best regards,

Michael

hsimon-hashicorp:
> Hi, no, it’s not but i could open one if this is required. Best regards, Michael
> […](#)
> On 22. Feb 2022, at 23:05, Heather Simon ***@***.***> wrote: Hello @gounselor - is this in relationship to an open issue? We ask that PRs be linked to an existing issue, so discussions about the desired solution can be hashed out there. Thanks! — Reply to this email directly, view it on GitHub, or unsubscribe. Triage notifications on the go with GitHub Mobile for iOS or Android. You are receiving this because you were mentioned.

Please do - as I mentioned, we prefer to encourage discussion of the problem as well as its solution, to make sure that all the angles are covered. Thanks! :)

PR #14196, opened 2022-02-22
Fix autoseal health check race by passing metrics sink in CoreConfig

sgmiller:
Fixes the race where postUnseal tries to start metrics sink while it's being replaced by the test function.


ncabatoff (approved):

ncabatoff:
Could I ask you to backport this to 1.10 please?

PR #14197, opened 2022-02-22
Ensure that fewer goroutines survive after a test completes

ncabatoff:
* add Core.ShutdownWait that doesn't return until shutdown is done
* create the usedCodes cache on seal and nil it out on pre-seal so that the finalizer kills the janitor goroutine
* stop seal health check when sealing rather than wait for it to discover the active context is done
* make sure all lease-loading goroutines are done before returning from restore
* make uniquePoliciesGc discover closed quitCh immediately instead of only when the ticker fires
* make sure all loading goroutines are done before returning from loadEntities, loadCachedEntitiesOfLocalAliases


raskchanky (approved):

sgmiller (approved):
The autoseal cleanup lgtm!

PR #14199, opened 2022-02-22
Vault 4864 vault crl generation non compliance in issuer field

kitography:
Adds the necessary fields for https://www.rfc-editor.org/rfc/rfc5280.html compliant CRL generation, fixes a bug where:

In essence, OP's certificate is issued with a DN which contains only single-entity RDNs, but in a different order than Go's default. Go's x509.Certificate.CreateCRL mangles the certificate's subject field by round-tripping it through the non-compliant x509/pkix.Name. By switching to x509.CreateRevocationList(...), we can fix this behavior.

Notably, x509.CreateRevocationList requires a RevocationList template and validates this template has provided a Number value (a presumed monotonically increasing counter similar to CRL version). We need to add this to storage (and thus, acquire a lock when building the CRL) on each node which has a CRL. We can probably start at 1 (and increment from the previous omitempty/zero value).

### DIFF
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -97,6 +97,7 @@ func Backend(conf *logical.BackendConfig) *backend {
 	}
 
 	b.crlLifetime = time.Hour * 72
+	b.crlCounter = 0
 	b.tidyCASGuard = new(uint32)
 	b.tidyStatus = &tidyStatus{state: tidyStatusInactive}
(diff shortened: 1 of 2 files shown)


cipherboy:
Closing as this was addressed as part of #15277. :-) Thanks @kitography! 

Also, sadly, on second inspection of `x509.CreateRevocationList`, it doesn't actually build the name correctly :sob:

PR #14201, opened 2022-02-22
Prefix mount to path

Monkeychip:
Per PR [here](https://github.com/hashicorp/vault/pull/14164/files).

Additionally, made another mirage handler that is the exact payload on the RFC. Tested against this and all worked as expected.


hashishaw (approved):
👏

Issue #14204, opened 2022-02-22
The doc does not mention that leases can be revoked from the UI


xens:
**Description**

The vault [Concepts / Lease documentation](https://www.vaultproject.io/docs/concepts/lease) does not mention that leases can be revoked from the UI, it only mentions the CLI or the API for manual operations.

**To Reproduce**
1. In the UI go inside Access / Leases
2. select a prefix or a lease
3. there's a button you can click to revoke either the lease or a certain prefix

**Expected behavior**

The documentation should mention that leases can be revoked from the UI.

PR #14205, opened 2022-02-22
fix(doc): mention that leases can be revoked on the UI

xens:
The vault [Concepts / Lease documentation](https://www.vaultproject.io/docs/concepts/lease) does not mention that leases can be revoked from the UI, it only mentions the CLI or the API for manual operations. This PR fixes this issue.

https://github.com/hashicorp/vault/issues/14204


taoism4504 (approved):
LGTM

PR #14208, opened 2022-02-22
update go mod and go sum for sdk

HridoyRoy:
This PR was made by running

`go mod tidy`
and 
`go mod download`

in the sdk folder.


mladlow:
There isn't actually an SDK update in there, and I'm not sure if the other changes should be or not.

mladlow:
I just saw your note about running this in the sdk folder. I think the commands should be run in the top-level `vault` directory. I'm not the right person to evaluate when we want to update the SDK deps themselves.

HridoyRoy:
Ah, my mistake! Closing this PR.

PR #8776, opened 2020-04-19
Clarify when service_registraion was introduced

spuder:
Resolves https://github.com/hashicorp/vault/issues/8768
Language is modeled after the nomad acl version limits

> ~> Version information ACLs are only available on Nomad 0.7.0 and above.

https://www.vaultproject.io/docs/configuration/service-registration/consul

PR #14213, opened 2022-02-23
Clarify service_registration stanza version

robmonte:
A refactoring of https://github.com/hashicorp/vault/pull/8776

This PR simply clarifies in the docs when the `service_registration` configuration option was added to Vault, and that it is not required when using Consul as the storage backend for Vault.


fairclothjm (approved):

ccapurso (approved):
This looks good, thank you for the contribution!

PR #14217, opened 2022-02-23
Change OpenAPI code generator to extract request objects

averche:
### Current generated OpenAPI json

The current generated spec has request bodies anonymously defined inline:

```json
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "type": "object",
                "properties": {
                  "token": {
                    "type": "string",
                    "description": "My token"
                  }
                }
              }
            }
          }
 ```

### New generated OpenAPI spec:

The new generated spec has the request bodies extracted out under `/components/schemas` and referenced with `$ref` tags. The advantage of this style is that the spec can now be used for code generation. The generated request structs will have human-readable names (e.g. `KvLookupRequest` in the example below):

```json
   { {
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/KvLookupRequest"
                  }
                }
              }
            }
          }
       }
  "components": {
    "schemas": {
      "KvLookupRequest": {
        "type": "object",
        "properties": {
          "token": {
            "type": "string",
            "description": "My token"
          }
        }
      }
    }
  }
```

### Implementation

The struct schemas are extracted individually for each plugin in `openapi.go` and then combined in `vault/logical_system.go`. Since the plugins don't have names (types) associated with them until they are mounted, we pass the plugin name to each plugin through its help endpoint within the "data" map.


scripts/gen_openapi.sh
```diff
@@ -17,7 +17,7 @@ then
     exit 1
 fi
 
-vault server -dev -dev-root-token-id=root &
+../bin/vault server -dev -dev-root-token-id=root &
```
AnPucel, line 20: I'm assuming this is for local testing?

averche: Yes, didn't mean to commit it 😃

sdk/framework/backend.go
```diff
@@ -486,9 +486,18 @@ func (b *Backend) handleRootHelp() (*logical.Response, error) {
 		return nil, err
 	}
 
+	// The plugin type (e.g. "kv", "cubbyhole") is only assigned at the time
+	// the plugin is enabled (mounted). If specified in the request, the type
+	// will be used as part of the request/response body names in the OAS
+	// document
+	var requestResponsePrefix string
```
digivava, line 493: Is there a possibility of this variable staying as an empty string? And if so will that cause any problems with the spec generation?

sdk/framework/openapi.go
```diff
@@ -32,6 +32,9 @@ func NewOASDocument() *OASDocument {
 			},
 		},
 		Paths: make(map[string]*OASPathItem),
+		Components: OASComponents{
+			Schemas: make(map[string]*OASSchema),
```
digivava, line 36: Good job utilizing this more modern OAS concept! (reference schemas)

sdk/framework/openapi.go
```diff
@@ -459,6 +470,30 @@ func documentPath(p *Path, specialPaths *logical.Paths, backendType logical.Back
...
+// constructRequestName joins the given prefix with the path elements into a
+// CamelCaseRequest string.
+//
+// For example, prefix="kv" & path=/config/lease/{name} => KvConfigLeaseRequest
+func constructRequestName(requestResponsePrefix string, path string) string {
+	var b strings.Builder
+
+	b.WriteString(strings.Title(requestResponsePrefix))
+
+	// split the path by / _ - seprarators
```
digivava, line 482: lil typo here on "separators"

digivava (approved):
Small typo in a comment but otherwise it looks good to me, good job thinking through how best to get the mount type in there.

sdk/framework/backend.go
averche: Yes, it is possible if whoever calls the plugin's /help endpoint does not specify the "requestResponsePrefix". In this case, the generated request name will be missing that prefix. For example "KvLookupRequest" would be replaced with just "LookupRequest". I'm not really sure what the best way of dealing with this is (or if it should be of concern). In the OpenAPI generation workflow, the prefix will always be set.

VinnyHC (approved):
This looks good to me! I briefly walked through the API Explorer in the FE and that also seems to be happy with your changes. 🎉

sdk/framework/backend.go
kalafut, line 493:
```suggestion
	requestResponsePrefix := req.GetString("requestResponsePrefix")
```
... and you can remove the assignment below.

averche: TIL, thanks!

sdk/framework/openapi.go
```diff
@@ -204,9 +213,9 @@ var (
 )
 
 // documentPaths parses all paths in a framework.Backend into OpenAPI paths.
-func documentPaths(backend *Backend, doc *OASDocument) error {
+func documentPaths(backend *Backend, requestResponsePrefix string, doc *OASDocument) error {
 	for _, p := range backend.Paths {
-		if err := documentPath(p, backend.SpecialPaths(), backend.BackendType, doc); err != nil {
+		if err := documentPath(p, backend.SpecialPaths(), requestResponsePrefix, backend.BackendType, doc); err != nil {
```
kalafut, line 218: Since we're changing this... As this has now accumulated 5 parameters, I wonder if it is worth considering a small refactor? Maybe `documentPath` could be a method on `*OASDocument`, and we could do like:

```go 
cfg := pathConfg{...}
if err := doc.appendPaths(cfg); err != nil {    // naming TBD
... 
}
```

Not a blocker, just an idea.

kalafut: I'd be fine either way. Probably cleanest as a separate PR.

sdk/framework/backend.go
```diff
@@ -486,9 +486,15 @@ func (b *Backend) handleRootHelp() (*logical.Response, error) {
 		return nil, err
 	}
 
+	// The plugin type (e.g. "kv", "cubbyhole") is only assigned at the time
+	// the plugin is enabled (mounted). If specified in the request, the type
```
kalafut, line 490: Can you explain a bit more why you'd want to specify `requestResponsePrefix`?

kalafut (approved):
LGTM. Does the generated doc still pass the swagger.io inspector OK?

sdk/framework/backend.go
averche: The plugins/backends currently don't have a notion of their type. It defaults to the name of the executable but can be overridden when the plugin is mounted. Since we need this type to form the Request & Response full names, I'm passing them as an optional request parameter to the plugin root help endpoint.

averche: Should I add more explanation to the comment?

averche:
> LGTM. Does the generated doc still pass the swagger.io inspector OK?

Yes, tested it in swagger.io 👍

sdk/framework/backend.go
kalafut: Yes I think an addition to the comment would be good.

PR #14222, opened 2022-02-23
Set header content type instead of overwriting all headers

### LINKED ISSUE #14221: JSONMergePatch overwrites all headers removing X-Vault-Namespace among others
**Describe the bug**
When trying to use JSONMergePatch with vault cloud that requires a namespace I get permission denied. This is because all of the headers are replaced in order to add the json patch header:

https://github.com/hashicorp/vault/blob/f5d96970038a878b5836a98f40748f1340854bee/api/logical.go#L149

**Expected behavior**
It is expected that headers such as `X-Vault-Namespace` remain.

**Environment:**
* Vault Go SDK

dynajoe:
Fixes #14221

### DIFF
--- a/api/logical.go
+++ b/api/logical.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"net/http"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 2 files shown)


ccapurso:
Thank you for the contribution, @dynajoe! Would you be able to add a changelog entry by adding the following to `changelog/14222.txt` include the following text?

````
```release-note: bug
core/api: Fix overwriting of request headers when using JSONMergePatch.
```
````

dynajoe:
Completed.

ccapurso:
Hi, @dynajoe. It looks like there is an extra 2 in your changelog filename. Would you be able to rename the file to match the PR number? Once the the changelog check passes this will be good to go and I will merge.

dynajoe:
Wow, good eye. Fixed.

ccapurso (approved):
Looks good, thanks for catching this!

PR #12789, opened 2021-10-10
Add agent/v1/quit endpoint

remilapeyre:
Closes https://github.com/hashicorp/vault/issues/11089

PR #14223, opened 2022-02-23
Quit agent endpoint with config

### LINKED ISSUE #11089: Support for shutting down Vault Agent when Job ends (possibly via HTTP POST)
We have several long-running Kubernetes batch/v1 Jobs that require credentials from an injected `vault agent`.  The token and credentials need to be renewed throughout the life of the Job and should be revoked when the Job ends (via [vault-k8s](https://github.com/hashicorp/vault-k8s)s "vault.hashicorp.com/agent-revoke-on-shutdown" annotation).

Currently, the Jobs never finish because agent's container will continue to live on indefinitely even after the Job's "main" container finishes its work and exits.

We'd like to have a mechanism for telling the injected vault agent to shut down.

We also run Istio in our cluster and its `istio-proxy` sidecar (which is an `envoy` proxy under the hood) has a similar issue.  The Istio project has worked around this problem by having it run a [HTTP listener](https://github.com/istio/istio/issues/15041) which gracefully exits when it receives a `POST /quitquitquit`.

Additional tooling has grown up around this convention, including the [`scuttle` helper](https://github.com/redboxllc/scuttle) which wraps the execution of the Job's "main" process and issues `POST` requests to each of the Pod's sidecars asking them to shut down.

I'd love to be able to configure the vault agent sidecar to listen for a POST on some path and cleanly shut down if/when it receives one.

We currently use a custom fork of the aforementioned `scuttle` command to `pkill -SIGINT vault-sidekick` but this forces us to relax our security settings by setting `shareProcessNamespace: true` and to run both the sidecar and main process under the same user.

We cannot use vault-k8s' "vault.hashicorp.com/agent-pre-populate-only" annotation (as suggested in [another issue](https://github.com/hashicorp/vault-k8s/issues/120#issuecomment-612442641)) because without the agent running as a sidecar the token/credentials would expire before the main Job completes (and could live beyond its completion since they're not actively revoked).

The same underlying issue has come up with `consul-k8s` and it looks like it's [adding a similar endpoint](https://github.com/hashicorp/consul-k8s/issues/162).

Envoy's [underlying endpoint](https://www.envoyproxy.io/docs/envoy/latest/operations/admin.html?highlight=quitquitquit#post--quitquitquit).

In the future this *might* be solved by Kubernetes allowing [dependencies between containers within a Pod to be specified](https://github.com/kubernetes/kubernetes/issues/65502), but that discussion has been ongoing for 2.5 years and the appropriate teams have not reached consensus on the shape/scope of the solution.

**Additional context**
I would be happy to help implement a solution if this sounds like a good idea.

I originally requested this over in the [vault-k8s repo](https://github.com/hashicorp/vault-k8s/issues/217), they [seemed interested](https://github.com/hashicorp/vault-k8s/issues/217#issuecomment-771175938), but pointed out that this would be a Vault Agent feature and redirected me here.

tomhjp:
Builds on #12789 and closes #11089

Support a quit API endpoint so that when Vault Agent can be programatically shut down. This is useful for instance when Agent is a sidecar in a Kubernetes job to stop the job from hanging forever. Currently the only workarounds are to use `shareProcessNamespace: true` for the container so that process kill signals can be sent, or to avoid the sidecar container entirely and just rely on an init container.


tomhjp:
cc @taoism4504 for documentation updates

command/agent_test.go
```diff
@@ -2082,3 +2082,118 @@ listener "tcp" {
...
+			},
+			LogicalBackends: map[string]logical.Factory{
+				"kv": logicalKv.Factory,
+			},
+		},
+		&vault.TestClusterOptions{
+			NumCores: 1,
+		})
+	cluster.Start()
+	defer cluster.Cleanup()
```
benashz, line 2105: nit: lately I have been using `t.Cleanup()` instead of a defer, ideally `vault.NewTestCluster()` would take care of this for us.
```suggestion
	t.Cleanup(cluster.Cleanup)
```

command/agent_test.go
```diff
@@ -2082,3 +2082,118 @@ listener "tcp" {
...
+
+	// Unset the environment variable so that agent picks up the right test
+	// cluster address
+	defer os.Setenv(api.EnvVaultAddress, os.Getenv(api.EnvVaultAddress))
+	err := os.Unsetenv(api.EnvVaultAddress)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	listenAddr := "127.0.0.1:18123"
```
benashz, line 2118: nit: using static ports here could cause problems in the case where the port is already bound by another process.

We could probably get the port dynamically doing something like:
```
	ln1, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		t.Fatal(err)
	}
	listenAddr := ln1.Addr().String()
	ln1.Close()
```

command/agent_test.go
```diff
@@ -2082,3 +2082,118 @@ listener "tcp" {
 		"Points",
 	})
 }
+
+func TestAgent_Quit(t *testing.T) {
```
benashz, line 2086: nit: it looks like we are testing two modes within the same test. I would consider perhaps making this test table driven?

command/agent.go
```diff
@@ -1047,3 +1050,19 @@ func (c *AgentCommand) handleMetrics() http.Handler {
 		}
 	})
 }
+
+func (c *AgentCommand) handleQuit() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Only handle POST/PUT requests
+		switch r.Method {
+		case http.MethodPost:
+		case http.MethodPut:
```
benashz, line 1059: I think `case` can take more than one argument.
```suggestion
		case http.MethodPost, http.MethodPut:
```

command/agent.go
```diff
@@ -1047,3 +1050,19 @@ func (c *AgentCommand) handleMetrics() http.Handler {
...
+		default:
+			w.WriteHeader(405)
+			return
+		}
+
+		c.logger.Debug("received quit request")
```
benashz, line 1065: Should we handle the case where `c.Logger` might be nil, assuming it is not required to be set somewhere else.

command/agent_test.go
```diff
@@ -2082,3 +2082,118 @@ listener "tcp" {
...
+		t.Fatalf("expected 404 but got: %d", resp.StatusCode)
+	}
+
+	// Now try on listener 2 where the quit API should be enabled.
+	err = client.SetAddress("http://" + listenAddr2)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = client.RawRequest(client.NewRequest("POST", "/agent/v1/quit"))
```
benashz, line 2187: We support both `POST` and `PUT`, should we be testing with both methods? Might be another reason to make this a table driven test.

benashz, line 2187: nit: use the method constant from`http`
```suggestion
	_, err = client.RawRequest(client.NewRequest(http.MethodPost, "/agent/v1/quit"))
```

command/agent.go
```diff
@@ -1047,3 +1050,19 @@ func (c *AgentCommand) handleMetrics() http.Handler {
 		}
 	})
 }
+
+func (c *AgentCommand) handleQuit() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Only handle POST/PUT requests
+		switch r.Method {
+		case http.MethodPost:
```
benashz, line 1058: Out of curiosity what is the reason for supporting multiple methods?

command/agent.go
```diff
@@ -1047,3 +1050,19 @@ func (c *AgentCommand) handleMetrics() http.Handler {
...
+		case http.MethodPut:
+		default:
+			w.WriteHeader(405)
```
benashz, line 1061:
```suggestion
			w.WriteHeader(http.StatusMethodNotAllowed)
```

command/agent.go
benashz, line 1065: Should we log something about the requesting peer? Could be useful to know who is killing us?

internalshared/configutil/listener.go
```diff
@@ -111,6 +113,10 @@ type Listener struct {
 	CustomResponseHeadersRaw interface{}                  `hcl:"custom_response_headers"`
 }
 
+type AgentAPI struct {
```
benashz, line 116: nit: might be worth adding some documentation.

website/content/docs/agent/index.mdx
```diff
@@ -109,6 +109,22 @@ Vault Agent allows client-side caching of responses containing newly created tok
...
+
+### Quit
+
+This endpoints triggers shutdown of the agent. By default, it is disabled, and can
+be enabled per listener using the [`agent_api`][agent-api] stanza. It is recommended
+to only enable this on trusted interfaces, as it does not require any authorization to use.
+
+| Method | Path             |
+| :----- | :--------------- |
+| `POST` | `/agent/v1/quit` |
```
benashz, line 122: nit: missing documentation for `PUT`

benashz (commented):
Looking good! Had a few questions/suggestions.

command/agent.go
```diff
@@ -716,6 +716,9 @@ func (c *AgentCommand) Run(args []string) int {
 			// Create a muxer and add paths relevant for the lease cache layer
 			mux := http.NewServeMux()
 			mux.Handle(consts.AgentPathCacheClear, leaseCache.HandleCacheClear(ctx))
+			if lnConfig.AgentAPI != nil && lnConfig.AgentAPI.EnableQuit {
+				mux.Handle(consts.AgentPathQuit, c.handleQuit())
+			}
```
tvoran, line 721: I noticed while testing this that if enable_quit=false, requests to /agent/v1/quit are proxied through to Vault which returns a 404 (as you'd expect). Alternatively we could decide to handle or not inside the handleQuit() function, which means the request would always stop at agent and not be proxied to Vault and back.

(Sorry, just thinking out loud here. I don't have a strong preference one way or the other.)

command/agent_test.go
```diff
@@ -2082,3 +2082,118 @@ listener "tcp" {
...
+		t.Fatal(err)
+	}
+
+	// First try on listener 1 where the API should be disabled.
+	resp, err := client.RawRequest(client.NewRequest("POST", "/agent/v1/quit"))
+	if err == nil {
+		t.Fatalf("expected error")
+	}
+	if resp != nil && resp.StatusCode != http.StatusNotFound {
+		t.Fatalf("expected 404 but got: %d", resp.StatusCode)
```
benashz, line 2178:
```suggestion
		t.Fatalf("expected %d but got: %d", http.StatusNotFound, resp.StatusCode)
```

website/content/docs/agent/index.mdx
```diff
@@ -334,6 +356,7 @@ template {
 [persistent-cache]: /docs/agent/caching/persistent-caches
 [template]: /docs/agent/template
 [template-config]: /docs/agent/template-config
+[agent-api]: /docs/agent/#agent-api-stanza
```
calvn, line 359:
```suggestion
[agent-api]: /docs/agent/#agent_api-stanza
```

command/agent_test.go
tomhjp: I actually prefer `defer` tbh - it's much more commonly understood and doesn't have any disadvantages in this instance from `t.Cleanup` AFAICT. Agreed it would be better if `vault.NewTestCluster()` did this for us though, that would be a nice improvement.

command/agent_test.go
tomhjp: Added, and updated the other tests in this file to use the same

command/agent_test.go
tomhjp: I actually explicitly want to test that the per-listener config is working properly, i.e. that it's not globally enabling  the API

command/agent.go
tomhjp: I don't it's necessary, it's set early on in the Agent code path.

command/agent.go
tomhjp: I didn't give it much thought tbh, I just didn't change it. Just POST is better though - changed.

command/agent.go
tomhjp: I think we need to be careful about logging IPs, as they are personal data according to GDPR regulations. Something to consider in another PR/broader context I think.

command/agent.go
tomhjp: Wow good catch - yeah I've updated this so we always serve the endpoint, but just serve the 404 from Agent instead when it's not enabled.

command/agent_test.go
benashz: Sounds good!

tomhjp:
Thanks for the reviews. I think that's all comments addressed. Docs site deployment link is here if you'd like to browse: https://vault-aq18lm9e5-hashicorp.vercel.app/docs/agent#api

tomhjp:
Thanks for the reviews. I think that's all comments addressed. Docs site deployment link is here if you'd like to browse: https://vault-aq18lm9e5-hashicorp.vercel.app/docs/agent#api

tvoran (approved):
Working great for me in my testing!

calvn (approved):
Looks great! 

Reminder that if there's a `release/1.10.x` by the time this gets merged, we'd need to also backport it to that branch.

taoism4504 (approved):
Text looks good.

benashz (approved):
LGTM!

command/agent_test.go
```diff
@@ -2082,3 +2095,133 @@ listener "tcp" {
...
+	wg.Wait()
+}
+
+// Get a randomly assigned port and then free it again before returning it.
+// There is still a race when trying to use it, but should work better
+// than a static port.
+func generateListenerAddress(t *testing.T) string {
+	t.Helper()
+
+	ln1, err := net.Listen("tcp", "127.0.0.1:0")
```
kalafut, line 2220: For awareness: I see we've recently extracted some "find me a free port" functionality into: https://github.com/hashicorp/go-freeport. Though since we can do `:0` it probably isn't necessary here.

command/agent_test.go
```diff
@@ -2082,3 +2095,133 @@ listener "tcp" {
...
+	case <-cmd.ShutdownCh:
+	case <-time.After(5 * time.Second):
+		t.Errorf("timeout")
+	}
+
+	wg.Wait()
+}
+
+// Get a randomly assigned port and then free it again before returning it.
+// There is still a race when trying to use it, but should work better
```
benashz, line 2215: The risk of a port collision should be relatively low since most kernel's incrementally assign ports from their dynamic range of many thousands of ports.

tomhjp: TIL, thanks

PR #14224, opened 2022-02-23
UI/fix db role ttl display

hashishaw:
This PR fixes a UI issue where the TTL looked like it never changed from 1h (the default) when really the adapter was not parsing the correct value from the API, which is `default_ttl`.

### DIFF
--- a/ui/app/templates/components/database-role-edit.hbs
+++ b/ui/app/templates/components/database-role-edit.hbs
@@ -63,6 +63,14 @@
           @label={{capitalize (or attr.options.label (humanize (dasherize attr.name)))}}
           @value={{stringify (get @model attr.name)}}
         />
+      {{else if (eq attr.options.editType "ttl")}}
+        <InfoTableRow
+          @alwaysRender={{true}}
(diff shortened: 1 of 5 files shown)


ui/app/templates/components/database-role-edit.hbs
```diff
@@ -63,6 +63,14 @@
           @label={{capitalize (or attr.options.label (humanize (dasherize attr.name)))}}
           @value={{stringify (get @model attr.name)}}
         />
+      {{else if (eq attr.options.editType "ttl")}}
```
Monkeychip, line 66: just checking that the editType here doesn't need to be default_ttl?

Monkeychip (approved):

ui/app/templates/components/database-role-edit.hbs
hashishaw: Nope, the `attr.name` is `default_ttl`, but here I'm checking if editType (as set in the options on the model) is ttl, as it is for max_ttl as well, so that we can format it correctly 👍

Issue #14114, opened 2022-02-16
The default schema used in the mysql backend is insufficient for KVv2 storage


drawks:
**Describe the bug**
When upgrading a KVv1 mount to KVv2 on a server using the mysql backend the column width of `vault_key` will be insufficient for storing the values used with the KVv2 mount.

**To Reproduce**
1.  Configure vault with a KVv1 mount `vault secrets enable -version=1 kv`
2. Store some secrets in the KVv1 mount at an arbitrary depth ala `vault kv put /kv/foo/bar/bing/band/bong/ding/dong baz=boo`
3. Upgrade mount with `vault kv enable-versioning secret/`
4. Observe command returns, but vault never completes KVv2 upgrade and mount becomes unusable.
```
$ vault kv get /kv/foo/bar/bing/band/bong/ding/dong
Error reading kv/data/foo/bar/bing/band/bong/ding/dong: Error making API request.

URL: GET http://127.0.0.1:8200/v1/kv/data/foo/bar/bing/band/bong/ding/dong
Code: 400. Errors:

* Upgrading from non-versioned to versioned data. This backend will be unavailable for a brief period and will resume service shortly.
```
And there is a single error in the logs detailing what happened
```
Feb 13 07:21:32 localhost.localdomain vault[30523]: 2022-02-13T07:21:32.529Z [INFO]  secrets.kv.kv_fb3379c9: collecting keys to upgrade
Feb 13 07:21:32 localhost.localdomain vault[30523]: 2022-02-13T07:21:32.530Z [INFO]  secrets.kv.kv_fb3379c9: done collecting keys: num_keys=2
Feb 13 07:21:32 localhost.localdomain vault[30523]: 2022-02-13T07:21:32.538Z [ERROR] secrets.kv.kv_fb3379c9: upgrading resulted in error: error="Error 1406: Data too long for column 'vault_key' at row 1" progress=1/2```

**Expected behavior**
Mount is successfully upgraded and continues working.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.9.1
* Vault CLI Version (retrieve with `vault version`): 1.9.1
* Server Operating System/Architecture: CentOS 7

**Additional context**
Manually updating the schema of the vault database before conducting the upgrade results in proper operation: `alter table vault modify vault_key varchar(3072);`

drawks:
@hsimon-hashicorp I've updated my issue with a bit more detail, I suspect that this is a very small PR to fix this.

[...]

PR #14231, opened 2022-02-23
Increase column width of vault_key on mysql

drawks:
* resolves The default schema used in the mysql backend is insufficient for KVv2 storage #14114
* increases column width of vault_key from 512 to 3072 in mysql physical backend

As noted in https://github.com/hashicorp/vault/issues/14114 the vault_key column is too narrow in the default schema. This fix will fix new installations, existing installations will still need to modify their schema in place.


hsimon-hashicorp:
Thanks, @drawks! Can you also add a changelog entry? :)

ccapurso:
Hi, @drawks. Thank you for the contribution! The increased width of 3072 is a reasonable change given how MySQL allocates for `VARBINARY` columns. Is the new width necessary or is it an attempt to provide so much headroom to completely avoid the issue?

drawks:
@ccapurso 3072 is just the max width for that column type. AFAICT the actual data width of the column is not logically bounded and increases relative to the number of "directory parts" of the key. In my use case a width of ~600 is sufficient, but I imagine there are others with arbitrarily deep paths in their kv stores and an upgrade from KVv1 to KVv2 is not recoverable if the column is too narrow so I just opted for the maximum width.

ccapurso:
Great, thanks for the context!

CHANGELOG.md
```diff
@@ -81,6 +81,7 @@ certificates. [[GH-13660](https://github.com/hashicorp/vault/pull/13660)]
 * identity: Fix possible nil pointer dereference. [[GH-13318](https://github.com/hashicorp/vault/pull/13318)]
 * identity: Fixes a panic in the OIDC key rotation due to a missing nil check. [[GH-13298](https://github.com/hashicorp/vault/pull/13298)]
 * kmip (enterprise): Fix locate by name operations fail to find key after a rekey operation.
+* physical/mysql: Create table with wider `vault_key` column when initializing database tables. [[GH-14231](https://github.com/hashicorp/vault/pull/14231)]
```
ccapurso, line 84: `CHANGELOG.md` should actually not be updated directly. Changelog entries for a PR can be added by creating a changelog file for the associated PR. In your case, this would be `changelog/14231.txt`. The contents would then be:

````
```release-note:improvement
physical/mysql: Create table with wider `vault_key` column when initializing database tables.
```
````

**Review by ccapurso — CHANGES_REQUESTED** [review]

**Comment by ccapurso** [comment]
This looks good overall, just one comment on the changelog entry which needs to be addressed prior to merging.

File: CHANGELOG.md (hunk 1 above)
Comment (drawks) [reply CHANGELOG.md]: I've just push a fix for this, I hope everything looks good now.

**Review by ccapurso — APPROVED** [review]
Looks good! I will merge once CircleCI checks pass.

**Comment by ypid-work** [comment]
> existing installations will still need to modify their schema in place.

In that case there should be either:

* A database migration script that does this change for existing setups.
* Mentioned in the changelog/upgrade guide that the following needs to be executed:

```SQL
`ALTER TABLE `vault` CHANGE `vault_key` `vault_key` VARBINARY(3072) NOT NULL;`
```

I am not that familiar with Vault but that is how other projects do it. If I miss something please excuse me and feel free to educate me.

PR #14233, opened 2022-02-23
UI: Fix incorrect validity modal on transit secrets engine

hashishaw:
Prior to this fix, validity was not being properly passed which resulted in incorrect information to the user. See before/after gifs.

**BEFORE**

before

**AFTER**
after


Monkeychip (approved):
Nice fix!

Issue #13530, opened 2021-12-28
pki documentation unclear about interactions between glob_domains and wildcards


drawks:
**Describe the bug**
pki documentation is not clear as to what the expected behavior is when combining `allow_glob_domains` and requests for literal wildcard domains. Behavior has changed at some point between 1.5.4 and 1.9.1, but no mention of changed behavior is present in any of the release notes/changelogs

**To Reproduce**
1. configure a pki role with `allow_glob_domains` and a domain allowed entry like `*.*.foo` on 1.5.4 and 1.9.1 
2. observe that 1.5.4 will allow cert issuance for wildcards like `*.bar.foo`
3. observe that 1.9.1 will not allow cert issuance for the same request

**Expected behavior**
Expectations are unclear because documentation doesn't describe what should be expected.

chongyangshi:
The use of `glob.Glob` is a much older line of code, but [the change in its input as described above for v1.9.3](https://github.com/hashicorp/vault/commit/5e016cd7c94051dd47ee972db6715b22ddccd534#diff-4ddfb6cce8278968d34e839e8c8989c06fb1514b53d824267fbf2d54306b28cbL348) started passing `sanitizedName` rather than `name` to `glob.Glob` . And this the `sanitizedName` with the leading `.*` already trimmed away would no longer be accepted by `glob.Glob`.

cipherboy:
Hello @dovys and @drawks -- we agree on the assessment here and we're working a fix internally.

drawks:
The change that restores the previous behavior is greatly appreciated, it would still be good if the documentation is clear about what the expected behavior should be. Part of the reason that this took a while to identify as a bug/regression is because the intended behavior is not clearly spelled out anywhere. Since the root problem of this issue is the lack of documentation, I'd appreciate if it could be re-opened until such time as the documentation is updated.

cipherboy:
That's in #14236 which also documents the stuff in #14238. :-) Let me know if you have suggestions for either of those PRs (or the one that landed fixing this issue).

PR #14235, opened 2022-02-23
Fix broken interactions between glob_domains and wildcards

cipherboy:
As documented in #13530, the behavior recently changed in Vault 1.8.0 as a result of an [earlier commit addressing case sensitivity during issuance](https://github.com/hashicorp/vault/commit/f267c3ad749f3f4f048fdc38898cb9e21133546a), preventing some certificates from being issued when they should've been.

We revert this behavior change and add regression tests for various issuance scenarios.

Resolves: https://github.com/hashicorp/vault/issues/13530

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -3904,6 +3904,275 @@ func TestBackend_Root_FullCAChain(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


kitography (approved):
Looking good :).

cipherboy:
Thanks @kitography!

Issue #7592, opened 2019-10-08
Add a feature to support disable wildcard subdomains


venky999:
**Is your feature request related to a problem? Please describe.**
At the moment

VAULT PKI supports wild card subdomains to generated.

like `*.myorg.com` or `*.finance.myorg.com`

Is it possible to add an option to explicitly disable this feature
so that I need to use a specific subdomain when generating cert

like 

`test.myorg.com` or `test.finance.myorg.com`

This is useful when using intermediate CA created for different namespaces from root CA and I want to Intermediate not to issue wildcard subdomains

jefferai:
If the user can pile on 200 subdomains in a single request, what's the logical difference between supporting * and supporting those 200 subdomains?

Just trying to understand the use case.

timarenz:
As I just worked on the same requirement for a customer this can be solved today using a Sentinel policy. However, this is a Vault Enterprise feature but could solve what you want to accomplish. 

Here is an example on how to do this:
[https://gist.github.com/timarenz/15a3b134e37af827c15304e196412b29](url)

venky999:
> If the user can pile on 200 subdomains in a single request, what's the logical difference between supporting * and supporting those 200 subdomains?
> 
> Just trying to understand the use case.

So that the PKI cannot issue subdomains that are not in the explicit list

arctiqtim:
So many large organizations' security teams have a policy against wildcard cert use.  Specifically Vault fits this use case to allow programmatic, policy-controlled user generated certificates.  The PKI secrets engine's biggest use case IMO is to provide a clean way to issue per-app per-service certificates that are finer grained, and very short TTLs.  An option in Vault to disable the creation of wildcard certs would allow admins to better enforce the idea of app-specific certs vs creation of wildcard certs.

Sentinel is a way to solve it for sure but I don't think it should be required.

AkashGuptaHC:
> As I just worked on the same requirement for a customer this can be solved today using a Sentinel policy. However, this is a Vault Enterprise feature but could solve what you want to accomplish.
> 
> Here is an example on how to do this: [https://gist.github.com/timarenz/15a3b134e37af827c15304e196412b29](url)

AkashGuptaHC:
> As I just worked on the same requirement for a customer this can be solved today using a Sentinel policy. However, this is a Vault Enterprise feature but could solve what you want to accomplish.
> 
[...]

PR #14236, opened 2022-02-23
Clarify documentation around certificate issuance

cipherboy:
This pull request clarifies the documentation around certificate issuance restrictions in roles. We document the new `allow_wildcard_certificates` field (added in #14238). 

A few clarifications are made to the documentation:

 - `allow_bare_domains`, `allow_glob_domains`, and `allow_subdomains` are all independent,
 - `enforce_hostnames` and `allow_wildcard_certificates` take precedence over `allow_any_name`,
 - We limit to RFC 6125 wildcards.
 - Clarify that both `allow_bare_domains` and `allow_glob_domains` will permit wildcard issuance in certain scenarios. 
 - `allow_localhost` is independent from `allowed_domains` listing (so `localhost` can be allowed in the latter explicitly and even if `allow_localhost=false`, will allowed to be issued for bare/subdomains of `localhost`).

---

```
Co-authored-by: mickael-hc <86245626+mickael-hc@users.noreply.github.com>
Co-authored-by: Kit Haines <kit.haines@hashicorp.com>
Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>
```

PR #14238, opened 2022-02-23
Add role parameter to restrict issuance of wildcard certificates

cipherboy:
This pull request allows Vault operators to restrict issuance of wildcard certificates via role configuration. We also fix several bugs in the parsing of wildcard certificates from a RFC 6125 perspective, permitting wildcard globs only in the left-most DNS label. 

For compatibility with existing Vault releases, this field defaults to `true` and must explicitly be set to `false` (at role creation time or via an update) in order to take advantage of this change.

Documented in: #14236
Resolves: https://github.com/hashicorp/vault/issues/7592


kitography (approved):
LGTM!  💯

PR #14244, opened 2022-02-24
builtin/logical/transit: fix dropped test error

alrs:
This fixes a dropped test `err` variable in `builtin/logical/transit`.

Could someone add the "no-changelog" label?


ccapurso (approved):
Good catch, @alrs. Thanks!

PR #14250, opened 2022-02-24
Client Count Styling fixes after design review

Monkeychip:
A couple of small styling changes after demoing with design.

TODO in next release:
- Only show attribution when one namespace is available.
- remove all caps in filter component (globally used component when is why we're pushing this until later).

### DIFF
--- a/ui/app/components/clients/horizontal-bar-chart.js
+++ b/ui/app/components/clients/horizontal-bar-chart.js
@@ -25,7 +25,7 @@ import { tracked } from '@glimmer/tracking';
 
 // SIZING CONSTANTS
 const CHART_MARGIN = { top: 10, left: 95 }; // makes space for y-axis legend
-const TRANSLATE = { down: 14 };
+const TRANSLATE = { down: 14, left: 99 };
 const CHAR_LIMIT = 15; // character count limit for y-axis labels to trigger truncating
 const LINE_HEIGHT = 24; // each bar w/ padding is 24 pixels thick
 
@@ -213,7 +213,7 @@ export default class HorizontalBarChart extends Component {
     // add client count total values to the right
     chartSvg
       .append('g')
-      .attr('transform', `translate(99, ${TRANSLATE.down})`)
+      .attr('transform', `translate(${TRANSLATE.left}, ${TRANSLATE.down})`)
       .selectAll('text')
       .data(dataset)
       .enter()
--- a/ui/app/components/clients/horizontal-bar-chart.js
+++ b/ui/app/components/clients/horizontal-bar-chart.js
@@ -25,7 +25,7 @@ import { tracked } from '@glimmer/tracking';
 
 // SIZING CONSTANTS
 const CHART_MARGIN = { top: 10, left: 95 }; // makes space for y-axis legend
-const TRANSLATE = { down: 13 };
+const TRANSLATE = { down: 14 };
 const CHAR_LIMIT = 15; // character count limit for y-axis labels to trigger truncating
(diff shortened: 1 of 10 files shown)


ui/app/styles/core/helpers.scss
```diff
@@ -165,13 +165,13 @@
   text-transform: lowercase;
 }
 .has-bottom-margin-xs {
-  margin-bottom: $spacing-xs;
+  margin-bottom: $spacing-xs !important;
```
hashishaw, line 168: Are these required because `chart-title` or `stat-title` also have a margin-bottom value? Could we not update the other classes to not include margin values instead of adding `!important` (which should be used sparingly), since this helper is far-reaching and the charts and stats are only used in client counts so far.

ui/lib/core/addon/templates/components/empty-state.hbs
```diff
@@ -13,7 +13,7 @@
         </p>
       {{/if}}
     {{else}}
-      <h3 class="empty-state-title" data-test-empty-state-title>
+      <h3 class={{concat "empty-state-title" (if this.subTitle " has-bottom-margin-m")}} data-test-empty-state-title>
```
hashishaw, line 16: 👏

hashishaw (commented):
We should avoid using !important if possible... let me know if it's reasonable

ui/app/styles/core/helpers.scss
Monkeychip: Yeah, I went back and forth on this. I can just add them to chart-title and stat-title, but sometimes they are overridden by the title class. Let me see if I can approach another way.

ui/app/styles/core/title.scss
```diff
@@ -2,6 +2,10 @@
 .subtitle:not(:last-child) {
   display: block;
   margin-bottom: 1rem;
+
```
Monkeychip, line 5: My original reason for using important was because title was overriding the margin-bottom class. So I added this here. It's a duplication of the has-bottom-margin-xs, but it doesn't use important. @hashishaw thoughts?

ui/app/components/clients/horizontal-bar-chart.js
```diff
@@ -213,7 +213,7 @@ export default class HorizontalBarChart extends Component {
     // add client count total values to the right
     chartSvg
       .append('g')
-      .attr('transform', `translate(${CHART_MARGIN.left}, ${TRANSLATE.down})`)
+      .attr('transform', `translate(${TRANSLATE.left}, ${TRANSLATE.down})`)
```
Monkeychip, line 216: This adds a space next to the number on the horizontal chart. 
image

ui/app/styles/core/title.scss
hashishaw: 👍 I think it's better than using !important 😄 thanks!

ui/app/styles/core/helpers.scss
```diff
@@ -171,7 +171,7 @@
   margin-bottom: $spacing-s;
 }
 .has-bottom-margin-m {
-  margin-bottom: $spacing-m;
+  margin-bottom: $spacing-m !important;
```
hashishaw: Can we do the same here?

Monkeychip: It should be removed in the latest files. I committed it here. https://github.com/hashicorp/vault/pull/14250/commits/85f9d139994f7002c3ed64de7d375c99322e146f

hashishaw: Ah thank you!

hashishaw (approved):
🚀

PR #14264, opened 2022-02-24
sdk changes for 1.10.0 release branch

HridoyRoy:
Steps to generate:

`go mod tidy`

`go mod download`

### DIFF
--- a/sdk/go.mod
+++ b/sdk/go.mod
@@ -13,7 +13,6 @@ require (
 	github.com/golang/protobuf v1.5.2
 	github.com/golang/snappy v0.0.4
 	github.com/hashicorp/errwrap v1.1.0
-	github.com/hashicorp/go-cleanhttp v0.5.2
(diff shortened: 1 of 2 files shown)


mladlow:
I don't actually think I have the right information to evaluate whether these changes are expected or not. Does it seem like these dependency removals would be expected based on a somewhat-recent change?

HridoyRoy:
> I don't actually think I have the right information to evaluate whether these changes are expected or not. Does it seem like these dependency removals would be expected based on a somewhat-recent change?

I can try to track down what PRs these changes could come from and post them in the description if I find anything.

calvn:
Looks like both of these were added in [this PR](https://github.com/hashicorp/vault/pull/10505/files#diff-7b5028b779639efd98a7f47865b168ca39285e1c46d94140c5d38c516d00b127R16) to the SDK. I don't see them referenced or imported anywhere in there though, so I think it should be safe to remove.

calvn (approved):

PR #14269, opened 2022-02-25
(WIP) [VAULT-5003] Use net/http client in Sys().RaftSnapshotRestore

VinnyHC:
Modifies `Sys().RaftSnapshotRestore` function to use a net/http client to prevent attempting to read the entire body of a request (which is the case in our `(*client).RawRequestWithContext()` )
The `go-retryablehttp` package [makes a copy of the request body on construction](https://github.com/hashicorp/go-retryablehttp/blob/ff6d014e72d968e0f328637b209477ee09393175/client.go#L219-L238) of a retry-able request in order to be able to re-play the request. This is problematic for the API client’s `Sys().RaftSnapshotRestore` function when large snapshots are streamed to the server.
This hasn't been identified as a problem in our CLI tool because we use files which `go-retryablehttp` can handle without reading the full value into memory.


api/sys_raft.go
```diff
@@ -272,24 +272,93 @@ func (c *Sys) RaftSnapshot(snapWriter io.Writer) error {
...
+	if resp == nil {
+		return nil
+	}
+
+	// Check for a redirect, only allowing for a single redirect
+	if resp.StatusCode == 301 || resp.StatusCode == 302 || resp.StatusCode == 307 {
+		// Parse the updated location
+		respLoc, err := resp.Location()
+		if err != nil {
+			return err
```
averche: These internal errors might be confusing for the caller without a context. Would it make sense to add `fmt.Errorf("...: %w", err)` around them?

api/sys_raft.go
```diff
@@ -272,24 +272,93 @@ func (c *Sys) RaftSnapshot(snapWriter io.Writer) error {
...
+	resp, err := c.c.config.HttpClient.Do(req)
 	defer resp.Body.Close()
 
+	if err != nil {
+		return err
+	}
+
+	if resp == nil {
+		return nil
+	}
```
averche: I believe this nil check might be [redundant](https://pkg.go.dev/net/http#Client.Do):

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close.

api/sys_raft.go
```diff
@@ -272,24 +272,93 @@ func (c *Sys) RaftSnapshot(snapWriter io.Writer) error {
...
+		req.Header.Set("X-Vault-Policy-Override", "true")
+	}
+
+	var result *Response
+	resp, err := c.c.config.HttpClient.Do(req)
 	defer resp.Body.Close()
 
+	if err != nil {
+		return err
+	}
```
averche, line 219: Should the `defer resp.Body.Close()` be done after the error check?

averche: Looks like you fixed it already 👍

api/sys_raft.go
VinnyHC: This was from [main](https://github.com/hashicorp/vault/blob/main/api/sys_raft.go#L177-L188) but indeed looks like it can be removed.

api/sys_raft.go
```diff
@@ -387,3 +315,86 @@ func (c *Sys) PutRaftAutopilotConfiguration(opts *AutopilotConfig) error {
...
+		}
+
+		// Update the request
+		req.URL = respLoc
+
+		// Retry the request
+		resp, err = c.c.config.HttpClient.Do(req)
+		if err != nil {
+			return nil, err
+		}
```
averche, line 391: I'm a bit unsure how the `defer resp.Body.Close()` would work here. Would it be applying to the new `resp` pointer in this case. If so, what happens to the original `resp.Body()`, would it not be closed?

VinnyHC: Decided caller should close body

api/client.go
```diff
@@ -1249,6 +1249,83 @@ START:
...
+		// Ensure a protocol downgrade doesn't happen
+		if req.URL.Scheme == "https" && respLoc.Scheme != "https" {
+			return nil, fmt.Errorf("redirect would cause protocol downgrade")
+		}
+
+		// Update the request
+		req.URL = respLoc
+
+		// Retry the request
+		resp, err = c.config.HttpClient.Do(req)
```
ncabatoff, line 1315: We don't need r.ResetJSONBody?

api/client.go
```diff
@@ -1249,6 +1249,83 @@ START:
 	return result, nil
 }
 
+// requestWithContext avoids the use of the go-retryable library found in RawRequestWithContext and is
+// useful when making requests where the req or resp body is likely larger than available memory
+func (c *Client) requestWithContext(ctx context.Context, r *Request) (*Response, error) {
```
ncabatoff, line 1254: Should also note in godoc that we ignore *all* config settings, including ones that aren't retryable-specific, i.e.. HttpClient, Timeout, OutputCurlString, Logger.

api/client.go
```diff
@@ -1249,6 +1249,83 @@ START:
...
+	req, err := http.NewRequestWithContext(ctx, r.Method, r.URL.RequestURI(), r.Body)
```
ncabatoff, line 1255: Curious why you don't use NewRequest?

api/client.go
VinnyHC: The existing redirect logic in the snapshot writer omitted this, but It seems that we should!

ncabatoff: Does it?  I was basing this comment on comparing your code with RawRequestWithContext.

VinnyHC: I'm pretty sure that is what's going on [here](https://github.com/hashicorp/vault/blob/main/api/sys_raft.go#L189-L210).
I've refactored this so both the read and write snapshot functions will use this new client method and is a good opportunity to clean up some of this one-off code and get closer to the logic `RawRequestWithContext` has implemented.

api/client.go
VinnyHC: godoc updated to be more clear as to what this method should be used for and what config options apply to it. I've also made it a bit more clear in the code to show that we use the `HttpClient` and fail-fast for `OutputCurlString`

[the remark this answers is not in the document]
VinnyHC: I do agree that my name could be more clear, but `nonRetriableRequestWithContext` might be misleading. I went with `httpRequestWithContext` to hopefully indicate that this uses `net/http` but happy to pivot.

api/client.go
```diff
@@ -1466,3 +1577,14 @@ func (w *replicationStateStore) states() []string {
 	copy(c, w.store)
 	return c
 }
+
+// tokenSanityCheck will check for non-printable characters to prevent a call that will fail at the api
```
ncabatoff, line 1581: We prefer to avoid the term "sanity check", see https://github.com/hashicorp/engineering-docs/blob/main/writing/style-guide.md#ableism

VinnyHC: Great call! Updating now and thanks for the doc!

api/client.go
ncabatoff: To be clear, I meant Client.NewRequest, not http.NewRequest.

VinnyHC: The client function `ToHTTP` has been deprecated (likely because it's encouraged to use `toRetryableHTTP()`) and in this case, I think it's more clear/succinct to see that we are using an `*http.Request` in this function.

api/client.go
```diff
@@ -53,6 +53,14 @@ const (
...
 	HeaderForward         = "X-Vault-Forward"
 	HeaderInconsistent    = "X-Vault-Inconsistent"
+	TLSErrorString        = "This error usually means that the server is running with TLS disabled\n" +
+		"but the client is configured to use TLS. Please either enable TLS\n" +
+		"on the server or run the client with -address set to an address\n" +
+		"that uses the http protocol:\n\n" +
+		"    vault <command> -address http://<address>\n\n" +
+		"You can also set the VAULT_ADDR environment variable:\n\n\n" +
+		"    VAULT_ADDR=http://<address> vault <command>\n\n" +
+		"where <address> is replaced by the actual address to the server."
```
averche, line 63: This error text might benefit from using the ` quote: https://play.golang.com/p/1834sdy0UMg . It's also OK to leave it as is, your choice 👍

api/client.go
```diff
@@ -1192,17 +1197,7 @@ START:
...
-					"This error usually means that the server is running with TLS disabled\n"+
-					"but the client is configured to use TLS. Please either enable TLS\n"+
-					"on the server or run the client with -address set to an address\n"+
-					"that uses the http protocol:\n\n"+
-					"    vault <command> -address http://<address>\n\n"+
-					"You can also set the VAULT_ADDR environment variable:\n\n\n"+
-					"    VAULT_ADDR=http://<address> vault <command>\n\n"+
-					"where <address> is replaced by the actual address to the server.",
-				err)
+			err = fmt.Errorf("%s\n\n%s", err, TLSErrorString)
```
averche, line 1200: Switching from `errwrap` to `fmt.Errorf` might cause issues for the clients checking the inner (wrapped) error for equality (maybe not, I'm not really familiar with `errwrap`)

api/client.go
```diff
@@ -1466,3 +1577,14 @@ func (w *replicationStateStore) states() []string {
 	copy(c, w.store)
 	return c
 }
+
+// tokenCheck will check for non-printable characters to prevent a call that will fail at the api
+func tokenCheck(t string) error {
```
averche, line 1582: Would `checkToken` or `validateToken` be a better name?

api/client.go
averche: If a client is currently using [this function](https://github.com/hashicorp/errwrap/blob/master/errwrap.go#L68-L73) to check for specific errors, it may start failing.

api/client.go
VinnyHC: Yes. They both would.

api/client.go
VinnyHC: Good call, I kept this as `errwrap` for backward compatibility.

api/client.go
```diff
@@ -1249,6 +1244,122 @@ START:
...
+
+	resp, err := httpClient.Do(req)
+
+	if resp != nil {
+		result = &Response{Response: resp}
+	}
+
+	if err != nil {
+		if strings.Contains(err.Error(), "tls: oversized") {
+			err = fmt.Errorf("%s\n\n%s", err, TLSErrorString)
```
averche, line 1323: This may also need to use `errwrap` for backwards compatibility (since RaftSnapshotRestore called RawRequestWithContext before).

averche (approved):
One optional comment, LGTM otherwise 👍

api/client.go
VinnyHC: Yeah, good call. Maybe a subsequent PR can address `errwrap`.

api/client.go
VinnyHC: @ncabatoff I'd love to get your take on this. For this PR I think it's a good call to leave `errwrap` for these specific cases, but we can add something to our backlog that free's up the last few uses of `errwrap` in `client.go` if it's a reasonable proposition.

ncabatoff (approved):
Nice work!

api/client.go
```diff
@@ -1249,6 +1244,122 @@ START:
...
+	if outputCurlString {
+		return nil, fmt.Errorf("output-curl-string is not implemented for this request")
+	}
+
+	req.URL.User = r.URL.User
+	req.URL.Scheme = r.URL.Scheme
+	req.URL.Host = r.URL.Host
+	req.Host = r.URL.Host
+
+	if r.Headers != nil {
```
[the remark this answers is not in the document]
VinnyHC: Gotcha! Makes sense. To prevent an unnecessary lock I implemented the header loop while grabbing other values from the client configuration.

PR #14270, opened 2022-02-25
Replace docs references to PUT with POST

kalafut:
The operations are handled identically, but ~85% of the references were
POST, and having a mix of PUT and POST has been a source of questions.

A subsequent commit will update the internal use of "PUT" such as by
the API client and `-output-curl-string`.


ncabatoff (approved):
Thanks for doing this!

PR #14271, opened 2022-02-25
Fix a bug with single-phase login MFA

hghaf099:
If a method ID is used that does not need a passcode,
for single-phase login, a user passes an empty string to the X-Vault-MFA header.
Currently, this causes an error to be returned that the header value is invalid.

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -108,7 +108,7 @@ var (
 	}
 
 	oidcProtectedPathRegex = regexp.MustCompile(`^identity/oidc/provider/\w(([\w-.]+)?\w)?/userinfo$`)
-	mfaMethodIDRegex       = regexp.MustCompile(logical.GenericOptionalUUIDRegex("method_id"))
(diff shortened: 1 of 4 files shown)


http/handler.go
```diff
@@ -1130,7 +1131,12 @@ func parseMFAHeader(req *logical.Request) error {
 			return fmt.Errorf("invalid data in header %q; missing method name or ID", MFAHeaderName)
 		}
 
-		if shardSplits[1] == "" {
+		// In cases where login MFA is used, and the MFA method does not use
```
ncabatoff, line 1134: Isn't this case handled above, i.e. for no passcode simply don't include a colon?

hghaf099: Ah, yes! :) 
I should be documenting this!

hghaf099:
Turns out there is simple solution for the issue as pointed out by Nick in the comment. I am going to close this PR.

PR #14276, opened 2022-02-25
Fix KV permissions error

Monkeychip:
This was a regression bug from this [PR](https://github.com/hashicorp/vault/pull/13872/files). When you first start an engine you are going to error on the model fetch because there are no secrets to list. This caused us to show the get credentials card on a fresh install of a secret engine. Now we're searching for a permissions error.

Will backport to 1.9.

### DIFF
--- a/ui/app/routes/vault/cluster/secrets/backend/list.js
+++ b/ui/app/routes/vault/cluster/secrets/backend/list.js
@@ -113,8 +113,9 @@ export default Route.extend({
         .catch((err) => {
           // if we're at the root we don't want to throw
           if (backendModel && err.httpStatus === 404 && secret === '') {
+            this.set('noMetadataPermissions', false);
             return [];
-          } else if (backendModel.engineType === 'kv' && backendModel.isV2KV) {
+          } else if (err.httpStatus === 403 && backendModel.isV2KV) {
             this.set('noMetadataPermissions', true);
             return [];
           } else {
--- a/ui/tests/acceptance/secrets/backend/kv/secret-test.js
+++ b/ui/tests/acceptance/secrets/backend/kv/secret-test.js
@@ -582,6 +582,8 @@ module('Acceptance | secrets/secret/create', function (hooks) {
     let userToken2 = consoleComponent.lastLogOutput;
     await settled();
     await listPage.visitRoot({ backend: enginePath });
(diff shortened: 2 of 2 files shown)


ui/app/routes/vault/cluster/secrets/backend/list.js
```diff
@@ -113,8 +113,9 @@ export default Route.extend({
         .catch((err) => {
           // if we're at the root we don't want to throw
           if (backendModel && err.httpStatus === 404 && secret === '') {
+            this.set('noMetadataPermissions', false);
```
hashishaw, line 116: If it's false by default this might be redundant

ui/tests/acceptance/secrets/backend/kv/secret-test.js
```diff
@@ -582,6 +582,8 @@ module('Acceptance | secrets/secret/create', function (hooks) {
     let userToken2 = consoleComponent.lastLogOutput;
     await settled();
     await listPage.visitRoot({ backend: enginePath });
+    // confirm they see an empty state and not the get-credentials card
+    await assert.dom('[data-test-empty-state-title]').hasText('No secrets in this backend');
```
hashishaw, line 586: nice! Can we add a test for the case that we do want to see the get-credentials card?

hashishaw (commented):
Nice! Couple small things

ui/app/routes/vault/cluster/secrets/backend/list.js
Monkeychip: It's true by default. I just tested without and I get the original error.

Monkeychip: https://github.com/hashicorp/vault/pull/14276/files/ca89590cacc2fe75575ed457c97d8c2e78ad63a7#diff-b30928193c4fb5f7951070126947426c653b4ee983ac47cb8c3d9c57c8cf9165R14

ui/tests/acceptance/secrets/backend/kv/secret-test.js
Monkeychip: Good call. Added in the latest commit.

hashishaw (approved):
thanks!

PR #14288, opened 2022-02-25
Client Count Testing

zofskeez:
Small changes after doing some client count testing. The mirage clients handler was updated with the response generated from vault-tools script and a small bug was fixed where removing a namespace from the filter did not remove the auth method filter input.


Monkeychip (approved):
Thanks for making these!

ui/app/components/clients/current.js
```diff
@@ -84,6 +84,7 @@ export default class Current extends Component {
     // value comes in as [namespace0]
     this.selectedNamespace = value;
     if (!value) {
+      this.authMethodOptions = [];
       // on clear, also make sure auth method is cleared
       this.selectedAuthMethod = null;
```
hashishaw, line 89: This line should remove the auth method filter when namespace is deselected, so not sure why that wasn't working 🤔

zofskeez: Yes that's what I added to fix the issue.

zofskeez: Sorry I was looking at your comment wrong :pensive:. The condition in the template was `{{#if (not (is-empty this.authMethodOptions))}}`. Looking at it again I suppose that could have been changed `{{#if this.selectedAuthMethod}}` which would have been cleaner but I wasn't sure how all this state was being managed. It seemed like updating `this.authMethodOptions` so it reflected the current state was the most straightforward.

hashishaw: I agree, this will be the most straightforward particularly since some namespaces won't have auth mounts to filter on, which is why it's that way  in the template. Thanks for the explanation!

PR #14286, opened 2022-02-25
Sync PKI API and FrameworkField descriptions

cipherboy:
As pointed out internally, a lot of the API docs and FrameworkField
descriptions of parameters were out of date. This syncs a number of
them, updating their descriptions where relevant.

`Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>`

---

There's definitely more of these to do, but this at least gets us started. I'm inclined to open this for review and merge, rather than trying to do them all at once.

PR #14292, opened 2022-02-25
Add warning when generate_lease=no_store=true when writing PKI role

cipherboy:
When `no_store=true`, the value of `generate_lease` is ignored completely
(and set to false). This means that when `generate_lease=true` is
specified by the caller of the API, it is silently swallowed. While
changing the behavior could break callers, setting a warning on the
response (changing from a 204->200 in the process) seems to make the
most sense.

---

This probably isn't that important (and can wait for 1.10 branching). I found it when looking at the docs around #14286 and thought it was interesting.

Can be tested with something like:

```
source devvault && vault secrets enable pki
vault write /pki/roles/example-com generate_lease=true no_store=true
```

### DIFF
--- a/builtin/logical/pki/path_roles.go
+++ b/builtin/logical/pki/path_roles.go
@@ -583,6 +583,7 @@ func (b *backend) pathRoleList(ctx context.Context, req *logical.Request, d *fra
 
 func (b *backend) pathRoleCreate(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
 	var err error
+	var resp *logical.Response
 	name := data.Get("name").(string)
 
 	entry := &roleEntry{
@@ -644,6 +645,10 @@ func (b *backend) pathRoleCreate(ctx context.Context, req *logical.Request, data
 	// no_store implies generate_lease := false
 	if entry.NoStore {
 		*entry.GenerateLease = false
+		if data.Get("generate_lease").(bool) {
+			resp = &logical.Response{}
+			resp.AddWarning("mutually exclusive values no_store=true and generate_lease=true were both specified; no_store=true takes priority")
+		}
 	} else {
 		*entry.GenerateLease = data.Get("generate_lease").(bool)
 	}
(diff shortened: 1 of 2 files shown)


stevendpclark (approved):
lgtm!

cipherboy:
Thanks for the review @stevendpclark!

builtin/logical/pki/path_roles.go
```diff
@@ -644,6 +645,10 @@ func (b *backend) pathRoleCreate(ctx context.Context, req *logical.Request, data
 	// no_store implies generate_lease := false
 	if entry.NoStore {
 		*entry.GenerateLease = false
+		if data.Get("generate_lease").(bool) {
```
harsimranmaan, line 648: Would it make sense to return an error when both options are set to true?

cipherboy: The [docs](https://github.com/hashicorp/vault/blob/main/builtin/logical/pki/path_roles.go#L370-L371) have mentioned [since the introduction of the `no_store` option](https://github.com/hashicorp/vault/commit/7cbc5d6e058647ea45f3254def7ffff060b59faa#diff-dae4006668fecb2f5554c7b3efa6f21df59a7055280dba36d435251372f0076cR215) that `no_store=true` sets `generate_lease=false` (regardless of its value), so I don't think we can necessarily make a breaking change at this point in time (which erring out would be). However, emitting a warning is the next best thing we can do for the time being. :-)

A future Vault v2 version potentially could make that change, or perhaps an equivalent amount of time with the warning. I dunno. But at least as it stands, we'd risk breaking people's existing e.g., Terraform definitions if we made that change now.

cipherboy: The nice thing is a [quick search](https://github.com/search?p=1&q=vault_pki_secret_backend_role+generate_lease+no_store&type=Code) shows that no _public_ Terraform users seem to have made this mistake, but [some use variables](https://github.com/adfinis-sygroup/terraform-vault-namespace/blob/ef2987f2efea2da331f9f9bc94bbc2497ff92db7/variables.tf#L168-L186) that could potentially be overridden to have the undesired behavior. Plus, I'd venture that most Terraform users keep their stuff private.

PR #14294, opened 2022-02-25
[WIP] Create new script to scrape registry for list of plugins

AnPucel:
Option 1: 
- We scrape the registry.go file with genplugins.go which uses an ast and creates a json
- We can then pipe those values into our vault commands and generate the OpenApi Spec as normal

### DIFF
--- a/scripts/genplugins.go
+++ b/scripts/genplugins.go
@@ -12,21 +12,19 @@ import (
 func main() {
 	fset := token.NewFileSet()
 
-	file, err := parser.ParseFile(fset, "../../vault/helper/builtinplugins/registry.go", nil, parser.ParseComments)
+	file, err := parser.ParseFile(fset, "../helper/builtinplugins/registry.go", nil, parser.ParseComments)
 	if err != nil {
 		panic(err)
 	}
 
 	cmap := ast.NewCommentMap(fset, file, file.Comments)
 
-	//var v *visitor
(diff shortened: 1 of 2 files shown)


scripts/genplugins.go
```diff
@@ -0,0 +1,89 @@
...
+	DbBackends []Backend `json:"DbPlugins"`
+}
+
+func (v *Visitor) Visit(n ast.Node) ast.Visitor {
+	if n == nil {
+		return nil
+	}
+
+	if keyValueExpr, ok := n.(*ast.KeyValueExpr); ok {
+		if ident, ok := keyValueExpr.Key.(*ast.Ident); ok {
```
averche, line 55: Style: it might be better to invert the `if` logic here to flatten out the code structure:

```go
keyValueExpr, ok := n.(*ast.KeyValueExpr)
if !ok {
     return v
}

ident, ok := keyValueExpr.Key.(*ast.Ident)
if !ok {
     return v
}

// the rest of the code
```

averche (commented):
In general, I prefer the shell script solution in #14311 (it's simpler and does not introduce go.mod dependencies). Leaving a few style comments 😃

AnPucel:
Going to close in favor of the purely shell script [solution](https://github.com/hashicorp/vault/pull/14311)

AnPucel:
Going to close this in favor of the purely shell script [solution](https://github.com/hashicorp/vault/pull/14311)

PR #14295, opened 2022-02-26
update build and release/ci to point to 1.10 release branch

HridoyRoy:


mladlow (approved):
Looks good! Can you please make a PR on main to remove 1.6.x and add 1.10.x from the `ci.hcl` file as well? I'll add that step to the parent checklist.

PR #14301, opened 2022-02-28
Output full secret path in certain kv commands

digivava:
This PR is meant to mitigate some of the confusion for new Vault users, who often struggle with the discrepancy between the shortened path used by the CLI (`secret/foo`) and the real secret path (`secret/data/foo`). This bites people when they've only worked with the CLI but then need to write a policy or make a logical call with our Go and Ruby client libraries.

With this change, `vault kv get`, `vault kv put`, `vault kv patch`, and `vault kv metadata get` will all output another top-level section when run in `table` output mode. I did not add this to the kv commands that produce just a simple success message, and I did not add it to kv rollback as it seemed irrelevant given the use case.

Examples:

**Creating a new secret:**
```
$ vault kv put secret/my-secret foo=bar
==== Secret Path ====
secret/data/my-secret

======= Metadata =======
Key                Value
---                -----
created_time       2022-02-28T17:29:34.110676Z
custom_metadata    <nil>
deletion_time      n/a
destroyed          false
version            1
```

**Reading a secret:**
```
$ vault kv get secret/my-secret
==== Secret Path ====
secret/data/my-secret

======= Metadata =======
Key                Value
---                -----
created_time       2022-02-28T17:29:34.110676Z
custom_metadata    <nil>
deletion_time      n/a
destroyed          false
version            1

=== Data ===
Key    Value
---    -----
foo    bar
```

For `kv metadata get`, the same logic is used, but the header is slightly different:
```
$ vault kv metadata get secret/my-secret
===== Metadata Path =====
secret/metadata/my-secret

========== Metadata ==========
Key                     Value
---                     -----
cas_required            false
created_time            2022-02-28T17:29:34.110676Z
current_version         1
custom_metadata         <nil>
delete_version_after    0s
max_versions            0
oldest_version          0
updated_time            2022-02-28T17:29:34.110676Z

====== Version 1 ======
Key              Value
---              -----
created_time     2022-02-28T17:29:34.110676Z
deletion_time    n/a
destroyed        false
```

While "Data Path" may seem more internally accurate and matching with "Metadata Path", I chose to go with "Secret Path" for the main use case (regular kv commands) because I want totally new users who are not familiar with the /data /metadata subcategories to immediately intuit that this is the "real path" to their secret.

### DIFF
--- a/command/kv_helpers.go
+++ b/command/kv_helpers.go
@@ -154,8 +154,12 @@ func kvParseVersionsFlags(versions []string) []string {
 	return versionsOut
 }
 
-func outputSecretPath(ui cli.Ui, path string) {
-	ui.Info(addEqualSigns("Secret Path", len(path)))
+func outputPath(ui cli.Ui, path string, isMetadataPath bool) {
+	header := "Secret Path"
+	if isMetadataPath {
+		header = "Metadata Path"
+	}
+	ui.Info(addEqualSigns(header, len(path)))
 	ui.Info(path)
 	ui.Info("")
 }
--- a/command/kv_helpers.go
+++ b/command/kv_helpers.go
@@ -142,7 +142,7 @@ func getHeaderForMap(header string, data map[string]interface{}) string {
 	// 4 for the column spaces and 5 for the len("value")
 	totalLen := maxKey + 4 + 5
 
-	return addEqualSigns(header, totalLen)
+	return padEqualSigns(header, totalLen)
 }
 
 func kvParseVersionsFlags(versions []string) []string {
@@ -154,18 +154,14 @@ func kvParseVersionsFlags(versions []string) []string {
 	return versionsOut
 }
 
-func outputPath(ui cli.Ui, path string, isMetadataPath bool) {
-	header := "Secret Path"
-	if isMetadataPath {
-		header = "Metadata Path"
-	}
-	ui.Info(addEqualSigns(header, len(path)))
+func outputPath(ui cli.Ui, path string, title string) {
+	ui.Info(padEqualSigns(title, len(path)))
 	ui.Info(path)
 	ui.Info("")
 }
 
 // Pad the table header with equal signs on each side
-func addEqualSigns(header string, totalLen int) string {
+func padEqualSigns(header string, totalLen int) string {
 	equalSigns := totalLen - (len(header) + 2)
 
 	// If we have zero or fewer equal signs bump it back up to two on either
--- a/command/kv_helpers.go
+++ b/command/kv_helpers.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/hashicorp/go-secure-stdlib/strutil"
 	"github.com/hashicorp/vault/api"
+	"github.com/mitchellh/cli"
 )
 
 func kvReadRequest(client *api.Client, path string, params map[string]string) (*api.Secret, error) {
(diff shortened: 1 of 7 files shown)


command/kv_metadata_get.go
```diff
@@ -117,6 +117,8 @@ func (c *KVMetadataGetCommand) Run(args []string) int {
 
 	delete(secret.Data, "versions")
 
+	outputPath(c.UI, path, true)
```
averche, line 120: Rather than passing a boolean here, I think it would be easier to read with a string:

```go
func outputPath(ui cli.Ui, path string, title string) {
      // ...
}

outputPath(c.UI, path, "Metadata Path")
```

Alternatively, it's also OK to have two functions, `outputSecretPath` and `outputMetadataPath`. I think I would lean towards just passing a string.

command/kv_patch.go
```diff
@@ -185,7 +185,17 @@ func (c *KVPatchCommand) Run(args []string) int {
 		return code
 	}
 
-	return OutputSecret(c.UI, secret)
+	if Format(c.UI) != "table" {
+		return OutputSecret(c.UI, secret)
```
averche, line 189: This is an early successful return, which is a bit non-standard in Go. Is it possible to fix this by inverting the `if` to `if Format(c.UI) == "table" {...}`. Alternatively, I would add a comment here: 
```go
return OutputSecret(c.UI, secret) // early return
```

averche (approved):
Small nit comments, otherwise LGTM 👍

command/kv_helpers.go
```diff
@@ -141,6 +142,30 @@ func getHeaderForMap(header string, data map[string]interface{}) string {
...
+	if isMetadataPath {
+		header = "Metadata Path"
+	}
+	ui.Info(addEqualSigns(header, len(path)))
+	ui.Info(path)
+	ui.Info("")
+}
+
+// Pad the table header with equal signs on each side
+func addEqualSigns(header string, totalLen int) string {
```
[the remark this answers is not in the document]
digivava: I was bouncing back and forth between these two names as well! I'll go with `padEqualSigns` after all then.

command/kv_patch.go
digivava: There were a few places already in the KV command code that were already doing `return OutputSecret` earlier based on a flag and then having `return 0` at the very end (such as kv get), so I tried to make it work similarly to those. OutputSecret is capable of failing and returning exit code 1 so I think that's the rationale behind it sometimes coming earlier.

command/kv_put.go
```diff
@@ -161,5 +161,15 @@ func (c *KVPutCommand) Run(args []string) int {
...
+		return OutputSecret(c.UI, secret)
+	}
+
+	outputPath(c.UI, path, "Secret Path")
+
+	metadata := secret.Data
+	c.UI.Info(getHeaderForMap("Metadata", metadata))
+	OutputData(c.UI, metadata)
+
+	return 0
```
VinnyHC, line 174: For the `OutputData` call, if I understand it correctly I think we might swallow an error code here. To @averche point you might want to introduce something like:
```go
	if Format(c.UI) == "table" {
		metadata := secret.Data
		c.UI.Info(getHeaderForMap("Metadata", metadata))
		return OutputData(c.UI, metadata)
	}

	return OutputSecret(c.UI, secret)
```

VinnyHC (commented):
This is a great addition that doesn't seem to muddy things up too much! Do you think there's an easy way to add some tests?

command/kv_put.go
digivava: Yeah I noticed that throughout the KV code they're actually doing this (calling OutputData without the return in front) so that's where I got that (look at the current kv get code, for example), so I figured there was just some "how to write CLI code" thing I wasn't understanding, but I agree that it seems kinda janky so maybe I will switch it to be like that.

digivava:
> This is great! So, we decided against adding the Metadata path when returning the secret path due to difficulty?

Yeah, I wasn't able to think of a way to get the metadata path at this point in the code, without breaking for paths that have words like "data" in the actual secret name or mount name.

command/kv_patch.go
digivava: Ended up switching it around like you'd suggested after all

command/kv_test.go
```diff
@@ -1230,3 +1208,73 @@ func TestKVPatchCommand_RWMethodPolicyVariations(t *testing.T) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if secret == nil || secret.Auth == nil || secret.Auth.ClientToken == "" {
+		return nil, fmt.Errorf("missing auth data: %#v", secret)
+	}
+
+	return secret.Auth, err
+}
```
VinnyHC, line 1280: Some weird `git` stuff going on here? Might be worth trying to clean it up to preserve the commit history.

digivava: Keeping this manual move-around I did as we discussed, but will try to avoid doing this next time for reviewers' sake

PR #14309, opened 2022-02-28
update SDK Version For 1.11

HridoyRoy:


mladlow (approved):

mladlow:
It's fine if this isn't in this PR, but we need to update `ci.hcl` to not have 1.6 and to have 1.10, I think. It _could_ be in this PR, I'm pretty sure? Or the one after this to update the SDK version in the go.mod.

PR #14311, opened 2022-02-28
Create OpenApi spec w/ dynamically generated plugins

AnPucel:
Currently, the gen_openapi.sh script only pulls hardcoded plugins which won't allow us to pick up new ones added. We're introducing a new script to scrape `registry.go` for a list of plugins that we can enable for the purposes of better OpenApi code gen coverage.


scripts/gen_dynamic_openapi.sh
```diff
@@ -0,0 +1,75 @@
...
+fi
+
+vault server -dev -dev-root-token-id=root &
+sleep 2
+VAULT_PID=$!
+
+echo "Mounting all builtin backends..."
+
+# Read auth backends
+flag=false
```
averche, line 27: Nit: the variable names in this file could be improved a bit for clarity:
```
flag -> codeLinesStarted
p -> line
re -> regex
var1 -> plugin
```

scripts/gen_dynamic_openapi.sh
```diff
@@ -0,0 +1,75 @@
...
+re='".*"'
```
averche, line 28: Wouldn't this regex match anything? We probably want to have a more narrow regex here.

scripts/gen_dynamic_openapi.sh
```diff
@@ -0,0 +1,75 @@
...
+while read p; do
+    if [[ $p == *"credentialBackends:"* ]] ; then
+        flag=true
+    elif [ $flag = true ] && [[ $p = *"}"* ]]  ; then
+        break
+    elif [ $flag = true ] && [[ $p =~ $re ]]; then
+        backend=${BASH_REMATCH[0]}
+        var1=$(sed -e 's/^"//' -e 's/"$//' <<<"$backend")
```
averche, line 36: Is this line going to exclude the `// deprecated` entries. It may be worthwhile to do that in the overall regex.

scripts/gen_dynamic_openapi.sh
```diff
@@ -0,0 +1,75 @@
...
+    fi
+done <../../vault/helper/builtinplugins/registry.go
+
+
+# Enable enterprise features
+if [[ ! -z "$VAULT_LICENSE" ]]
+then
+  vault write sys/license text="$VAULT_LICENSE"
+  vault secrets enable kmip
+  vault secrets enable transform
```
averche, line 62: It may be possible to extract these programmatically as well since the `registry.go` has these entries in vault-enterprise.

averche (commented):
I definitely like this direction 👍  I can help with tightening up the shell script & the regex.

scripts/gen_dynamic_openapi.sh
AnPucel: yes definitely!

scripts/gen_dynamic_openapi.sh
AnPucel: > I definitely like this direction 👍 I can help with tightening up the shell script & the regex.

Cool! I'll publish this one and start addressing your comments!

AnPucel:
> I definitely like this direction 👍 I can help with tightening up the shell script & the regex.

Great! I'll publish this and start addressing the comments

scripts/gen_dynamic_openapi.sh
AnPucel: This matches everything between quotes, which seemed specific enough to pick up what we need. But let me know improving regex. It could be narrower to include `:` and the key value pattern. But it seemed okay since we're just regexing one line. Let me know your thoughts though

scripts/gen_dynamic_openapi.sh
AnPucel: yep! This should be simple enough

scripts/gen_dynamic_openapi.sh
AnPucel: I actually haven't added the deprecation logic yet. My plan was to run it through a separate regex and not include those matches. B/c from what I can tell, to include this in the current regex would require something that isn't currently supported in bash regex. B/c we'd be attempting to negate the word, which I can't seem to find an easy way to do. But I could be missing something.

AnPucel: This line just removes the quotes that get matched by the regex

scripts/gen_dynamic_openapi.sh
AnPucel: Okay. I looked at this and they're actually in a separate file `registry_util_ent.go`. So, I could just add a check to see if that file exists and then pull those in, but we'd have to run it from the enterprise repo. Or a separate script for enterprise.

scripts/gen_dynamic_openapi.sh
```diff
@@ -0,0 +1,89 @@
...
+        echo $plugin
+        #vault secrets enable ${plugin}
+    fi
+  done <$entRegFile
+  
+fi
+
+# Output OpenAPI, optionally formatted
+if [ "$1" == "-p" ]; then
+  curl -H "X-Vault-Token: root" "http://127.0.0.1:8200/v1/sys/internal/specs/openapi" | jq > openapi.json
```
VinnyHC, line 81: You might want to check that `jq` can be found:
```shell
if ! command -v jq &> /dev/null
then
    echo "jq could not be found"
    exit
fi

Or maybe just add a not to the assumptions comment at the top.

**Review by VinnyHC — COMMENTED** [review]
Quick comment about `jq`, but this looks like a good direction!

File: scripts/gen_dynamic_openapi.sh (hunk 3 above)
Comment (AnPucel) [reply scripts/gen_dynamic_openapi.sh]: I added some deprecated logic that just checks if the line contains "deprecated"

File: scripts/gen_dynamic_openapi.sh [hunk 6]
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
averche, line 1: Would it be better to modify the existing `gen_openapi.sh` rather than introducing a new one? If you do so, be sure to check whether it's run from CI and whether it still works after the change.

scripts/gen_dynamic_openapi.sh
```diff
@@ -0,0 +1,89 @@
...
+regex='".*"'
+while read line; do
+    if [[ $line == *"credentialBackends:"* ]] ; then
+        codeLineStarted=true
+    elif [ $codeLineStarted = true ] && [[ $line = *"}"* ]]  ; then
+        break
+    elif [ $codeLineStarted = true ] && [[ $line =~ $regex ]] && [[ $line != *"Deprecated"* ]] ; then
+        backend=${BASH_REMATCH[0]}
+        plugin=$(sed -e 's/^"//' -e 's/"$//' <<<"$backend") 
+        vault auth enable ${plugin}
```
averche, line 38:
```suggestion
        vault auth enable "${plugin}"
```

scripts/gen_dynamic_openapi.sh
```diff
@@ -0,0 +1,89 @@
...
+
+# Read auth backends
+codeLineStarted=false
+regex='".*"'
+while read line; do
+    if [[ $line == *"credentialBackends:"* ]] ; then
+        codeLineStarted=true
+    elif [ $codeLineStarted = true ] && [[ $line = *"}"* ]]  ; then
+        break
+    elif [ $codeLineStarted = true ] && [[ $line =~ $regex ]] && [[ $line != *"Deprecated"* ]] ; then
```
averche, line 35: Maybe inline the regex since `Deprecated` is already inlined: `$line =~ '".*"'` (not sure if this works with shell). Otherwise, maybe call it `inQuotesRegex` or something similar to give an idea of what this is doing

scripts/gen_dynamic_openapi.sh
```diff
@@ -0,0 +1,89 @@
...
+VAULT_PID=$!
+
+echo "Mounting all builtin backends..."
+
+# Read auth backends
+codeLineStarted=false
+regex='".*"'
+while read line; do
+    if [[ $line == *"credentialBackends:"* ]] ; then
+        codeLineStarted=true
```
averche, line 32:
```suggestion
        codeLinesStarted=true
```

(missing an s 😃)

scripts/gen_dynamic_openapi.sh
AnPucel: Let me see if it's being used anywhere. I'm unclear on who actually uses the other script and I figured I'd leave it in case the structure of the registry file changes and it completely breaks this script.

averche: I think we want `gen_openapi.sh` script to be up to date with latest plugins and this change would ensure that.

AnPucel: Sure! I just figured we were the only ones using it, so adding an additional script is nbd. But I can overwrite the old one.

scripts/gen_dynamic_openapi.sh
AnPucel: The inline gives me a shell check warning and doesn't seem to be matching a regex. I can rename it though.

scripts/gen_openapi.sh
```diff
@@ -87,4 +85,4 @@ fi
 kill $VAULT_PID
 sleep 1
 
-echo "\nopenapi.json generated."
+echo "\nopenapi.json generated."
```
kalafut, line 88: Results in:

```
\nopenapi.json generated.
```

kalafut:
It might be good for the relative path to be immune to run location, so you can run it from the repo root, scripts folder, wherever. On my first run I hit:

```
Mounting all builtin backends...
scripts/gen_openapi.sh: line 40: ../../vault/helper/builtinplugins/registry.go: No such file or directory
```

Some sample techniques for achieving this: https://stackoverflow.com/a/6659698, https://code-maven.com/bash-shell-relative-path

scripts/gen_openapi.sh
AnPucel: Good catch! thank you

scripts/gen_openapi.sh
```diff
@@ -9,6 +9,9 @@ set -e
 #   1. Vault has been checked out at an appropriate version and built
 #   2. vault executable is in your path
 #   3. Vault isn't already running
+#   4. Jq is installed
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
```
averche, line 14: `cd` is usually not a good idea for shell scripts. If the directory doesn't exist, the subsequent operations will be done on the wrong directory.

scripts/gen_openapi.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
averche, line 1: Is there a reason to switch this to bash?

averche (commented):
Looks great! Left a couple comments

scripts/gen_openapi.sh
```diff
@@ -9,6 +9,9 @@ set -e
 #   1. Vault has been checked out at an appropriate version and built
 #   2. vault executable is in your path
 #   3. Vault isn't already running
+#   4. Jq is installed
```
averche, line 12:
```suggestion
#   4. jq is installed
```

scripts/gen_openapi.sh
AnPucel: Yeah! I was trying to keep it generic shell. But bash supports the inline regex and `BASH_REMATCH`

scripts/gen_openapi.sh
AnPucel: Hmm. This was the recommendation to make the relative path immune to the run location. It seems like BASH_SOURCE should always exist b/c it's where you're running the script from. But I could be misunderstanding this

scripts/gen_openapi.sh
averche: That's a good reason 👍

scripts/gen_openapi.sh
averche: I'm not really familiar with `BASH_SOURCE` so I might be wrong here. Would it be possible to instead prefix every path with `"${BASH_SOURCE[0]}"`

AnPucel: hmmm. lemme try it

AnPucel: so, `BASH_SOURCE` is the path to the script file. It will always exist. So, I could change it for convention purposes, but this seems to be pretty common practice. I could change it for convention, I'm just not super familiar with shell script convention.

averche: OK, if this always works then it should be good 👍

averche (approved):
LGTM 👍

PR #14312, opened 2022-02-28
fix ent changelog check

swayne275:


HridoyRoy (approved):

.github/workflows/changelog-checker.yml
```diff
@@ -25,11 +25,16 @@ jobs:
       - name: Check for changelog entry in diff
         run: |
           # check if there is a diff in the changelog directory
-          changelog_files=$(git --no-pager diff --name-only HEAD "$(git merge-base HEAD "origin/${{ github.event.pull_request.base.ref }}")" -- changelog/${{ github.event.pull_request.number }}.txt)
+          if [ ${{ github.event.repository.name }} == "vault-enterprise" ]; then
+            expected_changelog_file=changelog/_${{ github.event.pull_request.number }}.txt
+          else
+            expected_changelog_file=changelog/${{ github.event.pull_request.number }}.txt
+          fi
```
mladlow, line 32: This seems like it should work, though I always doubt my ability to predict shell scripts until I run them. I tried to run parts of this and they seemed good though. Do you think it would be worth potentially doing an echo of "Looking for expected changelog file named: BLAH"?

mladlow (approved):

.github/workflows/changelog-checker.yml
swayne275: I had that in testing, I can certainly add it for future debugging!

swayne275: a9c691e40e

PR #14313, opened 2022-02-28
UI/ Client count permissions empty states

hellobontempo:
Some client count permissions cleanup:

- Adds error handling on component level
- Shows correct empty state when config off for users with limited permissions
Screen Shot 2022-03-01 at 12 42 38 PM

- Shows correct empty state when no data exists and user doesn't have config, activity or license permissions, so has to input billing start date.
Screen Shot 2022-03-01 at 10 36 01 AM

### DIFF
--- a/ui/app/components/clients/history.js
+++ b/ui/app/components/clients/history.js
@@ -68,15 +68,16 @@ export default class History extends Component {
         id: namespace.label,
       }))
     : [];
+  @tracked selectedAuthMethod = null;
+  @tracked authMethodOptions = [];
 
   // TEMPLATE MESSAGING
   @tracked noActivityDate = '';
   @tracked responseRangeDiffMessage = null;
   @tracked isLoadingQuery = false;
   @tracked licenseStartIsCurrentMonth = this.args.model.activity?.isLicenseDateError || false;
-
-  @tracked selectedAuthMethod = null;
-  @tracked authMethodOptions = [];
+  @tracked isError = false;
+  @tracked errorObject = null;
 
   get versionText() {
     return this.version.isEnterprise
@@ -213,7 +214,7 @@ export default class History extends Component {
         end_time: this.endTimeRequested,
       });
       if (response.id === 'no-data') {
-        // empty response is the only time we want to update the displayed date with the requested time
+        // empty response (204) is the only time we want to update the displayed date with the requested time
         this.startTimeFromResponse = this.startTimeRequested;
         this.noActivityDate = this.startTimeDisplay;
       } else {
@@ -221,6 +222,7 @@ export default class History extends Component {
         this.startTimeFromResponse = response.formattedStartTime;
         this.endTimeFromResponse = response.formattedEndTime;
         this.storage().setItem(INPUTTED_START_DATE, this.startTimeFromResponse);
+        this.isError = false;
       }
       this.queriedActivityResponse = response;
       this.licenseStartIsCurrentMonth = response.isLicenseDateError;
@@ -238,6 +240,8 @@ export default class History extends Component {
         this.responseRangeDiffMessage = null;
       }
     } catch (e) {
+      this.isError = true;
+      this.errorObject = e;
       return e;
     } finally {
       this.isLoadingQuery = false;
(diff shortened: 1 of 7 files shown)


ui/app/templates/components/clients/error.hbs
```diff
@@ -0,0 +1,25 @@
...
+      You must be granted permissions to view this page. Ask your administrator if you think you should have access to the
+      <code>{{@error.path}}</code>
+      endpoint.
+    </p>
+  {{else}}
+    <ul>
+      {{#if @error.message}}
+        <li>{{@error.message}}</li>
+      {{/if}}
+      <hr />
```
hashishaw, line 17: Should this line only show if a message exists?

ui/app/components/clients/history.js
```diff
@@ -238,6 +240,8 @@ export default class History extends Component {
         this.responseRangeDiffMessage = null;
       }
     } catch (e) {
+      this.isError = true;
+      this.errorObject = e;
```
hashishaw, line 244: I think we can get away with having only `errorObject`, set to null to start, and then check for that object in the template so we don't need both of these.

ui/tests/acceptance/client-history-test.js
```diff
@@ -275,4 +274,30 @@ module('Acceptance | clients history tab', function (hooks) {
...
+      this.get('/v1/sys/health', this.passthrough);
+      this.get('/v1/sys/seal-status', this.passthrough);
+      this.post('/v1/sys/capabilities-self', this.passthrough);
+      this.get('/v1/sys/internal/ui/feature-flags', this.passthrough);
+    });
+    await visit('/vault/clients/history');
+    assert.equal(currentURL(), '/vault/clients/history', 'clients/history URL is correct');
+    await click('[data-test-popup-menu-trigger="month"]');
+    const dropdownListMonths = this.element.querySelectorAll('[data-test-month-list] button');
+    // TODO CMB change how selected b/c depending on time of year first month or year may be disabled
```
hashishaw, line 293: Could you look for the first button that isn't disabled?

hashishaw (commented):
Elegant approach! A couple things to tidy, but nice work 👍

ui/tests/acceptance/client-history-test.js
hellobontempo: That's what I was thinking!

ui/tests/acceptance/client-history-test.js
```diff
@@ -272,7 +273,33 @@ module('Acceptance | clients history tab', function (hooks) {
...
+    this.server = new Pretender(function () {
+      this.get('/v1/sys/license/status', () => sendResponse(null, 403));
+      this.get('/v1/sys/version-history', () => sendResponse(null, 403));
+      this.get('/v1/sys/internal/counters/config', () => sendResponse(null, 403));
+      this.get('/v1/sys/internal/counters/activity', () => sendResponse(null, 403));
+      this.get('/v1/sys/health', this.passthrough);
+      this.get('/v1/sys/seal-status', this.passthrough);
+      this.post('/v1/sys/capabilities-self', this.passthrough);
+      this.get('/v1/sys/internal/ui/feature-flags', this.passthrough);
+    });
```
zofskeez, line 290: No need to change but a suggestion for the future is Mirage might be better in this situation since there are some common endpoints between the `client-history` and `client-current` tests. By creating a mirage handler you could share it between tests and all the endpoints would be defined in one place along with the `sendResponse` function.

ui/app/templates/components/clients/error.hbs
```diff
@@ -0,0 +1,25 @@
+<EmptyState
+  @title={{if (eq @error.httpStatus 403) "You are not authorized" "Error"}}
+  @subTitle={{concat "Error " @error.httpStatus}}
```
zofskeez, line 3: Just a heads up that the `concat` helper is fine but you should be able to do
`@subTitle="Error {{@error.httpStatus}}"`

zofskeez (approved):
Looks good to me. Thanks and nice work!

ui/app/templates/components/clients/error.hbs
hellobontempo: Good catch! I think I had an `if` there and forgot to pull out the `concat`

PR #14320, opened 2022-03-01
Mirage clean up

Monkeychip:
Removing unneeded mirage endpoints in the mirage base handler. This was used early for client count work, but we've moved those all to their own handlers (clients). I confirmed that the enterprise tests all pass locally.


zofskeez (approved):
Looks good thanks for cleaning this up!

PR #14321, opened 2022-03-01
Vault-4964: Update Managed Key documentation for AWS KMS

rculpepper:
This updates the documentation for the Managed Key API to include parameters for using AWS KMS as a backend.


website/content/api-docs/system/managed-keys.mdx
```diff
@@ -132,6 +132,45 @@ $ curl \
   a read-write session to the HSM. This is a boolean expressed as a string (e.g.
   `"true"`). This key is mainly to work around a limitation within AWS's CloudHSM v5
   pkcs11 implementation.
+  
+#### AWS KMS Parameters
+- `access_key` `(string: <required>)`: The AWS access key to use. This may also be provided in the
+`AWS_ACCESS_KEY_ID` environment variable or as part of the AWS profile from the AWS CLI or instance
```
cipherboy, line 138: Does this still hold after we did the config-means-config change to g-k-w? Same elsewhere below.

stevendpclark: Correct, we ignore all environment variables now, so this and the AWS_SECRET_ACCESS_KEY below needs to be removed from this doc.

rculpepper: The AWS environment variables still work. If the fields aren't passed in as parameters, the AWS code will look at the environment variables

stevendpclark: Ah you are correct, the aws go library itself will leverage them when we fall through without an access key along with a whole slew of other envs.

All except for AWS_KMS_ENDPOINT, that one is completely ignored as we set the "disallow_env_vars".

stevendpclark (approved):

divyapola5 (approved):

website/content/api-docs/system/managed-keys.mdx
```diff
@@ -132,6 +132,44 @@ $ curl \
...
+type is `ECDSA`, the `curve` field is required.
+
+- `kms_key_id` `(string: <required>)`: An identifier for the key. If using an existing key, this
+value should be an AWS alias for the key, including the `alias/` prefix. If Vault should generate
+the key, the provided ID will be added as an alias after the key is created. To fully specify the
+alias for the generated key, the provided key ID should be the full alias, including the `alias/`
+prefix. Otherwise, Vault will create the alias as `alias/hashicorp/{kms_key_id}`. The alias must be
+unique with the AWS account.
+
+    e.g., If the value provided is `kms_key_id=example-key-id`, then Vault will create the alias
```
victorr, line 162: Due to the indentation, this paragraph is going to appear as a code block:
image

Perhaps it would be better to use `~>` to highlight it as a note, or to simply leave it as a regular paragraph?

victorr (changes requested):

website/content/api-docs/system/managed-keys.mdx
```diff
@@ -132,6 +132,44 @@ $ curl \
   a read-write session to the HSM. This is a boolean expressed as a string (e.g.
   `"true"`). This key is mainly to work around a limitation within AWS's CloudHSM v5
   pkcs11 implementation.
+  
+#### AWS KMS Parameters
+- `access_key` `(string: <required>)`: The AWS access key to use. This may also be provided in the
+`AWS_ACCESS_KEY_ID` environment variable or as part of the AWS profile from the AWS CLI or instance
+profile.
+
+- `curve` `(string: "P256")`: The curve to use for an ECDSA key. This field is used when
```
victorr, line 141: There is no actual default value for `curve`: 
https://github.com/hashicorp/vault-enterprise/blob/0984b99ab637d21b4118950c90406ed4731cc161/vault/logical_system_managed_keys_ent.go#L344-L347

website/content/api-docs/system/managed-keys.mdx
```diff
@@ -132,6 +132,44 @@ $ curl \
...
+the key, the provided ID will be added as an alias after the key is created. To fully specify the
+alias for the generated key, the provided key ID should be the full alias, including the `alias/`
+prefix. Otherwise, Vault will create the alias as `alias/hashicorp/{kms_key_id}`. The alias must be
+unique with the AWS account.
+
+    e.g., If the value provided is `kms_key_id=example-key-id`, then Vault will create the alias
+    `hashicorp/example-key-id`. If the value is `kms_key_id=alias/example-key-id`, then Vault will
+    create the alias `example-key-id`.
+
+- `region` `(string: "us-east-1")`: The AWS region where the keys are stored (or will be stored). This
```
victorr, line 166: There is no actual default value for `region`.

website/content/api-docs/system/managed-keys.mdx
```diff
@@ -132,6 +132,44 @@ $ curl \
   a read-write session to the HSM. This is a boolean expressed as a string (e.g.
   `"true"`). This key is mainly to work around a limitation within AWS's CloudHSM v5
   pkcs11 implementation.
+  
+#### AWS KMS Parameters
```
ncabatoff, line 136: Would be good to also provide details on how to create the key in KMS if the user wishes to do so.  e.g. with terraform, key_type of ESCSA and curve of P256 is customer_master_key_spec = "ECC_NIST_P256".  Also the key usage must be SIGN_VERIFY, not ENCRYPT_DECRYPT like seal users might expect.

website/content/api-docs/system/managed-keys.mdx
rculpepper: This happens in the AWS code. It defaults to `us-east-1` if a region isn't passed in and isn't set in environment variables or AWS config.

website/content/api-docs/system/managed-keys.mdx
rculpepper: This is handled in the parsing for the key type. If a curve isn't supplied for an ECDSA key, it just returns `ECDSA_P256` as the key type

website/content/docs/enterprise/managed-keys.mdx
```diff
@@ -24,7 +24,7 @@ as the Managed Key for which it intends to use.
 ## Backend Support
 
 Managed Keys were developed to support different types of external backends.
-At this time only a PKCS#11 backend is supported, with the possibility of
+At this time supported backends are PKCS#11 and AWS KMS, with the possibility of
```
taoism4504, line 27:
```suggestion
Currently, supported backends are PKCS#11 and AWS KMS. 
```

website/content/docs/enterprise/managed-keys.mdx
```diff
@@ -24,7 +24,7 @@ as the Managed Key for which it intends to use.
...
 future integrations being added.
```
taoism4504, line 28:
```suggestion
Support for additional integrations may be added in the future.
```

taoism4504 (commented):
minor nits, but content looks good.

PR #14324, opened 2022-03-01
auth/ldap: add username_as_alias configurable

jasonodonnell:
This adds a new config parameter to the LDAP util SDK `username_as_alias`. A change in https://github.com/hashicorp/vault/pull/11000 resulted in alias names changing, which has broken a number of users in 1.9+. Using this parameter, an operator can change alias names back.

For example, if LDAP auth was configured with no binddn or bindpass, in 1.8.8 the alias name would be:
```
[~] vault read identity/alias/id/66c3a58f-9a72-26cf-f5a9-438528398eb3
Key                          Value
---                          -----
...
mount_accessor               auth_ldap_6af4105c
mount_path                   auth/ldap/
mount_type                   ldap
name                         bob
namespace_id                 root
```

However, in 1.9.3, the same config would result in alias names like this:
```
[~] vault read identity/alias/id/67656061-c8d2-8379-a7a5-dfa2c89659c9
Key                          Value
---                          -----
...
mount_accessor               auth_ldap_01e39e22
mount_path                   auth/ldap/
mount_type                   ldap
name                         cn=bob,CN=Users,DC=corp,DC=example,DC=net
namespace_id                 root
```

### DIFF
--- a/builtin/credential/ldap/path_login.go
+++ b/builtin/credential/ldap/path_login.go
@@ -103,6 +103,10 @@ func (b *backend) pathLogin(ctx context.Context, req *logical.Request, d *framew
 		},
 	}
 
+	if cfg.UsernameAsAlias {
+		auth.Alias.Name = username
+	}
+
 	cfg.PopulateTokenAuth(auth)
 
 	// Add in configured policies from mappings
--- a/sdk/helper/ldaputil/config.go
+++ b/sdk/helper/ldaputil/config.go
@@ -112,6 +112,12 @@ Default: ({{.UserAttr}}={{.Username}})`,
 			},
 		},
 
+		"username_as_alias": {
+			Type:        framework.TypeBool,
+			Default:     false,
+			Description: "If true, sets the alias name to the username",
+		},
+
 		"userattr": {
 			Type:        framework.TypeString,
 			Default:     "cn",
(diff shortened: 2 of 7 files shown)


jasonodonnell:
Known issue PR opened here: https://github.com/hashicorp/vault/pull/14325.

sdk/helper/ldaputil/config.go
```diff
@@ -242,6 +248,10 @@ func NewConfigEntry(existing *ConfigEntry, d *framework.FieldData) (*ConfigEntry
 		cfg.AnonymousGroupSearch = d.Get("anonymous_group_search").(bool)
 	}
 
+	if _, ok := d.Raw["username_as_alias"]; ok || !hadExisting {
```
calvn, line 251: Would it be better to check and error on the mutual exclusivity with `userfilter` somewhere in here? It would bubble up the error sooner on config rather than expect this behavior on login.

builtin/credential/ldap/path_login.go
```diff
@@ -103,6 +103,10 @@ func (b *backend) pathLogin(ctx context.Context, req *logical.Request, d *framew
 		},
 	}
 
+	if cfg.UsernameAsAlias && cfg.UserFilter == "" {
```
calvn, line 106: If we don't allow `cfg` to get into this state on `NewConfigEntry` we could just check against `cfg.UsernameAsAlias`.

sdk/helper/ldaputil/config.go
jasonodonnell: This is kind of annoying for the user because `userfilter` has a default (`({{.UserAttr}}={{.Username}})`), so you have to explicitly clear it out in the config.

FWIW, I did a lot of testing and could not think of a way to get an identity collision, unless you ignore the warning about `userattr` not being set in `userfilter`. Thoughts on not having this mutual exclusivity?

builtin/credential/ldap/path_login.go
jasonodonnell: The UX of this is annoying because `userfilter` has a default. Considering dropping it 🤔

calvn (approved):

calvn:
Should the label be all the way back to 1.8.10?

PR #14328, opened 2022-03-01
treat logical.ErrRelativePath as 400 instead of 500

ccapurso:
The userpass auth backend may return a `500 Internal Server Error` to attempts to read or write usernames that contain `..`. The underlying cause is a check performed in `StorageView. SanityCheck`. It is theoretically possible for other endpoints to result in a 500 response for a path that contains `..`. The policies endpoints, for example `sys/policies/acl/:path`, have their own error handling which results in returning a 400 for various errors:

```
❯ bin/vault read sys/policies/acl/foo..bar
Error reading sys/policies/acl/foo..bar: Error making API request.

URL: GET http://127.0.0.1:8200/v1/sys/policies/acl/foo..bar
Code: 400. Errors:

* failed to read policy: relative paths not supported
```

The proposed fix is to treat `logical.ErrRelativePath` as a 400. Rather than do this directly within the userpass logic, it is done at a higher level in the request handling logic so that it has broad coverage across any backend.

```
❯ bin/vault read auth/userpass/users/foo..bar
Error reading auth/userpass/users/foo..bar: Error making API request.

URL: GET http://127.0.0.1:8200/v1/auth/userpass/users/foo..bar
Code: 400. Errors:

* 1 error occurred:
        * relative paths not supported
```

### DIFF
--- a/vault/request_handling.go
+++ b/vault/request_handling.go
@@ -321,7 +321,7 @@ func (c *Core) checkToken(ctx context.Context, req *logical.Request, unauth bool
 			// fail later via bad path to avoid confusing items in the log
 			checkExists = false
 		case logical.ErrRelativePath:
-			return nil, te, err
+			return nil, te, errutil.UserError{Err: err.Error()}
 		case nil:
 			if existsResp != nil && existsResp.IsError() {
 				return nil, te, existsResp.Error()
--- a/vault/request_handling.go
+++ b/vault/request_handling.go
@@ -320,6 +320,8 @@ func (c *Core) checkToken(ctx context.Context, req *logical.Request, unauth bool
 		case logical.ErrUnsupportedPath:
 			// fail later via bad path to avoid confusing items in the log
 			checkExists = false
+		case logical.ErrRelativePath:
+			return nil, te, err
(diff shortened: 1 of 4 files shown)


vault/request_handling.go
```diff
@@ -320,6 +320,8 @@ func (c *Core) checkToken(ctx context.Context, req *logical.Request, unauth bool
 		case logical.ErrUnsupportedPath:
 			// fail later via bad path to avoid confusing items in the log
 			checkExists = false
+		case logical.ErrRelativePath:
```
HridoyRoy, line 323: Is this a UserError (line 332 below), and if not, is it ok to return?

HridoyRoy (changes requested):

vault/request_handling.go
ccapurso: Ah, I think you're right! Thanks for calling that out. I addressed this in https://github.com/hashicorp/vault/pull/14328/commits/c1e527d4886939aaded77c1de010fc62b3d20eaf.

HridoyRoy (approved):
Lgtm!

PR #14335, opened 2022-03-02
validate mount accessor while configuring MFA enforcement

hghaf099:
The mount accessor on the mfa method cannot be different from the accessor being logged in on.
So say I have an entity entity1  with alias12 on userpass12 and alias23 on userpass23. My email address on Okta is alias12@hashicorp,com. I try to login on userpass23 with alias23 .  It detects I am entity1 , looks up my alias on userpass12 which is configured on the okta mfa method, renders out my okta username as [alias12@hashicorp.com](mailto:alias12@hashicorp.com) , and allows me to login to userpass23.

### DIFF
--- a/changelog/14335.txt
+++ b/changelog/14335.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+auth/login mfa: The mount accessor on the MFA method config cannot be different from the accessor being logged in on.
+```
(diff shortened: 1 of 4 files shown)


hghaf099:
I am going to close this issue as this is not considered a bug. It is up to the admin to decide if the mfa enforcement accessor should match the MFA methodIDs or not.

PR #14344, opened 2022-03-02
LinkTo remove tagName lint warning

Monkeychip:
This PR removes the tagName attr in the LinkTo components. In doing so I was able to also remove the nested LinkTo. It appears the issue was styling, which I fixed.

Additionally, because of the Toggle Components use of the tagName while removing it I glimmerized the component—it made it easier to remove.

Every instance where I remove the tagName I tested that link to make sure the look and functionality worked. I also tested the glimmerized Toggle component. However, I am going to 1.11 this. That way if there are issues we can spot them before the next binary. This is clean up work and not bug or functionality.


ui/app/styles/core/toggle.scss
```diff
@@ -1,34 +1,34 @@
 /* COPIED FROM BULMA/SWITCH */
-.toggle[type='checkbox'] {
```
Monkeychip, line 2: changed type to class for warning issue.

ui/app/templates/components/configure-aws-secret.hbs
```diff
@@ -61,6 +45,7 @@
       (hash access_key=@accessKey iam_endpoint=@iamEndpoint sts_endpoint=@stsEndpoint secret_key=@secretKey region=@region)
     }}
     data-test-aws-root-creds-form="true"
+    aria-label="save root creds form"
```
Monkeychip, line 48: helps with linting warning about two form element being on the same page that need to be uniquely identified.

ui/app/templates/vault/cluster/secrets/backend/roles.hbs
```diff
@@ -1,143 +0,0 @@
-<PageHeader as |p|>
-  <p.top>
```
Monkeychip, line 2: This file was not being used anywhere. Double checked in the test suite and the router.

ui/app/styles/components/tabs.scss
```diff
@@ -8,6 +8,17 @@
   ul {
     border-color: transparent;
     min-height: 3rem;
+
+    > a {
```
Monkeychip, line 12: the original styling for when it was inside a "li" element. Keeping that old styling just in case and duplicating it here when it's directly under a ul element.

ui/lib/core/addon/components/toggle.js
```diff
@@ -15,29 +15,40 @@
  * @param {string} [status='normal'] - Status can be normal or success, which makes the switch have a blue background when checked=true
  */
 
-import Component from '@ember/component';
-import { computed } from '@ember/object';
+import Component from '@glimmer/component';
 import layout from '../templates/components/toggle';
+import { setComponentTemplate } from '@ember/component';
```
zofskeez, line 20: Heads up that I don't think you need `setComponentTemplate` or to import the `layout` if you were to move the template file to the same directory as the component js file. The `autocomplete-input` component is an example of that in the core addon.

ui/lib/core/addon/components/toggle.js
```diff
@@ -15,29 +15,40 @@
...
-  },
-});
+  }
+
+  @action
+  handleChange(e) {
+    this.args.onChange(e.target.checked);
+  }
+}
+
```
zofskeez, line 53: Thanks for glimmerizing! :sparkler:

zofskeez (approved):
Thanks for tackling this and nice work! Getting rid of the nested LinkedTo's cleans things up so much.

ui/lib/core/addon/components/toggle.js
Monkeychip: Oh, let me try this!

Issue #14351, opened 2022-03-03
A way to forcibly unmount a secret engine


benvanstaveren:
A long, long time ago we ran vault with the Consul KV backend. Somewhere along the lines, Consul crapped itself. In the end we ended up with a PKI secrets engine that has no data, no configuration, no nothing except showing up in secrets list. Doing a secrets disable on it either results in a 500 error, or a timeout.

**Describe the solution you'd like**
I'd like to be able to forcibly disable a secret engine, no questions asked, no errors returned, just nuked completely. 

**Explain any additional use-cases**
I've not encountered any other issues that would require the "nuclear" option, but...

swayne275:
thanks for the suggestion! I'll raise it for discussion.

swayne275:
I'm going to close this based on the following (but please re-open if it doesn't resolve your issue):

When a secrets engine is revoked, Vault will [revoke secrets associated with this engine](https://www.vaultproject.io/docs/commands/secrets/disable). You are likely seeing the error during this process.

Since you mentioned a timeout, the best course of action is likely to increase that timeout such that there is enough time for the secrets to be revoked.

Failing this, you can "forcibly unmount" a secrets engine by doing a [force prefix revoke](https://www.vaultproject.io/docs/commands/lease/revoke) for the secrets mount in question (something of the form `vault lease revoke -force -prefix <secrets prefix>`

PR #14363, opened 2022-03-03
add tip for how to force a secrets engine disable

swayne275:
this is a follow-up to a [recently raised issue](https://github.com/hashicorp/vault/issues/14351)

### DIFF
--- a/website/content/api-docs/system/mounts.mdx
+++ b/website/content/api-docs/system/mounts.mdx
@@ -212,6 +212,23 @@ $ curl \
     http://127.0.0.1:8200/v1/sys/mounts/my-mount
 ```
 
+### Force Disable
+
+Because disabling a secrets engine revokes secrets associated with this mount,
+there can be errors that prevent the secrets engine from being disabled if the
+revocation fails.
+
+The best way to resolve this is to figure out the underlying issue, and then
+disable the secrets engine once that has been solved. Often, this can be as
+simple as increasing the timeout (in the event of timeout errors).
+
+For recovery situations where the secret has been manually removed from the
+secrets backing service, one can "force" a secrets engine disable in Vault by
+performing a [force revoke](https://www.vaultproject.io/api/system/leases#revoke-force)
+on the mount prefix, followed by a secrets disable when that completes. Note
+that this method may result in dangling credentials if the underlying secrets
+were not manually cleaned up. This is meant for extreme circumstances.
+
 ## Get the configuration of a Secret Engine
 
 This endpoint returns the configuration of a specific secret engine.
--- a/website/content/api-docs/system/mounts.mdx
+++ b/website/content/api-docs/system/mounts.mdx
@@ -222,12 +222,12 @@ The best way to resolve this is to figure out the underlying issue and then
 disable the secrets engine once the underlying issue is resolved. Often, this can be as
 simple as increasing the timeout (in the event of timeout errors).
 
-For recovery situations where the secret has been manually removed from the
+For recovery situations where the secret was manually removed from the
 secrets backing service, one can force a secrets engine disable in Vault by
 performing a [force revoke](https://www.vaultproject.io/api/system/leases#revoke-force)
 on the mount prefix, followed by a secrets disable when that completes. 
-This method may result in dangling credentials if the underlying secrets
-were not manually cleaned up. This is meant for extreme circumstances.
+If the underlying secrets were not manually cleaned up, this method might result
+in dangling credentials. This is meant for extreme circumstances.
 
 ## Get the configuration of a Secret Engine
 
--- a/website/content/api-docs/system/mounts.mdx
+++ b/website/content/api-docs/system/mounts.mdx
@@ -226,7 +226,7 @@ For recovery situations where the secret has been manually removed from the
 secrets backing service, one can "force" a secrets engine disable in Vault by
 performing a [force revoke](https://www.vaultproject.io/api/system/leases#revoke-force)
 on the mount prefix, followed by a secrets disable when that completes. 
-that this method may result in dangling credentials if the underlying secrets
+This method may result in dangling credentials if the underlying secrets
 were not manually cleaned up. This is meant for extreme circumstances.
 
 ## Get the configuration of a Secret Engine
--- a/website/content/api-docs/system/mounts.mdx
+++ b/website/content/api-docs/system/mounts.mdx
@@ -215,7 +215,7 @@ $ curl \
 ### Force Disable
 
 Because disabling a secrets engine revokes secrets associated with this mount,
-there can be errors that prevent the secrets engine from being disabled if the
+possible errors can prevent the secrets engine from being disabled if the
 revocation fails.
 
 The best way to resolve this is to figure out the underlying issue, and then
--- a/website/content/api-docs/system/mounts.mdx
+++ b/website/content/api-docs/system/mounts.mdx
@@ -219,7 +219,7 @@ possible errors can prevent the secrets engine from being disabled if the
 revocation fails.
 
 The best way to resolve this is to figure out the underlying issue and then
-disable the secrets engine once that has been solved. Often, this can be as
+disable the secrets engine once the underlying issue is resolved. Often, this can be as
 simple as increasing the timeout (in the event of timeout errors).
 
 For recovery situations where the secret has been manually removed from the
--- a/website/content/api-docs/system/mounts.mdx
+++ b/website/content/api-docs/system/mounts.mdx
@@ -218,7 +218,7 @@ Because disabling a secrets engine revokes secrets associated with this mount,
 possible errors can prevent the secrets engine from being disabled if the
 revocation fails.
 
-The best way to resolve this is to figure out the underlying issue, and then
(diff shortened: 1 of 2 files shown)


website/content/docs/commands/secrets/disable.mdx
```diff
@@ -30,3 +30,11 @@ $ vault secrets disable aws/
 
 There are no flags beyond the [standard set of flags](/docs/commands)
 included on all commands.
+
+## Force Disable
+
+Because `secrets disable` revokes secrets associated with this mount, there can
+be errors that prevent the secrets engine from being disabled if part of the
+revocation fails. One can get around this by performing a
+[prefix force revoke](/docs/commands/lease/revoke) on the mount, followed by a
```
ncabatoff, line 39: I would make sure to tell them to first try increasing the timeout.  And put some big warnings around the use of the force flag, since it may lead to dangling credentials (e.g. the dynamic secrets that aren't getting revoked as a result linger, the corresponding db/cloud creds are still around and a potential security issue.)

swayne275:
[this](https://docs.google.com/document/d/1MGQDNSdxDb2LsV82eVScz2W1ISfxLchldSIJCvgnutQ/edit#heading=h.brj6yyjc49yp) makes it seem like I don't actually need the `backport/1.9.x` label, but I'm not sure. does it cause any harm?

website/content/docs/commands/secrets/disable.mdx
swayne275: how's this: 7fe721670e

benvanstaveren: I'll throw in my 2 cents as the raiser of the issue referenced and say that it worked. In our case it was a PKI mount so no dangly bits.

website/content/api-docs/system/mounts.mdx
```diff
@@ -212,6 +212,23 @@ $ curl \
     http://127.0.0.1:8200/v1/sys/mounts/my-mount
 ```
 
+### Force Disable
+
+Because disabling a secrets engine revokes secrets associated with this mount,
+there can be errors that prevent the secrets engine from being disabled if the
```
taoism4504, line 218:
```suggestion
errors can prevent the secrets engine from being disabled if the
```

website/content/api-docs/system/mounts.mdx
```diff
@@ -212,6 +212,23 @@ $ curl \
...
+revocation fails.
+
+The best way to resolve this is to figure out the underlying issue, and then
```
taoism4504, line 221:
```suggestion
The best way to resolve this is to figure out the underlying issue and then
```

website/content/api-docs/system/mounts.mdx
```diff
@@ -212,6 +212,23 @@ $ curl \
...
+disable the secrets engine once that has been solved. Often, this can be as
```
taoism4504, line 222:
```suggestion
disable the secrets engine once the underlying issue is resolved. Often, this can be as
```

website/content/api-docs/system/mounts.mdx
```diff
@@ -212,6 +212,23 @@ $ curl \
...
+simple as increasing the timeout (in the event of timeout errors).
+
+For recovery situations where the secret has been manually removed from the
+secrets backing service, one can "force" a secrets engine disable in Vault by
+performing a [force revoke](https://www.vaultproject.io/api/system/leases#revoke-force)
+on the mount prefix, followed by a secrets disable when that completes. Note
+that this method may result in dangling credentials if the underlying secrets
```
taoism4504, line 229:
```suggestion
This method may result in dangling credentials if the underlying secrets
```

website/content/docs/commands/secrets/disable.mdx
```diff
@@ -30,3 +30,20 @@ $ vault secrets disable aws/
...
 included on all commands.
+
+## Force Disable
+
+Because `secrets disable` revokes secrets associated with this mount, there can
+be errors that prevent the secrets engine from being disabled if the revocation
+fails.
+
+The best way to resolve this is to figure out the underlying issue, and then
+disable the secrets engine once that has been solved. Often, this can be as
```
taoism4504, line 41: 'that' is an antecedent, so does 'that' refer to the underlying issue here? If so, consider rewriting:

The best way to resolve this is to figure out the underlying issue, and then disable the secrets engine once the underlying issue is resolved.

website/content/docs/commands/secrets/disable.mdx
```diff
@@ -30,3 +30,20 @@ $ vault secrets disable aws/
...
+simple as increasing the timeout (in the event of timeout errors).
+
+For recovery situations where the secret has been manually removed from the
+secrets backing service, one can "force" a secrets engine disable in Vault by
```
taoism4504, line 45: no need for quotations around the word force.

website/content/docs/commands/secrets/disable.mdx
```diff
@@ -30,3 +30,20 @@ $ vault secrets disable aws/
...
+performing a [prefix force revoke](/docs/commands/lease/revoke) on the mount
+prefix, followed by a `secrets disable` when that completes. Note that this
```
taoism4504, line 47:
```suggestion
prefix, followed by a `secrets disable` when that completes. 
```

website/content/docs/commands/secrets/disable.mdx
```diff
@@ -30,3 +30,20 @@ $ vault secrets disable aws/
...
+method may result in dangling credentials if the underlying secrets were not
```
taoism4504, line 48:
```suggestion
This method may result in dangling credentials if the underlying secrets were not
```

website/content/api-docs/system/mounts.mdx
```diff
@@ -212,6 +212,23 @@ $ curl \
...
+
+Because disabling a secrets engine revokes secrets associated with this mount,
+possible errors can prevent the secrets engine from being disabled if the
+revocation fails.
+
+The best way to resolve this is to figure out the underlying issue and then
+disable the secrets engine once the underlying issue is resolved. Often, this can be as
+simple as increasing the timeout (in the event of timeout errors).
+
+For recovery situations where the secret has been manually removed from the
```
taoism4504, line 225:
```suggestion
For recovery situations where the secret was manually removed from the
```

website/content/api-docs/system/mounts.mdx
```diff
@@ -212,6 +212,23 @@ $ curl \
...
+secrets backing service, one can force a secrets engine disable in Vault by
```
taoism4504, line 226: I wasn't sure if secrets engine disable should be wrapped as code or if you meant to say ...one can force a secrets engine to disable...

website/content/api-docs/system/mounts.mdx
```diff
@@ -212,6 +212,23 @@ $ curl \
...
+performing a [force revoke](https://www.vaultproject.io/api/system/leases#revoke-force)
+on the mount prefix, followed by a secrets disable when that completes. 
+This method may result in dangling credentials if the underlying secrets
```
taoism4504, line 229:
```suggestion
This method might result in dangling credentials if the underlying secrets
```

taoism4504: tl;dr
The use of 'may' or 'might' can be a bit tricky. This sentence is a mixed conditional with a past tense verb (were) in the if/when clause, and the future tense model 'may' appears in the other clause. In this case, 'might' would be the better option to use in place of 'may'.

website/content/api-docs/system/mounts.mdx
```diff
@@ -212,6 +212,23 @@ $ curl \
...
+revocation fails.
+
+The best way to resolve this is to figure out the underlying issue and then
+disable the secrets engine once the underlying issue is resolved. Often, this can be as
+simple as increasing the timeout (in the event of timeout errors).
+
+For recovery situations where the secret has been manually removed from the
+secrets backing service, one can force a secrets engine disable in Vault by
+performing a [force revoke](https://www.vaultproject.io/api/system/leases#revoke-force)
+on the mount prefix, followed by a secrets disable when that completes.
```
taoism4504, line 228: unsure whether secrets disable needs to be wrapped in code

website/content/api-docs/system/mounts.mdx
taoism4504: I'd also suggest rewriting the sentence to the following:
If the underlying secrets were not manually cleaned up, this method might result in dangling credentials.

website/content/docs/commands/secrets/disable.mdx
```diff
@@ -30,3 +30,20 @@ $ vault secrets disable aws/
 
 There are no flags beyond the [standard set of flags](/docs/commands)
 included on all commands.
+
+## Force Disable
```
taoism4504, line 34: Same comments/suggestions as above.

taoism4504 (approved):
A few suggestions/comments but otherwise LG.

website/content/api-docs/system/mounts.mdx
swayne275: I don't want it here since `secrets disable` is a command and this is for the API

PR #14369, opened 2022-03-03
rename Dockerfile build-arg VERSION to PRODUCT_VERSION

alvin-huang:
We pass in `PRODUCT_VERSION` in our [docker build action](https://github.com/hashicorp/actions-docker-build/blob/f24c46a02ebc09617d0ce5ecaa4dbcc79a150ef8/scripts/docker_build#L41) so we need to reference it in our Dockerfile here to set the right version label. 

The concern is there may be folks using the current Dockerfile with tooling that sets `--build-arg VERSION=x.y.z` and this will break them especially if they check the label for the current version like we do in the action. What do you think @samsalisbury ?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,10 +51,10 @@ dev-dynamic-mem: dev-dynamic
 # Creates a Docker image by adding the compiled linux/amd64 binary found in ./bin.
 # The resulting image is tagged "vault:dev".
 docker-dev: prep
-	docker build --build-arg PRODUCT_VERSION=$(GO_VERSION_MIN) --build-arg BUILD_TAGS="$(BUILD_TAGS)" -f scripts/docker/Dockerfile -t vault:dev .
+	docker build --build-arg VERSION=$(GO_VERSION_MIN) --build-arg BUILD_TAGS="$(BUILD_TAGS)" -f scripts/docker/Dockerfile -t vault:dev .
 
 docker-dev-ui: prep
-	docker build --build-arg PRODUCT_VERSION=$(GO_VERSION_MIN) --build-arg BUILD_TAGS="$(BUILD_TAGS)" -f scripts/docker/Dockerfile.ui -t vault:dev-ui .
+	docker build --build-arg VERSION=$(GO_VERSION_MIN) --build-arg BUILD_TAGS="$(BUILD_TAGS)" -f scripts/docker/Dockerfile.ui -t vault:dev-ui .
 
 # test runs the unit tests and vets the code
 test: prep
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -51,10 +51,10 @@ dev-dynamic-mem: dev-dynamic
 # Creates a Docker image by adding the compiled linux/amd64 binary found in ./bin.
 # The resulting image is tagged "vault:dev".
 docker-dev: prep
-	docker build --build-arg VERSION=$(GO_VERSION_MIN) --build-arg BUILD_TAGS="$(BUILD_TAGS)" -f scripts/docker/Dockerfile -t vault:dev .
+	docker build --build-arg PRODUCT_VERSION=$(GO_VERSION_MIN) --build-arg BUILD_TAGS="$(BUILD_TAGS)" -f scripts/docker/Dockerfile -t vault:dev .
 
 docker-dev-ui: prep
-	docker build --build-arg VERSION=$(GO_VERSION_MIN) --build-arg BUILD_TAGS="$(BUILD_TAGS)" -f scripts/docker/Dockerfile.ui -t vault:dev-ui .
+	docker build --build-arg PRODUCT_VERSION=$(GO_VERSION_MIN) --build-arg BUILD_TAGS="$(BUILD_TAGS)" -f scripts/docker/Dockerfile.ui -t vault:dev-ui .
```
samsalisbury, line 57: This and the target above seem to be referencing `./scripts/docker/Dockerfile[.ui]` so they shouldn't need updating based on the change to the `./Dockerfile` as far as I can see. The `VERSION` in these Dockerfiles is the Go version, as opposed to the product version. (This is why I always use `PRODUCT_VERSION` for the product version, not just `VERSION` as it helps to disambiguate all the different things with versions we're dealing with :) )

alvin-huang: Good catch! Just changed it back!

samsalisbury (approved):

alvin-huang:
@samsalisbury do you think renaming `VERSION` will be an issue for anyone that might happen to be building the Dockerfile themselves (externally)?

samsalisbury:
> @samsalisbury do you think renaming `VERSION` will be an issue for anyone that might happen to be building the Dockerfile themselves (externally)?

Hi @alvin-huang, if other people are using this particular Dockerfile it would break their build, but I wouldn't be too concerned about it because this is a new(ish) file, and is geared really specifically towards producing release builds using CRT. See especially the `COPY` on line 23.

alvin-huang:
Sounds good to me, thanks! @mladlow, tagged you as more of an FYI

alvin-huang:
@mladlow this should be backported to any release lines you will be releasing with CRT

mladlow:
@alvin-huang typically we'd ask the PR author to handle the backports. In the event of conflicts, they're the most familiar with how to resolve them. We shouldn't need any additional 1.7 releases, but would need backports to `release/1.10.x`, `release/1.9.x`, and `release/1.8.x`. Can you all take this on? If that's not possible, I can find someone else to do it.

PR #14378, opened 2022-03-04
Vault-4964: Update Managed Key documentation for AWS KMS

rculpepper:
This updates the documentation for the Managed Key API to include parameters for using AWS KMS as a backend.

(These are the same changes as #14321 but I had to remake the PR due to a git config problem.)


victorr (approved):
LGTM!

PR #14379, opened 2022-03-04
UI/ Fix version check typo

hellobontempo:


ui/app/routes/vault/cluster.js
```diff
@@ -40,7 +40,7 @@ export default Route.extend(ModelBoundaryRoute, ClusterRoute, {
     const currentTokenName = this.auth.get('currentTokenName');
     // if no namespace queryParam and user authenticated,
     // use user's root namespace to redirect to properly param'd url
-    if (this.featureFlagService.managedNamespaceRoot && this.version.isOss) {
+    if (this.featureFlagService.managedNamespaceRoot && this.version.isOSS) {
```
zofskeez, line 43: Good catch! I'm wondering if this check is necessary if we're only raising an error to the console? I'm not familiar with the feature flags though so it must have been added with purpose. As far as a test goes you could stub or spy the `console.error` method and make sure it gets hit but since this doesn't actually have any impact on the UI I'm not sure it's necessary.

zofskeez (approved):

ui/app/routes/vault/cluster.js
hellobontempo: Good call - thank you!

PR #14383, opened 2022-03-04
Updated Docs For Specific Call Out

joe-rua:
The Parameter `cidr_list` is not support for Key_Type CA, customer was confused on this, so I feel we should specifically call this out to ensure there is no confusion


joe-rua:
This is called out in this line of code https://github.com/hashicorp/vault/blob/dc086173df7ac318edf33af33f79b97f4c8e9555/builtin/logical/ssh/path_roles.go#L124-L133

cipherboy (approved):
Makes sense, thanks!

PR #14385, opened 2022-03-05
Set systemd service type to notify

### LINKED ISSUE #14371: systemd notify capabilities are not enabled by default
**Describe the bug**
In [#11517](https://github.com/hashicorp/vault/pull/11517), the ability to notify systemd with the status of the vault server was added but the default Vault systemd unit file provided in the official packages does not set `Type=notify` so this feature doesn't ever get used when installing through package managers. 

I only happened to realize this was even a feature due to having debug logs enabled on a dev environment where I saw the log message saying `would have sent systemd notification (systemd not present)` and got a bit worried since this machine definitely is a systemd box. Took a bit of digging to find the commit adding this and the comment about the systemd service having to have the `Type=notify` in it.

I don't see any reason that the unit file for Vault shouldn't be updated with `Type=notify` set and am happy to open a PR if there aren't any objections to it.

**To Reproduce**
1. Install vault from the official Apt repos
2. Start vault with debug logging enabled
3. Notice the message `would have sent systemd notification (systemd not present)`  shortly after starting up.
4. Run `systemctl edit vault` and paste the code below in the editor and save.
```
[Service]
Type=notify
```
5. Run `systemctl daemon-reload && systemctl restart vault` and you will now see the message `sent systemd notification: notification=READY=1`

**Expected behavior**
Vault's systemd unit file should properly configure it's type as `notify` when installed from a package.

**Environment:**
* Vault Server Version: 1.9.4
* Vault CLI Version: 1.9.4
* Server Operating System/Architecture: Ubuntu 20.04 AMD64

```hcl
max_lease_ttl      = "420h"
default_lease_ttl  = "1h"
log_level          = "Debug"
api_addr           = "https://192.168.1.69:8200"
cluster_addr       = "https://192.168.1.69:8201"
plugin_directory   = "/etc/vault.d/plugins"

seal "transit" {
  disable_renewal = "false"
  key_name = "autounseal"
  mount_path = "transit/"
}

ui = true

storage "raft" {
  path    = "/opt/vault/data"
  node_id = "vault-2"
}

listener "tcp" {
  address         = "0.0.0.0:8200"
  tls_disable     = 0
  tls_cert_file   = "/etc/vault.d/ssl/vault.crt"
  tls_key_file    = "/etc/vault.d/ssl/vault.key"
  tls_min_version = "tls12"
}

```

RickyGrassmuck:
Updates the systemd unit file shipped in the Linux packages to set the service type to `notify`

Fixes #14371

### DIFF
--- a/.release/linux/package/usr/lib/systemd/system/vault.service
+++ b/.release/linux/package/usr/lib/systemd/system/vault.service
@@ -8,6 +8,7 @@ StartLimitIntervalSec=60
 StartLimitBurst=3
 
 [Service]
+Type=notify
 EnvironmentFile=/etc/vault.d/vault.env
 User=vault
 Group=vault
--- a/changelog/14385.txt
+++ b/changelog/14385.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+core: Systemd unit file included with the Linux packages now sets the service type to notify.
+```


ncabatoff (approved):

ncabatoff:
Thanks @RickyGrassmuck !

Issue #14202, opened 2022-02-22
vault agent should be more verbose about tls_server_name errors


grahamc:
**Describe the bug**

If I configure a Vault Agent to connect to a server but provide an invalid `tls_server_name`, the agent fails in a hard to diagnose way.

**To reproduce**
With the following configuration:

```hcl
vault {
    address = "https://example.com"
    ca_cert = "./cacert.pem"
}
```

we get a reasonable error:

```
==> Vault agent started! Log data will stream in below:

==> Vault agent configuration:

                     Cgo: enabled
               Log Level: info
                 Version: Vault v1.9.3
             Version Sha: v1.9.3

2022-02-22T15:21:33.844-0500 [INFO]  template.server: starting template server
2022-02-22T15:21:33.844-0500 [INFO] (runner) creating new runner (dry: false, once: false)
2022-02-22T15:21:33.844-0500 [INFO]  auth.handler: starting auth handler
2022-02-22T15:21:33.844-0500 [INFO]  auth.handler: authenticating
2022-02-22T15:21:33.844-0500 [INFO]  sink.server: starting sink server
2022-02-22T15:21:33.844-0500 [INFO] (runner) creating watcher
2022-02-22T15:21:33.957-0500 [ERROR] auth.handler: error authenticating: error="Put \"https://example.com/v1/auth/approle/login\": x509: certificate signed by unknown authority" backoff=1s
```

However we get no diagnostics if we add a mismatched `tls_server_name` directive. For example, with the following configuration:

```hcl
vault {
    address = "https://example.com"
    ca_cert = "./cacert.pem"
    tls_server_name = "totally-bogus.com"
}
```

we get:

```
$ vault agent -config=./agent-config.hcl -log-level=debug
==> Vault agent started! Log data will stream in below:

==> Vault agent configuration:

                     Cgo: enabled
               Log Level: debug
                 Version: Vault v1.9.3
             Version Sha: v1.9.3

2022-02-22T15:23:09.879-0500 [INFO]  auth.handler: starting auth handler
2022-02-22T15:23:09.879-0500 [INFO]  auth.handler: authenticating
2022-02-22T15:23:09.879-0500 [INFO]  template.server: starting template server
2022-02-22T15:23:09.879-0500 [INFO]  sink.server: starting sink server
2022-02-22T15:23:09.879-0500 [INFO] (runner) creating new runner (dry: false, once: false)
[...]

PR #14386, opened 2022-03-07
Error handling for TLS server name and Vault address mismatch in Vault Agent

ryowright:
Added more verbose error handling for when there is a mismatch between the Vault address and the TLS server name in the configuration file (under the Vault stanza) for setting up Vault Agent. Fix for issue https://github.com/hashicorp/vault/issues/14202.

### DIFF
--- a/command/agent/auth/auth.go
+++ b/command/agent/auth/auth.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"math/rand"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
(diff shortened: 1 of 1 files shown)


hsimon-hashicorp:
Thanks @ryowright! Our engineers are having a hackweek this week, but I'll get this reviewed as soon as I can. Appreciate your efforts as always! :)

ryowright:
Of course! Always glad to help where I can.

command/agent/auth/auth.go
```diff
@@ -212,6 +216,20 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
 			}
 		}
 
+		clientAddr = clientToUse.Address()
+		tlsServerName = clientToUse.CloneConfig().HttpClient.Transport.(*http.Transport).TLSClientConfig.ServerName
+		if tlsServerName != "" && tlsServerName != clientAddr {
```
ncabatoff, line 221: There's no requirement that tls server name matches the address we use, which e.g. could be an IP address.

VioletHynes:
Hey there, thanks for your contribution! I'm going to close this. As mentioned before, there's no requirement that these two must match, and it would be a breaking change to introduce such a requirement. Since then, this has grown a little stale and has conflicts.

PR #14396, opened 2022-03-07
[QTI-188] Update the UI tests to be able to run against a cluster on a separate host.

mikebaum:
* fixed the `auth.js` file for the ui tests to receive the root token rather than use a hard-coded value
* fixed the `unseal-test` to be flexible to handle cases where there are more than one unseal/recovery key
* added a script (`enos-test-ember.js`), that can be used to configure and run `ember test` using the following configuration from the environment:
   * **VAULT_UNSEAL_KEYS** - a json string formatted list of recovery keys to use when unsealing vault.
   *  **VAULT_TOKEN** - the vault root token to use for authentication
   *  **VAULT_ADDR** - the address of the vault cluster to test
   *  **TEST_FILTERS** - a json formatted list of test filters to add to the `ember test` command via the option `-f=<filter>`
* Added the `package.json`: `test:enos` target to launch the ui tests against a real backend as configured via the environment variables listed above.
* Added a Makefile target to the root project `Makefile`, `test-ember-enos` that will install the npm dependencies and run the `package.json` target to run the tests.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -26,6 +26,7 @@
     "start2": "ember server --proxy=http://localhost:8202 --port=4202",
     "start:mirage": "start () { MIRAGE_DEV_HANDLER=$1 yarn run start; }; start",
     "test": "npm-run-all lint:js:quiet lint:hbs:quiet && node scripts/start-vault.js",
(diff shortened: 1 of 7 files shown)


hashishaw (approved):
Looks great, thank you! 🚀

PR #14397, opened 2022-03-07
Add backing code for the pki add-root command.

victorr:
Work in progress.


command/pkicli/high_level_ops.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, err
+	}
+
+	// 3. Set the config URLs -- Is this necessary for intermediates??
+	err = p.configUrls(params)
+	if err != nil {
+		return nil, err
+	}
+
+	// 4. Sign the intermediate CSR
```
cipherboy, line 107: We kinda discussed this yesterday, but if the mount is not provided, steps 4 and 5 here could be elided and a CSR returned to the user. In the case where the intermediate is signed successfully, we don't really need to return the intermediate cert as they're likely not doing anything with it.

command/pkicli/high_level_ops.go
```diff
@@ -0,0 +1,129 @@
...
+	signParams.put("_mount", rootMountPath)
+	signParams.put("csr", generateResp.csr)
+	signResp, err := p.signIntermediate(signParams)
+	if err != nil {
+		return nil, err
+	}
+
+	// 5. Set the signed certificate in the intermediate
+	setSignedParams := params.clone()
+	setSignedParams.put("certificate", signResp.certPem)
```
cipherboy, line 118: `certificate` here should probably be the concat of `signResp.certPem` and `root.chainPem`, though you don't fetch it here. This allows the chain to be constructed correctly; I think that'd have to use the new `ca_chain` field if it exists on the `/<root_mount>/certs/ca_chain` endpoint, otherwise fall back to `certificate` on the same endpoint.

command/pki.go
```diff
@@ -34,5 +35,49 @@ Usage: vault pki <subcommand> [options] [args]
 }
 
 func (c *PKICommand) Run(args []string) int {
+	//c.testCreateRoot()
```
cipherboy, line 38: I don't think this belongs here actually; this would belong in a new file and be added as a `pki add-intermediate` or `pki add-root` type behavior in the `command/commands.go` file.

command/pkicli/high_level_ops.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// 2. Generate the intermediate CA
+	generateParams := params.clone()
+	generateParams.put("_ca_type", "intermediate")
+	generateResp, err := p.generateCA(generateParams)
+	if err != nil {
+		return nil, err
+	}
+
+	// 3. Set the config URLs -- Is this necessary for intermediates??
```
cipherboy, line 101: Yes, I probably would configure this, though I'm not really sure _how_ to :-) In particular, we could point at Vault directly but I think we'd need to introspect `VAULT_ADDR` (or equivalent CLI options) and compute this path manually... Maybe we just do that from the client?

command/pkicli/high_level_ops.go
```diff
@@ -0,0 +1,129 @@
...
+	// The parameterMap accepts:
+	// - max_lease_ttl, to tune the backend
+	// - all the parameters for https://www.vaultproject.io/api/secret/pki#generate-intermediate
+	// - all the parameters for https://www.vaultproject.io/api-docs/secret/pki#sign-intermediate
+	// - all the parameters for https://www.vaultproject.io/api-docs/secret/pki#set-signed-intermediate
+	CreateIntermediate(signingMountPath, mountPath string, parameterMap mapStringAny) (*pkiCreateIntermediateResponse, error)
+
+}
+
+type pkiCreateRootParameters struct {
```
cipherboy, line 25: Looks unused? I think the `mapStringAny` approach is a little better here.

command/pkicli/low_level_ops.go
```diff
@@ -0,0 +1,134 @@
...
+	return err
+}
+
+func (p pkiOps) configUrls(params *Params) error {
+	// https://www.vaultproject.io/api/secret/pki#set-urls
+
+	if !params.hasAny("issuing_certificates", "crl_distribution_points", "ocsp_servers") {
+		return nil
+	}
+	params = params.clone()
```
cipherboy, line 116: Some API endpoints fail with extra parameters and others don't, I'm not really sure the rhyme or reason. Perhaps best to copy just the keys needed?

command/pki.go
```diff
@@ -34,5 +35,49 @@ Usage: vault pki <subcommand> [options] [args]
...
 }
+
+func (c *PKICommand) testCreateRoot() int {
+	client, err := c.Client()
+	if err != nil {
+		c.UI.Error(err.Error())
+	}
+	ops := pkicli.NewOperations(client)
+
+	vaultAddress := "http://localhost:8200"  // TODO: how to get the address?
```
cipherboy, line 51: Ah, I see your TODO now. I think `client.Address` would suffice?

command/pkicli/high_level_ops.go
cipherboy: See note below, it appears to be `client.Address`.

command/pki.go
victorr: Yes, this was just an easy place for me go invoke test code.

command/pkicli/low_level_ops.go
victorr: good point, hopefully we can use the endpoint API configuration to look up their parameters

command/pkicli/high_level_ops.go
victorr: Ok, I'll leave it to callers to add the parameters.

command/pkicli/high_level_ops.go
victorr: Pushed a changeset making these steps optional.

command/pkicli/high_level_ops.go
victorr: Are you referring to this API? https://www.vaultproject.io/api-docs/secret/pki#read-ca-certificate-chain

victorr: Done.

divyapola5 (approved):

PR #14400, opened 2022-03-07
UI/Hide empty masked PKI row values

hellobontempo:
After some changes made to the `InfoTableRow` component in PR #13585 empty masked values were displaying in the certificate show pages. This PR fixes so only masked inputs with values display

**Screeshot of bug**
image

**With fix**
Screen Shot 2022-03-07 at 2 50 33 PM

Screen Shot 2022-03-07 at 3 21 04 PM

Screen Shot 2022-03-07 at 3 23 35 PM

### DIFF
--- a/ui/app/templates/components/config-pki-ca.hbs
+++ b/ui/app/templates/components/config-pki-ca.hbs
@@ -16,7 +16,7 @@
       />
     {{/if}}
     {{#each this.model.attrs as |attr|}}
-      {{#if attr.options.masked}}
+      {{#if (and attr.options.masked (get this.model attr.name))}}
         <InfoTableRow
           data-test-table-row
           @label={{capitalize (or attr.options.label (humanize (dasherize attr.name)))}}
--- a/ui/app/templates/components/config-pki-ca.hbs
+++ b/ui/app/templates/components/config-pki-ca.hbs
@@ -8,7 +8,7 @@
     {{/if}}
   </h2>
   {{#if (or this.model.certificate this.model.csr)}}
-    {{#if (not (eq this.model.canParse true))}}
+    {{#if (eq this.model.canParse false)}}
       <AlertBanner
         @type="info"
         @message="There was an error parsing the certificate's metadata. As a result, Vault cannot display the issue and expiration dates. This will not interfere with the certificate's functionality."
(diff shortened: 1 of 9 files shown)


ui/app/templates/components/generate-credentials.hbs
```diff
@@ -75,12 +75,6 @@
             @label={{capitalize (or attr.options.label (humanize (dasherize attr.name)))}}
             @value={{date-format (get this.model attr.name) "MMM dd, yyyy hh:mm:ss a" isFormatted=true}}
           />
-        {{else if (and (get this.model attr.name) (eq attr.name "revocationTime"))}}
```
hellobontempo, line 78: `revocationTime` won't exist for a newly generated credential. (I had originally added this before understanding exactly when this template renders)

ui/app/templates/components/pki-cert-show.hbs
```diff
@@ -45,7 +45,7 @@
           @label={{capitalize (or attr.options.label (humanize (dasherize attr.name)))}}
           @value={{date-format (get this.model attr.name) "MMM dd, yyyy hh:mm:ss a" isFormatted=true}}
         />
-      {{else if (and (get this.model attr.name) (eq attr.name "revocationTime"))}}
```
hellobontempo, line 48: `(get this.model attr.name)` returns `0` so this block wasn't being hit and `Revocation time: 0` was displaying in the UI. Now zero is passed to the `date-format` helper which errors so revocation time isn't displayed unless it has a date value

ui/app/templates/components/config-pki-ca.hbs
```diff
@@ -8,15 +8,15 @@
     {{/if}}
   </h2>
   {{#if (or this.model.certificate this.model.csr)}}
-    {{#if (not (eq this.model.canParse true))}}
+    {{#if (eq this.model.canParse false)}}
```
hellobontempo, line 11: `canParse` is only set when it goes through the `parse-pki-certs` helper - so we want to check for `false` specifically, as it could be `undefined`

hashishaw (approved):
Great work, excellent test coverage, and thank you for the PR comments 👏

PR #14422, opened 2022-03-09
UI/add managed ns redirect prefix

hashishaw:
This PR updates the namespace redirect logic on managed vault clusters. 

Previous behavior caused a bug where if a namespace was provided in the URL but did not match the required format for managed clusters, it looked like you were logging into the `admin` namespace when in reality, the namespace passed to all requests is whatever is in the URL param. 

**UI Behavior when I try to log in with my token which is valid for the `admin` namespace:**
managed-login-error
Fails because request has the namespace which is in the URL param:
actual-request-ns

After this fix, whatever namespace passed is parsed and either appended to the root (`admin`) or kept as is, if the namespace param starts with `admin/`

**New Behavior**
managed-namespace-fix

### DIFF
--- a/ui/tests/acceptance/managed-namespace-test.js
+++ b/ui/tests/acceptance/managed-namespace-test.js
@@ -3,6 +3,7 @@ import { currentURL, visit, fillIn } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import Pretender from 'pretender';
 import logout from 'vault/tests/pages/logout';
+import { getManagedNamespace } from 'vault/routes/vault/cluster';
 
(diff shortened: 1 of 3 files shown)


ui/tests/acceptance/managed-namespace-test.js
```diff
@@ -50,4 +50,32 @@ module('Acceptance | Enterprise | Managed namespace root', function (hooks) {
       'Correctly prepends root to namespace'
     );
   });
+
+  test('getManagedNamespace helper works as expected', function (assert) {
```
Monkeychip, line 54: Thanks for adding the test coverage!

Monkeychip (approved):

PR #14424, opened 2022-03-09
Agent error log level is mismatched

swenson:
`logLevelToStringPtr` translates `go-hclog`'s `ERROR` to `"ERROR"` for
Consul Template's runner, but that expects `ERR` and is quite strict
about it.

This will address https://github.com/hashicorp/vault-k8s/issues/223
after it is set as the default image in `vault-k8s`.

I didn't find a simple way to test this other than starting up a full
server and agent and letting them run, which is unfortunately fairly
slow.

I confirmed that this addresses the original issue by modifying the helm
chart with the values in this commit and patching the log level to `err`.

### DIFF
--- a/changelog/14424.txt
+++ b/changelog/14424.txt
@@ -1,4 +1,3 @@
 ```release-note:bug
-COMPONENT: summary of chan
-agent: Log level mismatch between ERR and ERROR
+agent: Fix log level mismatch between ERR and ERROR
 ```
--- a/changelog/14424.txt
+++ b/changelog/14424.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+COMPONENT: summary of chan
+agent: Log level mismatch between ERR and ERROR
+```
--- a/command/agent/template/template_test.go
+++ b/command/agent/template/template_test.go
@@ -188,7 +188,7 @@ func TestCacheConfigNoListener(t *testing.T) {
 	assert.NotNil(t, ctConfig.Vault.Transport.CustomDialer)
 }
 
-func TestServerRun(t *testing.T) {
+func createHttpTestServer() *httptest.Server {
 	// create http test server
 	mux := http.NewServeMux()
 	mux.HandleFunc("/v1/kv/myapp/config", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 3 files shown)


changelog/14424.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
+COMPONENT: summary of chan
+agent: Log level mismatch between ERR and ERROR
```
kalafut, line 3:
```suggestion
agent: Fix log level mismatch between ERR and ERROR
```

tvoran (approved):
Worked great for me, thanks!

command/agent/template/template_test.go
```diff
@@ -400,6 +405,70 @@ func TestServerRun(t *testing.T) {
...
+			select {
+			case <-ctx.Done():
+				t.Fatal("timeout reached before templates were rendered")
+			case err := <-errCh:
+				if err != nil {
+					t.Fatalf("did not expect error, got: %v", err)
+				}
+			}
+			cancel()
+			server.Stop()
```
tomhjp, line 467: nit: Should these be called with `defer` further up so that we still clean up if we hit a `t.Fatal`?

tomhjp (approved):
LGTM!

command/agent/template/template_test.go
tomhjp: I think you're helped by the `t.Run` call, e.g. should be analogous to this?: https://go.dev/play/p/DmHsMW6Su96

calvn (approved):
Looks good, thanks for the fix!

command/agent/template/template_test.go
swenson: Ah, I think you're right! I'll commit this after re-running the tests.

command/agent/template/template_test.go
```diff
@@ -400,6 +405,70 @@ func TestServerRun(t *testing.T) {
 	}
 }
 
+// TestNewServerLogLevels tests that the server can be started with any log
+// level.
+func TestNewServerLogLevels(t *testing.T) {
+	ts := createHttpTestServer()
+	defer ts.Close()
+
+	tmpDir, err := ioutil.TempDir("", "agent-tests")
```
calvn, line 414: Since we're using Go 1.17+, it might be a good opportunity to swap this call to `os. MkdirTemp` in here and elsewhere in this file since that's what it really calls [under the hood](https://pkg.go.dev/io/ioutil#TempDir).

The doc mentions that it is preferred to call funcs from `os` over `ioutil`:
> As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.

swenson:
Thanks everyone!

PR #14426, opened 2022-03-09
identity/oidc: prevent key rotation on performance secondary clusters

austingebauer:
### Overview

This PR prevents the identity store's periodic func from running on performance secondary clusters. This fixes an issue where the key rotations were attempting to write to readonly storage.

Additionally, this adds some `continue` statements in a few places that I think are appropriate. This prevents some logging that didn't make sense (see delete logs below) and sets the `didUpdate` bool properly.

Example logs:
```
2022-03-08T15:09:41.745-0800 [DEBUG] secrets.identity.identity_2b8b62f2: rotating OIDC key: key=default
2022-03-08T15:09:41.862-0800 [WARN]  secrets.identity.identity_2b8b62f2: error rotating OIDC keys: err="cannot write to readonly storage"
2022-03-08T15:09:41.863-0800 [ERROR] secrets.identity.identity_2b8b62f2: error saving key: key="" error="cannot write to readonly storage"
2022-03-08T15:09:41.863-0800 [ERROR] secrets.identity.identity_2b8b62f2: error deleting OIDC public key: key_id=da41c8f4-0cac-f19c-16ad-b16f33a0d709 error="cannot write to readonly storage"
2022-03-08T15:09:41.863-0800 [DEBUG] secrets.identity.identity_2b8b62f2: deleted OIDC public key: key_id=da41c8f4-0cac-f19c-16ad-b16f33a0d709
```

### Testing

I manually tested this change using a performance secondary cluster to confirm that the key rotations in the periodic func are skipped on the active instance.

### DIFF
--- a/vault/identity_store_oidc.go
+++ b/vault/identity_store_oidc.go
@@ -24,6 +24,7 @@ import (
 	"github.com/hashicorp/vault/helper/identity"
 	"github.com/hashicorp/vault/helper/namespace"
 	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/helper/consts"
 	"github.com/hashicorp/vault/sdk/helper/identitytpl"
 	"github.com/hashicorp/vault/sdk/logical"
 	"github.com/patrickmn/go-cache"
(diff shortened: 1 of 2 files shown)


fairclothjm (approved):

vault/identity_store_oidc.go
```diff
@@ -1791,6 +1794,7 @@ func (i *IdentityStore) expireOIDCPublicKeys(ctx context.Context, s logical.Stor
 			if err := s.Delete(ctx, publicKeysConfigPath+keyID); err != nil {
 				i.Logger().Error("error deleting OIDC public key", "key_id", keyID, "error", err)
 				nextExpiration = now
+				continue
```
calvn, line 1797: This makes me think whether we should move L1793 (`didUpdate = true`) to be called on L1799 instead. This way we only flush the cache on successful deletion of the key(s).

austingebauer: I was just thinking to do this too 👍  Done in 8da0631.

calvn (approved):

PR #14435, opened 2022-03-10
Backport of UI/add managed ns redirect prefix into release/1.9.x

hc-github-team-secure-vault-core:
## Backport

This PR is auto-generated from #14422 to be assessed for backporting due to the inclusion of the label backport/1.9.x.

---

This PR updates the namespace redirect logic on managed vault clusters. 

Previous behavior caused a bug where if a namespace was provided in the URL but did not match the required format for managed clusters, it looked like you were logging into the `admin` namespace when in reality, the namespace passed to all requests is whatever is in the URL param. 

**UI Behavior when I try to log in with my token which is valid for the `admin` namespace:**
managed-login-error
Fails because request has the namespace which is in the URL param:
actual-request-ns

After this fix, whatever namespace passed is parsed and either appended to the root (`admin`) or kept as is, if the namespace param starts with `admin/`

**New Behavior**
managed-namespace-fix


hashishaw:
Automatic cherry-pick failed, closing this in favor of #14440

PR #14438, opened 2022-03-10
Command code for adding pki-intermediate

divyapola5:
Command code to add intermediate CA

### DIFF
--- a/command/pki_add_intermediate.go
+++ b/command/pki_add_intermediate.go
@@ -31,14 +31,11 @@ Usage: vault pki add-intermediate [options] [ROOT_MOUNT] PATH COMMON_NAME [K=V]
 
   Configures an intermediate mount and generate the intermediate certificate.
   The intermediate certificate is the one from which all leaf certificates will be generated.
-  This intermediate will be signed by the root. Other parameters can be specified in
-  K=V format (mirroring vault write).
+  This intermediate will be signed by the root if root-mount specified in the input parameters.
+  Other parameters can be specified in K=V format (mirroring vault write).
 
   Configure an intermediate mount at path pki-int with a specific ttl:
       $ vault pki add-intermediate pki pki-int example.com ttl=48000h
-
-  Configure an intermediate mount at path pki-int with CSR type and common name:
-      $ vault pki add-intermediate pki pki-int example.com csr=@example.csr format=pem
   
 
 ` + c.Flags().Help()
--- a/command/pki_add_intermediate.go
+++ b/command/pki_add_intermediate.go
@@ -16,9 +16,7 @@ var (
 type PKIAddIntermediateCommand struct {
 	*BaseCommand
 
-	flagMount string
 	flagRootMount string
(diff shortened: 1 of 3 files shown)


command/pki_add_intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+	flagCommonName string
+}
+
+func (c *PKIAddIntermediateCommand) Synopsis() string {
+	return "Generate intermediate certificate"
+}
+
+func (c *PKIAddIntermediateCommand) Help() string {
+	helpText := `
+Usage: vault pki add-intermediate [options] [ROOT_MOUNT] PATH COMMON_NAME [K=V]
```
cipherboy, line 30: root mount is an option I think from the flags below? :)

command/pki_add_intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+	return strings.TrimSpace(helpText)
+}
+
+func (c *PKIAddIntermediateCommand) Flags() *FlagSets {
+	set := c.flagSet(FlagSetHTTP)
+	f := set.NewFlagSet("Command Options")
+
+	f.StringVar(&StringVar{
+		Name:    "mount",
+		Target:  &c.flagMount,
```
cipherboy, line 52: I don't think this one is necessary? That's what the first positional argument should be, correct? `args[1]` in particular.

command/pki_add_intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+	f.StringVar(&StringVar{
+		Name:    "root-mount",
+		Target:  &c.flagRootMount,
+		Default: "pki",
+		EnvVar:  "",
+		Usage:   "PKI root mount",
+	})
+
+	f.StringVar(&StringVar{
+		Name:    "common_name",
```
cipherboy, line 67: Same as above `flagMount` above. Only `root-mount` is necessary.

command/pki_add_intermediate.go
```diff
@@ -0,0 +1,139 @@
...
+		return 1
+	}
+
+	args = f.Args()
+	if len(args) < 3 {
+		c.UI.Error(fmt.Sprintf("Not enough arguments (expected 3+, got %d)", len(args)))
+		return 1
+	}
+
+	rootMountPath := sanitizePath(args[0]) // TODO figure out how to check for optional fields
```
cipherboy, line 101: Ah I see now. You can remove this and just use `c.flagRootMount` everywhere below and it'll be handled for you, when you call `f.Args()` on line 95 above. :-) 

And then you'll need to update the rest of your `args` usage by one.

cipherboy (commented):
Mostly info about how to use optional flags, feel free to ignore if you don't want to :-)

command/pki_add_intermediate.go
divyapola5: Yup, updated the code now. Thanks.

cipherboy (approved):

PR #14462, opened 2022-03-11
Add vault pki role-update subcommand

cipherboy:
Writing over an existing role (`vault write /pki/roles/example-com ...`)
results in a loss of data unless all previously specified options are
re-specified. This means there's no update capability. Add a new
subcommand, `vault pki role-update /pki/roles/example-com ...` to
handle fetching existing role data and augmenting it with new options.


victorr (approved):
👍

PR #14471, opened 2022-03-11
UI/fix flaky clients current test

hellobontempo:
Adds `waitUntil` to fix flaky test

Screen Shot 2022-03-11 at 12 09 50 PM

### DIFF
--- a/ui/tests/acceptance/client-current-test.js
+++ b/ui/tests/acceptance/client-current-test.js
@@ -1,5 +1,5 @@
 import { module, test } from 'qunit';
-import { visit, currentURL, settled, click } from '@ember/test-helpers';
+import { visit, currentURL, settled, click, waitUntil, find } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import Pretender from 'pretender';
 import authPage from 'vault/tests/pages/auth';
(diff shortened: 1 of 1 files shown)


hashishaw (approved):
👏

hashishaw:
Just one note: why does this need a backport?

hellobontempo:
> Just one note: why does this need a backport?

Oh - I thought to mitigate flakiness in any PR that goes to 1.10? But am I thinking of that wrong?

hashishaw:
> Oh - I thought to mitigate flakiness in any PR that goes to 1.10? But am I thinking of that wrong?

Wrong is a strong word 😄  Our general approach to backports is they're for fixes, and test updates/developer happiness isn't typically considered a fix. I think this a place where we should use discretion, and if the test is failing VERY often before this fix then maybe this is a good exception to our rule!

hellobontempo:
> > Oh - I thought to mitigate flakiness in any PR that goes to 1.10? But am I thinking of that wrong?
> 
> Wrong is a strong word 😄 Our general approach to backports is they're for fixes, and test updates/developer happiness isn't typically considered a fix. I think this a place where we should use discretion, and if the test is failing VERY often before this fix then maybe this is a good exception to our rule!

Aha! Thank you. Now I'm remembering that we only backport fixes. _Completely_ spaced on this and was just in backport mode. Thanks for clarifying!

PR #14474, opened 2022-03-12
Add fields 'ttl' and 'num_uses' to SecretID generation.

### LINKED ISSUE #14390: Customizing num_uses and ttl per approle secret id
When I generate a new secret ID for an approle, I would like to customize the TTL and/or num_uses for that secret ID, similarly to how I can customize the [cidr_list](https://www.vaultproject.io/api-docs/auth/approle#cidr_list) and [token_bound_cidrs](https://www.vaultproject.io/api-docs/auth/approle#token_bound_cidrs-1). 

Additional API parameters for the [approle secret ID creation](https://www.vaultproject.io/api-docs/auth/approle#generate-new-secret-id) allowing custom values to be set for `num_uses` and `ttl`.

Since `num_uses` and `ttl` are currently tied to the role configuration, any use case for an approle that requires a different value for either parameter than the default means I have to create an entirely new approle with identical policies but different configuration. This is possible, but tedious.

In particular, this can be useful for the recommended [approle use pattern that involves wrapping](https://learn.hashicorp.com/tutorials/vault/pattern-approle?in=vault/recommended-patterns). For example, it would be very convenient to specify the `num_uses` for a new secret ID being passed to automation, even if the approle configuration did not have `num_uses` set.

RemcoBuddelmeijer:
Add the fields 'ttl' and 'num_uses' to the endpoints `/auth/approle/role/<role_name>/secret-id` and `/auth/approle/role/<role_name>/custom-secret-id`.
Edited the help synopsis and description to refer to these fields. Both for the endpoints respectively but also for the `secret-id-num-uses` and `secret-id-ttl` endpoints.

Fixes #14390


swayne275:
thanks! it appears we have some [formatting issues](https://app.circleci.com/pipelines/github/hashicorp/vault/29149/workflows/a6378d19-dd61-4796-b9ee-00e620c431db/jobs/404062)

swayne275:
sure, give it a shot and we can take a deeper look

RemcoBuddelmeijer:
The test seem to be passing now 😄

kent007:
Hi everyone,

Just want to keep the conversation moving on this feature. I see Vault 1.10 just rolled out, any idea when these changes might be generally released?

builtin/credential/approle/path_role.go
```diff
@@ -2355,9 +2375,28 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
 		return nil, err
 	}
 
+	var numUses int
+	// Check whether or not num_uses is defined, otherwise fallback to secret_id_num_uses
+	if numUsesRaw, ok := data.GetOk("num_uses"); ok {
+		if numUses < 0 {
```
hghaf099, line 2381: Please move this check to line 2388, as this is not doing much here.

builtin/credential/approle/path_role.go
```diff
@@ -2376,6 +2415,7 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
 			"secret_id":          secretID,
 			"secret_id_accessor": secretIDStorage.SecretIDAccessor,
 			"secret_id_ttl":      int64(b.deriveSecretIDTTL(secretIDStorage.SecretIDTTL).Seconds()),
+			"secret_id_num_uses": int(secretIDStorage.SecretIDNumUses),
```
hghaf099, line 2418: Is the int(...) casting necessary?

hghaf099, line 2418: Is the int(...) casting necessary?

builtin/credential/approle/path_role_test.go
```diff
@@ -1152,6 +1152,77 @@ func TestAppRole_RoleSecretID(t *testing.T) {
...
+	}
+
+	if resp.Data["secret_id"].(string) == "" {
+		t.Fatalf("failed to generate secret_id")
+	}
+	if resp.Data["secret_id_ttl"].(int64) != 5 {
+		t.Fatalf("secret_id_ttl has not been set by the 'ttl' field")
+	}
+	if resp.Data["secret_id_num_uses"].(int) != 5 {
+		t.Fatalf("secret_id_ttl has not been set by the 'num_uses' field")
```
hghaf099, line 1202:
```suggestion
		t.Fatalf("secret_id_num_uses has not been set by the 'num_uses' field")
```

hghaf099, line 1202:
```suggestion
		t.Fatalf("secret_id_ttl has not been set by the 'num_uses' field")
```
```suggestion
		t.Fatalf("secret_id_num_uses has not been set by the 'num_uses' field")
```

builtin/credential/approle/path_role_test.go
```diff
@@ -1152,6 +1152,77 @@ func TestAppRole_RoleSecretID(t *testing.T) {
...
+	}
+
+	if resp.Data["secret_id"] != "abcd123" {
+		t.Fatalf("failed to set specific secret_id to role")
+	}
+	if resp.Data["secret_id_ttl"].(int64) == 0 {
+		t.Fatalf("secret_id_ttl has not been set by the 'ttl' field")
+	}
+	if resp.Data["secret_id_num_uses"].(int) != 5 {
+		t.Fatalf("secret_id_ttl has not been set by the 'num_uses' field")
```
hghaf099, line 1222:
```suggestion
		t.Fatalf("secret_id_num_uses has not been set by the 'num_uses' field")
```

hghaf099, line 1222:
```suggestion
		t.Fatalf("secret_id_num_uses has not been set by the 'num_uses' field")
```

changelog/14474.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+auth/approle: add 'ttl' and 'num_uses' fields when generating SecretIDs
```
hghaf099, line 2: would you please be more elaborative here? Like to indicate that the two fields are per secretID, yet if they are not configured, the role specific values are used.

hghaf099, line 2: would you please be more elaborative here? Like to indicate that the two fields are per secretID, yet if they are not configured, the role specific values are used.

website/content/api-docs/auth/approle.mdx
```diff
@@ -501,12 +512,21 @@ Assigns a "custom" SecretID against an existing AppRole. This is used in the
...
+  after which this SecretID expires. Defaults to 0, meaning no expiration.
+  Overrides secret_id_ttl role option when supplied.
 
 ### Sample Payload
 
 ```json
 {
-  "secret_id": "testsecretid"
+  "secret_id": "testsecretid",
+  "metadata": "{ \"tag1\": \"production\" }",
```
hghaf099, line 527: can we keep the changes to what has actually changed please?

hghaf099, line 527: can we keep the changes to what has actually changed please?

builtin/credential/approle/path_role.go
```diff
@@ -2355,9 +2375,28 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
...
 
+	var numUses int
+	// Check whether or not num_uses is defined, otherwise fallback to secret_id_num_uses
+	if numUsesRaw, ok := data.GetOk("num_uses"); ok {
+		if numUses < 0 {
+			return logical.ErrorResponse("num_uses cannot be negative"), nil
+		}
+		numUses = numUsesRaw.(int)
+	} else {
+		numUses = role.SecretIDNumUses
```
hghaf099, line 2386: After some internal discussions, we have decided that we would like to bound the per SecretID num_uses and ttl by the corresponding values in the Role parameters. Like, if we set the num_uses for a secretID to 200 and the role. SecretIDNumUses is 100, the value for the role should be used. And the same for the ttl.

builtin/credential/approle/path_role.go
RemcoBuddelmeijer: You are correct, this is not necessary. Removed :)

builtin/credential/approle/path_role.go
RemcoBuddelmeijer: I have moved it to line 2382 instead. This check is only to validate the supplied field, not the role's configuration. Moved it instead of creating a second duplicate variable on the stack from `numUsesRaw`.

builtin/credential/approle/path_role.go
RemcoBuddelmeijer: Wouldn't this defeat the purpose of the feature request that was filed? Here it is specifically requested to separate the two and in particular, be able to override the role's configuration.
Without the possibility to override the role's configuration you would still have to create a separate role.

changelog/14474.txt
RemcoBuddelmeijer: I'll change this once the rest of the conversations are resolved :)

website/content/api-docs/auth/approle.mdx
RemcoBuddelmeijer: I didn't think too much of this, but changed and pushed!

builtin/credential/approle/path_role.go
kent007: This seems consistent with the existing rules for CIDR block restrictions...if the role value is set, then the per secret-id values must be a subset. Likewise, if the role value is _not_ set, then the per secret-id values are unbounded. I think this would still meet the original feature request, since the user can choose to set maximum values as part of the role config (or leave them unset and allow any value per-secret ID)

RemcoBuddelmeijer: > The override is still possible though, the only constraint is that the new secretID ttl/num_uses could not be more than the corresponding ones in the role config. So, this way, the Role should be configured with highest required values for those parameters, and this way, the secretIDs could be configured to have different values but in a defined range.

Okay this makes more sense to me. While I still would disagree with a couple of points, it looks like the original requestee has given his stamp of approval.
I'll think of a solution that would fit best for the situation.

RemcoBuddelmeijer: @hghaf099 @kent007 After a short consiledation, I came to the conclusion to simply lower bind the SecretID's ttl and num_uses to the Role's ttl and num_uses. This yields to a PR without any changes outside of the scope on the Role's configuration, possibly requiring users to change the approach they take when creating a Role.
For the best result possible I'll go for a change where validation is performed rather than implicitly overriding the SecretId's defined ttl/num_uses with the Role's ttl/num_uses.

kent007: Hi @RemcoBuddelmeijer, I'm wondering what else remains to be done for this implementation?  It sounds like we just need a few additional checks on the values presented at secret-id creation time...I have some time available this month to finish it  up if need be

RemcoBuddelmeijer: Hey @kent007 I was unavailable the past days but I have the weekend planned to finish up the implementation!

RemcoBuddelmeijer: Intermediate update. My days have been packed far more than expected. Right now I am out of the country and expecting to be back next weekend.

RemcoBuddelmeijer: Updated the branch with the new requirements. Updated the upstream and merged main back into this branch. Will change changelog. Please feel free to review!

builtin/credential/approle/path_role.go
```diff
@@ -2390,6 +2395,10 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
 	// Check whether or not ttl is defined, otherwise fallback to secret_id_ttl
 	if ttlRaw, ok := data.GetOk("ttl"); ok {
 		ttl = time.Second * time.Duration(ttlRaw.(int))
+		// If the ttl is less than the role's secret_id_ttl, throw an error rather than implicitly overriding
+		if ttl < role.SecretIDTTL {
+			return logical.ErrorResponse("ttl cannot be shorter than the role's secret_id_ttl"), nil
```
kent007, line 2400: Wouldn't this scenario make sense though? If the default TTL on the role was 90 days, and the client requested a secret-id with a TTL of 5 minutes for some process, then the shorter TTL should win. Just like num_uses, the role config would serve as the upper bound?

RemcoBuddelmeijer: You are 100% correct. I’ve gone ahead and change it.

RemcoBuddelmeijer:
Hello all! I was hoping to bump up this PR to get this looked at and possibly  merged. Thank you! 😄

kent007:
Hi everyone, also bumping this PR. Would love to see this feature merged.

builtin/credential/approle/path_role_test.go
```diff
@@ -1152,6 +1151,234 @@ func TestAppRole_RoleSecretID(t *testing.T) {
...
+	}
+
+	resp, err = b.HandleRequest(context.Background(), roleReq)
+	if err != nil || (resp != nil && resp.IsError()) {
+		t.Fatalf("err:%v resp:%#v", err, resp)
+	}
+
+	testCases := []testCase{
+		{
+			name:    "finit num_uses ttl",
```
raskchanky, line 1185:
```suggestion
			name:    "finite num_uses ttl",
```

builtin/credential/approle/path_role.go
```diff
@@ -2355,9 +2375,38 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
...
+	// Check whether or not num_uses is defined, otherwise fallback to secret_id_num_uses
+	if numUsesRaw, ok := data.GetOk("num_uses"); ok {
+		numUses = numUsesRaw.(int)
+		if numUses < 0 {
+			return logical.ErrorResponse("num_uses cannot be negative"), nil
+		}
+
+		// If the role's secret_id_num_uses is higher than the specified num_uses, throw an error rather than implicitly overriding
+		if (numUses == 0 && role.SecretIDNumUses > 0) || (role.SecretIDNumUses > 0 && numUses > role.SecretIDNumUses) {
+			return logical.ErrorResponse("num_uses cannot be higher than the role's secret_id_num_uses"), nil
```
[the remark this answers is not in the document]
RemcoBuddelmeijer: > This error message confuses me a little, specifically for the first case. If numUses == 0 && role.SecretIDNumUses > 0 then num_uses isn't higher than the role's secret_id_num_uses, is it? It's lower.

Whenever either num_uses or ttl is set to zero it equals to [unlimited/infinite](https://www.vaultproject.io/api-docs/auth/approle#:~:text=A%20value%20of%20zero%20will%20allow%20unlimited%20uses.). This guard statement's first part ensures that no unlimited custom num_uses can be used whenever the role's secret id configuration uses a limited value. The second part ensures the normal flow, as long as the role's secret id configuration does not use an unlimited num_uses value.
The same applies to the ttl parameter.

> Additionally, wouldn't this get triggered by default, e.g. if num_uses isn't set (because it would default to 0)?

No this wouldn't get triggered by default. The `data#GetOk` method checks if a parameter was specified, if not, it defaults to the Role's Secret ID num_uses. This was done to not break the current implementation. The same applies to the ttl parameter.

builtin/credential/approle/path_role.go
```diff
@@ -2355,9 +2375,38 @@ func (b *backend) handleRoleSecretIDCommon(ctx context.Context, req *logical.Req
...
+		}
+	} else {
+		numUses = role.SecretIDNumUses
+	}
+
+	var ttl time.Duration
+	// Check whether or not ttl is defined, otherwise fallback to secret_id_ttl
+	if ttlRaw, ok := data.GetOk("ttl"); ok {
+		ttl = time.Second * time.Duration(ttlRaw.(int))
```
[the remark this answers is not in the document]
RemcoBuddelmeijer: This piece of code was duplicated from the current approach taken to calculate the Role's Secret ID's ttl: https://github.com/hashicorp/vault/blob/22346ba24980073f8243aa5a1ee35dc18427da49/builtin/credential/approle/path_role.go#L966

Seeing that this is no longer valid, do we want to change this behaviour as well? The code was duplicated only with the intention to not change the way TTL is calculated for the SecretID.

builtin/credential/approle/path_role_test.go
```diff
@@ -1141,7 +1141,6 @@ func TestAppRole_RoleSecretID(t *testing.T) {
 		"secret_id": "abcd123",
 	}
 	roleSecretIDReq.Data = roleCustomSecretIDData
-	roleSecretIDReq.Operation = logical.UpdateOperation
```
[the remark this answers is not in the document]
RemcoBuddelmeijer: This was removed as the operation never changes for this request. It's not directly related to the PR, but more so a cleanup so it can be rollbacked if it's out of scope.

builtin/credential/approle/path_role.go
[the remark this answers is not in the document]
raskchanky: Ah, I'm following you now, thanks. I do think the comment might be incorrect though, as it states that an error should be thrown if the role's secret_id_num_uses is higher than the specified num_uses, but the code is opposite of that (and I think in this case the code is correct).

builtin/credential/approle/path_role.go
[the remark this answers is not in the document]
raskchanky: We don't want to change the behavior of the existing API, at least not in this PR. It's odd that the implementation you point out doesn't match the docs (which says it takes either a duration string or an integer). Regardless, since this is a new API, we should use `parseutil.ParseDurationSecond()`, since that will handle both duration strings and integers transparently.

builtin/credential/approle/path_role_test.go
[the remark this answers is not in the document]
raskchanky: Ah, I see what you mean now

builtin/credential/approle/path_role.go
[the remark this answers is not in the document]
RemcoBuddelmeijer: I would argue that this isn't a new API. We are currently modifying an already existing API endpoint with additional new fields. It would be odd to change this now as it defaults to the Role's Secret ID Configuration, which is calculated using the invalid approach.
How about we immediatelly create a new issue and link it after this PR?

builtin/credential/approle/path_role.go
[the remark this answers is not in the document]
RemcoBuddelmeijer: You are correct. I will port this together while rephrasing. This was already part of the plan as it wasn't sure yet before what the approach would be. 😄

builtin/credential/approle/path_role.go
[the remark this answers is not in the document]
raskchanky: You're right, I misspoke. This is not a new API. It's a new field that we're adding to an existing API. I'm arguing that since it's a new field, the value we extract from it should be parsed using the more flexible approach, that allows for either integers or duration strings. The fact that the existing TTL field doesn't do that is a bug, as far as I can tell, especially since the documentation indicates that it should. It doesn't make sense to me to merge this PR with a bug in it, when we can correct it before it's merged.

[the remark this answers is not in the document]
RemcoBuddelmeijer: While I agree, now we could have a potential mismatch in the way that TTL is parsed and returned, just to point that out. Since it's obviously not a good thing to leave bugs in, I will create an issue for the existing TTL fields. In the mean time I will patch the behaviour of the new field and we can go from there!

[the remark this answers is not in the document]
raskchanky: > This is a major change. We should be careful about this.

I disagree that it's a major change, since I'm not advocating for changing the existing behavior of anything. Everything that existed before this PR will continue to work the same way it always has. All we're doing is determining how to parse a brand new field that no one has used before.

builtin/credential/approle/path_role.go
```diff
@@ -2514,14 +2564,17 @@ defined on the role, can access the role.`,
...
-		`Duration in seconds, representing the lifetime of the SecretIDs
-that are generated against the role using 'role/<role_name>/secret-id' or
-'role/<role_name>/custom-secret-id' endpoints.`,
-		``,
+		"Duration in seconds of the SecretID generated against the role.",
+		`If the SecretIDs are generated/assigned against the role using the
+'role/<role_name>/secret-id' or 'role/<role_name>/custom-secret-id' endpoints,
+then the lifetime of the SecretID is by default defined by this option.
+Can be overriden by the 'ttl' field on either of the two endpoints,
+but is still the longest possible ttl.`,
```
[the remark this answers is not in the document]
RemcoBuddelmeijer: I will remove the ", but" clause and only specify that this option may be overriden by the ttl field

builtin/credential/approle/path_role.go
```diff
@@ -2514,14 +2564,17 @@ defined on the role, can access the role.`,
 		"Use limit of the SecretID generated against the role.",
 		`If the SecretIDs are generated/assigned against the role using the
 'role/<role_name>/secret-id' or 'role/<role_name>/custom-secret-id' endpoints,
-then the number of times that SecretID can access the role is defined by
-this option.`,
+then the number of times that SecretID can access the role is by default defined by this option.
+Can be overriden by the 'num_uses' field on either of the two endpoints,
+but is still the highest possible num_uses.`,
```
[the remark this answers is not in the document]
RemcoBuddelmeijer: I will remove the ", but" clause and only specify that this option may be overriden by the num_uses field. Also did some more rephrasing.

RemcoBuddelmeijer:
Hello everyone, periodic bump up of the PR to see if it can get merged. Thank you!

HridoyRoy (approved):
Lgtm

PR #14488, opened 2022-03-14
oss port: semgrep fixes

HridoyRoy:
This is the port of the semgrep fixes here: https://github.com/hashicorp/vault-enterprise/pull/2637/files


swayne275 (approved):
I think I remember these! I know it's a port (and this is a nit), but the triple nested ifs starting around line 178 - could those be unwound a bit?

HridoyRoy:
> I think I remember these! I know it's a port (and this is a nit), but the triple nested ifs starting around line 178 - could those be unwound a bit?

Having some trouble unwinding and checking semgrep on enterprise. I'm going to merge this PR so semgrep on ent is fixed, and then I'll open a new PR porting semgrep over to OSS and add the unwinding in the new PR, so the semgrep tests can run on the change in realtime.

PR #14489, opened 2022-03-14
UI/fix kv data cache

hashishaw:
This PR fixes some strange behavior within the KV secrets engine (see gifs below):

User `reader` has the following policy:
```
path "kv/metadata" {
  capabilities = ["list"]
}
path "kv/data/*" {
  capabilities = ["read", "update"]
}
```

User `admin` has the following policy:
```
path "kv/metadata" {
  capabilities = ["list"]
}
path "kv/data/*" {
  capabilities = ["read", "update"]
}
path "kv/metadata/*" {
  capabilities = ["read", "update"]
}
```
**BEFORE: reader cannot click button create new version, but experiences no issues when creating a version other than the new version doesn't show after redirect to show view**
kv-new-version-before

**AFTER: admin can still create new version from specific version (this behavior is preserved, if user has metadata read access)**
kv-new-version-after

**AFTER: reader can create new version, and it always creates from the most recent version due to no metadata read access**
kv-new-version-metadata

### DIFF
--- a/ui/app/templates/components/secret-edit-toolbar.hbs
+++ b/ui/app/templates/components/secret-edit-toolbar.hbs
@@ -92,12 +92,15 @@
       {{#let (concat "vault.cluster.secrets.backend." (if (eq @mode "show") "edit" "show")) as |targetRoute|}}
         {{#if @isV2}}
           <ToolbarLink
-            @params={{array targetRoute @model.id (query-params version=@modelForData.version)}}
+            {{! Always create new version from latest if no metadata read access }}
+            @params={{array
+              targetRoute
+              @model.id
+              (query-params version=(if @model.canReadMetadata @modelForData.version ""))
+            }}
             @data-test-secret-edit="true"
             @replace={{true}}
             @type="add"
-            @disabled={{@model.failedServerRead}}
-            @disabledTooltip="Metadata read access is required to create new version"
           >
             Create new version
           </ToolbarLink>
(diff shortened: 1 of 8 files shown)


ui/app/templates/components/secret-edit-toolbar.hbs
```diff
@@ -92,12 +92,15 @@
       {{#let (concat "vault.cluster.secrets.backend." (if (eq @mode "show") "edit" "show")) as |targetRoute|}}
         {{#if @isV2}}
           <ToolbarLink
-            @params={{array targetRoute @model.id (query-params version=@modelForData.version)}}
+            {{! Always create new version from latest if no metadata read access }}
```
Monkeychip, line 95: Is this case even if you're previewing an older version of the secret?

hashishaw: Right now through the UI there's not a way to view an older version if you don't have metadata read access, but if there's a use case there then we can calculate whether to include the version based on whether it's on the query param or if they can read metadata

ui/tests/acceptance/secrets/backend/kv/secret-test.js
```diff
@@ -683,8 +694,9 @@ module('Acceptance | secrets/secret/create', function (hooks) {
     await settled(); // eslint-disable-line
     await click('[data-test-secret-tab]');
     await settled(); // eslint-disable-line
-    let text = document.querySelector('[data-test-empty-state-title]').innerText.trim();
```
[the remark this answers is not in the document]
Catclaws12: - [ ] 🙀

ui/app/templates/components/secret-edit-toolbar.hbs
Monkeychip: Got it!

Monkeychip (approved):
Thanks for tackling this!

PR #14490, opened 2022-03-14
Update more PKI documentation

cipherboy:
Various changes here; would definitely like feedback on if these warrant a backport to 1.10 for clarity.

 - Serial number field is, IMO, misnamed and definitely was misdocumented. It does not impact what most users expect it to (namely, the actual serial number field of the X509 certificate object), but instead adds a `serialNumber=<value>` attribute _into the certificate's **Subject**_. I've clarified it in both the website docs and the path-help docs.
 - `key_bits` needed to be more consistently documented. 
 - Clarify documentation around `GET pki/config/crl` and `GET pki/config/urls` being empty when nothing is written yet.
 - Clarify that `pki/config/crl/rotate` should be called when the CRL is expired or close to expiring, if no revocations have happened.
 - Fix the path-help documentation of the fetch APIs, which referenced the `fetch/` path segment that no longer exists. 

---

Not addressed in this PR (and definitely not a candidate for 1.10 backport) is that `pki/intermediate/generate/:type` returns help text with `ttl` and `not_after` fields, but these fields cannot go into the CSR and thus shouldn't be exposed. However, TTL (and not `not_after`, which is inconsistent with the rest of the API) is used by `getGenerationParams` to build a role for generation (and is also used by `pki/root/generate/:type`). That requires some more thought to fix.


stevendpclark (approved):

cipherboy:
Thanks Steve!

PR #14493, opened 2022-03-15
UI/d3 DOM cleanup hover issue

hellobontempo:
**This hover bug was happening because not all of the D3 elements were removed from the page when the chart updated**
chart-hover-bug

**cleaned up HTML into groups with fix**
cleanup-chart-html

### DIFF
--- a/ui/tests/acceptance/client-history-test.js
+++ b/ui/tests/acceptance/client-history-test.js
@@ -337,7 +337,7 @@ module('Acceptance | clients history tab', function (hooks) {
     assert.equal(currentURL(), '/vault/clients/history', 'clients/history URL is correct');
     assert
       .dom(SELECTORS.emptyStateTitle)
-      .includesText('No start date found', 'Empty state shows no billing start date');
+      .includesText('start date found', 'Empty state shows no billing start date');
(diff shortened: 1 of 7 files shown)


ui/tests/acceptance/client-history-test.js
```diff
@@ -292,7 +337,7 @@ module('Acceptance | clients history tab', function (hooks) {
     assert.equal(currentURL(), '/vault/clients/history', 'clients/history URL is correct');
     assert
       .dom(SELECTORS.emptyStateTitle)
-      .includesText('No start date found', 'Empty state shows no billing start date');
+      .includesText('start date found', 'Empty state shows no billing start date');
```
hellobontempo, line 340: for enterprise this reads "No billing start date found" and for OSS "No start date found" 

this was previously passing, but started failing for some reason? so only checking for the words shared between both versions

hashishaw: Are you running Vault enterprise locally with the tests (assuming it's failing locally?). The tests run on whatever version of vault you have loaded so that could be why

hashishaw (approved):
Nice work, thanks for finding and fixing!

PR #14495, opened 2022-03-15
Introduce fips version-agnostic build tag

cipherboy:
Unlike `fips_140_3`, `fips` will be a (FIPS) version-agnostic build tag.
The listener support will remain in 140-3 only, but the `IsFIPS()` check
should apply regardless of FIPS version.

We add two FIPS-only build files which validate the constraints of FIPS
builds here: `fips` must be specified with either `fips_140_2` or `fips_140_3`
build tags, and `fips` and `cgo` must also be specified together.
Additionally, using only a version-specific FIPS build tag without the
version-agnostic FIPS tag should be a failure.

### DIFF
--- a/helper/constants/fips.go
+++ b/helper/constants/fips.go
@@ -1,4 +1,4 @@
-//go:build !fips_140_3
+//go:build !fips
 
 package constants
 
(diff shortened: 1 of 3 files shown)


stevendpclark (approved):

cipherboy:
Thanks for the review and comments @stevendpclark :)

PR #14506, opened 2022-03-15
Revert the OpenAPI reference objects change #14217

averche:
The change in the structure of OpenAPI spec to add reference objects (#14217) caused portions of the UI to break. Reverting this change for now while we investigate & fix the issue.


VinnyHC (approved):
This looks good to me, not sure what the official "revert" process is WRT changelog and such.

averche:
The UI issue has been fixed in #14508, closing this PR

PR #14507, opened 2022-03-15
Handle the empty mount accessor case in client counts

ncabatoff:


vault/activity_log.go
```diff
@@ -2099,11 +2099,15 @@ func (a *ActivityLog) precomputedQueryWorker(ctx context.Context) error {
...
 				for mountAccessor, mountData := range nsMap[nsID].Mounts {
 					var displayPath string
-					valResp := a.core.router.ValidateMountByAccessor(mountAccessor)
-					if valResp == nil {
-						displayPath = fmt.Sprintf("deleted mount; accessor %q", mountAccessor)
+					if mountAccessor == "" {
+						displayPath = "no mount accessor (pre-1.10 upgrade?)"
 					} else {
-						displayPath = valResp.MountPath
+						valResp := a.core.router.ValidateMountByAccessor(mountAccessor)
```
hghaf099, line 2105: The printed message is like `"deleted mount; accessor \"\""` before this PR. Based on that result, it seems that the mountAccessor is `\`. If this is the case, I am not sure if the code changes the behaviour I am seeing. Is this intentional?

PR #14508, opened 2022-03-15
UI: Parse OpenAPI response correctly if schema includes $ref

hashishaw:
This PR adds UI support for the changes in #14217 

Before this change, the UI was broken where we try to generate the form based on OpenAPI:
image

After, the form is rendered as expected:
image

### DIFF
--- a/ui/app/services/path-help.js
+++ b/ui/app/services/path-help.js
@@ -207,7 +207,7 @@ export default Service.extend({
       let props = {};
       const schema = pathInfo?.post?.requestBody?.content['application/json'].schema;
       if (schema.$ref) {
-        // $ref will be shaped like `#/component/schema/MyResponseType
+        // $ref will be shaped like `#/components/schemas/MyResponseType
         // which maps to the location of the item within the openApi response
         let loc = schema.$ref.replace('#/', '').split('/');
         props = loc.reduce((prev, curr) => {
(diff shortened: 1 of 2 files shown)


ui/app/services/path-help.js
```diff
@@ -204,10 +204,18 @@ export default Service.extend({
...
       }
 
-      // TODO: handle post endpoints without requestBody
-      const props = pathInfo.post
-        ? pathInfo.post.requestBody.content['application/json'].schema.properties
-        : {};
+      let props = {};
+      const schema = pathInfo?.post?.requestBody?.content['application/json'].schema;
+      if (schema.$ref) {
+        // $ref will be shaped like `#/component/schema/MyResponseType
```
hellobontempo, line 210: super small comment but components and schemas are plural in the `$ref` 
` "$ref": "#/components/schemas/KvLookupRequest"`

...but does that affect your mapping?

hellobontempo: oh nope! it doesn't, since you're using `reduce` and setting the keys on 212

hashishaw: no, but I should probably be accurate in my comment 😄

hellobontempo (approved):
Nice job! Thanks for the quick fix! 🚀

PR #14513, opened 2022-03-15
Add Semgrep Rules to OSS

HridoyRoy:


raskchanky (approved):

samsalisbury (commented):
Hi @HridoyRoy, I've not reviewed this in depth yet. Would it be possible to run this check in GitHub Actions instead of CircleCI? I believe we are trying to migrate everything to GHA in the long-run.

HridoyRoy:
> Hi @HridoyRoy, I've not reviewed this in depth yet. Would it be possible to run this check in GitHub Actions instead of CircleCI? I believe we are trying to migrate everything to GHA in the long-run.

Hi @samsalisbury , that sounds like a great idea! Let's port this to GH Actions later, but temporarily I'd like to just put the semgrep checks into OSS so that we can ensure no semgrep-related issues occur during OSS --> Ent Merges for 1.10. 
Thanks!

PR #14532, opened 2022-03-16
Addressing various Ember depreciations required for 4.0

Monkeychip:
- Removed [Ember.logger.](https://deprecations.emberjs.com/v3.x/#toc_ember-console-deprecate-logger) This was used in only one area, the not-found test and it honestly didn't seem to be doing anything. I removed it and the tests continued to pass. I also paused each test to make sure it was functioning as expected. 
- We have `jquery-intergration` set to false in our config, yet it still exists in our package json. Removed it.
- [Prevent](https://deprecations.emberjs.com/v3.x/#toc_ember-string-prototype_extensions) user from using Ember string methods on strings. 
- [Remove the reopen](https://deprecations.emberjs.com/v3.x/#toc_ember-built-in-components-reopen) on linkTo. This involved some scss changes and a restructure on the replication tabs page. Double checked all areas this effected look good.

### DIFF
--- a/ui/app/styles/components/tabs.scss
+++ b/ui/app/styles/components/tabs.scss
@@ -13,10 +13,16 @@
       &:focus {
         box-shadow: none;
       }
-      &.is-active,
-      &.is-active .tab {
+      &.active,
+      .tab {
         border-color: $blue;
         color: $blue;
+      }
+    }
+    li {
+      > a &.active {
+        border-color: $blue;
+        color: $blue !important;
       }
     }
   }
@@ -25,8 +31,8 @@
     &:focus {
       box-shadow: none;
     }
-    &.is-active a,
-    &.is-active .tab {
+    &.active a,
+    &.active .tab {
       border-color: $blue;
       color: $blue;
     }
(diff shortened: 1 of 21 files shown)


ui/app/styles/components/tabs.scss
```diff
@@ -13,20 +13,26 @@
       &:focus {
         box-shadow: none;
       }
-      &.is-active,
-      &.is-active .tab {
+      &.active,
```
zofskeez, line 16: This looks good. I wonder why the `activeClass` override was there in the first place? Was there an alternate styling for `.active` that was being applied maybe?

zofskeez (approved):
This all looks good to me nice work!

ui/app/styles/components/tabs.scss
```diff
@@ -13,20 +13,26 @@
...
+      .tab {
```
hashishaw, line 17: I think we still want to only target active tags here, right? 
```suggestion
      &.active .tab {
```

ui/app/styles/components/tabs.scss
Monkeychip: Honestly, no idea. But good idea. Let me do some more checking on .active anywhere else.

ui/app/styles/components/tabs.scss
Monkeychip: Good catch. I'll amend.

PR #14246, opened 2022-02-24
Make web_ui complient with Go's native embedding

ccapurso:
If directory is embeded using Go's embed package it should be present in git repository to avoid failures/erros messages in Go commands. For empty directory this can be done using common .gitkeep pattern.

I am opening a new PR from the original https://github.com/hashicorp/vault/pull/13646 due to CI issues.

Fixes https://github.com/hashicorp/vault/issues/13651

PR #14536, opened 2022-03-16
Update gitignore

hashishaw:
The previous `.gitignore` configuration was causing `http/web_ui/.gitkeep` to be deleted every time we created a new branch within the VSCode terminal. This update preserves the following behavior:

* Do not commit any of the files within `http/web_ui` 
* Preserve web_ui folder's complience with go_embed rules (#14246)

While removing the developer headache of making sure not to commit the deleted .gitkeep file in every branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,8 @@ _testmain.go
 /pkg/
 
 # Generated Web UI goes here
-/http/web_ui/**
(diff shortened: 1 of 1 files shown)


zofskeez (approved):
Many many thanks for this!!

PR #14542, opened 2022-03-16
Rename master key -> root key in docs

kalafut:
Replace more 'master key' references with 'root key' as part of inclusive language updates.


yhyakuna (approved):
👍

PR #14545, opened 2022-03-16
OIDC Logout Bug

zofskeez:
It was discovered that when logging out when previously authenticated via OIDC or JWT auth methods the `with` query param is set to `token` which hydrates the auth form with the `token` auth method selected rather than the previously logged in method. To ensure that the selected auth method persists I updated the auth service to use `sessionStorage` for this piece of state.

### DIFF
--- a/ui/app/components/auth-form.js
+++ b/ui/app/components/auth-form.js
@@ -211,15 +211,20 @@ export default Component.extend(DEFAULTS, {
   showLoading: or('isLoading', 'authenticate.isRunning', 'fetchMethods.isRunning', 'unwrapToken.isRunning'),
 
   authenticate: task(
-    waitFor(function* (backend, data) {
+    waitFor(function* (backendType, data) {
       const {
         selectedAuth,
         cluster: { id: clusterId },
       } = this;
       try {
         this.delayAuthMessageReminder.perform();
-        const authResponse = yield this.auth.authenticate({ clusterId, backend, data, selectedAuth });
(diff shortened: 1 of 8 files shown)


ui/app/components/auth-form.js
```diff
@@ -206,12 +211,15 @@ export default Component.extend(DEFAULTS, {
   showLoading: or('isLoading', 'authenticate.isRunning', 'fetchMethods.isRunning', 'unwrapToken.isRunning'),
 
   authenticate: task(
-    waitFor(function* (backendType, data) {
-      let clusterId = this.cluster.id;
+    waitFor(function* (backend, data) {
```
hashishaw, line 214: I do like that the previous name was clear that this is the backend type, not the backend name. Could we add a comment or keep the old name here?

hashishaw (approved):
LGTM! Thanks 🚀

ui/app/components/auth-form.js
```diff
@@ -118,14 +130,7 @@ export default Component.extend(DEFAULTS, {
     'selectedAuth',
     'selectedAuthIsPath',
     function () {
```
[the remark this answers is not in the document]
zofskeez: I thought so too and tried it but `this` becomes undefined.

ui/app/components/auth-form.js
zofskeez: Sure I can change it back. The auth service `authenticate` method has it named `backend` so I thought I would align it that way.

PR #14551, opened 2022-03-17
UI/Wrong sentinel error message for auth methods

hellobontempo:
### This PR cleans up some error handling that was slightly inaccurate and glimmerizes the `MessageError` component.

Previously the UI was intercepting the API error message and displaying an incorrect error message and warning flag. Below I was logged in as a user who did not have permission to enable the `userpass` auth method. The following warning displayed:

Screen Shot 2022-03-16 at 4 24 45 PM

<hr>

**With fix:**
Screen Shot 2022-03-17 at 12 47 26 PM

### DIFF
--- a/ui/app/components/mount-backend-form.js
+++ b/ui/app/components/mount-backend-form.js
@@ -131,7 +131,7 @@ export default Component.extend({
       try {
         yield mountModel.save();
       } catch (err) {
-        if (err.message === 'mountIssue') {
+        if (err.message === 'permissionIssue') {
           this.mountIssue = true;
           this.set('isFormInvalid', this.mountIssue);
           this.flashMessages.danger(
--- a/ui/app/components/mount-backend-form.js
+++ b/ui/app/components/mount-backend-form.js
@@ -112,43 +112,49 @@ export default Component.extend({
           throw err;
         }
       }
-      // TODO CMB: this flash message seems to be an over-generalization, so added the 'if...path !==' check, but want to revisit
-      // TODO should we check instead if config capabilities.capabilites.includes('deny')
-      if (!capabilities.get('canUpdate')) {
-        if (capabilities.get('path') !== 'userpass/config') {
-          // if there is no sys/mount issue then error is config endpoint.
-          this.flashMessages.warning(
-            'You do not have access to the config endpoint. The secret engine was mounted, but the configuration settings were not saved.'
-          );
-          // remove the config data from the model otherwise it will save it even if the network request failed.
-          [this.mountModel.maxVersions, this.mountModel.casRequired, this.mountModel.deleteVersionAfter] = [
-            0,
-            false,
-            0,
-          ];
-        }
-      }
+
+      let changedAttrKeys = Object.keys(mountModel.changedAttributes());
+      const updatesConfig =
+        mountModel.isV2KV &&
+        (changedAttrKeys.includes('casRequired') ||
+          changedAttrKeys.includes('deleteVersionAfter') ||
+          changedAttrKeys.includes('maxVersions'));
+
       try {
         yield mountModel.save();
       } catch (err) {
-        if (err.message === 'permissionIssue') {
+        if (err.httpStatus === 403) {
           this.mountIssue = true;
           this.set('isFormInvalid', this.mountIssue);
           this.flashMessages.danger(
             'You do not have access to the sys/mounts endpoint. The secret engine was not mounted.'
           );
           return;
-        } else if (err.errors) {
+        }
+        if (err.errors) {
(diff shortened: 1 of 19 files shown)


ui/app/components/mount-backend-form.js
```diff
@@ -112,18 +112,21 @@ export default Component.extend({
           throw err;
         }
       }
-
+      // TODO CMB: this flash message seems to be an over-generalization, so added the 'if...path !==' check, but want to revisit
+      // TODO should we check instead if config capabilities.capabilites.includes('deny')
       if (!capabilities.get('canUpdate')) {
```
hellobontempo, line 117: we have access to `capabilities: {capabilities: 'deny'}` wondering if that would be better here as `canUpdate` feels a little misleading

ui/app/components/mount-backend-form.js
```diff
@@ -112,18 +112,21 @@ export default Component.extend({
...
-        // if there is no sys/mount issue then error is config endpoint.
-        this.flashMessages.warning(
-          'You do not have access to the config endpoint. The secret engine was mounted, but the configuration settings were not saved.'
-        );
```
hellobontempo, line 120: this flash message was displaying when an API error was being returned (from either the path already existing or due to a permissions issue)

ui/app/mixins/policy-edit-controller.js
```diff
@@ -36,9 +36,8 @@ export default Mixin.create({
           }
           return this.transitionToRoute('vault.cluster.policy.show', m.get('policyType'), m.get('name'));
         })
-        .catch(() => {
```
hellobontempo, line 39: this stopped working after glimmerizing `message-error`

ui/app/templates/components/mount-backend-form.hbs
```diff
@@ -21,7 +21,7 @@
 <form {{action (perform this.mountBackend) on="submit"}}>
   <div class="box is-sideless is-fullwidth is-marginless">
     <NamespaceReminder @mode="enable" @noun={{if (eq this.mountType "auth") "Auth Method" "Secret Engine"}} />
-    <MessageError @model={{this.model}} @errorMessage={{this.errorMessage}} />
```
hellobontempo, line 24: is there a way to get the model to update when an AdapterError happens in the controller?

ui/lib/core/addon/components/message-error.js
```diff
@@ -11,48 +11,104 @@ import layout from '../templates/components/message-error';
...
-      const modelIsError = this.model?.isError;
-      if (errorMessage) {
-        return [errorMessage];
-      }
 
-      if (errors && errors.length > 0) {
-        return errors;
-      }
+class MessageError extends Component {
+  get errorMessage() {
```
hellobontempo, line 21: Is this a reasonable application of getters?

ui/lib/core/addon/components/message-error.js
```diff
@@ -11,48 +11,104 @@ import layout from '../templates/components/message-error';
...
+      return [adapterError.message];
     }
-  ),
-});
+    return null;
+  }
+}
+export default setComponentTemplate(layout, MessageError);
+
+// import { computed } from '@ember/object';
```
hellobontempo, line 60: Left for comparison and debugging. Will delete! (or revert if tests keep failing 🙃 )

ui/app/components/mount-backend-form.js
hashishaw: I don't think `deny` === `!canUpdate`, since they could have read capabilities but not update.

ui/app/components/mount-backend-form.js
```diff
@@ -135,9 +138,17 @@ export default Component.extend({
             'You do not have access to the sys/mounts endpoint. The secret engine was not mounted.'
           );
           return;
+        } else if (err.errors) {
+          let errors = err.errors.map((e) => {
+            if (typeof e === 'object') return e.title || e.message || JSON.stringify(e);
```
hashishaw, line 143: Nice handling of all the various ways an error could be sent!

ui/lib/core/addon/components/message-error.js
hashishaw: I would not use a getter unless you're doing something other than just getting the arg (for example, setting a default)

ui/app/templates/components/mount-backend-form.hbs
hashishaw: I think I need more clarification. We can sync offline

hashishaw (commented):
Looking good so far!

Monkeychip (approved):
Nice work

PR #14571, opened 2022-03-18
LinkTo Updates

zofskeez:
Updates link-to usages to remove the following deprecated patterns:
- curly bracket invocation
- positional params
- unknown args (invokeAction)

### DIFF
--- a/ui/app/templates/components/generated-item-list.hbs
+++ b/ui/app/templates/components/generated-item-list.hbs
@@ -23,13 +23,11 @@
       <nav class="tabs">
         <ul>
           {{#each tabs as |tab|}}
-            {{! template-lint-configure no-curly-component-invocation "warn" }}
-            {{#link-to params=tab.routeParams tagName="li" data-test-auth-section-tab=true}}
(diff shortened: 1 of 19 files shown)


ui/app/templates/components/generated-item-list.hbs
```diff
@@ -61,14 +59,10 @@
         this.itemType
       }} will be listed here. Create your first {{this.itemType}} to get started."
     >
-      {{! template-lint-configure no-curly-component-invocation "warn" }}
-      {{! template-lint-configure no-link-to-positional-params "warn" }}
-      {{#link-to "vault.cluster.access.method.item.create" this.itemType class="link"}}
+      <LinkTo @route="vault.cluster.access.method.item.create" @model={{this.itemType}} class="link">
```
Monkeychip, line 62: does the class="link" here do any styling still. e.g. is it still needed?

zofskeez: There is a `link` class in the `generic.scss` which appears to change the appearance from our standard link styling.

PR #14580, opened 2022-03-18
Redirect from the old API docs page to the new API docs page

sudomateo:
API documentation used to be served on `/api`. Now it's served on `/api-docs`. This pull request adds a permanent redirect so users can find the new API documentation using the old link.

### DIFF
--- a/website/redirects.next.js
+++ b/website/redirects.next.js
@@ -15,6 +15,11 @@ module.exports = [
     permanent: false,
   },
   {
+    source: '/api',
+    destination: '/api-docs',
+    permanent: true,
+  },
+  {
     source: '/api/secret/generic',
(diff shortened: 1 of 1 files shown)


website/redirects.next.js
```diff
@@ -14,6 +14,11 @@ module.exports = [
     destination: '/intro/getting-started',
     permanent: false,
   },
+  {
+    source: '/api',
+    destination: '/api-docs',
+    permanent: true,
```
sudomateo, line 20: I'll leave it up to the reviewers to determine if this should be a permanent redirect or not.

sudomateo:
Closing in favor of https://github.com/hashicorp/vault/pull/14579.

PR #14584, opened 2022-03-18
Remove record.toJSON()

Monkeychip:
Part of removing the deprecations - [see here](https://deprecations.emberjs.com/ember-data/v3.x/#toc_record-toJSON).

It's an easy fix largely because we were using this method inside an adapter and previously this method was more handy outside in say a model or component. Whereas before this method serialized the record, this new line does the same thing by calling this.serialize(). I compared the payloads, double checked the policies pages, and ran the enterprise test suite. All looked good.


hashishaw (approved):
Love these one-line fixes 💪

PR #14596, opened 2022-03-18
Runloop/Computed Dot Access Removal

zofskeez:
Accessing methods on run and computed functions is deprecated in Ember 4.0. This removes instances of those usages as well as accessing the Ember global which is also deprecated in 4.0.

https://deprecations.emberjs.com/v3.x/#toc_deprecated-run-loop-and-computed-dot-access
https://deprecations.emberjs.com/v3.x/#toc_ember-global


hashishaw (approved):
LGTM I had no idea we were using ember runloop so many places.

PR #14597, opened 2022-03-18
Ember With to Let

zofskeez:
As of Ember 4.0 the with helper will be deprecated. This converts all usages to the let helper.

https://deprecations.emberjs.com/v3.x/#toc_ember-glimmer-with-syntax


Monkeychip (approved):
Nice! Guessing no issues with yielded blocks?

zofskeez:
> Nice! Guessing no issues with yielded blocks?

No issues that I could see. We weren't using it with the optional `else` block which made the refactor to `let` pretty straightforward.

PR #14608, opened 2022-03-21
Vault documentation: added new client count faqs for vault 1.10

taoism4504:
Per my sync with Darshana, new and updated faqs were addressed for vault 1.10

:mag: [Deploy Preview](https://vault-git-client-count-changes-vault-1-10-hashicorp.vercel.app/docs/concepts/client-count/faq)


website/content/docs/concepts/client-count/faq.mdx
```diff
@@ -218,4 +226,22 @@ However, creating a new token across a parent/child namespace boundary could res
...
+
+When migrating mounts within a namespace, client counts are not impacted.
+
+- For each existing alias on the source mount and its corresponding entity, if the new namespace has an entity corresponding to it, then you can assign a corresponding alias for the mount and the client count will not be impacted. For example, say I am a user `prat` on namespace ns1/ and namespace ns2/, and I have an alias on the source mount in ns1/. After the move operation, I just need an alias added to the `prat` entity in ns2/ for the target mount. This keeps the client counts the same.
+- For the example above, say I do not have an entity `prat` in ns2/. That will need to be created in order to make an alias associating it with the target mount. This increases the client count by 1.
+- Following the above example, let's say after the move operation, the entity `prat` in ns1/ has no other aliases associated with it, i.e., the alias for the source mount was its only alias. At this point, we can clean up and remove the `prat` entity in ns1/. This will decrease the client count by 1.
```
yhyakuna, line 247:
```suggestion
- For each existing alias on the source mount and its corresponding entity, if the new namespace has an entity corresponding to it, then you can assign a corresponding alias for the mount and the client count will not be impacted. For example, say I am a user `prat` on namespace `ns1/` and namespace `ns2/`, and I have an alias on the source mount in `ns1/`. After the move operation, I just need an alias added to the `prat` entity in `ns2/` for the target mount. This keeps the client counts the same.
- For the example above, say I do not have an entity `prat` in `ns2/`. That will result in a creation of a default entity in the target mount (`ns2/`). This increases the client count by 1.
- Following the above example, let's say after the move operation, the entity `prat` in `ns1/` has no other aliases associated with it, i.e., the alias for the source mount was its only alias. At this point, we can clean up and remove the `prat` entity in `ns1/`. This will decrease the client count by 1.
```

yhyakuna (approved):
Minor re-wording _suggestion_ but looks good to me.

PR #14636, opened 2022-03-22
identity/oidc: updates documentation for PKCE, client types, and default resources

austingebauer:
This PR adds documentation for a few features that were recently added to Vault OIDC provider:
- PKCE (https://github.com/hashicorp/vault/pull/13917)
- Client types (https://github.com/hashicorp/vault/pull/13917)
- Default resources (https://github.com/hashicorp/vault/pull/14119)

Additionally, this PR reworks some of the content:
- General documentation page follows "Setup" -> "Usage" convention of other documentation
- General documentation uses default resources for a minimal working configuration
- General documentation references other HashiCorp products with OIDC auth methods that we've tested


fairclothjm (approved):

website/content/api-docs/secret/identity/oidc-provider.mdx
```diff
@@ -21,7 +21,7 @@ This endpoint creates or updates a Provider.
   Vault's `api_addr` as the `scheme://host:port` component and `/v1/:namespace/identity/oidc/provider/:name` as the path
   component. If provided explicitly, it must point to a Vault instance that is network reachable by clients for ID token validation.
 
-- `allowed_client_ids` `([]string: <optional>)` – The client IDs that are permitted to use the provider. If empty, no clients are allowed. If "*", all clients are allowed.
+- `allowed_client_ids` `([]string: <optional>)` – The client IDs that are permitted to use the provider. If empty, no clients are allowed. If `"*"` provided, all clients are allowed.
```
calvn, line 24:
```suggestion
- `allowed_client_ids` `([]string: <optional>)` – The client IDs that are permitted to use the provider. If empty, no clients are allowed. If `"*"` is provided, all clients are allowed.
```

website/content/docs/concepts/oidc-provider.mdx
```diff
@@ -7,32 +7,35 @@ description: >-
...
 
-The allowed client IDs parameter utilizes the list of client IDs that have been generated by Vault as a part of client registration. By default, all clients will be *disallowed*. Providing an asterisk(*) as the parameter value will allow all clients to use the provider. The scopes parameter employs a list of references to named scope resources. The values provided are discoverable by the `scopes_supported` key in the OIDC discovery document of the provider. By default, a provider will have the `openid` scope available. See the scopes section below for more details on the `openid` scope.
+The allowed client IDs parameter utilizes the list of client IDs that have been generated by Vault as a part of client registration. By default, all clients will be *disallowed*. Providing an asterisk(*) as the parameter value will allow all clients to use the provider.
```
calvn, line 36: Could go either way, but was thinking that being explicit here might help.

```suggestion
The allowed client IDs parameter utilizes the list of client IDs that have been generated by Vault as a part of client registration. By default, all clients will be *disallowed*. Providing `*` as the parameter value will allow all clients to use the provider.
```

website/content/docs/concepts/oidc-provider.mdx
```diff
@@ -85,39 +88,84 @@ Several named scopes can be made available on an individual provider. Note that
...
+
 ### Keys
 
-Key resources are referenced by clients via the key parameter. This parameter specifies the key that will be used to sign ID tokens for the client. See existing [documentation](/api-docs/secret/identity/tokens#create-a-named-key) for details on keyring management, supported signing algorithms, rotation periods, and verification TTLs. Currently, a key referenced by a client cannot be changed.
+Key resources are referenced by clients via the `key` parameter. This parameter specifies the key that will be used to sign ID tokens for the client. See existing [documentation](/api-docs/secret/identity/tokens#create-a-named-key) for details on keyring management, supported signing algorithms, rotation periods, and verification TTLs. Currently, a key referenced by a client cannot be changed.
+
+Each Vault namespace will contain a built-in key resource named `default`. Clients that don't
+specify the `key` parameter at creation time will use the `default` key. The `default` key
+will use the `RS256` signing algorithm, allow all client IDs, and have rotation and verification
+TTLs of `24h`. The `default` key can be modified but not deleted.
```
calvn, line 168: Wondering if this should be in list form so that it's easier to grok.

taoism4504 (approved):
LGTM

website/content/docs/upgrading/upgrade-to-1.10.x.mdx
```diff
@@ -52,3 +52,26 @@ were 26 characters; they now have a minimum of 95 characters. However, existing
...
+ - A "default" OIDC provider that's usable by all client applications
+ - A "default" key for signing and verification of ID tokens
+ - An "allow_all" assignment which authorizes all Vault entities to authenticate via a
+   client application
+
+If you created an [OIDC provider](/api-docs/secret/identity/oidc-provider#create-or-update-a-provider)
+with the name "default", [key](/api-docs/secret/identity/tokens#create-a-named-key) with the
+name "default", or [assignment](/api-docs/secret/identity/oidc-provider#create-or-update-an-assignment)
+with the name "allow_all" using the Vault 1.9 tech preview, the installation of these built-in
+resources will be skipped without an error. We recommend deleting any resources that have
```
calvn, line 74: Interesting, I didn't know that we simply skipped these builtin resources. Do we make any assumptions in the code on these literal values (e.g. "allow_all")?

austingebauer: Ya, we check if they exist, and if so, skip creation. See https://github.com/hashicorp/vault/blob/main/vault/identity_store_oidc_provider.go#L2386. 

There is an assumption made based on the literal value at https://github.com/hashicorp/vault/blob/main/vault/identity_store_oidc_provider.go#L2316. This makes me wonder if we should more strongly emphasize the need to delete preexisting resources that collide with these names.

website/content/docs/concepts/oidc-provider.mdx
austingebauer: Good idea. Done in a93251e.

website/content/docs/upgrading/upgrade-to-1.10.x.mdx
calvn: Yeah, I have similar thoughts too... Another thing to consider is that we disable modification of these entries after the upgrade so it'd be really hard for them to delete these entries afterwards unless they resort to using `sys/raw/`.

austingebauer: I've added a stronger emphasis on this in 55e14cf. I agree. There are already checks to ensure that deletion/modification don't happen in the 1.10 code. For example, https://github.com/hashicorp/vault/blob/main/vault/identity_store_oidc_provider.go#L660-L663 and https://github.com/hashicorp/vault/blob/main/vault/identity_store_oidc_provider.go#L760-L763.

calvn (approved):

PR #14646, opened 2022-03-22
Route renderTemplate Deprecation

zofskeez:
As of Ember 4.0 the `renderTemplate` method on routes is being deprecated. This PR removes the usages and refactors to retain the intended results.

https://deprecations.emberjs.com/v3.x/#toc_route-render-template

### DIFF
--- a/ui/app/routes/loading.js
+++ b/ui/app/routes/loading.js
@@ -1,25 +1,9 @@
 import Route from '@ember/routing/route';
-import { inject as service } from '@ember/service';
 
 export default Route.extend({
-  router: service(),
-  init() {
+  setupController(controller) {
     this._super(...arguments);
-    this.router.on('routeWillChange', (transition) => {
-      this.set('myTargetRouteName', transition.to.name);
-    });
-  },
-  renderTemplate() {
-    let targetName = this.myTargetRouteName;
-    let isCallback =
-      targetName === 'vault.cluster.oidc-callback' || targetName === 'vault.cluster.oidc-callback-namespace';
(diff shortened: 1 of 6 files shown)


ui/app/routes/loading.js
```diff
@@ -1,25 +1,9 @@
...
-import { inject as service } from '@ember/service';
 
 export default Route.extend({
-  router: service(),
-  init() {
+  setupController(controller) {
     this._super(...arguments);
-    this.router.on('routeWillChange', (transition) => {
-      this.set('myTargetRouteName', transition.to.name);
-    });
```
zofskeez, line 10: This pattern wasn't working when I tested prior to making the change since the router updates for intermediate transitions so `loading` was the value for `transition.to.name`. This resulted in the standard `LogoSplash` displaying rather than the OIDC callback. Using `location.pathname` solves the issue and the new `OidcCallbackSplash` component is displayed which prevents a change in display while transition from the loading state to the oidc-callback route.

ui/lib/core/addon/components/masked-input.js
```diff
@@ -52,7 +52,9 @@ export default Component.extend({
       this.onChange(value);
     },
     handleKeyUp(name, value) {
-      this.onKeyUp(name, value);
+      if (this.onKeyUp) {
+        this.onKeyUp(name, value);
+      }
```
zofskeez, line 57: I came across an error in the console while testing where `onKeyUp` was undefined so I added a check.

hashishaw (approved):
Nice work!

PR #14654, opened 2022-03-22
docs/ssct: change service side to server side

austingebauer:
This PR updates some documentation that uses "service" instead of "server" for server side consistent tokens.


mladlow:
I'm going to go ahead and merge this so it's included in the docs when they go live.

Issue #12350, opened 2021-08-18
Unable to copy inline text from vault secret


trutled3:
**Describe the bug**
After creating a json secret in vault, I am unable to copy inline text from the secret. I am only able to copy inline text after hitting "Edit" on the secret. It may be related to https://github.com/hashicorp/vault/pull/11500

**To Reproduce**
1. Create json secret in kv secrets engine
2. Attempt to highlight text and right click to copy and unable to do so.
3. See error

![image](https://user-images.githubusercontent.com/89163186/129967917-83573156-4a9d-4827-af65-329a53027c45.png)

**Expected behavior**
To be able to highlight inline comment and able to copy without having to click "Edit"

**Environment:**
* Vault Server Version (retrieve with `vault status`): v1.8.1
* Vault CLI Version (retrieve with `vault version`): N/A, was able to replicate with docker.
* Server Operating System/Architecture: N/A, was able to replicate with docker.

Vault server configuration file(s):
Can replicate with 
docker run -it --rm -p 8200:8200 vault:1.8.1

timonych:
> @timonych thank you for posting this issue. We're looking into it. It should be an easy fix, but there might be a little more to it, involving the removal of a deprecated plugin. I created a ticket and will hopefully tackle this soon.

Thanks for information, waiting for fix this trouble.

To be honest, @trutled3 has opened this issue. I've only add additional info about version and problem.

Monkeychip:
This has been fixed and backported to 1.8.x -1.10.x. The fix for 1.8 -1.10 will show a blinking cursor when in readOnly mode but you will now be able to copy/paste. The fix for 1.11 will not show a blinking cursor and will allow you to copy/paste. The fix for 1.11 was a complete refactor which gave us more control over the css and removed a deprecated wrapper. Thanks @trutled3 for opening the issue and everyone for upvoting or commenting on it.

PR #14659, opened 2022-03-23
Remove Ivy Codemirror

Monkeychip:
Replacing the deprecated ivycodemirror with an ember modifier. See Deprecation and instructions [here](https://github.com/IvyApp/ivy-codemirror).

Notes:
- This PR came about because of [this issue.](https://github.com/hashicorp/vault/issues/12350) Unfortunately without having some regression issues there was no simple way to allow copying while also not showing the cursor. In order to do this we needed to remove IvyCodemirror.
- We no longer use an options object.
- This is our first custom modifier. If you're curious how they work, [this documentation](https://www.npmjs.com/package/ember-modifier) was very helpful.
- This became a bigger PR than I had originally thought it would be. My apologies, but these changes should happen all at once.

https://user-images.githubusercontent.com/6618863/160458318-24c30605-7eef-414c-874e-623e7abed069.mp4

### DIFF
--- a/ui/app/components/json-editor.js
+++ b/ui/app/components/json-editor.js
@@ -16,7 +16,7 @@ import { action } from '@ember/object';
  * @param {string} [helpText] - helper text.
  * @param {Object} [extraKeys] - Provides keyboard shortcut methods for things like saving on shift + enter.
  * @param {Array} [gutters] - An array of CSS class names or class name / CSS string pairs, each of which defines a width (and optionally a background), and which will be used to draw the background of the gutters.
- * @param {string} [mode] - The mode defined for styling. Right now we only import ruby so must mode but be ruby or defaults to javascript. If you wanted another language you need to import it into the modifier.
+ * @param {string} [mode] - The mode defined for styling. Right now we only import ruby so mode must but be ruby or defaults to javascript. If you wanted another language you need to import it into the modifier.
  * @param {Boolean} [readOnly] - Sets the view to readOnly, allowing for copying but no editing. It also hides the cursor. Defaults to false.
  * @param {String} [theme] - Specify or customize the look via a named "theme" class in scss.
  * @param {String} [value] - Value within the display. Generally, a json string.


ui/app/styles/components/codemirror.scss
```diff
@@ -168,31 +168,12 @@ $gutter-grey: #2a2f36;
 }
 
 .readonly-codemirror {
-  .CodeMirror-cursors {
```
Monkeychip, line 171: this was not working. We have it working now under a new class that only shows if the editor is readOnly.

ui/app/templates/components/console/log-json.hbs
```diff
@@ -2,7 +2,10 @@
   <JsonEditor
     @showToolbar={{false}}
     @value={{stringify this.content}}
-    @options={{hash readOnly=true lineNumbers=false autoHeight=true gutters=false theme="hashi auto-height"}}
```
Monkeychip, line 5: all options objects were spread out so that the individual properties were sent to the json-editor.

ui/tests/integration/components/console/log-json-test.js
```diff
@@ -19,8 +19,7 @@ module('Integration | Component | console/log json', function (hooks) {
     this.set('content', objectContent);
 
     await render(hbs`{{console/log-json content=content}}`);
-    const instance = this.codeMirror.instanceFor(find('[data-test-component=json-editor]').id);
```
Monkeychip, line 22: this old way of selecting the codemirror instance no longer works. I suspect because we're initializing it through a modifier instead of on the component via ivycodemirror.

ui/ember-cli-build.js
```diff
@@ -26,10 +26,6 @@ const appConfig = {
       return `${config.rootURL.replace(/\/$/, '')}${filePath}`;
     },
   },
-  codemirror: {
```
Monkeychip, line 29: everything removed from this file was added into the modifier.

zofskeez (changes requested):
Great work on this thanks! Small wording issue on a comment and I wanted to get your thoughts on the `mode` arg before approving.

ui/app/components/json-editor.js
```diff
@@ -14,50 +14,24 @@ import { action } from '@ember/object';
  * @param {Function} [valueUpdated] - action to preform when you edit the codemirror value.
  * @param {Function} [onFocusOut] - action to preform when you focus out of codemirror.
  * @param {string} [helpText] - helper text.
- * @param {object} [options] - option object that overrides codemirror default options such as the styling.
+ * @param {Object} [extraKeys] - to provide keyboard shortcut methods for things like saving on shift + enter.
+ * @param {Array} [gutters] - An array of CSS class names or class name / CSS string pairs, each of which defines a width (and optionally a background), and which will be used to draw the background of the gutters.
+ * @param {string} [mode] - right now we only import ruby so must mode but be ruby or defaults to javascript. If you wanted another language you need to import it into the modifier.
```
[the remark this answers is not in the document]
Monkeychip: That's a valid question. I think however I prefer mode because it aligns with the codemirror property. It would be easy for someone to look up. I also don't want someone to think they only can use Ruby in the future. They can use lots of other options they just need to import it. They'll immediately error to if they use something other than ruby.

I'll amend the comment wording.

ui/app/templates/components/console/log-json.hbs
Monkeychip: Good catch. I removed this because before it was used to set the viewportMargin. [See here](https://github.com/hashicorp/vault/pull/14659/files#diff-9401613f84fc7b84c9749d59349bcac5eee3d99028b28d2241b6e3e985dd419aL43)

ui/app/modifiers/code-mirror.js
```diff
@@ -0,0 +1,68 @@
...
+  _setup() {
+    if (!this.element) {
+      throw new Error('CodeMirror modifier has no element');
+    }
+    const editor = codemirror(this.element, {
+      // IMPORTANT: `gutters` must come before `lint` since the presence of
+      // `gutters` is cached internally when `lint` is toggled
+      gutters: this.args.named.gutters || ['CodeMirror-lint-markers'],
+      matchBrackets: true,
+      lint: { lintOnChange: true },
```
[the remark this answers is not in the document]
Monkeychip: Good catch. This was left over from me playing with the linting. I'll change.

ui/app/components/json-editor.js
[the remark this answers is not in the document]
zofskeez: Sounds good so if I were to pass in `php` right now for example `codemirror` would throw an error since it's not imported?

ui/app/templates/components/console/log-json.hbs
zofskeez: Ahh ok I see now thanks!

ui/app/components/json-editor.js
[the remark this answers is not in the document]
Monkeychip: correct, it wouldn't work.

ui/app/modifiers/code-mirror.js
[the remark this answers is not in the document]
Monkeychip: Okay, strange. I changed back and it broke the linter. Let me investigate a little tomorrow because we are calling performLint() so we shouldn't need this.

zofskeez (approved):

PR #14661, opened 2022-03-23
PKI: Add missing default cases within switch statements

stevendpclark:
- Harden the code base a bit adding default's to switch statements
   to various error handlers and processing statements.
 - Fixup some error messages to include proper values we support.

### DIFF
--- a/builtin/logical/pki/cert_util.go
+++ b/builtin/logical/pki/cert_util.go
@@ -748,7 +748,7 @@ func signCert(b *backend,
 		}
 
 	default:
-		return nil, errutil.InternalError{Err: fmt.Sprintf("supported key type value: %s", data.role.KeyType)}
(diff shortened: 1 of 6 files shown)


builtin/logical/pki/cert_util.go
```diff
@@ -747,6 +747,8 @@ func signCert(b *backend,
 			return nil, errutil.UserError{Err: "RSA keys < 2048 bits are unsafe and not supported"}
 		}
 
+	default:
+		return nil, errutil.InternalError{Err: fmt.Sprintf("supported key type value: %s", data.role.KeyType)}
```
victorr, line 751: Should that be "unsupported key type..."?

victorr (approved):
Very nice! Just a possible correction.

builtin/logical/pki/cert_util.go
stevendpclark: Oops good eye, will fix.

Issue #13647, opened 2022-01-13
Policies can only have lower case names but upper case gets accepted and then mangled to lower case


firstdorsal:
Policies get accepted if they have an uppercase name and without any further notice mangled to their lowercase representation.

IMO policy names should be rejected if they contain anything else but lowercase and not silently be mangled to lowercase as this behaviour causes issues.

ncabatoff:
Hi @firstdorsal,

I would prefer that behaviour too.  I don't think it's likely we change this though due to backwards compatibility concerns.  I'll leave this open for now, possibly our developer experience team would like to think through whether we ought to change this and how it might be done safely.

maxb:
This problem also exists elsewhere in Vault - identity entity and identity group names, come to mind - although there the behaviour is case-insensitive but case-preserving. This can cause some decidedly awkward interactions when using other software to orchestrate Vault configuration, and that other software assumes names will simply be used as is, and can be simply compared (case-sensitively) to enforce uniqueness.

digivava:
Thanks for bringing this to our attention! For now we have just added a warning to both the CLI output and the API response object whenever the resulting policy name differs from the one in the request. Hopefully this at least calls some attention to this behavior when it occurs.

Fuco1:
This is also annoying when using terraform to automatically generate some policies for example for applications registered in azure or aws.  Often they have upper case names and the policies are created with upper case names but then converted to lowercase.  Every terraform run it now gives me a change set with the cases. :/  I'll update the terraform script to lower-case the policy names, but it is annoying.

PR #14670, opened 2022-03-23
Warn on upper case in policy name

digivava:
Provides a warning when an upper case character is given in a policy name, as our current behavior of just lower-casing anything that comes in without telling anyone has caused some users confusion (see https://github.com/hashicorp/vault/issues/13647).

### DIFF
--- a/command/policy_write.go
+++ b/command/policy_write.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"os"
 	"strings"
-	"unicode"
 
 	"github.com/mitchellh/cli"
 	"github.com/posener/complete"
@@ -92,10 +91,13 @@ func (c *PolicyWriteCommand) Run(args []string) int {
 	}
 
 	// Policies are normalized to lowercase
-	policyName := args[0]
-	warnOnUpperCase(c.UI, policyName)
-	formattedPolicyName := strings.TrimSpace(strings.ToLower(policyName))
+	name := args[0]
+	formattedName := strings.TrimSpace(strings.ToLower(name))
 	path := strings.TrimSpace(args[1])
+
+	if name != formattedName {
+		c.UI.Warn(fmt.Sprintf("Policy name was converted to %s", formattedName))
+	}
 
 	// Get the policy contents, either from stdin of a file
 	var reader io.Reader
@@ -122,20 +124,11 @@ func (c *PolicyWriteCommand) Run(args []string) int {
 	}
 	rules := buf.String()
 
-	if err := client.Sys().PutPolicy(formattedPolicyName, rules); err != nil {
+	if err := client.Sys().PutPolicy(formattedName, rules); err != nil {
 		c.UI.Error(fmt.Sprintf("Error uploading policy: %s", err))
 		return 2
 	}
 
-	c.UI.Output(fmt.Sprintf("Success! Uploaded policy: %s", formattedPolicyName))
+	c.UI.Output(fmt.Sprintf("Success! Uploaded policy: %s", formattedName))
 	return 0
-}
-
-func warnOnUpperCase(ui cli.Ui, s string) {
-	for _, r := range s {
-		if unicode.IsUpper(r) && unicode.IsLetter(r) {
-			ui.Warn("Policy name contains upper-case character(s) and will be converted to lower-case.")
-			break
-		}
-	}
 }
--- a/command/policy_write.go
+++ b/command/policy_write.go
@@ -91,8 +91,8 @@ func (c *PolicyWriteCommand) Run(args []string) int {
 	}
 
 	// Policies are normalized to lowercase
-	name := args[0]
-	formattedName := strings.TrimSpace(strings.ToLower(name))
+	policyName := args[0]
+	formattedName := strings.TrimSpace(strings.ToLower(policyName))
 	path := strings.TrimSpace(args[1])
 
 	// Get the policy contents, either from stdin of a file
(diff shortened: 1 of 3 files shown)


command/policy_write.go
```diff
@@ -91,7 +92,9 @@ func (c *PolicyWriteCommand) Run(args []string) int {
 	}
 
 	// Policies are normalized to lowercase
-	name := strings.TrimSpace(strings.ToLower(args[0]))
+	policyName := args[0]
+	warnOnUpperCase(c.UI, policyName)
+	name := strings.TrimSpace(strings.ToLower(policyName))
```
AnPucel, line 97: Maybe name this `normalizedName` or `formattedName`? to differentiate it from simple `policyName`

digivava: I didn't even notice that variable name but you're so right, thank you!

command/policy_write.go
```diff
@@ -127,3 +130,12 @@ func (c *PolicyWriteCommand) Run(args []string) int {
 	c.UI.Output(fmt.Sprintf("Success! Uploaded policy: %s", name))
 	return 0
 }
+
+func warnOnUpperCase(ui cli.Ui, s string) {
+	for _, r := range s {
+		if unicode.IsUpper(r) && unicode.IsLetter(r) {
```
AnPucel, line 136: Super nit-picky/ curious question. If `go` supports short circuiting, should this check the Letter first and then whether or not it's uppercase?

AnPucel: Or you may just need the `Upper` check b/c it seems like it also checks that it's a letter? idk

digivava: I think you're totally right, but resolving this conversation as I went with a different approach that doesn't iterate over the letters, after discussing with @VinnyHC .

vault/logical_system.go
```diff
@@ -2435,14 +2435,21 @@ func (b *SystemBackend) handlePoliciesSet(policyType PolicyType) framework.Opera
 			return nil, err
 		}
 
+		name := data.Get("name").(string)
```
averche, line 2438: I believe there is a wrapper `GetString` that does the safe string casting.

vault/logical_system.go
```diff
@@ -2435,14 +2435,21 @@ func (b *SystemBackend) handlePoliciesSet(policyType PolicyType) framework.Opera
...
 			Type:      policyType,
 			namespace: ns,
 		}
 		if policy.Name == "" {
 			return logical.ErrorResponse("policy name must be provided in the URL"), nil
 		}
+		if name != policy.Name {
+			if resp == nil {
+				resp = &logical.Response{}
+			}
```
averche, line 2450:
```suggestion
			resp = &logical.Response{}
```

I think `resp` is always `nil` at this point

vault/logical_system.go
digivava: Good thing for me to know about! I just looked at it though and it seems like the method (I assume you mean the one from go-secure-stdlib/strutil) takes a map[string]interface{}, whereas data is actually a *framework.FieldData.

command/policy_write.go
```diff
@@ -91,9 +91,14 @@ func (c *PolicyWriteCommand) Run(args []string) int {
...
 	// Policies are normalized to lowercase
-	name := strings.TrimSpace(strings.ToLower(args[0]))
+	name := args[0]
+	formattedName := strings.TrimSpace(strings.ToLower(name))
 	path := strings.TrimSpace(args[1])
 
+	if name != formattedName {
+		c.UI.Warn(fmt.Sprintf("Policy name was converted to %s", formattedName))
+	}
+
```
VinnyHC, line 101: Does it make sense to move this down after the client call? If we're going to error anyway no sense warning.

VinnyHC (commented):
Looks good! One comment about order of operations.

command/policy_write.go
digivava: Good point, would be weird to see the warning before that.

vault/logical_system.go
averche:
```suggestion
		name := req.GetString("name")
```

I believe it's this one.

command/policy_write.go
```diff
@@ -119,11 +120,15 @@ func (c *PolicyWriteCommand) Run(args []string) int {
...
 
-	if err := client.Sys().PutPolicy(name, rules); err != nil {
+	if err := client.Sys().PutPolicy(formattedName, rules); err != nil {
 		c.UI.Error(fmt.Sprintf("Error uploading policy: %s", err))
 		return 2
 	}
 
-	c.UI.Output(fmt.Sprintf("Success! Uploaded policy: %s", name))
+	if name != formattedName {
+		c.UI.Warn(fmt.Sprintf("Policy name was converted to %s", formattedName))
```
AnPucel, line 129: do we want to warn in the case that space was trimmed off? I'm wondering if that will look weird when warned b/c they'll just look identical haha

digivava: The trimming is actually happening on the path to the HCL file, not the policy name. So we don't need to worry about warning because the file path name isn't getting saved anywhere, it's just getting used to populate the policy.

vault/logical_system.go
digivava: As discussed offline, we're going to keep this one as is since this is a line from the original code, which seems to be intentionally passing in and using data *framework.FieldData instead of the string from the logical.Request.

command/policy_write.go
AnPucel: Maybe I'm completely misreading the code above. But I was assuming `args[0]` is the policy name. and then we trim it to the `formattedName`

command/policy_write.go
```diff
@@ -91,7 +91,8 @@ func (c *PolicyWriteCommand) Run(args []string) int {
 	}
 
 	// Policies are normalized to lowercase
-	name := strings.TrimSpace(strings.ToLower(args[0]))
+	name := args[0]
```
AnPucel, line 94: nit: `policyName` maybe?. Maybe it's obvious when you read the code though.

command/policy_write.go
digivava: Oh sorry I was looking at the TrimSpace on the next line for args[1]! You're right, we are trimming it. I'll change it so there's quotes around the formatted name in the warning, so the space is obvious.

AnPucel (approved):
LGTM

VinnyHC (approved):
Looks good to me!

PR #14673, opened 2022-03-23
Fix a few linting errors in gen_openapi.sh script

averche:
Fixing a few small linting errors in `scripts/gen_openapi.sh` picked up by `shellcheck`.

### DIFF
--- a/scripts/gen_openapi.sh
+++ b/scripts/gen_openapi.sh
@@ -29,28 +29,28 @@ echo "Mounting all builtin backends..."
 # Read auth backends
 codeLinesStarted=false
 inQuotesRegex='".*"'
-while read line; do
+while read -r line; do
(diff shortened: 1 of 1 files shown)


scripts/gen_openapi.sh
```diff
@@ -87,4 +86,5 @@ fi
 kill $VAULT_PID
 sleep 1
 
-printf "\nopenapi.json generated."
```
AnPucel, line 90: My shellcheck didn't pick this up. Just curious if you're not supposed to use printf.

AnPucel (approved):
Thanks for fixing these!

scripts/gen_openapi.sh
averche: `printf` prints an extra character at the end, echo seemed better.

PR #14388, opened 2022-03-07
Add context-aware functions to vault/api

averche:
### Credits

This pull request is based on a community contribution #14152 by [ilyakaznacheev](https://github.com/ilyakaznacheev) and resolves #14070

### Description

For each API wrapper function in `vault/api`, we are adding a context-aware equivalent. These new functions will have `WithContext` appended to their names.

### Implementation Notes

One of the issues that I discovered in the original PR (#14152) is a potential context leak due to this piece of code inside `RawRequestWithContext`:

```go
	if timeout != 0 {
		// Note: we purposefully do not call cancel manually. The reason is
		// when canceled, the request.Body will EOF when reading due to the way
		// it streams data in. Cancel will still be run when the timeout is
		// hit, so this doesn't really harm anything.
		ctx, _ = context.WithTimeout(ctx, timeout)
	}
```

To prevent context leaks, the timeout logic has been extracted out of `RawRequestWithContext` and into each `...WithContext` wrapper (using `withConfiguredTimeout` helper).  Additionally, since `RawRequestWithContext` is still used outside of `vautl/api`, its implementation has been moved into unexported `rawRequestWithContext`

See [this comment](https://github.com/hashicorp/vault/pull/14152#r810400167) for more details.
 
### Testing

All existing tests that were calling vault/api functions have been switched to call `WithContext` equivalents with `context.Background()` passed in. Since `WithContext` functions invoke the regular ones, this change should increase code coverage without any negative side effects.

PR #14677, opened 2022-03-23
Replace http method strings with net/http constants

averche:
This is a quick refactoring PR to follow up #14388 where @VinnyHC pointed out that we were using hard-coded strings instead of `net/http` string constants.


digivava:
Yay! I confirmed that GET, POST, PUT, PATCH, and DELETE don't have any more hard-coded strings, but it looks like "LIST" is still hard-coded.

averche:
Good catch!

digivava (approved):

averche:
It appears that `LIST` is not one of the [standard http methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) and it does not have a corresponding constant in [net/http](https://pkg.go.dev/net/http#pkg-constants).

PR #14685, opened 2022-03-24
Vault documentation: added known issues to 1.10.x upgrade guide

taoism4504:
Per Aarti and Meggie, known issues were added to the upgrade guide for 1.10.x. 
Since the known issues come from the release notes v.1.10.0, Meggie wanted to use the #@include directive so in case updates need to be made to known issues, the updates can be automatically propagated without having to access multiple files to make the updates. I'll need to figure out how to do this in the future.

:mag: [Deploy Preview](https://vault-git-update-upgrade-1-10-guide-hashicorp.vercel.app/docs/upgrading/upgrade-to-1.10.x)


yhyakuna (approved):
LGTM

PR #14695, opened 2022-03-24
Fix markdown formatting issue in 1.10 upgrade docs

daveadams:
The `allow_all` token on the ["Upgrading to Vault 1.10.x"](https://www.vaultproject.io/docs/upgrading/upgrade-to-1.10.x) page causes a run of unwanted italics (see attached screenshot).

formatting-issue

This commit changes the quoting on that token from double-quotes to backquotes to avoid this issue. It also changes the quoting of other tokens in the same section for formatting consistency.

An alternative approach would be to escape just the underscore in the `allow_all` token, but backquoting API tokens seems to be typical style.


daveadams:
Looks like this has been fixed elsewhere.

Issue #12329, opened 2021-08-15
Several Vault endpoints spuriously implement CreateOperation & advertise x-vault-createSupported in OpenAPI


maxb:
Several Vault endpoints spuriously implement CreateOperation and so advertise x-vault-createSupported in OpenAPI, when it is meaningless to do so.

This may go unnoticed by the vast majority of users, but if you are writing checks which consult valid operations in OpenAPI, to vet whether your users are making mistakes in writing their Vault policies, it is important.

It's a coding error in a Vault framework.Path object, to implement logical.CreateOperation in your Operations or Callbacks, if you do not also implement ExistenceCheck - yet there are a number of instances of doing so in Vault code.

Here are the ones I've found so far:

* ad/rotate-role/{name}
* ad/rotate-root
* kv-v2/config
* kv-v2/delete/{path}
* kv-v2/destroy/{path}
* kv-v2/undelete/{path}
* transit/cache-config
* auth/kubernetes/config

In each case, the logical.CreateOperation entry should be removed from the Operations or Callbacks map.

Ideally, Vault would log a warning on startup if such a coding error was present, when initialising plugins, to guide future codes to use the API correctly.

hsimon-hashicorp:
Hi @maxb! I spoke with our engineers, and they agree your proposed solution is the one they would also choose. Please feel free to tackle this, and submit a PR and link it to this issue. (Anything with a label of `good-first-issue` is one where we welcome community contributions!) I apologize about the delay in getting back to you, and if you have any questions, please feel free to let me know. :)

maxb:
Thanks, I will raise with my employer whether I can sign the CLA as an individual contributor, or the company needs to sign it. This might take a little while - if anyone else wants to take the idea and run with it in the meantime, that's fine too!

maxb:
It has been a long while since I opened this issue, but since my CLA is now sorted and I have some time over Christmas, I'm now in a position to start making some PRs for it.

maxb:
I have created the 10 PRs to implement this change.

PR #14703, opened 2022-03-24
Remove CreateOperation from transit/cache-config

AnPucel:
A `logical.CreateOperation` without the implementation of an `ExistenceCheck` defaults to becoming a `logical.UpdateOperation`. We're removing the `logical.CreateOperation` from `/transit/cache-config` to improve the output of the OpenApi Spec.

Verified OAS Spec doesn't advertise the `x-vault-createSupported` in OpenApi spec. Also verified the transit/cache-config `POST` still works. 

See this gh issue [#12329](https://github.com/hashicorp/vault/issues/12329)


digivava (approved):
Makes sense to me and it sounds like you tested that POST still works, so I'll approve but as discussed let's also get another non-devex engineer on here for visibility.

HridoyRoy (approved):

PR #14704, opened 2022-03-24
Vault-4010 Unauthenticated panic when processing "help" requests

akshya96:
Approved PR on enterprise: https://github.com/hashicorp/vault-enterprise/pull/2556
Jira: https://hashicorp.atlassian.net/browse/VAULT-4010
The Vault server's wrapHelpHandler function incorrectly processes the requests sent to the Vault server. It assumes that all request URL paths start with the "/v1/" string, and when they do not, it panics. This panic is then recovered by the Go's HTTP server and while the server does not crash, this process is likely resource exhaustive and can be used for a denial of service attack.
Solution: Fix the Vault's help request processing, so it doesn't panic when the request URL path does not start with the "/v1/" prefix. Validate it before calling the handleHelp function. This will prevent a possibility of triggering an unauthenticated panic in a Vault server which could be used for a denial of service attack.

### DIFF
--- a/changelog/14704.txt
+++ b/changelog/14704.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Fix panic for help request URL paths without /v1/ prefix   
+```
--- a/http/help.go
+++ b/http/help.go
@@ -1,7 +1,9 @@
 package http
 
 import (
+	"errors"
 	"net/http"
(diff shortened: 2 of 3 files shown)


ncabatoff (approved):
Don't forget the changelog please.

PR #14717, opened 2022-03-24
Vault documentation: added new vault lambda extension doc 

taoism4504:
Per Jim's request, a new vault lambda extension doc was added. An additional request to restructure the AWS left-nav was also addressed.

:mag: [Deploy Preview](https://vault-git-new-vle-caching-doc-hashicorp.vercel.app/docs/platform/aws/lambda-extension-cache)

image

PR #14719, opened 2022-03-24
Suggestion to split the commands into individual code block

yhyakuna:
🔍 [Deploy Preview](https://vault-git-format-fixes-hashicorp.vercel.app/docs/platform/aws/lambda-extension-cache)

This PR suggest some format fixes for [PR 14717](https://github.com/hashicorp/vault/pull/14717).

Split commands into individual code block; otherwise, the copy won't work properly .


taoism4504 (approved):
Thanks!

PR #14733, opened 2022-03-28
builtin/logical/transit: fix clobbered err

alrs:
This fixes an `err` variable passed into a test function but left unused.

### DIFF
--- a/builtin/logical/transit/stepwise_test.go
+++ b/builtin/logical/transit/stepwise_test.go
@@ -194,6 +194,9 @@ func testAccStepwiseDecrypt(
 		Path:      "decrypt/" + name,
 		Data:      decryptData,
 		Assert: func(resp *api.Secret, err error) error {
+			if err != nil {
+				return err
+			}
 			var d struct {
 				Plaintext string `mapstructure:"plaintext"`
 			}


alrs:
If I could, I'd add the `no-changelog` label.

builtin/logical/transit/stepwise_test.go
```diff
@@ -194,6 +194,9 @@ func testAccStepwiseDecrypt(
 		Path:      "decrypt/" + name,
 		Data:      decryptData,
 		Assert: func(resp *api.Secret, err error) error {
+			if err != nil {
```
stevendpclark, line 197: While I agree, this is probably the fix that makes the most sense, this test is now incongruent with all the other tests that we ignore the passed in error into the Assert function. 

Could we instead rename the err variable on line 208 or apply this fix to all the other tests within this file?

stevendpclark (changes requested):
Hi @alrs, 

Thanks for the PR submission! I have a minor comment on the fix that you propose, let me know your thoughts.

stevendpclark:
As this PR is fairly old now with no feedback on the requested changes, I'll close it out. 

Please feel free to re-open the PR if you feel strongly about the fix.

PR #14744, opened 2022-03-28
Ensure that URL encoded passwords are properly redacted.

benashz:
The URL password redaction operation did not handle the case where the
database connection URL was provided as a percent-encoded string, and
its password component contained reserved characters. It attempted to
redact the password by replacing the unescaped password in the
percent-encoded URL. This resulted in the password being revealed when
reading the configuration from Vault.

Given the follow URL configured in the DB secret engine:

`sqlserver://sa:yourStrong%28%21%29Password@127.0.0.1:55002`

On Vault read we would get:

```
vault read --format json tf-test-db-279243129191764308/config/db-6338640021812042769 | \
  jq -r .data.connection_details.connection_url
sqlserver://sa:yourStrong%28%21%29Password@127.0.0.1:55004
```

Which is unquoted to reveal the password:

```
vault read --format json tf-test-db-279243129191764308/config/db-6338640021812042769 | \
jq .data.connection_details.connection_url | python3 -c 'from urllib import parse; import sys; print(parse.unquote(sys.stdin.read()))'
"sqlserver://sa:yourStrong(!)Password@127.0.0.1:55004"
```

### DIFF
--- a/builtin/logical/database/backend_test.go
+++ b/builtin/logical/database/backend_test.go
@@ -3,7 +3,9 @@ package database
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"log"
+	"net/url"
 	"os"
 	"reflect"
 	"strings"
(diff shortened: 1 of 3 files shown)


swenson (approved):
LGTM;

I also like the fact that we aren't blindly replacing the password with `*****` now, just in case the password appears elsewhere in the URL.

tomhjp (approved):

PR #14745, opened 2022-03-28
UI/Implementing Design System color tokens

hellobontempo:
### Steps to use color token ds variables: 

- Include`@use 'tokens'` at the top of the `.scss` file 

- Color tokens can be accessed via `var(--token-color-css-variable-name)`

- Find the color from the design system [webpage](https://design-system-components-hashicorp.vercel.app/foundations/tokens) 

- Clicking "copy" automatically copies it in the correct format (wrapped by `var()`)
 > On clipboard after clicking 'copy' below: `var(--token-color-palette-neutral-500)`
Screen Shot 2022-03-28 at 2 11 30 PM


ui/app/styles/components/calendar-widget.scss
```diff
@@ -69,17 +67,17 @@ $dark-gray: #535f73;
       border-radius: 2px;
     }
     &.is-current-month {
-      border: 1px solid $ui-gray-900;
```
hellobontempo, line 72: substituted all instances of `ui-color-###` with their `readable-name` counterpart

ui/app/styles/utils/_colors.scss
```diff
@@ -1,17 +1,22 @@
+@use 'tokens';
 // Colors that will eventually be in Structure package
 
+// Design System Colors: https://design-system-components-hashicorp.vercel.app/foundations/tokens
+$token-black: var(--token-color-vault-brand-alt); // black #000000
```
hellobontempo, line 5: Thoughts on prefixing any design system tokens, not previously defined, with `$token-` ? or would we like it to be more specific `$ds-token-` 

Open to any other ideas!

PR #14746, opened 2022-03-28
add value length check to approle createHMAC

ccapurso:
Prevent clients from issuing approle login requests with unbounded input requiring SHA256 HMAC calculation by adding a limit check.


hghaf099 (approved):
LGTM

PR #14747, opened 2022-03-28
Remove Unknown Arguments From Builtin Components

zofskeez:
As of Ember 4.0, legacy arguments will be removed from builtin components (LinkTo, Input and Textarea). We still had a few templates where the lint rule was turned off to allow usage of some of the arguments. This PR removes those arguments along with the lint rule overrides.

https://deprecations.emberjs.com/v3.x/#toc_ember-built-in-components-legacy-arguments
https://deprecations.emberjs.com/v3.x/#toc_ember-built-in-components-legacy-attribute-arguments

### DIFF
--- a/ui/app/templates/components/cluster-info.hbs
+++ b/ui/app/templates/components/cluster-info.hbs
@@ -4,20 +4,18 @@
       {{#if (and this.activeCluster.unsealed this.auth.currentToken)}}
         {{#if @cluster.dr.isSecondary}}
           {{#if (has-permission "status" routeParams="replication")}}
-            <nav class="menu">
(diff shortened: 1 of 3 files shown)


ui/app/templates/components/cluster-info.hbs
```diff
@@ -78,22 +69,20 @@
         {{/if}}
       {{/if}}
     {{/unless}}
-    <nav class="menu">
+    <nav class="menu" aria-label="server menu">
```
Monkeychip, line 72: thanks for tackling these while you were in there.

Monkeychip (approved):
Nice work!

PR #14748, opened 2022-03-29
Ui/Remove [at] from date formatter

hellobontempo:
Left over syntax from when we used moment.js was showing a weird timestamp in the UI (`[at]` is not supported in date-fns.)

Removed "at" together following our standard practice for showing dates, since we don't typically include

**before fix:**
Screen Shot 2022-03-17 at 10 54 41 AM

**after**
Screen Shot 2022-03-28 at 5 09 31 PM


zofskeez (approved):
Good catch! Thanks.

Issue #14418, opened 2022-03-09
Move certificate extensions to entity alias metadata


MochaCaffe:
**Is your feature request related to a problem? Please describe.**
@peterverraedt submitted a PR #13348 to add TLS certificates ASN1 extensions as identity metadata in Vault when using the cert auth method. 
This would enable using certificates metadata in ACL templating to set a proper policy scope for each machine when accessing secrets.
However, it looks like extensions are not registered at all in Vault. It is expected that these should be set in the metadata property of an entity. 

When looking at an entity used by a client using cert method:
```bash
vault read identity/entity/id/<id> | grep metadata

Key                    Value
---                    -----
metadata               <nil>
```
  

**Describe the solution you'd like**
I propose to save certificate metadata in the alias instead of using the parent entity.
In addition to solving this issue, I think using the alias is a more coherent approach, as it is closely linked with the certificate unlike the parent entity which is more global to the user
Related PR: #14419

hsimon-hashicorp:
Hi @MochaCaffe! We discussed this request in our engineering team sync today. While the request is reasonable, we'd like to take some time to dig into the potential implementation details and validate that there won't be the potential for unintended consequences. As a result, it may take longer for a final decision to be made, whether it's that we accept a PR from the community, or handle the work internally as part of an upcoming release, or if the decision ends up being that it would be safer (from a functionality, compatibility, or security point of view) not to take this request on. We will link this request internally and work with product and engineering leaders to come to a decision. Thanks in advance for your patience! :)

HridoyRoy:
Hi @MochaCaffe ,
Thanks for submitting this issue! The change makes sense to us as the right path forward for enabling using certificates metadata in ACL templating. However, we're worried about storage concerns regarding this change, so we'd like for this to be an opt-in feature.

MochaCaffe:
Hi @HridoyRoy , currently, only cert metadata whose key is defined in the field "Allowed metadata extensions"  in the auth method are stored. By default, this field is empty, letting the feature acting as an "opt-in"

HridoyRoy:
> Hi @HridoyRoy , currently, only cert metadata having their keys present in the field "Allowed metadata extensions" in the auth method are stored. By default, this field is empty, letting the feature act as an "opt-in"

[...]

PR #14751, opened 2022-03-29
auth/cert: Add metadata to identity-alias

peterverraedt:
Add metadata to logical.Alias (the identity alias) too, in addition
to the metadata added to logical.Auth. This is analoguous to the
behaviour of the ldap and approle auth providers.

CC: @pmmukh 

Fixes: #14418

Signed-off-by: Peter Verraedt <peter.verraedt@kuleuven.be>


hghaf099:
Thank you for your contribution. There has been an update on the original GH issue. I am going to post a link here for reference. https://github.com/hashicorp/vault/issues/14418#issuecomment-1090500929

HridoyRoy (changes requested):
Hi @peterverraedt ,
Thanks for submitting this PR! 

This change makes sense to us as the right path forward for enabling using certificates metadata in ACL templating. However, we're worried about storage concerns regarding this change, so we'd like for this to be an opt-in feature. One option is to add this to the role config, but it might also be possible that an extra config endpoint is needed. 
Thanks again!

peterverraedt:
@HridoyRoy I've added a config value to the existing configuration endpoint of the auth method mount. As the identity aliases are linked to the auth mount as a whole rather than to the specific role, this seems to be the most logical place (either all identity aliases have some metadata, or none have metadata).

The amount of stored metadata is limited to these five fields <https://github.com/hashicorp/vault/blob/c5cf53fccb1923e2f17aacc100b6ba50881c23ee/builtin/credential/cert/path_login.go#L98-L102> and what the user specifies as "allowed_metadata_extensions". The latter is specified on role level as it could easily differ depending on the trusted CAs.

peterverraedt:
@HridoyRoy What is the current status of this PR?

pmmukh:
Hi @peterverraedt ! I'm currently taking a look at this PR, and i'll drop a proper review shortly, but off the bat I had a couple asks, if they make sense to you.
1. Since the goal is to use the certificate info for policy templating, could we potentially have a test that exercises that code path i.e store a secret in a path with say the cert common name, create a policy with that template, and check that a token with the policy works to read the secret ?
2. I think this PR should also remove the place where Metadata is currently being set i.e in auth.Metadata, as we can't use it for templating since its in the token metadata, and so I don't think there's a practical way of using it. wdyt ?

peterverraedt:
Hi @pmmukh, back from holidays now.

1. Yes, that is a good idea, I'll try to implement during the next days.
2. I'm following what is done in the ldap and approle auth providers: <https://github.com/hashicorp/vault/blob/main/builtin/credential/approle/path_login.go#L339> and <https://github.com/hashicorp/vault/blob/main/builtin/credential/ldap/path_login.go#L90>. The auth.Metadata is used in the metadata associated with the token that the user will receive after login, I am not sure whether it is merged with the alias.Metadata there? But I guess it could be removed, it is up to you.

peterverraedt:
Hi @pmmukh, I added a test case for testing whether usage of metadata in acl policies works.

I can also confirm for your second question that the metadata that is set in Auth.Metadata ends up as metadata associtated with the token, and while this does not relate at all to the acl policy templates, it still might be useful.

HridoyRoy (approved):
👍 lgtm!

PR #14752, opened 2022-03-29
Vault-4279 reporting redundant/unused keys in config

hghaf099:
The Vault server does not warn about configurations that contain invalid configuration keys.


command/server.go
```diff
@@ -458,6 +458,11 @@ func (c *ServerCommand) runRecoveryMode() int {
 		JSONFormat: logFormat == logging.JSONFormat,
 	})
 
+	// reporting Errors found in the config
+	for _, cErr := range configErrors {
+		c.logger.Error(cErr.String())
```
hghaf099, line 463: I used `Error` log level here. Should it be `warning` instead?

sgmiller: I would warn, yes, because the server may be able to continue.

sgmiller (changes requested):
Otherwise nice addition.

swayne275 (approved):

ccapurso (approved):

sgmiller (approved):

Issue #11722, opened 2021-05-28
Support passing certificate PEM string to API Client


lawliet89:
**Is your feature request related to a problem? Please describe.**

The [API Client](https://github.com/hashicorp/vault/blob/9e66901630e7c671039957f4dfdf0d91a6ac35e5/api/client.go#L216-L224) makes use of [`go-rootcerts`](https://github.com/hashicorp/go-rootcerts) to configure TLS for `HttpClient`. `go-rootcerts` supports passing in a PEM encoded string. 

This is useful for API client usage such as in the Terraform Vault provider where the CA string might be passed from other data/resources.

**Describe the solution you'd like**

Add a new `CACertificate` `[]byte` attribute andpass that to `rootcerts.Config`.

**Describe alternatives you've considered**

Writing PEM string to a temporary file.

**Explain any additional use-cases**
If there are any use-cases that would help us understand the use/need/value please share them as they can help us decide on acceptance and prioritization.

**Additional context**

If this is acceptable, I can take a stab at implementing it.

vishalnayak:
@lawliet89 I don't feel against this. Feel free to take a shot at implementing this. I feel most of the effort will be in plumbing the PEM string till that point. Let us know how it goes!

lawliet89:
Sure. I'll take a look. I think we can leave it to the user of the API client to figure out how to plumb the `[]byte` to the library.

hsimon-hashicorp:
Hi @lawliet89 - have you had any luck with this? If you need assistance, please feel free to ping back and we'll help you out! Thanks!

lawliet89:
Sorry, I've not had the time to implement this. If anyone wants to take it on, please feel free. I'll update if/when I implement something.

hsimon-hashicorp:
Thanks, @lawliet89 - I'll also check with our product owners to see about potentially roadmapping this ourselves. I can't make any guarantees, so please do feel free to still take a shot at it if you wish!

averche:
Hi @lawliet89, thank you for contributing this! I opened a PR #14753 which follows pretty closely to your description. Please let me know if that's what you had in mind.

lawliet89:
@averche looks great :) Thank you.

averche:
I have merged the fix, thanks for contributing to HashiCorp!

PR #14753, opened 2022-03-29
Add ability to pass certificate PEM bytes to vault/api

averche:
Fixes: #11722

> The [API Client](https://github.com/hashicorp/vault/blob/9e66901630e7c671039957f4dfdf0d91a6ac35e5/api/client.go#L216-L224) makes use of [`go-rootcerts`](https://github.com/hashicorp/go-rootcerts) to configure TLS for `HttpClient`. `go-rootcerts` supports passing in a PEM encoded string. 
> 
> This is useful for API client usage such as in the Terraform Vault provider where the CA string might be passed from other data/resources.

This PR will add the ability to pass certificate PEM bytes (on top of the existing file & directory entry).


VinnyHC (approved):
This looks good to me!

PR #14763, opened 2022-03-29
Ember upgrade to 3.28.6

Monkeychip:
Ember upgrade to 3.28 LTS

### Do not stub router service in tests
The following pattern results in test failures in 3.28:
```js
const routerService = Service.extend({
  urlFor() {
    return 'http://example.com';
  }
});
this.owner.register('service:router', routerService);
```
Instead, override the properties/method(s) on the service:
```js
this.owner.lookup('service:router').reopen({
  urlFor() {
    return 'http://example.com';
  }
});
```


ui/app/models/policy.js
```diff
@@ -4,7 +4,6 @@ import { computed } from '@ember/object';
 import lazyCapabilities, { apiPath } from 'vault/macros/lazy-capabilities';
 
 export default Model.extend({
-  errors: attr('array'),
```
zofskeez, line 7: `errors` is an internal property on the model class and overriding it was throwing errors on initial load of the policy model.

ui/app/mixins/policy-edit-controller.js
```diff
@@ -27,18 +27,13 @@ export default Mixin.create({
...
-        .save()
-        .then((m) => {
-          flash.success(`${policyType.toUpperCase()} policy "${name}" was successfully saved.`);
-          if (this.wizard.featureState === 'create') {
-            this.wizard.transitionFeatureMachine('create', 'CONTINUE', policyType);
-          }
-          return this.transitionToRoute('vault.cluster.policy.show', m.get('policyType'), m.get('name'));
-        })
-        .catch((e) => {
-          model.set('errors', e.errors);
```
zofskeez, line 40: The `errors` prop on Model is a getter now and attempting to set it throws an error. In this case the adapter sets the errors automatically on failure so this is not necessary.

ui/app/models/policy.js
hellobontempo: I had added [this](https://github.com/hashicorp/vault/pull/14551/files) because glimmerizing `MessageError` for some reason no longer triggered error messages, so I manually added it to the component. 

Definitely makes sense that overriding would cause issues -and in the `policy/create.hbs` file from the linked PR above, `this.model.errors` should still behave, right?

zofskeez: Yes I tested and the error message is still displayed. The `errors` property is set by the Adapter when there is a network request failure.

ui/app/components/secret-edit-metadata.js
```diff
@@ -54,6 +54,7 @@ export default class SecretEditMetadata extends Component {
     if (value) {
       if (name === 'customMetadata') {
         // cp validations won't work on an object so performing validations here
+        // JLR TODO: review this and incorporate into model-validations system
```
hashishaw, line 57: @zofskeez is this a TODO for this PR or future?

ui/lib/core/addon/components/form-field.js
```diff
@@ -31,7 +31,8 @@ import layout from '../templates/components/form-field';
 export default Component.extend({
   layout,
   'data-test-field': true,
-  classNames: ['field'],
+  attributeBindings: ['data-test-field'],
+  classNames: ['field-foo'],
```
hashishaw, line 35: Is this intentional, to keep `field-foo`? or is this a placeholder for something to be overwritten?

ui/lib/core/stories/toggle-button.md
```diff
@@ -1,24 +1,20 @@
...
 
+**Params**
 
 | Param | Type | Default | Description |
 | --- | --- | --- | --- |
-| toggleAttr | <code>String</code> | <code></code> | The attribute upon which to toggle. |
-| attrTarget | <code>Object</code> | <code></code> | The target upon which the event handler should be added. |
-| [openLabel] | <code>String</code> | <code>Hide</code> | options  - The message to display when the toggle is open. |
-| [closedLabel] | <code>String</code> | <code>More</code> | options  - The message to display when the toggle is closed. |
+| isOpen | <code>boolean</code> |  | determines whether to show open or closed label |
```
hashishaw, line 9: This interface is so much clearer 🙏

ui/tests/acceptance/oidc-auth-method-test.js
```diff
@@ -52,7 +52,7 @@ module('Acceptance | oidc auth method', function (hooks) {
   });
 
   test('it should login with oidc from listed auth mount tab', async function (assert) {
-    assert.expect(2);
+    assert.expect(3); // request is fired more than once -- so three assertions instead of two
```
hashishaw, line 55: Can we add a TODO to update the actual behavior, if it's firing more than expected?

hashishaw (approved):
Really great work on this!! One step closer to Ember 4.0 🚀 
Just a couple small comments, and then I think it looks good to go

ui/lib/core/addon/components/form-field.js
zofskeez: Hmm that can't be right. Maybe a hold over from some testing but the class should be `field`. Good catch!

ui/tests/acceptance/oidc-auth-method-test.js
zofskeez: I think this is from `fetchRole` being triggered in the `didReceiveAttrs` hook of the `auth-jwt` component. When the component is instantiated it fetches the role and then again when the auth method tab is clicked (triggers on re-render). `didUpdateAttrs` is probably better to use in this case but then again there may be a better approach when the component is glimmerized. I'll add a comment to the test to capture this.

PR #14771, opened 2022-03-29
Address semgrep 0.86.x breakage

stevendpclark:
Filed a bug against semgrep as they no longer seem to process the following code snippet, failing with the error 
`At line xxxx:1: (Failure "Found unexpected raw string literal without delimiters")`

```
package main

func main() {
	if err := func() error {
		return nil
	}(); err != nil {
		panic(err)
	}
}

const val = `
test literal
`
```

So fix up our codebase to not use that pattern and pin us to the latest version of semgrep 0.86.5

### DIFF
--- a/.circleci/config/commands/setup-semgrep.yml
+++ b/.circleci/config/commands/setup-semgrep.yml
@@ -7,7 +7,7 @@ steps:
       name: Setup Semgrep
       command: |
         apk add --no-cache python3 py3-pip make
-        python3 -m pip install --user semgrep
+        python3 -m pip install --user semgrep==0.86.5
         export PATH="$HOME/.local/bin:$PATH"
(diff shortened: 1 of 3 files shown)


.circleci/config/commands/setup-semgrep.yml
```diff
@@ -7,7 +7,7 @@ steps:
       name: Setup Semgrep
       command: |
         apk add --no-cache python3 py3-pip make
-        python3 -m pip install --user semgrep
+        python3 -m pip install --user semgrep==0.86.5
```
benashz, line 10: 👍

benashz (approved):
LGTM!

ncabatoff (approved):

Issue #14724, opened 2022-03-25
Use context in Kubernetes authentication


nicolascb:
**Is your feature request related to a problem? Please describe.**

Today the Login method via Kubernetes receives a context but does not use it: [api/auth/kubernetes/kubernetes.go#L70](https://github.com/hashicorp/vault/blob/main/api/auth/kubernetes/kubernetes.go#L70)

**Describe the solution you'd like**

Using the latest version of the api, we can solve this problem like this:

[api/auth/kubernetes/kubernetes.go#L77](https://github.com/hashicorp/vault/blob/main/api/auth/kubernetes/kubernetes.go#L77)
```diff
-       resp, err := client.Logical().Write(path, loginData)
+       resp, err := client.Logical().WriteWithContext(ctx, path, loginData)
```

**Describe alternatives you've considered**
N/A

**Explain any additional use-cases**
N/A

averche:
Hi @nicolascb!

Thank you for this suggestion! Now that we have `WithContext` methods, this change is quite easy. I have a PR that takes care of all of our current auth helpers in on shot: #14775

averche:
I have merged the PR. Thank you for contributing to HashiCorp!

PR #14725, opened 2022-03-25
Use context with Kubernetes Auth

nicolascb:
Fix #14724

PR #14775, opened 2022-03-29
Use WriteWithContext in auth helpers

averche:
Fixes: #14724

Propagate the context passed to the auth helpers by using WriteWithContext function

### DIFF
--- a/api/auth/approle/approle.go
+++ b/api/auth/approle/approle.go
@@ -100,6 +100,10 @@ func NewAppRoleAuth(roleID string, secretID *SecretID, opts ...LoginOption) (*Ap
 }
 
 func (a *AppRoleAuth) Login(ctx context.Context, client *api.Client) (*api.Secret, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
 	loginData := map[string]interface{}{
 		"role_id": a.roleID,
 	}
--- a/api/auth/approle/approle.go
+++ b/api/auth/approle/approle.go
@@ -125,7 +125,7 @@ func (a *AppRoleAuth) Login(ctx context.Context, client *api.Client) (*api.Secre
 
 	// if the caller indicated that the value was actually a wrapping token, unwrap it first
 	if a.unwrap {
-		unwrappedToken, err := client.Logical().Unwrap(secretIDValue)
+		unwrappedToken, err := client.Logical().UnwrapWithContext(ctx, secretIDValue)
 		if err != nil {
 			return nil, fmt.Errorf("unable to unwrap response wrapping token: %w", err)
 		}
@@ -135,7 +135,7 @@ func (a *AppRoleAuth) Login(ctx context.Context, client *api.Client) (*api.Secre
 	}
 
 	path := fmt.Sprintf("auth/%s/login", a.mountPath)
-	resp, err := client.Logical().Write(path, loginData)
+	resp, err := client.Logical().WriteWithContext(ctx, path, loginData)
 	if err != nil {
 		return nil, fmt.Errorf("unable to log in with app role auth: %w", err)
 	}
--- a/api/auth/aws/aws.go
+++ b/api/auth/aws/aws.go
@@ -84,6 +84,10 @@ func NewAWSAuth(opts ...LoginOption) (*AWSAuth, error) {
 // variables. To specify a path to a credentials file on disk instead, set
 // the environment variable AWS_SHARED_CREDENTIALS_FILE.
 func (a *AWSAuth) Login(ctx context.Context, client *api.Client) (*api.Secret, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
 	loginData := make(map[string]interface{})
 	switch a.authType {
 	case ec2Type:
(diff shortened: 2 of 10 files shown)


api/auth/approle/approle.go
```diff
@@ -125,7 +125,7 @@ func (a *AppRoleAuth) Login(ctx context.Context, client *api.Client) (*api.Secre
```
digivava, line 125: I think you may have missed the `gcp` submodule

averche: Thanks, good catch 👍

api/auth/aws/aws.go
```diff
@@ -182,7 +182,7 @@ func (a *AWSAuth) Login(ctx context.Context, client *api.Client) (*api.Secret, e
 	}
 
 	path := fmt.Sprintf("auth/%s/login", a.mountPath)
-	resp, err := client.Logical().Write(path, loginData)
+	resp, err := client.Logical().WriteWithContext(ctx, path, loginData)
```
hghaf099, line 185: Hi there, I stumbled in this PR when I was looking at https://github.com/hashicorp/vault/pull/14725 
Figured to add a comment here. It seems that the context in these functions is not used before this change. I think it is better to check if the ctx is nil first, and if so, assign a background context to it before passing it to the `WriteWithContext`.  This prevents a possible panic downstream in the context package.
```
if ctx == nil {
		ctx = context.Background()
}
```

averche: That's a good point. However, I don't think it's very common (or correct) pass `nil` to any function that expects context. Would it be more idiomatic to return an error in this case?

```go
if ctx == nil {
	return nil, fmt.Errorf("nil context")
}
```

This is similar to what the go standard library is doing in [net/http RequestWithContext](https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/net/http/request.go;l=866-868)

hghaf099: Well, I am not sure this practice is uncommon, especially when dealing with api code. There are numerous examples of that pattern. Example:
https://github.com/googleapis/google-api-go-client/blob/9d43e7dc08e2598d4dcb661324c4877dfd63c4f9/gmail/v1/gmail-gen.go#L3835

If we were dealing with the server code, I would agree that a nil context should be treated with returning an error. However, this is an api code, and unfortunately, the context has not been used before in there. 

With regards to the net/http package, what I understand is that the function requires a non-nil context by design and it is not necessarily a pattern to follow everywhere. What do you think?

averche: Looking at the gmail example, I see that the context is not passed directly to the method, rather it is initialized in a different method altogether [here](https://github.com/googleapis/google-api-go-client/blob/9d43e7dc08e2598d4dcb661324c4877dfd63c4f9/gmail/v1/gmail-gen.go#L3749-L3752). In that code it seems reasonable to check if the context was initialized for the object since the caller may have forgotten to set it.

In the `Login` methods, `ctx context.Context` is an explicit function parameter (effectively part of the function contract). To me, it seems like the `Login` functions should not work with a `nil` context and should require one (either return an error or panic). However, I'm not familiar with the pattern of nil-checking the context that was passed to the function so I'm trying to find the most idiomatic and common solution, particularly for API wrappers. Do you have other examples that we can look at?

averche: After some thinking and internal discussions, I'm going to go with your original suggestion just to keep the function fully backwards compatible. We can consider removing the extra nil check in future iterations of this library.

hghaf099: Actually, you raised very good point. I am just interested in learning about idiomatic ways of writing code as well, as I am fairly new to GO. I found this example that is kind of interesting. The ctx is passed in the func, and a nil check is done to decide what to do with the request:
https://github.com/googleapis/google-api-go-client/blob/9d43e7dc08e2598d4dcb661324c4877dfd63c4f9/internal/gensupport/send.go#L61

https://github.com/googleapis/google-api-go-client/blob/9d43e7dc08e2598d4dcb661324c4877dfd63c4f9/internal/gensupport/send.go#L27

command/agent/auth/auth.go
```diff
@@ -172,7 +172,7 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
 			ah.logger.Debug("lookup-self with preloaded token")
 			clientToUse.SetToken(ah.token)
 
-			secret, err = clientToUse.Logical().Read("auth/token/lookup-self")
+			secret, err = clientToUse.Logical().ReadWithContext(ctx, "auth/token/lookup-self")
```
VinnyHC, line 175:
```suggestion
			secret, err = clientToUse.Auth().Token().LookupSelfWithContext(ctx)
```

VinnyHC: Wondering if we should introduce our library where we can?

VinnyHC (approved):
This looks good to me. I added a small suggestion around moving away from `Logical()` but I don't have strong feelings one way or another.

api/auth/aws/aws.go
averche: From what I've seen, the most common way of handling contexts passed directly to a function is to just assume that it is never `nil`. Any `nil` contexts would cause a panic once the context is actually used. Here is another example from standard library where context is not nil checked: [database/sql](https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/database/sql/sql.go;l=1842).

However, since the Login functions never actually used the context, it seems OK to add a nil check for backwards compatibility. If these were new functions, I would probably lean towards not checking the context for nil at all.

command/agent/auth/auth.go
averche: Ah, good catch!

PR #14791, opened 2022-03-30
Fixing excessive unix file permissions

hghaf099:
I also ran https://github.com/securego/gosec#github-action

### DIFF
--- a/physical/raft/raft.go
+++ b/physical/raft/raft.go
@@ -274,7 +274,7 @@ func EnsurePath(path string, dir bool) error {
 	if !dir {
 		path = filepath.Dir(path)
 	}
-	return os.MkdirAll(path, 0o755)
+	return os.MkdirAll(path, 0o750)
 }
 
 // NewRaftBackend constructs a RaftBackend using the given directory
--- a/physical/raft/raft.go
+++ b/physical/raft/raft.go
@@ -274,7 +274,7 @@ func EnsurePath(path string, dir bool) error {
 	if !dir {
 		path = filepath.Dir(path)
 	}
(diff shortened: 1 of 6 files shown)


physical/raft/raft.go
```diff
@@ -274,7 +274,7 @@ func EnsurePath(path string, dir bool) error {
 	if !dir {
 		path = filepath.Dir(path)
 	}
-	return os.MkdirAll(path, 0o755)
+	return os.MkdirAll(path, 0o750)
```
akshya96, line 277: Is there a reason why this is 0o750 instead of 0o700?

hghaf099: I ran this tool  https://github.com/securego/gosec#github-action and the maximum allowed permission that it suggested was 0o750. So, I went with that. Would you think that 0o700 is better?

ccapurso: I think my question to follow up the original question would be is there a reason that we want to allow group read and execute access? It might be safest to only allow owner access but I be missing some context.

mickael-hc: These are permissions for a directory, and this would allow non-vault user to traverse and list the directory contents. While using `0o750` should only allow the disclose the filenames in the directory (which are likely not sensitive), it might make sense to remain consistent with other folder permissions, and remove group read/execute bits, as mentioned above (similar to https://github.com/hashicorp/vault/blob/main/sdk/physical/file/file.go#L239)

hghaf099: I guess 0o700 follows the least privilege rule the best. Done

akshya96 (approved):

PR #14796, opened 2022-03-30
Starter PKI CA Storage API

stevendpclark:
Extremely basic model and storage layer to persist a pki key and pki issuer.

### DIFF
--- a/builtin/logical/pki/storage.go
+++ b/builtin/logical/pki/storage.go
@@ -10,63 +10,65 @@ import (
 )
 
 const (
-	pkiKeyPrefix    = "/config/key/"
-	pkiIssuerPrefix = "/config/issuer/"
+	keyPrefix    = "/config/key/"
+	issuerPrefix = "/config/issuer/"
 )
 
 var (
-	emptyPkiKey = pkiKey{}
-	emptyIssuer = pkiIssuer{}
+	emptyKey        = key{}
+	emptyIssuer     = issuer{}
+	emptyCertBundle = certutil.CertBundle{}
 )
 
-type pkiKeyId string
+type keyId string
 
-func (p pkiKeyId) String() string {
+func (p keyId) String() string {
 	return string(p)
 }
 
-type pkiIssuerId string
+type issuerId string
 
-func (p pkiIssuerId) String() string {
+func (p issuerId) String() string {
 	return string(p)
 }
 
-type pkiKey struct {
-	ID             pkiKeyId                `json:"id" structs:"id" mapstructure:"id"`
+type key struct {
+	ID             keyId                   `json:"id" structs:"id" mapstructure:"id"`
 	PrivateKeyType certutil.PrivateKeyType `json:"private_key_type" structs:"private_key_type" mapstructure:"private_key_type"`
 	PrivateKey     string                  `json:"private_key" structs:"private_key" mapstructure:"private_key"`
 }
 
-type pkiIssuer struct {
-	ID           pkiIssuerId `json:"id" structs:"id" mapstructure:"id"`
-	PKIKeyID     pkiKeyId    `json:"pki_key_id" structs:"pki_key_id" mapstructure:"pki_key_id"`
-	Certificate  string      `json:"certificate" structs:"certificate" mapstructure:"certificate"`
-	CAChain      []string    `json:"ca_chain" structs:"ca_chain" mapstructure:"ca_chain"`
-	SerialNumber string      `json:"serial_number" structs:"serial_number" mapstructure:"serial_number"`
+type issuer struct {
+	ID           issuerId `json:"id" structs:"id" mapstructure:"id"`
+	Name         string   `json:"name" structs:"name" mapstructure:"name"`
+	KeyID        keyId    `json:"key_id" structs:"key_id" mapstructure:"key_id"`
+	Certificate  string   `json:"certificate" structs:"certificate" mapstructure:"certificate"`
+	CAChain      []string `json:"ca_chain" structs:"ca_chain" mapstructure:"ca_chain"`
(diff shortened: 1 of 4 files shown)


builtin/logical/pki/storage.go
```diff
@@ -0,0 +1,104 @@
...
+	pkiKeyPrefix    = "/config/key/"
+	pkiIssuerPrefix = "/config/issuer/"
+)
+
+var (
+	emptyPkiKey = pkiKey{}
+	emptyIssuer = pkiIssuer{}
+)
+
+type pkiKeyId string
```
cipherboy, line 22: The argument I've generally seen for eliding the `pki` name from the variables is that you're already in the `pki` package, so that's implied. Would there be conflicts if we drop the pki prefix everywhere? Just curious :-)

builtin/logical/pki/storage.go
```diff
@@ -0,0 +1,104 @@
...
+func writePKIKey(ctx context.Context, s logical.Storage, pkiKey pkiKey) error {
+	keyId := pkiKey.ID
+
+	json, err := logical.StorageEntryJSON(pkiKeyPrefix+keyId.String(), pkiKey)
+	if err != nil {
+		return err
+	}
+
+	return s.Put(ctx, json)
+}
```
cipherboy, line 75: I think we also want a list on both keys and issuers I think :-)

builtin/logical/pki/storage.go
stevendpclark: Yeah I was really torn, I could definitely drop the pki references a bit everywhere within this package. Will do.

builtin/logical/pki/storage.go
stevendpclark: Yup! I was gonna handle the list and name association next.

cipherboy: Cool, ty! :-)

builtin/logical/pki/storage.go
```diff
@@ -0,0 +1,130 @@
...
+
+func fetchCertBundleByIssuerId(ctx context.Context, s logical.Storage, issuerId issuerId) (certutil.CertBundle, error) {
+	issuer, err := fetchIssuerById(ctx, s, issuerId)
+	if err != nil {
+		return emptyCertBundle, err
+	}
+
+	if issuer.KeyID == "" {
```
cipherboy, line 114: Interesting, I had taken the other approach of allowing the bundle to be created, but then (in `fetchCAInfo`) validating the key there.

stevendpclark: Since so much of the existing code expects a private key to exist within a certbundle that would be a lot of places we would need to add that validation to no?

cipherboy: I think we'd only need to add it to `fetchCAInfo` from what I can tell, which is used in the signing path. In the non-signing path (future `/issuer` work), I think we won't need to validate that.

cipherboy: Here's what I had for what its worth:

```diff 
+ func fetchIssuerBundle(ctx context.Context, s logical.Storage, id pkiIssuer, loadKey bool) (*certutil.CertBundle, error) {
+       var bundle certutil.CertBundle
+
+       issuer, err := fetchPKIIssuerById(ctx, s, id)
+       if err != nil {
+               return nil, err
+       }
+
+       bundle.Certificate = issuer.Certificate
+       bundle.IssuingCA = issuer.IssuingCA
+       bundle.CAChain = issuer.CAChain
+       bundle.SerialNumber = issuer.SerialNumber
+
+       // Fetch the key if it exists. Sometimes we don't need the key immediately.
+       var key pkiKey
+       if loadKey && issuer.PKIKeyID != pkiKeyId("") {
+               key, err = fetchPKIKeyById(ctx, s, id)
+               if err != nil {
+                       return nil, err
+               }
+
+               bundle.PrivateKeyType = key.PrivateKeyType
+               bundle.PrivateKey = key.PrivateKey
+       }
+
+       return &bundle, nil
+}
```

I thought returning pointer was also cleaner. My 2c :-)

stevendpclark: How about two methods instead of the boolean (it's a bit of a pet peeve and harder to identify who does what). As for the pointers sure I can switch as I suppose that is the standard in Vault.

stevendpclark: Thinking about it a little more let's go with your implementation. I'll remove mine from my PR for now.

cipherboy: I've added it to my PR: https://github.com/hashicorp/vault/pull/14804

builtin/logical/pki/storage.go
```diff
@@ -0,0 +1,176 @@
...
+
+	json, err := logical.StorageEntryJSON(issuerPrefix+issuerId.String(), issuer)
+	if err != nil {
+		return err
+	}
+
+	return s.Put(ctx, json)
+}
+
+func setKeysConfig(ctx context.Context, s logical.Storage, config map[string]keyId) error {
```
cipherboy, line 132: Maybe make it `interface{}` ? We might have things like defaults or key restrictions in the future here, that we'd want to also store in the config.

builtin/logical/pki/storage.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	return s.Put(ctx, json)
+}
+
+func getKeysConfig(ctx context.Context, s logical.Storage) (map[string]keyId, error) {
+	keyConfigEntry, err := s.Get(ctx, keyConfig)
+	if err != nil {
+		return nil, err
+	}
```
cipherboy, line 145: I think you need a `keyConfigEntry == nil` check and return an empty config (no error). This shouldn't happen on migration, but aligns us with the `getConfig` from storage methods (`url` or `crl`).

PR #14803, opened 2022-03-30
sdk/useragent: adds optional ability to set additional comments

austingebauer:
This PR adds the ability to set additional comments in Vault's user agent string from the `useragent` package. The new comments parameter is variadic, so it shouldn't break API compatibility for the package.

The motivation for this change is to clean up code in plugins that need to add additional comments. For example, we currently add partner IDs in [vault-plugin-auth-azure](https://github.com/hashicorp/vault-plugin-auth-azure/blob/main/util.go#L54) and [vault-plugin-secrets-azure](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/provider.go#L118) for usage attribution. Without this change, adding the additional comments requires assumptions about the format and string replacement.

I've improved the tests by turning them into table tests and adding additional coverage.

### DIFF
--- a/sdk/helper/useragent/useragent.go
+++ b/sdk/helper/useragent/useragent.go
@@ -26,22 +26,39 @@ var (
 // String returns the consistent user-agent string for Vault.
 //
 // e.g. Vault/0.10.4 (+https://www.vaultproject.io/; go1.10.1)
-func String() string {
-	return fmt.Sprintf("Vault/%s (+%s; %s)",
-		versionFunc(), projectURL, rt)
+//
+// Given comments will be appended to the semicolon-delimited comment section.
+//
+// e.g. Vault/0.10.4 (+https://www.vaultproject.io/; go1.10.1; comment-0; comment-1)
+func String(comments ...string) string {
+	ua := fmt.Sprintf("Vault/%s (+%s; %s", versionFunc(), projectURL, rt)
+
+	for _, c := range comments {
+		ua = fmt.Sprintf("%s; %s", ua, c)
+	}
+
+	return fmt.Sprintf("%s)", ua)
 }
 
 // PluginString is usable by plugins to return a user-agent string reflecting
 // the running Vault version and an optional plugin name.
 //
 // e.g. Vault/0.10.4 (+https://www.vaultproject.io/; azure-auth; go1.10.1)
-func PluginString(env *logical.PluginEnvironment, pluginName string) string {
+//
+// Given comments will be appended to the semicolon-delimited comment section.
+//
+// e.g. Vault/0.10.4 (+https://www.vaultproject.io/; azure-auth; go1.10.1; comment-0; comment-1)
+func PluginString(env *logical.PluginEnvironment, pluginName string, comments ...string) string {
 	var name string
-
 	if pluginName != "" {
 		name = pluginName + "; "
 	}
 
-	return fmt.Sprintf("Vault/%s (+%s; %s%s)",
-		env.VaultVersion, projectURL, name, rt)
+	ua := fmt.Sprintf("Vault/%s (+%s; %s%s", env.VaultVersion, projectURL, name, rt)
+
+	for _, c := range comments {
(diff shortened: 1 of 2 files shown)


sdk/helper/useragent/useragent.go
```diff
@@ -26,22 +26,39 @@ var (
...
+//
+// Given comments will be appended to the semicolon-delimited comment section.
+//
+// e.g. Vault/0.10.4 (+https://www.vaultproject.io/; go1.10.1; comment-0; comment-1)
+func String(comments ...string) string {
+	ua := fmt.Sprintf("Vault/%s (+%s; %s", versionFunc(), projectURL, rt)
+
+	for _, c := range comments {
+		ua = fmt.Sprintf("%s; %s", ua, c)
+	}
```
kalafut, line 38: Maybe can use https://pkg.go.dev/strings#Join ?

austingebauer: Good idea. Done in ae8f4ba.

austingebauer:
I'd guess this doesn't need a changelog entry. Let me know if you'd prefer for me to add one.

sdk/helper/useragent/useragent.go
```diff
@@ -26,22 +27,39 @@ var (
...
+// e.g. Vault/0.10.4 (+https://www.vaultproject.io/; go1.10.1; comment-0; comment-1)
+func String(comments ...string) string {
+	c := strings.Join(comments, "; ")
+	if len(c) > 0 {
+		c = "; " + c
+	}
+
+	return fmt.Sprintf("Vault/%s (+%s; %s%s)",
+		versionFunc(), projectURL, rt, c)
 }
```
calvn, line 42: Total nit here but perhaps we could use `strings.Join` to construct the entire string here, something like:
```suggestion
	comment = append([]string{projectURL, rt}, comment...)
	c := strings.Join(comment, "; ")

	return fmt.Sprintf("Vault/%s (+%s)", versionFunc(), c)
}
```

calvn (approved):
One optional suggestion, but otherwise looks good :)

sdk/helper/useragent/useragent.go
calvn: Forgot to mention, but similar for `PluginString` as well.

austingebauer: Ya, I wrote up something using your suggested approach as an alternative. I'd be happy to change things if you think that's better.

Here is what I wrote for `String()`, which is similar to your suggestion:
```
func String(comments ...string) string {
	c := append([]string{"+" + projectURL, rt}, comments...)
	return fmt.Sprintf("Vault/%s (%s)", versionFunc(), strings.Join(c, "; "))
}
```

For `PluginString()`, the `pluginName` could be empty, so it doesn't end up being as clean. I wonder if it's important to preserve the order of the comments in case something depends on it. For example, a server could be splitting/indexing on the semicolons. The placement of `c = append(c, rt)` below preserves the original order where `pluginName` comes first in the comments. 
```
func PluginString(env *logical.PluginEnvironment, pluginName string, comments ...string) string {
	c := []string{"+" + projectURL}
	if pluginName != "" {
		c = append(c, pluginName)
	}
	c = append(c, rt)
	c = append(c, comments...)
	return fmt.Sprintf("Vault/%s (%s)", env.VaultVersion, strings.Join(c, "; "))
}
```

Let me know your thoughts!

tomhjp: I slightly prefer the versions in your comments. It's slightly cleaner to read and it's nice that the comments are all logically grouped together in a way that aligns very closely with the [RFC spec](https://httpwg.org/specs/rfc7231.html#header.user-agent).

tomhjp (approved):
Seems like a good addition 👍

sdk/helper/useragent/useragent.go
austingebauer: +1 - I like that all of the comments are grouped together. I've changed this in e8df12f.

PR #14804, opened 2022-03-31
Handle resolving key, issuer references

cipherboy:
The API context will usually have a user-specified reference to the key.
This is either the literal string "default" to select the default key,
an identifier of the key, or a slug name for the key. Here, we wish to
resolve this reference to an actual identifier that can be understood by
storage.

Also adds the missing Name field to keys.

---

This depends on four unwritten methods and thus can't be built yet:

```
builtin/logical/pki/storage.go:83:18: undefined: getKeysConfig
builtin/logical/pki/storage.go:91:15: undefined: listKeys
builtin/logical/pki/storage.go:151:18: undefined: getIssuersConfig
builtin/logical/pki/storage.go:159:18: undefined: listIssuers
```

In `fetchCAInfo`, we'd call `resolveIssuerReference` and then call `fetchCertBundleByIssuerId`.

### DIFF
--- a/builtin/logical/pki/storage.go
+++ b/builtin/logical/pki/storage.go
@@ -10,63 +10,65 @@ import (
 )
 
 const (
-	pkiKeyPrefix    = "/config/key/"
-	pkiIssuerPrefix = "/config/issuer/"
+	keyPrefix    = "/config/key/"
+	issuerPrefix = "/config/issuer/"
 )
 
 var (
-	emptyPkiKey = pkiKey{}
-	emptyIssuer = pkiIssuer{}
+	emptyKey        = key{}
+	emptyIssuer     = issuer{}
+	emptyCertBundle = certutil.CertBundle{}
 )
 
-type pkiKeyId string
+type keyId string
 
-func (p pkiKeyId) String() string {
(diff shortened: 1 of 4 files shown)


builtin/logical/pki/storage.go
```diff
@@ -76,6 +77,45 @@ func writeKey(ctx context.Context, s logical.Storage, key key) error {
...
+			return keyId("key-read"), err
+		}
+
+		if key.Name == reference {
+			return key.ID, nil
+		}
+	}
+
+	// Otherwise, we must not have found the key.
+	return keyId("not-found"), fmt.Errorf("unable to find PKI key for reference: %v", reference)
```
cipherboy, line 116: This should probably be a UserError along with the one below?

stevendpclark: Possibly, again it's a convention thing where all the existing codebase within pki only checked for two error types in error handling code (InternalError and UserError). I *believe* I fixed all those handlers but there were a surprising number of them.

cipherboy:
Manually merged. :-)

cipherboy:

PR #14805, opened 2022-03-31
Make fetchCAInfo aware of new storage layout

cipherboy:
This allows fetchCAInfo to fetch a specified issuer, via a reference
parameter provided by the user. We pass that into the storage layer and
have it return a cert bundle for us. Finally, we need to validate that
it truly has the key desired.

---

Just about everything (except for root deletion/generation which checks the `ca_bundle` storage entry directly) correctly uses `fetchCAInfo`; I think later work will be updating the calls to pass the desired issuer. But assuming we can create the bundle correctly (at the storage level) and find the right bundle (from API request), I think we should be in good shape.

### DIFF
--- a/builtin/logical/pki/cert_util.go
+++ b/builtin/logical/pki/cert_util.go
@@ -91,27 +91,29 @@ func getFormat(data *framework.FieldData) string {
 
 // Fetches the CA info. Unlike other certificates, the CA info is stored
 // in the backend as a CertBundle, because we are storing its private key
-func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request) (*certutil.CAInfoBundle, error) {
-	bundleEntry, err := req.Storage.Get(ctx, "config/ca_bundle")
+func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRef string) (*certutil.CAInfoBundle, error) {
+	id, err := resolveIssuerReference(ctx, req.Storage, issuerRef)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


stevendpclark (approved):

cipherboy:
Manually merged :-) Thanks @stevendpclark!

builtin/logical/pki/cert_util.go
```diff
@@ -91,28 +91,30 @@ func getFormat(data *framework.FieldData) string {
 
 // Fetches the CA info. Unlike other certificates, the CA info is stored
 // in the backend as a CertBundle, because we are storing its private key
-func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request) (*certutil.CAInfoBundle, error) {
-	bundleEntry, err := req.Storage.Get(ctx, "config/ca_bundle")
+func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRef string) (*certutil.CAInfoBundle, error) {
+	id, err := resolveIssuerReference(ctx, req.Storage, issuerRef)
```
kitography, line 95: I'm not clear that this structure can work as with discussion of today about functioning, that this structure of get reference + below fetchCertBundleByIssuerId would have to be different to enable "get the old one"

kitography (commented):
Mostly I'm really happy this is small, but as of dicussion today, I think it does need to change.

builtin/logical/pki/cert_util.go
cipherboy: Yeah, I think migration concerns need to be addressed more concisely in one place :/

cipherboy:

PR #14808, opened 2022-03-31
Show no config warning error bug fix

Monkeychip:
[This PR correctly](https://github.com/hashicorp/vault/pull/14551) set the conditional to only show this warning at the right time, but something with `mountModel` always returned false when called before the save.

I also fixed a test that should have been catching this 🤦🏻‍♀️  (a test I wrote and incorrectly used ok when I should have been checking for the text).


arnav28 (approved):
Nice 👏

PR #14809, opened 2022-03-31
Replace RadioButton With Internal Component

zofskeez:
During the Ember upgrade to 3.28 there were deprecation warnings coming from the `ember-radio-button` addon prompting that the latest version be installed. Since we are using the latest version and it's likely that the addon is no longer actively maintained this PR removes the dependency and replaces it with an internal component as part of the `Core` addon.

### DIFF
--- a/ui/app/templates/components/raft-join.hbs
+++ b/ui/app/templates/components/raft-join.hbs
@@ -34,7 +34,6 @@
         <RadioButton
           id="join"
           name="setup-pref"
-          class=""
           @value="join"
           @groupValue={{this.preference}}
           @onChange={{fn (mut this.preference)}}
--- a/ui/app/templates/components/raft-join.hbs
+++ b/ui/app/templates/components/raft-join.hbs
@@ -30,28 +30,29 @@
         <br />
         How do you want to get started?
       </h2>
-      <RadioButton
-        @value="join"
-        @groupValue={{this.preference}}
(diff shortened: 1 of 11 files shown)


ui/app/templates/components/raft-join.hbs
```diff
@@ -30,28 +30,29 @@
...
-        @changed={{action (mut this.preference)}}
-        @name="setup-pref"
-        @radioId="join"
-        @classNames="vlt-radio is-block"
-      >
+      <div class="vlt-radio is-block">
+        <RadioButton
+          id="join"
+          name="setup-pref"
+          class=""
```
hellobontempo, line 37: does `class` need to be here if it's empty? I noticed it wasn't in the component below

ui/tests/integration/components/radio-button-test.js
```diff
@@ -0,0 +1,69 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { click, render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
+module('Integration | Component | radio-button', function (hooks) {
```
hellobontempo, line 6: 👏

hellobontempo (approved):
LGTM! Thanks for tackling 🚀

PR #14812, opened 2022-03-31
add kv write retry logic for TestKVPatchCommand_RWMethodPolicyVariations

ccapurso:
A KV backend that is undergoing an upgrade from v1 to v2 will be temporarily unavailable. Any requests to it will be responded to with an error `Upgrading from non-versioned to versioned data. This backend will be unavailable for a brief period and will resume service shortly.` There are test helpers that will look for a substring of this error in the response and retry the request if necessary. There is a KV write performed for the test cases defined `TestKVPatchCommand_RWMethodPolicyVariations` that was not using a retry helper func.


hghaf099 (approved):
LGTM

PR #14767, opened 2022-03-29
Add extra test coverage to PKI 

stevendpclark:
* Add a test that validates the behavior of the delete role api
 * Add a test for the crl/rotate api 
 * Rework existing test for fetch raw certs to no longer write directly to storage, instead use the proper apis 
 * Add test for ca_chain fetch api (returning nothing on root)

PR #14813, opened 2022-03-31
Refactor raw PKI fetch endpoints

cipherboy:
Various endpoints (`/ca`, `/crl`, `/ca_chain`, `/certs`) all use the same core
handling logic with a complicated per-path detection logic of which to
return. Refactor out the common response formatting code, but let each
API handler be distinct to provide the right storage reference.

This refactors just the raw path (`/ca`, `/crl`, `/ca_chain`, and
`/cert/*/raw{,/pem}`).

---

This should be rebased and merged after #14767 probably. We'll want a second version for the non-raw endpoints, but this at least takes a step in the right direction.

### DIFF
--- a/builtin/logical/pki/path_fetch.go
+++ b/builtin/logical/pki/path_fetch.go
@@ -17,7 +17,7 @@ func pathFetchCA(b *backend) *framework.Path {
 		Pattern: `ca(/pem)?`,
 
 		Callbacks: map[logical.Operation]framework.OperationFunc{
-			logical.ReadOperation: b.pathFetchRead,
+			logical.ReadOperation: b.pathFetchCARawHandler,
 		},
 
 		HelpSynopsis:    pathFetchHelpSyn,
@@ -25,10 +25,22 @@ func pathFetchCA(b *backend) *framework.Path {
 	}
 }
 
+func (b *backend) pathFetchCARawHandler(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+	var serial, pemType, contentType string
+	serial = "ca"
+	contentType = "application/pkix-cert"
+	if req.Path == "ca/pem" {
+		pemType = "CERTIFICATE"
+		contentType = "application/pem-certificate-chain"
+	}
+
+	return b.pathFetchReadRaw(ctx, req, data, serial, pemType, contentType)
+}
+
 // Returns the CA chain
 func pathFetchCAChain(b *backend) *framework.Path {
 	return &framework.Path{
-		Pattern: `(cert/)?ca_chain`,
+		Pattern: `cert/ca_chain`,
 
 		Callbacks: map[logical.Operation]framework.OperationFunc{
 			logical.ReadOperation: b.pathFetchRead,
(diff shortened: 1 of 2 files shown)


builtin/logical/pki/path_fetch.go
```diff
@@ -66,14 +111,27 @@ hyphen-separated octal`,
...
-			logical.ReadOperation: b.pathFetchRead,
+			logical.ReadOperation: b.pathFetchCertificateRawHelper,
 		},
 
 		HelpSynopsis:    pathFetchHelpSyn,
 		HelpDescription: pathFetchHelpDesc,
 	}
 }
 
+func (b *backend) pathFetchCertificateRawHelper(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
```
kitography, line 122: is there a reason that this is a `Helper` and everything else is a `Handler`?

builtin/logical/pki/path_fetch.go
```diff
@@ -17,18 +17,30 @@ func pathFetchCA(b *backend) *framework.Path {
...
+}
+
 // Returns the CA chain
 func pathFetchCAChain(b *backend) *framework.Path {
 	return &framework.Path{
-		Pattern: `(cert/)?ca_chain`,
+		Pattern: `cert/ca_chain`,
 
 		Callbacks: map[logical.Operation]framework.OperationFunc{
 			logical.ReadOperation: b.pathFetchRead,
```
kitography, line 46: Why is this one in particular not using a helper?  I'm not sure why the serial is still set in the switch case statement there.

kitography, line 46: Why is this one in particular not using a helper?  I'm not sure why the serial is still set in the switch case statement there.

builtin/logical/pki/path_fetch.go
```diff
@@ -315,6 +333,79 @@ reply:
...
+	response = &logical.Response{
+		Data: map[string]interface{}{
+			logical.HTTPContentType: contentType,
+			logical.HTTPRawBody:     certificate,
+		},
+	}
+
+	if retErr != nil {
+		if b.Logger().IsWarn() {
+			b.Logger().Warn("possible error, but cannot return in raw response. Note that an empty CA probably means none was configured, and an empty CRL is possibly correct; call /crl/rotate to create a non-empty CRL", "error", retErr)
```
kitography, line 394: Could we not use this to set a non 2** status code?

kitography, line 394: Could we not use this to set a non 2** status code?

builtin/logical/pki/path_fetch.go
```diff
@@ -315,6 +333,79 @@ reply:
 	return
 }
 
+func (b *backend) readCertByAlias(ctx context.Context, req *logical.Request, alias string) ([]byte, error) {
```
kitography, line 336: this seems to duplicate what is already in pathFetchRead: is that right?

kitography, line 336: this seems to duplicate what is already in pathFetchRead: is that right?

kitography (approved):
I have comments, but this is 100% better than what it was, so whatever.

builtin/logical/pki/path_fetch.go
cipherboy: I was only handling the non-JSON ("raw") responses in this PR. :-) This endpoint (`cert/ca_chain`) returns a JSON blob, so I didn't touch it on this first pass.

builtin/logical/pki/path_fetch.go
cipherboy: Yes; if we later refactor the JSON endpoints, we'll refactor `pathFetchRead` and likely use this logic instead of the logic already in it. 

I didn't really see value in updating `pathFetchRead` to take out this piece of logic (avoiding the duplication) for the time being, since the bigger refactor of JSON endpoints would probably impact how we structure it anyways.

That's my 2c. at any rate.

builtin/logical/pki/path_fetch.go
cipherboy: Oh yeah, that'd be a good idea.

cipherboy:
This approach won't work, because the Framework logic doesn't understand paths created this way. Closing this.

PR #14816, opened 2022-03-31
Remove ember-basic-dropdown-hover Addon

zofskeez:
While upgrading to Ember 3.28, `ember-basic-dropdown-hover` raised a deprecation warning and it was suggested to upgrade to the latest version which we were already on and dates back to 2019. Since it appears that it is no longer actively maintained the dependency was removed and the ToolTip component was refactored to add the logic for opening and closing the dropdown on mouse enter and leave.


Monkeychip (approved):
Thanks for tackling!

PR #14822, opened 2022-04-01
To cope with new k8s CSR api spec

imcom:
ref: https://github.com/kubernetes/kubernetes/issues/99504

The modification is necessary to complete the tls setup with newer k8s cluster >=1.22


taoism4504:
@hsimon-hashicorp , wanted to confirm that this has been technically reviewed. Thanks~

tvoran:
Thanks for putting this up, though it looks like we updated the docs for 1.22+ in #16029, so I'll close this for now.

PR #14834, opened 2022-04-01
Remove AppID auth backend

ccapurso:
Per Vault's [deprecation docs](https://www.vaultproject.io/docs/deprecation), the AppID auth method has been deprecated and is to be removed in version 1.11. This PR handles the removal of the auth backend and any associated documentation. The feature removal notice remains in the deprecation docs but the deprecation notice in the **Resources** column has been hardcoded to point to version 1.10.x of the docs since it can not use a relative path for the current version of the docs as it has been removed.


hghaf099 (approved):
LGTM

ccapurso:
Moving this back to a draft PR as we will not be removing this just yet.

PR #14835, opened 2022-04-01
KMSE Enable/Config

zofskeez:
This PR addresses some initial phase 2 items for adding the key management secrets engine:
- add key management secrets engine as supported backend (enterprise only)
- disable and show tooltip if license does not include Advanced Data Protection module
- display correct icon for key management secret engine
- remove ttl fields from configuration options
- remove ttl fields from configuration view page

### DIFF
--- a/ui/app/templates/components/mount-backend-form.hbs
+++ b/ui/app/templates/components/mount-backend-form.hbs
@@ -47,7 +47,7 @@
               @onRadioChange={{queue (action (mut this.mountModel.type)) (action "onTypeChange" "type")}}
               @disabled={{if type.requiredFeature (not (has-feature type.requiredFeature)) false}}
               @tooltipMessage={{if
-                (or (eq type.type "transform") (eq type.type "kmip"))
+                (or (eq type.type "transform") (eq type.type "kmip") (eq type.type "keymgmt"))
                 (concat
                   type.displayName
                   " is part of the Advanced Data Protection module, which is not included in your enterprise license."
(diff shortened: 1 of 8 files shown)


ui/app/templates/components/mount-backend-form.hbs
```diff
@@ -47,7 +47,7 @@
               @onRadioChange={{queue (action (mut this.mountModel.type)) (action "onTypeChange" "type")}}
               @disabled={{if type.requiredFeature (not (has-feature type.requiredFeature)) false}}
               @tooltipMessage={{if
-                (or (eq type.type "transform") (eq type.type "kmip"))
+                (or (eq type.type "transform") (eq type.type "kmip") (eq type.type "keymgmt"))
                 (concat
                   type.displayName
                   " is part of the Advanced Data Protection module, which is not included in your enterprise license."
```
hashishaw, line 53: We should double-check that this is true -- not sure which module KMSE is actually in

hashishaw (approved):
Looks great, thank you!

hashishaw:
We may want to re-name this PR to be indicative of what this is (not the entire phase two, but enable + config)

ui/app/templates/components/mount-backend-form.hbs
zofskeez: I'm not sure where to find that out. I can add a TODO comment so we can track that down before we finish phase 2.

Issue #14638, opened 2022-03-22
LIfetime Watcher Ignores "increment" when computing grace period. 


vlaurenzano:
**Describe the bug**

In lifetime_watcher the grace period is calculated with the secret's lease duration, not the given "increment" value. 

https://github.com/hashicorp/vault/blob/da0155ba197e3dc9f50695f49c52288a18ba8f1e/api/lifetime_watcher.go#L260

This means that for long lived leases, a client cannot submit it's own increment value that is less than the leases grace period without triggering and immediate return (without sleeping). 

https://github.com/hashicorp/vault/blob/da0155ba197e3dc9f50695f49c52288a18ba8f1e/api/lifetime_watcher.go#L364

In my case, the initial lease duration is 12 hours, but I’m trying to check the secrete every 10 minutes (in case of a revocation). The grace period for 12 hours ends up being something near 2 hours which is greater than my increment value, forcing the return before the time.After call.

**Expected behavior**
Grace period should be calculated using the minimum value of the lease duration and the client increment value. 

**Additional context**

Discussion forum: https://discuss.hashicorp.com/t/vault-sdk-lifetime-watcher-client-increment-and-secret-lease-duration/36850

If there is agreement from a maintainer I'm willing to make the update.

swayne275:
Thanks for raising this! This seems like a bug. We'll queue it for investigation.

averche:
Hi @vlaurenzano,

Thank you very much for opening this issue! I was able to successfully reproduce it and have a fix ready in #14836 based on your suggestions! I will run it by a few more people who are more familiar with the code and the "increment" behavior before merging. Thank you again for contributing to HashiCorp!

PR #14836, opened 2022-04-01
Respect increment value in grace period calculations (api/LifetimeWatcher)

averche:
Fixes:  #14638

### Background 

`api.LifetimeWatcher` is a tool that helps to monitor and periodically renew expiring token or credential leases. It determines a grace period based on a given lease duration and attempts to renew the lease within the grace period. The callers of LifetimeWatcher can optionally specify `Increment` value in seconds, which is sent to the vault server and is meant to override the token's TTL value, if respected by the server.

### The issue

> This means that for long lived leases, a client cannot submit it's own increment value that is less than the leases grace period without triggering and immediate return (without sleeping).

### Change

This PR will make sure that the grace period is calculated based on the minimum of the remaining lease duration & interval (if specified).

### How was this tested

I added a `short_increment_duration` test renewer_test.go.

Additionally, I was able to successfully reproduce the issue in [hello-vault-go](https://github.com/hashicorp/hello-vault-go/tree/main/sample-app). Before the fix, setting an increment value lower than the token's actual TTL resulted in an early renewal and the following error every time renew was called:

```
URL: PUT http://vault-server:8200/v1/sys/wrapping/unwrap
Code: 400. Errors:

* wrapping token is not valid or does not exist. If the AppRoleAuth struct was initialized with the WithWrappingToken LoginOption, then the secret ID's filepath should be a path to a response-wrapping token
```

When I built against this branch, the error disappeared and the renew period was updated to the `increment` value minus grace period as expected.


VinnyHC (approved):
Nice! This looks good to me.

digivava (approved):
Good stuff! I'm approving because I think this makes sense, but just checking one thing: In the [docs](https://www.vaultproject.io/docs/concepts/lease#lease-durations-and-renewal) it says "[The increment value] is not an increment at the end of the current TTL; it is an increment from the current time." 

That doesn't change anything here, right? I don't think it does but just double-checking.

PR #14843, opened 2022-04-01
Handle Importing Key, Issuers to new PKI storage layout

cipherboy:
This PR is in two parts:

 - Small (non-exhaustive) refactoring of `certutil` to expose useful functionality.
 - Addition of two new methods, `importKey` and `importCert` to the PKI storage code from #14796.

Notably, importing a key or issuer is roughly the same:

 - Iterate all existing keys/issuers and check to see if we have a match; return it if we do.
 - Create a new entry instead.
    - Generate a UUID, compute field values.
    - Now, we need to iterate over all issuers/keys and see if there's a missing KeyID value we can assign.
 - Write the result to storage and return.

I'm sure there's more refactoring of `certutil` we could do with this if we wanted. And we definitely need some tests. :-) But wanted to get your thoughts before I get too far.

One thing I'm worried about is what do we do about failures in the detection loop? Should we still write the issuer/key (we don't in `importIssuer` but do in `importKey`)? In `importKey`, what do we do if the first write succeeds (e.g., writing the first `issuer`) but the second one fails (e.g., writing the second `issuer`)? Do we attempt to record which issuers we touched and revert it?

### DIFF
--- a/builtin/logical/pki/storage.go
+++ b/builtin/logical/pki/storage.go
@@ -2,8 +2,13 @@ package pki
 
 import (
 	"context"
+	"crypto"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
+	"strings"
 
+	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/vault/sdk/helper/certutil"
 	"github.com/hashicorp/vault/sdk/helper/errutil"
 	"github.com/hashicorp/vault/sdk/logical"
@@ -52,6 +57,20 @@ type issuerConfig struct {
 	DefaultIssuerId issuerId `json:"default" structs:"default" mapstructure:"default"`
 }
 
+func genKeyIssuerUUID() (string, error) {
+	newId, err := uuid.GenerateUUID()
+	if err != nil {
+		return "", err
+	}
+
+	return newId, nil
+}
+
+func (k key) GetSigner() (crypto.Signer, error) {
+	signer, _, err := certutil.ParsePEMKey(k.PrivateKey)
+	return signer, err
+}
+
 func listKeys(ctx context.Context, s logical.Storage) ([]keyId, error) {
 	strList, err := s.List(ctx, keyPrefix)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


sdk/helper/certutil/helpers.go
```diff
@@ -170,43 +210,11 @@ func ParsePEMBundle(pemBundle string) (*ParsedCertBundle, error) {
...
-			if parsedBundle.PrivateKeyType != UnknownPrivateKey {
-				return nil, errutil.UserError{Err: "more than one private key given; provide only one private key in the bundle"}
-			}
-			parsedBundle.PrivateKeyFormat = ECBlock
-			parsedBundle.PrivateKeyType = ECPrivateKey
+		if signer, format, err := ParseDERKey(pemBlock.Bytes); err == nil {
+			parsedBundle.PrivateKeyFormat = format
+			parsedBundle.PrivateKeyType = GetPrivateKeyTypeFromSigner(signer)
 			parsedBundle.PrivateKeyBytes = pemBlock.Bytes
 			parsedBundle.PrivateKey = signer
```
stevendpclark, line 217: I believe you need to add a test that GetPrivateKeyTypeFromSigner didn't return UnknownPrivateKey here?

builtin/logical/pki/storage.go
```diff
@@ -52,6 +57,20 @@ type issuerConfig struct {
 	DefaultIssuerId issuerId `json:"default" structs:"default" mapstructure:"default"`
 }
 
+func genKeyIssuerUUID() string {
+	newId, err := uuid.GenerateUUID()
+	if err != nil {
+		panic("uuid generation failed unexpectedly: " + err.Error())
```
stevendpclark, line 63: Seems a little harsh to panic here, even if it dirties up the function sig.

stevendpclark (approved):
Some little nits, nothing that prevents a merge to our feature branch.

builtin/logical/pki/storage.go
cipherboy: I should contribute that to `go-uuid`. But if `io.ReadFull` fails, your entropy source is completely broken and you should really just panic. There's no cryptographic operations you want to do with a busted entropy source.

builtin/logical/pki/storage.go
```diff
@@ -95,6 +114,118 @@ func writeKey(ctx context.Context, s logical.Storage, key key) error {
...
+
+	// Extracting the signer is necessary for two reasons: first, to get the
+	// public key for comparison with existing issuers; second, to get the
+	// corresponding private key type.
+	keySigner, err := result.GetSigner()
+	if err != nil {
+		return nil, false, err
+	}
+	keyPublic := keySigner.Public()
+	result.PrivateKeyType = certutil.GetPrivateKeyTypeFromSigner(keySigner)
```
stevendpclark, line 173: Should we check for UnknownPrivateKey return value here?

builtin/logical/pki/storage.go
```diff
@@ -177,6 +308,100 @@ func writeIssuer(ctx context.Context, s logical.Storage, issuer issuer) error {
...
+			return nil, false, err
+		}
+
+		equal, err := certutil.ComparePublicKeys(issuerCert.PublicKey, signer.Public())
+		if err != nil {
+			return nil, false, err
+		}
+
+		if equal {
+			result.KeyID = existingKey.ID
```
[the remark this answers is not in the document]
cipherboy: Ah, yeah, this one should have a break. The Keys can't have a break.

builtin/logical/pki/storage.go
stevendpclark: Yeah okay nvm for now...

stevendpclark (approved):

builtin/logical/pki/storage.go
```diff
@@ -52,6 +57,20 @@ type issuerConfig struct {
 	DefaultIssuerId issuerId `json:"default" structs:"default" mapstructure:"default"`
 }
 
+func genKeyIssuerUUID() (string, error) {
```
kitography, line 60: I'm not clear why this is a separate function - we need to check "err" anyway each time it is called

kitography: Nevermind, apparently due to needing to panic

kitography (approved):
Looks great

cipherboy:
Manually merged to `pod-pki-rotation` :-)

cipherboy:

PR #14846, opened 2022-04-01
Vault 3999 Change permissions for directory/archive created by debug command 

akshya96:
Jira: https://hashicorp.atlassian.net/browse/VAULT-3999
Approved Ent PR: https://github.com/hashicorp/vault-enterprise/pull/2549
Changing the permissions of the temporary directory and the archive created by the vault debug command, so they are not accessible by other users in the system


hghaf099 (approved):
LGTM

ccapurso (approved):

PR #14847, opened 2022-04-01
Start /issuer-related PKI APIs

cipherboy:
Based on top of #14843; will be rebased once that merges.

This pull request introduces several new APIs to the PKI Engine, based around the concept of an issuer:

 - `LIST /issuers` -- for listing issuers.
 - `GET /issuer/:ref{,/der,/pem}` - for fetching a specific issuer, both as JSON and in raw DER/PEM form. 
 - `POST /issuer/:ref` - for updating issuers (mostly their `name` field right now).
 - `POST /issuers/import/{cert,bundle}` - for importing issuers and keys.

This doesn't update any of the existing code to use the new paths. 

I'm also going to update this with delete code (not written yet) and signing operations as well.

### DIFF
--- a/builtin/logical/pki/path_fetch_issuers.go
+++ b/builtin/logical/pki/path_fetch_issuers.go
@@ -1,0 +1,252 @@
+package pki
+
+import (
+	"context"
+	"encoding/pem"
+	"regexp"
+	"strings"
+
+	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+var nameMatcher = regexp.MustCompile("^" + framework.GenericNameRegex("ref") + "$")
+
+func pathListIssuers(b *backend) *framework.Path {
+	return &framework.Path{
+		Pattern: "issuers/?$",
+
+		Callbacks: map[logical.Operation]framework.OperationFunc{
+			logical.ListOperation: b.pathListIssuersHandler,
+		},
+
+		HelpSynopsis:    pathListIssuersHelpSyn,
+		HelpDescription: pathListIssuersHelpDesc,
+	}
+}
+
+func (b *backend) pathListIssuersHandler(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+	var responseKeys []string
+	responseInfo := make(map[string]interface{})
+
+	entries, err := listIssuers(ctx, req.Storage)
+	if err != nil {
+		return nil, err
+	}
+
+	// For each issuer, we need not only the identifier (as returned by
+	// listIssuers), but also the name of the issuer. This means we have to
+	// fetch the actual issuer object as well.
+	for _, identifier := range entries {
+		issuer, err := fetchIssuerById(ctx, req.Storage, identifier)
+		if err != nil {
+			return nil, err
+		}
+
+		responseKeys = append(responseKeys, string(identifier))
+		responseInfo[string(identifier)] = map[string]interface{}{
+			"name": issuer.Name,
+		}
+	}
+
+	return logical.ListResponseWithInfo(responseKeys, responseInfo), nil
+}
+
+const (
+	pathListIssuersHelpSyn  = `Fetch a list of CA certificates.`
+	pathListIssuersHelpDesc = `
+This endpoint allows listing of known issuing certificates, returning
+their identifier and their name (if set).
+`
+)
+
+func pathGetIssuer(b *backend) *framework.Path {
+	pattern := "issuer/" + framework.GenericNameRegex("ref") + "(/der|/pem)?"
+	return buildPathGetIssuer(b, pattern)
+}
+
+func buildPathGetIssuer(b *backend, pattern string) *framework.Path {
+	return &framework.Path{
+		// Returns a JSON entry.
+		Pattern: pattern,
+		Fields: map[string]*framework.FieldSchema{
+			"ref": {
+				Type:        framework.TypeString,
+				Description: `Reference to issuer; either "default" for the configured default issuer, an identifier of an issuer, or the name assigned to the issuer.`,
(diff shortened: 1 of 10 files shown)


builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+var nameMatcher = regexp.MustCompile("^" + framework.GenericNameRegex("ref") + "$")
+
+func pathListIssuers(b *backend) *framework.Path {
+	return &framework.Path{
+		Pattern: "issuers/?$",
+
+		Callbacks: map[logical.Operation]framework.OperationFunc{
```
stevendpclark, line 19: I believe this is the deprecated way to specify framework.Paths? Could we update to adopt the new way, such as https://github.com/hashicorp/vault/blob/main/builtin/logical/pki/path_root.go#L28-L43

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -0,0 +1,258 @@
...
+			logical.ListOperation: b.pathListIssuers,
+		},
+
+		HelpSynopsis:    pathListIssuersHelpSyn,
+		HelpDescription: pathListIssuersHelpDesc,
+	}
+}
+
+func (b *backend) pathListIssuers(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
```
stevendpclark, line 28: Could we rename to pathListIssuersHandler, using the same function name only separated by a func vs receiver is triggering me ;)

stevendpclark, line 28: Could we rename to pathListIssuersHandler, using the same function name only separated by a func vs receiver is triggering me ;)

builtin/logical/pki/path_fetch_issuers.go
cipherboy: Should we do this afterwards? On `main` I see:

```
[cipherboy@xps15 pki]$ gif 'Operations: ' | wc -l
8
[cipherboy@xps15 pki]$ gif 'Callbacks: ' | wc -l
17
```

which to me says we'd benefit from an all-at-once move?

builtin/logical/pki/backend_test.go
```diff
@@ -2522,7 +2522,7 @@ func TestBackend_SignSelfIssued(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	signingBundle, err := fetchCAInfo(context.Background(), b, &logical.Request{Storage: storage})
+	signingBundle, err := fetchCAInfo(context.Background(), b, &logical.Request{Storage: storage}, "default")
```
stevendpclark, line 2525: Could we define a constant for "default"? Maybe a typed string of some sort. 
Basically please protect me from the inevitable typo I will write ;)

cipherboy: This is actually from #14805 -- if you see what I've done in some of the other commits (`sign-verbatim`, `sign`, `sign-intermediate`, &cp -- you'll see that we're mostly updating it to:

> ```go
> signingBundle, caErr := fetchCAInfo(ctx, b, req, issuerName)
> ```

and `"default"` is the default value of [`addNonCACommonFields`](https://github.com/hashicorp/vault/pull/14847/files#diff-22663f59eb7343dfccf79ce2731482c07b5d867d2a8942a7b81270d34a9c6f91R135)'s `ref` field.

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -0,0 +1,252 @@
...
+
+	issuer, err := fetchIssuerById(ctx, req.Storage, ref)
+	if err != nil {
+		return nil, err
+	}
+
+	if newName != issuer.Name {
+		issuer.Name = newName
+
+		err := writeIssuer(ctx, req.Storage, issuer)
```
stevendpclark, line 156: I have always been torn about this, but since we are passing around pointers to structs, shouldn't we write out a new Issuer object instead of updating an existing one?

cipherboy: Not sure I follow why? I've actually considered restructuring to something like `issuer.Write()`.

If you write out a new object, you've gotta copy _everything_ over, and if the fields change, that seems like a brittle problem waiting to happen. Updating the value seems easiest? Why would pointer versus not matter?

builtin/logical/pki/backend_test.go
stevendpclark: Okay we can adopt a wait and see approach.

builtin/logical/pki/path_fetch_issuers.go
stevendpclark: Keep it as is for now, like I said I'm torn about this, same thing as before about returning pointers to structs vs structs themselves. You raise a descent point about field addition and my counters to that at this time don't really apply to this codebase at the moment.

I'd rather not have the issuer.Write() though that is for sure, just have a method that takes an issuer to write out like we have now.

stevendpclark (approved):

builtin/logical/pki/backend_test.go
cipherboy: Sorry, I've changed the branch of that PR now so its clearer. :D 

 - There's the weird `backend_test.go` one, not sure what I'd do with that one.
 - `crl_util.go` will be handled by the later PR, that needs to be based on revoked cert's actual issuer.
 - `path_fetch.go` is the exception I mentioned above,
 - And both `path_issue_sign` and `path_root.go` were updated in this PR.

stevendpclark (approved):

cipherboy:
Thanks all! Merging as this'll unblock some of @stevendpclark's work... :-)

PR #14851, opened 2022-04-01
Adds Vault version prerelease and metadata to logical.PluginEnvironment

austingebauer:
## Overview

This PR adds [`VersionPrerelease`](https://github.com/hashicorp/vault/blob/main/sdk/version/version.go#L12) and [`VersionMetadata`](https://github.com/hashicorp/vault/blob/main/sdk/version/version.go#L13) information to `logical.PluginEnvironment`. This allows plugins to obtain this information via calls to [`PluginEnv()`](https://github.com/hashicorp/vault/blob/main/sdk/logical/system_view.go#L76).

The motivation for this change is to properly set user-agent headers in both builtin and external plugins. In some cases, we'd like to know if Vault is an enterprise build or not. Including `VersionMetadata` in the plugin environment allows external plugins to do so. Otherwise, the Vault version info in the plugin will be from the SDK version, which may not be accurate.

Related PRs:
- https://github.com/hashicorp/vault-plugin-secrets-azure/pull/93
- https://github.com/hashicorp/vault-plugin-auth-azure/pull/54

## Testing

I tested this change in both Vault and Vault enterprise using builtin and external plugins. Plugins were able to accurately obtain the plugin environment details.


helper/forwarding/types.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.27.1
-// 	protoc        v3.19.3
```
austingebauer, line 4: I can try to downgrade my protoc if this is a problem.

fairclothjm (approved):

PR #14869, opened 2022-04-04
deprecating Legacy MFA

hghaf099:
As mentioned in Vault's [deprecation docs](https://www.vaultproject.io/docs/deprecation), the Legacy MFA feature has been deprecated. It is going to be removed in version 1.11.0. In this PR, the legacy MFA code path is removed along with its documentation.
Addresses https://hashicorp.atlassian.net/browse/VAULT-5682


changelog/14869.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:change
+auth: Remove support for legacy MFA
+(https://www.vaultproject.io/docs/auth/mfa)
```
ccapurso, line 3: This link will likely not exist any longer given that the doc page has been removed, correct? Should we link to a specific version of the docs like in the deprecation table?

ccapurso (approved):
I left one comment regarding the changelog entry but other than that looks good!

changelog/14869.txt
hghaf099: For sure. Thanks for catching that!

Issue #14863, opened 2022-04-03
Consul Connect can't initialize with Vault 1.10.0


voanhduy1512:
**Environment:**

* Vault Version: v1.10.0
* Consul Version: v1.11.4
* Operating System/Architecture: NixOS 21.11/x86_64-linux

**Consul Config File:**

```hcl
connect {
  enabled = true
  ca_provider = "vault"
  ca_config {
    address = "http://localhost:8200"
    token = "root token from vault server -dev  log"
    root_pki_path = "connect_ca"
    intermediate_pki_path = "connect_intermediate"
  }
}
```

**Expected Behavior:**
Consul will  initialize Connect with Vault as CA Provider

**Actual Behavior:**
Consul throws error message and can't initialize Connect CA

```bash
consul agent -dev -config-file consul.hcl

==> Starting Consul agent...
           Version: '1.11.4'
           Node ID: 'df8c57a8-09cb-6d8c-5f40-e3b8637e4948'
         Node name: 'N1'
        Datacenter: 'dc1' (Segment: '<all>')
            Server: true (Bootstrap: false)
       Client Addr: [127.0.0.1] (HTTP: 8500, HTTPS: -1, gRPC: 8502, DNS: 8600)
      Cluster Addr: 127.0.0.1 (LAN: 8301, WAN: 8302)
           Encrypt: Gossip: false, TLS-Outgoing: false, TLS-Incoming: false, Auto-Encrypt-TLS: false

==> Log data will now stream in as it occurs:

2022-04-03T22:36:40.988+0800 [INFO]  agent.server.raft: initial configuration: index=1 servers="[{Suffrage:Voter ID:df8c57a8-09cb-6d8c-5f40-e3b8637e4948 Address:127.0.0.1:8300}]"
2022-04-03T22:36:40.988+0800 [INFO]  agent.server.raft: entering follower state: follower="Node at 127.0.0.1:8300 [Follower]" leader=
2022-04-03T22:36:40.989+0800 [INFO]  agent.server.serf.wan: serf: EventMemberJoin: N1.dc1 127.0.0.1
2022-04-03T22:36:40.989+0800 [INFO]  agent.server.serf.lan: serf: EventMemberJoin: N1 127.0.0.1
2022-04-03T22:36:40.989+0800 [INFO]  agent.router: Initializing LAN area manager
[...]

PR #14875, opened 2022-04-04
Fix handling of default zero SignatureBits value with Any key type in PKI Secrets Engine

cipherboy:
When using `KeyType="any"` on a role (whether explicitly or implicitly
via a `sign-verbatim` like operation), we need to update the value of
`SignatureBits` from its new value 0 to a per-key-type default value. This
will allow sign operations on these paths to function correctly, having
the correctly inferred default signature bit length.

Additionally, this allows the computed default value for key type to be
used for minimum size validation in the RSA/ECDSA paths. We additionally
enforce the 2048-minimum in this case as well.

`Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>

Fix defaults and validation of "any" KeyType

When `certutil` is given the placeholder any keytype, it attempts to
validate and update the default zero value. However, in lacking a
default value for SignatureBits, it cannot update the value from the
zero value, thus causing validation to fail.
    
Add more awareness to the placeholder "any" value to `certutil`.

`Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>

---

Resolves: #14863

This also includes two tests, including one written by @stevendpclark for checking this issue and ensuring we don't regress, one for roles and one for `sign-verbatim`.

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -1953,6 +1953,23 @@ func TestBackend_PathFetchCertList(t *testing.T) {
 }
 
 func TestBackend_SignVerbatim(t *testing.T) {
+	testCases := []struct {
+		testName string
+		keyType  string
+	}{
+		{testName: "RSA", keyType: "rsa"},
+		{testName: "ED25519", keyType: "ed25519"},
+		{testName: "EC", keyType: "ec"},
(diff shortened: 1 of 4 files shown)


stevendpclark (approved):
👍

sgmiller (approved):

PR #14894, opened 2022-04-04
Update Ember Test Selectors for future Ember upgrade

Monkeychip:
We ran into some issues on the latest 6.0.0 version upgrade of this package, so only upgrading to 5.5.0. Confirmed no issue with this.

We need to update this package for the ember 3.28 upgrade we're working on.


Monkeychip:
Done in another PR by Jordan

PR #14899, opened 2022-04-04
Global flag that outputs minimum policy HCL required for an operation

digivava:
Implementation of an idea that was added to the KV path discrepancy discussions by @tomhjp !

This is implemented in the same way as`-output-curl-string`, except instead of producing a curl example, it produces an example of the minimum policy HCL needed to run that command. 

The "sudo" capability is determined by querying the OpenAPI spec endpoint for paths that have "x-vault-sudo" set to true.


api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+		return nil, fmt.Errorf("unable to read response body from OpenAPI: %v", err)
+	}
+	if buf.Len() == 0 {
+		return nil, fmt.Errorf("OpenAPI response had no content")
+	}
+
+	oasInfo := make(map[string]interface{})
+	if err := jsonutil.DecodeJSONFromReader(&buf, &oasInfo); err != nil {
+		return nil, fmt.Errorf("unable to decode JSON from OpenAPI response: %v", err)
+	}
```
VinnyHC, line 153: Since we intend to error on zero-length and `nil` you might be able to skip the buffer and pass in the body:
```go
	if err := jsonutil.DecodeJSONFromReader(resp.Body, &oasInfo); err != nil {
		return nil, fmt.Errorf("unable to decode JSON from OpenAPI response: %v", err)
	}
```

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+func isSudoPath(client *Client, path string) (bool, error) {
+	sudoPaths, err := getSudoPaths(client)
+	if err != nil {
+		return false, fmt.Errorf("unable to retrieve list of paths that require sudo capability: %v", err)
+	}
+	if sudoPaths == nil || len(sudoPaths) < 1 {
+		// OpenAPI spec did not return any paths that require sudo for
+		// some reason, but the user probably still shouldn't see an error.
+		return false, nil
+	}
```
VinnyHC, line 93: I think this could happen if the client's token is restricted.

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+func (d *OutputPolicyError) parseRequest() {
+	capabilities := []string{}
+	switch d.Request.Method {
+	case "GET":
+		capabilities = append(capabilities, "read")
+	case "LIST":
+		capabilities = append(capabilities, "list")
+	case "POST", "PUT":
+		capabilities = append(capabilities, "create")
+		capabilities = append(capabilities, "update")
```
VinnyHC, line 47: Are there operations where we restrict creation but allow updating?

VinnyHC (commented):
This is awesome! Some notes/questions in a few areas. Resolving the `sudo` permission is an interesting problem to solve!

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+	if !ok {
+		return nil, fmt.Errorf("OpenAPI response did not include paths")
+	}
+
+	pathsMap, ok := paths.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("OpenAPI response did not return valid paths")
+	}
+
+	sudoPaths := make(map[string]bool) // this could be a slice, but we're just making it a map so we can do quick lookups for the paths that don't have any templating
```
averche, line 165: The performance of looking things up in a slice is actually pretty good due to cache locality. If you have < 15 elements or so, it will likely outperform a map (and has a nicer interface IMO).

Go 1.18 also adds a convenience function [IndexFunc](https://pkg.go.dev/golang.org/x/exp/slices#IndexFunc) for easier slice lookup. Otherwise, we probably have it implemented somewhere in the code.

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
+package api
```
averche, line 1: Looking at this file overall, it seems like it is using client rather than is a part of client. Would it be more appropriate to have it reside inside `command/` instead?

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+		if d.parsingError != nil {
+			return d.parsingError.Error()
+		}
+	}
+
+	return ErrOutputPolicyRequest
+}
+
+// Builds a sample policy document from the request
+func (d *OutputPolicyError) parseRequest() {
```
averche, line 38: This function is called `parseRequest()` but it actually sneakily makes several requests. It may be worth splitting it up into a couple functions -- one for making the requests and one for accumulating capabilities. You may also want to rename the later to `buildSamplePolicy()` or something similar.

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+func (d *OutputPolicyError) HCLString() string {
+	if d.parsedHCLString == "" {
+		d.parseRequest()
+	}
+	return d.parsedHCLString
+}
+
+// Determine whether the given path requires the sudo capability
+func isSudoPath(client *Client, path string) (bool, error) {
+	sudoPaths, err := getSudoPaths(client)
```
averche, line 85: Do we know how many sudo paths there are? If I understand correctly, there will be (1 + sudoPath) API requests made per call here, which could be quite expensive. If this is a static list of things, we could possibly hard code them or source them from elsewhere.

digivava: I had started with hard-coding but some discussion in the RFC made me reconsider it, as it would be hard to remember to keep it up to date as we add new sudo endpoints.

And can you clarify what you mean here by the expensive multiple API requests? As far as I can tell it's just the one API request, since we're only hitting the OpenAPI endpoint one time and then comparing against the response object after that.

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	return ErrOutputPolicyRequest
+}
+
+// Builds a sample policy document from the request
+func (d *OutputPolicyError) parseRequest() {
+	capabilities := []string{}
+	switch d.Request.Method {
+	case "GET":
```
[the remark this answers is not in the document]
digivava: Good catch on the empty string! I'll add that just in case.

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+			return d.parsingError.Error()
+		}
+	}
+
+	return ErrOutputPolicyRequest
+}
+
+// Builds a sample policy document from the request
+func (d *OutputPolicyError) parseRequest() {
+	capabilities := []string{}
```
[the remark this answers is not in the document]
digivava: You're right, I was mixing this up with maps, where you'll get nil errors if you don't fully create an empty one.

command/main.go
```diff
@@ -264,3 +254,47 @@ func printCommand(w io.Writer, name string, cmdFn cli.CommandFactory) {
...
+		runOpts.Stdout.Write([]byte(fmt.Sprintf("%s\n", api.LastOutputStringError.CurlString())))
+		return 0
+	}
+}
+
+func generatePolicy(exitCode int, runOpts *RunOptions) int {
+	if exitCode == 0 {
+		fmt.Fprint(runOpts.Stderr, "Could not generate policy")
+		return 1
+	} else {
```
tomhjp, line 284: The `else` isn't necessary here: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

command/main.go
```diff
@@ -264,3 +254,47 @@ func printCommand(w io.Writer, name string, cmdFn cli.CommandFactory) {
...
+		if api.LastOutputPolicyError == nil {
+			if exitCode == 127 {
+				// Usage, just pass it through
+				return exitCode
+			}
+			fmt.Fprint(runOpts.Stderr, "Required policy not clear from API operation\n")
+			return exitCode
+		}
+		if api.LastOutputPolicyError.Error() != api.ErrOutputPolicyRequest {
+			runOpts.Stdout.Write([]byte(fmt.Sprintf("Error assembling policy HCL: %s\n", api.LastOutputPolicyError.Error())))
```
tomhjp, line 294:
```suggestion
			runOpts.Stderr.Write([]byte(fmt.Sprintf("Error assembling policy HCL: %s\n", api.LastOutputPolicyError.Error())))
```

tomhjp (commented):
This is awesome! And works great. Thanks for the cc :) just a few suggestions, mainly focussed around the UX for errors/incorrect usage.

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+	// so save any custom value and restore after.
+	currentWrappingLookupFunc := client.CurrentWrappingLookupFunc()
+	client.SetWrappingLookupFunc(nil)
+	defer client.SetWrappingLookupFunc(currentWrappingLookupFunc)
+	currentOutputCurlString := client.OutputCurlString()
+	client.SetOutputCurlString(false)
+	defer client.SetOutputCurlString(currentOutputCurlString)
+	currentOutputPolicy := client.OutputPolicy()
+	client.SetOutputPolicy(false)
+	defer client.SetOutputPolicy(currentOutputPolicy)
```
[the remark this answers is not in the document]
digivava: Great idea, that cleaned this up a lot!

command/main.go
```diff
@@ -264,3 +254,47 @@ func printCommand(w io.Writer, name string, cmdFn cli.CommandFactory) {
...
+	if exitCode == 0 {
+		fmt.Fprint(runOpts.Stderr, "Could not generate policy")
+		return 1
+	} else {
+		if api.LastOutputPolicyError == nil {
+			if exitCode == 127 {
+				// Usage, just pass it through
+				return exitCode
+			}
+			fmt.Fprint(runOpts.Stderr, "Required policy not clear from API operation\n")
```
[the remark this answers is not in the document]
digivava: It does read strangely you're right, I'll change the first part to what you said. 

However, the reason I removed the "run without -output-policy to see the generated error" message I had in there before is because I realized the user often won't be able to run the command without the flag, they'll just get a 403, which may be a red herring in terms of what happened. (Since I imagine people will often be running this flag to see _what_ privileges their token needs in order to get past the 403.)

api/output_policy_test.go
```diff
@@ -0,0 +1,91 @@
...
+					}
+				}
+			}`,
+		))
+	}
+
+	config, ln := testHTTPServer(t, http.HandlerFunc(handler))
+	defer ln.Close()
+
+	config.Address = strings.ReplaceAll(config.Address, "127.0.0.1", "localhost")
```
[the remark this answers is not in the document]
digivava: Good question... I copied that over from all the other tests I saw doing it. But it doesn't seem to be making a difference in whether the test passes here, so I think I'll remove it. 🤔

command/main.go
digivava: Nice catch! Do you think I should change this for -output-curl-string at this time too, or will that possibly break someone's automation?

command/main.go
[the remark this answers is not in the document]
tomhjp: That seems fair. Hopefully we can do better than re-running anyway if the comment about saving stderr works out.

command/main.go
tomhjp: Given the nature of `-output-curl-string`, I think it seems reasonable to assume it's for humans and not machines. And especially given this is only a change in the case that the command fails anyway, it seems safe enough to change for `-output-curl-string` (IMHO).

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+	parsedHCLString string
+}
+
+func (d *OutputPolicyError) Error() string {
+	if d.parsedHCLString == "" {
+		d.parseRequest()
+		if d.parsingError != nil {
+			return d.parsingError.Error()
+		}
+	}
```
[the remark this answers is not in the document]
digivava: As we discussed separately, I agree that this error-based pattern (which I curbed from `-output-curl-string`) is very strange and confusing, but after some time spent learning about how this system works, I feel it's going to take a bit more of a concentrated effort to change it, which is outside the scope of this PR and RFC. 

Right now the consistency it shares with `-output-curl-string` at least makes it so when you understand how one of these flags works, you understand the other.

As discussed, we'll revisit refactoring the -output flags at a different time, likely as part of a larger api client overhaul.

api/output_policy.go
digivava: (see other resolved comment)

command/main.go
```diff
@@ -126,7 +132,7 @@ func RunCustom(args []string, runOpts *RunOptions) int {
 	}
 
 	uiErrWriter := runOpts.Stderr
-	if outputCurlString {
+	if outputCurlString || outputPolicy {
 		uiErrWriter = ioutil.Discard
```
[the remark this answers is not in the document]
digivava: Thanks for this great suggestion! I've now added that, as well as giving a hopefully more helpful error message for the issue that seems to cause this problem 99% of the time.

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+
+	retryablehttp "github.com/hashicorp/go-retryablehttp"
+	"github.com/hashicorp/vault/sdk/helper/jsonutil"
+)
+
+const (
+	ErrOutputPolicyRequest = "output policy request"
+)
+
+var LastOutputPolicyError *OutputPolicyError
```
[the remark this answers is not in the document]
digivava: I think I'll be leaving this for a future bigger refactor as well because currently this variable is relied on by both the api and command packages for the way these flags work. 

At least the docstrings for the OutputCurlString field on the Client struct already says `Note: It is not thread-safe to set this and make concurrent requests with the same client.` which at least maybe helps people be aware of this generally.

api/output_policy.go
digivava: This was a good question that had me investigating for awhile but as discussed offline I think we're good here. As mentioned in the [docs](https://www.vaultproject.io/docs/concepts/policies#capabilities), "Very few parts of Vault distinguish between create and update, so most operations require both create and update capabilities. Parts of Vault that provide such a distinction are noted in documentation.".

I locally modified the OpenAPI spec generation code to show endpoints that support Create and Update respectively, and while there is a substantial number of them that claim to support only Update operations, there isn't a great way for us to programmatically verify this at this point in the code, so worst case scenario I think we will just produce some occasional unused "create" capabilities. 

The backend itself should be rejecting operations that don't actually have an implementation, in any case.

api/output_policy.go
digivava: Thanks for the rename idea!

Based on your other comment I think you may be misreading some of this, which I definitely don't blame you for, because the pattern taken by this and `-output-curl-string` is rather convoluted for sure. The method doesn't make several requests; it only makes one, to the OpenAPI endpoint, to determine the list of paths that require sudo.

Because of that, there didn't seem like a great way to split up the parseRequest method any further than it already is, but I took your naming advice and renamed the whole method to `buildSamplePolicy` (and something similar for -output-curl-string), and also changed a few of the other properties to make their purpose hopefully a little more clear.

Now that I've changed those the whole thing is a little more readable now in my opinion. So thank you!

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		d.parsingError = err
+		return
+	}
+	if needsSudo {
+		capabilities = append(capabilities, "sudo")
+	}
+
+	capStr := strings.Join(capabilities, `", "`)
+	d.parsedHCLString = fmt.Sprintf(
```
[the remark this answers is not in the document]
digivava: Thiank you, this ended up being the comment that finally got me to refactor the whole pattern a bit (for both output flags)! Hooray. 

Just in changing it to return the string I found that storing the error on the object seemed unnecessary as well, so now hopefully the logic is a little simpler as it doesn't need to check Error() in generatePolicy(), it can just check HCLString() and see if that returns an error.

api/output_policy.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+		}
+	}
+
+	return sudoPaths, nil
+}
+
+// Replaces any template fields in a path with the characters ".+" so that
+// we can later allow any characters to match those fields.
+func buildPathRegexp(pathName string) string {
```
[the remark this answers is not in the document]
averche: Ah, I think I understand the code a bit better now, thanks for the detailed explanation!

I think the regex below can still be combined into a single statically-compiled regex:

```go
re := regexp.MustCompile("{path}|{header}|{prefix}|{name}|{type}")
pathWithRegexPatterns = r.ReplaceAllString(pathWithRegexPatterns, ".+")
```

As for constructing a new regex by replacing `{...}` with `.+`, this may run into a few issues. For example, a regex like `a/b/.+` will match not only `a/b/c` but also `a/b/c/d/e/f`, which is probably not what we want here. 

Additionally, the templated name may have characters which would yield the generated regex invalid (e.g. `+`, `*`, '[`, `(`, etc.). All these characters need to be properly escaped if we want to ensure that a valid regex is generated. I would strongly recommend writing a unit test with several test cases to make sure that the escaping logic works.

Perhaps there are other alternatives besides generating regexes (maybe a library that can match the paths as such?)

api/output_policy.go
averche: Sorry, I might have misread the code -- somehow I thought that we were making a new request for each sudo path.

> I had started with hard-coding but some discussion in the RFC made me reconsider it, as it would be hard to remember to keep it up to date as we add new sudo endpoints.

One option to consider is to introduce a unit test which would ensure that your static list of sudo paths matches exactly with all the sudo paths that are defined in OpenAPI spec. This would essentially shift the performance hit from run-time to test time. It would also avoid the pitfalls of regexes mentioned in the other comment.

command/kv_helpers.go
```diff
@@ -64,6 +67,15 @@ func kvPreflightVersionRequest(client *api.Client, path string) (string, int, er
 			return "", 1, nil
 		}
 
+		// if the original request had the -output-curl-string or -output-policy flag,
+		if (currentOutputCurlString || currentOutputPolicy) && resp.StatusCode == 403 {
+			// we provide a more helpful error for the user,
+			// who may not understand why the flag isn't working.
+			err = fmt.Errorf(
+				`For KV commands, this output flag requires the success of a
```
averche, line 75: Do we need to mention "For KV commands" here? I'm assuming this error will only happen for KV commands.

api/output_policy.go
VinnyHC: To further @averche's suggestion of a static list checked during testing, if you choose to go that way,  I believe it would also remove the need to store a client pointer in the error.

api/output_policy.go
```diff
@@ -0,0 +1,183 @@
...
+
+const (
+	ErrOutputPolicyRequest = "output a policy, please"
+)
+
+var LastOutputPolicyError *OutputPolicyError
+
+type OutputPolicyError struct {
+	*retryablehttp.Request
+	VaultClient    *Client
```
VinnyHC, line 22: Does VaultClient need to be exported?

command/main.go
```diff
@@ -126,8 +131,8 @@ func RunCustom(args []string, runOpts *RunOptions) int {
 	}
 
 	uiErrWriter := runOpts.Stderr
-	if outputCurlString {
-		uiErrWriter = ioutil.Discard
+	if outputCurlString || outputPolicy {
+		uiErrWriter = &bytes.Buffer{}
```
averche, line 135: Would changing it from `Discard` to an empty buffer not change the behavior?

command/main.go
```diff
@@ -264,3 +253,55 @@ func printCommand(w io.Writer, name string, cmdFn cli.CommandFactory) {
 	}
 	fmt.Fprintf(w, "    %s\t%s\n", name, cmd.Synopsis())
 }
+
+func generateCurlString(exitCode int, runOpts *RunOptions, preParsingErrCatcher *bytes.Buffer) int {
```
averche, line 257: It might be more idiomatic to specify an interface here as an input (e.g. `preParsingErrCatcher io.Writer`). However, if I understand correctly, this error is supposed to be discarded anyway for `if outputCurlString || outputPolicy {` cases. In this case, would it make sense to just remove this variable altogether?

command/main.go
digivava: The uiErrWriter gets passed into the actual commands, as the ErrorWriter for the VaultUI object on the next line. If an error is encountered during the cli.Run() step but before LastOutputPolicyError gets populated (such as the error encountered by KV commands' preflight request), previously it would just completely discard the error. 

Changing it to a buffer which we write those errors into is how I surface them again later in generatePolicy, as Tom's comment requested.

command/main.go
digivava: I'm passing it as a *bytes.Buffer because I specifically needed the `WriteTo` method.

```
		fmt.Fprint(preParsingErrCatcher, "cURL command not set by API operation\n")
		preParsingErrCatcher.WriteTo(runOpts.Stderr)
```

This is me writing whatever was caught in that buffer (the purpose of which I explain in my other comment) to runOpts.Stderr. If you know a way to do those two lines with just io.Writer.Write lemme know, I'm not super familiar with the various ways to work with buffers so I just went with the first one that I was able to figure out. 😄

command/kv_helpers.go
digivava: Yeah I was on the fence about this one too so I'll remove it.

VinnyHC (approved):
This is fantastic stuff, really nice work piecing all of this together!

command/main.go
averche: I think I can understand the logic a bit better now. In this case, would it be a bit more clear to remove the following block:

```diff
- if outputCurlString || outputPolicy {
-	uiErrWriter = &bytes.Buffer{}
- }
```

then change the `preParsingErrCatcher` to `io.Writer`, and change the lines where we write the errors to just write directly into `preParsingErrCatcher`?

```diff
  fmt.Fprint(preParsingErrCatcher, "cURL command not set by API operation\n")
- preParsingErrCatcher.WriteTo(runOpts.Stderr)
```

If a temporary buffer is required, I would define it within the scope of `generateCurlString`.

digivava: (We resolved this offline but just for posterity), this buffer needs to be defined in the same scope as it is now so it can be passed to the initialization of the CLI commands, then populated inside the command's Run() function, and finally written to stderr in generateCurlString.

It's adding to an already confusing and hacky part of the codebase so we're interested in having this part of both output-curl-string and output-policy in a future refactoring PR, likely as part of the effort to move this kind of functionality out of the `api` package.

averche (approved):
Great job, this will be an awesome feature! 👍

PR #14900, opened 2022-04-04
[Vault-5248] MFA support for api login helpers

VinnyHC:
This PR adds MFA support for auth methods (login helpers).

*Single Phase MFA with Login Helper*
Pass in credentials using same format as `(*client).SetMFACreds()` as variadic  parameter. This is a simple convenience approach around our existing code. 
```go
// existing code
(*client).SetMFACreds(fmt.Sprintf("%s:%s", methodID, passcode))
(*client).Auth().Login(context.Context, AuthMethod)

// PR simplification
(*client).Auth().MFALogin(context.Context, AuthMethod, fmt.Sprintf("%s:%s", methodID, passcode))
```

*Two Phase MFA with Login Helper*
Flow when no credentials are provided. Performs checks on secret to confirm it contains MFARequirements for follow up call.
```go
mfaSecret, err := (*client).Auth().MFALogin(context.Background(), &AuthMethod)
mfaPayload := map[string]interface{}{
  "method-id": []string{"a-passcode"},
}
authSecret, err := (*client).Auth().MFAValidate(context.Background(), mfaSecret, mfaPayload) 
```

This PR also adds the new method `(*client).Sys().MFAValidateWithContext()` a convenience method for calling `sys/mfa/validate`.

### DIFF
--- a/api/auth.go
+++ b/api/auth.go
@@ -51,9 +51,9 @@ func (a *Auth) MFALogin(ctx context.Context, authMethod AuthMethod, creds ...str
 	return a.twoPhaseMFALogin(ctx, authMethod)
 }
 
-// MFAValidate validates an MFA request using the appropriate payload and a secret containing Auth.MFARequirement,
-// like the returned by MFALogin when credentials are not provided. Upon successful validation the client token
-// will be set accordingly.
+// MFAValidate validates an MFA request using the appropriate payload and a secret containing
+// Auth.MFARequirement, like the one returned by MFALogin when credentials are not provided.
+// Upon successful validation the client token will be set accordingly.
 //
 // The Secret returned is the authentication secret, which if desired can be
 // passed as input to the NewLifetimeWatcher method in order to start
--- a/api/auth.go
+++ b/api/auth.go
@@ -50,8 +50,8 @@ func (a *Auth) MFALogin(ctx context.Context, authMethod AuthMethod, creds ...str
 	return a.twoPhaseLogin(ctx, authMethod)
 }
 
-// MFAValidate validates an MFA request using the secret returned by MFALogin where credentials
-// are not provided and the appropriate payload. Upon successful validation the client token
+// MFAValidate validates an MFA request using the appropriate payload and a secret containing Auth.MFARequirement,
+// like the returned by MFALogin when credentials are not provided. Upon successful validation the client token
 // will be set accordingly.
 func (a *Auth) MFAValidate(ctx context.Context, mfaSecret *Secret, payload map[string]interface{}) (*Secret, error) {
 	if mfaSecret == nil || mfaSecret.Auth == nil || mfaSecret.Auth.MFARequirement == nil {
--- a/api/auth.go
+++ b/api/auth.go
@@ -37,22 +37,27 @@ func (a *Auth) Login(ctx context.Context, authMethod AuthMethod) (*Secret, error
 // MFALogin is a wrapper that helps satisfy Vault's MFA implementation.
 // If optional credentials are provided a single-phase login will be attempted
 // and the resulting Secret will contain a ClientToken if the authentication is successful.
-// The client's token will also be set accordingly. If no credentials are provided a
-// two-phase MFA login will be assumed and the resulting Secret will have a
-// MFARequirement containing the MFARequestID to be used in a follow-up call to `sys/mfa/validate`
-// or by passing it to the method (*Auth).MFAValidate with the appropriate payload.
+// The client's token will also be set accordingly.
+//
+// If no credentials are provided a two-phase MFA login will be assumed and the resulting
+// Secret will have a MFARequirement containing the MFARequestID to be used in a follow-up
+// call to `sys/mfa/validate` or by passing it to the method (*Auth).MFAValidate.
 func (a *Auth) MFALogin(ctx context.Context, authMethod AuthMethod, creds ...string) (*Secret, error) {
 	if len(creds) > 0 {
 		a.c.SetMFACreds(creds)
 		return a.login(ctx, authMethod)
 	}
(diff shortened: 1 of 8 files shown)


api/auth_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	t.Run("Login should set token on success", func(t *testing.T) {
+		if a.c.Token() != "" {
+			t.Errorf("client token was %v expected to be unset", a.c.Token())
+		}
+
+		_, err := a.Login(context.Background(), &m)
+		if err != nil {
+			t.Errorf("Login() error = %v", err)
+			return
```
averche, line 41: Some of these could probably be simplified a bit with `testify` library (`assert.NoError(t, "...")`. I believe it's already used somewhere in vault/api.

api/auth_test.go
```diff
@@ -0,0 +1,120 @@
...
+			t.Errorf("client token was %v expected %v", a.c.Token(), m.mockedSecret.Auth.ClientToken)
+			return
+		}
+	})
+}
+
+func TestAuth_MFALogin(t *testing.T) {
+	t.Parallel()
+
+	t.Run("MFALogin() should succeed if credentials are passed in", func(t *testing.T) {
```
averche, line 54: Could these be refactored a bit into a more table-like test?

api/sys_mfa.go
```diff
@@ -0,0 +1,45 @@
...
+	defer cancelFunc()
+
+	body := map[string]interface{}{
+		"mfa_request_id": requestID,
+		"mfa_payload":    payload,
+	}
+
+	r := c.c.NewRequest(http.MethodPost, fmt.Sprintf("/v1/sys/mfa/validate"))
+	if err := r.SetJSONBody(body); err != nil {
+		return nil, err
```
averche, line 24: It may be a good idea to add a bit more context to these errors with `fmt.Errorf("...: %w", err)`

averche (commented):
Looks really good, left a couple small comments

api/auth_test.go
VinnyHC: hmm, I don't see it in `go.mod`, do you know of an API test that uses it?

api/auth_test.go
VinnyHC: I'll see if I can do something here. I was hoping the individual tests made it a bit easier to see how the function was working.

api/auth.go
```diff
@@ -31,16 +31,82 @@ func (a *Auth) Login(ctx context.Context, authMethod AuthMethod) (*Secret, error
...
+	if len(creds) > 0 {
+		a.c.SetMFACreds(creds)
+		return a.login(ctx, authMethod)
+	}
+
+	return a.twoPhaseMFALogin(ctx, authMethod)
+}
+
+// MFAValidate validates an MFA request using the appropriate payload and a secret containing Auth.MFARequirement,
+// like the returned by MFALogin when credentials are not provided. Upon successful validation the client token
```
digivava, line 55: typo: "like the one returned by"

api/sys_mfa.go
```diff
@@ -0,0 +1,45 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+func (c *Sys) MFAValidate(requestID string, payload map[string]interface{}) (*Secret, error) {
```
digivava, line 9: Isn't the only reason we provide a difference between MethodName and MethodNameWithContext because we're trying to maintain backwards compatibility for users who are already using the one that doesn't take a context?

In this case, this is a brand new method, so I thought we would just want to have MFAValidate take a context as its first arg.

VinnyHC: I could be convinced either way!
The motivation here was that this lives in the existing `(*client).Sys()` space where most methods have both `MethodName` and `MethodNameWithContext` whereas `(*client).Auth()` is the opposite (outside of `(*client).Auth().Token()`), so I wanted `(*client).Auth().MFALogin()` to be as close to the existing `(*client).Auth().Login()` as possible.

raskchanky (approved):
Looks great! I don't have anything to add beyond what's already been commented on. Thanks for updating the existing tests to use the new methods here!

api/auth_test.go
averche: I must have seen it in the base `vault/` directory. Using standard library is the way to go then 👍

averche (approved):
LGTM, awesome job!

vault/external_tests/identity/login_mfa_okta_test.go
```diff
@@ -322,12 +323,11 @@ func mfaGenerateOktaLoginMFATest(client *api.Client) error {
...
-		"mfa_request_id": secret.Auth.MFARequirement.MFARequestID,
-		"mfa_payload": map[string][]string{
-			methodID: {},
-		},
-	})
+	secret, err = client.Sys().MFAValidateWithContext(context.Background(),
+		secret.Auth.MFARequirement.MFARequestID,
+		map[string]interface{}{
+			methodID: []string{},
+		})
```
averche, line 330:
```suggestion
		},
	)
```

Nit: bracket alignment for style consistency

vault/external_tests/identity/login_mfa_totp_test.go
```diff
@@ -76,21 +78,26 @@ func doTwoPhaseLogin(client *api.Client, totpCodePath, methodID, username string
...
-			methodID: {totpPasscode},
-		},
-	})
+
+	secret, err := client.Auth().MFAValidate(
+		context.Background(),
+		mfaSecret,
+		map[string]interface{}{
+			methodID: []string{totpPasscode},
+		})
```
averche, line 93:
```suggestion
		},
	)
```

PR #14911, opened 2022-04-05
Replace mount-filter with path-filter

Monkeychip:
The mount-filter endpoint is being deprecated - see PR[ here](https://github.com/hashicorp/vault-enterprise/pull/2716). While we don't use this endpoint we do use the language "mount filter" in one spot. I confirmed that this action is using the path-filter endpoint and changed the language to match the docs that will be updated in 1.11.


zofskeez (approved):
LGTM. Thanks!

PR #14916, opened 2022-04-05
OIDC Login Bug

zofskeez:
Addressing the following bugs when logging in with OIDC from a listed auth mounts tab:
- error is thrown _role with oidc role_type is not allowed_
- when an error is encountered the loading state is not removed
- when clicking the sign in action immediately `ember-concurrency` could throw an error indicating that the task was cancelled when using `await` when performing the task

image


hashishaw (approved):
Thanks for tackling this! I wonder if we could add some test coverage to make sure that the correct `auth_url` endpoint is hit, in both scenarios (tab and manual typing)

zofskeez:
> Thanks for tackling this! I wonder if we could add some test coverage to make sure that the correct `auth_url` endpoint is hit, in both scenarios (tab and manual typing)

I considered that. It's a bit tricky mocking out the OIDC callback so I wasn't sure how much time to spend on this. I'll hold on merging and get a couple of tests written though. I think it's worthwhile.

ui/tests/acceptance/oidc-auth-method-test.js
```diff
@@ -6,19 +6,12 @@ import { fakeWindow, buildMessage } from '../helpers/oidc-window-stub';
...
   hooks.beforeEach(function () {
     this.openStub = sinon.stub(window, 'open').callsFake(() => fakeWindow.create());
-  });
-  hooks.afterEach(function () {
-    this.openStub.restore();
-  });
-
-  // coverage for bug where token was selected as auth method for oidc and jwt
-  test('it should populate oidc auth method on logout', async function (assert) {
     this.server.post('/auth/oidc/oidc/auth_url', () => ({
```
hashishaw, line 15: What do you think about also (or replacing) this coverage with oidc at a non-standard mount path (eg. `oidc-test`)?

zofskeez: Do you mean adding a test that covers logging in using the _Mount path_ input?

zofskeez: I think as far as this logout test goes only the previously selected auth method is re-hydrated in the dropdown. It doesn't look to me like _Mount path_ gets re-hyrdrated but I could be missing something.

hashishaw: You're right, on logout with oidc I think expected behavior is that the mount path does not get filled in. I was talking about making sure the `auth_url` is generated correctly from the form, but that might be too complicated 🤔

zofskeez: Let me know what you think of the last commit. I updated the test that uses the auth path tabs and returned `test-path` from the mounts request rather than `oidc` and then assert that the correct `auth_url` endpoint is hit.

hashishaw: Great! Thanks for updating that 👍

PR #14918, opened 2022-04-05
Vault 3992 documentation changes

akshya96:


website/content/docs/plugins/index.mdx
```diff
@@ -12,6 +12,12 @@ allows both built-in and external plugins to be treated like Legos. Any plugin
 can exist at multiple different locations. Different versions of a plugin may
 be at each location, with each version differing from Vault's version.
 
+By default, Vault expects the config directory and files to be owned by the
```
mladlow, line 15: I think it's good to have this text repeated frequently in the docs because it's hard to tell when a user might encounter it, but it might be easier for us to keep all the locations up to date if it was a re-usable snippet? The 1.9 upgrade guide has some examples of using `@include` to make snippets, but I'm not sure how it works within the "warning" box as you have on the plugin arch page.

akshya96: Used @include to make snippets but looks like it does not work with "warning" box. Used it in other places.

mladlow (approved):
Looks good, thanks for the change!

PR #14919, opened 2022-04-05
[VAULT-5501] Remove `CreateOperation` from /transit/cache-config

AnPucel:
A `logical.CreateOperation` without the implementation of an `ExistenceCheck` defaults to becoming a `logical.UpdateOperation`. We're removing the `logical.CreateOperation` from `/transit/cache-config` to improve the output of the OpenApi Spec.

Verified OAS Spec doesn't advertise the `x-vault-createSupported` in OpenApi spec. Also verified the transit/cache-config `POST` still works. 

See this gh issue [#12329](https://github.com/hashicorp/vault/issues/12329)


maxb:
Please see my PR #18492 which includes this change, and more.

AnPucel:
> Please see my PR #18492 which includes this change, and more.

Thanks! I will close this in favor of your PR

PR #14922, opened 2022-04-05
Implement PKI storage migration.

stevendpclark:
- Hook into the backend::initialize function, calling the migration on a primary only.
 - Migrate an existing certificate bundle to the new issuers and key layout

### DIFF
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -7,6 +7,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/hashicorp/vault/sdk/helper/consts"
+
 	"github.com/armon/go-metrics"
 	"github.com/hashicorp/vault/helper/metricsutil"
 	"github.com/hashicorp/vault/helper/namespace"
@@ -119,7 +121,8 @@ func Backend(conf *logical.BackendConfig) *backend {
 			secretCerts(&b),
 		},
 
-		BackendType: logical.TypeLogical,
+		BackendType:    logical.TypeLogical,
+		InitializeFunc: b.initialize,
 	}
 
 	b.crlLifetime = time.Hour * 72
@@ -232,4 +235,23 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
 		}
 		return resp, err
 	}
+}
+
+// initialize is used to perform a possible PKI storage migration if needed
+func (b *backend) initialize(ctx context.Context, req *logical.InitializationRequest) error {
+	logger := b.Logger().Named("initialize")
+
+	// Early exit if not a primary cluster or performance secondary with a local mount.
+	if b.System().ReplicationState().HasState(consts.ReplicationDRSecondary|consts.ReplicationPerformanceStandby) ||
+		(!b.System().LocalMount() && b.System().ReplicationState().HasState(consts.ReplicationPerformanceSecondary)) {
+		logger.Debug("skipping pki migration as we are not on primary or secondary with a local mount")
+		return nil
+	}
+
+	if err := migrateStorage(ctx, req, logger); err != nil {
+		logger.Error("Error during migration of PKI mount: " + err.Error())
+		return err
+	}
+
+	return nil
 }
--- a/builtin/logical/pki/config_util.go
+++ b/builtin/logical/pki/config_util.go
@@ -1,0 +1,37 @@
+package pki
+
+import (
+	"context"
+
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+func updateDefaultKeyId(ctx context.Context, s logical.Storage, id keyId) error {
+	config, err := getKeysConfig(ctx, s)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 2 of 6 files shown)


builtin/logical/pki/backend.go
```diff
@@ -233,3 +236,14 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
...
 }
+
+// initialize is used to perform a possible PKI storage migration if needed
+func (b *backend) initialize(ctx context.Context, req *logical.InitializationRequest) error {
+	// Only primary should perform the upgrade.
+	if b.System().ReplicationState().HasState(consts.ReplicationPerformanceSecondary | consts.ReplicationPerformanceStandby | consts.ReplicationDRSecondary) {
+		return nil
+	}
+
+	logger := b.Logger().Named("initialize")
```
cipherboy, line 247: Can it ever be that everyone thinks they're a PerfSecondary/PerfStandby/DRSecondary during initialization because the last leader wasn't brought up, perhaps with an auto-unseal mechanism? Or is that unlikely?

builtin/logical/pki/storage.go
```diff
@@ -288,7 +291,7 @@ func fetchIssuerById(ctx context.Context, s logical.Storage, issuerId issuerId)
 	return &issuer, nil
 }
 
-func writeIssuer(ctx context.Context, s logical.Storage, issuer issuer) error {
+func writeIssuer(ctx context.Context, s logical.Storage, issuer *issuer) error {
```
cipherboy, line 294: Thank you :D I was actually considering this.

builtin/logical/pki/config_util.go
```diff
@@ -0,0 +1,37 @@
+package pki
+
+import (
+	"context"
+
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+func updateDefaultKeyId(ctx context.Context, s logical.Storage, id keyId) error {
```
cipherboy, line 9: This is very useful, thank you!

builtin/logical/pki/storage.go
```diff
@@ -15,10 +15,13 @@ import (
 )
 
 const (
-	storageKeyConfig     = "/config/keys"
-	storeageIssuerConfig = "/config/issuers"
-	keyPrefix            = "/config/key/"
-	issuerPrefix         = "/config/issuer/"
+	storageKeyConfig    = "config/keys"
```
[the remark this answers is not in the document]
stevendpclark: Yup! We need to update that, I'll file an issue so we don't forget.

builtin/logical/pki/backend.go
stevendpclark: It's a common pattern I see in Vault so I don't believe so but I can't point to proof of that at the moment. One thing you made me think of that is missing in this statement though is local mounts...

builtin/logical/pki/backend.go
```diff
@@ -233,3 +236,14 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
...
+	return migrateStorage(ctx, req, logger)
```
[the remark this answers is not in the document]
stevendpclark: Yup oversight on my part, would be good to include the mount point in those log messages in case we are migrating multiple PKI mounts at once...

builtin/logical/pki/backend.go
stevendpclark: After the morning coffee, no there isn't a way that a cluster/node can be switched state during initialization of it's cluster type based on leader election. Technically a primary cluster could be converted to a secondary but it will get wiped out and reset anyways.

cipherboy (approved):

kitography (approved):
This looks great - I have a question, it isn't even a nit;

Nice tests; Panic in no randomness makes me happy; very clean code changes; thank you for fixing spelling; i think that migration strategy is impressive and will solve all our messes.

builtin/logical/pki/backend.go
```diff
@@ -233,3 +236,22 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
...
+
+	// Early exit if not a primary cluster or performance secondary with a local mount.
+	if b.System().ReplicationState().HasState(consts.ReplicationDRSecondary|consts.ReplicationPerformanceStandby) ||
+		(!b.System().LocalMount() && b.System().ReplicationState().HasState(consts.ReplicationPerformanceSecondary)) {
+		logger.Debug("skipping pki migration as we are not on primary or secondary with a local mount")
+		return nil
+	}
+
+	if err := migrateStorage(ctx, req, logger); err != nil {
+		logger.Error("Error during migration of PKI mount: " + err.Error())
```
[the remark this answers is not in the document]
stevendpclark: So during my testing the error appears within Vault's server log, and the first request to the mount gets an 500 error 
 as well as sees the error message.

Now part of a bigger problem I'm seeing is the mount is still technically active post migration failure, which is another question I will bring up.

cipherboy:

PR #14937, opened 2022-04-06
Upgrade ember-test-selectors

zofskeez:
In the latest version of `ember-test-selectors`, automatic attribute binding was removed. This causes issues with older components where either the `data-test` attribute was defined as a property on the component, the selector was passed in as an arg without defining it in `attributeBindings` or the selector was passed as `data-test-some-selector={{true}}`.

In some cases I was able to use the `attributeBindings` property but in others there were different test selectors defined throughout the code and it was more straightforward to glimmerize the component and spread the attributes on the root element.


ui/tests/acceptance/secrets/backend/kv/secret-test.js
```diff
@@ -213,12 +213,16 @@ module('Acceptance | secrets/secret/create', function (hooks) {
...
-    assert
-      .dom('[data-test-input="maxVersions"]')
-      .hasClass('has-error-border', 'shows border error on input with error');
-    assert.dom('[data-test-secret-save]').isDisabled('Save button is disabled');
+    /*
+     * commenting out for now until ember-cp-validations is updated or removed
+     * throws an error when attempting to use isHTMLSafe which is imported from @ember/string in the cp-validations code
+     * in ember 3.28 the import changed to @ember/template
+     */
+    // await typeIn('[data-test-input="maxVersions"]', 'abc');
```
zofskeez, line 221: Attempting to type in a string on a number field is triggering an uncaught error from `ember-cp-validations`. I tracked it down to this line [in the code](https://github.com/offirgolan/ember-cp-validations/blob/9f05d9be6e12e4e6db6e3ed3501de6593a9eafb1/addon/utils/utils.js#L4). Since this is a side branch from the upgrade I want to comment this out for now so the tests pass to unblock this merge and deal with cp-validations separately.

ui/app/templates/components/toolbar-secret-link.hbs
```diff
@@ -1,17 +1,10 @@
...
-  @data-test-credentials-link={{this.data-test-credentials-link}}
-  @data-test-backend-credentials={{this.data-test-backend-credentials}}
-  @data-test-transit-action-link={{this.data-test-transit-action-link}}
-  @data-test-transit-key-actions-link={{this.data-test-transit-key-actions-link}}
+  @mode={{@mode}}
+  @secret={{@secret}}
+  @replace={{@replace}}
+  @queryParams={{@queryParams}}
+  class="toolbar-link"
+  ...attributes
```
hashishaw, line 7: Since we're passing along attributes named the same, would those all be captured in `...attributes` so we can remove all the `@attrs`? If we do this, I'd love to see those attributes called out in the JSDocs of the class component

ui/tests/acceptance/secrets/backend/kv/secret-test.js
hashishaw: Would you add `TODO` to the comment so we can easily find it again?

ui/lib/core/addon/components/linked-block.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="linked-block" role="link" {{on "click" this.onClick}} ...attributes>
```
hashishaw, line 1: Should this be `role="button"` or `role="link"` since we're doing a function action on click?

ui/lib/core/addon/components/form-field.js
```diff
@@ -31,7 +31,8 @@ import layout from '../templates/components/form-field';
 export default Component.extend({
   layout,
   'data-test-field': true,
-  classNames: ['field'],
+  attributeBindings: ['data-test-field'],
```
hashishaw, line 34: Why are we adding back attributeBindings here when we removed them everywhere else?

hashishaw (commented):
Really great work so far, thanks for tackling this! I left a few comments.

One nit that you don't need to fix here but maybe we can update/standardize as we go along: I prefer to have all the `@attrs` and non-attrs grouped, so instead of 
```hbs
<Component
  @thing="thing"
  class="foo"
  @foo="bar" />
```
it would be 
```hbs
<Component
  @thing="thing"
  @foo="bar"
   class="foo" />
```

zofskeez:
> Really great work so far, thanks for tackling this! I left a few comments.
> 
> One nit that you don't need to fix here but maybe we can update/standardize as we go along: I prefer to have all the `@attrs` and non-attrs grouped, so instead of
> 
> ```
> <Component
>   @thing="thing"
>   class="foo"
>   @foo="bar" />
> ```
> 
> it would be
> 
> ```
> <Component
>   @thing="thing"
>   @foo="bar"
>    class="foo" />
> ```

Ya same here but given the lift on this I didn't want to spend more time rearranging manually. Since each developer may have a different preference, if we want to enforce this we should consider leveraging the [plugin system and custom rules for ember-template-lint](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/plugins.md).

I haven't tried defining a custom rule yet but as far as the transform aspect goes it would be pretty straightforward to group all the args first I think. We should be able to wire it up to `formatOnSave` for those using VSCode, and/or update our pre-commit lint hook to use --fix.

ui/lib/core/addon/components/form-field.js
zofskeez: I didn't glimmerize this component and now that automatic attribute binding has been removed defining `data-test-field: true` as a property does not do anything. Adding it to `attributeBindings` is necessary to bind it to the root element.

ui/lib/core/addon/components/linked-block.hbs
zofskeez: :thinking:. I suppose button would be more appropriate. Although the click action results in a transition so in that way it behaves as a link. I really don't know where to land :smile:. What do you think?

ui/app/templates/components/toolbar-secret-link.hbs
zofskeez: Sorry I'm not sure I fully understand. What do you mean by `@attrs`? `...attributes` will only spread html attributes. I thought I saw an Ember RFC for being able to spread component args as well which would be great but not sure when that will land. As far as attributes named the same, the values are concatenated. For example:
```hbs
<ToolbarSecretLink
 class="passed-in"
/>
{{!-- results in --}}
<a href="/ui/vault/secrets/secret/create" id="ember509" class="ember-view toolbar-link passed-in" data-test-secret-create="">
  Create secret
</a>
```

zofskeez: [Found the RFC I mentioned!](https://github.com/emberjs/rfcs/pull/593)

hashishaw: Oh nice! I think i was thinking about React-land, where non-html attributes are passed. Carry on!

ui/lib/core/addon/components/form-field.js
hashishaw: Should we remove `data-test-field: true` then?

ui/lib/core/addon/components/linked-block.hbs
hashishaw: hmm.... I'm going to ask our resident a11y experts if they have an opinion on this but I think, to your point, since the behavior is to navigate it probably makes sense to stay a link

ui/lib/core/addon/components/form-field.js
zofskeez: I think there were instances where `data-test-field` was not passed into the component and some tests were failing so I added it back as a default. If the component value for `data-test-field` is not truthy the attribute will not be bound to the element.

hashishaw: Thanks for the explanation! Can't wait till we glimmerize all the things.

zofskeez: We must celebrate when that day arrives!

PR #14941, opened 2022-04-06
Bug Fix and Glimmerize secret-edit component

Monkeychip:
This PR takes the long way around to fix the following bug:

_Only on the binary did this bug occur_
- create secret with several versions
- mark a version as deleted
- switch to another version, no option to delete where there should be one (on the delete modal).

By converting the secret-edit component to glimmer this handled the issue where the computed property modelForData was not being recomputed on a version change. To change in earlier versions I will simply add "selectedVersion" to the properties on the computed property. However, now that this is glimmerized it's MUCH easier to read. I added documentation and cleaned out some unused functionality: largely the two mixins that this component had extended. We were not using them, probably a left over from the earlier KV work.

I ran all the tests locally and they pass as well as testing this as a binary and making sure the bug didn't show up again.

### DIFF
--- a/ui/app/components/secret-edit-toolbar.js
+++ b/ui/app/components/secret-edit-toolbar.js
@@ -12,7 +12,6 @@
  * @secretDataIsAdvanced={{secretDataIsAdvanced}}
  * @showAdvancedMode={{showAdvancedMode}}
  * @modelForData={{this.modelForData}}
- * @navToNearestAncestor={{this.navToNearestAncestor}}
  * @canUpdateSecretData={{canUpdateSecretData}}
  * @codemirrorString={{codemirrorString}}
  * @wrappedData={{wrappedData}}
@@ -30,7 +29,6 @@
  * @param {boolean} secretDataIsAdvanced - used to determine if show JSON toggle
  * @param {boolean} showAdvacnedMode - used for JSON toggle
  * @param {object} modelForData - a modified version of the model with secret data
- * @param {string} navToNearestAncestor - route to nav to if press cancel
  * @param {boolean} canUpdateSecretData - permissions that show the create new version button or not.
  * @param {string} codemirrorString - used to copy the JSON
  * @param {object} wrappedData - when copy the data it's the token of the secret returned.
--- a/ui/app/components/secret-edit.js
+++ b/ui/app/components/secret-edit.js
@@ -120,7 +120,9 @@ export default class SecretEdit extends Component {
     return this.secretData.isAdvanced();
   }
 
-  @or('secretDataIsAdvanced', 'preferAdvancedEdit') showAdvancedMode;
+  get showAdvancedMode() {
+    return this.secretDataIsAdvanced || this.args.preferAdvancedEdit;
+  }
 
   get isWriteWithoutRead() {
     if (!this.args.model) {
--- a/ui/app/components/secret-edit.js
+++ b/ui/app/components/secret-edit.js
@@ -7,161 +7,180 @@
  * <SecretEdit @model={{model}}/>
  * ```
 /
- * @param {object} model - Model returned from route secret-v2
+ * @param {object} model - Model returned from secret-v2 which is generated in the secret-edit route
  */
 
 import { inject as service } from '@ember/service';
-import Component from '@ember/component';
-import { computed } from '@ember/object';
-import { alias, or } from '@ember/object/computed';
-import FocusOnInsertMixin from 'vault/mixins/focus-on-insert';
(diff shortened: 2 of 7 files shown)


ui/app/components/secret-edit-toolbar.js
```diff
@@ -12,7 +12,6 @@
  * @secretDataIsAdvanced={{secretDataIsAdvanced}}
  * @showAdvancedMode={{showAdvancedMode}}
  * @modelForData={{this.modelForData}}
- * @navToNearestAncestor={{this.navToNearestAncestor}}
```
Monkeychip, line 15: A mixin method we were not using. We passed it into the secret-delete-menu but didn't actually use it.

ui/app/templates/components/secret-edit-toolbar.hbs
```diff
@@ -19,7 +19,6 @@
       <SecretDeleteMenu
         @modelForData={{@modelForData}}
         @model={{@model}}
-        @navToNearestAncestor={{@navToNearestAncestor}}
```
Monkeychip, line 22: Again another mixin method we were passing through but never using.

ui/app/components/secret-edit.js
```diff
@@ -108,60 +88,59 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
...
+  @alias('checkMetadataCapabilities.canUpdate') canUpdateSecretMetadata;
+  @alias('checkMetadataCapabilities.canRead') canReadSecretMetadata;
 
-  requestInFlight: or('model.isLoading', 'model.isReloading', 'model.isSaving'),
+  @or('model.isLoading', 'model.isReloading', 'model.isSaving') requestInFlight;
+  @or('requestInFlight', 'model.isFolder', 'model.flagsIsInvalid') buttonDisabled;
 
-  buttonDisabled: or('requestInFlight', 'model.isFolder', 'model.flagsIsInvalid'),
-
-  modelForData: computed('isV2', 'model', function () {
```
Monkeychip, line 120: here will be the backport fix. I'll make this `computed('isV2', 'model', 'selectedVersion', function ...`

ui/app/components/secret-edit.js
```diff
@@ -1,84 +1,63 @@
...
-    let secrets = this.model.secretData;
-    if (!secrets && this.model.selectedVersion) {
-      this.set('isV2', true);
-      secrets = this.model.belongsTo('selectedVersion').value().secretData;
+  constructor() {
+    super(...arguments);
+    let secrets = this.args.model.secretData;
+    if (!secrets && this.args.model.selectedVersion) {
+      this.isV2 = true;
+      secrets = this.args.model.belongsTo('selectedVersion').value().secretData;
```
zofskeez, line 42: This is a bit of an interesting way of accessing this property. You should be able to do `this.args.model.selectedVersion.secretData`.

ui/app/components/secret-edit.js
```diff
@@ -1,84 +1,63 @@
...
     }
     const data = KVObject.create({ content: [] }).fromJSON(secrets);
-    this.set('secretData', data);
-    this.set('codemirrorString', data.toJSONString());
-    if (data.isAdvanced()) {
-      this.set('preferAdvancedEdit', true);
-    }
```
zofskeez, line 67: If `preferAdvancedEdit` is an arg is this check happening higher up now? I'm just wondering since the component was potentially overriding the original value if we will run into state issues.

zofskeez (commented):
Good work on ✨!

ui/app/components/secret-edit.js
Monkeychip: It's a copy over from how it was done before. I'm hesitant to change it in case I'm missing a reasoning for accessing the store again. maybe a refresh after choosing the version or something like that.

ui/app/components/secret-edit.js
```diff
@@ -108,60 +88,59 @@ export default Component.extend(FocusOnInsertMixin, WithNavToNearestAncestor, {
...
-  secretDataIsAdvanced: computed('secretData', 'secretData.[]', function () {
+  get secretDataIsAdvanced() {
     return this.secretData.isAdvanced();
-  }),
+  }
 
-  showAdvancedMode: or('secretDataIsAdvanced', 'preferAdvancedEdit'),
+  get showAdvancedMode() {
+    return this.secretDataIsAdvanced || this.args.preferAdvancedEdit;
+  }
```
zofskeez, line 120: Following up from our zoom conversation that this new getter will cover the scenario where `preferAdvancedEdit` was previously being set in the constructor based on the `isAdvanced` value.

PR #14942, opened 2022-04-06
UI/Add months to activity serializer

hellobontempo:
This PR adds the `months` key from the `/activity` response to the serializer so it can be consumed by the UI for the 1.11 client count designs.

This will be one of a series of PRs - will add a changelog for the feature work to the final PR

### DIFF
--- a/ui/app/models/clients/activity.js
+++ b/ui/app/models/clients/activity.js
@@ -1,10 +1,12 @@
 import Model, { attr } from '@ember-data/model';
 export default class Activity extends Model {
-  @attr('string') responseTimestamp;
+  @attr('array') byMonthTotalClients;
+  @attr('array') byMonthNewClients;
   @attr('array') byNamespace;
+  @attr('object') total;
   @attr('array') formattedEndTime;
   @attr('array') formattedStartTime;
   @attr('string') startTime;
   @attr('string') endTime;
-  @attr('object') total;
+  @attr('string') responseTimestamp;
 }
--- a/ui/app/serializers/clients/activity.js
+++ b/ui/app/serializers/clients/activity.js
@@ -39,24 +39,24 @@ export default class ActivitySerializer extends ApplicationSerializer {
         return {
           month: m.timestamp,
           entity_clients: m.new_clients.counts.entity_clients,
-          non_entity_clients: m['new_clients']['counts']['non_entity_clients'],
-          total: m['new_clients']['counts']['clients'],
-          namespaces: this.flattenDataset(m['new_clients']['namespaces']),
+          non_entity_clients: m.new_clients.counts.non_entity_clients,
+          total: m.new_clients.counts.clients,
+          namespaces: this.flattenDataset(m.new_clients.namespaces),
         };
       });
     } else {
       return payload.map((m) => {
         return {
           month: m.timestamp,
-          entity_clients: m['counts']['entity_clients'],
-          non_entity_clients: m['counts']['non_entity_clients'],
-          total: m['counts']['clients'],
-          namespaces: this.flattenDataset(m['namespaces']),
+          entity_clients: m.counts.entity_clients,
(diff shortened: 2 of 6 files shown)


ui/app/models/clients/activity.js
```diff
@@ -1,10 +1,12 @@
 import Model, { attr } from '@ember-data/model';
 export default class Activity extends Model {
-  @attr('string') responseTimestamp;
+  @attr('array') byMonthTotalClients;
+  @attr('array') byMonthNewClients;
   @attr('array') byNamespace;
+  @attr('object') total;
```
hashishaw, line 6: In my experience `@attr('object')` is less supported than just `@attr()` which leaves whatever object passed alone. We can leave this until implementation demands otherwise, but just an FYI of a potential gotcha

ui/app/serializers/clients/activity.js
```diff
@@ -29,12 +32,50 @@ export default class ActivitySerializer extends ApplicationSerializer {
...
+  }
+
+  // In 1.10 'distinct_entities' changed to 'entity_clients' and
   // 'non_entity_tokens' to 'non_entity_clients'
-  // accounting for deprecated API keys here and updating to latest nomenclature
   homogenizeClientNaming(object) {
-    // TODO CMB check with API payload, latest draft includes both new and old key names
-    // TODO CMB Delete old key names IF correct ones exist?
+    // if new key names exist, only return those key/value pairs
+    if (Object.keys(object).includes('entity_clients', 'non_entity_clients')) {
```
hashishaw, line 70: I think `Array.includes` only takes one parameter so this should be updated to an or statement, one for each. Is it reasonable to assume that if we have one of the updated names, we have both? In that case we can maybe just check for one.

ui/app/serializers/clients/activity.js
```diff
@@ -29,12 +32,50 @@ export default class ActivitySerializer extends ApplicationSerializer {
...
   }
 
-  // For 1.10 release naming changed from 'distinct_entities' to 'entity_clients' and
+  // for vault usage - vertical bar chart
+  flattenByMonths(payload, isNewClients = false) {
+    if (isNewClients) {
+      return payload.map((m) => {
+        return {
+          month: m.timestamp,
+          entity_clients: m['new_clients']['counts']['entity_clients'],
```
hashishaw, line 41: Since the selectors are static strings, you can just access like
```suggestion
          entity_clients: m.new_clients.counts.entity_clients,
```
and ignoring the linting. Alternatively, we can make this a little more concise by removing the outer `isNewClients` check and doing:
```js
const dataKey = isNewClients ? 'new_clients.counts' : 'counts';
return payload.map(m => {
  const flat = {
    entity_clients: m[dataKey].entity_clients,
    ...
  }
  if (isNewClients) {
    flat.new_clients = { ... }
  }
}
```

ui/app/serializers/clients/activity.js
```diff
@@ -45,7 +86,6 @@ export default class ActivitySerializer extends ApplicationSerializer {
         non_entity_clients,
       };
     }
-    return object;
```
hashishaw, line 48: What happens if the object doesn't include any of the keys we check for? Return an empty object?

hashishaw (commented):
Nice! Left a couple comments, let me know if you have any questions

ui/app/serializers/clients/activity.js
hellobontempo: Yes - it sure does 😅 - great idea thank you!

ui/app/models/clients/activity.js
hellobontempo: This was `@attr('object')` in the last release, so I don't _think_ it caused any problems? But good FYI

hashishaw (approved):
🚀

PR #14943, opened 2022-04-06
Fix handling of SignatureBits for ECDSA issuers

cipherboy:
When adding SignatureBits control logic, we incorrectly allowed
specification of SignatureBits in the case of an ECDSA issuer. As noted
in the original request, NIST and Mozilla (and others) are fairly
prescriptive in the choice of signatures (matching the size of the
NIST P-curve), and we shouldn't usually use a smaller (or worse, larger
and truncate!) hash.

Ignore the configuration of signature bits and always use autodetection
for ECDSA like ed25519.

---

This definitely warrants a changelog entry, but not sure if we need anything else? This was added in 1.10.0, we'll backport this to 1.10.1, fixing the issue. It only presently works on roles (it should probably be added to the other endpoints later?). 

Ticket: VAULT-5707

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -4476,8 +4476,11 @@ func TestBackend_Roles_IssuanceRegression(t *testing.T) {
 }
 
 type KeySizeRegression struct {
-	RoleKeyType       string
-	RoleKeyBits       []int
+	// Values reused for both Role and CA configuration.
+	RoleKeyType string
+	RoleKeyBits []int
+
+	// Signature Bits presently is only specified on the role.
 	RoleSignatureBits []int
 
 	// These are tuples; must be of the same length.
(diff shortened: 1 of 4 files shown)


sdk/helper/certutil/helpers.go
```diff
@@ -863,16 +858,25 @@ func createCertificate(data *CreationBundle, randReader io.Reader, privateKeyGen
...
+	// Lastly, note that pub above needs to be the _signer's_ public key;
+	// the issue with DefaultOrValueHashBits is that it is called at role
+	// configuration time, which might _precede_ issuer generation. Thus
+	// it only has access to the desired key type and not the actual issuer.
+	// The reference from that function is reproduced below:
+	//
+	// > To comply with BSI recommendations Section 4.2 and Mozilla root
+	// > store policy section 5.1.2, enforce that NIST P-curves use a hash
+	// > length corresponding to curve length. Note that ed25519 does not
+	// > the "ec" key type.
```
kitography, line 879: does not implement?

sdk/helper/certutil/helpers.go
```diff
@@ -863,16 +858,25 @@ func createCertificate(data *CreationBundle, randReader io.Reader, privateKeyGen
...
+	// suggest instead using the correct hash function, we should prefer that
+	// over the operator-specified signatureBits.
+	//
+	// Lastly, note that pub above needs to be the _signer's_ public key;
+	// the issue with DefaultOrValueHashBits is that it is called at role
+	// configuration time, which might _precede_ issuer generation. Thus
+	// it only has access to the desired key type and not the actual issuer.
+	// The reference from that function is reproduced below:
+	//
+	// > To comply with BSI recommendations Section 4.2 and Mozilla root
```
kitography, line 876: Confused looking at this list of curves below: we explicitly set ECDSAWithSHA256 for P224 and 256 (but not P192 which is set by default)?

I think it would be cleaner/safer if we added P192 to the first case, then either error if we don't recognize the alrogirthm?

kitography (approved):
Nice test, the only confusing part of this fix is not in code that you changed.

stevendpclark (approved):

sdk/helper/certutil/helpers.go
cipherboy: We actually restrict it to at least P-224:

https://github.com/hashicorp/vault/blob/74248e1b20a6f48bfc94f7d72551fc5cebdbd3ea/sdk/helper/certutil/helpers.go#L690-L694 

And Go lacks support for P-192: https://github.com/golang/go/blob/master/src/crypto/elliptic/elliptic.go#L442-L447 and https://pkg.go.dev/crypto/elliptic

So there's no way to do this comparison.

What is the hash function for P-192 anyways? [This](https://csrc.nist.gov/csrc/media/projects/cryptographic-standards-and-guidelines/documents/examples/ecdsa_prime.pdf) makes me think it is the insecure SHA-1 which isn't... good? :-)

sdk/helper/certutil/helpers.go
cipherboy: Ah that's right. :-)

builtin/logical/pki/backend_test.go
```diff
@@ -4543,10 +4577,12 @@ func TestBackend_Roles_KeySizeRegression(t *testing.T) {
...
 		// and any size RSA/Ed25519 keys.
 		/*  2 */ {"ec", []int{0}, []int{0}, []string{"rsa", "rsa", "rsa", "ec", "ed25519"}, []int{1024, 2048, 4096, 224, 0}, true},
-		// But it should work to issue larger EC keys.
-		/*  3 */ {"ec", []int{0, 256}, []int{0, 256}, []string{"ec"}, []int{256}, false},
-		/*  4 */ {"ec", []int{384}, []int{0, 384}, []string{"ec"}, []int{384}, false},
-		/*  5 */ {"ec", []int{521}, []int{0, 512}, []string{"ec"}, []int{521}, false},
+		// But it should work to issue larger EC keys. Note that we should be
+		// independent of signature bits as that's computed from the issuer
+		// type (for EC based issuers).
+		/*  3 */ {"ec", []int{0, 256}, []int{0, 256, 384, 521}, []string{"ec", "ec", "ec"}, []int{256, 384, 521}, false},
```
stevendpclark, line 4583: nit: Are we missing a testcase for P224?

cipherboy: Yes, and pushed :-)

PR #14947, opened 2022-04-06
Revert the WithContext changes to vault tests

averche:
These test changes are causing many conflicts in the automatic backport PRs and don't provide enough value to justify them.


cipherboy (approved):
Looks like we're in good shape:

```
$ git diff --name-only HEAD~ | sed '/_test\.go$/d'
```

reports no changed files that were non-tests. It looks like it'll make the rebase of PKI tests a lot easier.

Looking forward to OpenAPI :-)

Issue #14934, opened 2022-04-06
sys: don't set `X-Vault-Namespace` header for root-only paths 


aead:
**Describe the bug**
Vault exposes root-only API paths. For example: `[sys/health](https://www.vaultproject.io/docs/enterprise/namespaces#sys-health).
Ref: https://www.vaultproject.io/docs/enterprise/namespaces#root-only-api-paths

However, a `vault.Client` with a namespace set via [`SetNamespace`](https://pkg.go.dev/github.com/hashicorp/vault/api#Client.SetNamespace) sends the `X-Vault-Namespace` header to those root-only API paths.
For example:
```
User-Agent: [Go-http-client/1.1]
X-Vault-Namespace: [test]
X-Vault-Request: [true]
X-Vault-Token: [abc]
Accept-Encoding: [gzip]
```

This causes Vault `1.9.1` to respond with:
```
Code: 404. Errors:

    unsupported path
```  

**To Reproduce**

As a simple reproducer - the 1st curl request succeeds while the 2nd fails:
```
curl -L -X GET "https://localhost:8200/v1/sys/health?drsecondarycode=299&performancestandbycode=299&sealedcode=299&standbycode=299&uninitcode=299"

curl -L -X GET -H 'X-Vault-Namespace: test' "https://localhost:8200/v1/sys/health?drsecondarycode=299&performancestandbycode=299&sealedcode=299&standbycode=299&uninitcode=299"
```

**Expected behavior**
Either the`vault.Client` should not send the `X-Vault-Namespace` header for root-only API paths or Vault itself should ignore this header in such a case.

**Environment:**
* Vault Server Version (retrieve with `vault status`): `1.9.1` 
* Vault SDK Version (retrieve with `vault version`):  `1.5.0`
* Server Operating System/Architecture: linux/amd64

aead:
Just to elaborate on a possible use case:
A client is using a secret engine (K/V, transient, ...) and at the same time checks Vault's health status periodically.
Since, `sys/health` is currently a root-only API path you **have** to use two clients to achieve this. 
Using the same client results in the pretty confusing error `404 unsupported path`.

From the client perspective this is a total surprise. 
If the solution should be:
> My suggestion is that you either use two clients for the two different namespaces

Then this behavior should be documented more explicitly IMO. In particular since certain API paths are currently tied to one namespace.

VinnyHC:
[...]

PR #14948, opened 2022-04-06
Unset Namespace on root API

VinnyHC:
Fixes: #14934 

In the API `Client` should ignore namespaces on calls that are intended to be issued against the [root](https://www.vaultproject.io/docs/enterprise/namespaces#root-only-api-paths).

This PR modifies the following calls to unset the namespace, issue the call, and reset the namespace configuration:
`(*client).Sys().Health()` / `(*client).Sys().HealthWithContext()`
`(*client).Sys().InitStatus()` / `(*client).Sys().InitStatusWithContext()` 
`(*client).Sys().Init()` / `(*client).Sys().InitWithContext()`
`(*client).Sys().KeyStatus()` / `(*client).Sys().KeyStatusWithContext()`
`(*client).Sys().Leader()` / `(*client).Sys().LeaderWithContext()`
as well as all of the `Sys()` methods that target `sys/storage/raft/*`

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -826,6 +826,14 @@ func (c *Client) ClearNamespace() {
 	c.headers.Del(consts.NamespaceHeaderName)
 }
 
+// Namespace returns the namespace currently set by this client. It will
+// return the empty string if there is no namespace set.
+func (c *Client) Namespace() string {
+	c.modifyLock.Lock()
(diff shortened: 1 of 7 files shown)


VinnyHC:
It seems to make more sense to use one client per namespace so that an explicit choice is made.

PR #14953, opened 2022-04-07
Fix dead link for JWT supported algorithms

tomhjp:
The previous link was broken. I wasn't sure whether to try to pick an evergreen link (what I've currently opted for), or a more concrete link that will never get broken but may get stale (by specifying both the latest commit and the exact line numbers) - I'm open to input.


austingebauer (approved):
Thanks, @tomhjp!

Issue #14781, opened 2022-03-30
AWS Auth Login Metadata


jlestrada:
**Is your feature request related to a problem? Please describe.**
Yes, the proposed feature request is related to a problem with the audit trail when using AWS Auth Engine.

The scenario is that the AWS Auth engine is enabled and configured and the use of an AWS Auth Role is binded to an IAM Role as the principal entity. With the IAM Role bounded to a Vault Role, an IAM User is then capable of assuming this role to log into the Vault server. This authorization flow works well and allows for Vault access to be  controlled through IAM access to assume the role. Multiple IAM Users are capable of assuming this role. 

The issue specifically is the lack of observability on the audit trail. With the Vault server audit logs enabled, the following metadata format is supplied.

```
"metadata": {
	"client_user_id": <unique_id>,
	"role_id": <role_id>,
	"auth_type": "iam",
	"canonical_arn": "arn:aws:iam::<aws_account>:role/<iam-role-name>,
	"account_id": <aws_account>,
	"client_arn": "arn:aws:sts::<aws_account>:assumed-role/<iam-role-name>/vaultSession"
}
```

The extra metadata fields are configured via the [iam_metadata endpoint](https://www.vaultproject.io/api-docs/auth/aws#iam_metadata).

None of these fields are unique to the IAM User who is assuming the role. This lack of observability as a result does not allow for specific actions to be traced back to a user during an audit. 

**Describe the solution you'd like**
The request here is to develop a way in which a clear owner can be pinned to specific Vault actions with the use of IAM Roles. An AWS metadata field such as the `AWS::IAM::AccessKey` can be used to determine a unique individual based on cloudtrail lookups.

**Describe alternatives you've considered**
The alternative at the moment is to bind the IAM User arn explicitly to the Vault Role. This direct binding will explicitly use the ARN of the IAM User. The drawback to this method is the need to consistently update the Vault AWS Auth Role for changes to who should and shouldn't have access. Being able to control this from outside of Vault would be helpful from an operational toil perspective.

**Explain any additional use-cases**
The additional/extended use case here is the combined use of the Vault DB Engine configuration. The DB Engine is configured such that it is creating temporary and unique roles for users of this Vault Role binded to the IAM Role. This feature of credentials creation is useful as it allows a user to access the Database in a scoped manner with short lived credentials. 

[...]

PR #14954, opened 2022-04-07
aws auth displayName

jlestrada:
What: 
Modify the AWS Auth engine such that it is capable of setting the field `displayName` to include the `RoleSessionName` when assume role process is used to authenticate to vault server.

Why:
This update would allow users to be capable of setting DB Engine dynamic credentials managed by the username template to match IAM User names. Please read details [here](https://github.com/hashicorp/vault/issues/14781#issuecomment-1091830812).

### DIFF
--- a/builtin/credential/aws/path_login.go
+++ b/builtin/credential/aws/path_login.go
@@ -1407,6 +1407,11 @@ func (b *backend) pathLoginUpdateIam(ctx context.Context, req *logical.Request, 
 			Name: identityAlias,
 		},
 	}
+
+	if entity.Type == "assumed-role" {
+		auth.DisplayName = strings.Join([]string{entity.FriendlyName, entity.SessionInfo}, "/")
+	}
+
 	roleEntry.PopulateTokenAuth(auth)
 	if err := identityConfigEntry.IAMAuthMetadataHandler.PopulateDesiredMetadata(auth, map[string]string{
 		"client_arn":           callerID.Arn,
--- a/changelog/14954.txt
+++ b/changelog/14954.txt
@@ -1,0 +1,3 @@
+```release-note:change
+auth/aws: Add RoleSession to DisplayName when using assumeRole for authentication
+```


hghaf099:
@jlestrada Thanks for your submission. Would you please make sure to sign the CLA? And, some tests are also failing. Then, we can have someone to take a deeper look into the PR. Thanks

calvn:
It looks like `entity.FriendlyName` can also be used in the case were the role is not supplied to the request in here to determine the role name:

https://github.com/hashicorp/vault/blob/b17e3256dde937a6248c9a2fa56206aac93d07de/builtin/credential/aws/path_login.go#L1257-L1265

Instead of modifying `parseIamArn`, what do you think if we change the way we set `logical.Auth.DisplayName` such that it's based conditionally on `entity.Type`?

https://github.com/hashicorp/vault/blob/b17e3256dde937a6248c9a2fa56206aac93d07de/builtin/credential/aws/path_login.go#L1405

So something like the following before we return `auth`:

```
if entity.Type == "assume-role" {
   auth.DisplayName = strings.Join([]string{entity.FriendlyName, entity.SessionInfo}, "/")
}
```

jlestrada:
> It looks like `entity.FriendlyName` can also be used in the case were the role is not supplied to the request in here to determine the role name:
> 
> https://github.com/hashicorp/vault/blob/b17e3256dde937a6248c9a2fa56206aac93d07de/builtin/credential/aws/path_login.go#L1257-L1265
> 
> Instead of modifying `parseIamArn`, what do you think if we change the way we set `logical.Auth.DisplayName` such that it's based conditionally on `entity.Type`?
> 
> https://github.com/hashicorp/vault/blob/b17e3256dde937a6248c9a2fa56206aac93d07de/builtin/credential/aws/path_login.go#L1405
> 
> So something like the following before we return `auth`:
> 
> ```
> if entity.Type == "assume-role" {
>    auth.DisplayName = strings.Join([]string{entity.FriendlyName, entity.SessionInfo}, "/")
> }
> ```

Yeah this is a great suggestion. Felt funny deleting/removing code to modify the behavior as opposed to setting it to be conditional. I have since updated with your suggestion. Let me know what you think!

calvn (approved):
Looks great, thanks for working on this!

PR #14957, opened 2022-04-07
Add build date

ccapurso:
**Core Vault Changes**
The concept of a `BuildDate` has been added to the `version` package. This will ultimately be used for Vault Enterprise licensing but has also been added to existing endpoints and commands as follows:

- `sys/seal-status` now includes a `build_date` field
- `sys/version-history` now includes a `build_date` field _per_ version entry
- `vault version` will denote the build date of the CLI binary, for example `Vault v1.11.0-dev1 ('e07f614b1493c70a76591abcbb84424682ad8adf+CHANGES'), built 2022-04-07T13:35:26Z`
- `vault status` (which uses `sys/seal-status`) will include **Build Date** when using `--format=table` and a `build_date` field when using `--format=json`
- `vault version-history` (which uses `sys/version-history`) will include a **Build Date** when using `--format=table`. Using `--format=json` will output the raw `sys/version-history` response accounting for that endpoint's modified response outlined above.

The `sys/version-history` endpoint is backed by the `core/versions` storage entries. These entries are stored as JSON so accounting for the build date simply required providing an extra key. Upon startup, these entries are cached. The cache used to be a `map[string]time.Time` but has been modified in a future-proof way to allow for information other than the installation time.

**Build Tooling Changes**
- `scripts/build_date.sh` was added to fetch the RFC3339-formatted timestamp of the last commit to provide a stable build date across platforms
- `scripts/build.sh` has been modified to provide a build flag via `-ldflags` for `github.com/hashicorp/vault/sdk/version.BuildDate` using the `build_date.sh` script noted above
- `scripts/windows/build.bat` has also been modified to set the build date via `-ldflags` but does so without a separate script
- The `build` Make target has been modified to accept a `VAULT_BUILD_DATE`
- A `get-build-date` job has been added to the GitHub build workflows which will call `scripts/build_date.sh` and provide the generated string as an output
- The `build-*` jobs in the GitHub build workflows use the `build-date` output from above to specify `VAULT_BUILD_DATE` for `make build`
- Verified CI builds by temporarily adding this branch to `branches` list in `.github/workflows/build.yml`, subsequently [created custom release](https://artifactory.hashicorp.engineering/ui/repos/tree/General/hashicorp-crt-stable-local/vault/1.11.0-custom-build-date/9a98b0741eaa5b5a2d87e8e8be67bb6e272d813d/vault_1.11.0-custom-build-date_darwin_amd64.zip) binary to verify that `BuildDate` is set properly
- Verified `scripts/windows/build.bat` changes on Windows VM

**Example output:**
```
❯ vault version
Vault v1.11.0-dev1 ('e07f614b1493c70a76591abcbb84424682ad8adf+CHANGES'), built 2022-04-07T13:35:26Z
```

```
❯ vault status
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    1
Threshold       1
Version         1.11.0-dev1
Build Date      2022-04-07T13:35:26Z
Storage Type    inmem
Cluster Name    vault-cluster-e49e8815
Cluster ID      3ff45c8c-cc98-4160-71c0-7078a8efe2ba
HA Enabled      false
```

```
❯ vault version-history

Note:
Use of this command requires a server running Vault 1.10.0 or greater.
Version tracking was added in 1.9.0. Earlier versions have not been tracked.


Version  Installation Time     Build Date
-------  -----------------     ----------
1.11.0   2022-04-12T13:37:00Z  2022-04-07T13:35:26
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
     branches:    
       # Push events on main branch
       - main
+      - add-build-date
 
 env:
   PKG_NAME: "vault"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,7 +105,7 @@ jobs:
           CGO_ENABLED: 0
         run: |
           mkdir dist out
-          GO_TAGS="${{ env.GO_TAGS }}" VAULT_VERSION=${{ needs.get-product-version.outputs.product-base-version }} VAULT_REVISION="$(git rev-parse HEAD) VAULT_BUILD_DATE="${{ needs.get-build-date.outputs.build-date }}" make build
+          GO_TAGS="${{ env.GO_TAGS }}" VAULT_VERSION=${{ needs.get-product-version.outputs.product-base-version }} VAULT_REVISION="$(git rev-parse HEAD)" VAULT_BUILD_DATE="${{ needs.get-build-date.outputs.build-date }}" make build
           zip -r -j out/${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip dist/
       - uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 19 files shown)


modrake (approved):
The build changes here look good - if you were feeling extra you could validate the build date string when it is set, but it's an unlikely failure, happy to save it for if it ever crops up outside my pessimistic take 😂

ccapurso:
> could validate the build date string

@modrake, you and I share the same pessimistic take on this 😂.. My hope is to write a release smoke test to verify that the build date is set and properly formatted.

sarahethompson (approved):
LGTM!

hghaf099 (approved):
LGTM

PR #14960, opened 2022-04-07
Add chain-building logic to PKI issuers path

cipherboy:
With the one-entry-per-issuer approach, CA Chains become implicitly
constructed from the pool of issuers. This roughly matches the existing
expectations from `/config/ca` (wherein a chain could be provided) and
`/intemediate/set-signed` (where a chain may be provided). However, in
both of those cases, we simply accepted a chain. Here, we need to be
able to reconstruct the chain from parts on disk.

However, with potential rotation of roots, we need to be aware of
disparate chains. Simply concating together all issuers isn't
sufficient. Thus we need to be able to parse a certificate's Issuer and
Subject field and reconstruct valid (and potentially parallel)
parent<->child mappings.

This attempts to handle roots, intermediates, cross-signed
intermediates, cross-signed roots, and rotated keys (wherein one might
not have a valid signature due to changed key material with the same
subject).

--

This needs tests :-)


builtin/logical/pki/cert_util.go
```diff
@@ -1513,3 +1513,288 @@ func stringToOid(in string) (asn1.ObjectIdentifier, error) {
 	}
 	return ret, nil
 }
+
+func rebuildIssuersChains(ctx context.Context, s logical.Storage, referenceCert *issuer /* optional */) error {
+	// This function rebuilds the CAChain field all known issuers. This
```
cipherboy, line 1518: typo.

builtin/logical/pki/cert_util.go
```diff
@@ -1513,3 +1513,288 @@ func stringToOid(in string) (asn1.ObjectIdentifier, error) {
...
+		}
+
+		issuerIdCertMap[identifier] = cert
+		subjectIssuerIdsMap[string(cert.RawSubject)] = append(subjectIssuerIdsMap[string(cert.RawSubject)], identifier)
+	}
+
+	// Now that we have the subj->issuer map built, we can build the parent
+	// and child mappings. We iterate over all issuers and build it one step
+	// at a time.
+	for _, child := range issuers {
```
cipherboy, line 1623: This is worst case `O(n^2)` because all of the issuers could have the same name and be self-signed certs. That makes the chain building (below) fast.

builtin/logical/pki/cert_util.go
```diff
@@ -1513,3 +1513,288 @@ func stringToOid(in string) (asn1.ObjectIdentifier, error) {
...
+	// > compatibility, all implementations SHOULD be prepared to handle
+	// > potentially extraneous certificates and arbitrary orderings from any
+	// > TLS version, with the exception of the end-entity certificate which
+	// > MUST be first.
+	//
+	// So, we take this to mean we should build chains via DFS: each issuer is
+	// explored until an empty parent pointer (i.e., self-loop) is reached and
+	// then latest-most duplicate parent is then explored.
+	//
+	// However, we don't actually need to do a DFS (per issuer) here. We can
```
cipherboy, line 1691: I think just means it is a topographical sort, making it `O(|V| + |E|)`, which means the entire algorithm is `O(n^2)` in the number of certificates (either by the toposort or by the above construction).

builtin/logical/pki/cert_util.go
```diff
@@ -1513,3 +1514,371 @@ func stringToOid(in string) (asn1.ObjectIdentifier, error) {
...
+	sort.SliceStable(issuers, func(i, j int) bool {
+		return issuers[i] < issuers[j]
+	})
+
+	// We expect each of these maps to be the size of the number of issuers
+	// we have (as we're mapping from issuers to other values).
+	//
+	// The first caches the storage entry for the issuer, the second caches
+	// the parsed *x509.Certificate of the issuer itself, and the third maps
+	// that certificate back to the other issuers with that subject (note the
```
cipherboy, line 1585: Comment needs to be updated to describe all four maps here :P

builtin/logical/pki/cert_util.go
```diff
@@ -1513,3 +1514,371 @@ func stringToOid(in string) (asn1.ObjectIdentifier, error) {
...
+	issuerIdParentsMap := make(map[issuerId][]issuerId, len(issuers))
+	issuerIdChildrenMap := make(map[issuerId][]issuerId, len(issuers))
+
+	// For every known issuer, we map that subject back to the id of issuers
+	// containing that subject. This lets us build our issuerId -> parents
+	// mapping efficiently. Worst case we'll have a single linear chain where
+	// every entry has a distinct subject.
+	subjectIssuerIdsMap := make(map[string][]issuerId, len(issuers))
+
+	// First, ready every storage entry from disk. We'll propagate entries to
```
cipherboy, line 1598: typo `ready` -> `read`.

builtin/logical/pki/cert_util.go
```diff
@@ -1513,3 +1514,371 @@ func stringToOid(in string) (asn1.ObjectIdentifier, error) {
...
+	// simply invert the (pseudo-)directed graph, i.e., topologically sort it.
+	// Some number of certs (roots without cross-signing) lack parent issuers.
+	// These are already "done" from the PoV of chain building. We can thus
+	// iterating through the parent mapping to find entries without parents to
+	// start the sort. After processing, we can add all children and visit them
+	// if all parents have been processed.
+	//
+	// Note though, that while topographical sorting is equivalent to the DFS,
+	// we have to take care to make it a pseudo-DAG. This means handling the
+	// most common 2-star (2-clique) sub-graphs of reissued certificates,
```
cipherboy, line 1724: Presently we only handle cliques as the roots of trust. I think this is incorrect. In particular, consider the restrictions under which cliques form:

 - The subject and issuer of all certs in the clique needs to be the same, in order to generate the clique.
 - The signature needs to be valid for the parent relationship to exist. This means that all members of the clique need to use the same key material.

In particular, if there were another key using the same key material with a different Subject or Issuer, it would not be the part of the clique. If there were another key using the different key material with the same subject, it could not be in the clique as the signature (onto members of the clique) would be invalid. 

However, we do have to consider cross-signed certs (same subject as the clique but not itself part of the clique -- due to different issuer). Because of the different issuer they'd allow validation of the clique (from an non-clique root of trust), effectively making the clique an internal/terminal clique (from the purposes of validation).  

Since cliques are formed by reissuance of certs, this would concretely look like an old root being reissued (at least once, if not more). During rotation to a new key material for a new root, the old root could be cross-signed by the new root. This forms a clique (by reissuance of the old) that's not the root of trust (if the new root is imported into the mount). That suggests we probably need to handle the scenario where cliques aren't roots of trust, and thus refactor it into a method, to be called from within the `toVisit` queue. In particular, we wish to visit the clique after all external potential references into it have been processed, to ensure the chain is constructed correctly. Bounding clique size probably helps. 

This clique detection also detects two-cycles (as referenced by the circle cert example). We don't detect any larger cycles (as cycles are truly pathological and do not provide much value, if any unlike cliques which form from reissuance). Maybe the scenario where the old root cross-signs the new root, in which case we'd get a non-clique cycle of (old root (self-sig) -> new root crossed by old root -> new root (self sig) -> new root cross old root -> [repeat])

So, I think the net is that we need to go back and revise this clique detection logic and separately also add cycle detection with cross-signed roots. In handling the cycle detection, we'll still want to include all parts of the cycle (including the loop of new root back to the present value), as that'll allow devices using the new cert to validate the old root without needing to use a different issuer. Cycles need not be at the root of trust either.

Yikes!

builtin/logical/pki/cert_util.go
```diff
@@ -1513,3 +1515,1026 @@ func stringToOid(in string) (asn1.ObjectIdentifier, error) {
...
+		processedIssuers[issuer] = true
+
+		childrenCerts, ok := issuerIdChildrenMap[issuer]
+		if ok && len(childrenCerts) > 0 {
+			toVisit = append(toVisit, childrenCerts...)
+		}
+
+		// If (and only if) we're presently starved for next nodes to visit,
+		// attempt to resolve cliques and cycles again to fix that. This is
+		// because all-cycles cycle detection is as costly as
```
cipherboy, line 1824: Fix :-)

builtin/logical/pki/cert_util.go
```diff
@@ -1513,3 +1515,1039 @@ func stringToOid(in string) (asn1.ObjectIdentifier, error) {
...
+						closure = append(closure, node)
+					}
+				}
+			}
+		}
+
+		// If we lack a closure, go to the next node.
+		if len(closure) == 0 {
+			continue
+		}
```
cipherboy, line 2033: @kitography / @monoidmusician - Does this reasoning sound good?

I do not need a check here to ensure the detected closure has no external, unprocessed dependencies because the starvation condition that we use to call this method (see the `if len(toVisit) == 0` guards prior to this call) guarantee that everything that _could've_ been processed, _should've_ been. 


In particular, consider an external parent that should be processed first, _before_ this closure. This certificate would have to have an (unrelated) issuer that we may not know about, but the correct subject and public key to sign something in the cycle. Without the starvation condition, if we were just calling this randomly, we'd have to ensure that the external parent(s) of this closure was processed before the closure itself. However, because we're starved of nodes to visit (and afterwards, we'll have way too many nodes :-), we know that parent must've already been processed, and that child was unreachable because it had unprocessed parents (due to cycles, i.e., exactly what this code was attempting to prevent against).

builtin/logical/pki/chain_util.go
```diff
@@ -0,0 +1,1047 @@
...
+		// clique must include the cycle (in the CA Chain computation).
+		// However, its not sufficient to just do one and then the other:
+		// we need the closure of all cliques (and their incident cycles).
+		// Finally -- it isn't enough to consider this chain in isolation
+		// either. We need to consider _all_ parents and ensure they've been
+		// processed before processing this closure.
+
+		var cliques [][]issuerId
+		var cycles [][]issuerId
+		var closure []issuerId
```
cipherboy, line 431: Should probably make this a map, avoid unnecessary iterations for membership.

stevendpclark (approved):
There is a lot here so please have @kitography provide a review. 

The only feedback I have on this is we are most likely missing a reprocessing of CAChains if we call importKey and it updates the issuer.KeyID of an existing certificate. We call writeIssuer to do the update of that issuer but don't reprocess the chains from what I can tell.

cipherboy:
@stevendpclark Thanks! I don't think that's actually necessary; we rebuild all issuers regardless of their key presence; whether or not a private key is present doesn't actually affect whether or not we're building the chain for that issuer (we do it anyway) and the signature verification doesn't depend on key presence either (we're using the public key off of the cert itself).

builtin/logical/pki/cert_util.go
cipherboy: Per OOB discussion with @MonoidMusician, there's a counter-example here we need to take into account:

Suppose there's two cycles, with an intermediate joining them (i.e., there exists a cert with an issuer in the first cycle, which can validate a node in the second cycle).

This doesn't turn it into a larger cycle, and if the secondary cycle is processed first, its chain will be wrong (because it needs to include the former cycle). So we _do_ need to account for processing state and we _do_ need to take into account other (non-closure) parents.

cipherboy: OK, this has been updated.

cipherboy:
Manually merged, per out of band conversation with @kitography and @stevendpclark. Thanks! :-)

PR #14962, opened 2022-04-07
Warnings indicating ignored and replaced parameters

vishalnayak:


ccapurso (approved):
We probably want to add a changelog entry (maybe to denote an `improvement`?) but otherwise looks good!

sdk/framework/backend.go
```diff
@@ -275,7 +291,24 @@ func (b *Backend) HandleRequest(ctx context.Context, req *logical.Request) (*log
 		}
 	}
 
-	return callback(ctx, req, &fd)
+	resp, err := callback(ctx, req, &fd)
+	if err != nil {
+		return nil, err
+	}
+
+	// Handle the nil response case.
```
hghaf099, line 299: Seems like some tests are failing as they expect a nil response. Could this change cause breakage for the existing users that check if the resp is not nil, and if so rely on the response content? In this case, only Warnings field might be populated.

ccapurso: Ah, I missed this. I think @hghaf099 is right. There is [logic](https://github.com/hashicorp/vault/blob/6e6d17cfc18cef3c16575a1508c76d300e4a414e/sdk/logical/response_util.go#L19-L30) to convert a `nil` response for `ReadOperation` and `ListOperation` to a 404. This would likely break that, right?

ccapurso: Change looks good, thanks!

ccapurso (approved):

hghaf099 (approved):

hghaf099:
Would a changelog be needed for this change?

vishalnayak:
> Would a changelog be needed for this change?

Yes. Added one!

PR #14966, opened 2022-04-07
Fix edit capabilities call in auth method

arnav28:
- Capabilities call was not getting triggered correctly as apiPath
  method was missing the correct context.

### DIFF
--- a/changelog/14966.txt
+++ b/changelog/14966.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Fixes edit auth method capabilities issue
+```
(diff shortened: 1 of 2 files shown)


hashishaw (commented):
Nice work!

For completeness, would you make sure there aren't any other `return apiPath...` lurking inside functions? Thank you!

arnav28:
> Nice work!
> 
> For completeness, would you make sure there aren't any other `return apiPath...` lurking inside functions? Thank you!

Yup, this was the only instance. 👍

hashishaw (approved):
Thanks! Add a changelog and this is good to go

PR #14967, opened 2022-04-07
UI/Client count running totals component

hellobontempo:
### PR to add the Running total component to the client count history tab
Screen Shot 2022-04-08 at 10 46 10 AM


ui/app/serializers/clients/activity.js
```diff
@@ -34,20 +34,22 @@ export default class ActivitySerializer extends ApplicationSerializer {
 
   // for vault usage - vertical bar chart
   flattenByMonths(payload, isNewClients = false) {
+    const sortedPayload = [...payload];
+    sortedPayload.reverse();
```
hellobontempo, line 38: I initially created a method to sort months by date - but this requires iterating through and converting the timestamp to a date object, comparing each and _then_ formatting the timestamp into the `M/yy` format. 

The API always returns a chronological array of months in order from most recent to oldest. So to save computing this in the frontend opted to just reverse this order.

ui/app/serializers/clients/activity.js
```diff
@@ -88,6 +90,14 @@ export default class ActivitySerializer extends ApplicationSerializer {
     return object;
   }
 
+  // TODO CMB make these two functions utils?
+
+  // format ISO timestamp to M/yy
+  formatTimestamp(timestamp) {
```
hellobontempo, line 96: Timestamp looks like: `2021-03-21T00:00:00Z` - if you directly format this, it returns a date object associated with the user's timezone. We want to globally return the same year and month as the API, so by splitting on `T` and parsing _after_ we consistently get `3/21`

ui/app/components/clients/vertical-bar-chart.js
```diff
@@ -111,18 +121,14 @@ export default class VerticalBarChart extends Component {
...
     tooltipRect.on('mouseover', (data) => {
-      let hoveredMonth = data.month;
-      this.tooltipTotal = `${data.clients} ${data.new_clients ? 'total' : 'new'} clients`;
-      this.uniqueEntities = `${data.entity_clients} unique entities`;
-      this.nonEntityTokens = `${data.non_entity_clients} non-entity tokens`;
-      // let node = chartSvg
-      //   .selectAll('rect.tooltip-rect')
-      //   .filter(data => data.month === this.hoveredLabel)
-      //   .node();
+      let hoveredMonth = data[this.xKey];
```
zofskeez, line 128: I'm noticing `let` is used for variable assignment in this file but am wondering if any of the variables would ever be reassigned? If not I think `const` would be a better choice to signal that the variables will not be reassigned. This is probably a developer preference type of thing but wanted to offer my thoughts.

ui/app/routes/vault/cluster/clients/history.js
```diff
@@ -2,6 +2,7 @@ import Route from '@ember/routing/route';
 import { isSameMonth } from 'date-fns';
 import RSVP from 'rsvp';
 import getStorage from 'vault/lib/token-storage';
+import { parseRFC3339 } from '../../../../utils/date-formatters';
```
zofskeez, line 5: Rather than moving up the tree you could use the vault alias for this import:
`import { parseRFC3339 } from 'vault/utils/date-formatters';`

ui/app/serializers/clients/activity.js
```diff
@@ -88,13 +91,6 @@ export default class ActivitySerializer extends ApplicationSerializer {
     return object;
   }
 
-  parseRFC3339(timestamp) {
-    // convert '2021-03-21T00:00:00Z' --> ['2021', 2] (e.g. 2021 March, month is zero indexed)
-    return timestamp
-      ? [timestamp.split('-')[0], Number(timestamp.split('-')[1].replace(/^0+/, '')) - 1]
-      : null;
-  }
-
```
zofskeez, line 97: Nice moving this to a util!

ui/mirage/utils/date-formatters.js
```diff
@@ -0,0 +1,20 @@
...
+// convert ISO timestamp '2021-03-21T00:00:00Z' to ['2021', 2]
+// (e.g. 2021 March, month is zero indexed) (used by calendar widget)
+export const parseRFC3339 = (timestamp) => {
+  if (Array.isArray(timestamp)) {
+    // return if already formatted correctly
+    return timestamp;
+  }
+  let date = parseAPITimestamp(timestamp);
+  return date ? [`${date.getFullYear()}`, date.getMonth()] : null;
+};
```
zofskeez, line 20: It looks like this a duplicate of the util file. Was this intentional? You should be able to import the util file in the tests or mirage handlers if needed.

ui/app/components/clients/vertical-bar-chart.js
hellobontempo: Nope - i think `const` would be preferable here!

ui/mirage/utils/date-formatters.js
hellobontempo: I had trouble importing the util into the mirage handler?? Maybe it was a typo - but I'd definitely prefer that over duplicating the whole file

ui/app/routes/vault/cluster/clients/history.js
hellobontempo: Ah yes - thank you! It auto-imported like this and I didn't even think to clean it up!

ui/tests/acceptance/client-history-test.js
```diff
@@ -196,7 +196,7 @@ module('Acceptance | clients history tab', function (hooks) {
     assert.dom('[data-test-stat-text="total-clients"] .stat-value').hasText('15');
     assert.dom('[data-test-stat-text="entity-clients"] .stat-value').hasText('5');
     assert.dom('[data-test-stat-text="non-entity-clients"] .stat-value').hasText('10');
-    await waitUntil(() => find('[data-test-horizontal-bar-chart]'));
+    await settled();
```
hellobontempo, line 199: waitUntil kept timing out so going to try removing it..

zofskeez (approved):
Very cool thanks for this!

PR #14968, opened 2022-04-07
Don't clone OutputCurlString value

digivava:
The OutputCurlString field comment claims that the field is not copied over as part of cloning the client, when in fact, it currently is.

This fixes that bug and also adds mention of this extremely useful but hard-to-discover flag to the CLI command overview.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -588,7 +588,6 @@ func (c *Client) CloneConfig() *Config {
 	newConfig.CheckRetry = c.config.CheckRetry
 	newConfig.Logger = c.config.Logger
 	newConfig.Limiter = c.config.Limiter
-	newConfig.OutputCurlString = c.config.OutputCurlString
 	newConfig.SRVLookup = c.config.SRVLookup
 	newConfig.CloneHeaders = c.config.CloneHeaders
 	newConfig.CloneToken = c.config.CloneToken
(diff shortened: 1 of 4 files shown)


api/client.go
```diff
@@ -1002,22 +1001,21 @@ func (c *Client) clone(cloneHeaders bool) (*Client, error) {
...
+		Timeout:        config.Timeout,
+		Backoff:        config.Backoff,
+		CheckRetry:     config.CheckRetry,
+		Logger:         config.Logger,
+		Limiter:        config.Limiter,
+		AgentAddress:   config.AgentAddress,
+		SRVLookup:      config.SRVLookup,
+		CloneHeaders:   config.CloneHeaders,
+		CloneToken:     config.CloneToken,
+		ReadYourWrites: config.ReadYourWrites,
```
VinnyHC, line 1018: Nit: These lines look identical, not sure why they're in the changelog, but if they all are intended to be changed you could alphabetize them for easier reading.

VinnyHC (approved):
Good catch on this!

api/client.go
digivava: It's the removal of `OutputCurlString: config.OutputCurlString`, I feel like I've seen this happen in git sometimes when I remove things from a struct though I don't know why. I think the order is meant to reflect the order that they're declared in the struct.

PR #14969, opened 2022-04-07
updating legacy MFA doc with link to Login MFA

hghaf099:


ccapurso (approved):

hghaf099:
> I approved but one question - I notice that this is a PR against the 1.10 release branch. Should we instead push to main and then backport to 1.10?

Thanks for taking a look at this PR. Actually, the feature is deprecated in main, so no need to update anything in there.

PR #14970, opened 2022-04-08
Fix nightly tests and debug

akshya96:
Making a build on windows failed as syscall is undefined on windows.
Created a umask function in osutil and used it in debug.go
Also, setting an env variable to disable file  permissions check did not work and nightly tests still failed, used umask to create files with required permissions here.


swayne275 (commented):
Can you explain more about the bit where a previous file permissions check disable didn't work?

If a previous one didn't work, should we remove that here?

command/operator_diagnose_test.go
```diff
@@ -478,7 +479,7 @@ func TestOperatorDiagnoseCommand_Run(t *testing.T) {
 				t.Parallel()
 				client, closer := testVaultServer(t)
 				defer closer()
-
+				os.Setenv(consts.VaultDisableFilePermissionsCheckEnv, "true")
```
ccapurso, line 482: I'm not sure that I can provide any feedback on whether or not this makes sense in the context of this specific test. A comment describing why we're disabling the file permissions check might be useful.

akshya96:
> Can you explain more about the bit where a previous file permissions check disable didn't work?
> 
> If a previous one didn't work, should we remove that here?

Kept the previous file permissions check disable.

PR #14974, opened 2022-04-08
Update two packages for upcoming Ember upgrade

Monkeychip:
Pulling this into its own PR. Two packages that needed to be upgraded:
1. hashicorp-flight-icons
2. ember-basic-dropdown

Both upgrades cleared the test suite and I did some manual testing. However, it's nice to have these separated out from the large PR in case folks notice any odd changes before the upgrade. This way we can trace it back to this PR.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -63,7 +63,7 @@
     "@ember/test-waiters": "^3.0.0",
     "@glimmer/component": "^1.0.3",
     "@glimmer/tracking": "^1.0.3",
-    "@hashicorp/ember-flight-icons": "^1.1.0",
+    "@hashicorp/ember-flight-icons": "2.0.3",
     "@hashicorp/structure-icons": "^1.3.0",
     "@icholy/duration": "^5.1.0",
     "@storybook/cli": "^6.3.10",
@@ -93,7 +93,7 @@
     "doctoc": "^1.4.0",
     "ember-api-actions": "^0.2.8",
     "ember-auto-import": "^1.10.1",
-    "ember-basic-dropdown": "3.0.19",
+    "ember-basic-dropdown": "4.0.5",
     "ember-cli": "~3.24.0",
     "ember-cli-autoprefixer": "^0.8.1",
     "ember-cli-babel": "^7.23.0",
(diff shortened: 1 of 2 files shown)


zofskeez:
I can't seem to find a 2.0 release for the flight icons on the design systems repo but I saved an email from a while back that mentioned a breaking change:

_Notice: we have just released version 2.0.0 of the FlightIcons packages (both @hashicorp/flight-icons and  @hashicorp/ember-flight-icons) that contains a breaking change: the old icon vault has been renamed bank-vault to avoid conflicts with the newly released “product” icons (which contains a “vault” logo icon). When you update the version please make sure the old vault icon is not used in your codebase, otherwise please update it to the new name. Thanks_

There's a `vault-logo.svg` in the public folder that we likely use but wanted to call this out just in case there are conflicts.

ui/package.json
```diff
@@ -193,7 +193,7 @@
   "resolutions": {
     "cryptiles": "^4.1.2",
     "eslint-utils": "^1.4.1",
-    "ember-basic-dropdown": "3.0.19",
+    "ember-basic-dropdown": "4.0.5",
```
zofskeez, line 196: Do we need this anymore? I can't remember for sure but I recall that this was necessary for `ember-basic-dropdown-hover` which has been removed on the 3.28 branch. Would these updates be more relevant off of that branch rather than main?

Monkeychip:
> I can't seem to find a 2.0 release for the flight icons on the design systems repo but I saved an email from a while back that mentioned a breaking change:
> 
> _Notice: we have just released version 2.0.0 of the FlightIcons packages (both @hashicorp/flight-icons and @hashicorp/ember-flight-icons) that contains a breaking change: the old icon vault has been renamed bank-vault to avoid conflicts with the newly released “product” icons (which contains a “vault” logo icon). When you update the version please make sure the old vault icon is not used in your codebase, otherwise please update it to the new name. Thanks_
> 
> There's a `vault-logo.svg` in the public folder that we likely use but wanted to call this out just in case there are conflicts.

Great catch. Here is the release schedule. It was confusing to me to. It's mixed in with the design system releases. https://github.com/hashicorp/design-system/releases?page=1 I'll look for the icon. Right now nothing looks off with the vault icons, but I'll check the code.

ui/package.json
Monkeychip: Interesting. Let me run these updates on the 3.28 branch and see if i hit any conflicts.

Monkeychip:
Closing for this PR to sidebranch https://github.com/hashicorp/vault/pull/14976/files

PR #14975, opened 2022-04-08
Support root issuer generation

stevendpclark:
Add initial very basic support for creating issuers through the new /issuers/generate/root api.

- This needs some further refinement/refactoring but I'd like to get some early feedback on the approach. 
- Testing at the moment for the existing is a bit lacking, more to come.
- This is just for root at the moment, not the intermediate endpoint that will most likely be another PR.

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -4656,6 +4656,23 @@ func TestRootWithExistingKey(t *testing.T) {
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "unable to find PKI key for reference: my-key1")
 
+	// Fail if the specified key name is default.
+	_, err = client.Logical().WriteWithContext(ctx, "pki-root/issuers/generate/root/internal", map[string]interface{}{
+		"common_name": "root myvault.com",
+		"issuer_name": "my-issuer1",
+		"key_name":    "Default",
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "reserved keyword 'default' can not be used as key name")
+
+	// Fail if the specified issuer name is default.
+	_, err = client.Logical().WriteWithContext(ctx, "pki-root/issuers/generate/root/internal", map[string]interface{}{
+		"common_name": "root myvault.com",
+		"issuer_name": "DEFAULT",
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "reserved keyword 'default' can not be used as issuer name")
+
 	// Create the first CA
 	resp, err := client.Logical().WriteWithContext(ctx, "pki-root/issuers/generate/root/internal", map[string]interface{}{
 		"common_name": "root myvault.com",
@@ -4669,11 +4686,20 @@ func TestRootWithExistingKey(t *testing.T) {
 	require.NotEmpty(t, myIssuerId1)
 	require.NotEmpty(t, myKeyId1)
 
+	// Fail if the specified issuer name is re-used.
+	_, err = client.Logical().WriteWithContext(ctx, "pki-root/issuers/generate/root/internal", map[string]interface{}{
+		"common_name": "root myvault.com",
+		"issuer_name": "my-issuer1",
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "issuer name already used")
+
 	// Create the second CA
 	resp, err = client.Logical().WriteWithContext(ctx, "pki-root/issuers/generate/root/internal", map[string]interface{}{
 		"common_name": "root myvault.com",
 		"key_type":    "rsa",
 		"issuer_name": "my-issuer2",
+		"key_name":    "root-key2",
 	})
 	require.NoError(t, err)
 	require.NotNil(t, resp.Data["certificate"])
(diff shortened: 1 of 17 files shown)


builtin/logical/pki/storage.go
```diff
@@ -518,6 +525,56 @@ func fetchCertBundleByIssuerId(ctx context.Context, s logical.Storage, id issuer
...
+		if _, _, err = importIssuer(ctx, s, cert, ""); err != nil {
+			return nil, nil, err
+		}
+	}
+
+	keyDefaultSet, err := isKeyDefaultSet(ctx, s)
+	if err != nil {
+		return nil, nil, err
+	}
+	if len(allKeyIds) == 0 || !keyDefaultSet {
```
cipherboy, line 559: Should this just be handled by `import`? It seems cleaner than delegating to `writeBundle` logic, which won't be used in the `/import` paths for instance.

builtin/logical/pki/path_root.go
```diff
@@ -26,27 +26,7 @@ import (
...
-		HelpSynopsis:    pathGenerateRootHelpSyn,
-		HelpDescription: pathGenerateRootHelpDesc,
-	}
-
-	ret.Fields = addCACommonFields(map[string]*framework.FieldSchema{})
-	ret.Fields = addCAKeyGenerationFields(ret.Fields)
-	ret.Fields = addCAIssueFields(ret.Fields)
-
-	return ret
+	return commonGenerateRoot(b, "root/generate/"+framework.GenericNameRegex("exported"))
```
cipherboy, line 29: I was (trying) to use the `buildPath...` naming scheme here. Either `common` or `buildPath` makes sense to me, but we should probably use one and not both. :-)

cipherboy (commented):
Looking good :-) Some minor comments.

builtin/logical/pki/path_root.go
stevendpclark: Oops I hadn't noticed, I'll switch to buildPath...

builtin/logical/pki/storage.go
stevendpclark: Yup looking at this you are correct, it's been moved into the proper import methods.

builtin/logical/pki/backend_test.go
```diff
@@ -4613,6 +4613,213 @@ func TestBackend_Roles_KeySizeRegression(t *testing.T) {
...
+	require.NotEmpty(t, myKeyId1)
+
+	// Fail if the specified issuer name is re-used.
+	_, err = client.Logical().WriteWithContext(ctx, "pki-root/issuers/generate/root/internal", map[string]interface{}{
+		"common_name": "root myvault.com",
+		"issuer_name": "my-issuer1",
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "issuer name already used")
+
```
kitography, line 4696: what happens if we try generating a root credential with a name which is the Issuer's Key ID? (I'm hoping that this would fail)

builtin/logical/pki/backend_test.go
```diff
@@ -4613,6 +4613,213 @@ func TestBackend_Roles_KeySizeRegression(t *testing.T) {
...
+		"key_ref":     myKeyId1,
+	})
+	require.NoError(t, err)
+	// csr3 := resp.Data["csr"]
+	myKeyId3 := resp.Data["key_id"]
+	require.NotEmpty(t, myKeyId3)
+
+	require.NotEqual(t, myKeyId1, myKeyId2)
+	require.Equal(t, myKeyId1, myKeyId3)
+}
```
kitography, line 4821: key name reusue?  issuer name reuse (for non roots?)

builtin/logical/pki/storage_test.go
```diff
@@ -102,48 +102,53 @@ func Test_KeysIssuerImport(t *testing.T) {
...
 	issuer1.KeyID = ""
 
-	key1_ref1, existing, err := importKey(ctx, s, key1.PrivateKey)
+	key1_ref1, existing, err := importKey(ctx, s, key1.PrivateKey, "key1")
 	require.NoError(t, err)
 	require.False(t, existing)
 	require.Equal(t, key1.PrivateKey, key1_ref1.PrivateKey)
 
-	key1_ref2, existing, err := importKey(ctx, s, key1.PrivateKey)
+	key1_ref2, existing, err := importKey(ctx, s, key1.PrivateKey, "ignore-me")
```
kitography, line 110: what are we trying to test here?  Like, this is a test on a helper function, what are we trying to guarantee to the user?

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -61,25 +61,17 @@ their identifier and their name (if set).
 )
 
 func pathGetIssuer(b *backend) *framework.Path {
-	pattern := "issuer/" + framework.GenericNameRegex("ref") + "(/der|/pem)?"
+	pattern := "issuer/" + framework.GenericNameRegex("issuer_ref") + "(/der|/pem)?"
```
kitography, line 64: nit: the number of places this is referred to, i would love to see a constant

kitography (commented):
This is pretty damn good, super impressed by some of the generate tests, have quite a few nits (though not nearly as many as expected on 721 lines of code)

builtin/logical/pki/config_util.go
```diff
@@ -2,10 +2,29 @@ package pki
 
 import (
 	"context"
+	"strings"
 
 	"github.com/hashicorp/vault/sdk/logical"
 )
 
+func isKeyDefaultSet(ctx context.Context, s logical.Storage) (bool, error) {
```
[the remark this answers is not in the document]
stevendpclark: Definitely not an idiot, it was badly named. Updating to isDefaultKeySet and isDefaultIssuerSet in a hope to make this clearer.

builtin/logical/pki/backend_test.go
stevendpclark: Yes that would fail if an end-user were to try to use an existing id for a Name value. Now if they could predict the uuid we would generate in the future though, that would be a bad day...

builtin/logical/pki/backend_test.go
stevendpclark: Validating that the new CA we generated re-used a key that we had previous generated. Will add a clearer message on this test.

builtin/logical/pki/storage_test.go
stevendpclark: The "contract" as far as I understand it, is if we determine that the key/issuer already exists no updates occur from the importXXX functions and that the existing flag is set to true.

cipherboy:
Manually merging :-)

cipherboy:

PR #14976, opened 2022-04-08
Upgrade dropdown and icons

Monkeychip:
I had to delete Yarn lock and rebuild in order to avoid some build/broccoli failures.


zofskeez (approved):
Thanks! Interesting looking test failure. Maybe try and re-run from failed?

Issue #14535, opened 2022-03-16
Okta authentication MFA_REQUIRED case does not recognize GOOGLE provider TOTP type


wenslayer:
**Describe the bug**
When authenticating via Okta, Vault does not recognize the Google provider TOTP type.

**To Reproduce**
1. As a user in Okta, configure your "extra verification" to configure Google Authenticator.
2. Run `vault login -method=okta username=my.okta.username`
3. Get the following error:

```text
Error authenticating: Error making API request.

URL: PUT https://vault.my.domain/v1/auth/okta/login/my.okta.username
Code: 400. Errors:

* Okta Verify Push or TOTP factor is required in order to perform MFA
```

**Expected behavior**
The flow of being prompted for the TOTP.

**Environment:**
* Vault Server Version: 1.9.3
* Vault CLI Version: 1.9.4
* Server Operating System/Architecture: macOS 12.3

Vault server configuration file(s): 

```json
      {
        "storage": {
          "s3": {
            "bucket": "my.vault.bucket",
            "region": "${AWS_REGION}"
          }
        },
        "ha_storage": {
          "consul": {
            "address": "127.0.0.1:8500",
            "path": "vault"
          }
        },
        "service_registration": {
          "consul": {
            "address": "127.0.0.1:8500"
          }
        },
        "listener": {
          "tcp": {
            "address": "127.0.0.1:8200",
            "tls_disable": true
          }
        },
        "listener": {
          "tcp": {
            "address": "${IPV4}:8200",
            "tls_cert_file": "/etc/vault/certs/vault-pub.pem",
            "tls_key_file": "/etc/vault/certs/vault-priv.pem",
            "tls_min_version": "tls12"
          }
        },
        "ui": true,
        "api_addr": "https://${IPV4}:8200",
        "cluster_addr": "https://${IPV4}:8201"
      }
```

**Additional context**
It appears this code block is simply bypassing any provider that is NOT "OKTA", when it could accept the case when the provider is "GOOGLE" and the type matches `mfaTOTPMethod`.
* https://github.com/hashicorp/vault/blob/v1.9.4/builtin/credential/okta/backend.go#L182-L191

I see in the [Okta API documentation for supported factors for providers](https://developer.okta.com/docs/reference/api/factors/#supported-factors-for-providers):

Provider | Factor Type
-- | --
...
GOOGLE | token:software:totp
...
OKTA | token:software:totp

wenslayer:
[...]

PR #14985, opened 2022-04-08
supporting google authenticator with Okta auth

hghaf099:
When authenticating via Okta, Vault does not recognize the Google provider TOTP type.
Addresses https://github.com/hashicorp/vault/issues/14535


raskchanky (commented):
The change itself looks ok. Do you mind adding some tests to cover this new functionality?

hghaf099:
> The change itself looks ok. Do you mind adding some tests to cover this new functionality?

I have tested it manually, as it would need Okta credentials. I will push the manual script into the vault-tools repo shortly. Do you still think a GO test would be needed?

builtin/credential/okta/backend.go
```diff
@@ -191,8 +196,9 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username, pas
 			}
 		}
 
-		// Okta push and totp are currently supported. If a totp passcode is provided during
-		// login and is supported, that will be the preferred method.
+		// Okta push and totp, and Google totp are currently supported.
+		// If a totp passcode is provided during login and is supported,
+		// that will be the preferred method.
 		switch {
```
calvn, line 202: This switch case also handles push factor, but the Google provider [doesn't support that](https://developer.okta.com/docs/reference/api/factors/#supported-factors-for-providers). Should we update this switch so that push factor can only be set if the provider is Okta?

hghaf099: Sure, I can add a check for the provider to only use push factor for Okta.

changelog/14985.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+auth:When authenticating via Okta, Vault does not recognize the Google provider TOTP type
```
calvn, line 2:
```suggestion
auth/okta: Add support for Google provider TOTP type in the Okta auth method
```

builtin/credential/okta/backend.go
```diff
@@ -179,7 +180,11 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username, pas
 		for _, v := range result.Embedded.Factors {
 			v := v // create a new copy since we'll be taking the address later
 
-			if v.Provider != "OKTA" {
+			if preferredProvider != "" && preferredProvider != v.Provider {
+				continue
+			}
```
calvn, line 185: Is this check necessary, or does the `strutil.ListContains` call below handle this already (both the empty provider case and the matching provider checks)?

hghaf099: It is needed. Suppose I enabled Google totp, and Okta push and totp on my account. Then, this loop goes through an array containing all three enabled providers methods as the `result.Embedded.Factors` is populated by the third party Okta api.  In this case, suppose I pass in the Google totp to the login command. However, when the loop ends, the totp factor could be populated by the okta totp. Then, MFA validation fails since I have passed in the Google totp code. The reason I introduced the `provider` in the command line is this. Cause without it, there is no way to distinguish whether the totp code is for Okta or Google.
The below `strutil.ListContains` just makes sure that we only accept Google and Okta as there are more providers than these two as mentioned in the Okta documentation.

calvn: Ah, yes that makes sense! Somehow I overlooked where `preferredProvider` is coming from.

calvn (approved):
One last nit: Can we also update the docs page to make a mention of the new `provider` parameter that's present on the login endpoint?

hghaf099:
Should this be backported as well?

calvn:
This is more of an enhancement/improvement than a bug, so I'd think that it doesn't need to be backported.

builtin/credential/okta/backend.go
wenslayer: Does this mean that if I only have Google TOTP enabled, that I'd have to explicitly set Google as the provider? Because that seems like an unworkable situation -- our org has folks with a mix of either Okta Verify or Google TOTP (but not Okta TOTP), and we can't know in advance which provider to use.

Insisting that the provider be specified should only be required if both Okta and Google TOTP are present.

hghaf099: No, if you only have Google TOTP enabled in your Okta account, you don't need to pass in the provider, as  `result.Embedded.Factors` would only have one element. The problem raises when in a single Okta account, both Okta TOTP and Google TOTP are enabled as possible providers. Here, you would need to pass in the provider. Note that if you have enabled Okta Push and Google TOTP, you would NOT need to pass in the provider.

Issue #10140, opened 2020-10-13
Provide a query parameter in UI to allow specifying mount in login UI


shwuandwing:
**Is your feature request related to a problem? Please describe.**
Today, the Vault login url can take a query parameter specifying the login method.  e.g.  https://vault.com/ui/vault/auth?with=jwt     .  This lets people write documentation and share links to the Vault UI that has the jwt as the login method populated.   There is a query parameter to support prefilling the namespace as well for Enterprise.

We like a query parameter in the url to specify the mount path as well (for customers who create auth mounts at a custom path).   Curently, we have to share a url to the with jwt specified but document to users to specify the mount path

**Describe the solution you'd like**
A query parameter to specify the mount path

martincostello:
Similarly, I'd also like a query string parameter that can be used for the role name for a JWT/OpenID connection. Maybe there's scope to look at prefill parameters _in general_, rather than just specifically for adding mount path and this?

remilapeyre:
Hi, I opened a PR to fix this issue: https://github.com/hashicorp/vault/pull/14988

hsimon-hashicorp:
The linked PR was closed due to inactivity, so I wanted to come by and give a (slightly belated, apologies) update. Our engineers are planning on making some changes and improvement to the auth forms, which will hopefully help cover some of these pain points. We have linked this GitHub issue to our internal Jira tracking ticket, so if this gets closed, check the release notes to see what made it in. Thanks so much for your patience!

marcportabellaclotet-mt:
Any update on this topic?

hellobontempo:
Apologies for the delay here. This behavior actually already exists, admittedly the UX is confusing and we're working on improving the workflow as well as documentation

1. Tune the mount to have `listing_visibility="unauth"`. This can be done using the CLI or UI ([docs](https://developer.hashicorp.com/vault/api-docs/system/auth#listing_visibility-1)). This makes the custom path known to the web UI _prior_ to authentication.

<img width="700" alt="Image" src="https://github.com/user-attachments/assets/c5bb2ca7-31df-41e3-92d1-a8d2e1ea5b09" />

2. When tab is selected the `with=` query param designates the custom mount path

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/5fecbaf0-af92-4001-bc37-c085aac095de" />

PR #14988, opened 2022-04-10
Add the possibility to set the mount path for login in query params

remilapeyre:
This patch adds the possibility to set the mount path used for logging
in the UI using query params. When the mount path is given that way the
drawer is already open on page load.

Closes https://github.com/hashicorp/vault/issues/10140

### DIFF
--- a/changelog/14988.txt
+++ b/changelog/14988.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+ui: The mount path used when logging in Vault can now be specified with the
+`path` query param, e.g. https://vault.com/ui/vault/auth?with=jwt&path=my_custom_method.
+```
--- a/ui/app/components/auth-form-options.js
+++ b/ui/app/components/auth-form-options.js
@@ -1,3 +1,8 @@
 import Component from '@ember/component';
 
-export default Component.extend({});
+export default Component.extend({
+  init() {
+    this._super(...arguments);
+    this.set('isOpen', this.customPath !== '');
+  },
+});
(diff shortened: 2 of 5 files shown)


hghaf099:
@remilapeyre Thank you for submitting this PR. Would you please create a GH issue for this PR as well if there is none, and link the two? Also, please add a changelog.

remilapeyre:
Hi @hghaf099, I found an issue matching this new feature and added an entre for the changelog.

jgutierrezglez:
This improvement would be really useful for us - @remilapeyre would be it possible to resolve the conflicts and re-submit this PR for review again? Thanks

hsimon-hashicorp:
Hello, Remi! I spoke with our engineers, and I think we need to re-consider this approach. This PR changes the current behavior which we want to keep. That is, when you change the mount type, it should clear the mount path. Currently with their changes it keeps the mount path because it's in the query param. If you can figure out how to keep the existing clear-path-on-change behavior we would consider it!
Please let me know how you'd like to proceed here. Would you rather close this PR and file a new one with the changes requested? Thanks in advance!

remilapeyre:
I don't think a new PR will be necessary, I had not though about this part which obviously makes sense. I will look into this and update the PR.

hellobontempo:
Hi there 👋 I noticed this PR has been stale for about 6 months, are you still planning to follow-up with the requested changes? We'll close this PR in about a month if there are no updates.

thomas-maurice:
Having the ability to specify the mount path in the URL would make my life so much easier !

zofskeez:
Closing due to inactivity

PR #14994, opened 2022-04-11
Fix handling of SignatureBits for ECDSA issuers (#14943)

cipherboy:
Backport of #14943:

When adding SignatureBits control logic, we incorrectly allowed
specification of SignatureBits in the case of an ECDSA issuer. As noted
in the original request, NIST and Mozilla (and others) are fairly
prescriptive in the choice of signatures (matching the size of the
NIST P-curve), and we shouldn't usually use a smaller (or worse, larger
and truncate!) hash.

Ignore the configuration of signature bits and always use autodetection
for ECDSA like ed25519.

---

This definitely warrants a changelog entry, but not sure if we need anything else? This was added in 1.10.0, we'll backport this to 1.10.1, fixing the issue. It only presently works on roles (it should probably be added to the other endpoints later?). 

Ticket: VAULT-5707

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -4245,8 +4245,11 @@ func TestBackend_Roles_IssuanceRegression(t *testing.T) {
 }
 
 type KeySizeRegression struct {
-	RoleKeyType       string
-	RoleKeyBits       []int
+	// Values reused for both Role and CA configuration.
+	RoleKeyType string
+	RoleKeyBits []int
+
+	// Signature Bits presently is only specified on the role.
 	RoleSignatureBits []int
 
 	// These are tuples; must be of the same length.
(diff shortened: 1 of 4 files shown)


stevendpclark (approved):

cipherboy:
Thanks @stevendpclark!

martin31821:
Hi @cipherboy. 

We're using Vault to manage Certificates in an embedded-secureboot scenario and unfortunately, this PR broke our workflow. 
The actual hardware only supports SHA256 signatures on any EC curve (we're using secp521). 

Is there any way, the possibility to override the default behavior could be reintroduced?
I agree that for the overwhelming majority of the people this behavior is totally fine, but I would still like to see the possibility to override the setting.

Maybe there should be an enforcement in place that rejects the SHA512/256 downsampling problem?

PR #14999, opened 2022-04-11
Fix /intermediate/set-signed, /config/ca import and other minor test scenarios

cipherboy:
Based on top of #14975 and #14960; will be rebased after the former merges. Only the top 3 commits are new. 

---

We make three changes:

 - Update the two import paths to reuse the issuer-import logic (`/config/ca` and `/intermediate/set-signed`). 
 - Fix the error message on missing default issuer.
 - Fix tests related to imports &c to handle the new semantics.

We'll still need to update the root deletion code as well.


stevendpclark (approved):
lgtm!

cipherboy:
Thanks! Manually merging...

cipherboy:

PR #15001, opened 2022-04-11
Remove ModelWrap Component

zofskeez:
This PR addresses a bug in kmip views which used the _ModelWrap_ component. After the Ember 3.28 upgrade the concurrency task in the component was being cancelled and the success callback was not firing. Given it's limited use the component was removed and controllers were added for necessary views.


Monkeychip (approved):
Nice job on a quick fix!

PR #15003, opened 2022-04-12
Introduce defaultRef constant within PKI

stevendpclark:
- Replace hardcoded "default" references with a constant to easily identify various usages.
 - Use the addIssuerRefField function instead of redefining the field in various locations.

Builds on top of #14975


kitography (approved):
Awesome!  Also removing lines, even more awesome.

cipherboy:

PR #15004, opened 2022-04-12
Add support for deleting all keys, issuers

cipherboy:
The old `DELETE /root` code must now delete all keys and issuers for
backwards compatibility. We strongly suggest calling individual delete
methods (`DELETE /key/:key_ref` or `DELETE /issuer/:issuer_ref`) instead,
for finer control.

In the process, we detect whether the deleted key/issuers was set as the
default. This will allow us to warn (from the single key/deletion issuer
code) whether or not the default was deleted (while allowing the
operation to succeed).


cipherboy:
Thanks @stevendpclark! Manually merged...

cipherboy:

PR #15005, opened 2022-04-12
Remove duplicated issuers in full chain

cipherboy:
This fixes full chain construction so that we don't accidentally introduce duplicate issuers into the chain. This also removes our incorrect setting of the legacy `IssuingCA` field, which was used prior to `CAChain` introduction.

Strictly, I believe this is caused by `CAChain` containing not just the strict issuer hierarchy but also the leaf cert. I believe this makes the most sense from a bundle perspective (at the cost of a duplication), but perhaps it is worth trimming it out.

Additionally, we haven't yet switched the old logic (for `certificate` field on the `ca_chain` endpoint) to the new full chain. I'd like to make that change, given the goals of rotation, and have it be for the default issuer only.

Thoughts?

### DIFF
--- a/builtin/logical/pki/storage.go
+++ b/builtin/logical/pki/storage.go
@@ -85,6 +85,10 @@ func listKeys(ctx context.Context, s logical.Storage) ([]keyId, error) {
 }
 
 func fetchKeyById(ctx context.Context, s logical.Storage, keyId keyId) (*key, error) {
+	if len(keyId) == 0 {
+		return nil, errutil.InternalError{Err: fmt.Sprintf("unable to fetch pki key: empty key identifier")}
+	}
+
 	keyEntry, err := s.Get(ctx, keyPrefix+keyId.String())
 	if err != nil {
 		return nil, errutil.InternalError{Err: fmt.Sprintf("unable to fetch pki key: %v", err)}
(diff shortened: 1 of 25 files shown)


stevendpclark (approved):

builtin/logical/pki/storage.go
```diff
@@ -575,7 +575,6 @@ func fetchCertBundleByIssuerId(ctx context.Context, s logical.Storage, id issuer
 
 	var bundle certutil.CertBundle
 	bundle.Certificate = issuer.Certificate
-	bundle.IssuingCA = issuer.CAChain[0]
```
kitography, line 578: sorry, this is used in sdk/helper/certutil/types.go, I think?  Clarification that  `pemBlock, _ = pem.Decode([]byte(c.IssuingCA))` will still work?

kitography (commented):
Bug fix and tests look good, have question about effect on SDK

builtin/logical/pki/storage.go
cipherboy: Note its gated on size tho:

https://github.com/hashicorp/vault/blob/main/sdk/helper/certutil/types.go#L210-L212

So it won't really matter, we'll skip it.

cipherboy:
Manually merged, per out of band conversation with @kitography and @stevendpclark. Thanks!

cipherboy:

PR #15012, opened 2022-04-12
UI/Clients monthly usage component

hellobontempo:
Adds the monthly usage vertical bar chart and component for 1.11 client count work.

vault-usage


ui/app/components/clients/running-total.js
```diff
@@ -34,30 +34,24 @@ import { mean } from 'd3-array';
...
-  get getTotalClients() {
-    return (
-      this.args.chartLegend?.map((legend) => {
-        return {
-          label: legend.label,
-          total: this.args.runningTotals[legend.key],
-        };
-      }) || null
-    );
+  barChartData = this.args.barChartData;
```
hellobontempo, line 41: is there a benefit to making these variables here? could just get the mean from `this.args.barChartData`..

ui/app/templates/components/clients/monthly-usage.hbs
```diff
@@ -14,38 +13,29 @@
...
     </p>
   </div>
 
   <div class="data-details-top">
     <h3 class="data-details">Average total clients per month</h3>
     <p class="data-details">
-      {{#if this.averageTotalClients}}
-        {{format-number this.averageTotalClients}}
-      {{else}}
-        <span class="has-text-danger is-size-8">
```
hellobontempo, line 27: Removed because if there is an error collecting data for this component - then it shouldn't render at all

ui/app/components/clients/running-total.js
zofskeez: Assigning an arg to a class property this way just doesn't quite seem right to me. For one thing if the arg changes that will not be reflected in `this.barChartData`. If you did want to assign the initial value to a property it would be more appropriate to do so in the constructor.

ui/app/components/clients/monthly-usage.js
```diff
@@ -31,13 +33,15 @@ import { mean } from 'd3-array';
  * @param {array} chartLegend - array of objects with key names 'key' and 'label' so data can be stacked
  */
 export default class MonthlyUsage extends Component {
+  barChartData = this.args.verticalBarChartData;
```
zofskeez, line 36: See comment below.

ui/app/components/clients/running-total.js
```diff
@@ -34,30 +34,24 @@ import { mean } from 'd3-array';
...
+  runningTotals = this.args.runningTotals;
+
+  get getEntityClientData() {
+    return {
+      runningTotal: this.runningTotals.entity_clients,
+      averageNewClients: Math.round(mean(this.barChartData?.map((d) => d.entity_clients))),
+    };
```
zofskeez, line 48: This is more of a stylistic comment but using _get_ in a property name makes it sound like it's a function rather than a property. I think `entityClientData` would be more clear on what this represents.

ui/app/components/clients/running-total.js
```diff
@@ -34,30 +34,24 @@ import { mean } from 'd3-array';
...
-          label: legend.label,
-          average: Math.round(mean(this.args.barChartData?.map((d) => d[legend.key]))),
-        };
-      }) || null
-    );
+  get getNonEntityClientData() {
+    return {
+      runningTotal: this.runningTotals.non_entity_clients,
+      averageNewClients: Math.round(mean(this.barChartData?.map((d) => d.non_entity_clients))),
+    };
```
zofskeez, line 55: See comment above

ui/app/templates/components/clients/running-total.hbs
```diff
@@ -54,21 +49,17 @@
...
     <div class="data-details-bottom">
-      {{#let (get this.getAverageNewClients 1) as |stat|}}
-        <h3 class="data-details">Average new non-entity clients per month</h3>
-        <p class="data-details">
-          {{format-number stat.average}}
-        </p>
-      {{/let}}
+      <h3 class="data-details">Average new non-entity clients per month</h3>
+      <p class="data-details">
+        {{format-number (get this.getNonEntityClientData "averageNewClients")}}
```
zofskeez, line 61: The instances in this file that use the `get` helper are not necessary since you know the name of the property you are trying to access. Instead you can replace it with `this.getNonEntityClientData.averageNewClients`.

zofskeez (changes requested):
Nice work! I left a few comments for you.

ui/app/components/clients/running-total.js
hellobontempo: Thanks! I tried that earlier and it didn't work - which is what prompted me reaching out in slack. Might have been some other syntax or typo issue as it works now! 🤯

ui/app/templates/components/clients/running-total.hbs
hellobontempo: Thanks! I tried that earlier and it didn't work - which is what prompted me reaching out in slack. Might have been some other syntax or typo issue as it works now! 🤯

zofskeez (approved):
Thanks!

PR #15013, opened 2022-04-12
Rework PKI test TestBackend_Root_Idempotency

stevendpclark:
- Validate that generate/root calls are no longer idempotent, but the bundle importing
   does not generate new keys/issuers
 - As before make sure that the delete root api resets everything
 - Address a bug within the storage that we bombed when we had multiple different
   key types within storage.


cipherboy (approved):

cipherboy:

PR #15018, opened 2022-04-13
Update default value of agent-cache-listener-port

calvin620707:
Base on this line and my experiment, I think the default value of vault.hashicorp.com/agent-cache-listener-port should be 8200.

https://github.com/hashicorp/vault-k8s/blob/main/agent-inject/agent/agent.go#L30


tvoran (approved):
Good catch, thanks!

PR #15019, opened 2022-04-13
Address some small nits in PKI rotation

stevendpclark:
- Reading the RFC, I had missed the detail about the migrated issuer/keys should have the Name set to `current`.
- Fix issue spotted that we were no longer building up a CRL after the set-signed api call. Fix that issue and add a corresponding test validating a CRL exists post set-signed.


cipherboy (approved):

cipherboy:

PR #15026, opened 2022-04-13
Form Field Glimmer

zofskeez:
While setting out to update the validation message handling in the `FormField` component the opportunity presented itself to glimmerize the component as well as `FormFieldGroups`. The model validations handling was also updated in components that use the form field components.


ui/app/components/mount-backend-form.js
```diff
@@ -167,23 +163,11 @@ export default Component.extend({
...
-          : set(this.validationMessages, 'path', path.errors.join('. '));
-      }
-      // check maxVersions is a number
-      if (name === 'maxVersions') {
-        maxVersions.isValid
-          ? set(this.validationMessages, 'maxVersions', '')
-          : set(this.validationMessages, 'maxVersions', maxVersions.errors.join('. '));
-      }
-      this.set('isFormInvalid', !isValid);
+      const { isValid, state } = this.mountModel.validate();
```
Monkeychip, line 166: so much cleaner!

ui/lib/core/addon/components/form-field-groups.js
```diff
@@ -22,24 +22,21 @@ import layout from '../templates/components/form-field-groups';
  *  @validationMessages={{validationMessages}}
  * />
  * ```
- *
- * @param [renderGroup=null] {String} - An allow list of groups to include in the render.
- * @param model=null {DS.Model} - Model to be passed down to form-field component. If `fieldGroups` is present on the model then it will be iterated over and groups of `FormField` components will be rendered.
- * @param onChange=null {Func} - Handler that will get set on the `FormField` component.
- * @param onKeyUp=null {Func} - Handler that will set the value and trigger validation on input changes
- * @param validationMessages=null {Object} Object containing validation message for each property
+ * @callback onChangeCallback
```
Monkeychip, line 25: oh, I didn't know this was an option. I like it.

Monkeychip: (@callback)

ui/lib/core/addon/components/form-field-groups.js
```diff
@@ -22,24 +22,21 @@ import layout from '../templates/components/form-field-groups';
...
+ * @callback onKeyUpCallback
+ * @param {Model} model- Model to be passed down to form-field component. If `fieldGroups` is present on the model then it will be iterated over and groups of `FormField` components will be rendered.
+ * @param {string} [renderGroup] - An allow list of groups to include in the render.
+ * @param {onChangeCallback} [onChange] - Handler that will get set on the `FormField` component.
```
Monkeychip, line 29: these look like duplicates of the above on row 25 and 26?

ui/lib/core/addon/components/form-field-groups.js
```diff
@@ -22,24 +22,21 @@ import layout from '../templates/components/form-field-groups';
...
+ * @param {onKeyUpCallback} [onKeyUp] - Handler that will set the value and trigger validation on input changes
+ * @param {ModelValidations} [modelValidations] - Object containing validation message for each property
  *
  */
 
-export default Component.extend({
-  layout,
```
Monkeychip, line 35: you were able to remove the layout but not extend it as we've done in other addon glimmerized components? How is this not breaking the component?

Monkeychip: Example: 
```
class StatTextComponent extends Component {}

export default setComponentTemplate(layout, StatTextComponent);
```

ui/lib/core/addon/components/form-field-groups.js
zofskeez: The pattern is to define the callback using `@callback` and then use it as a param type. I found this when reviewing the [jsdoc documentation for params](https://jsdoc.app/tags-param.html) under the _Callback Functions_ section at the bottom.

ui/lib/core/addon/components/form-field-groups.js
Monkeychip: Is it because you removed the hbs files from the templates directory?

zofskeez: Yes if you move the template file so that it is colocated with the component js file ember figures it out! I would love to colocate all components and even use `pod` structure so that route/controller/template files exist together in the tree.

ui/lib/core/addon/components/form-field-groups.js
Monkeychip: Interesting. Nice find.

Monkeychip (approved):
Might be worth running the enterprise tests locally before merging. Just guessing that the enterprise suite may have some unique use cases for this component. Looks great. Thank you for tackling!

zofskeez:
> Might be worth running the enterprise tests locally before merging. Just guessing that the enterprise suite may have some unique use cases for this component. Looks great. Thank you for tackling!

Good call. I just did a run locally and everything passed!

ui/lib/core/addon/components/form-field-groups.js
Monkeychip: That's awesome. Going forward I'll try and remember this. It's nice to have some examples in the code.

Issue #13270, opened 2021-11-24
Obfuscated password values in WebUI shows if a password contains certain characters


kafeinnet:
**Describe the bug**

When a secret value contains some accentuated characters (é, à, è, ...), the obfuscated value in the WebUI shows where those characters are.

**To Reproduce**

1. Store some accentuated characters in a secret value, like é ou à
2. Look a the obfuscated value

![image](https://user-images.githubusercontent.com/4932189/143261506-5a7009a8-cb85-4ba7-a68d-a05704357311.png)

![image](https://user-images.githubusercontent.com/4932189/143261451-58504c96-2b11-4b83-b62b-741bea4c4f39.png)

**Expected behavior**

No hints about the stored value should be displayed (imho, not even the length of the password).

**Environment:**

```
Key             Value
---             -----
Seal Type       shamir
Initialized     true
Sealed          false
Total Shares    5
Threshold       3
Version         1.8.4
Storage Type    consul
Cluster Name    xxx
Cluster ID      xxx
HA Enabled      true
HA Cluster      https://xx.xx.xx.xx:8201
HA Mode         active
Active Since    2021-10-21T08:27:17.605793713Z
```

evsasha:
I confirm. Obfuscation  work correct on versions 1.6.7, 1.7.7 but not work in 1.9.1 and 1.8.6

arnav28:
Thanks @evsasha for pointing it out. We have taken both the changes (font discrepancy and length of obfuscated values) in consideration and will hopefully be pushing a fix as soon as it's prioritized. 👍

PR #15028, opened 2022-04-13
Backport of UI: Masked inputs always look the same when value is hidden into release/1.9.x

hc-github-team-secure-vault-core:
## Backport

This PR is auto-generated from #15025 to be assessed for backporting due to the inclusion of the label backport/1.9.x.

---

Previous to this fix, if a masked value had special characters such as é, à, è the masked value would give away where chose characters are in the string. 

After this fix, masked display will be completely agnostic to the actual value -- always showing the same length and size of character. This PR addresses [issue 13270](https://github.com/hashicorp/vault/issues/13270)

**Before - you can tell without showing the value where the special character is**
masked value display - before

**After - can no longer tell when masked the length or location of special characters**
masked value display - after
masked value show - after

It's worth noting that the input will still show the true length and location of the special character. This is so that the user can type in the field without unmasking the value.
masked value input - after


hashishaw:
No changes 😞

PR #15035, opened 2022-04-13
UI/Add double attribution chart to current

hellobontempo:
### Adds the separated new/total attribution chart to the current tab

attribution-split

**TO DO**
- [ ] Update CSV to include new counts - will be in a separate PR


ui/app/serializers/clients/monthly.js
```diff
@@ -29,22 +33,28 @@ export default class MonthlySerializer extends ApplicationSerializer {
...
+    // if object only has outdated key names, update naming
+    if (Object.keys(object).includes('distinct_entities')) {
+      let { clients, distinct_entities, non_entity_tokens } = object;
+      return {
+        clients,
+        entity_clients: distinct_entities,
+        non_entity_clients: non_entity_tokens,
+      };
+    }
+    // TODO CMB: test what to return if neither key exists
```
hashishaw, line 57: I think this TODO is handled by returning the original object

ui/app/templates/components/clients/horizontal-bar-chart.hbs
```diff
@@ -1,12 +1,17 @@
...
+  <svg
+    data-test-horizontal-bar-chart
+    class="chart is-horizontal"
+    {{on "mouseleave" this.removeTooltip}}
+    {{did-insert this.renderChart @dataset}}
+    {{did-update this.renderChart @dataset}}
+  >
+  </svg>
+{{else}}
+  <div class="chart-empty-state">
```
hashishaw, line 11: nice!

ui/mirage/handlers/clients.js
```diff
@@ -663,6 +664,7 @@ const handleMockQuery = (queryStartTimestamp, monthlyData) => {
     do {
       i++;
       let timestamp = formatRFC3339(sub(startDateByMonth, { months: i }));
+      // this is probably not what the empty object looks like but waiting to hear back from backend
```
hashishaw, line 667: Can we add a "TODO" here so we make sure to come back and clean it up once we do hear back from backend?

ui/tests/helpers/clients.js
```diff
@@ -58,7 +58,7 @@ export function generateConfigResponse(overrides = {}) {
     request_id: 'some-config-id',
     data: {
       default_report_months: 12,
-      enabled: 'default-enable',
```
hashishaw, line 61: FYI, when ENT vault is started this value is `default-enable` from the API until it's changed, and likewise for OSS and `default-disable`. We should make sure we don't lose support for those values (unless the API removed them and I wasn't aware)

hashishaw (approved):
Nice work! A couple small comments but nothing blocking

ui/mirage/handlers/clients.js
hellobontempo: Great idea!

ui/tests/helpers/clients.js
hellobontempo: Oh! I has looked and saw it was `enable` but I believe that was when I had turned it off and back on....I'll change back!

Issue #10346, opened 2020-11-06
Web UI allows to enter comma-separated token policies


manderson-it:
**Describe the bug**
When creating a new role for an Auth Method (e.g., Kubernetes) through the Web UI the Web UI EditField named **Generated Token's Policies** allows to enter comma-separated policies. This leads to a consistent Web UI display, but results in invalid referenced Vault policies.

**To Reproduce**
1. Web UI: Navigate to Access > Auth Method (e.g., Kubernetes)
2. Web UI: Click "..." > View Configuration > Tab Roles > Create Role
3. Web UI: Enter Name, Bound service account names, Bound service namespaces
4. Web UI: Expand Tokens to reveal EditField **Generated Token's Policies**
5. Web UI: Enter a comma-separated value referencing two different Vault policies 'default,helloworld' 

This results in a consistent display after the role has been created (see screenshot).
![vault-webui-bug-policies1](https://user-images.githubusercontent.com/8469956/98423982-5204e800-205e-11eb-9545-d86192424666.png)

EditField when comma-separating two different Vault policies in the field:
![vault-webui-bug-EditField-wrong](https://user-images.githubusercontent.com/8469956/98424033-7a8ce200-205e-11eb-97d2-ae03cb72ac02.png)

EditField when using the Add Button to the right:
![vault-webui-bug-EditField-correct](https://user-images.githubusercontent.com/8469956/98424049-84aee080-205e-11eb-9456-a33b33fa48bd.png)

**Expected behavior**
1. The Vault Web UI rejects any comma-separated input in the EditField **Generated Token's Policies**, or
2. the Web UI splits the comma-separated policies into two different entries, meaning one entry for each policy.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.5.5
* Vault CLI Version (retrieve with `vault version`): 1.5.5
* Server Operating System/Architecture: Alpine

Vault server configuration file(s):

```hcl
# Paste your Vault config here.
# Be sure to scrub any sensitive values
```

**Additional context**
After enabling audit log with raw output, you can see what Vault consumed.
`vault audit enable file file_path=stdout log_raw=true`

"token_policies":["default,helloworld"] <---- this is invalid
versus
"token_policies":["default","helloworld"] <---- correct, notice the double-quotes before and after comma

Monkeychip:
[...]

PR #15046, opened 2022-04-14
Custom tooltip for Generated Token Policies form field on auth methods

Monkeychip:
This PR intercepts the model on the `generatedItem` component to replace the `helpText` for the auth form field "Generated Token's Policies"

Before the help text read: _Comma-separated list of policies associated with the user._

However, this field (e.g form field = stringArray) does not accept comma separated lists. We instead want a user to add a row for each new policy. We cannot change the text [here](https://github.com/hashicorp/vault/blob/642a60ace0b4e3a92073f2a676b7c621edf6d1a2/builtin/credential/userpass/path_user_policies.go#L26) because that changes the CLI text which _does_ accept a comma separated list.  So instead, we interrupt the component, search for this field type and replace the helpText with the following: _Add policies that will apply to the generated token for this user. One policy per row._

This now looks like this:
image

This change effects TLS Certificates, Userpass, and Kubernetes who have this field. The other auth methods have a different field or no policy field at all.

Will backport to 1.8

See GH issue [here](https://github.com/hashicorp/vault/issues/10346).


ui/app/components/generated-item.js
```diff
@@ -48,6 +48,20 @@ export default Component.extend({
   ),
   init() {
     this._super(...arguments);
+    this.model.fieldGroups.forEach((element) => {
+      // overwriting the helpText for Token Polices.
+      // HelpText from the backend says add a comma separated list, which works on the CLI but not here on the UI.
+      // This effects TLS Certificates, Userpass, and Kubernetes. https://github.com/hashicorp/vault/issues/10346
+      if (element.Tokens) {
+        element.Tokens.forEach((attr) => {
```
hashishaw, line 56: Instead of iterating over all of them can we do `element.Tokens.find(attr => attr.name === 'tokenPolicies')` and then update it if found? That way iteration will stop once the item is found, or have the same behavior as iterating over each if not found

ui/tests/acceptance/auth-list-test.js
```diff
@@ -69,6 +70,18 @@ module('Acceptance | auth backend list', function (hooks) {
     await triggerKeyEvent('[data-test-input="username"]', 'keyup', 65);
     await fillIn('[data-test-textarea]', user2);
     await triggerKeyEvent('[data-test-textarea]', 'keyup', 65);
+    // test for modified helpText on generated token policies
+    await click('[data-test-toggle-group="Tokens"]');
```
hashishaw, line 74: 👏

hashishaw (changes requested):
Nice work tracking this down! Small change request, let me know if that doesn't work for some reason

ui/app/components/generated-item.js
Monkeychip: Great idea. Amending.

hashishaw (approved):
🚀

PR #15054, opened 2022-04-14
Add AWS_DYNAMODB_REGION Environment variable

sergeylanzman:
From eks 1.18 `AWS_REGION` and `AWS_DEFAULT_REGION` auto inject to pods. So if you can't use dynamodb table from other region it's read from `AWS_REGION` always and override on config file. Reorder setting can be break current vault settings. I added one new with `AWS_DYNAMODB_REGION`

https://github.com/aws/amazon-eks-pod-identity-webhook/issues/40

fix https://github.com/hashicorp/vault/issues/15071

changelog
```
storage/dynamodb: Added `AWS_DYNAMODB_REGION` environment variable
```

### DIFF
--- a/changelog/15054.txt
+++ b/changelog/15054.txt
@@ -1,5 +1,3 @@
-
-   
 ```release-note:feature
 storage/dynamodb: Added `AWS_DYNAMODB_REGION` environment variable.
 ```
--- a/changelog/15054.txt
+++ b/changelog/15054.txt
@@ -1,0 +1,5 @@
+
+   
+```release-note:feature
+storage/dynamodb: Added `AWS_DYNAMODB_REGION` environment variable.
+```
--- a/physical/dynamodb/dynamodb.go
+++ b/physical/dynamodb/dynamodb.go
@@ -163,13 +163,16 @@ func NewDynamoDBBackend(conf map[string]string, logger log.Logger) (physical.Bac
 	if endpoint == "" {
 		endpoint = conf["endpoint"]
 	}
-	region := os.Getenv("AWS_REGION")
+	region := os.Getenv("AWS_DYNAMODB_REGION")
 	if region == "" {
-		region = os.Getenv("AWS_DEFAULT_REGION")
+		region := os.Getenv("AWS_REGION")
 		if region == "" {
-			region = conf["region"]
+			region = os.Getenv("AWS_DEFAULT_REGION")
 			if region == "" {
-				region = DefaultDynamoDBRegion
+				region = conf["region"]
+				if region == "" {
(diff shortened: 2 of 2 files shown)


hsimon-hashicorp:
Hi! Can you please also open and link a GitHub issue? Please don't forget a changelog entry, too. Thanks! :)

sergeylanzman:
@hsimon-hashicorp Added issue. How I can add changelog? can you point me to  example?

hsimon-hashicorp:
> @hsimon-hashicorp Added issue. How I can add changelog? can you point me to example?

Thank you for the issue! Adding a changelog entry is described in the [CONTRIBUTING.md](https://github.com/hashicorp/vault/blob/main/CONTRIBUTING.md). Thanks! :)

swenson (dismissed):
LGTM

This seems pretty reasonable to me. Thanks for this!

changelog/15054.txt
```diff
@@ -0,0 +1,5 @@
+
```
swenson, line 1: Could you remove the blank lines from the beginning?

swenson (commented):
Oops, I just saw the extra empty lines in the changelog file. Please remove those.

sergeylanzman:
Done

swenson (approved):
LGTM

physical/dynamodb/dynamodb.go
```diff
@@ -163,13 +163,16 @@ func NewDynamoDBBackend(conf map[string]string, logger log.Logger) (physical.Bac
 	if endpoint == "" {
 		endpoint = conf["endpoint"]
 	}
-	region := os.Getenv("AWS_REGION")
+	region := os.Getenv("AWS_DYNAMODB_REGION")
```
calvn, line 166: @swenson I wonder if using a `switch` is cleaner instead of deep nesting. I know that we have this pattern for the other config params, but since we're evaluating several env vars for this one it might not be a bad idea?
```go
	var region string
	switch {
	case os.Getenv("AWS_DYNAMODB_REGION") != "":
		region = os.Getenv("AWS_DYNAMODB_REGION")
	case os.Getenv("AWS_REGION") != "":
		region = os.Getenv("AWS_REGION")
	case os.Getenv("AWS_DEFAULT_REGION") != "":
		region = os.Getenv("AWS_DEFAULT_REGION")
	case conf["region"] != "":
		region = conf["region"]
	default:
		region = DefaultDynamoDBRegion
	}
```

swenson: Oh, I was thinking that the deep nesting was a little bit much. The switch is much cleaner.

PR #15055, opened 2022-04-14
Remove duplicate policies when creating/updating identity groups

chrishoffman:
Currently when adding policies to identity groups, the list is not deduplicated resulting in multiple copies of the same policy in an identity group.


ccapurso (approved):
Nice find, looks good to me

PR #15058, opened 2022-04-15
UI: fix blank selection on search select field

hashishaw:
Fixes a bug related to cross-namespace model querying within an entity. 

Previous behavior:
* Search-select component queries against a model in the current namespace
* The API call returns a 404 because there are no results in the current namespace
* The item being loaded includes ID's for that model, but they exist in a different namespace
* **BUG**: the search-select component shows a blank selection

**Before**
entity-display-bug
Another workaround for this behavior is to add an item of the type being queried to the current namespace. For example, in the gif above we would create an `Entity` in the namespace `ns1`

**After**
entity-display-fix


ui/README.md
```diff
@@ -26,6 +26,15 @@
```
Monkeychip, line 26: thank you for updating!

Monkeychip (approved):

PR #15060, opened 2022-04-15
Update changelog checker for validating Go toolchain updates

cipherboy:

---

Going to push some test commits, but this should help enforce Go toolchain version update PRs have the right format in the changelog entries.

- Valid Go update changelog: https://github.com/hashicorp/vault/runs/6041130204?check_suite_focus=true
- Numbered changelog entry (should fail): https://github.com/hashicorp/vault/runs/6041182182?check_suite_focus=true 
- Regular changelog entry: https://github.com/hashicorp/vault/runs/6041201670?check_suite_focus=true

### DIFF
--- a/.github/workflows/changelog-checker.yml
+++ b/.github/workflows/changelog-checker.yml
@@ -24,7 +24,10 @@ jobs:
           fetch-depth: 0 # by default the checkout action doesn't checkout all branches
       - name: Check for changelog entry in diff
         run: |
-          # check if there is a diff in the changelog directory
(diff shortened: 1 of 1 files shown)


mladlow (approved):
Thanks for fixing this up! I feel like eventually we should refactor the check into something like https://github.com/hashicorp/go-changelog/tree/main/cmd/changelog-pr-body-check but it's a much bigger change.

cipherboy:
Thanks! Yeah, that makes sense :-)

PR #15063, opened 2022-04-15
Backport of UI: fix blank selection on search select field into release/1.10.x

hc-github-team-secure-vault-core:
## Backport

This PR is auto-generated from #15058 to be assessed for backporting due to the inclusion of the label backport/1.10.x.

---

Fixes a bug related to cross-namespace model querying within an entity. 

Previous behavior:
* Search-select component queries against a model in the current namespace
* The API call returns a 404 because there are no results in the current namespace
* The item being loaded includes ID's for that model, but they exist in a different namespace
* **BUG**: the search-select component shows a blank selection

**Before**
entity-display-bug
Another workaround for this behavior is to add an item of the type being queried to the current namespace. For example, in the gif above we would create an `Entity` in the namespace `ns1`

**After**
entity-display-fix


hashishaw (approved):

hellobontempo (approved):
LGTM!

PR #15064, opened 2022-04-15
Update Go version to 1.16.15

cipherboy:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,7 @@ jobs:
     environment:
     - CIRCLECI_CLI_VERSION: 0.1.5546
     - GO_TAGS: ''
-    - GO_VERSION: 1.16.14
+    - GO_VERSION: 1.16.15
     - GOTESTSUM_VERSION: 0.5.2
   algolia-index:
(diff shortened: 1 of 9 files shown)


calvn:
I see that we include changelog entries on go version upgrades. Should we do so in this case as well?

cipherboy:
@calvn This PR has an entry: https://github.com/hashicorp/vault/pull/15064/files#diff-273a5bc507d92c5e8a4d96bfeff603dd0334f7857565689be28cc8168c7207a8 

Just note that the checking helper expects the old format; I haven't backported https://github.com/hashicorp/vault/pull/15060. 

This means it'll err because it expects the `15064.txt` format rather than the new `go-ver-SOMETHING.txt` format.

Hence, the label :-)

mladlow (approved):
The changelog entry looks right. Everything else looks _probably_ right and I'm assuming you grepped but I did not. There's a ton of places to update.

cipherboy:
Second grep on a fresh date showed nothing new. Will do the OSS->ENT merge now. Thanks @mladlow!

Issue #15057, opened 2022-04-15
 interface conversion: interface {} is json.Number, not string in sdk/framework/field_data.go


Nikita1821:
I use framework.TypeCommaIntSlice for get emloyeeID and add it in group AD in my engine.

Send: curl -X POST -H "X-Vault-Request:..." -H "X-Vault-Token: $(vault print token)" -d '{"param":  213}' http://127.0.0.1:8200/v1/<engine path>

In vault system log: interface conversion: interface {} is json.Number, not string

You use StringToSliceHookFunc() in DecodeHook();

if reflect1.Kind != reflect.String || reflect2.Kind != reflect.Slice {return data, nil} 
you do it : raw := data.(string)
reflect.ValueOf(json.Number), reflect.Kind() => string, reflect.Type() => json.Number

sdk/framework/field_data.go, maybe remove function  mapstructure.StringToSliceHookFunc(",") from DecodeHook

commit: Brian Kassouf, 4 years ago   (March 3rd, 2018 2:01 AM) logical/framework: add a field data type for parsing integer arrays, [83a7d20]

VinnyHC:
Hello, @Nikita1821 thanks for opening this issue. Could you please let us know what version of Vault (found using `vault status`) you're experiencing this with?

Nikita1821:
hi, the last version

Nikita1821:
ok, v1.10.0

Nikita1821:
you can add to any engine parameter with type framework.TypeCommaIntSlice
send just int and get this error

ccapurso:
As @Nikita1821 noted, this occurs due to the type `string` assertion performed in `mapstructure.StringToSliceHookFunc` when the input is of type `json.Number`. A similar issue used to occur with `framework.TypeCommaStringSlice`. The parsing for that type is handled by `ParseCommaStringSlice` in `github.com/hashicorp/go-secure-stdlib/parseutil`. The issue was fixed in https://github.com/hashicorp/vault/pull/14522. If the type of the input is `json.Number` then we simply call `String()`. We can likely add the same conversion for `framework.TypeCommaIntSlice`.

PR #15072, opened 2022-04-18
fix TypeCommaIntSlice panic caused by json.Number input

ccapurso:
Vault will panic when it attempts to parse fields with a schema of `TypeCommaIntSlice` if the provided value is an integer specifically if provided within the JSON body of an HTTP request. This is similar to a recent fix that was implemented for `TypeCommaStringSlice` which was fixed in https://github.com/hashicorp/vault/pull/14522. The issue resides in `mapstructure.StringToSliceHookFunc`.

The fix provided in this pull request is to check if the raw input is of type `json.Number` and then calling `String()`.

Fixes https://github.com/hashicorp/vault/issues/15057

### DIFF
--- a/changelog/15072.txt
+++ b/changelog/15072.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Fix panic caused by parsing JSON integers for fields defined as comma-delimited integers
+```
--- a/sdk/framework/field_data.go
+++ b/sdk/framework/field_data.go
@@ -243,6 +243,12 @@ func (d *FieldData) getPrimitive(k string, schema *FieldSchema) (interface{}, bo
 
 	case TypeCommaIntSlice:
 		var result []int
+
+		jsonIn, ok := raw.(json.Number)
+		if ok {
+			raw = jsonIn.String()
+		}
+
 		config := &mapstructure.DecoderConfig{
 			Result:           &result,
 			WeaklyTypedInput: true,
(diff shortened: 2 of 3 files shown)


swayne275 (commented):
what happens if it is comma separated ints e.g. `1,2`?

ccapurso:
> what happens if it is comma separated ints e.g. `1,2`?

@swayne275, that's a great question. Providing `1,2` as an integer would not be valid JSON and I imagine that the JSON deserialization would fail. `TypeCommaIntSlice` does support a comma-delimited string of integers (e.g. `"1, 2"`) and would result in output `[]int{1, 2}`. We cover this in a test case:
 https://github.com/hashicorp/vault/blob/f5327afa4342c239178aa93d550e02120e1b5ea6/sdk/framework/field_data_test.go#L525-L535

If providing `1,2` via an HTTP request is parsed successfully resulting in a json.Number, (`json.Number("1, 2")` is actually valid), the `String()` call would then provide the same result as providing it as the string `"1, 2"`.

swayne275 (approved):

swayne275:
@ccapurso why do we use `TypeCommaIntSlice` in the first place then?

ccapurso:
@swayne275, it allows users to provide `[]int` in a flexible way. Inputs on the CLI are all strings. So providing a flag like `-foo=1,2,3` from the CLI or the JSON `{"foo": [1,2,3]}` or `{"foo": "1,2,3"}` will all result in `[]int{1, 2, 3}` on the backend.

PR #6496, opened 2019-03-28
Add storybook

noelledaley:
# Add Storybook
This PR implements Storybook as a living style guide to catalog components in the Vault UI.

Screen Shot 2019-03-29 at 4 11 41 PM

storybook-demo

### Features
- View live examples of Vault's components
- Change the viewport to see components as they appear on different devices
- View component documentation, example usage, and and links to source code on the 'Notes' tab
- Automatically generate story and notes files for each component

______

### Testing
- Boot up Storybook with `yarn storybook`
- Verify that all components show up in the left nav (note: you'll have to refresh upon initial page load. this is a known bug with Storybook Ember).
- Verify that each component has at least 1 example in the canvas tab
- Verify that each component notes tab has a description, links to source code, a table of properties, and an example of how to use the component
- Verify that you can generate a new story file `ember generate story <component-name>`
- Verify that you can generate the component notes markdown file `yarn gen-story-md [name-of-component]`

### Feedback
- Does the file structure and organization of components in the left nav make sense?
- Are the examples we've shown in the canvas tab helpful?
- Are there any components which should use knobs (ways to dynamically change a components properties) that don't?
- Do the component descriptions and properties table make sense?

Previous discussion related to this PR is on https://github.com/hashicorp/vault/pull/6375

PR #15074, opened 2022-04-18
remove storybook:

Monkeychip:
As a team we decided to remove storybook. We didn't use it often enough and it was becoming a burden to update it's dependencies. 

[Here](https://github.com/hashicorp/vault/pull/6496/files) was the original PR that added storybook. 

I also updated browserstack. Not entirely sure why, but after removing the storybook dependencies from the package.json file, browserstack started to fail. I updated it and it fixed it. I wish I knew the answer to these things, but sometimes the web is too thick.


hellobontempo (commented):
LGTM! But I'll hold off reviewing and leave it to someone with more experience changing dependencies. 

Let me know if you want to pair on tackling that vercel failure

hashishaw (approved):
Thank you!

zofskeez (approved):
This looks good to me thanks for tackling!

PR #15075, opened 2022-04-18
Allow explicit issuer chain path construction

cipherboy:
This allows operators to manually construct complex CA chains manually, bypassing the existing chain building logic.

We add a new field, `manual_chain`, to the update issuers path, which explicitly constructs a chain (by ID), ignoring any unknown issuers.

@stevendpclark -- I added a commit here that fixes handling of duplicate names w.r.t. updating issuers, let me know what you think.

I also update some of the tests to exercise this.

### DIFF
--- a/builtin/logical/pki/chain_util.go
+++ b/builtin/logical/pki/chain_util.go
@@ -240,6 +240,39 @@ func rebuildIssuersChains(ctx context.Context, s logical.Storage, referenceCert 
 	processedIssuers := make(map[issuerId]bool, len(issuers))
 	toVisit := make([]issuerId, 0, len(issuers))
 
+	// Handle any explicitly constructed certificate chains. Here, we don't
+	// validate much what the user provides; if they provide since-deleted
+	// refs, skip them; if they duplicate entries, add them multiple times.
+	// The other chain building logic will be able to deduplicate them when
+	// used as parents to other certificates.
+	for _, candidate := range issuers {
+		entry := issuerIdEntryMap[candidate]
+		if len(entry.ManualChain) == 0 {
+			continue
+		}
+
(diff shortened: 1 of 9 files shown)


builtin/logical/pki/chain_util.go
```diff
@@ -124,6 +125,9 @@ func rebuildIssuersChains(ctx context.Context, s logical.Storage, referenceCert
 		if stored == nil || len(stored.Certificate) == 0 {
 			return fmt.Errorf("bad issuer while building chain: missing certificate entry: %v", identifier)
 		}
+		if strings.Count(stored.Certificate, "BEGIN CERTIFICATE") > 1 {
```
stevendpclark, line 128: Should we move this check into importIssuer?

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -141,22 +148,85 @@ func (b *backend) pathUpdateIssuer(ctx context.Context, req *logical.Request, da
 		return nil, err
 	}
 
+	newName, err := getIssuerName(ctx, req.Storage, data)
+	if err != nil {
+		// If the error is name already in use, and the new name is the
+		// old name for this issuer, we're not actually updating the
+		// issuer name (or causing a conflict) -- so don't err out. Other
+		// errs should still be surfaced, however.
+		if !strings.Contains(err.Error(), "issuer name already in use") || newName != issuer.Name {
```
stevendpclark, line 157: Could we have a dedicated error for this use case? String comparison on the error msg feels a little fragile.

stevendpclark (approved):
Some small nits, but overall 👍

stevendpclark (approved):

PR #15092, opened 2022-04-19
Upgrade hashicorp/consul-template dependency

tomhjp:
* Includes sprig template functions
* Includes improvements to writeTo template function
* Adds new test case for sprig functions and slightly improves the error reporting when the test fails

### DIFF
--- a/changelog/15092.txt
+++ b/changelog/15092.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+agent: Upgrade hashicorp/consul-template version for sprig template functions and improved writeTo function
+```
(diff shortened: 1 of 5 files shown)


calvn (approved):
Thanks for adding the test against sprig!

tvoran:
Looks like the tests are failing because `TemplateConfig.Command` is no longer a string? https://github.com/hashicorp/vault/blob/99e21324fa3471a3c4226bd317fd6d300742cb4f/command/agent/config/config_test.go#L678
And apparently `Command` is deprecated in favor of `Exec` now.

tvoran:
Looks like we should check for Exec in those tests, something like:

```go
Exec:           &ctconfig.ExecConfig{Command: []string{"restart service foo"}},
```

PR #15093, opened 2022-04-19
VAULT-5827 Update mongodb, brotli

swenson:
* `brotli` 1.0.1 was withdrawn, so `go get` throws an error for Vault: `go: warning: github.com/andybalholm/brotli@v1.0.1: retracted by module author: occasional panics and data corruption`
* `go-client-mongodb-atlas` has an old dependency on a renamed repo, and
  has been renamed twice. This caused issues in
  https://github.com/hashicorp/vault-plugin-secrets-mongodbatlas/issues/11
  for example.

Closes https://github.com/hashicorp/vault-plugin-secrets-mongodbatlas/issues/11

### DIFF
--- a/builtin/logical/database/backend_test.go
+++ b/builtin/logical/database/backend_test.go
@@ -3,7 +3,6 @@ package database
 import (
 	"context"
 	"database/sql"
-	"encoding/base64"
 	"fmt"
 	"log"
 	"net/url"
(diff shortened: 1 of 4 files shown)


calvn (approved):
👍 , does this need backporting?

swenson:
Thanks @calvn -- I think we probably don't need to backport since this is more preventative maintenance rather than a fix for an active bug?

PR #15099, opened 2022-04-20
Generated Model Bug

zofskeez:
This bug was discovered when attempting to create a user for the userpass auth method. After the 3.28 Ember upgrade there was an error thrown when attempting to set the id field of the model after it was initially set. Rather than setting the id field on change a private value is set through a computed property and then the id field is set in the adapter on save success.

### DIFF
--- a/ui/app/services/path-help.js
+++ b/ui/app/services/path-help.js
@@ -13,6 +13,7 @@ import fieldToAttrs from 'vault/utils/field-to-attrs';
 import { resolve, reject } from 'rsvp';
 import { debug } from '@ember/debug';
 import { dasherize, capitalize } from '@ember/string';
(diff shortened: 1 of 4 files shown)


changelog/15099.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+ui: Fixes issue setting username field when creating a user for userpass auth method
```
Monkeychip, line 2: not sure you need a changelog if it's a bug that was created in the 1.11 release and is not being backported any farther back.

zofskeez: Ya that's a good call. Since the bug exists in a different way on older versions but doesn't seem to impact creating a user I don't think this is worth backporting. I'll remove the changelog. Thanks!

PR #15100, opened 2022-04-20
Update CRL handling for multiple issuers

cipherboy:

---

This updates the CRL handling code to support generating a per-issuer CRL. Notably, when two issuers use the same public/private key _and_ have the same Subject, their CRLs are functionally equivalent and can be combined.

We update revoked cert entries to have an issuer field, allowing us to do the (expensive) issuer check once. 

We also update to the CRLv2 format, using code from @kitography.

### DIFF
--- a/builtin/logical/pki/cert_util.go
+++ b/builtin/logical/pki/cert_util.go
@@ -154,7 +154,10 @@ func fetchCertBySerial(ctx context.Context, req *logical.Request, prefix, serial
 		legacyPath = "revoked/" + colonSerial
 		path = "revoked/" + hyphenSerial
 	case serial == "crl":
-		path = "crl"
+		path, err = resolveIssuerCRLPath(ctx, req.Storage, defaultRef)
+		if err != nil {
+			return nil, err
+		}
 	default:
 		legacyPath = "certs/" + colonSerial
 		path = "certs/" + hyphenSerial
--- a/builtin/logical/pki/crl_util.go
+++ b/builtin/logical/pki/crl_util.go
@@ -1,6 +1,7 @@
 package pki
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"crypto/x509"
@@ -15,10 +16,13 @@ import (
 	"github.com/hashicorp/vault/sdk/logical"
 )
 
+const revokedPath = "revoked/"
+
 type revocationInfo struct {
 	CertificateBytes  []byte    `json:"certificate_bytes"`
 	RevocationTime    int64     `json:"revocation_time"`
 	RevocationTimeUTC time.Time `json:"revocation_time_utc"`
+	CertificateIssuer issuerID  `json:"issuer_id"`
 }
 
 // Revokes a cert, and tries to be smart about error recovery
@@ -53,7 +57,7 @@ func revokeCert(ctx context.Context, b *backend, req *logical.Request, serial st
 	alreadyRevoked := false
 	var revInfo revocationInfo
 
-	revEntry, err := fetchCertBySerial(ctx, req, "revoked/", serial)
+	revEntry, err := fetchCertBySerial(ctx, req, revokedPath, serial)
 	if err != nil {
 		switch err.(type) {
 		case errutil.UserError:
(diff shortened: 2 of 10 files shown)


builtin/logical/pki/crl_util.go
```diff
@@ -233,6 +365,16 @@ WRITE:
 		}
 	}
 
+	signingBundle, caErr := parseCABundle(ctx, b, req, bundle)
+	if caErr != nil {
+        switch caErr.(type) {
+        case errutil.UserError:
```
stevendpclark, line 371: nit: indentation?

builtin/logical/pki/crl_util.go
```diff
@@ -219,11 +302,60 @@ func buildCRL(ctx context.Context, b *backend, req *logical.Request, forceNew bo
...
+// Builds a CRL by going through the list of revoked certificates and building
+// a new CRL with the stored revocation times and serial numbers.
+func buildCRL(ctx context.Context, b *backend, req *logical.Request, forceNew bool, thisIssuerId issuerId, revoked []pkix.RevokedCertificate) error {
+	crlInfo, err := b.CRL(ctx, req.Storage)
+	if err != nil {
+		return errutil.InternalError{Err: fmt.Sprintf("error fetching CRL config information: %s", err)}
+	}
+
+	crlLifetime := b.crlLifetime
+	var revokedCerts []pkix.RevokedCertificate
```
stevendpclark, line 336: Not sure why we are defining another variable here, we simply always pass through the `revoked` param? Am I missing something?

stevendpclark (approved):
Nothing completely wrong, assuming we are overlooking the fact that we are overriding the final CRL and not actually writing out CRL's based on different issuers...

A few nits to address.

builtin/logical/pki/crl_util.go
cipherboy: Fixed :-)

builtin/logical/pki/crl_util.go
cipherboy: This is expected; see the complex conditionals at the start of `buildCRL` -- there's several cases where we write out a signed (but empty-list) CRL, even if we may have revoked certs.

builtin/logical/pki/crl_util.go
```diff
@@ -219,11 +351,106 @@ func buildCRL(ctx context.Context, b *backend, req *logical.Request, forceNew bo
...
+		}
+
+		if !foundParent {
+			// If the parent isn't found, add it to the unassigned bucket.
+			unassignedCerts = append(unassignedCerts, newRevCert)
+		} else {
+			// When the CertificateIssuer field wasn't found on the existing
+			// entry (or was invalid), and we've found a new value for it,
+			// we should update the entry to make future CRL builds faster.
+			revokedEntry, err = logical.StorageEntryJSON("revoked/"+normalizeSerial(serial), revInfo)
```
stevendpclark, line 390: So our call to normalizeSerial here is a bit touchy imho. 

We list the revoked/ storage above and fetch serial storage as is (which makes sense), with this call to normalizeSerial there is a chance that we write out the same serial within a different entry than what we loaded.

stevendpclark: Also please consider using the constant `revokedPath`

builtin/logical/pki/cert_util.go
```diff
@@ -154,7 +154,10 @@ func fetchCertBySerial(ctx context.Context, req *logical.Request, prefix, serial
 		legacyPath = "revoked/" + colonSerial
 		path = "revoked/" + hyphenSerial
 	case serial == "crl":
-		path = "crl"
+		path, err = resolveIssuerCRLPath(ctx, req.Storage, "default")
```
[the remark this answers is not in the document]
cipherboy: d'oh :D

builtin/logical/pki/crl_util.go
cipherboy: Ah, I see. Yeah, I was looking at the code in `revokeCert`, which writes at the correct serial location. I've fixed it :-)

stevendpclark (approved):

builtin/logical/pki/crl_util.go
```diff
@@ -219,11 +351,99 @@ func buildCRL(ctx context.Context, b *backend, req *logical.Request, forceNew bo
...
+			}
+		}
 	}
 
+	return unassignedCerts, revokedCertsMap, nil
+}
+
+// Builds a CRL by going through the list of revoked certificates and building
+// a new CRL with the stored revocation times and serial numbers.
+func buildCRL(ctx context.Context, b *backend, req *logical.Request, forceNew bool, thisIssuerId issuerID, revoked []pkix.RevokedCertificate, identifier string, crlNumber int64) error {
```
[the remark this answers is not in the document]
cipherboy: Good idea, ty :)

cipherboy:
Manually merged, thanks all!

cipherboy:

PR #15103, opened 2022-04-20
Client Count banner warnings for upgraded of minor 9 or 10

Monkeychip:
### This PR handles showing a warning banner for the following upgrade scenarios

**Current Month Tab:** warning shows if a relevant upgrade during the current month happened

**History Tab:** warning shows if a relevant upgrade happened during the time span of data returned from API 

**Relevant upgrades:**
_(only the initial upgrade to the minor version is important, subsequent patch upgrades not important)_
 - to 1.9.x or 1.10.x 
 - to 1.9.x and 1.10.x
 - from pre-1.9 to anything greater than 1.10 
 
**Why:**
Upgrade to 1.9: client counting changed  and we added auth mount distribution
Upgrade to 1.10: monthly breakdown added, so new clients breakdown per month now available

<hr> 
For each of these scenarios there is a different warning message that displays.

Designs:
Screen Shot 2022-04-22 at 12 11 48 PM


ui/app/components/clients/history.js
```diff
@@ -92,7 +92,44 @@ export default class History extends Component {
...
     // {id: '1.9.0', previousVersion: null, timestampInstalled: '2021-11-03T10:23:16Z'}
     // version id is 1.9.0 or earliest upgrade post 1.9.0, timestamp is RFC3339
-    return this.args.model.versionHistory[0] || null;
+    // need to return the last in the array as the data is in ascending order
+    return this.args.model.versionHistory.at(-1) || null;
+  }
+
+  get allMinorVersionsInRange() {
+    let newArray = [];
+    // ARG TODO testing unsure if the getActivityResponse is the correct one here.
```
Monkeychip, line 101: @hellobontempo I figured I'd throw this question at you. Is this the correct param to be checking against the queried range?

hellobontempo: yes - `startTime` and `endTime` are the unformatted timestamps from the API's activity response

ui/app/components/clients/history.js
```diff
@@ -92,7 +92,45 @@ export default class History extends Component {
   get latestUpgradeData() {
     // {id: '1.9.0', previousVersion: null, timestampInstalled: '2021-11-03T10:23:16Z'}
     // version id is 1.9.0 or earliest upgrade post 1.9.0, timestamp is RFC3339
-    return this.args.model.versionHistory[0] || null;
+    // need to return the last in the array as the data is in ascending order
+    return this.args.model.versionHistory.at(-1) || null;
```
hellobontempo, line 96: Did this change? I thought the `keys` were in descending order, with the latest first, and then in the serializer we map over the keys:
```
 normalizeFindAllResponse(store, primaryModelClass, payload, id, requestType) {
    let normalizedPayload = [];
    payload.keys.forEach((key) => {
      normalizedPayload.push({ id: key, ...payload.key_info[key] });
    });
    return this._super(store, primaryModelClass, normalizedPayload, id, requestType);
  },
  ```

File: ui/app/components/clients/current.js [hunk 3]
```diff
@@ -18,7 +18,8 @@ export default class Current extends Component {
   get latestUpgradeData() {
     // e.g. {id: '1.9.0', previousVersion: null, timestampInstalled: '2021-11-03T10:23:16Z'}
     // version id is 1.9.0 or earliest upgrade post 1.9.0, timestamp is RFC3339
-    return this.args.model.versionHistory[0] || null;
+    // need to return the last in the array as the data is in ascending order
+    return this.args.model.versionHistory.at(-1) || null;
```
hellobontempo, line 22: Question about this - see below

ui/app/components/clients/history.js
```diff
@@ -92,7 +92,45 @@ export default class History extends Component {
...
+        // push the minor version number e.g. 9
+        let upgradeVersionArray = object.id.split('.');
+        newArray.push(upgradeVersionArray[1]);
+      }
+    });
+    return newArray; // return [9,9,10] shows all the minor numbers that fall within the range of the returned API response.
+  }
+
+  get versionUpdateText() {
+    // count how many times a minor number occurs inside the array of minors within the API range.
```
hellobontempo, line 117: Do we need to know the count here? I'm wondering if we could simplify this a by using `.includes` instead.
```
let includesVersion9 = this.allMinorVersionsInRange.includes(9)
let includesVersion10 = this.allMinorVersionsInRange.includes(10)

if (includesVersion 9 && includesVersion10) {
...text for both }
if (includesVersion9) {
...text for 1.9
}
if (includes Version10) {
..text for 1.10 
}

File: ui/app/templates/components/clients/history.hbs [hunk 5]
```diff
@@ -102,26 +102,19 @@
...
-                How we count clients changed in 1.9, so please keep that in mind when looking at the data below, and you can
+                {{this.versionUpdateText}}
                 <DocLink
                   @path="/docs/concepts/client-count/faq#q-which-vault-version-reflects-the-most-accurate-client-counts"
                 >
-                  read more here.
+                  Learn more here.
                 </DocLink>
               </li>
             {{/if}}
```
hellobontempo, line 120: Just a note to update the conditional on line 137 here for the line chart (it's still using `this.countsIncludeOlderData`

ui/mirage/handlers/clients.js
```diff
@@ -678,7 +678,7 @@ export default function (server) {
   // 1.10 API response
   server.get('sys/version-history', function () {
     return {
-      keys: ['1.10.0', '1.9.0', '1.9.1', '1.9.2'],
```
hellobontempo, line 681: oh whoops! that was my bad! thanks for fixing 😅

ui/app/components/clients/history.js
hellobontempo: ☝️ on that note, we only care about the timestamp for whichever key has `previousVersion:null` 

would it make sense to instead have something like : 

``` get findInitialUpgrade () {
return this.args.model.versionHistory.find( version => version.previousVersion === null) || false
 }

Comment (hellobontempo) [reply ui/app/components/clients/history.js:96]: We do still need some sort of object with version upgrade info because it's passed to the chart to display upgrade info.

File: ui/app/components/clients/current.js [hunk 7]
```diff
@@ -76,13 +97,46 @@ export default class Current extends Component {
...
+    // return all upgrades that happened within date range of queried activity
+    return upgradesWithinData.length === 0 ? null : upgradesWithinData;
+  }
+
+  get upgradeVersionAndDate() {
+    if (!this.upgradeDuringCurrentMonth) {
+      return null;
+    }
+    if (this.upgradeDuringCurrentMonth.length === 2) {
+      return 'Vault was upgraded to 1.9 and 1.10 during this month.';
```
hashishaw, line 118: For future-proofing, could we use the value dates here too? That would even allow us to remove the length check, so for example:
```js
let versions = this.upgradeDuringCurrentMonth.map(upgrade => upgrade.id).join(' and ');
return `Vault was upgraded to ${versions} during this month`;
```

ui/app/components/clients/history.js
```diff
@@ -89,10 +90,81 @@ export default class History extends Component {
...
+    // get upgrade data for initial upgrade to 1.9 and/or 1.10
+    let relevantUpgrades = [];
+    const importantUpgrades = ['1.9', '1.10'];
+    importantUpgrades.forEach((version) => {
+      let findUpgrade = versionHistory.find((versionData) => versionData.id.match(version));
+      if (findUpgrade) relevantUpgrades.push(findUpgrade);
+    });
+
+    // if no history for 1.9 or 1.10, get first upgrade available
+    if (Object.entries(relevantUpgrades).length === 0) {
```
hashishaw, line 108: same comments. I wonder if this logic is useful as a helper/util fn... 🤔

hashishaw (approved):
Nice work y'all! A couple minor things, but looks good

ui/app/components/clients/current.js
```diff
@@ -15,10 +16,30 @@ export default class Current extends Component {
...
+    // get upgrade data for initial upgrade to 1.9 and/or 1.10
+    let relevantUpgrades = [];
+    const importantUpgrades = ['1.9', '1.10'];
+    importantUpgrades.forEach((version) => {
+      let findUpgrade = versionHistory.find((versionData) => versionData.id.match(version));
+      if (findUpgrade) relevantUpgrades.push(findUpgrade);
+    });
+
+    // if no history for 1.9 or 1.10, get first upgrade available
+    if (Object.entries(relevantUpgrades).length === 0) {
```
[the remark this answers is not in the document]
hellobontempo: Oh thank you! Nice catch - It had been an object initially, until I realized we needed to store multiple upgrade info for the charts

ui/app/components/clients/current.js
```diff
@@ -15,10 +16,30 @@ export default class Current extends Component {
...
+      relevantUpgrades.push({
```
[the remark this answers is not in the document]
hellobontempo: If a customer skips the relevant upgrades, for example is on 1.8 and jumps to 1.11, then we need to capture whatever their first stored upgrade date is. We still want to show a warning to explain any potential data discrepancies: 

image

ui/app/components/clients/history.js
hellobontempo: That's a good suggestion, for the whole getter?  Or just line 108 - I'll make a note and pulling it out may be a good call come test-writing time.

ui/app/components/clients/current.js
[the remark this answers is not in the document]
hellobontempo: But good point that we do need to account for if a vault customer _starts_ at a version >= 1.11 
Since `previousVersion = null` whether a user upgrades from pre-1.9 or their first version of vault is >= 1.11 we have no way of telling the difference

ui/tests/integration/components/replication-dashboard-test.js
```diff
@@ -44,7 +44,7 @@ module('Integration | Component | replication-dashboard', function (hooks) {
     assert.dom('[data-test-table-rows]').exists();
     assert.dom('[data-test-selectable-card-container="secondary"]').exists();
     assert.dom('[data-test-replication-doc-link]').exists();
-    assert.dom('[data-test-alert-banner]').doesNotExist('no flash message is displayed on render');
+    assert.dom('[data-test-flash-message]').doesNotExist('no flash message is displayed on render');
```
hellobontempo, line 47: didn't mean to change this in a previous PR

PR #15104, opened 2022-04-20
[VAULT-5887] TypeInt64 support added to OpenApi Spec generation

AnPucel:
`framework.TypeInt64` was not caught when converting to OpenApi types and would result in a confusing warning when calling `/v1/sys/internal/specs/openapi`

Verified by building Vault and generating an OpenApi spec and verifying we don't have any unknown formats in our spec.


digivava (approved):
Makes sense!

averche (commented):
Looks good! It might be a good idea to add a unit test case for this as well in openapi_test.go

PR #15105, opened 2022-04-20
Add back metadata.json

hashishaw:
Adds back a file which the UI no longer needs but turned out to break the main vault build. 

The file was removed in #15074 and we'll need to investigate how to safely remove this file in the future


swenson (approved):
LGTM

arnav28 (approved):

PR #15106, opened 2022-04-20
Disabling LinkTo Component

zofskeez:
During some recent deprecation work the `@disabled` argument on `LinkTo` components was mistakenly converted to an attribute. After double checking [the deprecation](https://deprecations.emberjs.com/v3.x/#toc_ember-built-in-components-legacy-arguments), `@disabled` is still a named argument so it was converted back. There was also some styling issues with disabled links using the `toolbar-link` class which were fixed.

### DIFF
--- a/ui/app/styles/components/toolbar.scss
+++ b/ui/app/styles/components/toolbar.scss
@@ -111,6 +111,15 @@
     }
   }
 }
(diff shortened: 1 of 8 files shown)


Monkeychip (approved):
Nice work!

PR #15117, opened 2022-04-21
docs/agent: merge template and template config

jasonodonnell:
This merges the `template_config` and `template` stanza documentation into a single page, broken into global and per template sections.


website/content/docs/agent/template.mdx
```diff
@@ -80,50 +142,13 @@ The top level `template` block has multiple configurations entries:
...
-```
-{{ with secret "secret/my-secret" }}
-{{ .Data.data.foo }}
-{{ end }}
-```
 
-## Example Configuration
+### Example
 
 The following demonstrates configuring Vault Agent to template secrets using the
```
swenson, line 148: I think this paragraph no longer applies to this example.

website/content/docs/agent/template.mdx
```diff
@@ -171,3 +196,45 @@ using the certificates `validTo` field.
 
 This does not apply to certificates generated with `generate_lease: true`. If set
 Vault Agent template will apply the non-renewable, leased secret rules.
+
+## Example
```
swenson, line 200: Should this header be updated to be more specific than "Example"?

website/content/docs/agent/template.mdx
jasonodonnell: Good catch!

website/content/docs/agent/template.mdx
jasonodonnell: Updated.

swenson (approved):
LGTM. It's much better to have these in one place. :)

PR #15120, opened 2022-04-21
PKI read legacy CA bundle prior to migration

stevendpclark:
When the new Vault version starts up and becomes the leader node on a secondary performance cluster it should be able to still support certificate issue/revoke and list commands even when the storage was not migrated.

### Known limitations

The new api on the secondary does not return the proper information in regards to issuers and keys as we don't fallback for those api calls.

### Testing performed

- Started up a primary and secondary performance clusters on 1.10 setting up a full CA within two separate mounts. 
- Upgraded the secondary performance cluster and verified that I could issue a new certificate from it as well as revoking a cert. 
- Upgraded the primary and verified that the new issuers api started working without restarting Vault on the secondary performance cluster.

### DIFF
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -260,15 +260,15 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
 }
 
 // initialize is used to perform a possible PKI storage migration if needed
-func (b *backend) initialize(ctx context.Context, req *logical.InitializationRequest) error {
+func (b *backend) initialize(ctx context.Context, _ *logical.InitializationRequest) error {
 	// Early exit if not a primary cluster or performance secondary with a local mount.
 	if b.System().ReplicationState().HasState(consts.ReplicationDRSecondary|consts.ReplicationPerformanceStandby) ||
 		(!b.System().LocalMount() && b.System().ReplicationState().HasState(consts.ReplicationPerformanceSecondary)) {
-		b.Logger().Debug("skipping pki migration as we are not on primary or secondary with a local mount")
+		b.Logger().Debug("skipping PKI migration as we are not on primary or secondary with a local mount")
 		return nil
 	}
 
-	if err := migrateStorage(ctx, req, b.Logger()); err != nil {
+	if err := migrateStorage(ctx, b.storage, b.Logger()); err != nil {
 		b.Logger().Error("Error during migration of PKI mount: " + err.Error())
 		return err
 	}
@@ -291,11 +291,9 @@ func (b *backend) updatePkiStorageVersion(ctx context.Context) {
 	}
 
 	if info.isRequired {
-		b.Logger().Info("PKI migration status is required, reading cert bundle from legacy ca location")
+		b.Logger().Info("PKI migration is required, reading cert bundle from legacy ca location")
 		b.pkiStorageVersion.Store(0)
-	}
-
-	if !info.isRequired {
+	} else {
 		b.Logger().Debug("PKI migration completed, reading cert bundle from key/issuer storage")
 		b.pkiStorageVersion.Store(1)
 	}
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/hashicorp/vault/sdk/helper/consts"
@@ -139,12 +140,15 @@ func Backend(conf *logical.BackendConfig) *backend {
 
 		BackendType:    logical.TypeLogical,
 		InitializeFunc: b.initialize,
+		Invalidate:     b.invalidate,
 	}
 
 	b.crlLifetime = time.Hour * 72
 	b.tidyCASGuard = new(uint32)
 	b.tidyStatus = &tidyStatus{state: tidyStatusInactive}
 	b.storage = conf.StorageView
+
+	b.pkiStorageVersion.Store(0)
 
 	return &b
 }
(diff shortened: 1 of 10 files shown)


builtin/logical/pki/backend.go
```diff
@@ -252,19 +258,52 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
...
+}
+
+func (b *backend) invalidate(ctx context.Context, key string) {
+	switch {
+	case strings.HasPrefix(key, legacyMigrationBundleLogKey):
+		// This is for a secondary cluster to pick up that the migration has completed
+		// and reset its compatibility mode.
+		b.updatePkiStorageVersion(ctx)
+	}
+	// FIXME: We need to hook into CRL generation here for issuer/bundle updates.
```
cipherboy, line 308: Since we merged the CRL stuff, a call to `buildCRLs` should suffice here. :-)

builtin/logical/pki/storage_migrations.go
```diff
@@ -16,55 +16,88 @@ import (
...
+		isRequired:       false,
+		legacyBundle:     nil,
+		legacyBundleHash: "",
+	}
+	var err error
+
+	migrationInfo.legacyBundle, err = getLegacyCertBundle(ctx, s)
 	if err != nil {
-		return err
+		return migrationInfo, err
```
cipherboy, line 42: If  this has been deleted (or never created), does it mean we'll err out here?

builtin/logical/pki/backend.go
```diff
@@ -252,19 +258,52 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
...
+
+func (b *backend) useLegacyBundleCaStorage() bool {
+	version := b.pkiStorageVersion.Load()
+	return version == nil || version == 0
+}
+
+func (b *backend) updatePkiStorageVersion(ctx context.Context) {
+	info, err := getMigrationInfo(ctx, b.storage)
+	if err != nil {
+		b.Logger().Error(fmt.Sprintf("Failed loading PKI migration status, staying in legacy mode: %v", err))
```
cipherboy, line 286: And related to the above, wondering if that means the migration status will be incorrect here? Namely, the `if !info.isRequired {...}` check later might need to be moved up?

builtin/logical/pki/storage_migrations.go
cipherboy: Ah, I think we don't err. Only on other storage issues.

builtin/logical/pki/cert_util.go
```diff
@@ -134,6 +126,25 @@ func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRe
...
+func fetchCertBundle(ctx context.Context, b *backend, s logical.Storage, issuerRef string) (*certutil.CertBundle, error) {
+	if b.useLegacyBundleCaStorage() {
+		// We have not completed the migration so attempt to load the bundle from the legacy location
+		return getLegacyCertBundle(ctx, s)
+	}
+
+	id, err := resolveIssuerReference(ctx, s, issuerRef)
+	if err != nil {
+		// Usually a bad label from the user or misconfigured default.
+		return nil, err
```
cipherboy, line 142: Not to nitpick, but doesn't this turn this into an `InternalError` versus a `UserError`? Not sure how much the distinction matters, but seems like we'd want to preserve that.

builtin/logical/pki/path_intermediate.go
```diff
@@ -47,6 +47,10 @@ appended to the bundle.`,
 func (b *backend) pathGenerateIntermediate(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
 	var err error
 
+	if b.useLegacyBundleCaStorage() {
+		return logical.ErrorResponse("Can not create intermediary until migration has completed"), nil
```
cipherboy, line 51: nit: `s/intermediary/intermediate/g` perhaps, for consistent language?

cipherboy (approved):
Largely looks good, some minor nits I think. 

Will this PR handle updating the `buildCRL` path (now that it has merged) or should I do that after this merges? 

Also, not sure if the complete "new APIs should error" will be done here now or in a separate PR :-)

builtin/logical/pki/cert_util.go
stevendpclark: Yup good catch, that was a flub on my part.

builtin/logical/pki/storage_migrations.go
stevendpclark: Correct, but this comment led me to a missing nil check within fetchCAInfo so thanks 👍

builtin/logical/pki/backend.go
stevendpclark: If you are referring to the empty/non-existant ca-bundle then that won't be a problem. Please let me know if I have misunderstood your concern.

cipherboy: Yeah, it was about the above comment, but I see that's wrong, feel free to ignore this one :-)

builtin/logical/pki/backend.go
stevendpclark: Will defer to another PR, since we don't have a request object here this isn't as trivial as calling buildCRLs.

builtin/logical/pki/backend.go
```diff
@@ -254,20 +260,51 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
...
+		return
+	}
+
+	if info.isRequired {
+		b.Logger().Info("PKI migration status is required, reading cert bundle from legacy ca location")
```
cipherboy, line 294: nit: perhaps drop the status?

PR #15121, opened 2022-04-21
Refactor to account for chart indicator

hellobontempo:
TO DO: 
- [ ] Current tab view
<hr>

version-warnings

### DIFF
--- a/ui/app/components/clients/history.js
+++ b/ui/app/components/clients/history.js
@@ -120,7 +120,7 @@ export default class History extends Component {
     const activityStart = new Date(this.getActivityResponse.startTime);
     const activityEnd = new Date(this.getActivityResponse.endTime);
     const upgradesWithinData = this.upgradeVersionHistory.filter((upgrade) => {
-      // will timezones be an issue because new Date defaults to current timezone?
+      // TODO how do timezones affect this?
       let upgradeDate = new Date(upgrade.timestampInstalled);
       return isAfter(upgradeDate, activityStart) && isBefore(upgradeDate, activityEnd);
     });
--- a/ui/app/components/clients/history.js
+++ b/ui/app/components/clients/history.js
@@ -5,6 +5,7 @@ import { tracked } from '@glimmer/tracking';
 import { isSameMonth, isAfter, isBefore } from 'date-fns';
 import getStorage from 'vault/lib/token-storage';
 import { ARRAY_OF_MONTHS } from 'core/utils/date-formatters';
+import { dateFormat } from 'core/helpers/date-format';
 
 const INPUTTED_START_DATE = 'vault:ui-inputted-start-date';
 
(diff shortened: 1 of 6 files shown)


ui/app/components/clients/history.js
```diff
@@ -89,48 +90,68 @@ export default class History extends Component {
...
-    // version id is 1.9.0 or earliest upgrade post 1.9.0, timestamp is RFC3339
-    // need to return the last in the array as the data is in ascending order
-    return this.args.model.versionHistory.at(-1) || null;
+  get upgradeVersionHistory() {
+    const versionHistory = this.args.model.versionHistory;
+    if (!versionHistory || versionHistory.length === 0) {
+      return null;
+    }
+
+    // get upgrade data for initial upgrade to 1.9 and/or 1.10
```
hellobontempo, line 99: First - we only care about an upgrade to '1.9' and/or '1.10'. We are also only concerned when the _first_ upgrade to the relative minor version (subsequent patch releases, within either minor version, are irrelevant)

ui/app/components/clients/history.js
```diff
@@ -89,48 +90,68 @@ export default class History extends Component {
...
+    let relevantUpgrades = [];
+    const importantUpgrades = ['1.9', '1.10'];
+    importantUpgrades.forEach((version) => {
+      let findUpgrade = versionHistory.find((versionData) => versionData.id.match(version));
+      if (findUpgrade) relevantUpgrades.push(findUpgrade);
+    });
+
```
hellobontempo, line 106: If neither upgrade is present, then we want just the first upgrade that happened (all upgrades will be post 1.9, because that was when the `version-history` endpoint starts)

ui/app/components/clients/history.js
```diff
@@ -89,48 +90,68 @@ export default class History extends Component {
...
+  get upgradeDuringActivity() {
+    const activityStart = new Date(this.getActivityResponse.startTime);
+    const activityEnd = new Date(this.getActivityResponse.endTime);
+    const upgradesWithinData = this.upgradeVersionHistory.filter((upgrade) => {
+      // will timezones be an issue because new Date defaults to current timezone?
+      let upgradeDate = new Date(upgrade.timestampInstalled);
+      return isAfter(upgradeDate, activityStart) && isBefore(upgradeDate, activityEnd);
     });
-    return newArray; // return [9,9,10] shows all the minor numbers that fall within the range of the returned API response.
+    // return all upgrades that happened within date range of queried activity
```
hellobontempo, line 127: We need all the data for upgrades within a date range so we can display the pertinent information in the line chart

ui/app/components/clients/history.js
```diff
@@ -89,48 +90,68 @@ export default class History extends Component {
...
+      return `Vault was upgrade to ${upgrade.id} on ${dateFormat(
+        [upgrade.timestampInstalled, 'MMM d, yyyy'],
+        { isFormatted: true }
+      )}.`;
+    }
+  }
+
+  get versionSpecificText() {
+    if (this.upgradeDuringActivity.length === 1) {
+      let version = this.upgradeDuringActivity[0].id;
```
hellobontempo, line 145: `version.match('1.9') will match regardless of what the patch release is, so both '1.9.1' or '1.9.5' will hit the correct `if` block

ui/app/components/clients/line-chart.js
```diff
@@ -90,6 +93,10 @@ export default class LineChart extends Component {
 
     chartSvg.selectAll('.domain').remove();
 
+    const findUpgradeData = (datum) => {
```
hellobontempo, line 93: If multiple upgrades exist, we want to indicate all of them on the chart

ui/app/components/clients/history.js
Monkeychip: I like this approach, but don't we want to search the reverse of the array, e.g. from the back? Find the most recent first? Or maybe it doesn't matter.

Monkeychip (approved):

Issue #15124, opened 2022-04-22
Update golang.org/x/crypto/ssh to patch CVE-2022-27191


rcjames:
**Describe the bug**
There is a vulnerability in the version of the crypto/ssh package used in Vault. I have raised this as per the security policy and been informed that this is not exploitable, but as the email mentions "The Vault team also welcomes community contributions" I assumed this was a hint that a PR to update the version might be accepted.

CVE-2022-27191 - https://github.com/advisories/GHSA-8c26-wmh5-6g9v

PR #15125, opened 2022-04-22
Update golang.org/x/crypto/ssh

rcjames:
Patch vulnerability in golang.or/x/crypto. Addresses https://github.com/hashicorp/vault/issues/15124


cipherboy (approved):
Looks great, many thanks for the contribution @rcjames

PR #15128, opened 2022-04-22
change action to pull v1 instead of main

claire-labry:
The `generate-metadata` action needs to be pulling from tags for better release management. We run the risk of disruption if we pushed changes to the `main` branch of the generate-metadata action thus disrupting release workflows for Vault. 

Read more on [Using tags for release management](https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management)

I added the backport labels for this PR, but I'm unsure which release branches Vault is using, so feel free to adjust!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Generate metadata file
         id: generate-metadata-file
-        uses: hashicorp/actions-generate-metadata@main
+        uses: hashicorp/actions-generate-metadata@v1
         with:
           version: ${{ needs.get-product-version.outputs.product-version }}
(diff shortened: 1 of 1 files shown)


mladlow (approved):
You got the active release branches. Approving with two questions - definitely seems like this needs to run on main and all the release branches, but I wanted to confirm that.

Second - this should merge smoothly to enterprise? We have the same file over there, so I am pretty sure there's no further action required on the enterprise side but wanted to get you to confirm.

claire-labry:
That all sounds correct @mladlow -- I'll merge this in and keep an eye on the release branches and on the enterprise side to ensure this change gets in. If not, I'll manually open up a PR to get them in. 

Thank you!

PR #15140, opened 2022-04-22
Import SDK for 1.10.2 version

schultz-is:
Imports SDK version bump from #15137


mladlow:
I don't fully understand why but I believe this should be the second of two PRs you'll need, where the first is editing the files in the SDK, and then you'll need to update this PR after that first one has merged (unless I missed the first one somewhere).

schultz-is:
Yeah, sorry, I ran the first round of SDK version PRs by @ccapurso. They've been merged in for all three release branches.

mladlow (approved):

PR #15141, opened 2022-04-22
Import sdk for 1.9.6 version

schultz-is:


mladlow (commented):
I don't know how to validate the gocql and snappy changes, and I _think_ snappy caused trouble once in the past.

schultz-is:
I think that's a result of the `go mod tidy`. I'll add those back in just to be super safe.

sgmiller (approved):

PR #15143, opened 2022-04-22
KMSE Capabilities & Phase 1 Cleanup

zofskeez:
- Added capabilities to key and provider models
- Hid toolbar actions when user does not have capability
- Fixed some issues with key-edit component and updated some parts to match designs
- Wired up distribute actions in key and provider views

### DIFF
--- a/ui/app/models/keymgmt/provider.js
+++ b/ui/app/models/keymgmt/provider.js
@@ -2,6 +2,7 @@ import Model, { attr } from '@ember-data/model';
 import { tracked } from '@glimmer/tracking';
 import { expandAttributeMeta } from 'vault/utils/field-to-attrs';
 import { withModelValidations } from 'vault/decorators/model-validations';
+import lazyCapabilities, { apiPath } from 'vault/macros/lazy-capabilities';
 
 const CRED_PROPS = {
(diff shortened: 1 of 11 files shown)


ui/app/models/keymgmt/provider.js
```diff
@@ -104,18 +109,48 @@ export default class KeymgmtProviderModel extends Model {
...
+      try {
+        this.keys = await this.store.lazyPaginatedQuery('keymgmt/key', {
+          backend: 'keymgmt',
+          provider: this.name,
+          responsePath: 'data.keys',
+          page,
+        });
+      } catch (error) {
+        this.keys = [];
+        if (error.httpStatus !== 404) {
```
hashishaw, line 122: Gut check: This is because the response is a 404 if the provider has no distributed keys, correct?

hashishaw (approved):
Beautiful ✨

ui/app/models/keymgmt/provider.js
zofskeez: As far as I can remember yes :slightly_smiling_face:. That was added in phase 1 but from what I recall if the provider doesn't have any associated keys then we just want to update `this.keys` rather than surfacing an error.

PR #15152, opened 2022-04-25
Expand NotAfter limit enforcement behavior

cipherboy:
Vault previously strictly enforced `NotAfter`/`ttl` values on certificate
requests, erring if the requested TTL extended past the `NotAfter` date of
the issuer. In the event of issuing an intermediate, this behavior was
ignored, instead permitting the issuance.

Users generally do not think to check their issuer's `NotAfter` date when
requesting a certificate; thus this behavior was generally surprising.

[Per RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.5) however, issuers need to maintain status information
throughout the life cycle of the issued cert. If this leaf cert were to
be issued for a longer duration than the parent issuer, the CA must
still maintain revocation information past its expiration.

Thus, we add an option to the issuer to change the desired behavior:

 - `err`, to err out,
 - `permit`, to permit the longer `NotAfter` date, or
 - `truncate`, to silently truncate the expiration to the issuer's
   `NotAfter` date.

Since expiration of certificates in the system's trust store are not
generally validated (when validating an arbitrary leaf, e.g., during TLS
validation), permit should generally only be used in that case. However,
browsers usually validate intermediate's validity periods, and thus
truncate should likely be used (as with permit, the leaf's chain will
not validate towards the end of the issuance period).

### DIFF
--- a/builtin/logical/pki/path_fetch_issuers.go
+++ b/builtin/logical/pki/path_fetch_issuers.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/helper/certutil"
 	"github.com/hashicorp/vault/sdk/logical"
 )
 
@@ -66,10 +67,23 @@ func pathGetIssuer(b *backend) *framework.Path {
 func buildPathGetIssuer(b *backend, pattern string) *framework.Path {
 	fields := map[string]*framework.FieldSchema{}
 	fields = addIssuerRefNameFields(fields)
+
+	// Fields for updating issuer.
 	fields["manual_chain"] = &framework.FieldSchema{
 		Type: framework.TypeCommaStringSlice,
 		Description: `Chain of issuer references to use to build this
 issuer's computed CAChain field, when non-empty.`,
+	}
+	fields["leaf_not_after_behavior"] = &framework.FieldSchema{
+		Type: framework.TypeString,
+		Description: `Behavior of leaf's NotAfter fields: "err" to error
+if the computed NotAfter date exceeds that of this issuer; "truncate" to
+silently truncate to that of this issuer; or "permit" to allow this
+issuance to succeed (with NotAfter exceeding that of an issuer). Note that
+not all values will results in certificates that can be validated through
+the entire validity period. It is suggested to use "truncate" for
+intermediate CAs and "permit" only for root CAs.`,
+		Default: "err",
 	}
 
 	return &framework.Path{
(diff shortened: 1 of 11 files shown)


cipherboy:
Can be tested with:

```
source devvault
vault secrets enable pki 
vault write pki/root/generate/internal common_name="root example.com" issuer_name=root  ttl=15s
vault write /pki/roles/local-testing allow_any_name=true enforce_hostnames=false no_store=true
sleep 16s
vault write pki/issue/local-testing common_name=hello # should err
vault write pki/issuer/root issuer_name=root leaf_not_after_behavior=permit 
vault write pki/issue/local-testing common_name=hello # should be default ttl of a month or so
vault write pki/issuer/root issuer_name=root leaf_not_after_behavior=truncate
vault write pki/issue/local-testing common_name=hello # should be expired
vault write pki/issuer/root issuer_name=root leaf_not_after_behavior=err
vault write pki/issue/local-testing common_name=hello # should err
```

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -158,6 +173,18 @@ func (b *backend) pathUpdateIssuer(ctx context.Context, req *logical.Request, da
 	}
 
 	newPath := data.Get("manual_chain").([]string)
+	rawLeafBehavior := data.Get("leaf_not_after_behavior").(string)
```
stevendpclark, line 176: Does this not fail the update api call if we don't specify the `leaf_not_after_behavior` argument?

builtin/logical/pki/storage_migrations.go
```diff
@@ -150,21 +150,27 @@ func setLegacyBundleMigrationLog(ctx context.Context, s logical.Storage, lbm *le
 	return s.Put(ctx, json)
 }
 
-func getLegacyCertBundle(ctx context.Context, s logical.Storage) (*certutil.CertBundle, error) {
+func getLegacyCertBundle(ctx context.Context, s logical.Storage) (*issuerEntry, *certutil.CertBundle, error) {
```
stevendpclark, line 153: nit?: The migration works properly, matches with legacy behaviour, due to the iota definition of `NotAfterBehavior` the value `ErrNotAfterBehavior` comes first. Should we either be explicit within the migration perhaps of the value we want or test that ErrNotAfterBehavior. == 0?

stevendpclark (commented):
Overall 👍, I just have two comments that I *believe* might need some tweaks.

builtin/logical/pki/storage_migrations.go
cipherboy: Wouldn't `ImportIssuer...` be the correct place to set this explicitly? Not sure what makes import on migration different than others, but yes, zero value being err was intentional :-)

builtin/logical/pki/path_fetch_issuers.go
stevendpclark: Yup you are correct I keep forgetting about the defaults...

builtin/logical/pki/storage_migrations.go
stevendpclark: Well the others to my knowledge have the value set through either the default on the parameter definition or end-user values, this one really is just up to the iota. I don't think we can put it within importIssuer as there are so many end-users of that function.

How about just adding a check within `storage_migrations_test.go` that a migrated issuer has a value of `ErrNotAfterBehavior` set? I'd just want something to break if that iota accidently gets re-ordered.

cipherboy: `importIssuer` is already opinionated though, in other forms: we automatically construct a chain (rather than giving the option to specify a manual chain at import time), and when calling e.g., `/config/ca`, we allow importing multiple issuers, and so there's really no way to control name either.

stevendpclark: Good points, let's go with adding it to `importIssuer`.

cipherboy:
Thanks @stevendpclark!

cipherboy:

Issue #15147, opened 2022-04-23
panic: non-positive interval for NewTicker on 1.10.1 after upgrade


arren-ru:
**Describe the bug**
After upgrading from 1.10.0 to 1.10.1 unable to start vault, after unsealing it crashes with Go panic.

**Environment:**
* Vault Server and CLI Versions are 1.10.1
* Server Operating System/Architecture: archlinux with systemd

Vault server configuration file(s):
```hcl
ui = true
disable_mlock = true

storage "raft" {
  path = "/var/lib/vault"
}

api_addr = "https://127.0.0.1:8200"
cluster_addr = "https://127.0.0.1:8201"

listener "tcp" {
  address       = ":8200"
  tls_cert_file = "***.crt"
  tls_key_file  = "***.key"
}

service_registration "consul" {
  address         = "127.0.0.1:8501"
  service_address = "10.10.0.10"
  scheme          = "https"
  token           = "***"
}
```

Crash logs are:
```
Apr 24 01:12:26 helios vault[696416]: 2022-04-24T01:12:26.899+0700 [INFO]  core: starting raft active node
Apr 24 01:12:26 helios vault[696416]: 2022-04-24T01:12:26.899+0700 [INFO]  storage.raft: starting autopilot: config="&{false 0 10s 24h0m0s 1000 0 10s}" reconcile_interval=0s
Apr 24 01:12:26 helios vault[696416]: panic: non-positive interval for NewTicker
Apr 24 01:12:26 helios vault[696416]: goroutine 2346 [running]:
Apr 24 01:12:26 helios vault[696416]: time.NewTicker(0xdf8475800?)
Apr 24 01:12:26 helios vault[696416]:         time/tick.go:24 +0x10f
Apr 24 01:12:26 helios vault[696416]: github.com/hashicorp/vault/vault.(*Core).raftTLSRotatePhased.func3()
Apr 24 01:12:26 helios vault[696416]:         github.com/hashicorp/vault/vault/raft.go:486 +0x10a
Apr 24 01:12:26 helios vault[696416]: created by github.com/hashicorp/vault/vault.(*Core).raftTLSRotatePhased
Apr 24 01:12:26 helios vault[696416]:         github.com/hashicorp/vault/vault/raft.go:477 +0x42a
Apr 24 01:12:26 helios systemd[1]: vault.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
[...]

PR #15154, opened 2022-04-25
Add upgrade note for #15147.

ncabatoff:


website/content/docs/upgrading/upgrade-to-1.10.x.mdx
```diff
@@ -77,6 +77,8 @@ naming collisions could result unexpected default behavior. Additionally, we rec
 the corresponding details in the OIDC provider [concepts](/docs/concepts/oidc-provider) document
 to understand how the built-in resources are used in the system.
 
+@include 'raft-panic-old-tls-key.mdx'
```
mladlow, line 80: Should this be under the Known Issues section for each of these pages?

website/content/partials/raft-panic-old-tls-key.mdx
```diff
@@ -0,0 +1,17 @@
...
+replication don't lose the ability to speak with other nodes, the newly generated daily
+TLS cert only starts being used once we see that all nodes have received it.
+
+A recent security audit related change results in this rotation code [getting a
+panic](https://github.com/hashicorp/vault/issues/15147) when the current cert is
+more than 24h old.  This can happen if the cluster as a whole is down for a day
+or more. It can also happen if a single node is unreachable 24h, or sufficiently
+backlogged in applying raft logs that it's more than a day behind.
+
+Impacted versions: 1.10.1, 1.9.5, 1.8.10.
```
mladlow, line 15: It might be worth a parenthetical that 1.10.0 is not affected - this is pretty clearly implied by what you've written but I can imagine the question being asked anyway.

website/content/partials/raft-panic-old-tls-key.mdx
```diff
@@ -0,0 +1,17 @@
+## Integrated Storage panic related to old TLS key
```
mladlow, line 1: If you put this in the known issues the heading should be at the ### level.

website/content/partials/raft-panic-old-tls-key.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+Raft in Vault uses its own set of TLS certificates, independent of those that the user
```
mladlow, line 3: The whole issue writeup is really good - solid background and description of when it can occur.

mladlow (approved):
I'm ok with this as is though I left some suggestions. I don't feel the need to re-review it. It needs the "backport/1.10.x" label to show up on the website immediately, and as you've updated the upgrade notes for 1.8 and 1.9 could probably benefit by having all 3 backport labels.

website/content/docs/upgrading/upgrade-to-1.10.x.mdx
ncabatoff: I'm trying to follow the existing pattern for embedding partials, even though I don't like it, because otherwise I'll have to change a bunch of things.

ccapurso (approved):
Looks good!

Issue #2075, opened 2016-11-08
PKI CA chain == CA certificate?


prudnitskiy:
Hello all!

I built a two-level PKI (root PKI and intermediate PKI signed by root PKI) by instructions found [here](https://www.vaultproject.io/docs/secrets/pki/) and [here](http://cuddletech.com/?p=959). When I try to get ca_chain from intermediate PKI I can see only a CA certificate from that PKI (but not all other certificates in chain). Is it a normal behaviour?

Vault version is 0.6.2

frlod:
@jefferai Would this work if I provide a pem bundle (private key + CA chain) when submitting an already-signed intermediate certificate using the POST pki/config/ca path?

I would also like the whole ca_chain to be returned, but the difference is that I do not generate the intermediate key with vault, but I generate it and sign it with an outside root, and then submit it into vault.

Thanks!

bhperry:
@frlod Yes that works. Just tested it, and vault is able to correctly parse out the issuing_ca from the chain, while also returning the full ca_chain cert that was provided.

cyberwave:
hello everyone!  I want to learn about PKI and CA. Who has the relevant book recommendation?

chrishoffman:
@cyberwave If you are looking for some book recommendations, the vault mailing list would be a better resource to ask the community.  See https://groups.google.com/forum/#!forum/vault-tool.

aantono:
@jefferai Is there any way a more detailed step-by-step instructions can be provided?  I've been searching Google Groups and internet for examples, but nothing came up as to how to add the Root Public Certificate component to be returned when asking for a `ca_chain`! :(

yusuf-gunaydin:
You can get the root certificate from `/v1/pki/cert/ca` and append it to the signed intermediate certificate with a new line.
I use the following Ansible script:

```yml
- name: "Enable PKI secret engine."
  uri:
    url: "https://{{ vault.ingress.host_path }}/v1/sys/mounts/pki_{{ item }}"
    method: POST
    headers:
      X-Vault-Token: "{{ vault_root_token }}"
    body_format: json
    body:
      type: pki
      config:
        max_lease_ttl: "43800h" # 5 years
    status_code:
      - 200
      - 204

- name: "Generate intermediate certificate."
  uri:
    url: "https://{{ vault.ingress.host_path }}/v1/pki_{{ item }}/intermediate/generate/internal"
    method: POST
    return_content: yes
    headers:
[...]

PR #15155, opened 2022-04-25
Update CA Chain to report entire chain

cipherboy:
This merges the `ca_chain` JSON field (of the `/certs/ca_chain` path) with
the regular `certificate` field, returning the root of trust always. This
also affects the non-JSON (raw) endpoints as well.

We return the default issuer's chain here, rather than all known issuers
(as that may not form a strict chain).

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -1773,7 +1773,7 @@ func TestBackend_PathFetchValidRaw(t *testing.T) {
 	}
 	rootCaAsPem := resp.Data["certificate"].(string)
 
-	// The ca_chain call at least for now does not return the root CA authority
+	// Chain should contain the root.
 	resp, err = b.HandleRequest(context.Background(), &logical.Request{
 		Operation:  logical.ReadOperation,
 		Path:       "ca_chain",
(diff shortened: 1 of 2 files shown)


stevendpclark (approved):

stevendpclark:
Note: we will need to address the comment within the Vault documentation about this not returning the full chain and to use pki/cert/ca_chain `ca_chain` field https://www.vaultproject.io/api-docs/secret/pki#read-ca-certificate-chain

cipherboy:
Yeah, that's a good point. There's a lot of docs that will need to be updated. :-)

cipherboy:

epieddy:
Hello, I'm using vault 1.11.0 and this is still not working : when I issue a certificat from an intermediate CA,  the field ca_chain in the json response contains only the intermediate CA, not the root CA

epieddy:
> Hello, I'm using vault 1.11.0 and this is still not working : when I issue a certificat from an intermediate CA, the field ca_chain in the json response contains only the intermediate CA, not the root CA

My bad,  you need to upload the full chain (intermediate CA + root CA) when you upload the signed intermediate CA with set-signed (https://github.com/hashicorp/vault/issues/2075#issuecomment-259169330)

cipherboy:
@epieddy As a note, you can generically import chain certificates now in Vault 1.11.0; see: https://www.vaultproject.io/api-docs/secret/pki#import-ca-certificates-and-keys

PR #15156, opened 2022-04-25
fix raft tls key rotation panic when rotation time in past

ccapurso:
Raft TLS key rotation will panic if the active key's creation time is more than 24 hours the past. Calculating the duration using `time.Until` will produce a negative value which causes time.NewTicker` to panic. This PR introduces a fix for this issue by creating a ticker with a duration equal to 1 minute to effectively cause rotation to occur immediately.

Fixes: https://github.com/hashicorp/vault/issues/15147

### DIFF
--- a/changelog/15156.txt
+++ b/changelog/15156.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+rafft: fix Raft TLS key rotation panic that occurs if active key is more than 24 hours old
+```
--- a/vault/raft.go
+++ b/vault/raft.go
@@ -500,7 +500,6 @@ func (c *Core) raftTLSRotatePhased(ctx context.Context, logger hclog.Logger, raf
 		ticker := time.NewTicker(time.Until(nextRotationTime))
 		defer ticker.Stop()
 		for {
-			ticker.Reset(time.Until(nextRotationTime))
 			select {
 			case <-keyCheckInterval.C:
 				err := checkCommitted()
@@ -516,6 +515,8 @@ func (c *Core) raftTLSRotatePhased(ctx context.Context, logger hclog.Logger, raf
 				} else {
 					nextRotationTime = getNextRotationTime(next)
 				}
+
+				ticker.Reset(time.Until(nextRotationTime))
 
 			case <-stopCh:
 				return
--- a/vault/raft.go
+++ b/vault/raft.go
@@ -495,19 +495,11 @@ func (c *Core) raftTLSRotatePhased(ctx context.Context, logger hclog.Logger, raf
 		keyCheckInterval := time.NewTicker(1 * time.Minute)
 		defer keyCheckInterval.Stop()
 
-		var backoff bool
 		// ticker is used to prevent memory leak of using time.After in
 		// for - select pattern.
 		ticker := time.NewTicker(time.Until(nextRotationTime))
 		defer ticker.Stop()
 		for {
-			// If we encountered and error we should try to create the key
-			// again.
-			if backoff {
-				nextRotationTime = time.Now().Add(10 * time.Second)
-				backoff = false
-			}
-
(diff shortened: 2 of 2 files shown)


vault/raft.go
```diff
@@ -473,9 +473,24 @@ func (c *Core) raftTLSRotatePhased(ctx context.Context, logger hclog.Logger, raf
...
 
+	getNextRotationTime := func(next time.Time) time.Time {
+		now := time.Now()
+
+		// active key's CreatedTime + raftTLSRotationPeriod might be in
+		// the past (meaning it is ready to be rotated) which will cause
+		// NewTicker to panic when used with time.Until, prevent this by
+		// pushing out rotation time into very near future
+		if next.Before(now) {
+			return now.Add(1 * time.Minute)
```
ncabatoff, line 484: I think this is fine, though I think it's not quite what the behaviour was before, right?  Prior to the breaking change, I assume a negative next would be treated as "immediately", not "now+1m"?

vault/raft.go
```diff
@@ -509,7 +524,7 @@ func (c *Core) raftTLSRotatePhased(ctx context.Context, logger hclog.Logger, raf
 					continue
 				}
 
-				nextRotationTime = next
+				nextRotationTime = getNextRotationTime(next)
```
ncabatoff, line 527: Can we move the ticker.Reset to here, as well as eliminate the backoff variable and just put all the time calculation logic here?  Feels like it would be safer and clearer.

vault/raft.go
ccapurso: That's true but with `NewTicker` I think all you can really do is try to be as close to "immediately" as possible. The reason for adding 1m was just to try to add enough buffer to prevent the negative `NewTicker` issue. It needs to be some time in the future in other words.

vault/raft.go
ccapurso: Is your thought to put all of this logic into the current `getNextRotationTime` (maybe with a different name)? That might be tricky as we have to worry about rotation time for the initialization of the ticker, the reset, and the handling of the error returned by the `rotateKeyring` in the case of the backoff.

ncabatoff: No, I propose to leave getNextRotationTime alone, but get rid of the backoff bool, the `continue`, and just do all the calculations for nextRotationTime here in this `case`.

ncabatoff: That way we can have more confidence that the padding we add to nextRotationTime is big enough (and possibly can be smaller than 1m), because we'll update the ticker with it immediately.

ccapurso: Ah sorry I misunderstood, moved things around a bit.

vault/raft.go
ccapurso: I reduced the padding to be 10 seconds. I'll admit that the 1 minute was a bit aggressive.

vault/raft.go
```diff
@@ -505,12 +512,11 @@ func (c *Core) raftTLSRotatePhased(ctx context.Context, logger hclog.Logger, raf
 				next, err := rotateKeyring()
 				if err != nil {
 					logger.Error("failed to rotate TLS key", "error", err)
-					backoff = true
-					continue
+					nextRotationTime = time.Now().Add(10 * time.Second)
+				} else {
+					nextRotationTime = getNextRotationTime(next)
 				}
```
ncabatoff, line 518: Could move the ticker.Reset here too.

ccapurso: Ah, whoops. Done

ncabatoff (approved):

changelog/15156.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+rafft: fix Raft TLS key rotation panic that occurs if active key is more than 24 hours old
```
kalafut, line 2: typo: rafft

PR #15163, opened 2022-04-25
[VAULT-5813] Remove duplicate sha_256 in SystemCatalogRequest OAS

AnPucel:
Currently, there is a duplicate `sha256` and `sha_256` field in the OpenApi Spec for SystemCatalogRequest. This causes issues for codegen because they get mapped to the same variable name. We're removing the `sha_256` in the OpenApi spec to avoid this.


averche (approved):
LGTM, thanks for adding the comment 👍

PR #15164, opened 2022-04-25
Allow explicit issuer override on roles

cipherboy:
When a role is used to generate a certificate (such as with the `sign/`
and `issue/` legacy paths or the legacy `sign-verbatim/` paths), we prefer
that issuer to the one on the request. This allows operators to set an
issuer (other than default) for requests to be issued against,
effectively making the change no different from the users' perspective
as it is "just" a different role name.


stevendpclark (approved):

cipherboy:
Thanks @stevendpclark! Merging... :)

cipherboy:

PR #15166, opened 2022-04-25
VAULT-5827 Don't prepare SQL queries before executing them

swenson:
We don't support proper prepared statements, i.e., preparing once and
executing many times since we do our own templating. So preparing our
queries does not really accomplish anything, and can have severe
performance impacts (see
https://github.com/hashicorp/vault-plugin-database-snowflake/issues/13
for example).

This behavior seems to have been copy-pasted for many years but not for
any particular reason that we have been able to find. First use was in
https://github.com/hashicorp/vault/pull/15

So here we switch to new methods suffixed with `Direct` to indicate
that they don't `Prepare` before running `Exec`, and switch everything
here to use those. We maintain the older methods with the existing
behavior (with `Prepare`) for backwards compatibility.

### DIFF
--- a/builtin/logical/mssql/path_creds_create.go
+++ b/builtin/logical/mssql/path_creds_create.go
@@ -95,7 +95,7 @@ func (b *backend) pathCredsCreateRead(ctx context.Context, req *logical.Request,
 			"name":     username,
 			"password": password,
 		}
-		if err := dbtxn.ExecuteTxQuery(ctx, tx, m, query); err != nil {
+		if err := dbtxn.ExecuteTxQueryDirect(ctx, tx, m, query); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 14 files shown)


builtin/logical/mssql/path_creds_create.go
```diff
@@ -95,7 +95,7 @@ func (b *backend) pathCredsCreateRead(ctx context.Context, req *logical.Request,
 			"name":     username,
 			"password": password,
 		}
-		if err := dbtxn.ExecuteTxQuery(ctx, tx, m, query); err != nil {
+		if err := dbtxn.ExecuteTxQueryDirect(ctx, tx, m, query); err != nil {
```
calvn, line 98: The database secret engines under `builtin/logical/` that's not `database/` such as this one (aka "Standalone DB Engines") are deprecated and will be [removed in the near future](https://www.vaultproject.io/docs/deprecation). Doesn't hurt to update them, but note that we will be dropping them soon.

tomhjp (approved):
LGTM

builtin/logical/mssql/path_creds_create.go
swenson: I was just changing all uses of the existing functions in `vault` :)

swenson:
Thanks everyone!

PR #15167, opened 2022-04-25
Handle client count timezone

arnav28:
- Backend convert the timezone to UTC, to mitigate it's impact sending
  start and end date other than 1. Chose 10 and 20 randomly.

### DIFF
--- a/ui/app/adapters/clients/activity.js
+++ b/ui/app/adapters/clients/activity.js
@@ -2,19 +2,21 @@ import Application from '../application';
 import { formatRFC3339 } from 'date-fns';
 
 export default Application.extend({
+  // Since backend converts the timezone to UTC, sending the first (1) as start or end date can cause the month to change.
+  // To mitigate this impact of timezone conversion, hard coding the dates to avoid month change.
   formatTimeParams(query) {
     let { start_time, end_time } = query;
     // check if it's an array, if it is, it's coming from an action like selecting a new startTime or new EndTime
     if (Array.isArray(start_time)) {
       let startYear = Number(start_time[0]);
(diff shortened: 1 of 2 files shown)


ui/app/adapters/clients/activity.js
```diff
@@ -2,19 +2,21 @@ import Application from '../application';
...
       let startMonth = Number(start_time[1]);
-      start_time = formatRFC3339(new Date(startYear, startMonth));
+      start_time = formatRFC3339(new Date(startYear, startMonth, 10));
     }
     if (end_time) {
       if (Array.isArray(end_time)) {
         let endYear = Number(end_time[0]);
         let endMonth = Number(end_time[1]);
-        end_time = formatRFC3339(new Date(endYear, endMonth));
+        end_time = formatRFC3339(new Date(endYear, endMonth, 20));
```
hellobontempo, line 19: When testing did the API always return with the last day of the month? Might be worth having backend take a look (or maybe you already have) to see if sending an arbitrary middle of the month day has any implications we're unaware of.

hellobontempo (approved):
I think my original question still applies here - I think a backend dev reviewing would be helpful! Just to make sure this won't introduce anything unexpected. Otherwise thanks for tackling! Nice to have this finally merged :)

arnav28:
> I think my original question still applies here - I think a backend dev reviewing would be helpful! Just to make sure this won't introduce anything unexpected. Otherwise thanks for tackling! Nice to have this finally merged :)

Yes, confirmed with backend that they have startOfMonth, endOfMonth for start and end params. I tested it as well. 👍 
Sorry forgot to mention during our discussion.

PR #15169, opened 2022-04-25
CSV Export include monthly data

Monkeychip:
This chart cleans up the CSV export for the attribution chart and updates the component accordingly. There have been some design updates and that's why this PR got so large. New client data is only available at a monthly breakdown, for historical months. (Previously it was thought that new client data would be available in the current/partial month as well). 

There's a lot of cleanup here to get everything in a working state to move forward with refactoring to accommodate the data generated with the newly created tool, using the real Vault API.

Side-by-side attribution chart only if single month queried in history tab:

double-chart

<hr>
sample of CSV exports:

### clients by namespace 
Screen Shot 2022-05-02 at 5 41 36 PM

_single historical month (with new counts)_

Screen Shot 2022-05-02 at 5 51 08 PM

### clients by auth 
Screen Shot 2022-05-02 at 5 41 49 PM

_single historical month (with new counts)_

Screen Shot 2022-05-02 at 5 50 18 PM

### DIFF
--- a/ui/app/components/clients/attribution.js
+++ b/ui/app/components/clients/attribution.js
@@ -112,9 +112,10 @@ export default class Attribution extends Component {
     return [clients, entity_clients, non_entity_clients];
   }
 
-  constructCsvRow(namespaceColumn, mountColumn = null, otherColumns) {
+  constructCsvRow(namespaceColumn, mountColumn = null, totalColumns, newColumns = null) {
     // if namespaceColumn is a string, then we're at mount level attribution, otherwise it is an object
     // if constructing a namespace row, mountColumn=null so the column is blank, otherwise it is an object
+    let otherColumns = newColumns ? [...totalColumns, ...newColumns] : [...totalColumns];
     return [
       `${typeof namespaceColumn === 'string' ? namespaceColumn : namespaceColumn.label}`,
       `${mountColumn ? mountColumn.label : ''}`,
@@ -122,7 +123,8 @@ export default class Attribution extends Component {
     ];
   }
   generateCsvData() {
-    let graphData = this.args.totalClientAttribution;
+    const totalAttribution = this.args.totalClientAttribution;
+    const newAttribution = this.barChartNewClients ? this.args.newClientAttribution : null;
     let csvData = [],
       csvHeader = [
         'Namespace path',
@@ -132,18 +134,35 @@ export default class Attribution extends Component {
         'Non-entity clients',
       ];
 
-    graphData.forEach((dataObject) => {
-      let namespace = this.isSingleNamespace ? this.args.selectedNamespace : dataObject;
-      let mount = this.isSingleNamespace ? dataObject : null;
-      let otherColumns = this.destructureCountsToArray(dataObject);
-      csvData.push(this.constructCsvRow(namespace, mount, otherColumns));
-      // constructCsvRow returns an array that corresponds to a row in the csv file:
-      // ['ns label', 'mount label', total client #, entity #, non-entity #]
+    if (newAttribution) {
+      csvHeader = [...csvHeader, 'Total new clients, New entity clients, New non-entity clients'];
+    }
 
-      // only iterate through mounts if NOT filtered and viewing a single namespace
+    totalAttribution.forEach((totalClientsObject) => {
+      let namespace = this.isSingleNamespace ? this.args.selectedNamespace : totalClientsObject;
+      let mount = this.isSingleNamespace ? totalClientsObject : null;
+
+      // find new client data for namespace/mount object we're iterating over
+      let newClientsObject = newAttribution
+        ? newAttribution.find((d) => d.label === totalClientsObject.label)
+        : null;
+
+      let totalClients = this.destructureCountsToArray(totalClientsObject);
+      let newClients = newClientsObject ? this.destructureCountsToArray(newClientsObject) : null;
+
+      csvData.push(this.constructCsvRow(namespace, mount, totalClients, newClients));
+      // constructCsvRow returns an array that corresponds to a row in the csv file:
+      // ['ns label', 'mount label', total client #, entity #, non-entity #, ...new client #'s]
+
+      // only iterate through mounts if NOT viewing a single namespace
       if (!this.isSingleNamespace && namespace.mounts) {
         namespace.mounts.forEach((mount) => {
-          csvData.push(this.constructCsvRow(namespace, mount, otherColumns));
+          let newMountData = newAttribution
+            ? newClientsObject?.mounts.find((m) => m.label === mount.label)
+            : null;
+          let mountTotalClients = this.destructureCountsToArray(mount);
+          let mountNewClients = newMountData ? this.destructureCountsToArray(newMountData) : null;
+          csvData.push(this.constructCsvRow(namespace, mount, mountTotalClients, mountNewClients));
         });
       }
     });
--- a/ui/app/components/clients/attribution.js
+++ b/ui/app/components/clients/attribution.js
@@ -13,8 +13,8 @@ import { inject as service } from '@ember/service';
  *    @chartLegend={{this.chartLegend}}
  *    @totalUsageCounts={{this.totalUsageCounts}}
  *    @newUsageCounts={{this.newUsageCounts}}
- *    @totalClientsData={{this.totalClientsData}}
- *    @newClientsData={{this.newClientsData}}
+ *    @totalClientAttribution={{this.totalClientAttribution}}
+ *    @newClientAttribution={{this.newClientAttribution}}
  *    @selectedNamespace={{this.selectedNamespace}}
  *    @startTimeDisplay={{date-format this.responseTimestamp "MMMM yyyy"}}
  *    @isDateRange={{this.isDateRange}}
@@ -25,8 +25,8 @@ import { inject as service } from '@ember/service';
  * @param {array} chartLegend - (passed to child) array of objects with key names 'key' and 'label' so data can be stacked
  * @param {object} totalUsageCounts - object with total client counts for chart tooltip text
  * @param {object} newUsageCounts - object with new client counts for chart tooltip text
- * @param {array} totalClientsData - array of objects containing a label and breakdown of client counts for total clients
- * @param {array} newClientsData - array of objects containing a label and breakdown of client counts for new clients
+ * @param {array} totalClientAttribution - array of objects containing a label and breakdown of client counts for total clients
+ * @param {array} newClientAttribution - array of objects containing a label and breakdown of client counts for new clients
  * @param {string} selectedNamespace - namespace selected from filter bar
  * @param {string} startTimeDisplay - string that displays as start date for CSV modal
  * @param {string} endTimeDisplay - string that displays as end date for CSV modal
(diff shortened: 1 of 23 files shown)


ui/app/components/clients/attribution.js
```diff
@@ -36,6 +36,7 @@ import { inject as service } from '@ember/service';
 export default class Attribution extends Component {
   @tracked showCSVDownloadModal = false;
   @service downloadCsv;
+  @service store;
```
Monkeychip, line 39: i think store can be removed

ui/lib/core/addon/utils/client-count-utils.js
```diff
@@ -0,0 +1,113 @@
+import { parseAPITimestamp } from 'core/utils/date-formatters';
+
+export const flattenDataset = (object) => {
+  if (Object.values(object).every((e) => e === null)) return object;
+  let flattenedObject = {};
+  Object.keys(object['counts']).forEach((key) => (flattenedObject[key] = object['counts'][key]));
+  return homogenizeClientNaming(flattenedObject);
+};
+
+export const formatByMonths = (monthsArray) => {
```
hellobontempo, line 10: TODO: this function will be refactored now that we have a way to generate mock data with the real API

ui/lib/core/addon/utils/client-count-utils.js
```diff
@@ -0,0 +1,113 @@
...
+        new_clients: {
+          month: parseAPITimestamp(m.timestamp, 'M/yy'),
+          ...newClients,
+          namespaces: formatByNamespace(m.new_clients.namespaces),
+        },
+      };
+      return nestCountsWithinNamespaceKey(m);
+    }
+    // TODO CMB below is an assumption, need to test
+    // if no monthly data (no counts key), month object will just contain a timestamp
```
hellobontempo, line 32: Following up, this is not true. Month object will not exist

ui/lib/core/addon/utils/client-count-utils.js
```diff
@@ -0,0 +1,113 @@
...
+    return {
+      clients,
+      entity_clients: distinct_entities,
+      non_entity_clients: non_entity_tokens,
+    };
+  }
+  return object;
+};
+
+export const nestCountsWithinNamespaceKey = (month) => {
```
hellobontempo, line 90: WIP function to make filtering by namespace id easier

zofskeez (commented):
Good work! Nothing really stood out to me as far as changes but given the size maybe it's a good idea to get another review from someone closer to the project?

ui/app/components/clients/attribution.js
```diff
@@ -103,9 +104,24 @@ export default class Attribution extends Component {
     }
   }
 
-  get getCsvData() {
+  destructureCountsToArray(object) {
+    let { clients, entity_clients, non_entity_clients } = object;
```
Monkeychip, line 108: are clients, entity_clients, etc. here are numbers?  e.g. clients = 100? or are they returning objects. It maybe helpful for a future developer if making changes to understand what is expected. So maybe just a comment explaining the type returned. Because we don't use something like typescript, it's not always easy to understand what is expected.

ui/app/components/clients/attribution.js
```diff
@@ -103,9 +104,24 @@ export default class Attribution extends Component {
...
+    return [clients, entity_clients, non_entity_clients];
+  }
+
+  constructCsvRow(firstColumn, secondColumn = null, otherColumns) {
+    // ['ns label', 'mount label', 'total clients', 'entity', 'non-entity']
+    // if constructing namespace row, mount column will be blank
```
Monkeychip, line 114: maybe ... "_the_ mount _label_ column will be blank. On first read I was expecting there to be a column named mount.

ui/app/components/clients/attribution.js
```diff
@@ -103,9 +104,24 @@ export default class Attribution extends Component {
...
+    // if firstColumn is a string, then we're at mount level attribution
```
Monkeychip, line 115: NIT: but is it an array if it's not a string? Maybe add to comment.

Monkeychip: Again this is where typescript helps a user read the code. i can tell from the firstColumn.label below it's an object, but might help a reader understand what's expected if the type is called out.  (maybe a larger discussion to have as a team about calling out types in comments since we don't typescript).

ui/app/components/clients/attribution.js
```diff
@@ -103,9 +104,24 @@ export default class Attribution extends Component {
...
+    return [
+      `${typeof firstColumn === 'string' ? firstColumn : firstColumn.label}`,
+      `${secondColumn ? secondColumn.label : ''}`,
+      ...otherColumns,
+    ];
+  }
+  generateCsvData() {
+    let graphData = this.args.totalClientAttribution;
```
Monkeychip, line 123: what happens if totalClientAttribution is empty?  Does it return an empty array?  I ask because of the forEach on 134.

ui/app/components/clients/attribution.js
```diff
@@ -114,24 +130,20 @@ export default class Attribution extends Component {
...
-        if (ns.mounts) {
-          ns.mounts.forEach((m) => {
-            csvData.push([ns.label, m.label, m.clients, m.entity_clients, m.non_entity_clients]);
-          });
-        }
-      });
-    }
+    // each array is a row in the csv file
+    graphData.forEach((data) => {
+      let namespace = this.isSingleNamespace ? this.args.selectedNamespace : data;
```
Monkeychip, line 135: I might be wrong about this for this situation, but if this.isSingleNamespace changes, you might want to make it a tracked property. The way I think of it, is in the old non-glimmer computed property style. If namespace here was a computed property and needed to watch if isSingleNamespace changed, then namespace should be a getter and isSingleNamespace should be a tracked property.

Again, this may not be relevant for this code structure, but thought it worth pointing out.

ui/app/components/clients/attribution.js
```diff
@@ -114,24 +130,20 @@ export default class Attribution extends Component {
...
+      csvData.push(this.constructCsvRow(namespace, mount, otherColumns));
+      if (!this.isSingleNamespace && namespace.mounts) {
+        namespace.mounts.forEach((mount) => {
+          csvData.push(this.constructCsvRow(namespace, mount, otherColumns));
+        });
+      }
+    });
     csvData.unshift(csvHeader);
-    // make each nested array a comma separated string, join each array in csvData with line break (\n)
+    // make each nested array a comma separated string, join each array "row" in csvData with line break (\n)
```
Monkeychip, line 146: great comment.

ui/app/components/clients/history.js
```diff
@@ -85,8 +86,8 @@ export default class History extends Component {
 
   get isDateRange() {
     return !isSameMonth(
-      new Date(this.getActivityResponse.startTime),
-      new Date(this.getActivityResponse.endTime)
+      parseAPITimestamp(this.getActivityResponse.startTime),
```
Monkeychip, line 89: does this function become irrelevant if the backend makes any of its "tasked" changes? If so maybe make a comment to revisit if so and so it done. That way if someone else does the work in the future the old pattern can easily be tracked down. Ignore me if this will always be here regardless of future backend changes.

ui/app/components/clients/history.js
```diff
@@ -193,38 +185,58 @@ export default class History extends Component {
     return this.queriedActivityResponse || this.args.model.activity;
   }
 
+  get byMonthTotalClients() {
+    return this.getActivityResponse?.byMonth;
```
Monkeychip, line 189: This situation is a little confusing because you use this.getActivityResponse in the definition of a tracked property but getActivityResponse itself is not a tracked property but is used to define a getter. Normally any property in a getter I add as a tracked property, but I'm not sure in this case. Just throwing it out there as maybe something to test. Re-query the date range and put a console log in this getter and make sure it fires when getActivityResponse changes.

ui/lib/core/addon/utils/client-count-utils.js
```diff
@@ -0,0 +1,89 @@
+import { parseAPITimestamp } from 'core/utils/date-formatters';
```
Monkeychip, line 1: nice call on making this a util.

ui/app/components/clients/attribution.js
hellobontempo: great idea! thank you

ui/app/components/clients/attribution.js
hellobontempo: `firstColumn` is a string if we're within a namespace, but an object if we're iterating through a namespace. I had abstracted to be `firstColumn` and `secondColumn` - but as this is the only place this is used right now, changed the variable names for clarity. 

I definitely want the comment to be clear! Is this better? 

```
  constructCsvRow(namespaceColumn, mountColumn = null, otherColumns) {
    // returns an array that corresponds to a row in the csv file:
    // ['ns label', 'mount label', total client #, entity #, non-entity #]
    // if namespaceColumn is a string, then we're at mount level attribution, otherwise it is an object
    // if constructing a namespace row, the mountColumn=null so the column is blank, otherwise it is an object
    return [
      `${typeof namespaceColumn === 'string' ? namespaceColumn : namespaceColumn.label}`,
      `${mountColumn ? mountColumn.label : ''}`,
      ...otherColumns,
    ];
  }
```

ui/app/components/clients/attribution.js
hellobontempo: higher up in this file there is this getter, so the export button only shows if there is data to iterate and export

  ```
  get hasCsvData() {
    return this.args.totalClientAttribution ? this.args.totalClientAttribution.length > 0 : false;
  }
  ```

ui/app/components/clients/attribution.js
hellobontempo: Hmm, `this.isSingleNamespace` relies on the `selectedNamespace` arg which _is_ a tracked property in the parent. So the getter should update appropriately based on the actions from the parent, right? So the tracked property in the parent changing, would trigger the getter to update - right?

ui/app/components/clients/history.js
hellobontempo: I was planning to make a ticket on our end, that's linked to the backend, to revisit warning banners if/when those backend changes happen. I'll mention this part of the code!

ui/app/components/clients/history.js
hellobontempo: Yep! It all updates. `this.getActivityResponse` changes anytime there is a network request. My understanding is that tracked properties are best used for state changes from the user? 

Also confirmed with console logs all works as expected!

ui/app/components/clients/current.js
```diff
@@ -29,71 +29,38 @@ export default class Current extends Component {
...
     const auth = this.selectedAuthMethod;
     if (!namespace && !auth) {
-      return this.byNamespaceTotalClients;
+      return this.byNamespace;
     }
     if (!auth) {
-      return this.byNamespaceTotalClients.find((ns) => ns.label === namespace);
+      return this.byNamespace.find((ns) => ns.label === namespace);
     }
-    return this.byNamespaceTotalClients
```
hellobontempo, line 82: Removing as no longer showing new vs total client counts in current/partial month view

ui/mirage/handlers/clients.js
```diff
@@ -1,15 +1,128 @@
...
-  isAfter,
-  isBefore,
-  sub,
-  isSameMonth,
-  startOfMonth,
-} from 'date-fns';
+import { formatISO, isAfter, isBefore, sub, isSameMonth, startOfMonth } from 'date-fns';
 import { parseAPITimestamp } from 'core/utils/date-formatters';
 const MOCK_MONTHLY_DATA = [
+  {
```
hellobontempo, line 4: add month out of order to make sure `.sort()` in serializer functions properly

hellobontempo (approved):

PR #15170, opened 2022-04-25
Use precompiled regexes and move OpenAPI call to tests

digivava:
The `digivava/output-policy` branch currently grabs the sudo paths from the OpenAPI spec at runtime.

This way, like @averche 's suggestion, the sudo paths will be a static list which we will update only when the TestSudoPaths test fails.

The keys in the `sudoPaths` map are strings so we can quickly grab a URL by its string-key name in the scenario where no regex matching is needed, and so that anyone who wants a list of sudo paths can fetch it from this endpoint in the format the OpenAPI spec originally returned. The values of the map are regexes, so that consumers of the map can do regex matching when needed.


api/plugin_helpers.go
```diff
@@ -23,6 +24,35 @@ var (
 	// PluginUnwrapTokenEnv is the ENV name used to pass unwrap tokens to the
 	// plugin.
 	PluginUnwrapTokenEnv = "VAULT_UNWRAP_TOKEN"
+
+	sudoPaths = map[string]*regexp.Regexp{
+		"/auth/token/accessors/":                        regexp.MustCompile("^/auth/token/accessors/$"),
```
averche, line 29: This path ends with a `/` whereas others don't. Is this a typo or intended?

api/plugin_helpers.go
```diff
@@ -23,6 +24,35 @@ var (
...
+		"/sys/config/ui/headers/":                       regexp.MustCompile("^/sys/config/ui/headers/$"),
+		"/sys/config/ui/headers/{header}":               regexp.MustCompile("^/sys/config/ui/headers/.+$"),
+		"/sys/leases":                                   regexp.MustCompile("^/sys/leases$"),
+		"/sys/leases/lookup/":                           regexp.MustCompile("^/sys/leases/lookup/$"),
+		"/sys/leases/lookup/{prefix}":                   regexp.MustCompile("^/sys/leases/lookup/.+$"),
+		"/sys/leases/revoke-force/{prefix}":             regexp.MustCompile("^/sys/leases/revoke-force/.+$"),
+		"/sys/leases/revoke-prefix/{prefix}":            regexp.MustCompile("^/sys/leases/revoke-prefix/.+$"),
+		"/sys/plugins/catalog/{name}":                   regexp.MustCompile("^/sys/plugins/catalog/.+$"),
+		"/sys/plugins/catalog/{type}":                   regexp.MustCompile("^/sys/plugins/catalog/.+$"),
+		"/sys/plugins/catalog/{type}/{name}":            regexp.MustCompile("^/sys/plugins/catalog/.+/.+$"),
```
averche, line 48: Should we limit then non-`{path}` regexes to something like `[\w-_]+` instead of `.+` or would it be too restrictive?

api/output_policy.go
```diff
@@ -88,96 +86,22 @@ func (d *OutputPolicyError) buildSamplePolicy() (string, error) {
...
-		return false, fmt.Errorf("unable to retrieve list of paths that require sudo capability: %v", err)
-	}
-	if sudoPaths == nil || len(sudoPaths) < 1 {
-		// OpenAPI spec did not return any paths that require sudo,
-		// but the user probably still shouldn't see an error.
-		return false, nil
-	}
-
-	// Return early if the path is clearly one of the sudo paths.
+func isSudoPath(path string) (bool, error) {
```
VinnyHC, line 89: I don't think we need to return an error

VinnyHC: I don't think we need to return an error

api/output_policy.go
```diff
@@ -88,96 +82,22 @@ func (d *OutputPolicyError) buildSamplePolicy() (string, error) {
...
 	// so we can check if our path matches against them.
-	for sudoPath := range sudoPaths {
-		r, err := regexp.Compile(fmt.Sprintf("^%s$", sudoPath))
-		if err != nil {
-			continue
-		}
-
-		match := r.Match([]byte(fmt.Sprintf("/%s", path))) // the OpenAPI response has a / in front of each path
+	for _, sudoPathRegexp := range sudoPaths {
+		match := sudoPathRegexp.Match([]byte(fmt.Sprintf("/%s", path))) // the OpenAPI response has a / in front of each path
```
averche, line 96:
```suggestion
		match := sudoPathRegexp.MatchString(fmt.Sprintf("/%s", path)) // the OpenAPI response has a / in front of each path
```

api/output_policy.go
```diff
@@ -68,11 +66,11 @@ func (d *OutputPolicyError) buildSamplePolicy() (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to unescape request URL characters: %v", err)
 	}
-	apiAddrPrefix := fmt.Sprintf("%sv1/", d.VaultClient.config.Address)
+	apiAddrPrefix := fmt.Sprintf("%sv1/", d.vaultServerAddress)
```
VinnyHC, line 69: I don't necessarily mind this logic, but I'm surprised that the bits you're interested in aren't available as part of `*retryablehttp.Request`

VinnyHC (commented):
This all seems reasonable, one comment about simplifying `isSudoPath`; I have some follow up questions but I think they are better suited for the main PR.

api/output_policy.go
```diff
@@ -68,15 +66,11 @@ func (d *OutputPolicyError) buildSamplePolicy() (string, error) {
...
```
averche, line 69: Is `vaultServerAddress` used anywhere else? If not, maybe it's possible to get rid of it and just take the string after `"v1"` as the path.

api/output_policy.go
averche: It also seems like the "/" is removed first before the call to `isSudoPath` and then added back here. Maybe it would be simpler to keep it in the original?

api/plugin_helpers.go
```diff
@@ -23,6 +24,35 @@ var (
 	// PluginUnwrapTokenEnv is the ENV name used to pass unwrap tokens to the
 	// plugin.
 	PluginUnwrapTokenEnv = "VAULT_UNWRAP_TOKEN"
+
+	sudoPaths = map[string]*regexp.Regexp{
```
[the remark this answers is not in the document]
VinnyHC: In other parts of Vault I've seen the variables declared with a `Re` suffix when it's a regexp; not positive this applies here but might help explain what's going on.

[OpenAPI Example RegExp naming](https://github.com/hashicorp/vault/blob/main/sdk/framework/openapi.go#L203)

averche (commented):
I definitely like this new approach 👍 Left a few nit-picky comments/suggestions.

api/output_policy.go
VinnyHC: I think you might be able to use `strings.TrimPrefix(d.URL.Path, "/v1")` and avoid passing in the server address altogether.

api/client.go
```diff
@@ -1253,16 +1253,8 @@ START:
...
-		outputPolicyClient, err := c.Clone()
-		if err != nil {
-			return nil, fmt.Errorf("unable to clone client for -output-policy: %v", err)
-		}
-		outputPolicyClient.config.OutputCurlString = false
-		outputPolicyClient.config.OutputPolicy = false
 		LastOutputPolicyError = &OutputPolicyError{
-			Request:     req,
-			VaultClient: outputPolicyClient,
+			Request: req,
```
VinnyHC, line 1257: This is a little confusing since I think `Request` is actually part of the `*retryablehttp.Request` that you're embedding. Since you've done all this cleanup I'd suggest just taking in the method and path during the creation of the `OutputPolicyError`

api/plugin_helpers.go
```diff
@@ -192,3 +228,28 @@ func VaultPluginTLSProviderContext(ctx context.Context, apiTLSConfig *TLSConfig)
 		return tlsConfig, nil
 	}
 }
+
+func GetSudoPaths() map[string]*regexp.Regexp {
+	return sudoPaths
+}
+
+// Determine whether the given path requires the sudo capability
+func IsSudoPath(path string) bool {
```
averche, line 237: Great idea moving the func here 👍

api/plugin_helpers.go
```diff
@@ -192,3 +228,28 @@ func VaultPluginTLSProviderContext(ctx context.Context, apiTLSConfig *TLSConfig)
...
+	if _, ok := sudoPaths[path]; ok {
+		return true
+	}
+
+	// Some sudo paths have templated fields in them.
+	// (e.g. /sys/revoke-prefix/{prefix})
+	// The values in the sudoPaths map are actually regular expressions,
+	// so we can check if our path matches against them.
+	for _, sudoPathRegexp := range sudoPaths {
+		match := sudoPathRegexp.Match([]byte(path))
```
averche, line 248:
```suggestion
		match := sudoPathRegexp.MatchString(path)
```

averche (approved):
LGTM 👍

api/plugin_helpers.go
```diff
@@ -192,3 +228,28 @@ func VaultPluginTLSProviderContext(ctx context.Context, apiTLSConfig *TLSConfig)
 		return tlsConfig, nil
 	}
 }
+
+func GetSudoPaths() map[string]*regexp.Regexp {
```
VinnyHC, line 232: Tiny Nit: I don't think I've seen many cases where `Get` prefixes a "getter". For example, in a lot of the [client code](https://github.com/hashicorp/vault/blob/main/api/client.go#L689-L707), I see setters prefixed with `Set` and getters just the name of the item.

VinnyHC (approved):
This is really awesome and the sudo paths approach gives a solid framework to build on!

PR #15178, opened 2022-04-26
Add warning on keyless default issuer

cipherboy:

---

```
[cipherboy@xps15 pki]$ vault write pki/config/issuers default=5a3ae525-d25f-57e6-9cf5-b7acb4c8db9b

  * This issuer has no key associated with it. Some operations like issuing
  certificates and signing CRLs will be unavailable with the requested default
  issuer.

Key        Value
---        -----
default    5a3ae525-d25f-57e6-9cf5-b7acb4c8db9b

[cipherboy@xps15 vault]$ vault write pki/issuers/import/cert pem_bundle=@$HOME/root-no-key.pem

  * The default issuer has no key associated with it. Some operations like
  issuing certificates and signing CRLs will be unavailable with the requested
  default issuer until a key is imported or the default issuer is changed.

Key                 Value
---                 -----
imported_issuers    [5f7bafc5-ddf8-6e7c-2e62-dbb19ba43d20]
imported_keys       <nil>
mapping             map[5f7bafc5-ddf8-6e7c-2e62-dbb19ba43d20:]

```

### DIFF
--- a/builtin/logical/pki/path_config_ca.go
+++ b/builtin/logical/pki/path_config_ca.go
@@ -83,16 +83,29 @@ func (b *backend) pathCAIssuersWrite(ctx context.Context, req *logical.Request, 
 		return logical.ErrorResponse("Error resolving issuer reference: " + err.Error()), nil
 	}
 
+	response := &logical.Response{
+		Data: map[string]interface{}{
+			"default": parsedIssuer,
+		},
+	}
+
+	entry, err := fetchIssuerById(ctx, req.Storage, parsedIssuer)
+	if err != nil {
+		return logical.ErrorResponse("Unable to fetch issuer: " + err.Error()), nil
+	}
+
+	if len(entry.KeyID) == 0 {
+		msg := "This selected default issuer has no key associated with it. Some operations like issuing certificates and signing CRLs will be unavailable with the requested default issuer until a key is imported or the default issuer is changed."
+		response.AddWarning(msg)
+		b.Logger().Error(msg)
+	}
+
 	err = updateDefaultIssuerId(ctx, req.Storage, parsedIssuer)
 	if err != nil {
 		return logical.ErrorResponse("Error updating issuer configuration: " + err.Error()), nil
 	}
 
-	return &logical.Response{
-		Data: map[string]interface{}{
-			"default": parsedIssuer,
-		},
-	}, nil
+	return response, nil
 }
 
 const pathConfigIssuersHelpSyn = `Read and set the default issuer certificate for signing.`
--- a/builtin/logical/pki/path_manage_issuers.go
+++ b/builtin/logical/pki/path_manage_issuers.go
@@ -187,6 +187,14 @@ func (b *backend) pathImportIssuers(ctx context.Context, req *logical.Request, d
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


builtin/logical/pki/path_config_ca.go
```diff
@@ -83,16 +83,27 @@ func (b *backend) pathCAIssuersWrite(ctx context.Context, req *logical.Request,
...
+		},
+	}
+
+	entry, err := fetchIssuerById(ctx, req.Storage, parsedIssuer)
+	if err != nil {
+		return logical.ErrorResponse("Unable to fetch issuer: " + err.Error()), nil
+	}
+
+	if len(entry.KeyID) == 0 {
+		response.AddWarning("This issuer has no key associated with it. Some operations like issuing certificates and signing CRLs will be unavailable with the requested default issuer.")
```
stevendpclark, line 98: Since this is so detrimental to the legacy api functioning, would it be worth also logging this to the Vault's log to possibly have support identify this as well?

stevendpclark (approved):
Overall +1, with a suggestion

builtin/logical/pki/path_config_ca.go
cipherboy: The issue is this isn't the only place where the default issuer can become disassociated from a key. We'd have to check, e.g., delete issuers, importing an issuer (with a cert and no key as this'll take the default slot), and perhaps some other paths for logging. Maybe its better to check at time of use? But `fetchCA...` currently already has resolved the issuer to an ID prior to the call I think, so it'd be hard to detect if it was the default, unless we cache that on the backend (which might actually be good)?

cipherboy: Ok, I've updated this for the import case, @kitography will need to add it to their delete keys case to see if the deleted key was for the default issuer.

cipherboy: Fixed.

builtin/logical/pki/path_config_ca.go
```diff
@@ -83,16 +83,29 @@ func (b *backend) pathCAIssuersWrite(ctx context.Context, req *logical.Request,
...
+
+	entry, err := fetchIssuerById(ctx, req.Storage, parsedIssuer)
+	if err != nil {
+		return logical.ErrorResponse("Unable to fetch issuer: " + err.Error()), nil
+	}
+
+	if len(entry.KeyID) == 0 {
+		msg := "This selected default issuer has no key associated with it. Some operations like issuing certificates and signing CRLs will be unavailable with the requested default issuer until a key is imported or the default issuer is changed."
+		response.AddWarning(msg)
+		b.Logger().Error(msg)
```
stevendpclark, line 100: ultimate nit: Warn vs Error?

cipherboy: I think we should error; fundamentally a portion of the feature space (and backwards compatibility) is broken in this specific configuration setup. My 2c.

stevendpclark: Sounds good.

stevendpclark (approved):

cipherboy:
Thanks! Merged...

cipherboy:

PR #15180, opened 2022-04-26
Update PKI to new Operations framework

cipherboy:
The backend Framework has updated `Callbacks` (used extensively in PKI) to
become deprecated; `Operations` takes their place and clarifies forwarding
of requests.

We switch to the new format everywhere, updating some bad assumptions
about forwarding along the way. Anywhere writes are handled (that should
be propagated to all nodes in all clusters), we choose to forward the
request all the way up to the performance primary cluster's primary
node. This holds for issuers/keys, roles, and configs (such as CRL
config, which is globally set for all clusters despite all clusters
having their own separate CRL).

### DIFF
--- a/builtin/logical/pki/path_revoke.go
+++ b/builtin/logical/pki/path_revoke.go
@@ -22,8 +22,14 @@ hyphen-separated octal`,
 			},
 		},
 
-		Callbacks: map[logical.Operation]framework.OperationFunc{
-			logical.UpdateOperation: b.metricsWrap("revoke", noRole, b.pathRevokeWrite),
+		Operations: map[logical.Operation]framework.OperationHandler{
+			logical.UpdateOperation: &framework.PathOperation{
+				Callback: b.metricsWrap("revoke", noRole, b.pathRevokeWrite),
+				// This should never be forwarded. See backend.go for more information.
+				// If this needs to write, the entire request will be forwarded to the
+				// active node of the current performance cluster, but we don't want to
+				// forward invalid revoke requests there.
+			},
 		},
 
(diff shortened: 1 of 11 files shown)


builtin/logical/pki/path_revoke.go
```diff
@@ -22,8 +22,11 @@ hyphen-separated octal`,
 			},
 		},
 
-		Callbacks: map[logical.Operation]framework.OperationFunc{
-			logical.UpdateOperation: b.metricsWrap("revoke", noRole, b.pathRevokeWrite),
+		Operations: map[logical.Operation]framework.OperationHandler{
+			logical.UpdateOperation: &framework.PathOperation{
+				Callback: b.metricsWrap("revoke", noRole, b.pathRevokeWrite),
+				// This should never be forwarded. See backend.go for more information.
```
cipherboy, line 28: @stevendpclark Should this be updated to mirror the Tidy operation endpoint, and forward this to the primary node of the cluster since it may write in the case that the cert wasn't set to no-store?

stevendpclark: I think this is better, just leave the lower layers within Vault do their thing. We might as well get rid of the `ForwardPerformanceStandby: true` on the tidy operations while we are at it, it's not incorrect but again let the lower levels of Vault do their thing.

cipherboy: From our conversation, I've left this one and the Tidy one alone, but updated the `crl/rotate` operation to forward automatically.

stevendpclark (approved):

cipherboy:
Thanks! Merged...

PR #15188, opened 2022-04-26
[VAULT-14990] Support retrieving kv secret paths with trailing spaces

AnPucel:
Currently you can create secret paths with trailing spaces with the api in KV. We sanitize those in the cli, which means you cannot retrieve the secret path you previously created. 

- Update the CLI to allow you to retrieve secret paths with trailing spaces: `vault kv get "secret/test "`. The trailing space will not be removed.

### DIFF
--- a/command/base_helpers.go
+++ b/command/base_helpers.go
@@ -38,18 +38,6 @@ func sanitizePath(s string) string {
 	return ensureNoTrailingSlash(ensureNoLeadingSlash(s))
 }
 
-// hasTrailingSpace checks to see if there is trailing space
-func hasTrailingSpace(s string) bool {
-	return strings.HasSuffix(s, " ")
-}
-
-// hasTrailingSlash boolean checks if a trailing slash exists
-func hasTrailingSlash(s string) bool {
-	s = strings.TrimSuffix(s, " ")
-
-	return strings.HasSuffix(s, "/")
-}
-
 // ensureTrailingSlash ensures the given string has a trailing slash.
 func ensureTrailingSlash(s string) string {
 	s = strings.TrimSpace(s)
--- a/command/base_helpers.go
+++ b/command/base_helpers.go
@@ -38,7 +38,6 @@ func sanitizePath(s string) string {
 	return ensureNoTrailingSlash(ensureNoLeadingSlash(s))
 }
 
-
 // hasTrailingSpace checks to see if there is trailing space
 func hasTrailingSpace(s string) bool {
 	x := strings.TrimSuffix(s, " ")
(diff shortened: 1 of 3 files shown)


command/base_helpers.go
```diff
@@ -38,6 +38,26 @@ func sanitizePath(s string) string {
 	return ensureNoTrailingSlash(ensureNoLeadingSlash(s))
 }
 
+// hasTrailingSpace checks to see if there is trailing space
+func hasTrailingSpace(s string) bool {
+	x := strings.TrimSuffix(s, " ")
+	if x != s {
+		return true
+	}
+	return false
```
averche, line 47:
```suggestion
	return x != s
```

Nit: this expression can be simplified

command/base_helpers.go
```diff
@@ -38,6 +38,26 @@ func sanitizePath(s string) string {
...
+}
+
+// hasTrailingSlash boolean checks if a trailing slash exists
+func hasTrailingSlash(s string) bool {
+	s = strings.TrimSpace(s)
+
+	if s[len(s)-1] == '/' {
```
averche, line 54: I think `strings.HasSuffix` might be a better choice here, same for `hasTrailingSpace`

averche, line 54: I think `strings.HasSuffix` might be a better choice here, same for `hasTrailingSpace`

command/kv_get.go
```diff
@@ -124,7 +124,13 @@ func (c *KVGetCommand) Run(args []string) int {
 	// Parse the paths and grab the KV version
 	if mountFlagSyntax {
 		// In this case, this arg is the secret path (e.g. "foo").
+		origPath := args[0]
 		partialPath = sanitizePath(args[0])
+
+		if hasTrailingSpace(origPath) && !hasTrailingSlash(origPath) {
+			partialPath = partialPath + " "
```
averche, line 131: Will this work if there are multiple trailing spaces? Maybe a safer option would be to not fully sanitize the path in case of the `kv get`?

averche, line 131: Will this work if there are multiple trailing spaces? Maybe a safer option would be to not fully sanitize the path in case of the `kv get`?

AnPucel: If you have something like "secret/path____" it will treat it as a single trailing space. Not sure if that's the right call

command/base_helpers.go
AnPucel: yes! Thank you

command/kv_get.go
AnPucel: I'll add all of the edge cases to the description

PR #15204, opened 2022-04-27
agent/auto-auth: Add `min_backoff` to set first backoff value

jasonodonnell:
This adds a new config to Agent's auto-auth.method's stanza: `min_backoff`.

Agent's auto-auth has an initial hardcoded retry duration of 1 second, which grows after each failed retry. I'm making this value configurable because there may be instances you want to retry faster/slower. While writing this feature I found that the SDK client used by auto-auth has it's own built in retry logic. I disabled that retry logic (set MaxRetries to 0) because the auto-auth code has its own backoff function and we were effectively doubling our retry attempts.

Templating also relies on auto-auth and has it's own retry logic. I found an instance where Vault failed to get a token but the templating server tried to fetch secrets. This resulted in yet another variation of retries and more requests to Vault. To combat this, I synced the Template server's initial backoff value to that of auto-auths.

Example config:

```
auto_auth = {
  method = {
    config = {
      role_id_file_path = "/tmp/agent/roleid"
      secret_id_file_path = "/tmp/agent/secretid"
    }

    type = "approle"
    min_backoff = "5s"
    max_backoff = "30s"
  }

  sink = {
    config = {
      path = "/tmp/file-foo"
    }
    type = "file"
  }
}

exit_after_auth = false

template = {
  contents = "{{ with secret \"secret/hashiconf\" }}{{ .Data | toJSONPretty }}{{ end }}"
  destination = "/tmp/agent/kv"
}

vault = {
  address = "http://127.0.0.1:8200"
}
```

### DIFF
--- a/command/agent/auth/auth.go
+++ b/command/agent/auth/auth.go
@@ -55,7 +55,7 @@ type AuthHandler struct {
 	random                       *rand.Rand
 	wrapTTL                      time.Duration
 	maxBackoff                   time.Duration
-	initialBackoff               time.Duration
+	minBackoff                   time.Duration
 	enableReauthOnNewCredentials bool
 	enableTemplateTokenCh        bool
 }
@@ -65,14 +65,14 @@ type AuthHandlerConfig struct {
 	Client                       *api.Client
 	WrapTTL                      time.Duration
 	MaxBackoff                   time.Duration
-	InitialBackoff               time.Duration
+	MinBackoff                   time.Duration
 	Token                        string
 	EnableReauthOnNewCredentials bool
 	EnableTemplateTokenCh        bool
 }
 
 func NewAuthHandler(conf *AuthHandlerConfig) *AuthHandler {
-	initial := conf.InitialBackoff
+	initial := conf.MinBackoff
 	if initial <= 0 {
 		initial = defaultInitialBackoff
 	}
@@ -87,7 +87,7 @@ func NewAuthHandler(conf *AuthHandlerConfig) *AuthHandler {
 		client:                       conf.Client,
 		random:                       rand.New(rand.NewSource(int64(time.Now().Nanosecond()))),
 		wrapTTL:                      conf.WrapTTL,
-		initialBackoff:               initial,
+		minBackoff:                   initial,
 		maxBackoff:                   conf.MaxBackoff,
 		enableReauthOnNewCredentials: conf.EnableReauthOnNewCredentials,
 		enableTemplateTokenCh:        conf.EnableTemplateTokenCh,
@@ -112,11 +112,11 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
 		return errors.New("auth handler: nil auth method")
 	}
 
-	if ah.initialBackoff <= 0 {
-		ah.initialBackoff = defaultInitialBackoff
+	if ah.minBackoff <= 0 {
+		ah.minBackoff = defaultInitialBackoff
 	}
 
-	backoff := newAgentBackoff(ah.initialBackoff, ah.maxBackoff)
+	backoff := newAgentBackoff(ah.minBackoff, ah.maxBackoff)
 
 	if backoff.current >= backoff.max {
 		return errors.New("auth handler: initial backoff cannot be greater than max backoff")
@@ -278,7 +278,7 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
 			}
 
 			am.CredSuccess()
-			backoff.reset(ah.initialBackoff)
+			backoff.reset(ah.minBackoff)
 
 			select {
 			case <-ctx.Done():
@@ -310,7 +310,7 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
 			}
 
 			am.CredSuccess()
-			backoff.reset(ah.initialBackoff)
+			backoff.reset(ah.minBackoff)
 		}
 
 		if watcher != nil {
--- a/command/agent/auth/auth.go
+++ b/command/agent/auth/auth.go
@@ -15,8 +15,8 @@ import (
 )
 
 const (
-	initialBackoff    = 1 * time.Second
-	defaultMaxBackoff = 5 * time.Minute
+	defaultInitialBackoff = 1 * time.Second
+	defaultMaxBackoff     = 5 * time.Minute
 )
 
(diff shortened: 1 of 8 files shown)


command/agent/auth/auth.go
```diff
@@ -80,6 +87,7 @@ func NewAuthHandler(conf *AuthHandlerConfig) *AuthHandler {
 		client:                       conf.Client,
 		random:                       rand.New(rand.NewSource(int64(time.Now().Nanosecond()))),
 		wrapTTL:                      conf.WrapTTL,
+		minBackoff:                   initial,
```
calvn, line 90: I think that it's okay to pass `conf.MinBackoff` here, similar to since we're doing the same check in `newAgentBackoff` which is where the value gets used. This is similar how we treat `conf.MaxBackoff`. We can skip the other checks `if initial <= 0 {}` that we have elsewhere.

command/agent/auth/auth.go
```diff
@@ -371,7 +395,7 @@ func (b *agentBackoff) next() {
 	b.current = maxBackoff - time.Duration(trim)
 }
 
-func (b *agentBackoff) reset() {
+func (b *agentBackoff) reset(initialBackoff time.Duration) {
 	b.current = initialBackoff
```
calvn, line 399: If you add (and set) an `initial` private field on `agentBackoff`, reset can remain without args and `b.current` can be set to `b.initial` like so:

```suggestion
func (b *agentBackoff) reset() {
	b.current = b.initial
```

command/agent/auth/auth.go
```diff
@@ -104,7 +112,15 @@ func (ah *AuthHandler) Run(ctx context.Context, am AuthMethod) error {
...
 	}
 
-	backoff := newAgentBackoff(ah.maxBackoff)
+	if ah.minBackoff <= 0 {
+		ah.minBackoff = defaultInitialBackoff
+	}
+
+	backoff := newAgentBackoff(ah.minBackoff, ah.maxBackoff)
+
+	if backoff.current >= backoff.max {
```
calvn, line 121: If we add `initial`, the check could be against that. We could also cap the max to initial instead of returning an error, but I'll leave it up to you on how we what to handle that.

command/agent/template/template.go
```diff
@@ -311,6 +311,15 @@ func newRunnerConfig(sc *ServerConfig, templates ctconfig.TemplateConfigs) (*ctc
 		Enabled:  &enabled,
 	}
 
+	// Sync Consul Template's retry with user set auto-auth initial backoff value.
+	// This is helpful if Auto Auth cannot get a new token and CT is trying to fetch
+	// secrets.
+	if sc.AgentConfig.AutoAuth != nil && sc.AgentConfig.AutoAuth.Method != nil {
+		if sc.AgentConfig.AutoAuth.Method.MinBackoff > 0 {
+			conf.Vault.Retry.Backoff = &sc.AgentConfig.AutoAuth.Method.MinBackoff
```
calvn, line 319: If we decide to go with this, can we add a note in the docs that `min_backoff` also affects backoff behavior for templating retries?

website/content/docs/agent/autoauth/index.mdx
```diff
@@ -138,8 +138,13 @@ These are common configuration values that live within the `method` block:
   structure. Values can be an integer number of seconds or a stringish value
   like `5m`.
 
+- `min_backoff` `(string or integer: "1s")` - The initial backoff time Agent
```
calvn, line 141: Should be parameter be `min_backoff` or `initial_backoff`? I see mixed references between min and initial even though they refer to the same thing.

calvn, line 141: Should be parameter be `min_backoff` or `initial_backoff`? I see mixed references between min and initial even though they refer to the same thing.

command/agent/template/template.go
calvn: If we're push the `min_backoff` to CT, I think that it makes sense to push the `max_backoff` as well. Otherwise CT could get into a state where the `max_backoff` is greater than the min_backoff value.

Note that pushing either value to templating changes the defaults though since they're different:

| Parameter    | Default auto-auth | Default templating
| ----------- | ----------- | ----------- |
| min_backoff      | 1s       | 250ms | 
| max_backoff   | 5min        | 1min |

Related to this (though it can be tackled separately), I wonder if `min_backoff` and `max_backoff` should live outside of the `auto_auth` block if this is the behavior.

website/content/docs/agent/autoauth/index.mdx
jasonodonnell: I changed it to `min_backoff` in https://github.com/hashicorp/vault/pull/15204/commits/baba936a9c75f8ef0454dca7878a8dd4594c64e1 (it was `initial_backoff`) because it aligned better with the existing parameter `max_backoff`.

command/agent/template/template.go
jasonodonnell: Sure, no problem.

command/agent/auth/auth.go
jasonodonnell: Good call out, I will update this on my end.

website/content/docs/agent/autoauth/index.mdx
jasonodonnell: I removed all initial references.

command/agent/auth/auth.go
jasonodonnell: Done!

website/content/docs/agent/autoauth/index.mdx
```diff
@@ -138,8 +138,14 @@ These are common configuration values that live within the `method` block:
...
+- `min_backoff` `(string or integer: "1s")` - The minimum backoff time Agent 
+  will delay before retrying after a failed auth attempt. The backoff will start 
+  at the configured value and double (with some randomness) after successive 
+  failures, capped by `max_backoff.` If Agent templating is being used, this 
+  value is also used as the min backoff time for the templating server.
+
 - `max_backoff` `(string or integer: "5m")` - The maximum time Agent will delay
-  before retrying after a failed auth attempt. The backoff will start at 1 second
+  before retrying after a failed auth attempt. The backoff will start at `min_backoff` 
   and double (with some randomness) after successive failures, capped by `max_backoff.`
```
calvn, line 149: We should also add the note about templating server in here.

calvn (approved):
One last doc comment, otherwise looks like it's good to go!

PR #15213, opened 2022-04-28
Allow callers to choose the entropy source for the random endpoints.

sgmiller:
They can choose "platform" to get the status quo behavior, "seal" to get
entropy augmentation, or "all" to get an XORed combination of both.

Addresses VAULT-5850.

### DIFF
--- a/builtin/logical/transit/path_random.go
+++ b/builtin/logical/transit/path_random.go
@@ -44,8 +44,8 @@ func (b *backend) pathRandom() *framework.Path {
 	}
 }
 
-func (b *backend) pathRandomWrite(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
-	return random.HandleRandomAPI(ctx, req, d, b.GetRandomReader())
+func (b *backend) pathRandomWrite(_ context.Context, _ *logical.Request, d *framework.FieldData) (*logical.Response, error) {
+	return random.HandleRandomAPI(d, b.GetRandomReader())
 }
 
 const pathRandomHelpSyn = `Generate random bytes`
--- a/builtin/logical/transit/path_random.go
+++ b/builtin/logical/transit/path_random.go
@@ -2,21 +2,10 @@ package transit
 
 import (
 	"context"
-	"crypto/rand"
-	"encoding/base64"
-	"encoding/hex"
-	"fmt"
-	"github.com/hashicorp/vault/sdk/helper/xor"
-	"io"
-	"k8s.io/utils/strings/slices"
-	"strconv"
-
-	uuid "github.com/hashicorp/go-uuid"
+	"github.com/hashicorp/vault/helper/random"
 	"github.com/hashicorp/vault/sdk/framework"
 	"github.com/hashicorp/vault/sdk/logical"
 )
-
-const maxBytes = 128 * 1024
 
 func (b *backend) pathRandom() *framework.Path {
 	return &framework.Path{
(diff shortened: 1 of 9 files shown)


lucymhdavies:
Interesting. I hadn't considered the possibility that callers would want to specify what entropy source to use.
I cannot think of a use-case where a user would explicitly want to use a lower entropy source of randomness, but I also cannot rule out people needing to do that for some reason.

What's the thinking behind having `platform` as the default source? Thinking in terms of secure defaults, if `seal` or `all` was the default source, then we would default to higher entropy where available, with a fallback to `platform` where that's not available.

Also not entirely sure what the use-case would be for `all`. I'm not a crypto expert, so I don't know whether the final random bytes you'd end up with from `platform XOR seal` would be higher entropy than just using `seal` directly.

builtin/logical/transit/path_random.go
```diff
@@ -74,9 +97,39 @@ func (b *backend) pathRandomWrite(ctx context.Context, req *logical.Request, d *
...
+		randBytes, err = uuid.GenerateRandomBytes(bytes)
+		if err != nil {
+			return nil, err
+		}
+	case "seal":
+		if rand.Reader == b.GetRandomReader() {
+			warning = "no seal/entropy augmentation available, using platform entropy source"
+		}
+		randBytes = make([]byte, bytes)
+		_, err = io.ReadFull(b.GetRandomReader(), randBytes)
```
cipherboy, line 113: There is `uuid.GenerateRandomBytesWithReader`, which handles this for you if you want.

website/content/api-docs/secret/transit.mdx
```diff
@@ -664,6 +664,11 @@ This endpoint returns high-quality random bytes of the specified length.
 - `format` `(string: "base64")` – Specifies the output encoding. Valid options
   are `hex` or `base64`.
 
+- `source` `(string: "platform")` - Specifies the source of the requested entropy.
+  `platform`, the default, sources entropy from the platform's entropy source.
```
cipherboy, line 668: My understanding, and correct me if I'm wrong here, but there's a subtle distinction here between _random data_ (i.e., output of a RNG construction) and _entropy_ (which is slower, used to seed the RNG, and arguably of higher quality (e.g,. thermal noise measurements) -- but of limited size). While `/dev/urandom` is frequently used to seed an RNG (chaining), it internally has a RNG and should be considered the former, not the latter.

My  2c., but since we've so far (IMO correctly) called it "random bytes" or wording to that effect, I don't think we should give the impression that _we're_ maintaining an internal RNG construction of our own...  (that we're seeding from an external entropy source). Instead, we should clarify we simply ferry it from another RNG

(I guess you could argue `all` creates its own RNG, albeit with a trivial construction :D).

cipherboy (approved):
Two nits otherwise looks good. :-) 

I do think it gets a little messy with both "source detection" and "multi-path reading", the option of using `GenerateRandomBytesWithReader` might help clean that up.

lucymhdavies:
> Any external RNG is going to have a hard time beating that

Good point. Still not sure if the quick response from Vault for these requests is enough of a big deal compared to the benefits of sourcing external entropy from an HSM. Quantitative measures of HSM speed would indeed be beneficial on that point.

i.e. I'm still not entirely convinced that `platform` should remain the default simply because it's maintaining status quo; I can't imagine that switching to `seal` as the default would be backwards-incompatible in any meaningful way.
But this is probably just personal preference (or my lack of imagination). I'm happy either way.

website/content/api-docs/secret/transit.mdx
sgmiller: That's quite fair, will change.

cipherboy (approved):
The refactoring looks great, though there are some `make fmt` issues :-)

I'm not sure I understand the `logical_system_test.go:3459` failure though.

sgmiller:
Somehow I lost a commit from main to introduce that test failure.  <shrug>

lucymhdavies:
🎉

PR #15214, opened 2022-04-28
MFA Push Methods Login Update

zofskeez:
Feedback from the mfa login project indicated that it may be confusing to have different workflows for push only mfa requirements versus passcode or multi-method requirements. As originally implemented, push only requirements would result in mfa validation being handled behind the scenes on the main login screen. After 5 seconds a message would be displayed to indicate that the user _may_ need to check their device for a push notification.

To address the feedback and unify the experience, all login attempts with mfa requirements will be transitioned to the mfa form where validation will begin automatically and the user will see a message indicating the method(s) that must be satisfied.

image.

In addition, to accommodate rate limiting errors which were introduced in #14864, the validation error response handling was updated to parse the new message and display the countdown in the same manner as a code that has already been used.

image

### DIFF
--- a/ui/app/components/mfa-form.js
+++ b/ui/app/components/mfa-form.js
@@ -15,9 +15,10 @@ import { numberToWord } from 'vault/helpers/number-to-word';
  * @param {string} clusterId - id of selected cluster
  * @param {object} authData - data from initial auth request -- { mfa_requirement, backend, data }
  * @param {function} onSuccess - fired when passcode passes validation
+ * @param {function} onError - fired for multi-method or non-passcode method validation errors
  */
 
-export const VALIDATION_ERROR =
+export const TOTP_VALIDATION_ERROR =
   'The passcode failed to validate. If you entered the correct passcode, contact your administrator.';
 
 export default class MfaForm extends Component {
@@ -25,6 +26,18 @@ export default class MfaForm extends Component {
 
   @tracked countdown;
   @tracked error;
+  @tracked codeDelayMessage;
+
+  constructor() {
+    super(...arguments);
+    // trigger validation immediately when passcode is not required
+    const passcodeOrSelect = this.constraints.filter((constraint) => {
+      return constraint.methods.length > 1 || constraint.methods.findBy('uses_passcode');
+    });
+    if (!passcodeOrSelect.length) {
+      this.validate.perform();
+    }
+  }
 
   get constraints() {
     return this.args.authData.mfa_requirement.mfa_constraints;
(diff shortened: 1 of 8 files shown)


ui/app/components/mfa-form.js
```diff
@@ -15,16 +15,29 @@ import { numberToWord } from 'vault/helpers/number-to-word';
...
 
   @tracked countdown;
   @tracked error;
+  @tracked codeDelayMessage;
+
+  constructor() {
+    super(...arguments);
+    // trigger validation immediately when passcode is not required
+    const passcodeOrSelect = this.constraints.filter((constraint) => {
+      return constraint.methods.length > 1 || constraint.methods.findBy('uses_passcode');
```
Monkeychip, line 35: is there any case where methods would come back undefined?  Given this is in the constructor, I would want it to bork here.

ui/app/components/mfa-form.js
```diff
@@ -66,19 +79,26 @@ export default class MfaForm extends Component {
...
-        this.newCodeDelay.perform(seconds);
+      const errors = error.errors || [];
+      const codeUsed = errors.find((e) => e.includes('code already used;'));
+      const rateLimit = errors.find((e) => e.includes('maximum TOTP validation attempts'));
+      const delayMessage = codeUsed || rateLimit;
+
+      if (delayMessage) {
+        const reason = codeUsed ? 'This code has already been used' : 'Maximum validation attempts exceeded';
+        this.codeDelayMessage = `${reason}. Please wait until a new code is available.`;
+        this.newCodeDelay.perform(delayMessage);
```
Monkeychip, line 90: great use case for concurrency!

ui/app/controllers/vault/cluster/auth.js
```diff
@@ -81,8 +80,10 @@ export default Controller.extend({
       this.authSuccess(authResponse);
     },
     onMfaErrorDismiss() {
-      this.set('mfaAuthData', null);
-      this.auth.set('mfaErrors', null);
+      this.setProperties({
```
Monkeychip, line 83: I realize this is a non-glimmerized controller, but I'm curious if you have any thoughts on `setProperties`? When recently converting a component to glimmer, I needed to continue to use setProperties, but with the new glimmer format had to import it from ember/object. I see the API has it document in 4.x, but what's the benefit in this case of using it instead of doing something like:

this.mfaAuthData = null,
this.mfaErrors = null,

ui/app/components/mfa-form.js
zofskeez: Good question. This data gets setup in the the Auth service from the backend response to a login request when mfa is required. Since a login enforcement must have at least 1 associated method, `methods` will always return as an array.

Monkeychip (approved):
LFTM, only a couple of questions.

ui/app/controllers/vault/cluster/auth.js
zofskeez: Similar to `set`, `setProperties` will update properties on an object and then trigger a re-render in the case of a classic component. With glimmer components this pattern was replaced with tracked properties. For example, in a classic component:
```js
this.setProperties({ foo: true, bar: false });
```
In Glimmer would become:
```js
@tracked foo;
@tracked bar;
this.foo = true;
this.bar = false;
```
`setProperties` when imported from `@ember/object` was handy when using a POJO in a classic component where some of the properties were bound in the template.
```js
const foo = { bar: true };
setProperties(foo, { bar: false });
```
This would result in the template being re-rendered and the value of `bar` would be updated.

This is my understanding of it and as far as I can tell you shouldn't need `set` or `setProperties` in glimmer components but instead use tracked properties.

Monkeychip: Awesome, thank you for the info! When I finally get around to having that glimmerize component ready for review, I'll show you the use case in which I kept the setProperties. It had to do with a spread of properties.

PR #15216, opened 2022-04-28
Use %q for quoted strings where appropriate

swayne275:
Not a huge value add, but I figured I might as well clean up some of our string formatting.

I didn't touch anywhere that looked like formatting mattered. lmk what you think!


api/client_test.go
```diff
@@ -254,7 +254,7 @@ func TestDefaulRetryPolicy(t *testing.T) {
 				t.Fatalf("expected to retry request: '%t', but actual result was: '%t'", test.expect, retry)
 			}
 			if err != test.expectErr {
-				t.Fatalf("expected error from retry policy: '%s', but actual result was: '%s'", err, test.expectErr)
+				t.Fatalf("expected error from retry policy: %q, but actual result was: %q", err, test.expectErr)
```
calvn, line 257: There's a slight difference between these two in that any quoted strings within those values will be escaped if using `%q` but not with `'%s'`, but if the spirit of the output here is to quote the error message itself and not so much about escaping quoted values within `err` then I think this is fine to do.

https://go.dev/play/p/xZQeqklVxRq

swayne275: oh that's great to know! I'd imagine most/all of these cases were from not knowing/thinking about using the `%q` verb, but if it looks like there was some other intention it shouldn't be changed

ccapurso (approved):
Looks good to me! The point about escaping is an interesting one. I think you're likely correct in that the use of `'%s'` was likely not intentional to prevent escaping though.

ccapurso:
> So what do we think about this, worth updating with current main and merging?

Seems reasonable to me

swayne275:
assuming tests pass i'm planning to merge this today or tomorrow unless there are any objections? @ccapurso @calvn

PR #15217, opened 2022-04-28
Allow looking up mount entries by their backend UUIDs

stevendpclark:
This is to allow the managed key registry (within enterprise) to lookup mount entries through either mount paths as it does now, or by a backend UUID value.


cipherboy (approved):
Looks good based on tests, but I'm admittedly not an expert on this part of the code. Re-requesting review from Scott...

sgmiller (approved):

cipherboy:

PR #15218, opened 2022-04-28
Model for mfa method

arnav28:


ui/app/routes/vault/cluster/access/mfa/index.js
```diff
@@ -0,0 +1,37 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+export default class MfaRoute extends Route {
```
Monkeychip, line 3: NIT: space between import and the export.

ui/app/templates/vault/cluster/access.hbs
```diff
@@ -11,6 +11,13 @@
         </LinkTo>
       </li>
     {{/if}}
+    {{#if (has-permission "access" routeParams="mfa")}}
+      <li>
+        <LinkTo @route="vault.cluster.access.mfa" data-test-link={{true}}>
```
Monkeychip, line 16: maybe a more specific data-test something to indicate it's for MFA?  When debugging tests, or expanding on tests I'm often searching the code for data-test-xx and the more specific the data-test is the easier it is for me to narrow down the specific one I'm looking for.

ui/app/templates/vault/cluster/access/mfa/configure.hbs
```diff
@@ -0,0 +1,34 @@
...
+  <p>
+    <b>Step 1:</b>
+    Set up an MFA configuration using one of the methods; TOTP, Okta, Duo or Pingid.
+  </p>
+  <p>
+    <b>Step 2:</b>
+    Set up an enforcement to map the MFA configuration to your chosen auth method(s).
+  </p>
+</div>
+<div class="box is-fullwidth is-shadowless">
```
Monkeychip, line 32: What is this div for if it's not being used? e.g. there is nothing inside of it.

ui/app/adapters/mfa-method.js
```diff
@@ -0,0 +1,31 @@
...
+    });
+  }
+
+  query() {
+    return this.ajax(this.urlForQuery(), 'GET', {
+      data: {
+        list: true,
+      },
+    }).then((resp) => {
+      return resp;
```
Monkeychip, line 28: do we plan on handling errors inside the route? I know we don't normally handle them in the adapter, so just curious if that's the plan.

ui/app/adapters/mfa-method.js
```diff
@@ -0,0 +1,31 @@
+import ApplicationAdapter from './application';
+
+export default class MfaMethodAdapter extends ApplicationAdapter {
+  namespace = 'v1';
+
+  urlForQuery(methodType) {
+    let baseUrl = this.buildURL() + '/identity/mfa/method';
```
zofskeez, line 7: Ember Data uses the model name among other things when building the url so the path here would likely end up being `v1/mfa-methods/identity/mfa/method`. If you override `pathForType` to return `identity/mfa/method` then you could update this to `const baseUrl = this.buildURL();` and append the method type.

Here's some more information on customizing adapters
https://guides.emberjs.com/v3.28.0/models/customizing-adapters/#toc_path-customization

ui/app/adapters/mfa-method.js
zofskeez, line 28: Unless we plan on doing something with the response here there is no need for the `.then`

ui/app/routes/vault/cluster/access/mfa/index.js
```diff
@@ -0,0 +1,37 @@
...
+  queryParams = {
+    page: {
+      refreshModel: true,
+    },
+  };
+
+  model(params) {
+    return this.store
+      .lazyPaginatedQuery('mfa-method', {
+        responsePath: 'data.keys',
```
zofskeez, line 15: I'm not sure we want to use the `keys` in this case since we have all the data returned under `data.key_info`. Simply updating the `responsePath` might not work though since `key_info` is an object rather than an array. In the query method on the adapter it might be best to mutate the response and loop through the `key_info` objects to turn it into an array.

ui/app/routes/vault/cluster/access/mfa/index.js
```diff
@@ -0,0 +1,37 @@
...
+        page: params.page || 1,
+      })
+      .then((model) => {
+        return model;
+      })
```
zofskeez, line 20: The promise will resolve the response so unless you need to do something with it the `.then` is not needed.

ui/app/routes/vault/cluster/access/mfa/index.js
```diff
@@ -0,0 +1,37 @@
...
+          return [];
+        } else {
+          throw err;
+        }
+      });
+  }
+  afterModel(model) {
+    if (model.get('length') === 0) {
+      this.router.transitionTo('vault.cluster.access.mfa.configure');
+    }
```
zofskeez, line 32: Nice redirect to the landing page here! Just a nit on using `get`. You can just do `model.length`.

ui/app/serializers/mfa-method.js
```diff
@@ -0,0 +1,18 @@
...
+  normalizeItems(payload) {
+    if (payload.data.keys && Array.isArray(payload.data.keys)) {
+      let data = payload.data.keys.map((key) => {
+        let model = payload.data.key_info[key];
+        model.id = key;
+        return model;
+      });
+      return data;
+    }
+    assign(payload, payload.data);
```
zofskeez, line 14: There's a deprecation warning on ember `assign`. You could use `Object.assign` or the spread operator instead.

zofskeez (commented):
Good work! I left a few comments to review.

ui/app/templates/vault/cluster/access/mfa/configure.hbs
arnav28: Was waiting to get the image, update the div.

ui/app/adapters/mfa-method.js
arnav28: I tried using pathForType and buildUrl but for whatever reason it doesn't modify the path. We can revisit this later.

ui/app/routes/vault/cluster/access/mfa/index.js
arnav28: Yup responsePath cannot use key_info. Instead of handling this in adapter, I have updated the serializer to normalize the response based on keys and key_info.

ui/app/templates/vault/cluster/access.hbs
arnav28: Good point. I will revisit this as part of tests. Currently tests are using this nav collection based on data-test-link attr.

PR #15219, opened 2022-04-28
Add enterprise sudo paths to api.SudoPaths map

digivava:
The previous list of SudoPaths did not contain the sudo paths from the enterprise version of Vault. 

This PR adds those paths, and also changes the test to only check that the api.SudoPaths map is missing a path from the OpenAPI spec but not the other way around, since the OpenAPI spec for OSS will not show these paths.


VinnyHC (approved):
This works for me! Only tiny nit would be possibly putting a comment on the handful of enterprise sudo paths.

ccapurso:
> Only tiny nit would be possibly putting a comment on the handful of enterprise sudo paths.

I'm going to merge this but went ahead and added a comment to denote this first.

Issue #15210, opened 2022-04-28
Upgrading cluster to v1.9.4 fails


fransguelinckx:
**Describe the bug**
When trying to upgrade a 3-node vault cluster from v1.8.9 to v1.9.4 by replacing one of the old nodes, the new node encounters an error preventing it from starting up.

**To Reproduce**
1. Start from v1.8.9 cluster with 3 nodes.
2. Terminate one of the follower nodes and replace it with a new node running v 1.9.4.

**Expected behavior**
The v1.9.4 node joins the cluster.

**Environment:**
* Vault Server Version: v1.9.4
* Server Operating System/Architecture: EC2 instances running ubuntu linux 5.4.0-1071-aws.

Vault server configuration file(s):
```hcl
ui = true

log_level = "trace"

storage "raft" {
  path = "/ebs1/vault"
  node_id = "vault-bb28de72914b59dfb.development.us-east-1.mycorp.com"
  retry_join {
    auto_join = "provider=aws region=us-east-1 tag_key=VaultCluster tag_value=vault-server-development-us-east-1"
    leader_tls_servername = "*.development.us-east-1.mycorp.com"
    leader_client_cert_file = "/opt/vault/tls/vault.crt"
    leader_client_key_file = "/opt/vault/tls/vault.key"
  }
}

listener "tcp" {
  address = "0.0.0.0:8200"
  tls_min_version = "tls12"
  tls_cert_file = "/opt/vault/tls/vault.crt"
  tls_key_file = "/opt/vault/tls/vault.key"
  proxy_protocol_behavior = "use_always"
  proxy_protocol_authorized_addrs = "0.0.0.0:8200"

  telemetry {
    unauthenticated_metrics_access = true
  }
}

# These values come from /usr/local/sbin/setup-vault-server-config.sh
api_addr = "https://vault-bb28de72914b59dfb.development.us-east-1.mycorp.com:8200"
cluster_addr = "https://vault-bb28de72914b59dfb.development.us-east-1.mycorp.com:8201"

disable_mlock = true

telemetry {
  disable_hostname = true
  prometheus_retention_time = "744h"
}

seal "awskms" {
  kms_key_id = "f1d7b410-e984-4028-a30c-7c4993d0f3e2"
}
```

Error log:
```text
Apr 28 06:17:38 vault-bb28de72914b59dfb systemd[1]: Started "HashiCorp Vault - A tool for managing secrets".
[...]

PR #15224, opened 2022-04-29
Fix/gh 15210/sockaddr template parsing when needed

peteski22:
Proposed fix for GitHub issue: https://github.com/hashicorp/vault/issues/15210

Before we attempt to make calls to parse any templates, we check to see if there are even templates in the config. If not, just carry on with what we have.


internalshared/configutil/listener.go
```diff
@@ -410,7 +411,16 @@ func ParseListeners(result *SharedConfig, list *ast.ObjectList) error {
 
 // ParseSingleIPTemplate is used as a helper function to parse out a single IP
 // address from a config parameter.
+// If the input doesn't appear to contain the 'template' format,
+// it will return the specified input unchanged.
 func ParseSingleIPTemplate(ipTmpl string) (string, error) {
+	m, err := regexp.MatchString("(.*?)({{.*?}})(.*)", ipTmpl)
```
ncabatoff, line 417:
```suggestion
	m, err := regexp.MatchString("{{.*?}}", ipTmpl)
```

We just need to know that somewhere in the input there's a `{{` followed eventually by a `}}`, right?

ncabatoff: Also check out regexp.MustCompile.  Any chance you get to turn a run-time error into a compile-time error is usually worth taking.

internalshared/configutil/listener_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseSingleIPTemplate(t *testing.T) {
+	type args struct {
+		ipTmpl string
+	}
+	tests := []struct {
+		name    string
+		args    args
```
ncabatoff, line 16: Since there's only one argument, is it worth defining this struct?  You could replace this field with `input string` and make it a little simpler.

internalshared/configutil/listener_test.go
```diff
@@ -0,0 +1,49 @@
...
+		want    string
+		wantErr assert.ErrorAssertionFunc
+	}{
+		{
+			name:    "test https addr",
+			args:    args{"https://vaultproject.io:8200"},
+			want:    "https://vaultproject.io:8200",
+			wantErr: assert.NoError,
```
ncabatoff, line 24: Neat!  Hadn't seen this testify trick before.

ncabatoff (approved):

PR #15227, opened 2022-04-29
Handle issuer, keys locking

cipherboy:
We need a write lock around writes to ensure serialization of
modifications. We use a single lock for both issuer and key
updates, in part because certain operations (like deletion) will
potentially affect both.

---

Also updated to make sure we grab the issuers lock.

### DIFF
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -185,6 +185,9 @@ type backend struct {
 
 	pkiStorageVersion atomic.Value
 	crlBuilder        *crlBuilder
+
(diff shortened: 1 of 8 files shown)


cipherboy:
Most of those are updated now :-) The CRL shouldn't need it as its read-only and can happen on any node (this lock only gets grabbed on primary active).

stevendpclark (approved):
At least as an initial stab this lgtm! I would definitely like to see an approval from @kitography as well on this one though before merging.

cipherboy:
Rebased to fix conflicts.

cipherboy:

PR #15244, opened 2022-04-29
MFA Login Enforcement Model

zofskeez:
This PR adds a model, adapter and serializer for mfa login enforcement.

### DIFF
--- a/ui/app/adapters/mfa-login-enforcement.js
+++ b/ui/app/adapters/mfa-login-enforcement.js
@@ -1,0 +1,14 @@
+import ApplicationAdapter from '../application';
+
+export default class KeymgmtKeyAdapter extends ApplicationAdapter {
+  namespace = 'v1';
+
+  pathForType() {
+    return 'identity/mfa/login-enforcement';
+  }
+
+  async query(store, type, query) {
+    const url = this.buildURL(type.modelName, null, null, 'query', query);
+    return this.ajax(url, 'GET', { data: { list: true } }).then((resp) => resp.data);
+  }
+}
--- a/ui/app/adapters/mfa-login-enforcement.js
+++ b/ui/app/adapters/mfa-login-enforcement.js
@@ -8,7 +8,7 @@ export default class KeymgmtKeyAdapter extends ApplicationAdapter {
   }
 
   async query(store, type, query) {
-    const url = this.buildURL(type.modelName, null, null, 'query', query);
+    const url = this.urlForQuery(query, type.modelName);
     return this.ajax(url, 'GET', { data: { list: true } }).then((resp) => resp.data);
   }
 }
--- a/ui/app/adapters/mfa-login-enforcement.js
+++ b/ui/app/adapters/mfa-login-enforcement.js
@@ -1,4 +1,4 @@
-import ApplicationAdapter from '../application';
+import ApplicationAdapter from './application';
 
 export default class KeymgmtKeyAdapter extends ApplicationAdapter {
   namespace = 'v1';
--- a/ui/app/models/mfa-login-enforcement.js
+++ b/ui/app/models/mfa-login-enforcement.js
@@ -1,0 +1,52 @@
+import Model, { attr, hasMany } from '@ember-data/model';
+import { tracked } from '@glimmer/tracking';
+import { expandAttributeMeta } from 'vault/utils/field-to-attrs';
+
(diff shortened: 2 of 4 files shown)


ui/app/adapters/mfa-login-enforcement.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default class KeymgmtKeyAdapter extends ApplicationAdapter {
+  namespace = 'v1';
+
+  pathForType() {
+    return 'identity/mfa/login-enforcement';
+  }
+
+  async query(store, type, query) {
+    const url = this.buildURL(type.modelName, null, null, 'query', query);
```
Monkeychip, line 11: curious why null, null here. Maybe give example in a comment what this URL looks like. Normally it's not necessary, but this one is hard to read.

ui/app/adapters/mfa-login-enforcement.js
```diff
@@ -0,0 +1,14 @@
+import ApplicationAdapter from './application';
+
+export default class KeymgmtKeyAdapter extends ApplicationAdapter {
+  namespace = 'v1';
+
+  pathForType() {
```
Monkeychip, line 6: From what I understand of pathForType it's when we want to interrupt the conventional URL. Example in the docs is for underscore situations. Why are we using it here? https://guides.emberjs.com/release/models/customizing-adapters/

ui/app/models/mfa-login-enforcement.js
```diff
@@ -0,0 +1,28 @@
+import Model, { attr, hasMany } from '@ember-data/model';
+import { expandAttributeMeta } from 'vault/utils/field-to-attrs';
+
+export default class mfaLoginEnforcementModel extends Model {
+  @attr('string', {
+    label: 'Enforcement name',
+    subText:
+      'The name for this enforcement. Giving it a name means that you can refer to it again later. This name will not be editable later.',
```
Monkeychip, line 8: More of curious question: are there any characters that the enforcement cannot have e.g. commas, spaces, *, etc. If so maybe later we can add validation.

ui/app/models/mfa-login-enforcement.js
```diff
@@ -0,0 +1,28 @@
...
+  })
+  name;
+  @hasMany('mfa-method', {
```
Monkeychip, line 11: nice!

ui/tests/unit/serializers/mfa-login-enforcement-test.js
```diff
@@ -0,0 +1,33 @@
+import { module, test } from 'qunit';
```
Monkeychip, line 1: thanks for adding a serializer test. we don't do these very often, but it's a great call in this scenario.

Monkeychip (approved):
Just a couple of questions, but nothing blocking.

ui/app/adapters/mfa-login-enforcement.js
zofskeez: The second argument to `buildURL` is `id` and the third is `snapshot`. When doing a query both are not available which is why I'm explicitly passing null. I think for this particular adapter, rather than using `buildURL` we could use `urlForQuery` to achieve the same thing and it would read more clearly.

ui/app/adapters/mfa-login-enforcement.js
zofskeez: Ember data will attempt to build the url automatically based on the model name which in this case would result in `/v1/mfa-login-enforcements` (pluralizes and dasherizes the name by default) https://guides.emberjs.com/release/models/customizing-adapters/#toc_path-customization.

ui/app/models/mfa-login-enforcement.js
zofskeez: That's a good question. If there were I would expect the API to return an error so that's something we should test once the form is built and then we can add validation on our end.

ui/tests/unit/serializers/mfa-login-enforcement-test.js
zofskeez: Thanks! I wanted to be sure that the keys were transformed as expected since I'm not at a place to actually test with a model until the method model work is merged.

ui/app/adapters/mfa-login-enforcement.js
Monkeychip: got it, thank you!

PR #15248, opened 2022-04-30
Globally scoped MFA method Get/List endpoints

hghaf099:


vault/login_mfa.go
```diff
@@ -199,6 +228,11 @@ func (i *IdentityStore) handleMFAMethodRead(ctx context.Context, req *logical.Re
 	if !(ns.ID == mfaNs.ID || mfaNs.HasParent(ns) || ns.HasParent(mfaNs)) {
 		return logical.ErrorResponse("request namespace does not match method namespace"), logical.ErrPermissionDenied
 	}
+
+	if methodType != "" && respData["type"] != methodType {
+		return logical.ErrorResponse("failed to find the method ID under MFA type %s. Method ID belongs to %s MFA type", methodType, respData["type"]), nil
+	}
```
ccapurso, line 234: I'm not sure we want to expose the actual MFA type here.

ccapurso (approved):
I provided one small comment about potentially exposing unwanted information in an error string but otherwise looks good!

vault/login_mfa.go
hghaf099: So, there is no sensitive information in the MFA configs, and MFAs could be actually configured by an operator on a top level namespace and be used everywhere. In addition, there is a check right above this which make sure the namespace rules are met during reading an MFA config. So, I believe this could help the user to figure out how to resolve the issue. Though I don't have a strong feeling on keeping this as a simple list command would also reveal this information. I will remove it.

Issue #13608, opened 2022-01-10
SSH: add not_before_duration feature, like what pki has


dhbaird:
**Is your feature request related to a problem? Please describe.**
Would like to backdate SSH certificates as is already possible in PKI, see:
* https://github.com/hashicorp/vault/issues/4639 PKI: new certificates valid ahead of time
* https://github.com/hashicorp/vault/pull/5325 Add ability to configure the NotBefore property of certificates in role api

**Describe the solution you'd like**
See same issue as above.

**Describe alternatives you've considered**
See same issue as above.

**Explain any additional use-cases**
See same issue as above.

**Additional context**
Would be nice to have an option to specify exact valid date/time in the past or future, as well as relative, but in a pinch just relative will dates in the past work (as in issue above).

Thanks!

cipherboy:
Hey @dhbaird -- this makes to me. If you happen to get around to it before we do (likely will be a bit), happy to review a PR for it.

Do note that we've recently modified the API structure of the PKI fields (for `NotAfter` at any rate) to allow both a numeric duration and an absolute timestamp... For _eventual_ consistency, it might be nice support both? My 2c. :-)

Gabrielopesantos:
Hello, I have never used vault but seems seems an interesting issue. Can I take it? Not sure If I will be able to produce any kind of solution but I will try. Thanks

cipherboy:
@Gabrielopesantos Sure! There's a similar PR on the PKI side right now: https://github.com/hashicorp/vault/pull/14178

You could likely model the SSH changes (design wise) off of that. :-) Let us know if you have any questions.

Gabrielopesantos:
Hello, I already made some changes based on the PR you linked but I am now having a few problems.

Before trying to implement the support of both numeric duration and absolute timestamp I would like to test if numeric is working as expected, however, I am having some issues running these tests. They take too much time and seem to fail to running out of memory.

cipherboy:
Hi @Gabrielopesantos,

If I understand you correctly, yeah, `not_before_duration` would simply take a numeric duration (in seconds). I don't (personally) see the value here in _both_ a numeric value and an absolute timestamp. Unlike NotAfter (where the Y10K date is a good reason to specify the exact NotAfter value -- because as an offset in seconds, it'd overflow a 64-bit numeric type), I don't really see the value in backdating by that much? 

[...]

PR #15250, opened 2022-04-30
not_before_duration added to SSH

Gabrielopesantos:
# Description
Added a configurable field that allows users to specify the amount of seconds an SSH certificate is backdated.
Issue: #13608

### DIFF
--- a/builtin/logical/ssh/backend_test.go
+++ b/builtin/logical/ssh/backend_test.go
@@ -1682,6 +1682,30 @@ func TestBackend_DefExtTemplatingDisabled(t *testing.T) {
 	}
 }
 
+func TestSSHBackend_ValidateNotBeforeDuration(t *testing.T) {
+	config := logical.TestBackendConfig()
+
+	b, err := Factory(context.Background(), config)
+	if err != nil {
+		t.Fatalf("Cannot create backend: %s", err)
+	}
+	testCase := logicaltest.TestCase{
+		LogicalBackend: b,
+		Steps: []logicaltest.TestStep{
+			configCaStep(testCAPublicKey, testCAPrivateKey),
+
+			createRoleStep("testing", map[string]interface{}{
+				"key_type":                "ca",
+				"allow_user_key_ids":      false,
+				"allow_user_certificates": true,
+				"not_before_duration":     "300s",
+			}),
+		},
+	}
+
+	logicaltest.Test(t, testCase)
+}
+
 func getSshCaTestCluster(t *testing.T, userIdentity string) (*vault.TestCluster, string) {
 	coreConfig := &vault.CoreConfig{
 		CredentialBackends: map[string]logical.Factory{
--- a/builtin/logical/ssh/backend_test.go
+++ b/builtin/logical/ssh/backend_test.go
@@ -1696,10 +1696,80 @@ func TestSSHBackend_ValidateNotBeforeDuration(t *testing.T) {
 
 			createRoleStep("testing", map[string]interface{}{
 				"key_type":                "ca",
-				"allow_user_key_ids":      false,
-				"allow_user_certificates": true,
-				"not_before_duration":     "300s",
+				"allow_host_certificates": true,
+				"allowed_domains":         "example.com,example.org",
+				"allow_subdomains":        true,
+				"default_critical_options": map[string]interface{}{
+					"option": "value",
+				},
+				"default_extensions": map[string]interface{}{
+					"extension": "extended",
+				},
+				"not_before_duration": "300s",
 			}),
+
+			signCertificateStep("testing", "vault-root-22608f5ef173aabf700797cb95c5641e792698ec6380e8e1eb55523e39aa5e51", ssh.HostCert, []string{"dummy.example.org", "second.example.com"}, map[string]string{
+				"option": "value",
+			}, map[string]string{
+				"extension": "extended",
+			},
+				2*time.Hour+5*time.Minute-30*time.Second, map[string]interface{}{
+					"public_key":       publicKey2,
+					"ttl":              "2h",
+					"cert_type":        "host",
+					"valid_principals": "dummy.example.org,second.example.com",
+				}),
+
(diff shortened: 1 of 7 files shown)


builtin/logical/ssh/backend_test.go
```diff
@@ -1682,6 +1682,100 @@ func TestBackend_DefExtTemplatingDisabled(t *testing.T) {
...
+				"option": "value",
+			}, map[string]string{
+				"extension": "extended",
+			},
+				2*time.Hour+5*time.Minute-30*time.Second, map[string]interface{}{
+					"public_key":       publicKey2,
+					"ttl":              "2h",
+					"cert_type":        "host",
+					"valid_principals": "dummy.example.org,second.example.com",
+				}),
```
cipherboy, line 1721: As an aside: wow, I know that passes `make fmt` but that's kinda weird formatting.

builtin/logical/ssh/path_roles.go
```diff
@@ -363,6 +364,17 @@ func pathRoles(b *backend) *framework.Path {
 					Name: "Signing Algorithm",
 				},
 			},
+			"not_before_duration": {
+				Type:    framework.TypeDurationSecond,
+				Default: 30,
+				Description: `
+                The duration before now the SSH cert needs to be signed.
```
cipherboy, line 371: Just for context, if you run `vault path-help ssh/roles` this aught to show up there. I think you'll find that it should probably be on one line, not on two.

cipherboy: Also, as a nit, perhaps `The duration before the issuance time that the SSH certificate should be backdated by.` might be a touch clearer?

cipherboy (commented):
So, my understanding is:

You've add a new parameter to the role, called `not_before_duration` that's a `time.Duration == int64` type. (Good!)

You then use this parameter when setting the `ValidAfter` value. (Good!)

You've added tests & docs & changelog and this is also good. :-)

But there's one small thing missing that I forgot earlier, due to the new role parameter: migration :/

If you have an existing role (under an older Vault version), because the new field will take on the zero value, we'll change behavior from backdating by 30 seconds to only 0 seconds, which might break some deployments. :/

So in net, you'll probably want to:
 - Bump the role version indicator to 3: https://github.com/hashicorp/vault/blob/436e7f38445d0f0976c87906ec4fe4ec3c95a44a/builtin/logical/ssh/path_roles.go#L28-L31
 - Add a new role version check, that if the version is less than three, you'll mark it modified and set the `NotAfterDuration` field to 30 seconds: https://github.com/hashicorp/vault/blob/436e7f38445d0f0976c87906ec4fe4ec3c95a44a/builtin/logical/ssh/path_roles.go#L687

Otherwise, looks great to me!

builtin/logical/ssh/path_roles.go
cipherboy: Or maybe `The duration that the SSH certificate should be backdated by at issuance.` ?

builtin/logical/ssh/backend_test.go
Gabrielopesantos: Not sure how is the formatting weird there but it seems to be in accordance with go fmt how could it be changed.

Gabrielopesantos: Not sure how is the formatting weird there but it seems to be in accordance with go fmt how could it be changed.

Gabrielopesantos: Not sure how is the formatting weird there but it seems to be in accordance with go fmt how could it be changed.

Gabrielopesantos: Not sure how is the formatting weird there but it seems to be in accordance with go fmt so I don't know how could it be improved.

builtin/logical/ssh/path_roles.go
Gabrielopesantos: Makes sense, I took inspiration on the comment of the PR you linked but your descriptions are better.

cipherboy: Ah, no, I like the `notBeforeDuration` as it leaves us open to add a `validAfter` field later with an explicit date if we want that (rather than a relative offset; not really sure why we'd want that in general though, maybe for Y10K certs backdated to 1970).

builtin/logical/ssh/path_roles.go
```diff
@@ -672,6 +684,11 @@ func (b *backend) checkUpgrade(ctx context.Context, s logical.Storage, n string,
 		err = nil
 	}
 
+	if result.Version < 3 {
+		modified = true
+		result.NotBeforeDuration = 30 * time.Second
```
cipherboy, line 689: One last tiny nit, sorry: this needs a `result.Version = 3` so we don't keep overwriting the roles :-)

This is done in the conditional so that all-at-once updates from older versions of Vault work (incrementally applying new versions of the role from the 0-valued ones). If we don't set it here, I think this will run every load of the role, potentially overwriting the specified value. 

```suggestion
		result.NotBeforeDuration = 30 * time.Second
		result.Version = 3
```

builtin/logical/ssh/backend_test.go
cipherboy: Unless I'm misreading this diff, it looks like there's an extra level of indentation. It starts by placing all parameters on the same line, indents for the map. Next map goes to the same indentation level as the function invocation (fine), but then we _indent_ for a non-map parameter, add the next map parameter, and indent that again!

Seems like we should outdent 1716 and line everything up nicely, or expand the function invocation so all parameters are on the same level (indented one like 1716), forcing all map's k=v entries to be on the level of 1717 instead of 1712. 

But like you said, `go fmt` is happy with it and other tests in the same file look like that, so I don't really care :D

builtin/logical/ssh/path_roles.go
cipherboy: Eh, or not :D Still, no point holding up the ACK on that.

cipherboy (approved):

PR #15254, opened 2022-05-02
Fix deletion of issuers, CRLs

cipherboy:
Apparently I had this branch sitting around but haven't proposed it yet? :-D 

We:

 - Delete the legacy CRL entry when we call `DELETE /root`.
 - When deleting an issuer, remove its corresponding CRL entry on next rebuild. This means CRLs will persist (on disk) until next CRL rebuild, even if all issuers using that CRL have been removed. I think this is fine. 
 - Fix bug where chain is incorrect after deletion. :-) Oops.

### DIFF
--- a/builtin/logical/pki/crl_util.go
+++ b/builtin/logical/pki/crl_util.go
@@ -360,6 +360,40 @@ func buildCRLs(ctx context.Context, b *backend, req *logical.Request, forceNew b
 		}
 	}
 
+	// Before persisting our updated CRL config, check to see if we have
+	// any dangling references. If we have any issuers that don't exist,
+	// remove them, remembering their CRLs IDs. If we've completely removed
+	// all issuers pointing to that CRL number, we can remove it from the
+	// number map and from storage.
+	for mapIssuerId := range crlConfig.IssuerIDCRLMap {
+		stillHaveIssuer := false
+		for _, listedIssuerId := range issuers {
+			if mapIssuerId == listedIssuerId {
+				stillHaveIssuer = true
+				break
+			}
+		}
+
+		if !stillHaveIssuer {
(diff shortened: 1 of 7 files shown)


builtin/logical/pki/crl_util.go
```diff
@@ -360,6 +360,40 @@ func buildCRLs(ctx context.Context, b *backend, req *logical.Request, forceNew b
 		}
 	}
 
+	// Before persisting our updated CRL config, check to see if we have
```
stevendpclark, line 363: At the moment we do not call a CRL rebuild on a deleteIssuer API call, that feels wrong but maybe I'm missing something?

stevendpclark (approved):
This looks good, a comment about the deleteIssuer but technically the next revoke/tidy operation I believe will resolve that issue anyways.

builtin/logical/pki/crl_util.go
cipherboy: If you delete the issuer, the lookup (by ref) will fail, so you can't fetch it any more. The deletion of the actual CRL can be lazily done; the remaining pieces will either be the same (no certs were actually revoked, so its equivalent to previous CRLs) or there's presently not-deleted yet unaccessible CRL that nobody really cares about since they can't see it, that'll get removed on next CRL rebuild (if it was the last of its kind).

So I think its fine?

stevendpclark: Yup I figured after I wrote the comment it was something along these lines, no need for additional changes.

One interesting but non-important point is that the secondary clusters now will force the rebuild I believe of the CRL while the primary will have lagged behind :)

cipherboy:
Thanks @stevendpclark! Manually merging...

cipherboy:

PR #15255, opened 2022-05-02
Add issuer usage restrictions

cipherboy:
~Built on top of #15254; will be rebased when that merges.~

This handles adding usage restrictions to issuers, limiting if they can be used for signing certs and CRLs (separately). 

We updated the legacy bundle shim (showing it can be used for both), the issuer update/fetch paths, and mostly enforce the constraint in `fetchCAInfo`. When fetching a cert (not for signing certs or CRLs), we use `false, false` as arguments.

I'm wondering if we want to make this an actual usage vector set? Or if separate flags are better? Seems like its all just about the same amount of work, minus the arguments to `fetchCAInfo`, which is an internal function.

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -2660,7 +2660,7 @@ func TestBackend_SignSelfIssued(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	signingBundle, err := fetchCAInfo(context.Background(), b, &logical.Request{Storage: storage}, defaultRef)
+	signingBundle, err := fetchCAInfo(context.Background(), b, &logical.Request{Storage: storage}, defaultRef, ReadOnlyUsage)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/builtin/logical/pki/cert_util.go
+++ b/builtin/logical/pki/cert_util.go
@@ -80,7 +80,7 @@ func getFormat(data *framework.FieldData) string {
 }
 
 // fetchCAInfo will fetch the CA info, will return an error if no ca info exists.
-func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRef string) (*certutil.CAInfoBundle, error) {
+func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRef string, usage issuerUsage) (*certutil.CAInfoBundle, error) {
 	entry, bundle, err := fetchCertBundle(ctx, b, req.Storage, issuerRef)
 	if err != nil {
 		switch err.(type) {
@@ -91,6 +91,10 @@ func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRe
 		default:
 			return nil, errutil.InternalError{Err: fmt.Sprintf("error fetching CA info: %v", err)}
 		}
+	}
+
+	if err := entry.EnsureUsage(usage); err != nil {
+		return nil, errutil.InternalError{Err: fmt.Sprintf("error while attempting to use issuer %v: %v", issuerRef, err)}
 	}
 
 	if bundle == nil {
--- a/builtin/logical/pki/path_fetch_issuers.go
+++ b/builtin/logical/pki/path_fetch_issuers.go
@@ -91,6 +91,14 @@ the entire validity period. It is suggested to use "truncate" for
 intermediate CAs and "permit" only for root CAs.`,
 		Default: "err",
 	}
+	fields["usage"] = &framework.FieldSchema{
+		Type: framework.TypeCommaStringSlice,
+		Description: `Comma-separated list (or string slice) of usages for
+this issuer; valid values are "read-only", "issuing-certificates", and
+"crl-signing". Multiple values may be specified. Read-only is implicit
+and always set.`,
+		Default: []string{"read-only", "issuing-certificates", "crl-signing"},
+	}
 
 	return &framework.Path{
 		// Returns a JSON entry.
@@ -162,6 +170,7 @@ func (b *backend) pathGetIssuer(ctx context.Context, req *logical.Request, data 
 			"manual_chain":            respManualChain,
 			"ca_chain":                issuer.CAChain,
 			"leaf_not_after_behavior": issuer.LeafNotAfterBehavior,
+			"usage":                   issuer.Usage.Names(),
 		},
 	}, nil
 }
@@ -202,6 +211,10 @@ func (b *backend) pathUpdateIssuer(ctx context.Context, req *logical.Request, da
 		// errs should still be surfaced, however.
 		return logical.ErrorResponse(err.Error()), nil
 	}
+	if err == errIssuerNameInUse && issuer.Name != newName {
+		// When the new name is in use but isn't this name, throw an error.
+		return logical.ErrorResponse(err.Error()), nil
+	}
 
 	newPath := data.Get("manual_chain").([]string)
 	rawLeafBehavior := data.Get("leaf_not_after_behavior").(string)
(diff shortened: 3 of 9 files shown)


builtin/logical/pki/cert_util.go
```diff
@@ -80,7 +80,7 @@ func getFormat(data *framework.FieldData) string {
 }
 
 // fetchCAInfo will fetch the CA info, will return an error if no ca info exists.
-func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRef string) (*certutil.CAInfoBundle, error) {
+func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRef string, issuance bool, crl bool) (*certutil.CAInfoBundle, error) {
```
stevendpclark, line 83: This feels like as we add more usages in the future could just expand to be really easy to pass in the wrong params. 

Could we instead introduce, internally, an iota of usage constants, and passing them them in kinda like the replication [HasState](https://github.com/hashicorp/vault/blob/main/sdk/helper/consts/replication.go#L147) flagging system?

stevendpclark: :) Oops, well I suppose you know my opinion. 

I think it would make things easier if we pushed the bitset all the way to the entry as well, I believe it will make the comparison easier at the sacrifice of making possibly a future patch update a little more complicated (fetch, update and set)

sgmiller: I agree.  Positional arguments of the same type are fragile.

builtin/logical/pki/backend_test.go
```diff
@@ -2660,7 +2660,7 @@ func TestBackend_SignSelfIssued(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	signingBundle, err := fetchCAInfo(context.Background(), b, &logical.Request{Storage: storage}, defaultRef)
+	signingBundle, err := fetchCAInfo(context.Background(), b, &logical.Request{Storage: storage}, defaultRef, false /* issue */, false /* crl */)
```
sgmiller, line 2663: And switching to constants makes the call sites readable.

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -162,6 +179,8 @@ func (b *backend) pathGetIssuer(ctx context.Context, req *logical.Request, data
 			"manual_chain":            respManualChain,
 			"ca_chain":                issuer.CAChain,
 			"leaf_not_after_behavior": issuer.LeafNotAfterBehavior,
+			"use_for_issuance":        issuer.UseForIssuance,
```
sgmiller, line 182: Rather than bool arguments on the API, would a StringSlice of usage types be both more future proof and a little easier to use?  e.g. usages="issuance,crl_signing".  I don't feel strongly though.

cipherboy: Eh, then you've gotta iterate over the usage and order isn't guaranteed so you're stuck doing multiple passes.

I think a map (or a proper set type if Go had one) is a better construct for that, but then you might as well just use bitvectors I guess. :/

cipherboy: Ah, no, I see what you mean, definitely.

cipherboy: I thought Scott was advocating for this also on the storage entry.

cipherboy:
Testing:

```bash
source devvault
vault secrets enable pki
vault write pki/root/generate/internal common_name="root x1" issuer_name=x1
vault write pki/root/generate/internal common_name="root x2" issuer_name=x2

vault write pki/issuer/default issuer_name=x1 usage="read only,issuing certificates"
vault write pki/roles/testing enforce_hostnames=false client_flag=true allow_any_name=true require_cn=false organization=hashicorp key_type=rsa
vault write pki/issue/testing common_name=localhost ttl=1s

vault write pki/issuer/default issuer_name=x1 usage="read only"
vault write pki/issue/testing common_name=localhost ttl=1s # should fail
```

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -91,6 +91,14 @@ the entire validity period. It is suggested to use "truncate" for
 intermediate CAs and "permit" only for root CAs.`,
 		Default: "err",
 	}
+	fields["usage"] = &framework.FieldSchema{
+		Type: framework.TypeCommaStringSlice,
+		Description: `Comma-separated list (or string slice) of usages for
+this issuer; valid values are "read only", "issuing certificates", and
+"CRL signing". Multiple values may be specified. Read-only is implicit
+and always set.`,
+		Default: []string{"issuing certificates", "CRL signing"},
```
stevendpclark, line 100: nit: Not a huge fan with the mixed casing and spaces for values here. What would you think of using the following instead?

- read-only
- issuing-certificates
- crl-signing

cipherboy: Well, CRL is an acronym and calling it "crl" felt weird if I was going to use spaces ;-) 

I'm fine with slugs, but note this gets used as both "identifiers" and "names for error messages" so that's why I went with the space approach (better error messages, worse names). Fewer people will modify the values than read the errors (since it is user-returnable with a bad-ish configuration).

builtin/logical/pki/storage.go
```diff
@@ -67,6 +67,67 @@ func (e keyEntry) isManagedPrivateKey() bool {
...
+	AllIssuerUsages issuerUsage = ReadOnlyUsage | IssuanceUsage | CRLSigningUsage
+)
+
+var namedIssuerUsages = map[string]issuerUsage{
+	"read only":            ReadOnlyUsage,
+	"issuing certificates": IssuanceUsage,
+	"CRL signing":          CRLSigningUsage,
+}
+
+func (i *issuerUsage) ToggleUsage(usage issuerUsage) *issuerUsage {
```
stevendpclark, line 89: It's a bit weird here that we are taking a pointer reference and returning the same pointer ref, it would lead me to think that I'm getting a new copy of the object, but I actually have two references to the same thing and I've modified the original one.

I think you did this to daisy chain the Toggles which we could more or less accomplish with the following api?

```
func (i *issuerUsage) ToggleUsage(usages ...issuerUsage) {
	for _, usage := range usages {
		*i ^= usage
	}
}
```

builtin/logical/pki/path_fetch_issuers.go
stevendpclark: Fair enough, the poor command line users needing to escape the values now ;)

cipherboy: Eh, that's a good point, I'll change them.

builtin/logical/pki/storage.go
cipherboy: Ah that's nicer, I'll do that.

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -91,6 +91,14 @@ the entire validity period. It is suggested to use "truncate" for
 intermediate CAs and "permit" only for root CAs.`,
 		Default: "err",
 	}
+	fields["usage"] = &framework.FieldSchema{
+		Type: framework.TypeCommaStringSlice,
+		Description: `Comma-separated list (or string slice) of usages for
+this issuer; valid values are "read only", "issuing certificates", and
```
stevendpclark, line 97: nit: Documentation is wrong now :)

stevendpclark (approved):
Small nit on docs, but 👍

builtin/logical/pki/path_fetch_issuers.go
cipherboy: Ah, I updated the other place... :/

cipherboy:
Thank you!

PR #15256, opened 2022-05-02
Address PKI to properly support managed keys

stevendpclark:
This is just a draft PR for now so that people can see the various changes that were required to get this to work within managed keys and for others to work off of.

The main massive change was to the importKey behaviour where now we tell it the type of key it really is on the inbound instead of it trying to determine what it is for itself. This was mainly done as within that function it did not really have the context needed to determine if it was a managed key or not.

Another rather large amount of change was to support a managed key within the context of an existing key.

### DIFF
--- a/builtin/logical/pki/key_util.go
+++ b/builtin/logical/pki/key_util.go
@@ -1,0 +1,116 @@
+package pki
+
+import (
+	"context"
+	"crypto"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"github.com/hashicorp/vault/sdk/helper/certutil"
+	"github.com/hashicorp/vault/sdk/helper/errutil"
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+type managedKeyContext struct {
+	ctx        context.Context
+	b          *backend
+	mountPoint string
+}
+
+func newManagedKeyContext(ctx context.Context, b *backend, mountPoint string) managedKeyContext {
+	return managedKeyContext{
+		ctx:        ctx,
+		b:          b,
+		mountPoint: mountPoint,
+	}
+}
+
+func comparePublicKey(ctx managedKeyContext, key *keyEntry, publicKey crypto.PublicKey) (bool, error) {
+	publicKeyForKeyEntry, err := getPublicKey(ctx, key)
+	if err != nil {
+		return false, err
+	}
+
+	return certutil.ComparePublicKeysAndType(publicKeyForKeyEntry, publicKey)
+}
+
+func getPublicKey(mkc managedKeyContext, key *keyEntry) (crypto.PublicKey, error) {
+	if key.PrivateKeyType == certutil.ManagedPrivateKey {
+		keyId, err := extractManagedKeyId([]byte(key.PrivateKey))
+		if err != nil {
+			return nil, err
+		}
+		return getManagedKeyPublicKey(mkc, keyId)
+	}
+
+	signer, _, _, err := getSignerFromKeyEntryBytes(key)
+	if err != nil {
+		return nil, err
+	}
+	return signer.Public(), nil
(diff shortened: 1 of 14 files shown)


builtin/logical/pki/path_manage_keys.go
```diff
@@ -23,15 +23,27 @@ func pathGenerateKey(b *backend) *framework.Path {
...
+				Type: framework.TypeString,
+				Description: `The name of the managed key to use when the exported
+type is kms. When kms type is the key type, this field or managed_key_name
+is required. Ignored for other types.`,
+			},
 		},
 
 		Operations: map[logical.Operation]framework.OperationHandler{
-			logical.CreateOperation: &framework.PathOperation{
+			logical.UpdateOperation: &framework.PathOperation{
```
stevendpclark, line 46: This was a bit subtle, but Create is only for PUT operations, we want to support POST I believe.

builtin/logical/pki/key_util.go
```diff
@@ -0,0 +1,126 @@
...
+
+func newManagedKeyContext(ctx context.Context, b *backend, mountPoint string) managedKeyContext {
+	return managedKeyContext{
+		ctx:        ctx,
+		b:          b,
+		mountPoint: mountPoint,
+	}
+}
+
+func comparePublicKey(ctx managedKeyContext, key *keyEntry, publicKey crypto.PublicKey) (bool, error) {
```
cipherboy, line 29: Maybe it makes sense to add a "fast path" and compare on bytes? Are there any cases where two bytes are _equal_ but their public keys would be unequal? 

That would avoid having to (unnecessarily) parse string->PEM->ASN.1->GoStruct blobs repeatedly. Just a thought.

builtin/logical/pki/storage.go
```diff
@@ -470,18 +494,12 @@ func importIssuer(ctx context.Context, s logical.Storage, certValue string, issu
...
-
-		// Fetch the signer for its Public() value.
-		signer, err := existingKey.GetSigner()
+		existingKey, err := fetchKeyById(ctx.ctx, s, identifier)
 		if err != nil {
 			return nil, false, err
 		}
 
-		equal, err := certutil.ComparePublicKeysAndType(issuerCert.PublicKey, signer.Public())
+		equal, err := comparePublicKey(ctx, existingKey, issuerCert.PublicKey)
```
cipherboy, line 502: Ah I guess this might make a fast-path comparison harder. :/

builtin/logical/pki/key_util.go
stevendpclark: Yeah as you pointed out below not sure if it's worth it at least for this round to add in a fast path.

cipherboy: Nah that's fair.

builtin/logical/pki/storage.go
```diff
@@ -197,25 +193,30 @@ func importKey(ctx context.Context, s logical.Storage, keyValue string, keyName
...
 
 	// Finally, we can write the key to storage.
-	if err := writeKey(ctx, s, result); err != nil {
+	if err := writeKey(mkc.ctx, s, result); err != nil {
+		return nil, false, err
+	}
+
+	// Before we return below, we need to iterate over _all_ issuers and see if
+	// one of them has a missing KeyId link, and if so, point it back to
+	// ourselves. We fetch the list of issuers up front, even when don't need
```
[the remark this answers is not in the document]
stevendpclark: Actually the comment is good I believe but we don't need to re-fetch the public key as we have already done it, will remove that instead.

cipherboy (approved):
Per discussion and demo today, this looks good. Managed key type to be addressed in a future PR.

cipherboy:

PR #15271, opened 2022-05-03
Add Transit BYOK wrapping key endpoint

schultz-is:
# Overview
Transit BYOK outlines a key import process that makes use of a shared public wrapping key. This key is generated lazily on first request to the endpoint and is used for subsequent imports. The key is stored in seal wrap storage, makes use of the `keysutil` policy management framework, and the private key will never leave the Vault cryptographic barrier.

# Testing
- Unit tests are included that cover the endpoint
- The first request to the endpoint should generate a 4096-bit RSA key and return its public component
- Subsequent requests to the endpoint should return the same public key

### DIFF
--- a/builtin/logical/transit/path_wrapping_key.go
+++ b/builtin/logical/transit/path_wrapping_key.go
@@ -5,12 +5,13 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
-	"github.com/hashicorp/vault/sdk/framework"
-	"github.com/hashicorp/vault/sdk/helper/keysutil"
-	"github.com/hashicorp/vault/sdk/logical"
 	"io"
 	"path"
 	"strconv"
+
+	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/helper/keysutil"
+	"github.com/hashicorp/vault/sdk/logical"
 )
 
 const WrappingKeyName = "wrapping-key"
@@ -92,6 +93,8 @@ func generateWrappingKey(ctx context.Context, storage logical.Storage, rand io.R
 	return p, nil
 }
 
-const pathWrappingKeyHelpSyn = "Returns the public key to use for wrapping imported keys"
-const pathWrappingKeyHelpDesc = "This path is used to retrieve the RSA-4096 wrapping key" +
-	"for wrapping keys that are being imported into transit."
+const (
+	pathWrappingKeyHelpSyn  = "Returns the public key to use for wrapping imported keys"
+	pathWrappingKeyHelpDesc = "This path is used to retrieve the RSA-4096 wrapping key" +
+		"for wrapping keys that are being imported into transit."
+)
--- a/builtin/logical/transit/path_wrapping_key.go
+++ b/builtin/logical/transit/path_wrapping_key.go
@@ -6,7 +6,10 @@ import (
 	"encoding/pem"
 	"fmt"
 	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/helper/keysutil"
 	"github.com/hashicorp/vault/sdk/logical"
+	"io"
+	"path"
 	"strconv"
 )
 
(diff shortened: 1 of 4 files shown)


builtin/logical/transit/path_wrapping_key.go
```diff
@@ -0,0 +1,100 @@
...
+		Callbacks: map[logical.Operation]framework.OperationFunc{
+			logical.ReadOperation: b.pathWrappingKeyRead,
+		},
+		HelpSynopsis:    pathWrappingKeyHelpSyn,
+		HelpDescription: pathWrappingKeyHelpDesc,
+	}
+}
+
+func (b *backend) pathWrappingKeyRead(ctx context.Context, req *logical.Request, _ *framework.FieldData) (*logical.Response, error) {
+	p, err := getWrappingKey(ctx, req.Storage)
```
sgmiller, line 31: This could technically race, leading to two keys being generated and returned to two callers.  Other users of the keysutil system use a LockManager, whose GetPolicy gets the instantiated key or creates it if missing, covered by a lock.  You might consider having a LockManager instance for this, even though it'll only be managing the one key.

sgmiller (changes requested):

builtin/logical/transit/path_wrapping_key.go
schultz-is: Updated the storage path for the wrapping key to be `policy/import/wrapping-key` rather than `import/policy/wrapping-key` which allowed use of the existing LockManager instance on the Transit backend. Since the `policy/` prefix is already in seal wrap storage, the extra `import/` prefix was removed. As far as I'm aware this shouldn't cause any name collisions, but if I'm missing a case I'll be more than happy to revisit this.

builtin/logical/transit/path_wrapping_key_test.go
```diff
@@ -0,0 +1,58 @@
...
+		Path:      "wrapping_key",
+	}
+	resp, err := b.HandleRequest(context.Background(), req)
+	if err != nil {
+		t.Fatalf("unexpected request error: %s", err)
+	}
+	if resp == nil || resp.Data == nil || resp.Data["public_key"] == nil {
+		t.Fatal("expected non-nil response")
+	}
+	pubKey := resp.Data["public_key"]
```
stevendpclark, line 40: Would it be worth testing that this is actually an RSA public key by attempting to parse it?

stevendpclark (approved):
Some small suggestions, but overall +1 from me.

builtin/logical/transit/path_wrapping_key_test.go
schultz-is: Good idea. I had done this in the upcoming import tests because I need to actually use the key for wrapping, but it's worth including it here as well.

builtin/logical/transit/path_wrapping_key.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+	if p == nil {
+		return nil, fmt.Errorf("error retrieving wrapping key: returned policy was nil")
+	}
+	if b.System().CachingDisabled() {
+		p.Unlock()
+	}
+
+	rsaPublicKey := p.Keys[strconv.Itoa(p.LatestVersion)]
```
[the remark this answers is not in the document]
schultz-is: Yep, true. Renamed to `wrappingKey` for clarity and correctness.

builtin/logical/transit/path_wrapping_key.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+	}
+
+	return resp, nil
+}
+
+const (
+	pathWrappingKeyHelpSyn  = "Returns the public key to use for wrapping imported keys"
+	pathWrappingKeyHelpDesc = "This path is used to retrieve the RSA-4096 wrapping key" +
+		"for wrapping keys that are being imported into transit."
```
sgmiller, line 80: missing a space due to the two line construction

sgmiller (changes requested):
tiny nit

builtin/logical/transit/path_wrapping_key.go
schultz-is: Good eye, thanks! Should be fixed now.

sgmiller (approved):

PR #15272, opened 2022-05-03
UBI Containerfile - CRT Enablement 

cipherboy:
This takes the recently released CRT tooling for UBI container building and enables OSS Vault to be built with it.

We merge the [UBI containerfile](https://github.com/hashicorp/docker-vault/blame/master/ubi/Dockerfile) into the existing `/Dockerfile` under a separate target (`default`). We make the following small modifications:

 - Provision the `vault` binary similarly to the existing `Dockerfile` from the CRT pipeline.
 - Pull in the UBI-specific entry point helper.
 - Export the `VAULT_DISABLE_FILE_PERMISSIONS_CHECK=1` flag; OpenShift requires separate permissions (notably the `vault` user is added to the `root` group), making these permissions a bit harder. 
 - Build all of this in the CRT pipeline.

A sample build tested by @tvoran on OpenShift 4 is available here: https://github.com/hashicorp/vault/actions/runs/2245939279

An updated run is available here: https://github.com/hashicorp/vault/runs/6289985171?check_suite_focus=true

After downloading the `vault_ubi_linux_amd64_1.11.0-dev1_720494da31f164b4ba5bf9983b0f3f55f2ae708c.docker.redhat.tar` file, run:

> ```
> unzip vault_ubi_linux_amd64_1.11.0-dev1_720494da31f164b4ba5bf9983b0f3f55f2ae708c.docker.redhat.tar.zip
> podman load --input vault_ubi_linux_amd64_1.11.0-dev1_720494da31f164b4ba5bf9983b0f3f55f2ae708c.docker.redhat.tar
> podman run <imported image spec from previous command output>
> ```

Thanks to all who helped with this!

~My intention is to backport this into the 1.10.x release branch for use with Vault Enterprise builds; are there objections or should we go further back to 1.8.x?~ Not making for 1.10.3

Does this need a Changelog entry here? It seems like Vault release changes should perhaps be documented elsewhere, though this shouldn't affect downstream consumers much (if I did the enablement correctly). :-D

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,13 +1,6 @@
 name: build
 
-on:
-  push:
-    # Sequence of patterns matched against refs/heads
-    branches:    
-      # Push events on main branch
-      - main
-  workflow_dispatch:
-    # Allow manual dispatch of CRT to allow triggering it from PR branches.
+on: [ workflow_dispatch, push ]
 
 env:
   PKG_NAME: "vault"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,8 @@ on:
     branches:    
       # Push events on main branch
       - main
+  workflow_dispatch:
+    # Allow manual dispatch of CRT to allow triggering it from PR branches.
 
 env:
   PKG_NAME: "vault"
@@ -257,3 +259,26 @@ jobs:
           tags: |
             docker.io/hashicorp/${{env.repo}}:${{env.version}}
             public.ecr.aws/hashicorp/${{env.repo}}:${{env.version}}
+
+  build-ubi:
+    name: Red Hat UBI ${{ matrix.arch }} build
+    needs:
+      - get-product-version
+      - build-linux
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: ["amd64"]
+    env:
+      repo: ${{github.event.repository.name}}
+      version: ${{needs.get-product-version.outputs.product-version}}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Docker Build (Action)
+        uses: hashicorp/actions-docker-build@v1
+        with:
+          version: ${{env.version}}
+          target: ubi
+          arch: ${{matrix.arch}}
+          zip_artifact_name: ${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_linux_${{ matrix.arch }}.zip
+          redhat_tag: scan.connect.redhat.com/ospid-f0a92725-d8c6-4023-9a87-ba785b94c3fd/${{env.repo}}:${{env.version}}-ubi
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,8 @@ LABEL name="Vault" \
       maintainer="Vault Team <vault@hashicorp.com>" \
       vendor="HashiCorp" \
       version=${PRODUCT_VERSION} \
-      release=${PRODUCT_VERSION} \
+      release=${PRODUCT_REVISION} \
+      revision=${PRODUCT_REVISION} \
       summary="Vault is a tool for securely accessing secrets." \
       description="Vault is a tool for securely accessing secrets. A secret is anything that you want to tightly control access to, such as API keys, passwords, certificates, and more. Vault provides a unified interface to any secret, while providing tight access control and recording a detailed audit log."
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+## DOCKERHUB DOCKERFILE ##
 FROM alpine:3.15 as default
 
 ARG BIN_NAME
@@ -8,8 +9,17 @@ ARG PRODUCT_VERSION
 # TARGETARCH and TARGETOS are set automatically when --platform is provided.
 ARG TARGETOS TARGETARCH
 
-LABEL maintainer="Vault Team <vault@hashicorp.com>"
-LABEL version=${PRODUCT_VERSION}
+# Additional metadata labels used by container registries, platforms
+# and certification scanners.
+LABEL name="Vault" \
+      maintainer="Vault Team <vault@hashicorp.com>" \
+      vendor="HashiCorp" \
+      version=${PRODUCT_VERSION} \
+      release=${PRODUCT_VERSION} \
+      summary="Vault is a tool for securely accessing secrets." \
+      description="Vault is a tool for securely accessing secrets. A secret is anything that you want to tightly control access to, such as API keys, passwords, certificates, and more. Vault provides a unified interface to any secret, while providing tight access control and recording a detailed audit log."
+
+COPY LICENSE /licenses/mozilla.txt
 
 # Set ARGs as ENV so that they can be used in ENTRYPOINT/CMD
 ENV NAME=$NAME
@@ -52,6 +62,89 @@ EXPOSE 8200
 COPY .release/docker/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
 ENTRYPOINT ["docker-entrypoint.sh"]
 
+
+# # By default you'll get a single-node development server that stores everything
+# # in RAM and bootstraps itself. Don't use this configuration for production.
+CMD ["server", "-dev"]
+
+
+## UBI DOCKERFILE ##
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5 as ubi
+
+ARG BIN_NAME
+# NAME and PRODUCT_VERSION are the name of the software in releases.hashicorp.com
+# and the version to download. Example: NAME=vault PRODUCT_VERSION=1.2.3.
+ARG NAME=vault
+ARG PRODUCT_VERSION
+# TARGETARCH and TARGETOS are set automatically when --platform is provided.
+ARG TARGETOS TARGETARCH
+
+# Additional metadata labels used by container registries, platforms
+# and certification scanners.
+LABEL name="Vault" \
+      maintainer="Vault Team <vault@hashicorp.com>" \
+      vendor="HashiCorp" \
+      version=${PRODUCT_VERSION} \
+      release=${PRODUCT_VERSION} \
+      summary="Vault is a tool for securely accessing secrets." \
+      description="Vault is a tool for securely accessing secrets. A secret is anything that you want to tightly control access to, such as API keys, passwords, certificates, and more. Vault provides a unified interface to any secret, while providing tight access control and recording a detailed audit log."
+
+COPY LICENSE /licenses/mozilla.txt
+
+# Set ARGs as ENV so that they can be used in ENTRYPOINT/CMD
+ENV NAME=$NAME
+ENV VERSION=$VERSION
+
+# Set up certificates, our base tools, and Vault. Unlike the other version of
+# this (https://github.com/hashicorp/docker-vault/blob/master/ubi/Dockerfile),
+# we copy in the Vault binary from CRT.
+RUN set -eux; \
+    microdnf install -y ca-certificates gnupg openssl libcap tzdata procps shadow-utils util-linux
+
+# Create a non-root user to run the software.
+RUN groupadd --gid 1000 vault && \
+    adduser --uid 100 --system -g vault vault && \
+    usermod -a -G root vault
+
+# Copy in the new Vault from CRT pipeline, rather than fetching it from our
+# public releases.
+COPY dist/$TARGETOS/$TARGETARCH/$BIN_NAME /bin/
+
+# /vault/logs is made available to use as a location to store audit logs, if
+# desired; /vault/file is made available to use as a location with the file
+# storage backend, if desired; the server will be started with /vault/config as
+# the configuration directory so you can add additional config files in that
+# location.
+RUN mkdir -p /vault/logs && \
+    mkdir -p /vault/file && \
+    mkdir -p /vault/config && \
+    mkdir -p $HOME && \
+    chown -R vault /vault && chown -R vault $HOME && \
+    chgrp -R 0 $HOME && chmod -R g+rwX $HOME && \
+    chgrp -R 0 /vault && chmod -R g+rwX /vault
+
+# Expose the logs directory as a volume since there's potentially long-running
+# state in there
+VOLUME /vault/logs
+
+# Expose the file directory as a volume since there's potentially long-running
+# state in there
+VOLUME /vault/file
+
+# 8200/tcp is the primary interface that applications use to interact with
+# Vault.
+EXPOSE 8200
+
+# The entry point script uses dumb-init as the top-level process to reap any
+# zombie processes created by Vault sub-processes.
+#
+# For production derivatives of this container, you shoud add the IPC_LOCK
+# capability so that Vault can mlock memory.
+COPY .release/docker/ubi-docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+ENTRYPOINT ["docker-entrypoint.sh"]
+
+# Use the Vault user as the default user for starting this container.
+USER vault
 
 # # By default you'll get a single-node development server that stores everything
 # # in RAM and bootstraps itself. Don't use this configuration for production.
--- a/Dockerfile
+++ b/Dockerfile
@@ -140,8 +140,8 @@ EXPOSE 8200
 #
 # For production derivatives of this container, you shoud add the IPC_LOCK
 # capability so that Vault can mlock memory.
-COPY .release/docker/ubi-docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
-ENTRYPOINT ["docker-entrypoint.sh"]
+COPY .release/docker/ubi-docker-entrypoint.sh /usr/local/bin/
+ENTRYPOINT ["ubi-docker-entrypoint.sh"]
 
 # Use the Vault user as the default user for starting this container.
 USER vault
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,10 +72,10 @@ CMD ["server", "-dev"]
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5 as ubi
 
 ARG BIN_NAME
-# NAME and PRODUCT_VERSION are the name of the software in releases.hashicorp.com
-# and the version to download. Example: NAME=vault PRODUCT_VERSION=1.2.3.
-ARG NAME=vault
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -257,3 +257,26 @@ jobs:
...
+
+  build-ubi:
+    name: Red Hat UBI ${{ matrix.arch }} build
+    needs:
+      - get-product-version
+      - build-linux
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: ["amd64"]
```
swenson, line 269: No support for `aarch64`?

mladlow:
I think we're going to change the default file permissions check behavior to make it opt-in, but there's probably no harm in keeping it explicit.

.github/workflows/build.yml
cipherboy: No, according to here, it is amd64 only: https://catalog.redhat.com/software/containers/hashicorp/vault-enterprise/5fda5633ac3db90370a26443?container-tabs=technical-information

swenson: :+1:

Dockerfile
```diff
@@ -53,6 +54,78 @@ COPY .release/docker/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
...
+# TARGETARCH and TARGETOS are set automatically when --platform is provided.
+ARG TARGETOS TARGETARCH
+
+LABEL maintainer="Vault Team <vault@hashicorp.com>"
+LABEL version=${PRODUCT_VERSION}
+
+# Set ARGs as ENV so that they can be used in ENTRYPOINT/CMD
+ENV NAME=$NAME
+ENV VERSION=$VERSION
+
```
tvoran, line 79: This needs to have a license to pass the [redhat certification](https://connect.redhat.com/zones/containers/container-certification-policy-guide) scan: https://github.com/hashicorp/docker-vault/blob/e4a92e52ee3d480e4fb45f26e5a7d1c295ce0d38/ubi/Dockerfile#L16-L17

Dockerfile
```diff
@@ -53,6 +54,78 @@ COPY .release/docker/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
...
+# The entry point script uses dumb-init as the top-level process to reap any
+# zombie processes created by Vault sub-processes.
+#
+# For production derivatives of this container, you shoud add the IPC_LOCK
+# capability so that Vault can mlock memory.
+COPY .release/docker/ubi-docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+ENTRYPOINT ["docker-entrypoint.sh"]
+
+
```
tvoran, line 128: This needs a non-root user directive to pass the [redhat certification](https://connect.redhat.com/zones/containers/container-certification-policy-guide) scan: https://github.com/hashicorp/docker-vault/blob/e4a92e52ee3d480e4fb45f26e5a7d1c295ce0d38/ubi/Dockerfile#L83

.github/workflows/build.yml
```diff
@@ -257,3 +257,26 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Docker Build (Action)
+        uses: hashicorp/actions-docker-build@v1
+        with:
+          version: ${{env.version}}
+          target: ubi
+          arch: ${{matrix.arch}}
+          zip_artifact_name: ${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_linux_${{ matrix.arch }}.zip
+          redhat_tag: scan.connect.redhat.com/ospid-62211e0d8bf2cabc69a39c7d/${{env.repo}}:${{env.version}}-ubi
```
tvoran, line 282: This doesn't look like the correct ospid for OSS Vault.

Dockerfile
```diff
@@ -53,6 +54,78 @@ COPY .release/docker/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
...
+ARG BIN_NAME
+# NAME and PRODUCT_VERSION are the name of the software in releases.hashicorp.com
+# and the version to download. Example: NAME=vault PRODUCT_VERSION=1.2.3.
+ARG NAME=vault
+ARG PRODUCT_VERSION
+# TARGETARCH and TARGETOS are set automatically when --platform is provided.
+ARG TARGETOS TARGETARCH
+
+LABEL maintainer="Vault Team <vault@hashicorp.com>"
+LABEL version=${PRODUCT_VERSION}
```
tvoran, line 74: Any reason we're not using the same labels as previously? https://github.com/hashicorp/docker-vault/blob/e4a92e52ee3d480e4fb45f26e5a7d1c295ce0d38/ubi/Dockerfile#L6-L14

.github/workflows/build.yml
cipherboy: Ah, that's a good point, I found this in the helloworld CRT example.

Dockerfile
cipherboy: Do you know how we should provision this? The LICENSE was grabbed from externally (in the old `docker-vault`) and I was a little hesitant to do that in CRT. Should I just `wget` it here too?

Dockerfile
cipherboy: I pulled them from the Alpine based image, should/can we sync them uniformly?

Dockerfile
tvoran: I think you'd just need to copy LICENSE from the repo root; that's how [docker-vault/ubi/Dockerfile](https://github.com/hashicorp/docker-vault/blob/e4a92e52ee3d480e4fb45f26e5a7d1c295ce0d38/ubi/Dockerfile#L16-L17) is doing it. And they're both Mozilla Public License 2.0 so I think we're good there.

cipherboy: Oh d'oh. What about EULA and `TermsOfEvaluation` too? Both are also missing from this UBI copy:

https://github.com/hashicorp/docker-vault/blob/e4a92e52ee3d480e4fb45f26e5a7d1c295ce0d38/ubi/Dockerfile#L42-L43

vs our existing Alpine image:

https://github.com/hashicorp/vault/blob/main/Dockerfile#L23

tvoran: I don't think those are necessary for redhat certification. Looks like it's just doing the same thing as the library alpine image (since the downloaded release includes those files): https://github.com/hashicorp/docker-vault/blob/e4a92e52ee3d480e4fb45f26e5a7d1c295ce0d38/0.X/Dockerfile#L42-L43

.github/workflows/build.yml
cipherboy: @tvoran This has been updated.

cipherboy:
Rerunning CRT after these updates... :-)

.github/workflows/build.yml
```diff
@@ -6,6 +6,7 @@ on:
     branches:    
       # Push events on main branch
       - main
+      - cipherboy-docker-ubi-crt-enablement
```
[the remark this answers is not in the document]
cipherboy: Added.

tvoran (approved):

tvoran:
I tested the ubi image from [this build](https://github.com/hashicorp/vault/actions/runs/2272413057) and it all seems to work! Thanks for doing all this!

cipherboy:
Bump @kpenfound and @modrake, any chance I could get a releng ACK before merging?

claire-labry:
@cipherboy Kyle + Morgan are out this week -- pulling in @samsalisbury on this PR [its after-hours in the UK] but I believe he has most context and knowledge around this 🙏🏻

cipherboy:
Thanks @claire-labry! @samsalisbury tomorrow or any time this week is fine, we're not yet at 1.11 release date and we've missed the 1.10.3 window. :-)

Dockerfile
```diff
@@ -56,3 +66,86 @@ ENTRYPOINT ["docker-entrypoint.sh"]
...
+# Vault.
+EXPOSE 8200
+
+# The entry point script uses dumb-init as the top-level process to reap any
+# zombie processes created by Vault sub-processes.
+#
+# For production derivatives of this container, you shoud add the IPC_LOCK
+# capability so that Vault can mlock memory.
+COPY .release/docker/ubi-docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+ENTRYPOINT ["docker-entrypoint.sh"]
```
samsalisbury, line 144: I wonder if keeping the filename consistent (i.e. `ubi-docker-entrypoint.sh`) could be clearer here? (First read through I missed the rename.)

```suggestion
COPY .release/docker/ubi-docker-entrypoint.sh /usr/local/bin/
ENTRYPOINT ["ubi-docker-entrypoint.sh"]
```

Dockerfile
```diff
@@ -56,3 +66,86 @@ ENTRYPOINT ["docker-entrypoint.sh"]
...
+      version=${PRODUCT_VERSION} \
+      release=${PRODUCT_VERSION} \
+      summary="Vault is a tool for securely accessing secrets." \
+      description="Vault is a tool for securely accessing secrets. A secret is anything that you want to tightly control access to, such as API keys, passwords, certificates, and more. Vault provides a unified interface to any secret, while providing tight access control and recording a detailed audit log."
+
+COPY LICENSE /licenses/mozilla.txt
+
+# Set ARGs as ENV so that they can be used in ENTRYPOINT/CMD
+ENV NAME=$NAME
+ENV VERSION=$VERSION
```
samsalisbury, line 96: I don't see these referenced in the `ENTRYPOINT`, the `ubi-docker-entrypoint.sh`, or the `CMD` are we sure this is needed?

```suggestion
```

Dockerfile
```diff
@@ -56,3 +66,86 @@ ENTRYPOINT ["docker-entrypoint.sh"]
...
+
+
+## UBI DOCKERFILE ##
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5 as ubi
+
+ARG BIN_NAME
+# NAME and PRODUCT_VERSION are the name of the software in releases.hashicorp.com
+# and the version to download. Example: NAME=vault PRODUCT_VERSION=1.2.3.
+ARG NAME=vault
+ARG PRODUCT_VERSION
```
samsalisbury, line 78: I don't think we need the `NAME` arg here as I don't see it referenced anywhere. However we should have a `PRODUCT_REVISION` arg as that's passed in automatically by the Action.

```suggestion
# PRODUCT_VERSION is the version built dist/$TARGETOS/$TARGETARCH/$BIN_NAME,
# which we COPY in later. Example: PRODUCT_VERSION=1.2.3.
ARG PRODUCT_VERSION
ARG PRODUCT_REVISION
```

Dockerfile
```diff
@@ -56,3 +66,86 @@ ENTRYPOINT ["docker-entrypoint.sh"]
...
+# TARGETARCH and TARGETOS are set automatically when --platform is provided.
+ARG TARGETOS TARGETARCH
+
+# Additional metadata labels used by container registries, platforms
+# and certification scanners.
+LABEL name="Vault" \
+      maintainer="Vault Team <vault@hashicorp.com>" \
+      vendor="HashiCorp" \
+      version=${PRODUCT_VERSION} \
+      release=${PRODUCT_VERSION} \
```
samsalisbury, line 88: I'm not sure what "release" means here, but could it be helpful to have `revision=${PRODUCT_REVISION}` here in the metadata?
```suggestion
      release=${PRODUCT_VERSION} \
      revision=${PRODUCT_REVISION} \
```

.github/workflows/build.yml
```diff
@@ -6,6 +6,8 @@ on:
     branches:    
       # Push events on main branch
       - main
```
samsalisbury, line 8: For continuous feedback on PRs, I would remove this branch filter so that we can immediately see the effects of changes made. For most projects, I would make this a permanent change on `main` so that all branches run the build, as this will help catch build issues early and avoid them polluting main. (These cross-platform builds often get broken by accidentally non-cross-platform code, as well as any changes to the build config itself as in this PR.) See [this newsletter article](https://hashicorp.atlassian.net/wiki/spaces/ENSV/pages/2299233161/Feb+18+2022+-+Faster+Build+Issue+Feedback) on why we made this the general recommendation back in Feb: 

```suggestion
    branches:    
      # Push events on main branch
      - main
```

samsalisbury (commented):
I think there are a few tweaks that could be helpful here, I hope it all makes sense, thanks.

Dockerfile
cipherboy: Release is `PRODUCT_VERSION`, set via CRT: https://github.com/hashicorp/vault/pull/15272/files/8ba72bd42342ad2d01e65d1d11641a08cff93964#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R280  Because we set version and not revision, I think that means PRODUCT_REVISION isn't useful for us to set?

tvoran: As far as I know more labels are fine from RedHat's perspective. Their certification preflight tool does require these labels to be present: name, vendor, version, release, summary, description. [[RedHat docs](https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.45/html/red_hat_openshift_software_certification_policy_guide/assembly-requirements-for-container-images_openshift-sw-cert-policy-introduction#con-image-metadata-requirements_openshift-sw-cert-policy-container-images)]

cipherboy: Ah, should release be set to `PRODUCT_REVISION` then, or some combination of the two?

tvoran: Oh that's an interesting idea, `PRODUCT_REVISON` is the git sha, right? So that would seem to fit in the description for `release`.

Dockerfile
samsalisbury: Definitely, `PRODUCT_VERSION` and `PRODUCT_REVISION` are both set automatically by the Action when it runs `docker build` so all Dockerfiles and stages within them should have those two ARGs, even if they're not otherwise referenced, as I think the build would fail if an unrecognised build arg is set.

Dockerfile
samsalisbury: Got it, so `release` is a Red Hat required field.. Seems worth using `PRODUCT_REVISION` there I guess, at least it's some extra info rather than just repeating the version?

.github/workflows/build.yml
samsalisbury: I think this is worth committing sooner rather than later so we can see the latest results of the build here in the PR @cipherboy, what do you think?

cipherboy: Ah that's fine, we had added `workflow_dispatch` to this PR for that, but if this is better, we should remove the `workflow_dispatch`.

I think the right change is to remove both, right? As I read it the suggestion is a no-op which isn't what we want, we want to remove it completely. I'll make these changes now that we've settled on them.

Dockerfile
cipherboy:
```suggestion
      release=${PRODUCT_REVISION} \
      revision=${PRODUCT_REVISION} \
```

Dockerfile
```diff
@@ -8,8 +9,17 @@ ARG PRODUCT_VERSION
...
 
-LABEL maintainer="Vault Team <vault@hashicorp.com>"
-LABEL version=${PRODUCT_VERSION}
+# Additional metadata labels used by container registries, platforms
+# and certification scanners.
+LABEL name="Vault" \
+      maintainer="Vault Team <vault@hashicorp.com>" \
+      vendor="HashiCorp" \
+      version=${PRODUCT_VERSION} \
+      release=${PRODUCT_VERSION} \
```
cipherboy, line 18:
```suggestion
      release=${PRODUCT_REVISION} \
      revision=${PRODUCT_REVISION} \
```

Dockerfile
```diff
@@ -8,8 +9,18 @@ ARG PRODUCT_VERSION
 # TARGETARCH and TARGETOS are set automatically when --platform is provided.
```
cipherboy, line 9:
```suggestion
ARG PRODUCT_REVISION
# TARGETARCH and TARGETOS are set automatically when --platform is provided.
```

cipherboy:
Ok, I think most of this is updated.

Dockerfile
tvoran: I think it was kept as `docker-entrypoint.sh` on the image since that's is how it's referenced in the helm chart when vault is deployed: https://github.com/hashicorp/vault-helm/blob/4005767e87c29ca676103c24f44d20f4d8a54117/templates/_helpers.tpl#L174

So changing it like this will break the helm chart.

cipherboy: https://github.com/hashicorp/vault/pull/15443 -- fix reverting this change.

mdeggies:
@samsalisbury should the project ID be set to `5fda55bd2937386820429e0c`? That is the one in the URL when I navigate to vault on RedHat's registry, e.g. https://catalog.redhat.com/software/containers/hashicorp/vault/5fda55bd2937386820429e0c. If that's not correct and the project ID is found in a different spot, let me know, as I'm rolling this out to consul here https://github.com/hashicorp/consul/pull/13232 🙏  Thanks!

mdeggies:
I believe we also need the EULA/TOE files present in the container: https://github.com/hashicorp/vault-enterprise/blob/main/Dockerfile#L35-L36. Will open a PR to get this added.

cipherboy:
@mdeggies said:
> @samsalisbury should the project ID be set to `5fda55bd2937386820429e0c`? That is the one in the URL when I navigate to vault on RedHat's registry, e.g. https://catalog.redhat.com/software/containers/hashicorp/vault/5fda55bd2937386820429e0c. If that's not correct and the project ID is found in a different spot, let me know, as I'm rolling this out to consul here [hashicorp/consul#13232](https://github.com/hashicorp/consul/pull/13232) pray Thanks!

The ospid is from the https://connect.redhat.com/projects page for the project, and this value (`ospid-f0a92725-d8c6-4023-9a87-ba785b94c3fd`) matches what's in the **PID** field on the Vault page. It looks like Consul's matches the [value from the registry URL](https://catalog.redhat.com/software/containers/hashicorp/consul/612d01d49f14588c41ebf67c) but Vault's doesn't for some reason. I have no idea why :-D

mdeggies:
Yeah the mismatch is super odd, but redhat has had a lot of inconsistencies/weird devex issues.. @alvin-huang mentioned this is one of them. :D

samsalisbury:
Ah, yes, this took a while to figure out @mdeggies, it is odd. At the moment, we assume the OSPID will work, and if it doesn't then publishing fails with a log message guiding the team what to do, [here](https://github.com/hashicorp/crt-workflows-common/commit/36dda0a16382910235570394ec56e9dc7ea77dfd#diff-7c6cc1cdd4ea60afb47761069609c53d9226fdd8805a95a667e51b4e9c5de235R54-R60) (internal only link).

PR #15273, opened 2022-05-03
Sidebranch: MFA end user setup

Monkeychip:
This PR handles the end user setup for MFA.

image

image

image


ui/app/components/mfa-setup-step-one.js
```diff
@@ -0,0 +1,75 @@
...
+export default class MfaSetupStepOne extends Component {
+  @service store;
+  @tracked error;
+  @tracked warning = '';
+  @tracked qrCode = '';
+
+  @action
+  redirectPreviousPage() {
+    this.args.restartFlow();
+    window.history.back();
```
Monkeychip, line 26: later we can work on a service that records route history and allows us to go to the previous route. For now, given our time constraints this will work.

ui/app/components/mfa-setup-step-one.js
```diff
@@ -0,0 +1,75 @@
...
+      this.args.saveUUIDandQrCode(this.UUID, response.data?.url); // parent needs to keep track of UUID and qrCode.
+      // if there was a warning it won't fail but needs to be handled here and the flow needs to be interrupted
+      let warnings = response.warnings || [];
+      if (warnings.length > 0) {
+        this.UUID = ''; // clear UUID
+        const alreadyGenerated = warnings.find((w) =>
+          w.includes('Entity already has a secret for MFA method')
+        );
+        if (alreadyGenerated) {
+          // replace warning because it comes in with extra quotes: "Entity already has a secret for MFA method ""' "
```
Monkeychip, line 61: this is something I need to follow up with the backend about. It looks like it should be returning a method name but it is not so instead it's a extra quotes at the end of the warning. If the backend address this we can remove this replace.

ui/app/components/splash-page.js
```diff
@@ -7,6 +7,7 @@ export default Component.extend({
   auth: service(),
   store: service(),
   tagName: '',
+  showTruncatedNavBar: true,
```
Monkeychip, line 10: we have the one case where we are using the splash page component but need to show the navbar hence the new conditional.

(I have a todo in another PR to glimmerize the splash page component and add documentation. However, it's outside the scope of this already large PR).

ui/app/styles/components/list-item-row.scss
```diff
@@ -16,6 +16,12 @@
     font-weight: $font-weight-semibold;
     color: $ui-gray-500;
   }
+
```
Monkeychip, line 19: used to center the qrCode.

ui/app/styles/core/buttons.scss
```diff
@@ -183,6 +183,10 @@ $button-box-shadow-standard: 0 3px 1px 0 rgba($black, 0.12);
     font-weight: $font-weight-semibold;
   }
 
+  &.has-text-danger {
```
Monkeychip, line 186: effects one other button that uses has-text-danger. confirmed with design this fix for both buttons is desired (The MFA designs had the border).

ui/app/styles/core/helpers.scss
```diff
@@ -81,6 +81,7 @@
 .is-flex-start {
   display: flex !important;
   justify-content: flex-start;
+  gap: $spacing-m;
```
Monkeychip, line 84: spacing between buttons. confirmed only impacts the desired row of buttons.

ui/app/templates/components/mfa-setup-step-one.hbs
```diff
@@ -0,0 +1,39 @@
...
+  you are not prevented from logging into Vault in the future, once MFA is fully enforced.
+</p>
+<form id="mfa-setup-step-one" {{on "submit" this.verifyUUID}}>
+  <MessageError @errorMessage={{this.error}} class="has-top-margin-s" />
+  <div class="field has-top-margin-l">
+    <label class="is-label">
+      Method ID
+    </label>
+
+    {{! template-lint-disable no-autofocus-attribute}}
```
Monkeychip, line 12: thoughts on removing the autofocus attr on the Input?

ui/package.json
```diff
@@ -124,6 +124,7 @@
     "ember-modifier": "^3.1.0",
     "ember-page-title": "^6.2.2",
     "ember-power-select": "^5.0.3",
+    "ember-qrcode-shim": "^0.4.0",
```
Monkeychip, line 127: Jordan I'll leave some additional comments in slack about this, but for now, this solves the qrCode issue. It's a wrapped that goes around the qrCode.js library, which ALSO hasn't been updated in a long time, but is still recommended as the way to handle qr codes.

ui/app/styles/core/helpers.scss
zofskeez: I wonder if this would be better in it's own utility class in case `is-flex-start` needs to be used without that gap spacing?

ui/app/templates/components/mfa-setup-step-one.hbs
zofskeez: I'm not sure if we have a consistent pattern with autofocusing but I don't have a problem with removing it.

ui/package.json
zofskeez: This looks good to me. Let's roll with it if it suits our needs and doesn't introduce deprecation warnings or those type of things in the build.

zofskeez (approved):
Great work on this! I left some comments but nothing to block merging.

ui/app/templates/components/mfa-setup-step-two.hbs
```diff
@@ -0,0 +1,35 @@
...
+</p>
+<div class="field has-top-margin-l">
+  <MessageError @errorMessage={{this.error}} class="has-top-margin-s" />
+  {{#if @warning}}
+    <AlertBanner @type="info" @title="MFA enabled" @message={{@warning}} class="has-top-margin-l" />
+  {{else}}
+    <div class="list-item-row">
+      <div class="center-display">
+        {{! template-lint-disable no-curly-component-invocation }}
+        {{qr-code text=@qrCode colorLight="#F7F7F7" width=155 height=155 correctLevel="L"}}
```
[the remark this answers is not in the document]
Monkeychip: I tried, it borked everything. hastag oldLibrary

ui/app/styles/core/helpers.scss
Monkeychip: Good call. Amended.

PR #15274, opened 2022-05-03
clone vault agent client with headers

ccapurso:
https://github.com/hashicorp/vault/pull/15204 introduced a change in which the Vault client is cloned for the auth handler. The headers are not cloned, however, which results in headers like `X-Vault-Namespace` being dropped. This PR changes the `Clone` call to a `CloneWithHeaders`.


jasonodonnell (approved):
Good catch and thanks for fixing!

Issue #15108, opened 2022-04-21
MFA TOTP is deleted after vault restart


hennadii2012:
**Describe the bug**
For vault SSO 1.10.0 in docker mfa totp method is deleted after restart all nodes in raft cluster

**To Reproduce**
1. Run `curl     --header "X-Vault-Token: $VAULT_TOKEN"     --request POST     --data @issuer.json    ${VAULT_ADDR}/v1/identity/mfa/method/totp
`
content of issuer.json:
{
  "issuer": "vault",
  "algorithm": "SHA512"
}

2. RUN docker stop vault-container-name  on all three nodes of raft cluster
3. RUN docker start   vault-container-name  on all three nodes of raft cluster
4. RUN curl     --header "X-Vault-Token: $VAULT_TOKEN"     --request LIST    ${VAULT_ADDR}/v1/identity/mfa/method/totp. It return nothing, but must return methoid from first step

**Expected behavior**
Return methoid from first step after restarts

**Environment:**
* Vault Server Version 1.10.0
* Vault CLI Version 1.10.0
* Server Operating System/Architecture: docker (vault:1.10.0) on ubuntu 20.04

Vault server configuration file(s):

```hcl
storage "raft" {
  path    = "/vault/file"
  node_id = "vault1"


  retry_join {
    leader_tls_servername   = "vault1.example.com"
    leader_api_addr         = "https://vault1.example.com:8200"
    leader_ca_cert_file     = "/opt/vault/tls/vault-ca.pem"
    leader_client_cert_file = "/opt/vault/tls/vault-cert.pem"
    leader_client_key_file  = "/opt/vault/tls/vault-key.pem"
  }
  retry_join {
    leader_tls_servername   = "vault2.example.com"
    leader_api_addr         = "https://vault2.example.com:8200"
    leader_ca_cert_file     = "/opt/vault/tls/vault-ca.pem"
    leader_client_cert_file = "/opt/vault/tls/vault-cert.pem"
    leader_client_key_file  = "/opt/vault/tls/vault-key.pem"
  }
  retry_join {
    leader_tls_servername   = "vault3.example.com"
    leader_api_addr         = "https://vault3.example.com:8200"
    leader_ca_cert_file     = "/opt/vault/tls/vault-ca.pem"
    leader_client_cert_file = "/opt/vault/tls/vault-cert.pem"
    leader_client_key_file  = "/opt/vault/tls/vault-key.pem"
  }
}
listener "tcp" {
[...]

PR #15275, opened 2022-05-03
Add upgrade note for #15108.

ncabatoff:


website/content/docs/upgrading/upgrade-to-1.10.x.mdx
```diff
@@ -109,3 +109,16 @@ set to `unauth`.
 There is a workaround for this error that will allow you to sign in to Vault using the OIDC
 auth method. Select the "Other" tab instead of selecting the specific OIDC auth mount tab.
 From there, select "OIDC" from the "Method" select box and proceed to sign in to Vault.
+
+### Login MFA configuration not loaded at startup
```
mladlow, line 113:
```suggestion
### Login MFA configuration does not persist after startup
```

mladlow (commented):
I'm not 100% sold on my comment but I think we should make the heading sound more severe.

website/content/docs/upgrading/upgrade-to-1.10.x.mdx
ncabatoff: How about "login mfa config disappears after restart"?

mladlow (approved):

PR #15276, opened 2022-05-03
Update deps for consul-template 0.29.0

swenson:
This requires bumping https://github.com/mitchellh/go-testing-interface. For this new version, we have to create a wrapper to convert the stdlib `testing.TB` interface to the`mitchellh/go-testing-interface` `T` interface, since it uses `Parallel()` now, which is not supported by `testing.TB`. This had to be added to a new package, `benchhelpers`, to avoid a circular dependency in `testhelpers`.

We also have to *unbump* https://github.com/armon/go-metrics since updating it breaks our usage of https://github.com/google/go-metrics-stackdriver

I verified that the new `pkiCert` template function works with agent injection using annotations like:

```yaml
vault.hashicorp.com/agent-inject-secret-sample.crt: "pki/issue/example-dot-com"
vault.hashicorp.com/agent-inject-template-sample.crt: |
  {{ pkiCert "pki/issue/example-dot-com" "common_name=foo.example.com" "ttl=1h" }}
```


helper/benchhelpers/benchhelpers.go
```diff
@@ -0,0 +1,19 @@
...
+
+type tbWrapper struct {
+	testing.TB
+}
+
+func (b tbWrapper) Parallel() {
+	// no-op
+}
+
+func TBtoT(tb testing.TB) testing2.T {
```
calvn, line 17: This felt a bit odd at first, but I see why you're doing it since `go-testing-interface` explicitly only supports satisfying `testing.T` as [mentioned here](https://github.com/mitchellh/go-testing-interface/pull/5#issuecomment-624214115).

swenson: Yeah. This wasn't my favorite solution, but I don't see a better way. :/

swenson:
Ignore this PR for now while I fight with `go mod`. I'll probably close and reopen it once I get this resolved.

swenson:
Going to close and open a new PR for this

PR #15277, opened 2022-05-03
Add Multiple Issuers to PKI Secret Engine Mounts - PKI Pod

cipherboy:
## Overview

This pull request adds support for multiple issuers (mixed roots and intermediates) under the same PKI secrets engine mount point. We expose two new major API paths, `/issuer{s}` and `/key{s}`, for managing issuing certificates and their associated keys respectively.

The intent of this design was to enable a forward-rotation strategy for root CAs, with additional goals of aiding the live rotation of intermediates as well. Previously, to rotate issuing certificates in Vault 1.10.x and earlier, either the mount would have to be completely removed (and replaced with a new one -- with all audit and role information recreated) or the root/intermediate would have to have been deleted.

Existing APIs are preserved and will continue to function. We introduce the notion of a "default" issuer (`/config/issuers`) to handle underlying selection of issuer when existing APIs elide that information. Role now allow selection of an explicit issuer, allowing operators to choose issuers for users without them necessarily noticing the impact.

Unless otherwise unauthenticated, many of these new APIs should be considered privileged and adequately ACL'd.

Additionally, as a design choice, all certificate authorities (whether the mount's issuing certificate with backing key or a parent root certificate without key) in the mount have an issuer entry. This allows a mount's _functional_ issuers to change over time (including the introduction of a soft delete, wherein both keys and certificates for an issuer are perserved, but the ability to revoke its certificates remains). There's now explicit chain building support as well; previously on many intermediate mounts, the parent root certificate(s) were not imported -- they can now be imported and exposed to callers.

However, as a result of this design, we open the possibility for many new, streamlined usages of the PKI secrets engine. To give just one example, a twin-intermediate approach is now viable, wherein high-impact certs could be signed by an HSM-backed intermediate and low-impact certs could be signed by a Vault-backed intermediate, all chosen by the operator based on user role selection. Users could be transparently switched between intermediates, without needing to update their client configuration. This is enabled by allowing multiple _active_ issuers in the same mount.

---

## Review Strategy

This PR is segmented into a number of incremental commits; see incremental commit messages and changes for details.

At a high level, we'd love to see the following questions answered:

 1. Are there any UX gaps in this API as proposed now, that need to be addressed (either as part of this PR or separately)? 
 2. Is forwarding of requests handled correctly? (Active node of the primary cluster needs to handle much of issuer&key responses along with role and configuration information).
 3. Do we need a read lock on the active primary/secondary &c for issuer consistency? These _should_ be relatively low-traffic APIs, but there's a risk. How would such a distributed lock be created and held? Notably, issuers are only modified on the active node of the primary cluster, but these get replicated on storage to other nodes in secondary clusters, and thus they _may_ have a temporarily inconsistent view of storage that may allow requests to suceed/fail until they've sufficiently caught up.
 4. There's many places `RawSubject` and `RawIssuer` are used directly for comparisons, but I believe this is incorrect; in particular, RDNs are `SEQUENCE`s of `SET`s, which means the internal comparison _should_ be an unordered comparison, meaning we may incorrectly deny some valid links due to differing order of RDN pieces. Is this critical? I don't believe so right now as we haven't previously had support for this.
 5. As part of this work, CRLs will now be rebuilt on secondary clusters when necessary; in particular, this means that when adding a new secondary cluster, it will no longer serve a `nil` CRL until a certificate is revoked (or the CRL is manually rotated). 
 6. Are there any new security concerns or are the privileged APIs largely understood?
 
## Outstanding Work

There's a bit of work left to enable managed keys in Enterprise. @stevendpclark has most of this work already in branches and it will be opened in parallel. When this merges, we'll do an OSS->ENT merge and then merge that PR.

We aim to store the issuer ID used to issue requests (when `no_store=false`); this work isn't done but doesn't hold up this merge.

There's another piece of correctness around issuer import (ensuring correct comparisons even when PEM contents differ); this also shouldn't hold up the merge.

There's additional work to allow revoking an issuer if _another_ issuer in the same mount signed it; this can happen after the `main` merge.

The many UI changes to incoprorate this feature set will come at a separate point in time.

Extensive documentation for this change set is still outstanding; API docs are _started_ in a [separate PR](https://github.com/hashicorp/vault/pull/15238). Other docs, learn guides, and tutorials are still outstanding.

## Miscellanea 

<details>

<summary> Incremental PRs </summary>

In working towards this feature, many incremental PRs were made towards this feature. We've used [the `rotation` label](https://github.com/hashicorp/vault/pulls?q=is%3Aopen+is%3Apr+label%3Arotation) for this; please feel free to walk through earlier discussions on earlier PRs for some discussion on these topics.

</details>

<details>

<summary>Hashicorp Internal Context</summary>

See internal RFC VLT-191 for extensive information about this change set; feel free to ask questions on `#team-vault-crypto`.

Meeting invites will be scheduled to give crypto team and core team a place to discuss this PR and start reviewing it. In discussion with @sgmiller, we felt it was best to open a single large PR, rather than attempting to recreate the incremental PRs above. This PR is essentially a large refactoring of the PKI code base, breaking several 

</details>

If possible, I'd like to see this rebased (rather than squash merged), to give adequate credit to everyone who's contributed along the way. Many internal PRs were squashed merged and I believe the commit series makes most sense separately rather than as a large single change. Thanks @kitography and @stevendpclark!


builtin/logical/pki/storage.go
```diff
@@ -0,0 +1,857 @@
...
+
+type keyConfigEntry struct {
+	DefaultKeyId keyID `json:"default" structs:"default" mapstructure:"default"`
+}
+
+type issuerConfigEntry struct {
+	DefaultIssuerId issuerID `json:"default" structs:"default" mapstructure:"default"`
+}
+
+func listKeys(ctx context.Context, s logical.Storage) ([]keyID, error) {
```
cipherboy, line 157: Per bk: storage list operations are not cached.

cipherboy (commented):
leaf_not_after_behavior should be in upgrade notes.

cipherboy (commented):
Feature: root has no `ca_chain` on issuance response, but intermediate does? 

Part of certutil: https://github.com/hashicorp/vault/blob/main/sdk/helper/certutil/helpers.go#L843-L855

Not really a bug, is a feature.

cipherboy (commented):
Discussion: @briankassouf / @sgmiller  -- what should we do when we fail to read the existing cert bundle on migration? See: e.g. 

 - https://github.com/hashicorp/vault/blob/bba4bcb19515e8ca5ea35931c15bff216951d5e0/builtin/logical/pki/backend.go#L324
 - https://github.com/hashicorp/vault/blob/bba4bcb19515e8ca5ea35931c15bff216951d5e0/builtin/logical/pki/storage_migrations.go#L34

cipherboy (commented):
Upgrade note: First rebuild of CRL on migrated cluster will be more expensive.

builtin/logical/pki/chain_util.go
```diff
@@ -0,0 +1,1305 @@
...
+
+func prettyIssuer(issuerIdEntryMap map[issuerID]*issuerEntry, issuer issuerID) string {
+	if entry, ok := issuerIdEntryMap[issuer]; ok && len(entry.Name) > 0 {
+		return "[id:" + string(issuer) + "/name:" + entry.Name + "]"
+	}
+
+	return "[" + string(issuer) + "]"
+}
+
+func rebuildIssuersChains(ctx context.Context, s logical.Storage, referenceCert *issuerEntry /* optional */) error {
```
swenson, line 21: There are a couple of giant functions in this file that I would love to see broken down to make reasoning about them easier, if possible. (Granted, a lot of these are detailed comments.)

Also, is it possible to use some standard-ish libraries, like [gonum](https://github.com/gonum/gonum), to do some of these graph computations?

builtin/logical/pki/cert_util.go
```diff
@@ -89,32 +79,45 @@ func getFormat(data *framework.FieldData) string {
...
 
-	var bundle certutil.CertBundle
-	if err := bundleEntry.DecodeJSON(&bundle); err != nil {
-		return nil, errutil.InternalError{Err: fmt.Sprintf("unable to decode local CA certificate/key: %v", err)}
+	if bundle == nil {
+		return nil, errutil.UserError{Err: "no CA information is present"}
 	}
 
-	parsedBundle, err := parseCABundle(ctx, b, req, &bundle)
+	parsedBundle, err := parseCABundle(ctx, b, req, bundle)
```
sgmiller, line 104: Why was this changed to pass by value?

builtin/logical/pki/path_fetch_keys.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	for _, identifier := range entries {
+		key, err := fetchKeyById(ctx, req.Storage, identifier)
+		if err != nil {
+			return nil, err
+		}
+
+		responseKeys = append(responseKeys, string(identifier))
+		responseInfo[string(identifier)] = map[string]interface{}{
```
swenson, line 54: I don't know if it is too late for this, but it seems like we could simplify the output of this endpoint a little with something like (paraphrased):

```json
{"keys": [
  {"id": "uuid...", other info ... },
  {"id": "uuid2...", other info ... },
]}
```  
 
Mostly, I'm always hesitant to put semantic meaning in the keys of a map (i.e., the keyIds being present as keys), when really what you have is a list. Plus, this means it is easier in the future to have multiple entries with the same `id` (like, if you had previous versions, or multiple valid versions, etc.)

builtin/logical/pki/path_manage_keys.go
```diff
@@ -0,0 +1,195 @@
...
+			},
+			keyTypeParam: {
+				Type:        framework.TypeString,
+				Default:     "rsa",
+				Description: `Type of the secret key to generate`,
+			},
+			keyBitsParam: {
+				Type:        framework.TypeInt,
+				Default:     2048,
+				Description: `Type of the secret key to generate`,
```
swenson, line 29: I don't think this description is correct. :)

builtin/logical/pki/chain_util.go
sgmiller: I'd love to see benchmark profiles for this code.

builtin/logical/pki/path_manage_issuers.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	// While we're here, check if we should warn about a bad default key. We
+	// do this unconditionally if the issuer or key was modified, so the admin
+	// is always warned. But if unrelated key material was imported, we do
+	// not warn.
+	config, err := getIssuersConfig(ctx, req.Storage)
+	if err == nil && len(config.DefaultIssuerId) > 0 {
+		// We can use the mapping above to check the issuer mapping.
+		if keyId, ok := issuerKeyMap[string(config.DefaultIssuerId)]; ok && len(keyId) == 0 {
```
swenson, line 231: If `!ok` would we also want to give a warning for missing a key for the default issuer?

changelog/15277.txt
```diff
@@ -0,0 +1,11 @@
+```release-note:feature
+**Allows Multiple Issuer Certificates to enable Non-Disruptive
+Intermediate/Root Certificate Rotation**: This introduces /keys and /issuers
+endpoints to allow import, generation and configuration of any number of keys
+or issuers that can be used to issue and revoke certificates.  Keys and Issuers
+can be referred to by (a) a unique UUID; (b) a name; (c) “default”.  If an
+issuer existed prior to this feature, that issuer will be tagged by a migration
+as “default” to allow backwards compatible calls which don’t specify an issuer.
+Creation of new roles will assume an issuer of “default” unless otherwise
+specified.  This default can be configured at /config/issuers and /config/keys.
```
sgmiller, line 10: This may be too much detail for a changelog.  How to use it should be in the docs.  I'd omit the details about referencing keys and issuers, but add a sentence or two about how having multiple issuers allows in place rotation of CA certificates.

sgmiller (commented):
Comments from my first pass through.

builtin/logical/pki/path_fetch_keys.go
cipherboy: That's sadly the format of `ListResponseWithInfo` that we've gotta obey. Its a standard Vault Response type, :/

builtin/logical/pki/cert_util.go
cipherboy: This wasn't, the function was also updated to be pass-by-reference.

builtin/logical/pki/chain_util.go
cipherboy: Just as an aside, like the idea of benchmarks, but this function gets called one of two places:

 - Importing/generating an issuer,
 - Deleting an issuer.
 
And is otherwise static.

Deleting might be a little surprising for a perf hit but importing/generating I don't think is that surprising.

@sgmiller What are you thinking of for benchmarks? How many issuers until it stalls? Complexity of chains? No changes and force-trigger rebuilds? &c.

builtin/logical/pki/path_fetch_keys.go
swenson: :+1:

builtin/logical/pki/path_manage_issuers.go
cipherboy: There's two scenarios where we're not ok:

 1. The config is completely bad. We warn a few other places in this case; [see the origin PR](https://github.com/hashicorp/vault/pull/15178) for context or the `deleteIssuer` code in `storage.go`.
 2. On imports of _unrelated certs/keys_. 

In case #2, we could be importing some other issuer completely unrelated to the configured default. Say, a future root or its key. Here, the import-local `issuerKeyMap` won't contain the default issuer, even if its otherwise perfectly valid.

So, warning on `!ok` alone is not sufficient information by itself.

That's what the comment 4 lines above was trying to convey, let me know if you have other wording suggestions :-)

> But if unrelated key material was imported, we do not warn.

builtin/logical/pki/chain_util.go
cipherboy: See #15315 :-) Looks like those numbers are fairly representative, so delete overhead isn't all that much.

builtin/logical/pki/path_manage_keys.go
```diff
@@ -0,0 +1,195 @@
...
+	exportPrivateKey := false
+	var keyBundle certutil.KeyBundle
+	var actualPrivateKeyType certutil.PrivateKeyType
+	switch {
+	case strings.HasSuffix(req.Path, "/exported"):
+		exportPrivateKey = true
+		fallthrough
+	case strings.HasSuffix(req.Path, "/internal"):
+		keyType := data.Get(keyTypeParam).(string)
+		keyBits := data.Get(keyBitsParam).(int)
```
stevendpclark, line 84: I'm adding extra input validation to keyType, keyBits

builtin/logical/pki/crl_util.go
```diff
@@ -1,24 +1,92 @@
...
+	return nil
+}
+
+// rebuild is to be called by various write apis that know the CRL is to be updated and can be now.
+func (cb *crlBuilder) rebuild(ctx context.Context, b *backend, request *logical.Request, forceNew bool) error {
+	return cb._doRebuild(ctx, b, request, forceNew, _ignoreForceFlag)
+}
+
+// requestRebuildIfActiveNode will schedule a rebuild of the CRL from the next read or write api call assuming we are the active node of a cluster
+func (cb *crlBuilder) requestRebuildIfActiveNode(b *backend) {
```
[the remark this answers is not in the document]
cipherboy: Can we restructure to avoid this problem? I think if we use `forceRebuild` as a true atomic, we could avoid holding the lock when _setting_ it and only grab the lock when reading it (inside `_doRebuild`). I think we can also change line 82 from `defer atomic.Store...` into a straight `atomic.Store` and the lock will prevent us from rebuilding the CRL at the same time -- the in-flight CRL build will be "wrong" if we've been requested to rebuild it for some reason anyways.

builtin/logical/pki/path_manage_keys.go
cipherboy: Addressed in #15319.

builtin/logical/pki/path_manage_keys.go
cipherboy: In #15319.

builtin/logical/pki/storage.go
```diff
@@ -0,0 +1,857 @@
...
+		return config.DefaultKeyId, nil
+	}
+
+	keys, err := listKeys(ctx, s)
+	if err != nil {
+		return keyID("list-error"), err
+	}
+
+	// Cheaper to list keys and check if an id is a match...
+	for _, keyId := range keys {
```
stevendpclark, line 421: A note from @sgmiller, he would prefer we use a Get operation to lookup a key id and only perform the list and iteration approach for name lookups (assuming the Get operation return nil,nil)

builtin/logical/pki/backend.go
```diff
@@ -233,3 +290,72 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
...
+	if err != nil {
+		b.Logger().Error(fmt.Sprintf("Failed loading PKI migration status, staying in legacy mode: %v", err))
+		return
+	}
+
+	if info.isRequired {
+		b.Logger().Info("PKI migration is required, reading cert bundle from legacy ca location")
+		b.pkiStorageVersion.Store(0)
+	} else {
+		b.Logger().Debug("PKI migration completed, reading cert bundle from key/issuer storage")
```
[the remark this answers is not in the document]
stevendpclark: Addressed in #15322

builtin/logical/pki/storage.go
stevendpclark: Addressed in #15322

builtin/logical/pki/crl_util.go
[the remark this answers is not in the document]
stevendpclark: Reworked in #15325

sgmiller:
> Discussion: @briankassouf / @sgmiller -- what should we do when we fail to read the existing cert bundle on migration? See: e.g.

We're in a pretty bad state at that point right? As in an unusable backend?  At the very least we should proceed no further with the upgrade.  We could perhaps start a goroutine that retries this periodically in case storage came back.

sgmiller (approved):
Not finding anything in my second pass to bring up.

briankassouf:
> Discussion: @briankassouf / @sgmiller -- what should we do when we fail to read the existing cert bundle on migration? See: e.g.

Do we have the ability to operate backwards compat on the old storage layout? Supposedly this would be required if a secondary or standby are upgraded ahead of a primary/active. If so, could we just fall back to old layout on the active node as well?

cipherboy:
@briankassouf said:
> @cipherboy said:
> > Discussion: @briankassouf / @sgmiller -- what should we do when we fail to read the existing cert bundle on migration? See: e.g.
> 
> Do we have the ability to operate backwards compat on the old storage layout? Supposedly this would be required if a secondary or standby are upgraded ahead of a primary/active. If so, could we just fall back to old layout on the active node as well?

We do, but note that this is on attempting to _read_ the old cert bundle, for migration. As @sgmiller pointed out: 

> We're in a pretty bad state at that point right? As in an unusable backend? At the very least we should proceed no further with the upgrade. We could perhaps start a goroutine that retries this periodically in case storage came back.

I'm not sure if there's the expectation that we retry the migration, or if we should just fall back unconditionally to the old bundle, hoping that after some node/service restart, it becomes available again and we can re-try the migration...

@stevendpclark If the active node of the primary cluster has read errors during migration and fails, and we subsequently demote it and promote a new active node that was already a standby in the same cluster, will that newly elected standby-turned-active reattempt migration?

stevendpclark:
@cipherboy said:
> > > Discussion: @briankassouf / @sgmiller -- what should we do when we fail to read the existing cert bundle on migration? See: e.g.
> > 
> > 
> > Do we have the ability to operate backwards compat on the old storage layout? Supposedly this would be required if a secondary or standby are upgraded ahead of a primary/active. If so, could we just fall back to old layout on the active node as well?
> 
> We do in general (not sure about the active primary node in particular), but note that this is on attempting to _read_ the old cert bundle, for migration. As @sgmiller pointed out:
> 
> > We're in a pretty bad state at that point right? As in an unusable backend? At the very least we should proceed no further with the upgrade. We could perhaps start a goroutine that retries this periodically in case storage came back.
> 

Agreed, for us to fail to read the legacy bundle either storage is corrupted or an unfortunate read blip occurred right at the point of migration. If it really is a failure to read the legacy bundle, most likely something was wrong prior to the migration with the mount and we are now exposing it. 

> I'm not sure if there's the expectation that we retry the migration, or if we should just fall back unconditionally to the old bundle, hoping that after some node/service restart, it becomes available again and we can re-try the migration... If we're still there and the storage later comes back, I _think_ we'll attempt to use it just fine, but I'm admittedly not really sure how to test this particular scenario.

At the moment the migration is attempted once per start of Vault. If it fails, Vault will remain in legacy storage mode (for that given mount), so we can issue new leaf certificates and accept requests to revoke them. We will not handle new intermediary/root certificate generation and all the new apis will fail with the migration error message. 

> 
> @stevendpclark If the active node of the primary cluster has read errors during migration and fails, and we subsequently demote it and promote a new active node that was already a standby in the same cluster, will that newly elected standby-turned-active reattempt migration?

Tested and yes, for some reason the active node fails, and we step-down/stop Vault on that node, another standby node within the cluster will wake up and attempt the migration.

stevendpclark (approved):

cipherboy:
Per internal discussion and consensus, squashing and merging. :-) Thanks everyone!!

PR #15278, opened 2022-05-03
removing prem/pro references as part of removing some build targets

hghaf099:
This PR aims to remove refrences to prem and pro build targets. Those targets are going to be removed on the ENT side.


hghaf099:
@ccapurso Would you please review this PR as well. Left over files in the OSS with regards to removing prem and pro targets.

ccapurso (approved):

PR #15279, opened 2022-05-04
UI/filter monthly graphs

hellobontempo:
Add filtering for time series graphs in client counts, and testing for serializer util functions

history-filtering


ui/app/components/clients/history.js
```diff
@@ -253,12 +261,27 @@ export default class History extends Component {
       .mounts?.find((mount) => mount.label === auth);
   }
 
+  get filteredActivityByMonth() {
+    const namespace = this.selectedNamespace;
+    const auth = this.selectedAuthMethod;
+    const namespaceData = this.getActivityResponse?.byMonth.map((m) => m.namespaces_by_key[namespace]);
```
hashishaw, line 267: Since namespace can be null if unset, I would move this line below line 270 so you don't get weird data or a failure when namespace is null

ui/app/components/clients/history.js
```diff
@@ -253,12 +261,27 @@ export default class History extends Component {
...
+    if (!namespace && !auth) {
+      return this.getActivityResponse?.byMonth;
+    }
+    if (!auth) {
+      return namespaceData;
+    }
+    const mountData = namespaceData.map((namespace) => namespace.mounts_by_key[auth]);
+    // TODO CMB come up with a better check when testing
+    return mountData.includes(undefined) ? null : mountData;
```
hashishaw, line 276: This is probably undefined because of line 267 so rearranging might help this

ui/lib/core/addon/utils/client-count-utils.js
```diff
@@ -87,3 +71,89 @@ export const homogenizeClientNaming = (object) => {
   }
   return object;
 };
+
+const flattenDataset = (object) => {
```
hashishaw, line 75: Would love to see tests for these!

ui/lib/core/addon/utils/client-count-utils.js
```diff
@@ -87,3 +71,89 @@ export const homogenizeClientNaming = (object) => {
...
+  if (Object.keys(object).includes('counts') && object.counts) {
```
hashishaw, line 76: This seems verbose, since if `counts` does not exist on object the following will by falsy, as well as if object is undefined or null. Are we trying to safeguard against other types? 
```suggestion
  if (object?.counts) {
```

ui/lib/core/addon/utils/client-count-utils.js
```diff
@@ -87,3 +71,89 @@ export const homogenizeClientNaming = (object) => {
...
+  // right now months are ordered in descending client count number
+  const sortedPayload = [...monthsArray];
+  return sortedPayload.sort((a, b) =>
+    compareAsc(parseAPITimestamp(a.timestamp), parseAPITimestamp(b.timestamp))
+  );
+};
+
+const namespaceArrayToObject = (totalClientsByNamespace, newClientsByNamespace, month) => {
+  // first nest 'new_clients' within each namespace and mount object respectively
+  totalClientsByNamespace.forEach((ns) => {
```
hashishaw, line 95: If we're mutating the data, it would be better to return a map of the `totalClientsByNamespace` instead. So eg 
```suggestion
  return totalClientsByNamespace.map((ns) => {
```
and then rather than deleting keys you can only include the items you want.

hashishaw (commented):
Good progress!

ui/app/components/clients/history.js
hellobontempo: Good call! I even had it there and then during "cleanup" moved all my variables to the top of the method 🤦

ui/lib/core/addon/utils/client-count-utils.js
hellobontempo: I've gotten inconsistent data back...in some cases the `counts` and `new_clients` keys exist on the object but are `null`, in other cases the keys don't exist at all. Not sure if this answers your question?

ui/lib/core/addon/utils/client-count-utils.js
hellobontempo: I cleaned up the comment a bit. Do you think it still makes sense to map, as I'm only manipulating the data to construct and then return the new object `namespaces_by_key` below. Happy to refactor accordingly! Just want to make sure we're on the same page.

ui/lib/core/addon/utils/client-count-utils.js
hashishaw: In that case I _think_ `if (object?.counts)` will suffice, but a great opportunity to spot check with tests!

ui/app/components/clients/history.js
```diff
@@ -185,12 +185,19 @@ export default class History extends Component {
...
+      return this.filteredActivityByMonth;
+    } else {
+      return this.getActivityResponse?.byMonth;
+    }
   }
 
   get byMonthNewClients() {
-    return this.byMonthTotalClients.map((m) => m.new_clients);
+    if (this.byMonthActivityData) {
+      return this.byMonthActivityData.map((m) => m.new_clients);
```
hashishaw, line 198: `byMonthActivityData` could return null if `this.getActivityResponse` is undefined since we have optional chaining on line 192, so I'd either return a fallback empty array from `byMonthByActivityData` or do optional chaining here as well

ui/app/components/clients/history.js
```diff
@@ -201,45 +208,43 @@ export default class History extends Component {
...
   get totalUsageCounts() {
-    return this.selectedNamespace ? this.filteredActivity : this.getActivityResponse.total;
+    return this.selectedNamespace ? this.filteredActivityByNamespace : this.getActivityResponse.total;
   }
 
-  get newUsageCounts() {
-    return this.selectedNamespace
-      ? this.filteredNewClientAttribution
-      : this.byMonthTotalClients[0]?.new_clients;
+  // (object) single month new client data with total counts + array of namespace breakdown
```
hashishaw, line 216: 😍 the comments!

hashishaw (approved):
Just one small item. Otherwise LGTM

ui/app/components/clients/history.js
hellobontempo: good call, thank you!

PR #15285, opened 2022-05-04
Compare issuer certificates using cert, signature algo and signature fields

stevendpclark:
- Instead of comparing the strings of a certificate, instead leverage the Go Raw attribute within a parsed certificate to compare. The Raw attribute is a byte array of an ASN.1 DER containing the cert, signature algo and signature.
- Rework a bit of the importIssuers function as well to fail checks on the inbound issuer earlier as well as load keys/issuers just before we need them
- The first commit just refactors some of the PKI test helper methods into a dedicated file called test_helpers.go

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -9,16 +9,12 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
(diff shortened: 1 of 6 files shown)


cipherboy (approved):
Looks good and we've been using it for a bit. :-) Thanks!

PR #15286, opened 2022-05-04
base sys/license/status mock times based on current timestamp

ccapurso:


ui/mirage/handlers/clients.js
```diff
@@ -1,4 +1,4 @@
-import { formatISO, isAfter, isBefore, sub, isSameMonth, startOfMonth } from 'date-fns';
+import { addDays, formatISO, formatRFC3339, isAfter, isBefore, sub, isSameMonth, startOfMonth } from 'date-fns';
 import { parseAPITimestamp } from 'core/utils/date-formatters';
 const MOCK_MONTHLY_DATA = [
```
hellobontempo, line 3: Ah, so this mirage file is actually not used for running tests at the moment. (That's the end goal, but right now it's just used for local development)

ccapurso: Ah, I my recollection seems to be incorrect and that I modified the wrong `sys/license/status` mock. Whoops, thanks for calling that out!

ui/mirage/handlers/clients.js
```diff
@@ -814,11 +814,13 @@ export default function (server) {
...
 
   server.get('sys/license/status', function () {
+    const startTime = new Date();
+
     return {
       data: {
         autoloading_used: true,
         autoloaded: {
-          expiration_time: '2022-05-17T23:59:59.999Z',
+          expiration_time: formatRFC3339(addDays(startTime, 365))
```
hellobontempo, line 823: Each of these lines need commas at the end. I'm curious if tests passed because this endpoint failed to build and so it didn't intercept the tests' network request...just spitballin'

hellobontempo (approved):
LGTM! Thanks for tackling 🚀

PR #15215, opened 2022-04-28
Upgrade CircleCI machine image

hghaf099:
Addresses https://hashicorp.atlassian.net/browse/VAULT-5370
Update circleci jobs using Ubuntu 14.04/16.04 as stated in CircleCI post [here](https://circleci.com/blog/ubuntu-14-16-image-deprecation/?mkt_tok=NDg1LVpNSC02MjYAAAGDLWMxR7BjchCqBFmLcj0YNCo6xEPnR5QCU4QWXC4SEf1-8i0oDF4e_BhNQkVC9V6BiHzKwMXNTz14QrTlPUqjFJVCByEpp4bJXNzOfSnqlmLO4w)

PR #15288, opened 2022-05-04
GOPATH is being set incorrectly in latest CircleCI machine image

swenson:
For some reason, GOPATH is being overridden in the `go mod download`
step after the latest machine image update in #15215.

This causes all of the modules to be downloaded to the
`/home/circleci/.go_workspace` cache instead of `/home/circleci/go` like
we require for the build (which will otherwise fail since we build with
`GOPROXY=off`).

Without this fix, the build will start to fail once the existing cache
is no longer used (after the root `go.mod`) is updated.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,7 +127,7 @@ jobs:
     - checkout
     - restore_cache:
         keys:
-        - v1.4-{{checksum "go.sum"}}-{{checksum "sdk/go.sum"}}-{{checksum "api/go.sum"}}
+        - v1.5-{{checksum "go.sum"}}-{{checksum "sdk/go.sum"}}-{{checksum "api/go.sum"}}
         name: Restore exact go modules cache
     - attach_workspace:
         at: .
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -127,7 +127,7 @@ jobs:
     - checkout
     - restore_cache:
         keys:
-        - v1.4-{{checksum "go.sum"}}-{{checksum "sdk/go.sum"}}-{{checksum "api/go.sum"}}
+        - v1.5-{{checksum "go.sum"}}-{{checksum "sdk/go.sum"}}-{{checksum "api/go.sum"}}
```
tvoran, line 130: Why does the cache version need to change?

swenson: If any caches were uploaded in the last few days without this `GOPATH` fix, they would be missing modules, so might cause future builds to break.

tvoran: Oh right, the cache are still being stored, they're just empty.

.circleci/config/commands/@caches.yml
```diff
@@ -40,6 +40,8 @@ refresh_go_mod_cache:
     - run:
         name: go mod download
         command: |
+          # set GOPATH explicitly to download to the right cache
+          export GOPATH=/home/circleci/go
```
benashz, line 44: I wonder if we could rely on `$HOME` rather than hardcoding the path; also ensure the path exists?

```suggestion
          export GOPATH="${HOME}/go"
          mkdir -p "$GOPATH"
```

benashz (approved):
Minor suggestion, otherwise +1

tvoran (approved):

calvn (approved):

.circleci/config/commands/@caches.yml
swenson: If it doesn't exist, we're going to have a bad time. :) But I'll add the `$HOME` bit.

swenson:
Thanks!

PR #15289, opened 2022-05-04
Prevent revoking all issuer certificates

cipherboy:
We prevent the revocation of issuer certificates via the `/revoke` endpoint. This is because the endpoint lacks enough specificity in the case where multiple issuers have the same serial number. 

Along the way, we handle two cases where we failed to address performance replication concerns. 

We also add a warning when revoking a certificate that's already expired. :-) 

---

My inclination is not to merge this until after #15277 has merged.


stevendpclark (approved):
lgtm!

sgmiller (approved):

PR #15293, opened 2022-05-04
Update deps for consul-template 0.29.0

swenson:
This requires bumping https://github.com/mitchellh/go-testing-interface.
For this new version, we have to create a wrapper to convert
the stdlib `testing.TB` interface to the
`mitchellh/go-testing-interface` `T` interface, since it uses
`Parallel()` now, which is not supported by `testing.TB`. This had to be
added to a new package, `benchhelpers`, to avoid a circular dependency
in `testhelpers`.

We also have to *unbump* https://github.com/armon/go-metrics since
updating it breaks our usage of
https://github.com/google/go-metrics-stackdriver

I verified that the new `pkiCert` template function works with agent
injection using annotations like:

```yaml
vault.hashicorp.com/agent-inject-secret-sample.crt: "pki/issue/example-dot-com"
vault.hashicorp.com/agent-inject-template-sample.crt: |
  {{ pkiCert "pki/issue/example-dot-com" "common_name=foo.example.com" "ttl=1h" }}
```


helper/benchhelpers/benchhelpers.go
```diff
@@ -0,0 +1,19 @@
+package benchhelpers
+
+import (
+	"testing"
+
+	testing2 "github.com/mitchellh/go-testing-interface"
```
tomhjp, line 6:
```suggestion
	testinginterface "github.com/mitchellh/go-testing-interface"
```

tomhjp (commented):
LGTM, but I just want to make sure I understand before hitting the green button. I also don't see any "unbump" for go-metrics, is that comment out of date now?

swenson:
When bumping `consul-template`, you also get a transitive bump of `go-metrics`. So the "unbump" refers to me having to manually roll that back in `go.mod`.

helper/benchhelpers/benchhelpers.go
swenson: :+1: I initially had something like this, but GoLand "helpfully" changed it when I refactored something. I'll switch it back.

tomhjp (approved):
Thanks for the explainer, LGTM!

swenson:
Thanks!

PR #10607, opened 2020-12-20
 Add suport for Consul tokens with roles and instances (take 3)

kula:
This adds support to the Consul secret engine to support generating tokens with Consul roles (available since 1.5), Consul Service Identities (available since 1.5) and Consul Node Identities (available since 1.8.1).

Includes updated test suite and updated documentation.

A re-do of https://github.com/hashicorp/vault/pull/10605 as I believe I did something wrong with module vendoring there.

PR #15295, opened 2022-05-04
Add suport for Consul service-identities and node-identities

robmonte:
This PR is a rebase of https://github.com/hashicorp/vault/pull/10607

This adds Consul secrets engine support for node-identities and service-identities.

When writing a Vault role, you can specify node-identity or service-identity like so:
`vault write consul/roles/serv-id-role service_identities="webserv-1:dc1"`
or
`vault write consul/roles/node-id-role node_identities="client-1:dc1"`

You can specify multiple of each identity on a Vault role. Service-identities are semicolon-separated because you can have a list of datacenters, while node-identities are just comma-separated.
`vault write consul/roles/serv-id-role service_identities="webserv-1:dc1,dc2,dc3;webserv-2;webserv-3:dc1"`
or
`vault write consul/roles/node-id-role node_identities="client-1:dc1,client-2;client-3:dc2"`

Then to obtain a token from Consul, you simply give the role like normal:
`vault read consul/creds/serv-id-role`
`vault read consul/creds/node-id-role`

### DIFF
--- a/builtin/logical/consul/path_roles.go
+++ b/builtin/logical/consul/path_roles.go
@@ -91,13 +91,13 @@ will be created within. Defaults to 'default'. Available in Consul 1.11 and abov
 			},
 
 			"service_identities": {
-				Type: framework.TypeString,
+				Type: framework.TypeStringSlice,
 				Description: `List of Service Identities to attach to the
 token, separated by semicolons. Available in Consul 1.5 or above.`,
 			},
 
 			"node_identities": {
-				Type: framework.TypeCommaStringSlice,
+				Type: framework.TypeStringSlice,
 				Description: `List of Node Identities to attach to the
 token. Available in Consul 1.8.1 or above.`,
 			},
@@ -176,7 +176,7 @@ func (b *backend) pathRolesWrite(ctx context.Context, req *logical.Request, d *f
 	policy := d.Get("policy").(string)
 	policies := d.Get("policies").([]string)
 	roles := d.Get("consul_roles").([]string)
-	serviceIdentities := d.Get("service_identities").(string)
+	serviceIdentities := d.Get("service_identities").([]string)
 	nodeIdentities := d.Get("node_identities").([]string)
 
 	switch tokenType {
@@ -254,7 +254,7 @@ type roleConfig struct {
 	Policy            string        `json:"policy"`
 	Policies          []string      `json:"policies"`
 	ConsulRoles       []string      `json:"consul_roles"`
-	ServiceIdentities string        `json:"service_identities"`
+	ServiceIdentities []string      `json:"service_identities"`
 	NodeIdentities    []string      `json:"node_identities"`
 	TTL               time.Duration `json:"lease"`
 	MaxTTL            time.Duration `json:"max_ttl"`
--- a/builtin/logical/consul/path_roles.go
+++ b/builtin/logical/consul/path_roles.go
@@ -181,11 +181,8 @@ func (b *backend) pathRolesWrite(ctx context.Context, req *logical.Request, d *f
 
 	switch tokenType {
 	case "client":
-		if policy == "" &&
-			len(policies) == 0 &&
-			len(roles) == 0 &&
-			len(serviceIdentities) == 0 &&
-			len(nodeIdentities) == 0 {
+		if policy == "" && len(policies) == 0 && len(roles) == 0 &&
+			len(serviceIdentities) == 0 && len(nodeIdentities) == 0 {
 			return logical.ErrorResponse(
 				"Use either a policy document, a list of policies or roles, or a set of service or node identities, depending on your Consul version"), nil
 		}
--- a/builtin/logical/consul/path_roles.go
+++ b/builtin/logical/consul/path_roles.go
@@ -91,7 +91,7 @@ will be created within. Defaults to 'default'. Available in Consul 1.11 and abov
 			},
 
 			"service_identities": {
-				Type: framework.TypeSemicolonStringSlice,
+				Type: framework.TypeString,
 				Description: `List of Service Identities to attach to the
 token, separated by semicolons. Available in Consul 1.5 or above.`,
 			},
@@ -175,10 +175,9 @@ func (b *backend) pathRolesWrite(ctx context.Context, req *logical.Request, d *f
 	tokenType := d.Get("token_type").(string)
 	policy := d.Get("policy").(string)
 	policies := d.Get("policies").([]string)
-	serviceIdentities := d.Get("service_identities").([]string)
-	nodeIdentities := d.Get("node_identities").([]string)
-	local := d.Get("local").(bool)
 	roles := d.Get("consul_roles").([]string)
+	serviceIdentities := d.Get("service_identities").(string)
+	nodeIdentities := d.Get("node_identities").([]string)
 
 	switch tokenType {
 	case "client":
@@ -258,7 +257,7 @@ type roleConfig struct {
 	Policy            string        `json:"policy"`
 	Policies          []string      `json:"policies"`
 	ConsulRoles       []string      `json:"consul_roles"`
-	ServiceIdentities []string      `json:"service_identities"`
+	ServiceIdentities string        `json:"service_identities"`
 	NodeIdentities    []string      `json:"node_identities"`
 	TTL               time.Duration `json:"lease"`
 	MaxTTL            time.Duration `json:"max_ttl"`
--- a/builtin/logical/consul/path_roles.go
+++ b/builtin/logical/consul/path_roles.go
@@ -188,7 +188,7 @@ func (b *backend) pathRolesWrite(ctx context.Context, req *logical.Request, d *f
 			len(serviceIdentities) == 0 &&
 			len(nodeIdentities) == 0 {
 			return logical.ErrorResponse(
-				"Use either a policy document, a list of policies or roles, or a set of service identities and/or node identities, depending on your Consul version"), nil
+				"Use either a policy document, a list of policies or roles, or a set of service or node identities, depending on your Consul version"), nil
 		}
 	case "management":
 	default:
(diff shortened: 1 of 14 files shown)


builtin/logical/consul/path_roles.go
```diff
@@ -89,6 +89,18 @@ created within. Defaults to 'default'. Available in Consul 1.7 and above.`,
 				Description: `Indicates which admin partition that the token
 will be created within. Defaults to 'default'. Available in Consul 1.11 and above.`,
 			},
+
+			"service_identities": {
+				Type: framework.TypeString,
```
swenson, line 94: Is it worth adding a `TypeSemicolonStringSlice`?

builtin/logical/consul/path_token.go
```diff
@@ -103,13 +106,41 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
...
+	if err != nil {
+		return nil, err
+	}
+	for _, serviceIdentity := range parsedServiceIdentities {
+		entry := &api.ACLServiceIdentity{}
+		components := strings.SplitN(serviceIdentity, ":", 2)
+		entry.ServiceName = components[0]
+		if len(components) == 2 {
+			entry.Datacenters = strings.Split(components[1], ",")
+		}
```
swenson, line 120: nit: I think this could be simplified to only call `Split` once. Maybe:

```suggestion
		components := strings.Split(serviceIdentity, ":")
		entry.ServiceName = components[0]
		if len(components) > 1 {
			entry.Datacenters = components[1:]
		}
```

builtin/logical/consul/path_token.go
```diff
@@ -132,3 +163,25 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
...
 }
+
+// Parses the provided string-like value as a semicolon-separated list of values.
+func parseSemicolonStringSlice(in interface{}) ([]string, error) {
+	rawString, ok := in.(string)
+	if ok && rawString == "" {
+		return []string{}, nil
+	}
+	var result []string
+	config := &mapstructure.DecoderConfig{
```
swenson, line 174: This seems like a lot of machinery for `strings.Split(rawString, ";")`? Or am I missing something?

robmonte: So, the original PR actually had implemented a TypeSemicolonStringSlice (as you mentioned in another comment). Jim thought it might be better to move the logic out of parseutil and just have a function here because it's the only place in the code that currently needs semicolon-delimited input.

This is just lifting the logic from that into here. I agree this can be cut down significantly.

builtin/logical/consul/path_token.go
robmonte: I'm trying to think if this creates the same result.
If the provided value is "serv1:dc1,dc2,dc3" then the result is
```
ServiceName: "serv1"
Datacenters: [dc1 dc2 dc3]
```

I think this would result in Datacenters still being a string containing commas, or am I mistaken?

swenson: Oh, I'm mistaken. I thought it was splitting twice on ":", but now that I read it, there are two separate split characters. My bad!

builtin/logical/consul/path_token.go
```diff
@@ -103,13 +106,41 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
 		})
 	}
 
+	aclServiceIdentities := []*api.ACLServiceIdentity{}
+	parsedServiceIdentities, err := parseSemicolonStringSlice(roleConfigData.ServiceIdentities)
```
calvn, line 110: If we're going with a plugin-specific function that parses the service_identities string value, would it be worth to parse it all the way such that what gets returned is `[]*api.ACLNodeIdentity{}`? That we we avoid doing half of the parsing there and the other half down below.

builtin/logical/consul/path_token.go
```diff
@@ -103,13 +106,41 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
...
+	if err != nil {
+		return nil, err
+	}
+	for _, serviceIdentity := range parsedServiceIdentities {
+		entry := &api.ACLServiceIdentity{}
+		components := strings.SplitN(serviceIdentity, ":", 2)
```
calvn, line 116: Why is `strings.SplitN` 2 here for `n`?

builtin/logical/consul/path_token.go
calvn: +1 on using `strings.Split` here since we're not doing anything particularly fancy like decoding to a custom struct.

builtin/logical/consul/path_roles.go
```diff
@@ -157,17 +176,22 @@ func (b *backend) pathRolesWrite(ctx context.Context, req *logical.Request, d *f
 	policy := d.Get("policy").(string)
 	policies := d.Get("policies").([]string)
 	roles := d.Get("consul_roles").([]string)
+	serviceIdentities := d.Get("service_identities").(string)
```
calvn, line 179: Should we do any sort of validation in here, especially for `service_identities` since the input is a just string?

builtin/logical/consul/path_token.go
robmonte: Good idea, I decided to move both node identity and service identity parsing out into functions because they're by far the bulkiest part of that code section.

builtin/logical/consul/path_token.go
robmonte: It made sense to keep this at first, because it would ensure that the string is only ever split into 2 parts on the first colon. After thinking about it more, I believe there there should only ever be one colon per serviceIdentity in this section of code when used correctly, so Split vs SplitN should have no difference.

Given that, I'd rather change it to regular Split for code clarity.

builtin/logical/consul/path_token.go
robmonte: Please let me know what you think of the update

builtin/logical/consul/path_roles.go
robmonte: I'm not quite sure how you would validate the entry for these. Since the name of a service or node can be anything and providing datacenters is optional, the presence of commas or semicolons is optional. Is there another form of validation I'm missing?

builtin/logical/consul/path_roles.go
austingebauer, line 94: Using `TypeString` with semicolon delimited service identities seems like it will work, but an alternative to consider would be `TypeStringSlice`. The comma splitting doesn't happen automatically by the framework using the type. 

Supplying the service identities would look something like:

```go
vault write consul/roles/serv-id-role \
    service_identities="webserv-1:dc1,dc2,dc3" \
    service_identities="webserv-2" \
    service_identities="webserv-3:dc1"
```

It might simplify the parsing a bit too. Let me know your thoughts 👍

builtin/logical/consul/path_token.go
```diff
@@ -132,3 +138,39 @@ func (b *backend) pathTokenRead(ctx context.Context, req *logical.Request, d *fr
 
 	return s, nil
 }
+
+func parseServiceIdentities(data string) []*api.ACLServiceIdentity {
```
austingebauer, line 142: It'd be nice to get some tests on these parsing functions to ensure they work as expected given various forms of input strings.

robmonte: Do you have any suggestions for what kind of approach you're thinking of for this?

austingebauer: Discussed outside of this PR -- was thinking something like a table style unit test.

builtin/logical/consul/path_roles.go
robmonte: This was a great idea as it eliminated the second string split.

Additionally, this matches the user experience of Consul itself, where you can specify `-service-identity=...` multiple times.

builtin/logical/consul/path_token.go
robmonte: This has now changed quite a bit by using Austin's suggestion to receive the two parameters as TypeStringSlice. This lets a user input the parameter multiple times and automatically builds the slice out of those inputs to eliminate the second strings.Split, thus simplifying the parsing logic.

calvn (approved):
Looks great! Thanks for adding tests around the changes, and good call on using `TypeStringSlice` instead -- definitely looks much cleaner.

website/content/api-docs/secret/consul.mdx
```diff
@@ -125,6 +137,10 @@ To create a client token within a particular Consul admin partition:
 - `consul_roles` `(list: <role or roles>)` – The list of Consul roles to assign to the
   generated token. Either `policies` or `consul_roles` are required for Consul 1.5 and above.
 
+- `service_identities` `(list: <service identity or identities>)` - The list of
+  service identities to assign to the generated token. This may be a semicolon-separated list to
```
austingebauer, line 141: I think this should be updated now that they're not semicolon-separated?

website/content/api-docs/secret/consul.mdx
```diff
@@ -163,6 +179,14 @@ To create a client token with defined roles:
 }
 ```
 
+To create a client token with service identities attached:
+
+```json
+{
+  "service_identities": "myservice-1:dc1,dc2;myservice-2:dc1"
```
austingebauer, line 186: Same here -- I think this will look different now.

builtin/logical/consul/path_token_test.go
```diff
@@ -0,0 +1,104 @@
+package consul
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/hashicorp/consul/api"
+)
+
+func Test_parseServiceIdentities(t *testing.T) {
```
austingebauer, line 10: Nice 👍

austingebauer (approved):
Looks good to me. Nice job, @robmonte 👍

website/content/api-docs/secret/consul.mdx
robmonte: You're right, I forgot about the doc changes here. I'm going to add this to my docs-specific task rather than update this PR again. (Same for other doc file). Thanks for catching this!!

PR #15315, opened 2022-05-06
Benchmark chain building

cipherboy:
Per request in #15277... :-)

I ended up having to rewrite the code to use `createBackendWithStorage` rather than creating clusters as it was way too slow.

Results:

```
$ go test -v -run=BenchmarkChainBuilding -bench=. github.com/hashicorp/vault/builtin/logical/pki -cpuprofile profile.out
goos: linux
goarch: amd64
pkg: github.com/hashicorp/vault/builtin/logical/pki
cpu: Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
BenchmarkChainBuilding
BenchmarkChainBuilding/test-case-0
BenchmarkChainBuilding/test-case-0-16         	     596	   2122381 ns/op
BenchmarkChainBuilding/test-case-1
BenchmarkChainBuilding/test-case-1-16         	    1105	   1099540 ns/op
BenchmarkChainBuilding/test-case-2
BenchmarkChainBuilding/test-case-2-16         	     495	   2341228 ns/op
BenchmarkChainBuilding/test-case-3
BenchmarkChainBuilding/test-case-3-16         	     480	   2469348 ns/op
BenchmarkChainBuilding/test-case-4
BenchmarkChainBuilding/test-case-4-16         	    1536	    758453 ns/op
BenchmarkChainBuilding/test-case-5
BenchmarkChainBuilding/test-case-5-16         	   49120	     24237 ns/op
BenchmarkChainBuilding/test-case-6
BenchmarkChainBuilding/test-case-6-16         	   10000	    125839 ns/op
BenchmarkChainBuilding/test-case-7
BenchmarkChainBuilding/test-case-7-16         	    5127	    236430 ns/op
BenchmarkChainBuilding/test-case-8
BenchmarkChainBuilding/test-case-8-16         	    2370	    507779 ns/op
```

Test cases are from the chain building tests, the call to `rebuildIssuersChains` is just benchmarked afterwards.

### DIFF
--- a/builtin/logical/pki/chain_test.go
+++ b/builtin/logical/pki/chain_test.go
@@ -1,6 +1,7 @@
 package pki
 
 import (
+	"context"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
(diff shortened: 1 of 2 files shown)


sgmiller (approved):

builtin/logical/pki/backend_test.go
```diff
@@ -5119,7 +5119,7 @@ var (
 	edCACert  string
 )
 
-func mountPKIEndpoint(t *testing.T, client *api.Client, path string) {
+func mountPKIEndpoint(t testing.TB, client *api.Client, path string) {
```
stevendpclark, line 5122: Why did we switch all the t pointer params to value ones?

stevendpclark: Oh and please fix fmt errors.

cipherboy: [`testing.TB` is the interface](https://pkg.go.dev/testing#TB) of things both [`testing.T` the struct](https://pkg.go.dev/testing#T) and [`testing.B` the struct](https://pkg.go.dev/testing#B) have. We directly pass `t` (the `*testing.T`) type into `mountPKIEndpoint` as it satisfies the `testing.TB` interface, so we don't change the _passed_ type, just the _accepting_ type to allow this function to be used by `testing.B` benchmark functions.

stevendpclark: Ah! that makes sense, sorry should have looked that up myself.

stevendpclark (approved):

builtin/logical/pki/backend_test.go
cipherboy: Nah that's fine :-) When we were just testing, it made sense. But I had a `testing.B` and nothing wanted to work with the `*testing.T` type sigs until I found `testing.TB`. Then I swapped it out and hoped nothing broke... :D

PR #15322, opened 2022-05-06
Use Get lookup within PKI resolveXXX functions instead of list iterations

stevendpclark:
- Leverage a Get lookup operation to see if our reference field is a UUID
   instead of listing all key/issuers and iterating over the list.
 - This should be faster and we get a cached lookup possibly if it was a
   UUID entry that we previously loaded.
 - Address some small feedback about migration wording as well.

### DIFF
--- a/builtin/logical/pki/storage.go
+++ b/builtin/logical/pki/storage.go
@@ -24,6 +24,9 @@ const (
 	legacyMigrationBundleLogKey = "config/legacyMigrationBundleLog"
 	legacyCertBundlePath        = "config/ca_bundle"
 	legacyCRLPath               = "crl"
+
+	// Used as a quick sanity check for a reference id lookups...
+	uuidLength = 36
 )
 
 type keyID string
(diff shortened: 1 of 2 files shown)


builtin/logical/pki/storage.go
```diff
@@ -412,19 +415,22 @@ func resolveKeyReference(ctx context.Context, s logical.Storage, reference strin
...
+		if err != nil {
+			return keyID("key-read"), err
+		}
+		if entry != nil {
+			return keyID(reference), nil
 		}
 	}
 
 	// ... than to pull all keys from storage.
+	keys, err := listKeys(ctx, s)
```
cipherboy, line 430: Seeing this part in particular makes me think its not actually that great of an idea @sgmiller. We maybe save the `id` case slightly, but for a chunk of these calls, we'll likely have names which still incurs the list. We're just moving around the list and perhaps most of the time deferring it until after a failed `get`. 

Are failed `Get`'s cached?

I'm thinking a proper issuer caching mechanism would be better than this fix, at least in my opinion.

PR #15329, opened 2022-05-08
Vault documentation: added disable parameter to seal stanza-related doc pages

taoism4504:
Per [Asana](https://app.asana.com/0/1200328878173626/1202157875971193) and my conversation w/ Kevin, a number of pages mentioning the seal stanza have been updated to include the missing `disable` parameter. I modified the wording based on Kevin's initial suggestion so please let me know if the description for `disable` needs to be modified. Thanks.

### DIFF
--- a/website/content/docs/concepts/seal.mdx
+++ b/website/content/docs/concepts/seal.mdx
@@ -81,7 +81,7 @@ access to the root key shards.
 
 ## Auto Unseal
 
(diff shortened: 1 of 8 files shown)


kkredit (commented):
This looks good to me :) 

The Asana ticket also mentions
>Furthermore, our general seal migration documentation (https://www.vaultproject.io/docs/concepts/seal#seal-migration) has some details about specifically what kinds of migrations are supported. We have a complicated bug that's preventing us from supporting AWSKMS <→ AWSKMS migrations currently, and we should flag that as an unsupported migration.

Is that going to be addressed as well? (perhaps in a separate PR?)

mladlow (approved):
These changes look good, thank you @taoism4504.

I'm fine with the note you suggested in a comment, I think it would be good to add that note though.

PR #15336, opened 2022-05-09
Return names for leaf_not_after_behavior responses

cipherboy:
Previously we'd return the raw enum value, which the entity accessing
the API wouldn't have any easy way of translating back into string
values. Return the string value directly instead.

### DIFF
--- a/sdk/helper/certutil/types.go
+++ b/sdk/helper/certutil/types.go
@@ -689,6 +689,20 @@ const (
 	PermitNotAfterBehavior
 )
 
+var notAfterBehaviorNames = map[NotAfterBehavior]string{
+	ErrNotAfterBehavior:      "err",
+	TruncateNotAfterBehavior: "truncate",
+	PermitNotAfterBehavior:   "permit",
+}
+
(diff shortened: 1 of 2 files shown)


sdk/helper/certutil/types.go
```diff
@@ -689,6 +689,20 @@ const (
 	PermitNotAfterBehavior
 )
 
+func (n NotAfterBehavior) String() string {
+	names := map[NotAfterBehavior]string {
```
stevendpclark, line 693: nit: Should this be pulled out into a "const" var (I know an actual const isn't possible).

stevendpclark (approved):
Small nit, but 👍

sdk/helper/certutil/types.go
cipherboy: We could refactor it when we need it, since this is in the SDK I'd rather not expose it until we need to.

cipherboy: Ah, I see on godbolt that could impact the performance, so I'll move it to a private thingy.

PR #15342, opened 2022-05-09
add mount accessor to audit request and responses

swayne275:


audit/format.go
```diff
@@ -293,14 +295,15 @@ func (f *AuditFormatter) FormatResponse(ctx context.Context, w io.Writer, config
 		},
 
 		Response: &AuditResponse{
-			MountType: req.MountType,
-			Auth:      respAuth,
```
pmmukh, line 297: only question i have is, do we need this field in both the request and response blocks of the response audit log ? Since it feels like having it in just one place should be good enough, but I do see we do this duplication for other fields like mountType and headers, so maybe there's a rationale i'm missing here.

pmmukh (approved):
just one q, looks solid to me!

ccapurso (approved):
Looks good to me! The one suggestion that I have is to provide some basic test coverage.

audit/format.go
sgmiller: Yes, we duplicate these because there may be cases where a request occurs but the response doesn't due to error or what have you.

sgmiller (approved):

audit/format.go
pmmukh: aah, gotcha!

PR #15344, opened 2022-05-10
Remove spurious fmt.Printf, including one of a key

swenson:


helper/dhutil/dhutil.go
```diff
@@ -90,7 +90,6 @@ func DeriveSharedKey(secret, ourPublic, theirPublic []byte) ([]byte, error) {
 	if n != 32 {
 		return nil, errors.New("short read from hkdf")
 	}
-	fmt.Printf("Key: %s\n", hex.EncodeToString(key[:]))
```
tvoran, line 93: Looks like this was the only thing using "encoding/hex" so the import can be removed?

swenson: Ah, thanks. That's what I get for editing in a different editor than usual. :)

tvoran (approved):
The Printf removals look fine to me, though you may want to get someone from core to sign off on the semgrep addition.

tools/semgrep/ci/fmt-printf.yml
```diff
@@ -0,0 +1,11 @@
+rules:
+  - id: fmt.Printf
+    languages: [go]
+    message: fmt.Printf is forbidden outside of cmd and test files
+    pattern: fmt.Printf
+    severity: ERROR
+    paths:
+      exclude:
+        - "*test*.go"
```
raskchanky, line 9: Don't go test files have to end in `_test.go`? Seems like this will catch any file with `test` in the name, rather than actual test files.

tools/semgrep/ci/fmt-printf.yml
```diff
@@ -0,0 +1,11 @@
...
+        - "cmd/*.go"
+        - "cmd/**/*.go"
```
raskchanky, line 11: Should this be `command` instead of `cmd`? I don't see a `cmd` directory.

raskchanky: Oh, I see a bunch under `builtin`, but I think these paths apply from the root, no?

swenson: These don't apply from the root.

For example, if I remove them:

```
$ make semgrep-ci
semgrep --error --include '*.go' --exclude 'vendor' -f tools/semgrep/ci .
...
  serviceregistration/kubernetes/client/cmd/kubeclient/main.go
     tools.semgrep.ci.fmt.Printf
        fmt.Printf is forbidden outside of cmd and test files

         68┆ fmt.Printf("pod: %s\n", b)
```

tools/semgrep/ci/fmt-printf.yml
swenson: :+1:

tools/semgrep/ci/fmt-printf.yml
```diff
@@ -0,0 +1,11 @@
+rules:
+  - id: fmt.Printf
+    languages: [go]
+    message: fmt.Printf is forbidden outside of cmd and test files
+    pattern: fmt.Printf
+    severity: ERROR
+    paths:
+      exclude:
+        - "*_test*.go"
```
raskchanky, line 9: Is the final asterisk necessary?

tools/semgrep/ci/fmt-printf.yml
raskchanky: Ah, TIL. So `cmd` is on purpose then. 👍

tools/semgrep/ci/fmt-printf.yml
```diff
@@ -0,0 +1,11 @@
+rules:
+  - id: fmt.Printf
+    languages: [go]
+    message: fmt.Printf is forbidden outside of cmd and test files
+    pattern: fmt.Printf
```
briankassouf, line 5: Might as well throw in a `fmt.Println` check too!

swenson: :+1:

tools/semgrep/ci/fmt-printf.yml
swenson: Oh, probably not. I was just being extra cautious about catching test files. I'll test and remove it.

swenson: Turns out, it is necessary. There are a few extra test files:

```
  builtin/credential/aws/pkcs7/verify_test_dsa.go
     tools.semgrep.ci.fmt.Printf
        fmt.Printf/Println is forbidden outside of cmd and test files

        111┆ fmt.Printf("%s\n", pemSignature)


  command/server/config_test_helpers.go
     tools.semgrep.ci.fmt.Printf
        fmt.Printf/Println is forbidden outside of cmd and test files

        323┆ fmt.Printf("\n config.Entropy: %#v", config.Entropy)
```

raskchanky: I don't think tests in those files are actually being run though. For example, the first test at the top of `verify_test_dsa.go` is called `TestVerifyEC2`.

CleanShot 2022-05-18 at 15 13 38

swenson: Possibly, but I think that is out of scope for this PR. Not sure why those tests are effectively disabled. Maybe they are only run manually?

raskchanky: I'm sorry, I've done a poor job of explaining my point. What I'm trying to say is that I think `"*_test.go"` is adequate for the exclude pattern, because `go test` ignores files that don't end in `_test.go`. So `verify_test_dsa.go` is not considered a test file by the Go test runner, as my screenshot shows. My concern with `*_test*.go` is having it exclude files that we do want checked, because they're not actual Go test files, e.g. `litmus_test_client.go`, etc.

swenson: Okay. I'll go ahead and clean up those `fmt.Print*` statements from those files as well then.

raskchanky (approved):

swenson:
Thanks!

PR #15352, opened 2022-05-10
Query and Precompute Non-Contiguous Segments in the Activity Log

HridoyRoy:
This PR:

1. Changes the query get logic from an algorithm to find the most recent contiguous chunk of segments to finding the closest sub-interval to the user-specified query interval. 
2. Changes the precompute logic to precompute on a minimum of 12 prior segments of activity log data if the number of contiguous segments in storage is less than 12.


vault/activity/query.go
```diff
@@ -134,6 +134,30 @@ func (s *PrecomputedQueryStore) listEndTimes(ctx context.Context, startTime time
...
+	maxEndTime := time.Time{}
+	for _, raw := range rawEndTimes {
+		val, err := strconv.ParseInt(raw, 10, 64)
+		if err != nil {
+			s.logger.Warn("could not parse precomputed query end time", "key", raw)
+			continue
+		}
+		endTime := time.Unix(val, 0).UTC()
+		s.logger.Trace("end time in consideration is", "end time", endTime, "end time bound", endTimeBound)
+		if endTime.After(maxEndTime) && (endTime.Before(endTimeBound) || endTime.Equal(endTimeBound)) {
```
ccapurso, line 152: We can probably simplify this check to `endTime.After(maxEndTime) && !endTime.After(endTimeBound)`. It might be a bit clearer to read and removes the potential need for 3 comparisons. What do you think?

vault/activity_log_test.go
```diff
@@ -2776,6 +2776,186 @@ func TestActivityLog_Precompute(t *testing.T) {
 	}
 }
 
+// TestActivityLog_Precompute_SkipMonth will put two non-contiguous chunks od
```
ccapurso, line 2779:
```suggestion
// TestActivityLog_Precompute_SkipMonth will put two non-contiguous chunks of
```

vault/activity_log_test.go
```diff
@@ -2776,6 +2776,186 @@ func TestActivityLog_Precompute(t *testing.T) {
...
+			0,
+			entityRecords[30:45],
+		},
+	}
+
+	// Note that precomputedQuery worker doesn't filter
+	// for times <= the one it was asked to do. Is that a problem?
+	// Here, it means that we can't insert everything *first* and do multiple
+	// test cases, we have to write logs incrementally.
+	doInsert := func(i int) {
```
ccapurso, line 2832: Should we add `t.Helper()` to this func?

vault/activity/query.go
HridoyRoy: Sure, sounds good!

vault/activity_log_test.go
HridoyRoy: Sure!

vault/activity_log_test.go
```diff
@@ -2776,6 +2776,186 @@ func TestActivityLog_Precompute(t *testing.T) {
...
+		}
+		data, err := proto.Marshal(eal)
+		if err != nil {
+			t.Fatal(err)
+		}
+		path := fmt.Sprintf("%ventity/%v/%v", ActivityLogPrefix, segment.StartTime, segment.Segment)
+		WriteToStorage(t, core, path, data)
+	}
+
+	expectedCounts := []struct {
```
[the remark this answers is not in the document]
HridoyRoy: I'm not quite sure how to do this. Do we absolutely need this? If so, could you elaborate a little more on how I'd go about doing it?
Thanks!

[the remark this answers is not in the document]
ccapurso: Nope, we don't need to do it. I just thought it was potentially a way to add a bit of readability so that each test case was somewhat self-contained.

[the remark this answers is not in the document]
HridoyRoy: Great to hear, thanks!

vault/activity/query.go
```diff
@@ -196,42 +220,45 @@ func (s *PrecomputedQueryStore) Get(ctx context.Context, startTime, endTime time
...
+			s.logger.Warn("missing end times", "start time", testStartTime)
+			continue
+		}
+		s.logger.Trace("retrieved max end time from storage", "endTime", testEndTime)
+		diff := testEndTime.Sub(testStartTime)
+		if diff >= maxTimeDifference {
+			closestStartTime = testStartTime
+			closestEndTime = testEndTime
+			maxTimeDifference = diff
+			s.logger.Trace("updating closest times")
```
ccapurso, line 245: I might be breading this logic incorrectly but it does not necessarily seem like it aligns with the following:

> Changes the query get logic from an algorithm to find the most recent contiguous chunk of segments to finding the closest sub-interval to the user-specified query interval.

It seems like instead what is calculated is the _widest_ contiguous chunk. That may not necessarily be the most recent, correct?

HridoyRoy: That's the same as what I said in the above quote, right? The "widest" contiguous chunk within the user-specified start and end time is the closest sub-interval to what the user specified.

ccapurso: I was focusing on the "closest sub-interval to the user-specified query interval" part. To provide an example, say the user specified Jan-August and there are the following contiguous chunks, Jan-Feb, Feb-Mar, and Mar-May. Based on the logic, I _think_ that Mar-May would be selected. Maybe that's the expectation and I'm misunderstanding the intention of the change and if so, sorry for the confusion! I interpreted "closest sub-interval" as being either close to the start time or end time.

HridoyRoy: The sub-interval isn't contiguous. So in the example you gave, all 3 intervals Jan-Feb, Feb-Mar, and Mar-May, should be returned with skipped nil months. 

Now, if the time interval specified was over 12 months (say we queried for Jan 2021 - April 2022), and the largest contiguous chunk of data is over 12 months (Jan 2021 - Feb 2022, say), then we will just return that contiguous chunk. The reason for this is that we simply need an upper bound for the maximum non-contiguous precomputation, so I chose a year.

ccapurso: We met directly so that I could get a little more insight into this discussion. I initially did not account for the impact that the change from using `getMostRecentActivityLogSegment` to `getMostRecentNonContiguousActivityLogSegments` in `precomputedQueryWorker` had on `Get`. That discussion paired with the example above resolved my confusion. Thanks, @HridoyRoy!

HridoyRoy: No problem, thanks for digging into it!

ccapurso (approved):
Looks good, thanks for taking the time to answer my questions!

vault/activity/query.go
```diff
@@ -196,42 +220,45 @@ func (s *PrecomputedQueryStore) Get(ctx context.Context, startTime, endTime time
...
-	var actualEndTime time.Time
-	for _, t := range endTimes {
-		if timeutil.InRange(t, startTime, endTime) {
-			if actualEndTime.IsZero() || t.After(actualEndTime) {
-				actualEndTime = t
-			}
+	closestStartTime := time.Time{}
+	closestEndTime := time.Time{}
+	maxTimeDifference := time.Duration(0)
+	for i := len(filteredList) - 1; i >= 0; i-- {
```
swayne275, line 227: all nits

1) is there a reason why we're doing a reverse for loop? i think it might be a bit easier to read with a range loop

2) do you think it makes sense to break this off into its own separate function? it might make it a bit easier to test this logic, and it also might make it more understandable for future engineers

i only ask these things because activity log is growing fairly hard to dive into and understand for those less familiar

vault/activity_log.go
```diff
@@ -483,6 +483,23 @@ func (a *ActivityLog) getMostRecentActivityLogSegment(ctx context.Context) ([]ti
 	return timeutil.GetMostRecentContiguousMonths(logTimes), nil
 }
 
+// getMostRecentActivityLogSegment gets the times (in UTC) associated with the most recent
```
swayne275, line 486: love the UTC callout!

vault/activity_log.go
```diff
@@ -483,6 +483,23 @@ func (a *ActivityLog) getMostRecentActivityLogSegment(ctx context.Context) ([]ti
...
+// contiguous set of activity logs, sorted in decreasing order (latest to earliest)
```
swayne275, line 487: does `latest to earliest` == `most to least recent`?

swayne275 (commented):
it looks like this got merged while i was reviewing, but overall i think it looks good. here were the comments i was in the process of leaving

PR #15353, opened 2022-05-10
MFA method and enforcement list view

arnav28:
- Added new route for list views
- List mfa methods along with id, type and icon
- Added client side pagination to list views

Screen Shot 2022-05-10 at 12 11 08 PM


ui/app/adapters/mfa-method.js
```diff
@@ -3,24 +3,26 @@ import ApplicationAdapter from './application';
...
   namespace = 'v1';
 
-  urlForQuery(methodType) {
-    let baseUrl = this.buildURL() + '/identity/mfa/method';
-    if (methodType) {
-      return `${baseUrl}/${methodType}`;
-    }
-    return baseUrl;
+  pathForType() {
+    return 'identity/mfa/method';
```
zofskeez, line 7: :+1:

ui/app/adapters/mfa-method.js
```diff
@@ -3,24 +3,26 @@ import ApplicationAdapter from './application';
...
   }
 
-  queryRecord(type, id) {
-    return this.ajax(this.urlForQuery(type), 'POST', {
+  queryRecord(store, type, query) {
+    const { id } = query;
+    if (!id) {
+      return;
+    }
```
zofskeez, line 14: Do you think throwing an error would be better here if id is required in the query?

ui/app/router.js
```diff
@@ -59,9 +59,13 @@ Router.map(function () {
           this.route('section', { path: '/:section_name' });
         });
         this.route('mfa', function () {
-          // lookup
-          this.route('index', { path: '/' });
           this.route('configure', { path: '/landing' });
+          this.route('methods', { path: '/' });
+          this.route('method', { path: '/method/:method_id' }, function () {
+            this.route('index', { path: '/' });
```
zofskeez, line 65: Is it necessary to define an index route here?

zofskeez (commented):
Thanks! I left a few comments but looks good.

ui/app/adapters/mfa-method.js
arnav28: Good point, will update.

ui/app/router.js
arnav28: I don't think, I just kept as a placeholder for adding future routes.

zofskeez: Ok sounds good :+1:

PR #15354, opened 2022-05-10
docs: update dep table to include okta auth changes

calvn:
Update deprecation table to include deprecation notice for the `api_token` parameter in the okta auth method. The parameter will be change from being optional to being required in 1.12. This will allow us to drop the usage of the non-official SDK library in the auth method.


swenson (approved):
LGTM

austingebauer (approved):

PR #15355, opened 2022-05-10
Updating Okta MFA to use official SDK

chrishoffman:
The PR migrates the login MFA functionality to use the official Okta SDK to initiate and verify the MFA enforcement.


swayne275 (approved):

hghaf099:
This looks great! I just have a couple of questions one of which might be out of the scope of this work.

vault/login_mfa.go
```diff
@@ -1848,50 +1853,36 @@ func (c *Core) validateOkta(ctx context.Context, mConfig *mfa.Config, username s
...
+	var userFactor *okta.UserFactor
+	for _, factor := range factors {
+		if factor.IsUserFactorInstance() {
+			userFactor = factor.(*okta.UserFactor)
+			if userFactor.FactorType == "push" {
+				factorFound = true
+				break
+			}
+		} else {
+			return errors.New("factor did not return user factor")
```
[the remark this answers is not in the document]
chrishoffman: Nice catch. You are correct that this is probably the wrong thing to do here. I'll remove.

vault/login_mfa.go
```diff
@@ -1848,50 +1850,34 @@ func (c *Core) validateOkta(ctx context.Context, mConfig *mfa.Config, username s
 		return fmt.Errorf("more than one user found for e-mail address")
 	}
 
-	user := &users[0]
+	user := users[0]
 
-	_, err = client.Users.PopulateMFAFactors(user)
+	factors, _, err := client.UserFactor.ListFactors(ctx, user.Id)
```
calvn, line 1855: This might be out of scope as well, but should we check for MFA enrollment status like we do on the Okta Auth Method? https://github.com/hashicorp/vault/blob/50afd00b0b92aad4282edcfeeea7d3aec8a64b0d/builtin/credential/okta/backend.go#L152

vault/login_mfa.go
```diff
@@ -1848,50 +1850,34 @@ func (c *Core) validateOkta(ctx context.Context, mConfig *mfa.Config, username s
...
-	}
-
-	req, err := client.NewRequest("POST", fmt.Sprintf("users/%s/factors/%s/verify", user.ID, factorID), nil)
-	if err != nil {
-		return err
-	}
-
-	var result pushResult
-	_, err = client.Do(req, &result)
+	result, _, err := client.UserFactor.VerifyFactor(ctx, user.Id, userFactor.Id, okta.VerifyFactorRequest{}, userFactor, nil)
```
calvn, line 1880: Should we check that the response on the verify call has the status `MFA_CHALLENGE` before polling against the factor result?

chrishoffman: According to the [documentation](https://developer.okta.com/docs/reference/api/factors/#issue-a-push-factor-challenge), I am only seeing the available statuses of `WAITING`, `SUCCESS`, `REJECTED`, and `TIMEOUT`. It could be that is not an option for push challenges which we filtered down to earlier.

changelog/15355.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+mfa/okta: migrate to use official Okta SDK
```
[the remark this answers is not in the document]
chrishoffman: I'm going to resolve this comment since we realized the issue was only related to dev accounts on Okta.

chrishoffman:
I also submitted this [issue](https://github.com/okta/okta-sdk-golang/issues/300) to the SDK. There is an easy enough workaround but it would be nice to use the transaction id from the verify response natively.

vault/login_mfa.go
```diff
@@ -1848,50 +1853,36 @@ func (c *Core) validateOkta(ctx context.Context, mConfig *mfa.Config, username s
...
-	var factorID string
-	for _, factor := range user.MFAFactors {
-		if factor.FactorType == "push" {
-			factorID = factor.ID
-			break
+	var factorFound bool
+	var userFactor *okta.UserFactor
+	for _, factor := range factors {
+		if factor.IsUserFactorInstance() {
+			userFactor = factor.(*okta.UserFactor)
```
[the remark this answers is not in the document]
chrishoffman: I'm going to leave that out of scope for now.

chrishoffman (commented):
@calvn @hghaf099 Can I get another quick review?

calvn (approved):

vault/login_mfa.go
```diff
@@ -1850,68 +1853,72 @@ func (c *Core) validateOkta(ctx context.Context, mConfig *mfa.Config, username s
...
+		// the VerifyFactor respone. This code effectively reimplements that method.
+		rq := client.CloneRequestExecutor()
+		req, err := rq.WithAccept("application/json").WithContentType("application/json").NewRequest("GET", url.String(), nil)
 		if err != nil {
 			return err
 		}
-		var result pushResult
-		_, err = client.Do(req, &result)
+		var result *okta.VerifyUserFactorResponse
+		_, err = rq.Do(ctx, req, &result)
```
hghaf099, line 1917: Seems like result is already a pointer now. Should `rq.Do` be fixed for that?

chrishoffman: I agree with you but this code is just mostly copied from the Okta SDK. https://github.com/okta/okta-sdk-golang/blob/66ce9052167bc3da973e2685c378a46937ae1dd5/okta/userFactor.go#L214-L232

PR #15319, opened 2022-05-06
Add parameter validation to the new generate key PKI api

stevendpclark:
- Validate the key_type and key_bits arguments that were provided and
   perform the same default processing of 0 as we used to do for the
   generateRoot/generateIntermediate apis
 - Add a test that validates the behaviour
 - Update the field description blurbs.

Addresses feedback from main PR: https://github.com/hashicorp/vault/pull/15277/files/bba4bcb19515e8ca5ea35931c15bff216951d5e0#r866289930

PR #15360, opened 2022-05-10
Fix revoking certificates in pre-migration state within PKI

stevendpclark:
This PR addresses some issues found when testing revoking certificates in a pre-migration state. The first commit addresses that issue and the second adds a suite of tests that validate all apis' behaviour within the pre-migration state, so it will be harder to regress them in the future. 

As part of the test suite some fixes were needed, mainly adding some missing assertions within the codebase that we aren't in legacy storage mode and one issue with a defined operation.

This PR was built on top of the many in-flight, so the following PRs will need to be merged prior to this one I believe.

#15285, #15289, #15303, #15306, #15315, #15319, #15322, #15325, #15336

### DIFF
--- a/builtin/logical/pki/cert_util.go
+++ b/builtin/logical/pki/cert_util.go
@@ -80,9 +80,32 @@ func getFormat(data *framework.FieldData) string {
 	return format
 }
 
-// fetchCAInfo will fetch the CA info, will return an error if no ca info exists.
+// fetchCAInfo will fetch the CA info, will return an error if no ca info exists, this does NOT support
+// loading using the legacyBundleShimID and should be used with care. This should be called only once
+// within the request path otherwise you run the risk of a race condition with the issuer migration on perf-secondaries.
 func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRef string, usage issuerUsage) (*certutil.CAInfoBundle, error) {
-	entry, bundle, err := fetchCertBundle(ctx, b, req.Storage, issuerRef)
+	var issuerId issuerID
+
+	if b.useLegacyBundleCaStorage() {
+		// We have not completed the migration so attempt to load the bundle from the legacy location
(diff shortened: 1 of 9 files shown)


builtin/logical/pki/cert_util.go
```diff
@@ -80,9 +80,32 @@ func getFormat(data *framework.FieldData) string {
 	return format
 }
 
-// fetchCAInfo will fetch the CA info, will return an error if no ca info exists.
+// fetchCAInfo will fetch the CA info, will return an error if no ca info exists, this does NOT support
+// loading using the legacyBundleShimID and should be used with care. This should be called only once
+// within the request path otherwise you run the risk of a race condition with the issuer migration on perf-secondaries.
 func fetchCAInfo(ctx context.Context, b *backend, req *logical.Request, issuerRef string, usage issuerUsage) (*certutil.CAInfoBundle, error) {
-	entry, bundle, err := fetchCertBundle(ctx, b, req.Storage, issuerRef)
+	var issuerId issuerID
```
cipherboy, line 87: Remind me again why it wasn't possible to use `fetchCertBundle` as the flex point again, rather than adding this here? It seems like the two would be equivalent: there's a single check for `useLegacyBundleCaStore` in both, and we should never go backwards (from `true->false`) on that variable. 

About the only material change I think I can see is two places that otherwise handled this fine now just directly call the `fetchCAInfoByIssuerId` / `fetchCertBundleByIssuerId` and the legacy bundle shim now is better which I _think_ is the core of the change to make CRL building work.

Maybe I'm misunderstanding that though? Regardless, I think I'm fine with it once I've wrapped my head around it.

cipherboy (approved):

builtin/logical/pki/cert_util.go
cipherboy: Ah, ok, that's fair :-) Sounds good to me then!

PR #15361, opened 2022-05-10
auth/okta: Add support for Okta number challenge

calvn:
This PR adds the ability for the Okta Auth Method to perform the [3-number verification challenge](https://help.okta.com/oie/en-us/Content/Topics/identity-engine/authenticators/configure-okta-verify-options.htm#:~:text=In%20the%20Push%20notification%3A%20Number%20Challenge%20section) if that's enabled on an Okta Organization.

The login flow is as follows, which happens within a single Vault login CLI command:
1. Perform a Vault login using Okta credentials
```
$ vault login -method=okta username=foo@example.com
Password (will be hidden):
...
```

2. Receive and confirm MFA prompt

3. Receive a number challenge in the CLI
```
...
In Okta Verify, tap the number '32'
...
```
4. Validate number on the user's device

5. Sucessful login
```
...
Success! You are now authenticated. The token information displayed below
is already stored in the token helper. You do NOT need to run "vault login"
again. Future Vault requests will automatically use this token.

Key                    Value
---                    -----
token                  hvs.CAESIF6yzMNIoAiH6WqTvZ9W_bhI-ux6qsLLBHfKoBCZegnKGh4KHGh2cy4wZVVQTmdnbDUxbUVzTGRTN25MNzlPS3k
token_accessor         GrGamdHye5dZKRr092bydAfO
token_duration         768h
token_renewable        true
token_policies         ["default" "foo"]
identity_policies      []
policies               ["default" "foo"]
token_meta_policies    default,foo
token_meta_username    foo@example.com
```


builtin/credential/okta/cli.go
```diff
@@ -48,6 +51,20 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
...
 	}
 
+	nonce := base62.MustRandom(20)
+	data["nonce"] = nonce
+
+	go func() {
+		for {
+			resp, _ := c.Logical().Read(fmt.Sprintf("auth/%s/verify/%s", mount, nonce))
+			if resp != nil {
+				fmt.Printf("In Okta Verify, tap the number '%s'\n", resp.Data["correctAnswer"].(json.Number))
```
swenson, line 61: For consistency, I think we would want to print this to `os.Stderr`

builtin/credential/okta/cli.go
```diff
@@ -48,6 +51,20 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
 		data["provider"] = provider
 	}
 
+	nonce := base62.MustRandom(20)
+	data["nonce"] = nonce
+
+	go func() {
+		for {
+			resp, _ := c.Logical().Read(fmt.Sprintf("auth/%s/verify/%s", mount, nonce))
```
swenson, line 59: What happens if the user responds with an incorrect number, or forgets about it? Will this loop forever?

calvn: The call to b.Login would unblock (failed auth), and the login would clean up the cached value through the defer call [in here](https://github.com/hashicorp/vault/pull/15361/files#diff-8ae1cf8247bed192b76102070c547ceda736bdad7243c035ef829add132c9205R88). The next verify call in this loop will return a nil response and we return from this goroutine.

builtin/credential/okta/path_login.go
```diff
@@ -34,6 +34,12 @@ func pathLogin(b *backend) *framework.Path {
 				Type:        framework.TypeString,
 				Description: "TOTP passcode.",
 			},
+			"nonce": {
+				Type: framework.TypeString,
+				Description: `Nonce provided if performing login that requires 
+number verification challenge. Logins through the vault login CLI command will 
+automatically generate a nonce`,
```
kalafut, line 41:
```suggestion
automatically generate a nonce.`,
```

builtin/credential/okta/cli.go
swenson: Hmm. But if the `verify` call continues to return a `nil` response, won't this loop forever?

calvn: You're right, I got things the other way around. The goroutine returns when we've received the number challenge from Okta, but before the user taps on that number on their device. For the case where the response is nil (no e.g. number challenge), this goroutine will run indefinitely until the program exits, i.e. the CLI finishes with either successful or failed auth.

What blocks during the login call is the login call on L68, so if the taps on the wrong number challenge, the request ends with failed auth pretty instantaneously:
```
Password (will be hidden):
In Okta Verify, tap the number '83'
Error authenticating: Error making API request.

URL: PUT http://localhost:8200/v1/auth/okta/login/foo@example.com
Code: 400. Errors:

* multi-factor authentication denied
```

Since this is not running on the server side, it will get GC'ed when the program exits but it's not great. I'll refactor this to make it cleaner.

builtin/credential/okta/path_login.go
```diff
@@ -172,6 +182,43 @@ func (b *backend) pathLoginRenew(ctx context.Context, req *logical.Request, d *f
...
+				Type: framework.TypeString,
+				Description: `Nonce provided during a login request to
+retrieve the number verification challenge for the matching request.`,
+			},
+		},
+		Operations: map[logical.Operation]framework.OperationHandler{
+			logical.ReadOperation: &framework.PathOperation{
+				Callback:                    b.pathVerify,
+				ForwardPerformanceSecondary: true,
+				ForwardPerformanceStandby:   true,
```
austingebauer, line 199: Just gave this a test using a perf standby. I think we'll need to either forward both the login/verify requests or not forward both of them. Right now, only the verify request is forwarded, which means that logins initiated on perf standby/secondary aren't able to complete the verify step (the nonce isn't cached on the instance the request is forwarded to).

I'd guess that it's probably fine to not forward this. The client would complete the full login/verify on the Vault instance it started the login with.

austingebauer (commented):
Just a couple of comments. Looks good to me otherwise 👍

builtin/credential/okta/path_login.go
calvn: You're right. The login request should be handled locally on perf secondaries, and the login handler does not seem to have any storage writes so it can be processed by perf standbys.

builtin/credential/okta/cli.go
calvn: Cleaned up the logic a bit so that we return from this goroutine whenever the we're done with `Auth`. Going to mark as resolved but feel free to re-open if you've got further comments!

austingebauer (approved):
👍

builtin/credential/okta/backend.go
```diff
@@ -246,7 +268,7 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username, pas
 				}
 
 				select {
-				case <-time.After(500 * time.Millisecond):
+				case <-time.After(1 * time.Second):
```
calvn, line 271: Taking the opportunity in this PR to change the polling frequency from 500ms to 1s. The rationale behind this change is that Okta dev accounts has a rate limit of 100req/min on `authn/` endpoints, and at this frequency calls made by a vault login will trigger a warning and eventually lead to rate limit exceeded errors. By lowering this to a max of 60req/min avoid hitting this rate limit and getting into flaky testing scenarios. The bump should not be that noticeable since this is a purely human-based flow.

Login MFA uses a 1s polling frequency, so it should be acceptable for us here as well: https://github.com/hashicorp/vault/blob/a2fc885b03d1121e64796632c5cc63320cc5d510/vault/login_mfa.go#L1928

swenson (approved):
LGTM

adrian-mi-romero:
@calvn how can I know if you fixed that was merged to main was in deed included in a vault release version? thanks in advance.

tvoran:
Hi @adrian-mi-romero, if you click on the commit sha from the "calvn merged commit" message above, and then expand the tags, it'll show which releases included this patch. Example screenshot:

Screenshot 2023-02-23 at 9 20 36 AM

adrian-mi-romero:
Thanks a lot @tvoran so it should be in v1.12.X ... wondering about it coz if you check the doc for v1.12.X it says your fix it is not supported/included.
https://developer.hashicorp.com/vault/docs/auth/login-mfa/faq#q-what-is-single-phase-mfa-vs-two-phase-mfa
image

tvoran:
@adrian-mi-romero Best to open another issue about that.

bobby-mack:
@calvn By chance did a recent update in 1.15.0 result in this feature breaking for Okta Verify Number Challenge? We're no longer presented with the corresponding matching number in the UI (previously working) as part of this PR.

PR #15362, opened 2022-05-10
UI: Fix version-history serializer and update mirage to reflect real API

hashishaw:
Fixes an issue where `version-history` keys weren't being properly accessed from the payload.


hellobontempo (approved):
Thanks for catching!! 🚀 edit: I noticed I missed some periods in the warning banner for the "Current" tab - so added those!

PR #15364, opened 2022-05-10
UI/ fix firefox not recognizing csv export

hellobontempo:
Without the explicit `.csv` extension in the filename, Firefox doesn't recognize the download file type.

### DIFF
--- a/ui/app/services/download-csv.js
+++ b/ui/app/services/download-csv.js
@@ -8,7 +8,8 @@ import Service from '@ember/service';
 
 export default class DownloadCsvService extends Service {
   download(filename, content) {
-    let formattedFilename = filename?.replace(/\s+/g, '-') || 'vault-data.csv';
+    // even though Blob type 'text/csv' is specified below, some browsers (ex. Firefox) require the filename has an explicit extension
+    let formattedFilename = `${filename?.replace(/\s+/g, '-')}.csv` || 'vault-data.csv';
     let { document, URL } = window;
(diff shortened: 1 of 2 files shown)


ui/app/services/download-csv.js
```diff
@@ -8,7 +8,8 @@ import Service from '@ember/service';
 
 export default class DownloadCsvService extends Service {
   download(filename, content) {
-    let formattedFilename = filename?.replace(/\s+/g, '-') || 'vault-data.csv';
+    // even though Blob type 'text/csv' is specified below, some browsers (ex. Firefox) require the filename has an explicit extension
+    let formattedFilename = `${filename?.replace(/\s+/g, '-')}.csv` || 'vault-data.csv';
```
arnav28, line 12: nit: filename param passed is without any extension right?

arnav28 (approved):

ui/app/services/download-csv.js
hellobontempo: Good question! But correct it's generated in the attribution component. I thought it'd simplify things to have the service take care of all csv/download related things. 

``` 
 get getCsvFileName() {
    let endRange = this.isDateRange ? `-${this.args.endTimeDisplay}` : '';
    let csvDateRange = this.args.startTimeDisplay + endRange;
    return this.isSingleNamespace
      ? `clients_by_auth_method_${csvDateRange}`
      : `clients_by_namespace_${csvDateRange}`;
  }

PR #15376, opened 2022-05-11
secrets/database: adds ability to manage alternative credential types and configuration

austingebauer:
## Overview

This PR adds the ability for the database secrets engine to manage alternative credential types and configuration. This provides the foundation for the secrets engine to start managing database credentials other than passwords (e.g., asymmetric keys, x.509 client certificates). 

Initially, this will be used to enable Vault to manage RSA key pair credentials to be consumed by the [Snowflake database plugin](https://github.com/hashicorp/vault-plugin-database-snowflake). This will enable clients to use [key pair authentication](https://docs.snowflake.com/en/user-guide/key-pair-auth.html) using a Vault-managed credential. See https://github.com/hashicorp/vault-plugin-database-snowflake/pull/15 for example usage on the plugin side.

**Dynamic Role Usage:**
```sh
vault write database/roles/my-dynamic-keypair-role \
	db_name=my-snowflake-database \
	creation_statements="CREATE USER {{name}} RSA_PUBLIC_KEY='{{public_key}}'
		DAYS_TO_EXPIRY = {{expiration}} DEFAULT_ROLE=myrole;
		GRANT ROLE myrole TO USER {{name}};" \
	default_ttl="1h" \
	credential_type="rsa_private_key" \
        credential_config=key_bits=2048 \ # Optional
        credential_config=format="pkcs8"  # Optional
```

**Static Role Usage:**
```sh
vault write database/static-roles/my-static-role \
    db_name=my-snowflake-database \
    rotation_statements="ALTER USER {{name}} SET RSA_PUBLIC_KEY='{{public_key}}'" # Optional
    username="static_user_keypair" \
    rotation_period="5m" \
    credential_type="rsa_private_key" \
    credential_config=key_bits=2048 \ # Optional
    credential_config=format="pkcs8"  # Optional
```

## Testing

Basic testing has been included in this PR. Tests that integrate with Snowflake to use RSA key pair credentials are included in the Snowflake database plugin PR.

Additional tests are in progress and will be included in further commits.


builtin/logical/database/rotation.go
```diff
@@ -330,61 +351,124 @@ func (b *databaseBackend) setStaticAccount(ctx context.Context, s logical.Storag
...
+			output.WALID = ""
+		case !wal.credentialIsSet():
+			b.Logger().Error("expected WAL to have a new credential set, but empty", "role", input.RoleName, "WAL ID", output.WALID)
+			if err := framework.DeleteWAL(ctx, s, output.WALID); err != nil {
+				b.Logger().Warn("failed to delete WAL with no new credential", "error", err, "WAL ID", output.WALID)
+			}
+
+			// Generate a new WAL entry and credential
+			output.WALID = ""
 		default:
```
austingebauer, line 383: This is a bit of a refactor that improves readability IMO. The existing static role rotation code is pretty defensive (perhaps for a good reason, but it's not totally obvious). Happy to consider thoughts around how we can further improve readability here.

swenson: This method is about 200 lines long and does many things. It might be more readable/maintainable to split it up into a few functions that do fewer things?

builtin/logical/database/rotation.go
```diff
@@ -330,61 +351,124 @@ func (b *databaseBackend) setStaticAccount(ctx context.Context, s logical.Storag
...
-			if wal == nil {
-				b.Logger().Error("expected role to have WAL, but WAL not found in storage", "role", input.RoleName, "WAL ID", output.WALID)
-			} else {
-				b.Logger().Error("expected WAL to have a new password set, but empty", "role", input.RoleName, "WAL ID", output.WALID)
-				err = framework.DeleteWAL(ctx, s, output.WALID)
-				if err != nil {
-					b.Logger().Warn("failed to delete WAL with no new password", "error", err, "WAL ID", output.WALID)
+			// Use the credential from the existing WAL entry
+			switch wal.CredentialType {
```
swenson, line 385: I feel like this switch could be squashed up into the previous switch, maybe?

sdk/database/dbplugin/v5/proto/database.proto
```diff
@@ -27,6 +27,8 @@ message NewUserRequest {
     google.protobuf.Timestamp expiration = 3;
     Statements statements = 4;
     Statements rollback_statements = 5;
+    int32 credential_type = 6;
+    bytes public_key = 7;
```
swenson, line 31: Maybe a minor thing, but since we're only using PEM-encoded keys, those should all be valid `string`s, right?

I mostly worry that `bytes` (or `[]byte`) above might lead someone to think we're using something like DER.

builtin/logical/database/rotation.go
austingebauer: I think you're right. Let me see what that would look like 👍

builtin/logical/database/rotation.go
austingebauer: I agree. I'd like to spend some time trying to reduce the complexity of this static rotation code in general. I'll take a look to see if there are any small, non-risky improvements that can be made as you've suggested. Otherwise, I'd prefer to leave any larger refactor out of this PR.

sdk/database/dbplugin/v5/proto/database.proto
austingebauer: Yep, I originally used a `string` type here. I would be fine going back to `string` since we are going with PEM encoding. Will do so if others agree.

builtin/logical/database/path_creds_create.go
```diff
@@ -107,21 +101,75 @@ func (b *databaseBackend) pathCredsCreateRead() framework.OperationFunc {
...
-			Password:   password,
 			Expiration: expiration,
 		}
 
-		// Overwriting the password in the event this is a legacy database plugin and the provided password is ignored
+		respData := make(map[string]interface{})
+
+		// Generate the credential based on the role's credential type
+		switch role.CredentialType {
+		case v5.CredentialTypePassword:
```
calvn, line 111: I think that we need to also include `""` for existing roles that didn't have this set prior to this field existing.

builtin/logical/database/path_creds_create.go
```diff
@@ -107,21 +101,75 @@ func (b *databaseBackend) pathCredsCreateRead() framework.OperationFunc {
...
-
-		respData := map[string]interface{}{
-			"username": newUserResp.Username,
-			"password": password,
+		respData["username"] = newUserResp.Username
+
+		// Database plugins using the v4 interface generate and return the password.
+		// If the returned password is not equal to the credential, then we need to
+		// set the credential to the password returned from the v4 database plugin.
+		if role.CredentialType == v5.CredentialTypePassword &&
```
calvn, line 168: I'm a bit confused about this this conditional. If `role.CredentialType == v5.CredentialTypePassword`, then it looks like we always set `respData["password"]` to  `password` in [L134](https://github.com/hashicorp/vault/pull/15376/files#diff-a83220b81620363186d8af7542730e768ccab09cf9a55c30cf24a5819110177bR134), so can this case ever be true?

builtin/logical/database/path_creds_create.go
```diff
@@ -158,14 +206,22 @@ func (b *databaseBackend) pathStaticCredsRead() framework.OperationFunc {
...
 
+		respData := map[string]interface{}{
+			"username":            role.StaticAccount.Username,
+			"ttl":                 role.StaticAccount.CredentialTTL().Seconds(),
+			"rotation_period":     role.StaticAccount.RotationPeriod.Seconds(),
+			"last_vault_rotation": role.StaticAccount.LastVaultRotation,
+		}
+
+		switch role.CredentialType {
+		case v5.CredentialTypePassword:
```
calvn, line 217: We should also account for the `""` case and default this to the password behavior for existing, older roles. This should be applicable to all cases with similar switch on the role's credential type.

builtin/logical/database/path_roles.go
```diff
@@ -356,6 +371,23 @@ func (b *databaseBackend) pathRoleCreateUpdate(ctx context.Context, req *logical
 		if role.DBName == "" {
 			return logical.ErrorResponse("database name is required"), nil
 		}
+
+		if credentialTypeRaw, ok := data.GetOk("credential_type"); ok {
+			credentialType := credentialTypeRaw.(string)
+			if err := role.setCredentialType(credentialType); err != nil {
```
calvn, line 377: Where do we set the default to be password if not specified?

builtin/logical/database/credentials.go
```diff
@@ -0,0 +1,172 @@
...
+
+	if pg.PasswordPolicy == "" {
+		return random.DefaultStringGenerator.Generate(ctx, b.GetRandomReader())
+	}
+	return b.System().GeneratePasswordFromPolicy(ctx, pg.PasswordPolicy)
+}
+
+// toKVPairs returns the configuration of the passwordGenerator
+// as a map from string to string.
+func (pg passwordGenerator) toKVPairs() map[string]string {
```
calvn, line 62: Nit: `configMap` might be a bit more descriptive than `toKVPairs` on what this is returning.

builtin/logical/database/path_roles_test.go
```diff
@@ -154,7 +337,7 @@ func TestBackend_StaticRole_Config(t *testing.T) {
 
 			expected := tc.expected
 			actual := make(map[string]interface{})
-			for _, key := range dataKeys {
+			for _, key := range []string{"username", "password", "last_vault_rotation", "rotation_period"} {
```
calvn, line 340: Nit: Should we keep `dataKeys := []string{...}` but have it be local to `TestBackend_StaticRole_Config`?

builtin/logical/database/path_creds_create.go
austingebauer: Ya, this caused me some confusion as well. I'm trying to preserve some behavior that existed prior to these changes. This [comment](https://github.com/hashicorp/vault/blob/main/builtin/logical/database/path_creds_create.go#L114) in the main branch sort of mentions what the issue is.

If we assume we're working with a v4 db plugin, it turns out our existing code for reaching out to the plugin to generate the password at [path_creds_create.go#L93](https://github.com/hashicorp/vault/blob/main/builtin/logical/database/path_creds_create.go#L93) is not needed. This is because the password is always generated on the db plugin side and returned in the v4 CreateUser response. Here is an example of a v4 implementation that shows this: [postgresql.go#L181](https://github.com/hashicorp/vault/blob/44fdbc7dc92ac9e9d801556c40ff9974f8dbfcd8/plugins/database/postgresql/postgresql.go#L181).

This means for v4 plugins, we're currently asking it to generate a password, and then not using it. Happy to consider how we can clean this up.

builtin/logical/database/path_creds_create.go
austingebauer: [CredentialType](https://github.com/hashicorp/vault/blob/db-secrets-credential-types/sdk/database/dbplugin/v5/database.go#L124-L130) is intentionally an `int` so that we don't have to check for the empty string on existing roles. The zero value will default to `CredentialTypePassword` for existing roles, which is what we want.

builtin/logical/database/path_creds_create.go
austingebauer: Similar comment above. Don't need to check for "" as it's an int.

builtin/logical/database/path_roles.go
austingebauer: The zero value for `CredentialType` on the [roleEntry](https://github.com/hashicorp/vault/blob/db-secrets-credential-types/builtin/logical/database/path_roles.go#L614) is password. So if you don't provide the `credential_type`, it'll default to password.

builtin/logical/database/path_roles_test.go
austingebauer: Will do 👍

builtin/logical/database/credentials.go
austingebauer: I like that better. Done in e9e3536.

builtin/logical/database/rotation.go
austingebauer: That simplified things (a little bit 🙂). Done in b923e86. Thanks!

builtin/logical/database/credentials.go
```diff
@@ -0,0 +1,172 @@
...
+		Bytes: private,
+	}
+
+	return pem.EncodeToMemory(publicBlock), pem.EncodeToMemory(privateBlock), nil
+}
+
+// configMap returns the configuration of the rsaKeyGenerator
+// as a map from string to string.
+func (kg rsaKeyGenerator) configMap() map[string]string {
+	return interfaceValuesToString(structs.Map(kg))
```
swenson, line 161: Any particular reason we're using `structs` (which seems to be unmaintained) instead of `mapstructure`?

builtin/logical/database/rotation.go
swenson: Sounds good.

swenson (approved):
LGTM. Whew, big change, but I like it. A minor possible nit/suggestion. :)

builtin/logical/database/credentials.go
austingebauer: We're using `mapstructure` to weakly decode `map[string]string` to the `rsaKeyGenerator` struct, but I didn't know you could go back the other way 🤔 Looks like maybe you can. Will give it a try.

I used `structs` because it solved the problem and is already used in other locations of Vault. I'd like to not use it though, so let me see what I can do. Thanks!

austingebauer: I wasn't able to get the conversion from `rsaKeyGenerator` struct to `map[string]string` working using `mapstructure`.. I ended up removing the use of `structs` and going with something simple instead. That's done in 3eaa2ec. Let me know if you have any thoughts around this approach.

swenson: :+1: definitely simpler

builtin/logical/database/path_creds_create.go
calvn: I see. We used to always overwrite the password through since `databaseVersionWrapper.NewUser` either returns the provided password, `req.Password`, or the password returned by the plugin. I think that just the `if role.CredentialType == v5.CredentialTypePassword {...}` might be sufficient here, and we could skip setting that on [L134](https://github.com/hashicorp/vault/pull/15376/files#diff-a83220b81620363186d8af7542730e768ccab09cf9a55c30cf24a5819110177bR134)

builtin/logical/database/credentials.go
```diff
@@ -0,0 +1,174 @@
...
+	return b.System().GeneratePasswordFromPolicy(ctx, pg.PasswordPolicy)
+}
+
+// configMap returns the configuration of the passwordGenerator
+// as a map from string to string.
+func (pg passwordGenerator) configMap() map[string]string {
+	config := make(map[string]string)
+
+	// Configuration without defaults are included if not empty
+	if pg.PasswordPolicy != "" {
```
calvn, line 65: We discussed offline on how we could dynamically generate this with `mapstructure`, so I'll let you explore this a bit more :)

sdk/database/dbplugin/v5/proto/database.proto
calvn: I don't have a strong preference either way. FWIW the standard library call `pem.EncodeToMemory` also returns it as `[]byte`.

calvn (approved):

builtin/logical/database/credentials.go
austingebauer: Thanks for the idea! I've changed this to use `mapstructure` in ab58627. I like this better, since we can add additional config fields without needing to update the `configMap` methods.

builtin/logical/database/path_creds_create.go
austingebauer: Ah, that's right. I didn't realize the password came back from `NewUser` for both v4 and v5. Just simplified this as you suggested in fe3e50a. Thanks!

PR #15377, opened 2022-05-11
Added support for VAULT_PROXY_ADDR + Updated docs

peteski22:
Updated documentation to describe the behavior when supplying `VAULT_HTTP_PROXY`. 
Also added support for `VAULT_PROXY_ADDR` as a 'better name' for `VAULT_HTTP_PROXY`.


website/content/docs/commands/index.mdx
```diff
@@ -379,9 +379,26 @@ used.
 
 ### `VAULT_HTTP_PROXY`
 
-HTTP proxy location which should be used to access Vault. When present, this
-overrides any other proxies found in the environment. Format should be
-`http://server:port`.
+HTTP or HTTPS proxy location which should be used by all requests to access Vault.
```
taoism4504, line 382:
```suggestion
HTTP or HTTPS proxy location should be used by all requests to access Vault.
```

website/content/docs/commands/index.mdx
```diff
@@ -379,9 +379,26 @@ used.
...
+ When present, this overrides the default proxy resolution behavior.
+Format should be `http://server:port` or `https://server:port`.
+
+(See `VAULT_PROXY_ADDR` below).
+
+### `VAULT_PROXY_ADDR`
+
+HTTP or HTTPS proxy location which should be used by all requests to access Vault.
```
taoism4504, line 390:
```suggestion
HTTP or HTTPS proxy location should be used by all requests to access Vault.
```

website/content/docs/commands/index.mdx
```diff
@@ -379,9 +379,26 @@ used.
...
+ When present, this overrides the default proxy resolution behavior.
+Format should be `http://server:port` or `https://server:port`.
+
+~> Note: When using `VAULT_HTTP_PROXY` or `VAULT_PROXY_ADDR` any of the standard
+ proxy variables found in the environment will be ignored.
+Specifically `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY`.
+All requests will resolve the specified proxy, there is no way to exclude
```
taoism4504, line 397:
```suggestion
All requests will resolve the specified proxy; there is no way to exclude
```

taoism4504 (approved):
minor nits; but LGTM.  I'm not sure if you require other reviewers since there seems to be new code introduced.

api/client_test.go
```diff
@@ -1201,3 +1201,145 @@ func TestClientWithNamespace(t *testing.T) {
 		t.Fatalf("Expected original namespace: \"%s\", got \"%s\"", ogNS, client.Namespace())
 	}
 }
+
+func TestVaultHttpProxyUsedWhenNoProxyEnvVarDoesntIncludeRequestHost(t *testing.T) {
```
ncabatoff, line 1205: I think these test names are a bit long.  One way you might address this: turn them into a table-driven test using t.Run.  Then the test case name can be as descriptive as you like, while the test method name is of a more reasonable length.  It would also make it easier to see at a glance what combinations of settings are covered.

ncabatoff (approved):
I'm not 100% convinced that the better name is worth having two different ways to do something and the potential confusion that introduces.  That said, I like the tests and the docs, and my minor misgivings aren't enough for me to reject this.  Perhaps in time we could retire the old name.

api/client_test.go
```diff
@@ -1201,3 +1200,88 @@ func TestClientWithNamespace(t *testing.T) {
 		t.Fatalf("Expected original namespace: \"%s\", got \"%s\"", ogNS, client.Namespace())
 	}
 }
+
+func TestVaultProxy(t *testing.T) {
+	const VaultHttpProxy string = "VAULT_HTTP_PROXY"
```
ncabatoff, line 1205: Curious why you're not using the existing constants?

api/client_test.go
```diff
@@ -1201,3 +1200,88 @@ func TestClientWithNamespace(t *testing.T) {
...
+			if c.Error != nil {
+				t.Fatalf("Expected no error reading config, found error %v", c.Error)
+			}
+
+			r, _ := http.NewRequest("GET", tc.requestUrl, nil)
+			proxyUrl, err := c.HttpClient.Transport.(*http.Transport).Proxy(r)
+
+			if err != nil {
+				t.Fatalf("Expected no error resolving proxy, found error %v", err)
+			} else if proxyUrl == nil || proxyUrl.String() == "" {
```
ncabatoff, line 1280: You don't need else if here, t.Fatal will terminate the test.

api/client_test.go
```diff
@@ -1201,3 +1200,88 @@ func TestClientWithNamespace(t *testing.T) {
...
+	const VaultProxyAddr string = "VAULT_PROXY_ADDR"
+	const NoProxy string = "NO_PROXY"
+
+	tests := map[string]struct {
+		name                  string
+		vaultHttpProxy        string
+		vaultProxyAddr        string
+		noProxy               string
+		requestUrl            string
+		checkProxyVarPriority bool
```
ncabatoff, line 1215: Instead of a bool, would it make sense to have this be a string that's the expected value for proxyUrl (or omit that check if empty string)?

peteski22:
> I'm not 100% convinced that the better name is worth having two different ways to do something and the potential confusion that introduces. That said, I like the tests and the docs, and my minor misgivings aren't enough for me to reject this. Perhaps in time we could retire the old name.

Yes, ideally we could get rid of the `VAULT_HTTP_PROXY` at some point. 👍🏼

PR #15378, opened 2022-05-11
Fix OIDC callback query params

arnav28:
- Value of namespace was getting stripped from the state query param
- Used native URL search param api to fetch the values

### DIFF
--- a/changelog/15378.txt
+++ b/changelog/15378.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: fix OIDC callback query params
+```
--- a/changelog/15378.txt
+++ b/changelog/15378.txt
@@ -1,3 +1,3 @@
-```release-note:bug
-ui: fix OIDC callback query params
+```release-note:improvement
+ui: Allow namespace param to be parsed from state queryParam
 ```
--- a/ui/app/routes/vault/cluster/oidc-callback.js
+++ b/ui/app/routes/vault/cluster/oidc-callback.js
@@ -6,11 +6,7 @@ export default Route.extend({
     // left blank so we render the template immediately
   },
   afterModel() {
-    let queryString = window.location.search;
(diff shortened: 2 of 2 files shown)


ui/app/routes/vault/cluster/oidc-callback.js
```diff
@@ -6,7 +6,18 @@ export default Route.extend({
     // left blank so we render the template immediately
   },
   afterModel() {
-    let { auth_path: path, code, state } = this.paramsFor(this.routeName);
+    let queryString = window.location.search;
+    // Check if url is encoded
+    if (this.containsEncodedComponents(queryString)) {
+      queryString = decodeURIComponent(queryString);
+    }
```
zofskeez, line 13: Is this check necessary? I believe if you use `decodeURIComponent` on a string that is not encoded it will return the same string. For example:
```js
decodeURIComponent('?foo=bar') === decodeURIComponent('%3Ffoo%3Dbar'); // true
```
Unless I'm not accounting for something this could be simplified to:
```js
const queryString = decodeURIComponent(window.location.search);
```

zofskeez (commented):
This looks good to me overall! I left a comment with a suggestion to simplify things for your review.

ui/app/routes/vault/cluster/oidc-callback.js
arnav28: Oh got it. Thanks, updated!! 👍

zofskeez (approved):
LGTM!

changelog/15378.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+ui: fix OIDC callback query params
```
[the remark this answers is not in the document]
arnav28: True, updated the change log. 👍

PR #15379, opened 2022-05-11
docs: update for vault-k8s 0.16.0

tvoran:
https://github.com/hashicorp/vault-k8s/blob/main/CHANGELOG.md#0160-may-11-2022


swenson (approved):
LGTM

PR #15380, opened 2022-05-11
VAULT-5935 agent: redact renew-self if using auto auth

swenson:
Vault agent redacts the token and accessor for `/auth/token/lookup-self` (and `lookup`)
if the token is the auto auth token to prevent it from leaking.

Similarly, we need to redact the token and accessor from `renew-self`
and `renew`, which also leak the token and accessor.

I tested this locally by starting up a Vault agent and querying the
agent endpoints, and ensuring that the accessor and token were set to
the empty string in the response.


calvn (approved):
Looks great!

tomhjp (approved):
LGTM

swenson:
Thanks!

Issue #15331, opened 2022-05-09
unknown or unsupported field - consul and tcp


jhrcz-ls:
**Describe the bug**
After upgrade from 1.10.0 to 1.10.2, new warning is shown on every restart or reload, using the same json configuration.

```
May  6 10:38:41 vault2 vault[xxx]: 2022-05-06T10:38:41.233+0200 [WARN]  unknown or unsupported field consul found in configuration at /etc/vaultserver/config.json
May  6 10:38:41 vault2 vault[xxx]: 2022-05-06T10:38:41.233+0200 [WARN]  unknown or unsupported field tcp found in configuration at /etc/vaultserver/config.json

```
it may be connected with this change... https://github.com/hashicorp/vault/pull/14752

confirmed it happens with the same after minor version change to 1.10.1 a 1.9.5 

**To Reproduce**
1. Run `docker run --rm -ti -v "$(pwd)/config.json:/config.json:ro" hashicorp/vault:1.10.1 vault server -log-level=trace -config /config.json`
2. See error

```
2022-05-09T12:30:35.891Z [WARN]  unknown or unsupported field consul found in configuration at /config.json
2022-05-09T12:30:35.891Z [WARN]  unknown or unsupported field tcp found in configuration at /config.json
```

**Expected behavior**

error not reported the same way as with older version:
`docker run --rm -ti -v "$(pwd)/config.json:/config.json:ro" hashicorp/vault:1.10.0 vault server -log-level=trace -config /config.json`

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.10.1 or 1.9.5
* Vault CLI Version (retrieve with `vault version`): x
* Server Operating System/Architecture: linux (downloaded binary or container `hashicorp/vault:1.10.1` / `hashicorp/vault:1.9.5`

Vault server configuration file(s):

```json
{
    "api_addr": "https://localhost:8200",
    "default_lease_ttl": "6h",
    "disable_mlock": true,
    "listener": {
        "tcp": {
            "address": "0.0.0.0:8200"
        }
    },
    "log_level": "info",
    "storage": {
        "consul": {
            "address": "127.0.0.1:8500"
        }
    },
    "ui": true
}

```

**Additional context**

looks like just warning report, function is not affected (error reports storage and tcp field, and tcp endpoints and storage works ok)

lenaing:
Hello,

I have the same issue with the following JSON config (Vault v1.10.2), hopefully everything works as expected :

```
[...]

PR #15382, opened 2022-05-11
Remove Storage Type and Listener Type From Unused Keys List

HridoyRoy:
This PR solves https://github.com/hashicorp/vault/issues/15331


HridoyRoy:
Was not on main branch when opening this pull request, will open a new one from the main branch.

PR #15383, opened 2022-05-11
report listener and storage types as found keys

HridoyRoy:
This PR fixes https://github.com/hashicorp/vault/issues/15331

### DIFF
--- a/command/server/config.go
+++ b/command/server/config.go
@@ -562,6 +562,7 @@ func ParseConfig(d, source string) (*Config, error) {
 		if err := ParseStorage(result, o, "storage"); err != nil {
 			return nil, fmt.Errorf("error parsing 'storage': %w", err)
 		}
+		result.found(result.Storage.Type, result.Storage.Type)
 	} else {
 		delete(result.UnusedKeys, "backend")
 		if o := list.Filter("backend"); len(o.Items) > 0 {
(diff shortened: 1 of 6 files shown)


command/server/config.go
```diff
@@ -967,3 +968,8 @@ func (c *Config) Prune() {
 		c.Telemetry.UnusedKeys = nil
 	}
 }
+
+func (c *Config) found(s, k string) {
```
hghaf099, line 972: From both instances that used this function, wondering when would `s` and `k` be different?

hghaf099 (commented):
I just have one question, otherwise looks good!

sgmiller (approved):
Approve apart from hghaf099's comment

command/server/config.go
HridoyRoy: My primary reason for introducing this function is to keep parity with our helper in SharedConfig:

```func (c *SharedConfig) found(s, k string) {
	delete(c.UnusedKeys, s)
	c.FoundKeys = append(c.FoundKeys, k)
}```

There are often cases where we transform the capitalization of keys in configs. For example, `result.found("seal", "Seal")` and `result.found("entropy", "Entropy")`. While we currently aren't doing this for the top level config (only the SharedConfig) I'd like to keep this functionality for future development.

PR #15392, opened 2022-05-12
refactor: replace strings.Replace with strings.ReplaceAll

Juneezee:
Replace `strings.Replace(s, old, new, -1)` with `strings.ReplaceAll(s, old, new)`. `strings.ReplaceAll` is a wrapper function for `strings.Replace`, but `strings.ReplaceAll` is more readable and removes the hardcoded `-1`.


ccapurso (approved):
Definitely more readable. Thank you for the submission!

PR #15401, opened 2022-05-12
Fix unknown type parameter during CA generation

cipherboy:
When generating a new PKI CA from the Web UI, the UI incorrectly sends
the type parameter in the POST body. The server will now warn on unknown
parameters, resulting in the UI surfacing that up to the caller. Since
the type is part of the URL, we don't need to duplicate it in the POST
body, so elide it.

---

Marking no-changelog as this is part of the server warning feature new in 1.11 and while affecting earlier versions, didn't cause a warning. 

Before:

image

After:

image

### DIFF
--- a/ui/app/adapters/pki-ca-certificate.js
+++ b/ui/app/adapters/pki-ca-certificate.js
@@ -32,6 +32,11 @@ export default ApplicationAdapter.extend({
       data = { certificate: snapshot.attr('certificate') };
     } else {
       data = serializer.serialize(snapshot, requestType);
+
+      // The type parameter is serialized but is part of the URL. This means
+      // we'll get an unknown parameter warning back from the server if we
+      // send it. Remove it instead.
+      delete data['type'];
     }
 
     return this.ajax(this.url(snapshot, action), 'POST', { data }).then((response) => {
--- a/ui/app/adapters/pki-ca-certificate.js
+++ b/ui/app/adapters/pki-ca-certificate.js
@@ -36,7 +36,7 @@ export default ApplicationAdapter.extend({
       // The type parameter is serialized but is part of the URL. This means
       // we'll get an unknown parameter warning back from the server if we
       // send it. Remove it instead.
-      delete data['type'];
+      delete data.type;
     }
 
     return this.ajax(this.url(snapshot, action), 'POST', { data }).then((response) => {


ui/app/adapters/pki-ca-certificate.js
```diff
@@ -32,6 +32,11 @@ export default ApplicationAdapter.extend({
       data = { certificate: snapshot.attr('certificate') };
     } else {
       data = serializer.serialize(snapshot, requestType);
+
+      // The type parameter is serialized but is part of the URL. This means
+      // we'll get an unknown parameter warning back from the server if we
+      // send it. Remove it instead.
+      delete data['type'];
```
hellobontempo, line 39: 😍  the comments! wondering if we should wrap the delete in a conditional, in case the form changes. Although the failure at attempting the delete a non-existent attr might be a good thing...
```suggestion
      if (data.type) delete data.type;
```

hellobontempo (approved):
Thanks for tackling!

ui/app/adapters/pki-ca-certificate.js
cipherboy: The `type` field of the form is used above for URL building:

https://github.com/hashicorp/vault/blob/5f696cbad4a8e6172a90fe42ed6447ff08d44a86/ui/app/adapters/pki-ca-certificate.js#L7-L19 

Which is called from below:

https://github.com/hashicorp/vault/blob/5f696cbad4a8e6172a90fe42ed6447ff08d44a86/ui/app/adapters/pki-ca-certificate.js#L42

If the form changes, we've got bigger issues (URL will now be invalid, causing _everything_ to fail). My 2c? :-)

hashishaw: Typically we'd do transformations like this in the serializer, but since that doesn't exist for this model type I think this is fine. Deleting an unknown key should not throw an error, but stylistically we should make it `delete data.type` instead. Thank you!

cipherboy:
```suggestion
      delete data.type;
```

cipherboy: Thank you!

hashishaw (approved):
Thank you!

cipherboy:
Thanks all! Merging...

PR #15407, opened 2022-05-12
UI/remove chart check

hellobontempo:


hashishaw (approved):

ui/tests/acceptance/client-current-test.js
```diff
@@ -121,27 +100,6 @@ module('Acceptance | clients current', function (hooks) {
...
-      }
-      if (key === 'yLabels') {
-        elementArray.forEach((element, i) => {
-          assert
-            .dom(element.children[1])
-            .hasTextContaining(`${mounts[i].mount_path}`, 'displays correct auth label');
-        });
-      }
-    }
-
```
zofskeez, line 144: I'm guessing the flaky failures came from within these loops? Is there a plan to revisit this and add back in more stable coverage for checking the chart elements and values?

zofskeez (approved):

ui/tests/acceptance/client-current-test.js
hellobontempo: The tests were flaky prior, but yes - the tests failing after removing the original failing assertions definitely indicates that the loops were slowing things down. 

Definitely want to revisit - but will have to brainstorm the best way to approach. I'm thinking of moving it to the component test, rather than have it be incorporated in the acceptance.

zofskeez: Nice! Sounds good to me.

PR #15408, opened 2022-05-12
UI/cleanup client charts for 1.11

hellobontempo:
In this PR: 
* Consume backend changes from #15420 (so charts match queried date range, even if some months have no data) 
* Cleanup vertical bar chart so bars are centered over x-axis labels, regardless of dataset length: 
<hr> 

**before:**

image

<hr> 

**with fix:** 
>  _In the gif the graphs display the current month (May). Since this is the History tab, it shouldn't be included at all. This is a known backend bug related to the ticket above and a fix is in the works!_

clients-1 11


arnav28 (approved):

ui/app/components/clients/line-chart.js
```diff
@@ -56,8 +56,7 @@ export default class LineChart extends Component {
       );
     }
     const filteredData = dataset.filter((e) => Object.keys(e).includes(this.yKey)); // months with data will contain a 'clients' key (otherwise only a timestamp)
-    const dataMax = max(filteredData.map((d) => d[this.yKey]));
-    const domainMax = Math.ceil(dataMax / 10) * 10; // we want to round UP to the nearest tens place ex. dataMax = 102, domainMax = 110
```
hellobontempo, line 60: This made things weird actually, so removed

PR #15411, opened 2022-05-13
Add tests for usage-based restrictions of issuers

cipherboy:
This adds tests for usage-based restriction of issuers.

Since CRL building is complicated w.r.t. if parallel issuers exists, I felt it was useful to add to the existing complex multi-issuer tests. 

If issuance succeeds, we'll try revoking the cert and make sure it gets revoked somewhere, unless it is the default issuer and then there's a chance it won't show up on a CRL right away.

### DIFF
--- a/builtin/logical/pki/chain_test.go
+++ b/builtin/logical/pki/chain_test.go
@@ -432,15 +432,34 @@ func (c CBValidateChain) Run(t testing.TB, b *backend, s logical.Storage, knownK
 type CBUpdateIssuer struct {
 	Name    string
 	CAChain []string
+	Usage   string
 }
 
 func (c CBUpdateIssuer) Run(t testing.TB, b *backend, s logical.Storage, knownKeys map[string]string, knownCerts map[string]string) {
 	url := "issuer/" + c.Name
 	data := make(map[string]interface{})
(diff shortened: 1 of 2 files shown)


builtin/logical/pki/chain_test.go
```diff
@@ -525,14 +566,37 @@ func (c CBIssueLeaf) Run(t testing.TB, b *backend, s logical.Storage, knownKeys
...
+			if err != nil {
+				t.Fatalf("failed to read default issuer config: %v", err)
+			}
+			if resp == nil {
+				t.Fatalf("failed to read default issuer config: nil response")
+			}
+			defaultID := resp.Data["default"].(issuerID).String()
+			c.Issuer = defaultID
+			issuer = nil
+			CBRead(b, s, "crl/rotate")
```
stevendpclark, line 583: I don't understand why the crl/rotate call would need to be done here? The revoke should have forced the cert to be on an expected CRL?

cipherboy: Ah that's right. I had a bug earlier in my tests where I forgot about duplicate issuers, so I had added this call and we don't actually need it.

stevendpclark (approved):
lgtm!

PR #15414, opened 2022-05-13
Transit byok import endpoints

schultz-is:
# Summary
This adds two new endpoints to the Transit backend that enable BYOK (bring your own key):
- `keys/:name/import`, which imports an external key into a new Transit key for use within Vault.
- `keys/:name/import_version`, which imports an external key into a new version of an existing Transit key for use within Vault.

The intended use cases for Transit BYOK are seamless migration to Vault from legacy HSM/KMS solutions and import of externally-generated shared keys into an organization that already uses Transit. Since the security of imported keys cannot be guaranteed, and because the import process is complicated, this feature is not recommended for use without a full understanding of security implications.


builtin/logical/transit/path_import.go
```diff
@@ -0,0 +1,363 @@
...
+				Description: `The type of key being imported. Currently, "aes128-gcm96" (symmetric), "aes256-gcm96" (symmetric), "ecdsa-p256"
+(asymmetric), "ecdsa-p384" (asymmetric), "ecdsa-p521" (asymmetric), "ed25519" (asymmetric), "rsa-2048" (asymmetric), "rsa-3072"
+(asymmetric), "rsa-4096" (asymmetric) are supported.  Defaults to "aes256-gcm96".
+`,
+			},
+			"hash_function": {
+				Type:    framework.TypeString,
+				Default: "SHA256",
+				Description: `The hash function used as a random oracle in the OAEP wrapping of the user-generated,
+ephemeral AES key. Can be one of "SHA1", "SHA224", "SHA256" (default), "SHA384", or "SHA512"`,
```
sgmiller, line 45: Is this capitalization consistent with our other usages of hash names?

builtin/logical/transit/path_import.go
```diff
@@ -0,0 +1,363 @@
...
+				Type: framework.TypeBool,
+				Description: `Enables key derivation mode. This
+allows for per-transaction unique
+keys for encryption operations.`,
+			},
+
+			"convergent_encryption": {
+				Type: framework.TypeBool,
+				Description: `This field is not currently supported for import operations!
+Whether to support convergent encryption.
```
sgmiller, line 66: I know we're testing for this explicitly.  Can we omit it from the docs here, and test for it in data.Raw to reject it?

builtin/logical/transit/path_import.go
```diff
@@ -0,0 +1,363 @@
...
+				Default: "SHA256",
+				Description: `The hash function used as a random oracle in the OAEP wrapping of the user-generated,
+ephemeral AES key. Can be one of "SHA1", "SHA224", "SHA256" (default), "SHA384", or "SHA512"`,
+			},
+			"ciphertext": {
+				Type: framework.TypeString,
+				Description: `The base64-encoded ciphertext of the keys. The AES key should be encrypted using OAEP 
+with the wrapping key and then concatenated with the import key, wrapped by the AES key.`,
+			},
+			"allow_rotation": {
```
sgmiller, line 52: Where/how do we prevent non-automatic e.g. manual rotations of imported keys?

builtin/logical/transit/path_import.go
```diff
@@ -0,0 +1,363 @@
...
+
+	if convergent {
+		return nil, errors.New("import cannot be used on keys with convergent encryption enabled")
+	}
+
+	polReq := keysutil.PolicyRequest{
+		Storage:                  req.Storage,
+		Name:                     name,
+		Derived:                  derived,
+		Convergent:               convergent,
```
sgmiller, line 170: If we aren't allowing convergence, lets just remove it from the policy request.

builtin/logical/transit/path_import.go
```diff
@@ -0,0 +1,363 @@
...
+	ephKey, err := rsa.DecryptOAEP(hashFn, b.GetRandomReader(), privWrappingKey, wrappedEphKey, []byte{})
+	if err != nil {
+		return nil, err
+	}
+
+	// Zero out the ephemeral AES key just to be extra cautious. Note that this
+	// isn't a guarantee against memory analysis! See the documentation for the
+	// `vault.memzero` utility function for more information.
+	defer func() {
+		for i := range ephKey {
```
sgmiller, line 312: Should we just use memzero (even though it doesn't do anything differently?)

builtin/logical/transit/path_import.go
```diff
@@ -0,0 +1,363 @@
...
+			ephKey[i] = 0
+		}
+	}()
+
+	// Ensure the ephemeral AES key is 256-bit
+	if len(ephKey) != 32 {
+		return nil, errors.New("expected ephemeral AES key to be 256-bit")
+	}
+
+	kwp, err := subtle.NewKWP(ephKey)
```
sgmiller, line 322: Oooh, shiny.

sdk/helper/keysutil/policy.go
```diff
@@ -1355,13 +1362,128 @@ func (p *Policy) VerifySignature(context, input []byte, hashAlgorithm HashType,
...
+}
+
 // Rotate rotates the policy and persists it to storage.
 // If the rotation partially fails, the policy state will be restored.
 func (p *Policy) Rotate(ctx context.Context, storage logical.Storage, randReader io.Reader) (retErr error) {
 	priorLatestVersion := p.LatestVersion
 	priorMinDecryptionVersion := p.MinDecryptionVersion
 	var priorKeys keyEntryMap
 
+	if p.Imported && !p.AllowImportedKeyRotation {
```
sgmiller, line 1483: NM, here it is!

sgmiller (changes requested):
Pretty good overall, just a few comments.  Appreciate the thorough test coverage.

builtin/logical/transit/path_import.go
schultz-is: Yeah, snagged it from TOTP: https://www.vaultproject.io/api-docs/secret/totp#algorithm
The `parseHashFn` that's introduced and used here tests against `strings.ToUpper(input)`, though, so it should accept any lower, upper, or some weird combination.

builtin/logical/transit/path_import.go
schultz-is: That was my initial plan, but it's an unexported function in the `vault` package, so I figured it's easier to recreate than it is to export and update any calls to it.

builtin/logical/transit/path_import.go
schultz-is: TIL! Was looking for a way to not mention the field in the docs, but still reject the import request when the field was provided. Thanks!

builtin/logical/transit/path_import.go
schultz-is: Good point; updated.

builtin/logical/transit/path_import.go
```diff
@@ -0,0 +1,346 @@
...
+		Storage:                  req.Storage,
+		Name:                     name,
+		Derived:                  derived,
+		Exportable:               exportable,
+		AllowPlaintextBackup:     allowPlaintextBackup,
+		AutoRotatePeriod:         autoRotatePeriod,
+		AllowImportedKeyRotation: allowRotation,
+	}
+
+	switch keyType {
```
stevendpclark, line 159: nit: since we test within parseHashFn in a case-insensitive fashion, should we do the same here?

sdk/helper/keysutil/lock_manager.go
```diff
@@ -439,6 +442,63 @@ func (lm *LockManager) GetPolicy(ctx context.Context, req PolicyRequest, rand io
...
+	err = p.Import(ctx, req.Storage, key, rand)
+	if err != nil {
+		return fmt.Errorf("error importing key: %s", err)
+	}
+
+	if lm.useCache {
+		lm.cache.Store(req.Name, p)
+	}
+
+	lock.Unlock()
```
stevendpclark, line 497: Shouldn't this have been written as a deferred statement after the lock.Lock() on lie 466? If we error out we aren't releasing the lock correct?

stevendpclark (commented):
Sorry late to the review party, a few comments but overall 👍

PR #15417, opened 2022-05-13
Vault CLI: show detailed information with ListResponseWithInfo

cipherboy:
This adds the ability (via switch `-detailed`) to show the additional metadata returned by endpoints which use the `ListResponseWithInfo` format helper. We add support for both table and JSON formatted response printing. 

See the individual commit messages for examples of each. 

---

An unofficial survey of endpoints returns:

 - PKI: `/issuers` and `/keys`, which are new to 1.11
 - SSH: `/roles`, which has been around for a while. It returns the `key_type` field.
 
A few other uses:

```
vault/identity_store_entities.go:666:			resp := logical.ListResponseWithInfo(keys, entityInfo)
vault/identity_store_entities.go:717:	return logical.ListResponseWithInfo(keys, entityInfo), nil
vault/identity_store_groups.go:533:	return logical.ListResponseWithInfo(keys, groupInfo), nil
vault/identity_store_util.go:2485:	return logical.ListResponseWithInfo(aliasIDs, aliasInfo), nil
vault/logical_system.go:4385:	return logical.ListResponseWithInfo(respKeys, respKeyInfo), nil
vault/login_mfa.go:191:	return logical.ListResponseWithInfo(keys, configInfo), nil
vault/login_mfa.go:746:	return logical.ListResponseWithInfo(keys, configInfo), nil
```

### DIFF
--- a/command/list.go
+++ b/command/list.go
@@ -15,6 +15,8 @@ var (
 
 type ListCommand struct {
 	*BaseCommand
+
+	flagDetailed bool
 }
 
 func (c *ListCommand) Synopsis() string {
@@ -42,7 +44,18 @@ Usage: vault list [options] PATH
 }
 
 func (c *ListCommand) Flags() *FlagSets {
-	return c.flagSet(FlagSetHTTP | FlagSetOutputFormat)
+	set := c.flagSet(FlagSetHTTP | FlagSetOutputFormat)
+	f := set.NewFlagSet("List Options")
+
+	f.BoolVar(&BoolVar{
+		Name:    "detailed",
+		Target:  &c.flagDetailed,
+		Default: false,
+		EnvVar:  "VAULT_DETAILED_LISTS",
+		Usage:   "Enables additional metadata during some LIST operations",
+	})
+
+	return set
 }
 
 func (c *ListCommand) AutocompleteArgs() complete.Predictor {
@@ -117,6 +130,9 @@ func (c *ListCommand) Run(args []string) int {
 		c.UI.Error(fmt.Sprintf("No entries found at %s", path))
 		return 2
 	}
+
+	// Hijack the secret's data to show the list with information.
+	secret.Data[doListWithInfoConstant] = c.flagDetailed
 
 	return OutputList(c.UI, secret)
 }
--- a/command/list.go
+++ b/command/list.go
@@ -131,8 +131,10 @@ func (c *ListCommand) Run(args []string) int {
 		return 2
 	}
 
-	// Hijack the secret's data to show the list with information.
-	secret.Data[doListWithInfoConstant] = c.flagDetailed
+	switch ui := c.UI.(type) {
+	case *VaultUI:
+		ui.detailed = c.flagDetailed
+	}
 
 	return OutputList(c.UI, secret)
 }
--- a/command/list.go
+++ b/command/list.go
@@ -15,8 +15,6 @@ var (
 
 type ListCommand struct {
 	*BaseCommand
-
-	flagDetailed bool
 }
 
 func (c *ListCommand) Synopsis() string {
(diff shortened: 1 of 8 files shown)


command/list.go
```diff
@@ -118,5 +131,8 @@ func (c *ListCommand) Run(args []string) int {
 		return 2
 	}
 
+	// Hijack the secret's data to show the list with information.
+	secret.Data[doListWithInfoConstant] = c.flagDetailed
```
ccapurso, line 135: I am not a huge fan of modifying the response data to add a flag meant to signal how `OutputList` should work. It might be cleaner to branch on `c.flagDetailed` here rather and potentially add a separate function like `OutputListWithDetails` rather than pushing the decision further down. It adds a bit of cognitive overhead as there is now a somewhat hidden flag.

ccapurso:
I provided a comment regarding the flag usage but overall, I _really_ like the concept. As it would have prevented me from having to add a new command recently.

command/list.go
cipherboy: @ccapurso Some thoughts:

1. What do you think about avoiding the `-detailed` flag altogether and just showing the extra info always, if the API endpoint returns it? The one argument in favor (of `-detailed`) might be the JSON format though... If we do that, we wouldn't need the changes here, and it'd make some of the detection logic simpler. The other reason in favor of that is that we wouldn't have to fork _all_ of the List functions (you'd have to fork `OutputList` -> `OutputListWithDetails`, then all of the formatters would have to again be updated to have two methods, mostly duplicating the code except for the extra code we add for Outputs.
2. The other option might be if `c.flagDetailed == false`, we could just do a `delete(secret.Data, "key_info")` and that'd be simpler. Then the details can be unconditionally and we wouldn't need to add the extra flag. Other callers would benefit from it the same way.

The other thought was I was thinking was that it might be neat if we pushed this down server side (i.e., set that flag server side for the `/issuers` and `/keys` endpoints), but that might be an argument for removing the `-detailed` flag altogether and having it just be on the `ListResponseWithInfo`.

Perhaps @ivana-mcconnell has some thoughts on the matter, too?

ccapurso: I imagine that the flag is necessary at this point given that changing the format of the output will introduce a backward incompatible change. As you mention, this is important for the JSON format. If someone pipes to `jq` currently when using `-format=json` for one of these endpoints they will instead get a map whereas before they got an array. In other words, I think an output change like this needs to be opt-in.

I agree, having to fork all the formatters would not be ideal. I wonder if it might make sense to add a flag to the `VaultUI` struct that basically acts as the `vault_cli_do_list_with_info` field. What do you think?

cipherboy: @ccapurso That makes sense. I realize I have no idea how the `ui.format` field gets set, so I had to do it manually in `list.go`. If you know how that gets done, let me know and I can update it.

ccapurso: One reason that I was thinking about having it on the `ui` level is that _maybe_ we would want to support a more expanded output for other commands in the future, similar to the `-o wide` option in `kubectl`. Not sure if others agree, just an idea, of course!
 
In terms of where `ui.format` is set, see [setupEnv](https://github.com/hashicorp/vault/blob/62415ecc4c880b98f9e170efc425c76d3ba3de24/command/main.go#L28).

cipherboy: Oh huh, TIL. Does that mean https://github.com/hashicorp/vault/blob/62415ecc4c880b98f9e170efc425c76d3ba3de24/command/base.go#L57 and https://github.com/hashicorp/vault/blob/62415ecc4c880b98f9e170efc425c76d3ba3de24/command/base.go#L515-L525 are unused?

ccapurso: That is a really good question. I think you can actually use `c.flagFormat` or `Format(c.UI)` to determine the output format. I think that I've used both. I think theoretically they should be same. It looks like `Format(c.UI)` is far more widely used.

cipherboy: Okie dokie, I'll merge things together then :-)

cipherboy:
@ccapurso Updated, let me know what you think. I tried to mimic the existing structure.

ccapurso (approved):
The changes look good to me! Thanks for adding this. You might want to get some feedback from someone on the devex side too.

cipherboy:
Thanks all! I've fixed the broken tests, will merge now.

PR #15420, opened 2022-05-13
append nil months to query get to cover all requested months (OSS)

HridoyRoy:
This PR fills out nil response months in activity log query handle response based on requested month data.

Unit test is in Ent PR: https://github.com/hashicorp/vault-enterprise/pull/2817/files

### DIFF
--- a/vault/activity_log.go
+++ b/vault/activity_log.go
@@ -510,7 +510,8 @@ func (a *ActivityLog) getLastEntitySegmentNumber(ctx context.Context, startTime 
 // WalkEntitySegments loads each of the entity segments for a particular start time
 func (a *ActivityLog) WalkEntitySegments(ctx context.Context,
 	startTime time.Time,
-	walkFn func(*activity.EntityActivityLog, time.Time)) error {
+	walkFn func(*activity.EntityActivityLog, time.Time),
+) error {
 	basePath := activityEntityBasePath + fmt.Sprint(startTime.Unix()) + "/"
 	pathList, err := a.view.List(ctx, basePath)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


vault/activity_log.go
```diff
@@ -1698,12 +1699,40 @@ func (a *ActivityLog) handleQuery(ctx context.Context, startTime, endTime time.T
...
+	if err != nil || err2 != nil {
+		return months
+	}
+	for start.Before(firstMonth) {
+		monthPlaceholder := &ResponseMonth{Timestamp: start.UTC().Format(time.RFC3339)}
+		modifiedResponseMonths = append(modifiedResponseMonths, monthPlaceholder)
+		start = timeutil.StartOfMonth(start.AddDate(0, 1, 0))
+	}
+	modifiedResponseMonths = append(modifiedResponseMonths, months...)
+	for lastMonth.Before(end) {
```
swayne275, line 1728: up until this point we've been building `modifiedResponseMonths` up in chronological order, but here we'll be going in reverse for the months past what's included in `months`

e.g. if `months` goes from june-aug of <some_year>, with `start` being january and `end` being december, this would look like:

[jan feb mar apr may june july aug dec nov oct sept]

is that a problem? e.g. should we instead count from the end of `months` up until `end`, resulting in:

[jan feb mar apr may june july aug sept oct nov dec]

swayne275 (commented):
generally approve, but curious about the comment i made

ccapurso (approved):
Looks good to me!

vault/activity_log.go
hellobontempo: whoops! didn't mean to un-resolve 🥴 was just checking the status

swayne275 (approved):
nice!

PR #15428, opened 2022-05-13
VAULT-5885: Fix erroneous success message in case of two-phase MFA, and provide MFA information in table format

VioletHynes:
Two issues have been fixed:
- The success message being erroneously reported during a two-phase MFA flow
- The MFA information being missing during a two-phase MFA flow unless -format=json was provided

Two-step MFA will now look something like this:
```
violethynes@violethynes-C02CW1WWMD6R Repositories % vault login -method userpass username=alice password=kQ3BlE+5M1libWJc
A login request was issued that is subject to MFA validation. Please make sure
to validate the login by sending another request to sys/mfa/validate endpoint.

WARNING! The VAULT_TOKEN environment variable is set! The value of this
variable will take precedence; if this is unwanted please unset VAULT_TOKEN or
update its value accordingly.

  * A login request was issued that is subject to MFA validation. Please
  make sure to validate the login by sending another request to mfa/validate
  endpoint.

Key                                        Value
---                                        -----
mfa_request_id                             20424e2f-e963-b23d-00de-01f41209fbf9
mfa_constraint_l22_pingid_id               ac25e939-4b7e-41ec-db50-bc65da01a642
mfa_constraint_l22_pingid_uses_passcode    false
mfa_constraint_l22_totp_id                 34dade06-351a-e576-537e-0c7d5b0e6d84
mfa_constraint_l22_totp_uses_passcode      true
```

### DIFF
--- a/changelog/15428.txt
+++ b/changelog/15428.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-auth: Fixed erroneous success message when using vault login in case of two-phase MFA, and fixed MFA information missing from table format when using vault login.
+auth: Fixed erroneous success message and token info when using vault login in case of two-phase MFA, and fixed MFA information missing from table format when using vault login.
 ```
--- a/changelog/15428.txt
+++ b/changelog/15428.txt
@@ -1,3 +1,9 @@
 ```release-note:bug
-auth: Fixed erroneous success message and token info when using vault login in case of two-phase MFA, and fixed MFA information missing from table format when using vault login.
+auth: Fixed erroneous success message when using vault login in case of two-phase MFA
+```
+```release-note:bug
+auth: Fixed erroneous token information being displayed when using vault login in case of two-phase MFA
+```
+```release-note:bug
+auth: Fixed two-phase MFA information missing from table format when using vault login
 ```
--- a/changelog/15428.txt
+++ b/changelog/15428.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+login/mfa: Fixed erroneous success message in case of two-phase MFA, and fixed MFA information missing from table format.
+```
--- a/changelog/15428.txt
+++ b/changelog/15428.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-login/mfa: Fixed erroneous success message when using vault login in case of two-phase MFA, and fixed MFA information missing from table format when using vault login.
+auth: Fixed erroneous success message when using vault login in case of two-phase MFA, and fixed MFA information missing from table format when using vault login.
 ```
--- a/changelog/15428.txt
+++ b/changelog/15428.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-login/mfa: Fixed erroneous success message in case of two-phase MFA, and fixed MFA information missing from table format.
+login/mfa: Fixed erroneous success message when using vault login in case of two-phase MFA, and fixed MFA information missing from table format when using vault login.
 ```
--- a/command/format.go
+++ b/command/format.go
@@ -385,23 +385,6 @@ func (t TableFormatter) OutputSecret(ui cli.Ui, secret *api.Secret) error {
 	}
 
 	if secret.Auth != nil {
-		out = append(out, fmt.Sprintf("token %s %s", hopeDelim, secret.Auth.ClientToken))
-		out = append(out, fmt.Sprintf("token_accessor %s %s", hopeDelim, secret.Auth.Accessor))
-		// If the lease duration is 0, it's likely a root token, so output the
-		// duration as "infinity" to clear things up.
-		if secret.Auth.LeaseDuration == 0 {
-			out = append(out, fmt.Sprintf("token_duration %s %s", hopeDelim, "∞"))
-		} else {
-			out = append(out, fmt.Sprintf("token_duration %s %v", hopeDelim, humanDurationInt(secret.Auth.LeaseDuration)))
-		}
-		out = append(out, fmt.Sprintf("token_renewable %s %t", hopeDelim, secret.Auth.Renewable))
-		out = append(out, fmt.Sprintf("token_policies %s %q", hopeDelim, secret.Auth.TokenPolicies))
-		out = append(out, fmt.Sprintf("identity_policies %s %q", hopeDelim, secret.Auth.IdentityPolicies))
-		out = append(out, fmt.Sprintf("policies %s %q", hopeDelim, secret.Auth.Policies))
-		for k, v := range secret.Auth.Metadata {
-			out = append(out, fmt.Sprintf("token_meta_%s %s %v", k, hopeDelim, v))
-		}
-
 		if secret.Auth.MFARequirement != nil {
 			out = append(out, fmt.Sprintf("mfa_request_id %s %s", hopeDelim, secret.Auth.MFARequirement.MFARequestID))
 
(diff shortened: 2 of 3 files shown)


command/format.go
```diff
@@ -401,6 +401,17 @@ func (t TableFormatter) OutputSecret(ui cli.Ui, secret *api.Secret) error {
 		for k, v := range secret.Auth.Metadata {
 			out = append(out, fmt.Sprintf("token_meta_%s %s %v", k, hopeDelim, v))
 		}
+
+		if secret.Auth.MFARequirement != nil {
+			out = append(out, fmt.Sprintf("mfa_request_id %s %s", hopeDelim, secret.Auth.MFARequirement.MFARequestID))
+
+			for k, constraintSet := range secret.Auth.MFARequirement.MFAConstraints {
```
swayne275, line 408: i think we may want to capture the "and" vs "or" requirements for MFA (right now that's just the "any" requirement, which translates to an "or" of the MFA options).

@hghaf099 do you have thoughts on this? might there be a future where MFA requirements can be AND'd together (right now it sounds like they're just OR'd/any'd)?

command/format.go
```diff
@@ -401,6 +401,17 @@ func (t TableFormatter) OutputSecret(ui cli.Ui, secret *api.Secret) error {
...
+				for _, constraint := range constraintSet.Any {
+					out = append(out, fmt.Sprintf("mfa_constraint_%s_%s_id %s %s", k, constraint.Type, hopeDelim, constraint.ID))
+					out = append(out, fmt.Sprintf("mfa_constraint_%s_%s_uses_passcode %s %t", k, constraint.Type, hopeDelim, constraint.UsesPasscode))
```
swayne275, line 411: good formatting!

command/login.go
```diff
@@ -372,3 +368,14 @@ func (c *LoginCommand) extractToken(client *api.Client, secret *api.Secret, unwr
 		return nil, false, fmt.Errorf("no auth or wrapping info in response")
 	}
 }
+
+// Warn if the VAULT_TOKEN environment variable is set, as that will take
+// precedence. We output as a warning, so piping should still work since it
+// will be on a different stream.
+func (c *LoginCommand) checkForAndWarnAboutLoginToken() {
```
swayne275, line 375: nice!

command/login.go
```diff
@@ -240,6 +240,10 @@ func (c *LoginCommand) Run(args []string) int {
 		c.UI.Warn(wrapAtLength("A login request was issued that is subject to "+
 			"MFA validation. Please make sure to validate the login by sending another "+
 			"request to sys/mfa/validate endpoint.") + "\n")
+
+		// We return early to prevent success message from being printed
```
swayne275, line 244: which code path was being hit before this return?

swayne275 (commented):
a few little things, but looking good!

command/login.go
ccapurso: Agreed, this is great addition! I could imagine having a token set on the environment to in conjunction with issuing a login attempt to be confusing.

changelog/15428.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+login/mfa: Fixed erroneous success message in case of two-phase MFA, and fixed MFA information missing from table format.
```
hghaf099, line 2: Would you think just "auth" be better as the category? I think we went with that before, worth checking it out. 
Also, I guess it would be nice to note the commands in question in the message.

command/format.go
hghaf099: We used to have both "and" and "or" within a single MFAConstraints. However, we deviated from that to just have "or". In cases where a user needs to enforce multiple MFA methods they should define multiple MFAEnforcements.

command/format.go
hghaf099: +1 on that

command/login.go
```diff
@@ -240,6 +240,10 @@ func (c *LoginCommand) Run(args []string) int {
...
+		c.checkForAndWarnAboutLoginToken()
```
hghaf099, line 245: I am not sure if this call is needed here. Cause at this point, there is no Token to be returned as the user would need to validate the MFA. What do you think? Worth discussing with the devex team on this.

command/format.go
```diff
@@ -401,6 +401,17 @@ func (t TableFormatter) OutputSecret(ui cli.Ui, secret *api.Secret) error {
 		for k, v := range secret.Auth.Metadata {
 			out = append(out, fmt.Sprintf("token_meta_%s %s %v", k, hopeDelim, v))
 		}
+
+		if secret.Auth.MFARequirement != nil {
```
hghaf099, line 405: I am wondering if we should print the token related information when the MFA requirements is not nil (which means there is no token returned and the user needs to validate MFA)?

hghaf099 (commented):
This looks great! I just have some questions that might need discussion with the devex team.

command/format.go
VioletHynes: I had the same thought, and it would be pretty easy to not print them. However, the conversation on Slack seemed to lean towards keeping them in: https://hashicorp.slack.com/archives/C025608TAR0/p1652446657666779

command/login.go
VioletHynes: In particular, line 329 - if it continued then it assumes that it is successful.

command/login.go
VioletHynes: The conversation with devex: https://hashicorp.slack.com/archives/C025608TAR0/p1652448116320469 seemed to support this behaviour - in essence, even if there is no token set, the user's primary goal with this command is to set a token, so warning them still makes sense

changelog/15428.txt
VioletHynes: I went with login/mfa to try and make it clear that it was the non-enterprise MFA - but I'm open to changing it. I couldn't find any PR that touched a similar area but had a changelog.

Would you suggest auth/mfa, login/auth, or just auth? 

You're right about including the command in the log - I'll update that!

hghaf099: I would recommend to go with just "auth". The ENT MFA does not have the concept of two phase login, so that part of the message shows which MFA system is fixed.

VioletHynes: Updated

command/format.go
hghaf099: Although I marked this line, my main point is about whether we want to print the token related information when we are following the two-phase login work flow. I think the slack thread is in favor of printing relevant information. 
I feel the same way as in maybe we should avoid printing the following fields when two-phase login workflow is executed:
`token`, `token_accessor`, `token_duration`, `token_renewable`, token_policies`, `identity_policies`, and `policies`. 

In fact, in two-phase login, none of the auth response information is returned to the login command. So, I think that information is not relevant. What do you think about that?

VioletHynes: I agree with you, the whole token section above the section I added could essentially be wrapped in an if/else as it's not relevant for 2FA. I'd be absolutely in favour of making the change (I think we are aligned) but the sense I got from the Slack thread was that this should be discussed a separate issue.

Would you like to leave a comment on the thread with your opinion? If there is alignment that this should be done in this PR I'd be more than happy to make the change!

hghaf099 (approved):
LGTM

command/format.go
VioletHynes: Updated as per discussion in sprint planning. Now, two-phase MFA response might look like this:
```
violethynes@violethynes-C02CW1WWMD6R Repositories % vault login -method userpass username=alice password=kQ3BlE+5M1libWJc
A login request was issued that is subject to MFA validation. Please make sure
to validate the login by sending another request to sys/mfa/validate endpoint.

WARNING! The VAULT_TOKEN environment variable is set! The value of this
variable will take precedence; if this is unwanted please unset VAULT_TOKEN or
update its value accordingly.

  * A login request was issued that is subject to MFA validation. Please
  make sure to validate the login by sending another request to mfa/validate
  endpoint.

Key                                        Value
---                                        -----
mfa_request_id                             20424e2f-e963-b23d-00de-01f41209fbf9
mfa_constraint_l22_pingid_id               ac25e939-4b7e-41ec-db50-bc65da01a642
mfa_constraint_l22_pingid_uses_passcode    false
mfa_constraint_l22_totp_id                 34dade06-351a-e576-537e-0c7d5b0e6d84
mfa_constraint_l22_totp_uses_passcode      true
```

hghaf099: This looks great!

command/format.go
```diff
@@ -385,21 +385,32 @@ func (t TableFormatter) OutputSecret(ui cli.Ui, secret *api.Secret) error {
 	}
 
 	if secret.Auth != nil {
```
hghaf099, line 387: Just a minor comment, otherwise looks great!
Would you think we could avoid at least one level of nesting logic? like combining `secret.Auth != nil` with `secret.Auth.MFARequirement != nil` into a single if statement, then fixing the else part?

hghaf099 (approved):
I posted a comment, otherwise LGTM

command/format.go
VioletHynes: I tried this out and I think I prefer the current formatting - I think it logically flows better, otherwise we have two separate code paths where `Auth != nil`, which doesn't feel as nice to me. I'll lean towards keeping it as-is unless there's a strong inclination to change, I think.

swayne275 (approved):
I think it would be good to split up the changelog, but approved either way!

changelog/15428.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
```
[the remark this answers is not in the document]
VioletHynes: Thanks, good suggestion, I didn't see any example of this in any changelogs that I looked at, I'll make this change :)

Issue #10174, opened 2020-10-19
pki/ca/sign to allow CSR supplied values - to overcome not_before_duration limitation in sign-verbatim


sooslaca:
Hi,

In #10104 I was told use sign instead of sign-verbatim. 
We try that but there are other 'issues' and still cannot achieve dating back `not_before `properly.

" It sounds like if the sign endpoint supported custom x509v3 extensions, there wouldn't be a need to use sign-verbatim in this case" <- correct, we don't need extension
But would need the ability to set CSR supplied O, E and L   (Org, Email, Locality) values.

I understand the purpose of /sign is not to trust user provided values, correct?
But I see https://www.vaultproject.io/api-docs/secret/pki#use_csr_common_name allows CN to be supplied but not other values.

1) would it be possible to have flag for other values as well to be used?
2) would it be possible in sign-verbatim to use `notbefore` as defined in the role?
3) any other solution by which we can achieve the desired certificate to be signed?

Thanks

sooslaca:
To make it more precise reg sign-verbatim, previously I was referred to @jefferai's comment which is:
"Yes, the point of sign-verbatim is that you are taking the values verbatim from the CSR rather than applying role parameters to it."

Which is fine.
Quote: @Naugrimm "The problem is, that notBefore- and notAfter-date of a certificate can not be taken from the CSR, as they cannot be specified"
Eg: in CSR there cannot be such values defined.

Exactly the same problem we ran into, hence suggesting to let be able to define `notbefore` as a json attribute in the request for `sign-verbatim`.

Thanks

cipherboy:
@Gabrielopesantos I re-read this issue and point number two (copying `not_before_duration` from the role) might be a good idea, since [we already copy other relevant TTL fields there](https://github.com/hashicorp/vault/blob/main/builtin/logical/pki/path_issue_sign.go#L205-L211).

This one might be a little easier than the other one. :-)

Gabrielopesantos:
With point two you mean this one "would it be possible in sign-verbatim to use `notbefore` as defined in the role?", right?
I agree, does not seem as complex as the other. I'll start with this one.

cipherboy:
@Gabrielopesantos Yep! And thank you! Let me know if you have any questions.

Gabrielopesantos:
Hey, investigated a bit and, am I missing something or this is just adding an `if` statement for the `NotBeforeDuration` field just like the `TTL` or `MaxTTL`?

[...]

PR #15429, opened 2022-05-14
pki/sign-verbatim uses role not before duration

Gabrielopesantos:
## Description
pki/sign-verbatim uses `not_before_duration` defined role.
Related to issue: #10174

### DIFF
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -2178,9 +2178,10 @@ func runTestSignVerbatim(t *testing.T, keyType string) {
 
 	// create a role entry; we use this to check that sign-verbatim when used with a role is still honoring TTLs
 	roleData := map[string]interface{}{
-		"ttl":      "4h",
-		"max_ttl":  "8h",
-		"key_type": keyType,
+		"ttl":                 "4h",
+		"max_ttl":             "8h",
+		"key_type":            keyType,
+		"not_before_duration": "2h",
 	}
 	resp, err = b.HandleRequest(context.Background(), &logical.Request{
 		Operation:  logical.UpdateOperation,
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -2248,7 +2248,10 @@ func runTestSignVerbatim(t *testing.T, keyType string) {
 	}
 	cert := certs[0]
 	if math.Abs(float64(time.Now().Add(12*time.Hour).Unix()-cert.NotAfter.Unix())) < 10 {
-		t.Fatalf("sign-verbatim did not properly cap validity period on signed CSR")
+		t.Fatalf("sign-verbatim did not properly cap validity period (notAfter) on signed CSR: was %v vs requested %v but should've been %v", cert.NotAfter, time.Now().Add(12*time.Hour), time.Now().Add(8*time.Hour))
+	}
+	if math.Abs(float64(time.Now().Add(-2*time.Hour).Unix()-cert.NotBefore.Unix())) > 10 {
+		t.Fatalf("sign-verbatim did not properly cap validity period (notBefore) on signed CSR: was %v vs expected %v", cert.NotBefore, time.Now().Add(-2*time.Hour))
 	}
 
 	// Now check signing a certificate using the not_after input using the Y10K value
--- a/changelog/15429.txt
+++ b/changelog/15429.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-secrets/pki: pki/sign-verbatim uses not_before_duration defined role
+secrets/pki: Add support for pki/sign-verbatim use the not_before_duration field defined in the role
 ```
--- a/changelog/15429.txt
+++ b/changelog/15429.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+secrets/pki: pki/sign-verbatim uses not_before_duration defined role
+```
--- a/changelog/15429.txt
+++ b/changelog/15429.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-secrets/pki: Add support for pki/sign-verbatim use the not_before_duration field defined in the role
+secrets/pki: Have pki/sign-verbatim use the not_before_duration field defined in the role
 ```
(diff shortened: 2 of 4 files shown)


builtin/logical/pki/backend_test.go
```diff
@@ -2178,9 +2178,10 @@ func runTestSignVerbatim(t *testing.T, keyType string) {
 
 	// create a role entry; we use this to check that sign-verbatim when used with a role is still honoring TTLs
 	roleData := map[string]interface{}{
-		"ttl":      "4h",
-		"max_ttl":  "8h",
-		"key_type": keyType,
+		"ttl":                 "4h",
+		"max_ttl":             "8h",
+		"key_type":            keyType,
+		"not_before_duration": "2h",
```
cipherboy, line 2184: One small test improvement :-) If you look below around line 2250, we test that the requested 12h TTL wasn't honored (and is presumably 8h then).

If you don't mind, it'd be great to add a test to make sure the 2h not before duration was honored though :-) 

I ended up doing something like:

```diff
diff --git a/builtin/logical/pki/backend_test.go b/builtin/logical/pki/backend_test.go
index b1402ac82b..6ddd8b4a70 100644
--- a/builtin/logical/pki/backend_test.go
+++ b/builtin/logical/pki/backend_test.go
@@ -2248,7 +2248,10 @@ func runTestSignVerbatim(t *testing.T, keyType string) {
        }
        cert := certs[0]
        if math.Abs(float64(time.Now().Add(12*time.Hour).Unix()-cert.NotAfter.Unix())) < 10 {
-               t.Fatalf("sign-verbatim did not properly cap validity period on signed CSR")
+               t.Fatalf("sign-verbatim did not properly cap validity period (notAfter) on signed CSR: was %v vs requested %v but should've been %v", cert.NotAfter, time.Now().Add(12*time.Hour), time.Now().Add(8*time.Hour))
+       }
+       if math.Abs(float64(time.Now().Add(-2*time.Hour).Unix()-cert.NotBefore.Unix())) > 10 {
+               t.Fatalf("sign-verbatim did not properly cap validity period (notBefore) on signed CSR: was %v vs expected %v", cert.NotBefore, time.Now().Add(-2*time.Hour))
        }
 
        // Now check signing a certificate using the not_after input using the Y10K value
```

Otherwise, this looks good :-)

changelog/15429.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+secrets/pki: pki/sign-verbatim uses not_before_duration defined role
```
victorr, line 2: This is a bit terse perhaps?

builtin/logical/pki/backend_test.go
Gabrielopesantos: Well, the PR  was "ready" to be reviewed but still, thanks for feedback.
I think I am a bit confused with these `ttl` and `max_ttl` values. In the first `if` clause we are validating if the `ttl` defined in CSR (12 hours) is not honored, by checking if the difference between `notAfter` value and `now + 12 hours` is more than 10 seconds, right?

My biggest confusion is with the second `if` statement, we want to validate if the 2h `not_before` is honored. Isn't checking that the difference between `notBefore` and `now - 2 hours` is less than 10 seconds what we are looking for?

I have printed the `notAfter` and `notBefore` values of a certificate created now (2022/05/16 17:56:33 in my timezone) and `notAfter` has the value of `now` + 7 hours and `notBefore` now - 2 hours. Without considering the the `notBefore` that's 8 around which is the role `max_ttl`. Does that mean that `notBefore` is not included in the `ttl`? And `notBefore` has always 1 hour by default? These values are based on the test I am reviewing.

I should have known this already but a clarification would be appreciated. Thanks

changelog/15429.txt
Gabrielopesantos: Hello, how about "Add support for pki/sign-verbatim use the not_before_duration field defined in the role"?

builtin/logical/pki/backend_test.go
cipherboy: Correct, `notBefore` is not included in the TTL. Max TTL is the maximum allowable TTL the certificate can be valid for (from server time now, to the end of the cert.NotAfter period). TTL is the _default_ TTL the certificate is valid for (again, from server now). 

So, this role says, by default, issue for 4 hours, but let users request up to 8 hour validity (from now).

By allowing a 2 hour backdate, if the servers were somehow that far behind, they could theoretically request up to a 10 hour valid certificate (two in the past for clock sync issues, and 8 more in the future). 

So either way, the requested 12h validity is too much, and is truncated to 8 hours.

---

So, back to your question, we want the success condition to be `(our not_before_duration was respected)`. This expands to something like `(delta (two hours before now) to (cert.NotBefore)) == 0`. However, some of these checks take some amount of time, in, e.g., slow CI systems, so we check `(delta ...) <= 10 seconds` to give some time. Also, we throw in an absolute value for some reason. Maybe clock skew or something, I dunno.

Negating that, gives us `abs(delta) > 10 seconds` as the conditional for which we should abort the test (`t.Fatalf(...)`) on. So I think it is correct?

changelog/15429.txt
cipherboy: Not to nitpick this to death, but maybe `Have pki/sign-verbatim use the ...`? This isn't conditional is why I suggest that.

Gabrielopesantos: Makes sense.

builtin/logical/pki/backend_test.go
Gabrielopesantos: Yes, makes total sense. I was confusing everything in the beginning.

cipherboy: If you want, it'd be totally appropriate to add a note to the [API docs](https://github.com/hashicorp/vault/blob/main/website/content/api-docs/secret/pki.mdx#createupdate-role) to notate that `not_before_duration` doesn't impact the `ttl` and that the `ttl` is strictly used for future validity (now to NotAfter).

Gabrielopesantos: Good idea, I think that would avoid possible confusions from others.
Just pushed a proposal with the additions.

cipherboy (approved):
Looks great! Thank you!

PR #15432, opened 2022-05-16
MFA method form

arnav28:
- Updated model for form attributes
- Form for editing, creating mfa methods

Screen Shot 2022-05-16 at 12 04 53 AM

Screen Shot 2022-05-16 at 12 05 02 AM

### DIFF
--- a/ui/app/adapters/mfa-method.js
+++ b/ui/app/adapters/mfa-method.js
@@ -9,8 +9,8 @@ export default class MfaMethodAdapter extends ApplicationAdapter {
 
   createOrUpdate(store, type, snapshot) {
     const data = store.serializerFor(type.modelName).serialize(snapshot);
-    let id = snapshot.id;
-    return this.ajax(`${this.urlForQuery(type.modelName, snapshot)}/${data.type}/${id}`, 'POST', {
+    const { id } = snapshot;
+    return this.ajax(this.buildURL(type.modelName, data.type, id, 'POST'), 'POST', {
       data,
     }).then(() => {
       // TODO: Check how 204's are handled by ember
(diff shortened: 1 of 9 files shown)


ui/app/adapters/mfa-method.js
```diff
@@ -28,4 +39,12 @@ export default class MfaMethodAdapter extends ApplicationAdapter {
...
     });
   }
+
+  buildURL(modelName, id, snapshot, requestType) {
+    if (requestType === 'POST') {
+      let url = `${super.buildURL(modelName)}/${snapshot.attr('type')}`;
+      return id ? `${url}/${id}` : url;
+    }
+    return super.buildURL(...arguments);
+  }
```
zofskeez, line 49: :tada:

ui/app/adapters/mfa-method.js
```diff
@@ -7,19 +7,30 @@ export default class MfaMethodAdapter extends ApplicationAdapter {
...
+    return this.ajax(this.buildURL(type.modelName, id, snapshot, 'POST'), 'POST', {
+      data,
+    }).then(() => {
+      // TODO: Check how 204's are handled by ember
+      return {
+        data: {
+          id,
+          ...data,
+        },
+      };
```
zofskeez, line 22: I'm not sure this will work as expected. In the case of creating a new record the id from the snapshot will be undefined but the id is returned in the response, whereas updating returns no content. Also, I don't think we want to return the data under the data key.

zofskeez (commented):
Nice work! Just a small note on the adapter response when creating/updating.

ui/app/adapters/mfa-method.js
zofskeez: Following up after looking at the serializer. The `normalizeItems` method expects a data object in the payload so disregard that part of the comment.

PR #15434, opened 2022-05-16
api: make ListPlugins parse only known plugin types

tomhjp:
Currently, adding a new type to the top-level keys of response data will break client code if it's not a plugin type that the client already knows about. This PR makes it only iterate over known plugin types instead to make it more resilient to any future changes.


swenson (approved):
LGTM

PR #15435, opened 2022-05-16
Include default information in LIST keys, issuers

cipherboy:
This shows whether the specified key or issuer is default, along with
the private key type in the case of a `LIST /keys` (authenticated) call.

### DIFF
--- a/builtin/logical/pki/path_fetch_keys.go
+++ b/builtin/logical/pki/path_fetch_keys.go
@@ -44,6 +44,11 @@ func (b *backend) pathListKeysHandler(ctx context.Context, req *logical.Request,
 		return nil, err
 	}
 
+	config, err := getKeysConfig(ctx, req.Storage)
+	if err != nil {
+		return nil, err
+	}
+
 	for _, identifier := range entries {
 		key, err := fetchKeyById(ctx, req.Storage, identifier)
 		if err != nil {
@@ -53,6 +58,8 @@ func (b *backend) pathListKeysHandler(ctx context.Context, req *logical.Request,
 		responseKeys = append(responseKeys, string(identifier))
 		responseInfo[string(identifier)] = map[string]interface{}{
 			keyNameParam: key.Name,
+			"is_default": identifier == config.DefaultKeyId,
+			"key_type":   key.PrivateKeyType,
 		}
 
 	}
--- a/builtin/logical/pki/path_fetch_issuers.go
+++ b/builtin/logical/pki/path_fetch_issuers.go
@@ -39,6 +39,11 @@ func (b *backend) pathListIssuersHandler(ctx context.Context, req *logical.Reque
 		return nil, err
 	}
 
+	config, err := getIssuersConfig(ctx, req.Storage)
+	if err != nil {
+		return nil, err
+	}
+
 	// For each issuer, we need not only the identifier (as returned by
 	// listIssuers), but also the name of the issuer. This means we have to
 	// fetch the actual issuer object as well.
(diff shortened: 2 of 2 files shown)


builtin/logical/pki/path_fetch_keys.go
```diff
@@ -53,6 +58,8 @@ func (b *backend) pathListKeysHandler(ctx context.Context, req *logical.Request,
 		responseKeys = append(responseKeys, string(identifier))
 		responseInfo[string(identifier)] = map[string]interface{}{
 			keyNameParam: key.Name,
+			"is_default": identifier == config.DefaultKeyId,
+			"key_type":   key.PrivateKeyType,
```
stevendpclark, line 62: A bit late but this will output `ManagedPrivateKey` instead of a proper value? Do you want me to fix this later within my upcoming PR that I fixed it within the GET output?

stevendpclark: It will also slow it down most likely if we have a lot of managed keys, but at least we will be consistent across the apis.

cipherboy: I was looking at our existing GET code, and unless I was mistaken, we'll return the managed key type there as well:

https://github.com/hashicorp/vault/blob/ea3365d29632f5301f83750801599d90de206b51/builtin/logical/pki/path_fetch_keys.go#L149-L155

Am I missing something?

stevendpclark: Yeah my upcoming PR fixes this to return the proper key type based on the managed key's public key.

cipherboy: Ah I see, I missed that. :-) I dunno, I'm kinda torn tbh. I think the LIST operation should remain fast generally, so I'm inclined not to fetch it?

Maybe for consistency, the managed key's type should be placed in a separate field? But I guess the fact that it is a managed key can be inferred from the presence of the uuid, I dunno.

stevendpclark: I'd much prefer that key_type is consistent across the board within the api, and so that that has always been the real types, managed key type has never appeared within it. Can we just drop the key_type from the list operation?

cipherboy: Sure, we could do that.

What do you mean by "managed key type has never appeared within it"? As far as I'm aware, until the multi-issuers change, we never had any indication of key types at all, except on creation? As a result, there was never a way to tell the backing _algorithm_ of the key, except via fetching `/cert/ca` and parsing the certificate. But on 1.10 with the introduction of managed keys, there was no way of telling the creation type (`type=internal` vs `type=kms`) afterwards.

So net, I think we're free to make whatever decision we want to. I'm fine with `key_type` being strictly the algorithm, but it'd also be nice to have a sense of if a given key was a managed key or not up front.

Maybe change the field to `is_kms` on the LIST?

stevendpclark: All correct statements, what I meant about the `key_type` was all existing apis that return key_type in the past (mainly the generate Root and generate Intermediate off the top of my head), always returned the algorithm. 

At least now they will be able to tell within the new PR upcoming if it was a managed key type based on the new fields being added to the `GET /key/<key id>` response. 

I'm fine with the `is_kms` approach as well, not entirely sure if there is a huge win from it though either.

cipherboy: Yeah, that's fair, they could just name the key indicating it is HSM backed. Eh, lets elide it. So much for that idea. :-D

stevendpclark: To close the loop, the `key_type` field within the response from removed within #15468

PR #15438, opened 2022-05-16
Fix UBI setting environment variable

cipherboy:
The `VAULT_DISABLE_FILE_PERMISSIONS_CHECK` presently only takes the value `true` and doesn't use `ParseBool`; change the behavior to match most other environment variables (since it is new), but also update the UBI image to use the literal `true`. 

---

@akshya96 If this will be reverted, I'm happy to remove the updates to the config. But since its still here, I figured I'd aim for consistency. I think, if you wanted to avoid a complete revert and allow users to manually enable the feature, you could update this new code to just set `skipCheckPermissions := true` by default and then just update the docs. 

My 2c.


tvoran (approved):
The entrypoint change looks good to me.

PR #15440, opened 2022-05-16
Add default timeout to legacy ssh.ClientConfig

cipherboy:
When using the deprecated Dynamic SSH Keys method, Vault will make an
outbound SSH connection to an arbitrary remote host to place SSH keys.
We now set a timeout of 1 minute for this connection.

It is strongly recommended consumers of this SSH secrets engine feature
migrate to the more secure, and otherwise equivalent, SSH certificates
method.


sgmiller (approved):

cipherboy:
Thanks! Merging...

PR #15442, opened 2022-05-16
Update considerations for multiple issuers

cipherboy:
Also adds auditing section about suggested un-HMAC'd request/response
parameters.

---

Let me know if I missed anything, but I don't think the other docs pages mention single issuer as a restriction (so there should be no need to update them).


sgmiller (approved):

cipherboy:
Thanks all! Merging...

PR #15443, opened 2022-05-16
Fix ubi-docker-entrypoint.sh->docker-entrypoint.sh

cipherboy:
It is believed to break the helm chart; see also:

https://github.com/hashicorp/vault/pull/15272#discussion_r873927744


tvoran (approved):
It works! Thanks.

cipherboy:
Thanks for your review! Merging... :-)

PR #15444, opened 2022-05-16
docs: fix typo in jwt k8 provider

jasonodonnell:
A small typo was found in the JWT auth documentation for Kubernetes. In the example given, the ISSUER env is set to the address of the OIDC well-known endpoint. This should instead be `https://kubernetes.default.svc`.


tomhjp:
Which distributions did you test on? Unfortunately I think this varies between the various different providers

jasonodonnell:
@tomhjp I tested this originally on EKS and Kind, but just double checked on GKE and you're right, it varies. I'm looking how to identify the audience easily so I can update the documentation to do so.

PR #15449, opened 2022-05-16
Store migrated issuer, key ID in migration log

cipherboy:
If necessary, this will let us correlate migrated values afterwards.

---

@sgmiller and @stevendpclark I think we talked about this but never got around to doing it. I think it was supposed to be done around the same time as adding the (present) issuer to the certificate storage entry, but because that didn't happen (wasn't a fat format), this didn't happen either.

I opted to add it to the log, even though there's no (present) way of returning the log info to the user, rather than to the issuers config. Thoughts?


stevendpclark (approved):
This seems good to me. We could easily cache this info in the future within the backend if we need to use it.

cipherboy:
Thanks! Merging...

PR #15451, opened 2022-05-16
Log less around the current status of the PKI migration

stevendpclark:
- No point in writing any logs if no previous bundle exists
 - Only log output and schedule a CRL rebuild is we actually migrated something
 - Do not log on PKI storage version set/checks.

Logs when needing to migration an existing bundle

```
2022-05-16T17:19:12.085-0400 [INFO]  secrets.pki.pki_e3586de1: performing PKI migration to new keys/issuers layout
2022-05-16T17:19:12.086-0400 [INFO]  secrets.pki.pki_e3586de1: Migration generated the following ids and set them as defaults: issuer id=691b8102-1e26-ddc3-a891-5ac8145ce081 key id=7396a3a8-f72a-4eaf-e6fc-547f295a45d9
2022-05-16T17:19:12.086-0400 [INFO]  secrets.pki.pki_e3586de1: Scheduling PKI CRL rebuild.
```

There are no logs printed out if there isn't an existing bundle.

We output the following on non-active nodes when a migration has completed.
```
2022-05-16T17:19:12.485-0400 [INFO]  secrets.pki.pki_e3586de1: Detected a migration completed, resetting pki storage version
```

### DIFF
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -344,10 +344,8 @@ func (b *backend) updatePkiStorageVersion(ctx context.Context, grabIssuersLock b
 	}
 
 	if info.isRequired {
-		b.Logger().Info("PKI migration is required, reading cert bundle from legacy ca location")
 		b.pkiStorageVersion.Store(0)
 	} else {
-		b.Logger().Debug("PKI migration previously completed, reading cert bundle from key/issuer storage")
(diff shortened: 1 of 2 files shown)


stevendpclark:
Fixed merge conflicts from main

cipherboy (approved):
I think this addresses my comments :-) Thanks Steve!

PR #15452, opened 2022-05-16
Vault-6037 making filesystem permissions check opt-in

akshya96:
https://hashicorp.atlassian.net/browse/VAULT-6037
Making this change opt-in, by renaming the environment variable VAULT_DISABLE_FILE_PERMISSIONS_CHECK to VAULT_ENABLE_FILE_PERMISSIONS_CHECK and adjusting the logic

### DIFF
--- a/changelog/15452.txt
+++ b/changelog/15452.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: renaming the environment variable VAULT_DISABLE_FILE_PERMISSIONS_CHECK to VAULT_ENABLE_FILE_PERMISSIONS_CHECK and adjusting the logic
+```
--- a/command/operator_diagnose_test.go
+++ b/command/operator_diagnose_test.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/hashicorp/vault/sdk/helper/consts"
 	"github.com/hashicorp/vault/vault/diagnose"
 	"github.com/mitchellh/cli"
 )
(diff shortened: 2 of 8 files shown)


HridoyRoy (commented):
Looks good overall -- just wondering if we are ok with removing the `consts.VaultDisableFilePermissionsCheckEnv` environment variable entirely, as it seems like this change is also going to be backported into minor versions?

akshya96:
> Looks good overall -- just wondering if we are ok with removing the `consts.VaultDisableFilePermissionsCheckEnv` environment variable entirely, as it seems like this change is also going to be backported into minor versions?
Looks like these changes were removed from back ports

HridoyRoy (approved):
Lgtm!

cipherboy:
@akshya96 As I mentioned in #15438, do you think we can switch this from a literal value to `strconv.ParseBool` [like most of the other server envvars are](https://github.com/hashicorp/vault/blob/main/command/server.go#L1161-L1184)? This'd give us a little more consistency :-)

akshya96:
> @akshya96 As I mentioned in #15438, do you think we can switch this from a literal value to `strconv.ParseBool` [like most of the other server envvars are](https://github.com/hashicorp/vault/blob/main/command/server.go#L1161-L1184)? This'd give us a little more consistency :-)
> 
> I'm happy to open a PR once this lands if you'd prefer.

Sure, It makes sense to be more consistent. Will make this change and update the PR

cipherboy:
Thank you!!

cipherboy (approved):
Thank you! :100:

PR #15467, opened 2022-05-17
Add more CA usage best practices

cipherboy:
First pass at some additional CA best practices.

Still missing:

 - [x] ACLing of paths by persona

### DIFF
--- a/website/content/docs/secrets/pki/considerations.mdx
+++ b/website/content/docs/secrets/pki/considerations.mdx
@@ -180,6 +180,40 @@ associated leases, to prevent unintended revocation when not using a token with
 a long enough lifetime. To revoke these certificates, use the `pki/revoke`
 endpoint.
 
+## Safe Usage of Roles
+
+The Vault PKI Secrets Engine supports many options to limit issuance via
+[Roles](http://localhost:3000/api-docs/secret/pki#create-update-role).
+Careful consideration of construction is necessary to ensure that more
+permissions are not given than necessary. Additionally, roles should generally
+do _one_ thing; multiple roles should be preferable over having too permissive
+roles that allow arbitrary issuance (e.g., `allow_any_name` should generally
+be used sparingly, if at all).
+
+ - `allow_any_name` should generally be set to `false`; this is the default.
+ - `allow_localhost` should generally be set to `false` for production
+   services, unless listening on `localhost` is expected.
+ - Unless necessary, `allow_wildcard_certificates` should generally be set to
+   `false`. This is **not** the default due to backwards compatibility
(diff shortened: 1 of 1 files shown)


website/content/docs/secrets/pki/considerations.mdx
```diff
@@ -217,6 +297,71 @@ nature:
...
+   revocation on behalf of an operator; may also handle delegated
+   issuance.
+ - *Advanced*; potentially a power-user that has access to additional
+   issuance APIs.
+ - *Requester*; a low-level user that simply requests certificates.
+ - *Unauthed*; any arbitrary user or service that lacks a Vault token.
+
+For these personas, we suggest the following ACLs, in condensed, tabular form:
+
+| Path | Operations | Operator | Agent | Advanced | Requester | Unauthed |
```
cipherboy, line 323: @ivana-mcconnell made a comment out of band that these might be good to have per-persona; currently they're sorted as such (so all the persona's paths are grouped together and you can then read across the table to find exactly those paths), but I'm not sure if it makes sense expanding these per-persona explicitly (duplicating endpoints across various personas). Does anyone know if we can do those fancy tables like on the tutorials pages (which I don't think are hosted on this repo)? 

This present table makes the additive nature visually clear, but sadly it is already too long to fit on a single (horizontal) screen, so maybe it doesn't matter and breaking it up would be more useful in general.

Should we make a better effort to make these in an ACL-able format (i.e., remove the regex patterns)? Or should we explicitly leave them in regex-style paths, to force people to think about the ACL actions they're potentially choosing here, as they are in general, risky?

Thoughts?

stevendpclark: This is really nice, but as you say quite long sadly. It's even missing some paths :( Pre-canned policies that enforce these personas might be interesting?

issuer/:issuer_ref/sign-intermediate
issuer/:issuer_ref/sign-self-issued
root/rotate/:exported

cipherboy: The concern with pre-canned policies in particular is that you might rarely want to actually open up _all_ users to issue against _all_ roles; I think more securely, we'd recommend limiting users to specific roles, which is up to the operator to design and implement.

That alone makes the pre-canned approach harder, as we're just doing the "easy" (operator/agent) side and leaving the (advanced/requester) roles more broad.

stevendpclark: Fair, but we could have a pre-canned Requester policy with <:role_name:> or something in that regard to push them down that path?

stevendpclark (approved):
💯

PR #15468, opened 2022-05-17
secret/pki: Return correct algorithm type from key fetch API for managed keys

stevendpclark:
Fix an issue that key_type field returned from the key fetch API had the ManagedPrivateKey value instead of the real algorithm of the managed key.

Also return additional information about the managed key such as the managed key name and it's UUID.


cipherboy:
@stevendpclark Did you want to undo my list addition here too? Or does that have to be a separate PR?

cipherboy (approved):
Match of Ent PR.

stevendpclark:
> @stevendpclark Did you want to undo my list addition here too? Or does that have to be a separate PR?

Ah yes let me do that here as well. I'll add another commit to this PR.

cipherboy (approved):
Thanks!

PR #15469, opened 2022-05-17
possibly forward cached MFA auth response to leader

hghaf099:
This PR is the OSS part of this [one](https://github.com/hashicorp/vault-enterprise/pull/2749).
It removes the logic for forwarding all login requests from a perfStandby node to the active node. Instead, it enables forwarding cached MFA auth response to the leader so that MFA validation works.


ccapurso (approved):
Looks good to me

PR #15470, opened 2022-05-17
Always return PKI configs for CRLs, URLs

cipherboy:
Vault currently returns a `nil` response when the default CRL and URL configuration is being used. This doesn't really make sense, especially in the CRL case, as this default CRL config actually has 72h expiration (and isn't disabled). 

Instead, always return config structs for both of these two configs. This matches the behavior introduced in the multi-issuer changes  where `config/issuers` and `config/keys` always returns structs (even if empty of values).


stevendpclark (approved):

cipherboy:
Thanks @stevendpclark!

PR #15473, opened 2022-05-17
Vault 6122 pki role issuer name validation

kitography:
This covers two things:
* @cipherboy 's comments that we should probably say something if the floating reference referred to by a role does not exist. 
* @ivana-mcconnell comments during review that it would be nice if we warned when we were deleting a floating reference to an issuer that was used by a role.

### DIFF
--- a/builtin/logical/pki/path_fetch_issuers.go
+++ b/builtin/logical/pki/path_fetch_issuers.go
@@ -410,6 +410,9 @@ func (b *backend) pathDeleteIssuer(ctx context.Context, req *logical.Request, da
 	response = &logical.Response{}
 
 	issuer, err := fetchIssuerById(ctx, req.Storage, ref)
+	if err != nil {
+		return nil, err
+	}
 	if issuer.Name != "" {
 		addWarningOnDereferencing(issuer.Name, response, ctx, req.Storage)
 	}
--- a/builtin/logical/pki/path_fetch_issuers.go
+++ b/builtin/logical/pki/path_fetch_issuers.go
@@ -248,7 +248,9 @@ func (b *backend) pathUpdateIssuer(ctx context.Context, req *logical.Request, da
 
 	modified := false
 
+	var oldName string
 	if newName != issuer.Name {
+		oldName = issuer.Name
 		issuer.Name = newName
 		modified = true
 	}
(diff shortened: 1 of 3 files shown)


builtin/logical/pki/storage.go
```diff
@@ -27,6 +27,9 @@ const (
 
 	// Used as a quick sanity check for a reference id lookups...
 	uuidLength = 36
+
+	maxRolesToScanOnIssuerChange = 10000
```
cipherboy, line 31: nit: maybe drop this to 50 or so? This could incur a lot of overhead. /cc @sgmiller

cipherboy (commented):
Looks fairly good. It looks like the storage migration test [expects a nil response](https://github.com/hashicorp/vault/blob/main/builtin/logical/pki/storage_migrations_test.go#L175), but we now return a warning. 

Strictly, I think we should perhaps condition the resolution of issuer reference (on roles) to being on the newer storage version? My 2c.

builtin/logical/pki/path_roles.go
```diff
@@ -726,6 +726,18 @@ func (b *backend) pathRoleCreate(ctx context.Context, req *logical.Request, data
 	if len(entry.Issuer) == 0 {
 		entry.Issuer = defaultRef
 	}
+	// Check that the issuers reference set resolves to something
+	issuerId, err := resolveIssuerReference(ctx, req.Storage, entry.Issuer)
```
stevendpclark, line 730: Missing err check here.

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -399,15 +406,22 @@ func (b *backend) pathDeleteIssuer(ctx context.Context, req *logical.Request, da
 		return logical.ErrorResponse("unable to resolve issuer id for reference: " + issuerName), nil
 	}
 
+	var response *logical.Response
+	response = &logical.Response{}
+
+	issuer, err := fetchIssuerById(ctx, req.Storage, ref)
```
stevendpclark, line 412: Missing err check

builtin/logical/pki/storage.go
```diff
@@ -889,3 +892,37 @@ func isKeyInUse(keyId string, ctx context.Context, s logical.Storage) (inUse boo
...
+	}
+
+	inUseBy = 0
+	checkedRoles := 0
+
+	for _, roleName := range roleEntries {
+		entry, err := storage.Get(ctx, "role/"+roleName)
+		if err != nil {
+			return false, 0, err
+		}
```
cipherboy, line 909: And I think this is the missing `nil` check that semgrep is complaining about.

stevendpclark (approved):
👍

PR #15474, opened 2022-05-17
UI: Better default transit auto-rotation

hashishaw:
Update the default transit auto-rotation from `1h` to `30d`, to represent a more reasonable default value for keys.

Screen Shot 2022-05-17 at 11 08 41 AM

### DIFF
--- a/ui/lib/core/addon/components/ttl-picker2.js
+++ b/ui/lib/core/addon/components/ttl-picker2.js
@@ -74,7 +74,14 @@ export default TtlForm.extend({
 
     if (typeOf(value) === 'number') {
       // if the passed value is a number, assume unit is seconds
-      time = value;
+      if (value % secondsMap.d === 0) {
+        unit = 'd';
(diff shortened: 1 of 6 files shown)


ui/lib/core/addon/components/ttl-picker2.js
```diff
@@ -74,7 +74,14 @@ export default TtlForm.extend({
 
     if (typeOf(value) === 'number') {
       // if the passed value is a number, assume unit is seconds
```
Monkeychip, line 76: does this comment still make sense? Because we're not assuming that it's seconds anymore, correct?

Monkeychip (approved):
LGTM. Only one clarifying question.

ui/lib/core/addon/components/ttl-picker2.js
hashishaw: We're still assuming that the number passed is referring to the number of seconds, not a number of minutes for example. I can add another comment explaining what the next chunk is doing though

PR #15475, opened 2022-05-17
Start documentation for FIPS variants of Vault Enterprise

cipherboy:
This restructures the FIPS documentation a little, turning it into a proper section, and adds the FIPS 140-2 inside documentation. 

We add notes to various other documentation sections, stating which algorithms cannot compliantly be used with the FIPS edition; these still will work, but aren't certified/certifiable.


sgmiller (approved):
Nothing to add, looks good.

PR #15481, opened 2022-05-17
Protect against key and issuer name re-use

stevendpclark:
- While importing keys and issuers verify that the provided name if any has not been used by another key that we did not match against.
 - Validate an assumption within the key import api, that we were provided a single key
 - Add additional tests on the new key generation and key import handlers.

### DIFF
--- a/builtin/logical/pki/path_manage_keys.go
+++ b/builtin/logical/pki/path_manage_keys.go
@@ -1,7 +1,9 @@
 package pki
 
 import (
+	"bytes"
 	"context"
+	"encoding/pem"
 	"strings"
 
 	"github.com/hashicorp/vault/sdk/framework"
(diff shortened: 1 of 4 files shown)


builtin/logical/pki/path_manage_keys.go
```diff
@@ -171,9 +173,9 @@ func pathImportKey(b *backend) *framework.Path {
 }
 
 const (
-	pathImportKeyHelpSyn  = `Import the specified key.`
+	pathImportKeyHelpSyn  = `Import a specified key.`
```
cipherboy, line 176: nit: I think `the` is correct fwiw.

stevendpclark: Fixed.

cipherboy (approved):

cipherboy:
Oh, while you're here and this hasn't merged, do you want to add this to the API docs since its missing from the import CA/keys section? Just add a new row I think is fine and note the limitation of one key.

PR #15485, opened 2022-05-17
Enforcement view at MFA method level

arnav28:
- List enforcements for each mfa method
- Delete MFA method if no enforcements are present
- Moved method, enforcement list item component to mfa folder

Screen Shot 2022-05-17 at 2 08 27 PM

Screen Shot 2022-05-17 at 2 08 39 PM


ui/app/routes/vault/cluster/access/mfa/methods/method.js
```diff
@@ -1,8 +1,22 @@
 import Route from '@ember/routing/route';
-
+import { hash } from 'rsvp';
```
Monkeychip, line 2: Jordan might have some thoughts on Hash. I think there's a work around he mentioned, but I know we're under a time crunch so this is more of a if you're interested reach out to him about it.

ui/app/routes/vault/cluster/access/mfa/methods/method.js
```diff
@@ -1,8 +1,22 @@
...
 export default class MfaMethodRoute extends Route {
   model({ id }) {
-    return this.store.findRecord('mfa-method', id);
+    return hash({
+      method: this.store.findRecord('mfa-method', id).then((data) => data),
```
Monkeychip, line 6: what if no record can be found? could that situation happen?

Monkeychip (approved):
LGTM. Just a couple of comments, but nothing blocking.

ui/app/routes/vault/cluster/access/mfa/methods/method.js
zofskeez: After looking into the documentation for `hash` further I think it is fine to use it in a route model hook if the resolved data from **all** of the promises is required for the route. `hash` returns a promise that will be rejected if any of the supplied promises are rejected which will then trigger the error event in the route.

With that said, I think a method can exist on it's own which makes the result of the promise to fetch enforcements not required to load the route. Since you are swallowing the error it shouldn't be a problem here but just wanted to provide a bit of background on this pattern.

PR #15486, opened 2022-05-17
MFA Guided Setup Config View

zofskeez:
Adds mfa config view in guided setup for creating a method and login enforcement inline.

image

image

image


ui/app/controllers/vault/cluster/access/mfa/methods/create.js
```diff
@@ -3,25 +3,87 @@ import { inject as service } from '@ember/service';
...
+      } catch (error) {
+        this.handleError(error, 'Error saving enforcement');
+      }
+    } catch (error) {
+      this.handleError(error, 'Error saving method');
+    }
+  }
+  handleError(error, message) {
+    const errorMessage = error?.errors ? `${message}: ${error.errors.join(', ')}` : message;
+    this.flashMessages.danger(errorMessage);
```
Monkeychip, line 87: just curious if we're doing anything other than flash messages? e.g. displaying errors in the template that are static like an AlertBanner or something.

Monkeychip (approved):
Nice work!

ui/app/controllers/vault/cluster/access/mfa/methods/create.js
zofskeez: Not that I'm aware of, or at least there was nothing in the designs that indicated alert banners. We will be adding model validations for required fields that will be checked on save and any errors from that check will be displayed inline in the forms.

PR #15493, opened 2022-05-18
Forward autopilot state reqs, avoid self-dialing

ncabatoff:
Make sure that autopilot is disabled when we step down from active node state.  Forward autopilot state requests to the active node.  Avoid self-dialing due to stale advertisement from when we used to be active node.


ccapurso (approved):
Looks good!

raskchanky (approved):

PR #15494, opened 2022-05-18
Warn on empty Subject field for issuers

cipherboy:
When generating a root or signing an intermediate certificate, it is
possible to have Vault generate a certificate with an empty Subject.
These don't validate in most TLS implementations well, so add a warning.
Note that non-Common Name fields could be present to make a non-empty
subject, so simply requiring a CommonName isn't strictly the best.

For example:

    $ vault write pki/root/generate/exported common_name=""
      * This issuer certificate was generated without a Subject; this makes
      it likely that issuing leaf certs with this certificate will cause TLS
      validation libraries to reject this certificate.
    ....

### DIFF
--- a/builtin/logical/pki/path_root.go
+++ b/builtin/logical/pki/path_root.go
@@ -165,6 +165,19 @@ func (b *backend) pathCAGenerateRoot(ctx context.Context, req *logical.Request, 
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


stevendpclark (approved):

cipherboy:
Thanks!

PR #15495, opened 2022-05-18
Fix a generic PKI description for key_name and issuer_name fields

stevendpclark:
- The field could be used to be applied to keys/issuers being generated or to update the name on existing values.


kitography (approved):

stevendpclark:
Thanks @kitography

PR #15499, opened 2022-05-18
UI/mfa breadcrumbs and small fixes

Monkeychip:
This PR does the following:

* adds breadcrumbs to the MFA views.
* Makes the sidebar menu active when on the config page. 
image
* removes the box shadow on the config page to match designs.
_before there was a line underneath the second box._ 
image
* the config page on refresh was not checking for new mfa-methods. You had to change routes to trigger the check. Fixed this so on refresh the model query's the mfa-method.
* Adds an empty state when there are no enforcements.
image


ui/app/routes/vault/cluster/access/mfa/index.js
```diff
@@ -1,3 +1,15 @@
 import Route from '@ember/routing/route';
 
-export default class MfaConfigureRoute extends Route {}
+export default class MfaConfigureRoute extends Route {
```
Monkeychip, line 3: adding this here triggers the model to call the mfa-method query method on the mfa-method adapter. This is what checks if there are mfa-methods. Before you had to change the route to trigger this, now whenever the page is refreshed it checks.

zofskeez (approved):
LGTM thanks for this! Nothing blocking but had a thought about the methods check.

ui/app/routes/vault/cluster/access/mfa/index.js
Monkeychip: it seems to work just fine. amended. Thanks for the suggestion.

PR #15504, opened 2022-05-18
Fix minor typos and change terminology slightly in database secrets engines docs

robmonte:
Change terminology from "symbol characters" to "special characters" as they may not all be symbols. Also fix several instances of the typo "for for".


swenson (approved):
LGTM

PR #15507, opened 2022-05-19
Updated  Name reported by k8s auth

kaparora:
Since 1.9 k8s auth method supports setting  Name reported by auth method to Service Account name which is not reflected in this doc


taoism4504 (approved):
LGTM

tomhjp (approved):
LGTM, I've just merged in main to re-trigger the CI as I wasn't able to manually re-run it. I'll set up a queued auto-squash now. Thanks for the addition!

PR #15510, opened 2022-05-19
Vault 5917 allow patch operations to pki roles issuers

kitography:
PATCH functionality for Roles and Issuers


cipherboy (approved):
This largely looks good! I have passing tests for role based patching but I think we can add this in a separate PR.

Thank you @kitography!

builtin/logical/pki/path_fetch_issuers.go
```diff
@@ -319,6 +325,156 @@ func (b *backend) pathUpdateIssuer(ctx context.Context, req *logical.Request, da
...
+			// old name for this issuer, we're not actually updating the
+			// issuer name (or causing a conflict) -- so don't err out. Other
+			// errs should still be surfaced, however.
+			return logical.ErrorResponse(err.Error()), nil
+		}
+		if err == errIssuerNameInUse && issuer.Name != newName {
+			// When the new name is in use but isn't this name, throw an error.
+			return logical.ErrorResponse(err.Error()), nil
+		}
+		if newName != issuer.Name {
```
stevendpclark, line 377: Verify if we are writing a new name, if non-zero, that it matches `nameMatcher.MatchString(newName)`  so that we have valid names to be use within URLs. 

I'm okay with deferring this to another PR as we are missing it here and within the `pathUpdateIssuer` handler.

stevendpclark (approved):
👍

PR #14178, opened 2022-02-20
PKI: Add not_before_duration API parameter to Root/Intermediate CA generation

guysv:
Add not_before_duration API parameter to:
- Root CA generation
- Intermediate CA generation
- Intermediate CA signing

closes https://github.com/hashicorp/vault/issues/10631

PR #15511, opened 2022-05-19
Rebase #14178 / Add not_before_duration API parameter to Root/Intermediate CA generation

cipherboy:
Rebase of #14178 so we can land it in this release cycle :-)

This builds on the excellent work of @guysv with some minor updates:

1. This shouldn't occur on the Intermediate _CSR_ generation, as the validity period isn't conveyed at all on the CSR request and is solely at the discretion of the CA Issuer.
2. Decided to add this to the existing field helper rather than a new one.
3. Needed a test case :-)

Since we changed a lot of the structure of the PKI engine to support multiple issuers, this mostly just needed a rebase to catch the new `buildPath...` method locations. :-)

### DIFF
--- a/builtin/logical/pki/fields.go
+++ b/builtin/logical/pki/fields.go
@@ -399,3 +399,15 @@ to the key.`,
 	}
 	return fields
 }
+
+func addNotBeforeField(fields map[string]*framework.FieldSchema) map[string]*framework.FieldSchema {
+	fields["not_before_duration"] = &framework.FieldSchema{
+		Type:        framework.TypeDurationSecond,
+		Description: "Not-Before backdating in seconds",
+		DisplayAttrs: &framework.DisplayAttributes{
+			Name: "Not-before duration",
+		},
+	}
+
+	return fields
+}
--- a/builtin/logical/pki/fields.go
+++ b/builtin/logical/pki/fields.go
@@ -250,10 +250,20 @@ If you want more than one, specify alternative names in the alt_names
 map using OID 2.5.4.5. This has no impact on the final certificate's
 Serial Number field.`,
 	}
+
 	fields["not_after"] = &framework.FieldSchema{
 		Type: framework.TypeString,
 		Description: `Set the not after field of the certificate with specified date value.
 The value format should be given in UTC format YYYY-MM-ddTHH:MM:SSZ`,
+	}
+
+	fields["not_before_duration"] = &framework.FieldSchema{
+		Type:        framework.TypeDurationSecond,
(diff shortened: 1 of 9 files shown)


builtin/logical/pki/fields.go
```diff
@@ -250,12 +250,22 @@ If you want more than one, specify alternative names in the alt_names
...
 	fields["not_after"] = &framework.FieldSchema{
 		Type: framework.TypeString,
 		Description: `Set the not after field of the certificate with specified date value.
 The value format should be given in UTC format YYYY-MM-ddTHH:MM:SSZ`,
 	}
 
+	fields["not_before_duration"] = &framework.FieldSchema{
+		Type:        framework.TypeDurationSecond,
+		Default:     30,
+		Description: `The duration before now the cert needs to be created / signed.`,
```
kitography, line 263: I don't understand this description.

kitography (approved):
LGTM

builtin/logical/pki/fields.go
cipherboy: That's actually consistent with the existing description. I think perhaps we should update both... Maybe 

> The duration before now which the certificate needs to be backdated by.

is clearer?

builtin/logical/pki/fields.go
```diff
@@ -250,12 +250,22 @@ If you want more than one, specify alternative names in the alt_names
...
+
 	fields["not_after"] = &framework.FieldSchema{
 		Type: framework.TypeString,
 		Description: `Set the not after field of the certificate with specified date value.
 The value format should be given in UTC format YYYY-MM-ddTHH:MM:SSZ`,
 	}
 
+	fields["not_before_duration"] = &framework.FieldSchema{
+		Type:        framework.TypeDurationSecond,
+		Default:     30,
```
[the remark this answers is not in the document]
cipherboy: Also, its generally helpful if you issue an intermediate and immediately issue a leaf, the leaf will be backdated by 30s, but your issuer isn't, so if the target system is in that (-30, -1) window that the intermediate isn't valid for, you've gotta wait 30s before the leaf will be considered valid/trusted.

stevendpclark (approved):

builtin/logical/pki/fields.go
kitography: Yeah, that's much clearer, thank you

cipherboy:
Thanks all!

guysv:
Being lazy pays off 😅 thanks for the merge!!

PR #15512, opened 2022-05-19
UI: KeyMgmt Secret Engine

hashishaw:
This is the final piece of work to enable Key Management Secrets Engine management in the UI

**Enable KeyMgmt Secret Engine**
keymgmt-enable

**Create provider form with validation**
keymgmt-provider-form

**Creating a key**
keymgmt-key-form

**Key Rotation**
keymgmt-rotate-key

**Distribution works from both key and provider**
keymgmt-distribute-key
keymgmt-distribute-provider

Previous work that this builds off of:
https://github.com/hashicorp/vault/pull/13638
https://github.com/hashicorp/vault/pull/13797
https://github.com/hashicorp/vault/pull/13840

### DIFF
--- a/ui/app/components/keymgmt/key-edit.js
+++ b/ui/app/components/keymgmt/key-edit.js
@@ -54,7 +54,6 @@ export default class KeymgmtKeyEdit extends Component {
     const { model } = this.args;
     try {
       yield model.save();
-      yield model.reload();
       this.router.transitionTo(SHOW_ROUTE, model.name);
     } catch (error) {
       let errorMessage = error;
--- a/ui/app/components/keymgmt/key-edit.js
+++ b/ui/app/components/keymgmt/key-edit.js
@@ -54,16 +54,19 @@ export default class KeymgmtKeyEdit extends Component {
     const { model } = this.args;
     try {
       yield model.save();
+      yield model.reload();
       this.router.transitionTo(SHOW_ROUTE, model.name);
     } catch (error) {
       this.flashMessages.danger(error.errors.join('. '));
     }
   }
 
-  @action
-  async removeKey() {
+  @task
+  @waitFor
+  *removeKey() {
     try {
-      await this.keyAdapter.removeFromProvider(this.args.model);
+      yield this.keyAdapter.removeFromProvider(this.args.model);
+      yield this.args.model.reload();
       this.flashMessages.success('Key has been successfully removed from provider');
     } catch (error) {
       this.flashMessages.danger(error.errors?.join('. '));
@@ -84,11 +87,13 @@ export default class KeymgmtKeyEdit extends Component {
       });
   }
 
-  @action
-  rotateKey(id) {
-    const backend = this.args.model.get('backend');
+  @task
+  @waitFor
+  *rotateKey() {
+    const id = this.args.model.name;
+    const backend = this.args.model.backend;
     const adapter = this.keyAdapter;
-    adapter
+    yield adapter
       .rotateKey(backend, id)
       .then(() => {
         this.flashMessages.success(`Success: ${id} connection was rotated`);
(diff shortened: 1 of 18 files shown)


ui/app/components/keymgmt/key-edit.js
```diff
@@ -54,16 +54,29 @@ export default class KeymgmtKeyEdit extends Component {
     const { model } = this.args;
     try {
       yield model.save();
+      yield model.reload();
```
zofskeez, line 57: Curious does the response from the save request return no content but values are being updated on the backend?

ui/app/models/keymgmt/provider.js
```diff
@@ -109,7 +112,10 @@ export default class KeymgmtProviderModel extends Model {
   }
 
   async fetchKeys(page) {
-    if (this.canListKeys) {
+    if (this.canListKeys === false) {
```
zofskeez, line 115: Curious if there are other falsy values that can be returned from the lazyCapabilities check that should be ignored? If the value was `null` for example is that considered as having the permission?

ui/app/templates/components/input-search.hbs
```diff
@@ -1,6 +1,13 @@
 <div class="field is-grouped">
   <div class="control is-expanded">
+    {{#if @label}}
+      <label for={{@id}} class="is-label">{{@label}}</label>
+    {{/if}}
+    {{#if @subText}}
+      <p class="sub-text">{{@subText}}</p>
+    {{/if}}
```
zofskeez, line 8: :tada:

ui/tests/integration/components/keymgmt/distribute-test.js
```diff
@@ -78,33 +78,53 @@ module('Integration | Component | keymgmt/distribute', function (hooks) {
...
-    await select(SELECTORS.providerInput, 'provider-aws');
-    await settled();
     assert.dom(SELECTORS.operationsSection).doesNotHaveAttribute('disabled');
-    await select(SELECTORS.providerInput, 'provider-azure');
+    // Remove selection
+    await ssComponent.deleteButtons.objectAt(0).click();
+    // Select Azure
+    await settled();
+    await clickTrigger();
+    await settled();
```
zofskeez, line 118: I wonder if `settled` needs to be called after `clickTrigger`? I had a look and that helper calls `click` from _@ember/test-helpers_ which calls settled within that method.

ui/tests/integration/components/keymgmt/distribute-test.js
```diff
@@ -141,15 +162,19 @@ module('Integration | Component | keymgmt/distribute', function (hooks) {
...
     await render(
       hbs`<Keymgmt::Distribute @backend="keymgmt" @providers={{providers}} @key="example-1" @onClose={{fn (mut this.onClose)}} />`
     );
     assert.dom(SELECTORS.providerInput).exists('Provider input shown');
     assert.dom(SELECTORS.keySection).doesNotExist('Key input not shown');
-    await select(SELECTORS.providerInput, 'provider-azure');
+    await clickTrigger();
+    await settled();
+    await typeInSearch('azure');
+    await settled();
```
zofskeez, line 174: Same as the previous comment but with this one calls `fillIn`

zofskeez (approved):
LGTM thanks for getting this across the line!

ui/app/components/keymgmt/key-edit.js
hashishaw: Correct, PUT and POST both result in a 204 httpstatus.

I was being generous with model reloading (because there are a lot of actions that change the data) but it turns out this one is superfluous :)

ui/tests/integration/components/keymgmt/distribute-test.js
hashishaw: I got failures that such-and-such didn't exist when I removed `settled` but I can try again

hashishaw: Update: removed settled and it worked fine! 🎉

swayne275 (approved):
the `path_random.go` change was the right move, looks good!

edit: just approved for that bit, i don't pretend to know the other parts!

hashishaw:
Having build problems, closing in favor of #15523

PR #15513, opened 2022-05-19
Remove reference to stored license

arnav28:
- Stored license was deprecated in 1.8 and from 1.11 all licenses will be
auto loaded.

### DIFF
--- a/changelog/15513.txt
+++ b/changelog/15513.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Remove stored license references
+```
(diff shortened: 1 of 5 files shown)


Monkeychip (approved):
Maybe beyond the scope, but the property `autoloaded` is a boolean and used in more files throughout the app. The boolean no longer makes sense (I don't believe) so should this be changed? This is a question, because maybe I'm missing something but the definition of the autoloaded property is " Whether the license is autoloaded." maybe there are more than two options—a third I'm unaware of and if such then the property still makes sense.

arnav28:
> Maybe beyond the scope, but the property `autoloaded` is a boolean and used in more files throughout the app. The boolean no longer makes sense (I don't believe) so should this be changed? This is a question, because maybe I'm missing something but the definition of the autoloaded property is " Whether the license is autoloaded." maybe there are more than two options—a third I'm unaware of and if such then the property still makes sense.

Yup, I was planning to remove the autoloaded boolean but then noticed that we have something called persisted_autoload. Honestly not sure what the difference is!!

PR #15519, opened 2022-05-19
VAULT-4306 Ensure /raft/bootstrap/challenge call ignores erroneous namespaces set

VioletHynes:
This ensures servers started on an environment where $VAULT_NAMESPACE is set do not use such a variable when making this kind of call, causing an error.

Notably, this file seems to be the only place where stripping the namespace makes sense (other API clients come from a CLI/api client call, where the namespace makes sense), which is why I did not make it a configuration item for ApiClient or similar. If it becomes a more general need, it makes sense to make the configuration change then, but it didn't make sense to me to add a new configuration item or method for what appears to be a snowflake.

### DIFF
--- a/changelog/15519.txt
+++ b/changelog/15519.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+raft: /raft/bootstrap/challenge call now properly ignores $VAULT_NAMESPACE environment variable when set on server attempting to join a raft cluster
+```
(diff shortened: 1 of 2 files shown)


swayne275 (approved):

changelog/15519.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+raft: /raft/bootstrap/challenge call now properly ignores $VAULT_NAMESPACE environment variable when set on server attempting to join a raft cluster
```
ncabatoff, line 2:
```suggestion
raft: joining a node to a cluster now ignores any VAULT_NAMESPACE environment variable set on the server process
```

ncabatoff (approved):

PR #15522, opened 2022-05-19
UI/mfa small bugs

Monkeychip:
This PR address the issue where, because of lazyPaginatedQuery, the list views for methods and enforcements were not being updated after deleting a method and/or enforcement. It appears the is issue is lazyPaginatedQuery does not use the store for its internal record and so was caching and therefore not calling the query again after the transition. We removed this so no pagination now happens, but neither does the bug.


zofskeez (approved):
LGTM thanks!

PR #15524, opened 2022-05-19
Return the signed ca in the ca_chain response field within sign-intermediate api call.

stevendpclark:
- When signing a CA certificate we should include it along with the issuers CA chain in the response.

### DIFF
--- a/builtin/logical/pki/path_root.go
+++ b/builtin/logical/pki/path_root.go
@@ -383,32 +383,30 @@ func (b *backend) pathIssuerSignIntermediate(ctx context.Context, req *logical.R
 		resp.AddWarning("This mount hasn't configured any authority access information fields; this may make it harder for systems to find missing certificates in the chain or to validate revocation status of certificates. Consider updating /config/urls with this information.")
 	}
 
+	caChain := append([]string{cb.Certificate}, cb.CAChain...)
+
 	switch format {
 	case "pem":
 		resp.Data["certificate"] = cb.Certificate
 		resp.Data["issuing_ca"] = signingCB.Certificate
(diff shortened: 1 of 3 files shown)


builtin/logical/pki/path_root.go
```diff
@@ -383,32 +383,30 @@ func (b *backend) pathIssuerSignIntermediate(ctx context.Context, req *logical.R
...
 	case "pem":
 		resp.Data["certificate"] = cb.Certificate
 		resp.Data["issuing_ca"] = signingCB.Certificate
-		if cb.CAChain != nil && len(cb.CAChain) > 0 {
-			resp.Data["ca_chain"] = cb.CAChain
-		}
+		resp.Data["ca_chain"] = caChain
 
 	case "pem_bundle":
 		resp.Data["certificate"] = cb.ToPEMBundle()
```
cipherboy, line 395: This is what makes me think this is a good idea: this gives you leaf + chain of the cert in the `certificate` field. :-) Avoids the need for concat.

changelog/15524.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:change
+secrets/pki: the signed CA certificate from the /root/sign-intermediate api will now appear within the ca_chain
```
cipherboy, line 2: Want to elide the `/root` portion, since there's now two of these APIs (`/issuer/:issuer_ref/sign-intermediate` as well)?

cipherboy (approved):

changelog/15524.txt
stevendpclark: tweaked.

PR #15525, opened 2022-05-19
Fix handling of username_as_alias during LDAP authentication

remilapeyre:
There is a bug that was introduced in the LDAP authentication method by https://github.com/hashicorp/vault/pull/11000.
It was thought to be backward compatible but has broken a number of users. Later
a new parameter `username_as_alias` was introduced in https://github.com/hashicorp/vault/pull/14324
to make it possible for operators to restore the previous behavior.
The way it is currently working is not completely backward compatible thought
because when username_as_alias is set, a call to GetUserAliasAttributeValue() will
first be made, then this value is completely discarded in pathLogin() and replaced
by the username as expected.

This is an issue because it makes useless calls to the LDAP server and will break
backward compatibility if one of the constraints in GetUserAliasAttributeValue()
is not respected, even though the resulting value will be discarded anyway.

In order to maintain backward compatibility here we have to only call
GetUserAliasAttributeValue() if necessary.

Since this change of behavior was introduced in 1.9, this fix will need to be
backported to the 1.9, 1.10 and 1.11 branches.

### DIFF
--- a/builtin/credential/ldap/backend.go
+++ b/builtin/credential/ldap/backend.go
@@ -198,8 +198,8 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username stri
 	if usernameAsAlias {
 		return username, policies, ldapResponse, allGroups, nil
 	}
-	
-	entityAliasAttribute, err = ldapClient.GetUserAliasAttributeValue(cfg.ConfigEntry, c, username)
+
+	entityAliasAttribute, err := ldapClient.GetUserAliasAttributeValue(cfg.ConfigEntry, c, username)
 	if err != nil {
 		return "", nil, logical.ErrorResponse(err.Error()), nil, nil
 	}
--- a/builtin/credential/ldap/backend.go
+++ b/builtin/credential/ldap/backend.go
@@ -195,15 +195,16 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username stri
 	// Policies from each group may overlap
 	policies = strutil.RemoveDuplicates(policies, true)
 
-	entityAliasAttribute := username
-	if !usernameAsAlias {
-		entityAliasAttribute, err = ldapClient.GetUserAliasAttributeValue(cfg.ConfigEntry, c, username)
-		if err != nil {
-			return "", nil, logical.ErrorResponse(err.Error()), nil, nil
-		}
-		if entityAliasAttribute == "" {
-			return "", nil, logical.ErrorResponse("missing entity alias attribute value"), nil, nil
-		}
+	if usernameAsAlias {
+		return username, policies, ldapResponse, allGroups, nil
+	}
+	
+	entityAliasAttribute, err = ldapClient.GetUserAliasAttributeValue(cfg.ConfigEntry, c, username)
+	if err != nil {
+		return "", nil, logical.ErrorResponse(err.Error()), nil, nil
+	}
+	if entityAliasAttribute == "" {
+		return "", nil, logical.ErrorResponse("missing entity alias attribute value"), nil, nil
 	}
 
 	return entityAliasAttribute, policies, ldapResponse, allGroups, nil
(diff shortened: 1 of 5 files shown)


calvn:
Thanks for the PR to address this! Can you include a test case to cover this scenario?

> will break backward compatibility if one of the constraints in GetUserAliasAttributeValue()
is not respected, even though the resulting value will be discarded anyway.

remilapeyre:
Hi @calvn, I added two tests, one for `username_as_alias = true` and the other for `username_as_alias = false`. I also ttok this opportunity to fix the error message at https://github.com/hashicorp/vault/blob/v1.9.6/sdk/helper/ldaputil/client.go#L247.

Can you please make a review?

builtin/credential/ldap/backend.go
```diff
@@ -195,12 +195,15 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username stri
...
+	entityAliasAttribute := username
+	if !usernameAsAlias {
+		entityAliasAttribute, err = ldapClient.GetUserAliasAttributeValue(cfg.ConfigEntry, c, username)
+		if err != nil {
+			return "", nil, logical.ErrorResponse(err.Error()), nil, nil
+		}
+		if entityAliasAttribute == "" {
+			return "", nil, logical.ErrorResponse("missing entity alias attribute value"), nil, nil
+		}
 	}
```
calvn, line 207: Nit: Can we flip the check here and do an early return so that we don't end up wrapping the calls within the conditional?

```suggestion
	if usernameAsAlias {
		return username, policies, ldapResponse, allGroups, nil
	}
	
	entityAliasAttribute, err = ldapClient.GetUserAliasAttributeValue(cfg.ConfigEntry, c, username)
	if err != nil {
		return "", nil, logical.ErrorResponse(err.Error()), nil, nil
	}
	if entityAliasAttribute == "" {
		return "", nil, logical.ErrorResponse("missing entity alias attribute value"), nil, nil
	}
```

remilapeyre:
Hi @calvn do you think you will be able to do the review shortly. This is blocking one of our Vault enterprise client.

jasonodonnell:
Hi @remilapeyre, CI is failing due to a formatting error (please run `make fmt`), as well as an undefined error: https://app.circleci.com/pipelines/github/hashicorp/vault/34706/workflows/89abcf43-92ae-49d3-9b9e-6cbb56aa0374/jobs/450756.

Can you take a look at these?

builtin/credential/ldap/backend.go
```diff
@@ -195,7 +195,11 @@ func (b *backend) Login(ctx context.Context, req *logical.Request, username stri
 	// Policies from each group may overlap
 	policies = strutil.RemoveDuplicates(policies, true)
 
-	entityAliasAttribute, err := ldapClient.GetUserAliasAttributeValue(cfg.ConfigEntry, c, username)
+	if usernameAsAlias {
+		return username, policies, ldapResponse, allGroups, nil
+	}
+	
+	entityAliasAttribute, err = ldapClient.GetUserAliasAttributeValue(cfg.ConfigEntry, c, username)
```
jasonodonnell, line 202: Undefined error here because `:=` is missing.

remilapeyre: Yes, I did not check the previous commit enough sorry about that. It should be good now.

changelog/15525.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+auth/ldap: A bug during login has been fixed when `username_as_alias` is set.
```
calvn, line 2: Nit: Can we update this entry to be a little bit more descriptive on the nature of the bug? Maybe also include the keyword `userattr` somewhere in here to to make it a bit more discoverable/searchable.

calvn (approved):
Changes look good, though I'd like to hold off on the merge until @jasonodonnell is all good with some validation he is performing on his end.

jasonodonnell (approved):
LGTM!

remilapeyre:
Hi @calvn @jasonodonnell, thanks for reviewing the PR. Do you know when 1.9.7 will be released? We can't build it ourselves because we are using the enterprise version.

Issue #13364, opened 2021-12-08
PKI: /sign-verbatim - key_usage and ext_key_usage request params are used as defaults instead of restrictions


djivko:
**Describe the bug**
/sign-verbatim API documentation states that "key_usage" and "ext_key_usage" request params should be used as restrictions for the allowed values of those certificate extensions. From docs:
`key_usage (list: ["DigitalSignature", "KeyAgreement", "KeyEncipherment"]) – Specifies the allowed key usage constraint on issued certificates...`
`ext_key_usage (list: []) – Specifies the allowed extended key usage constraint on issued certificates...`

However if the CSR includes those extentitions the values of the request params are discarded and the certificate is issued with "Key Usage" and "Extended Key Usage" extensions as they are defined in the CSR.

**To Reproduce**
1. Generate a CSR with "Extended Key Usage"  = "CodeSigning" in it.
2. Sign the certificate using /sign-verbatim endpoint and add "ext_key_usage: ["ServerAuth"]" param in the request. 
3. Parse the issued certificate and see that it includes  "Extended Key Usage"  extension has a value equal to "CodeSigning" (as it was defined in the CSR)

**Expected behavior**
1. If the CSR does not contain "Key Usage" and/or "Extended Key Usage" extensions the ones specified in the request are used for issuing the certificate (that's the current behaviour as well)
2. If the CSR contains "Key Usage" and/or "Extended Key Usage" extensions only the values that match the request params will be kept and the others will be discarded.
3.  If the CSR contains "Key Usage" and/or "Extended Key Usage" extensions but none of them matches the request params the signing of the certificate will be rejected with a proper error.  

**Additional context**
The issue seems to be that:
`x509.CreateCertificate`
method discards any "certTemplate" member values if they are already part of "certTemplate.ExtraExtensions"

hsimon-hashicorp:
Hi @djivko - just wanted to let you know that engineering and product are discussing more about how we want to handle x.509 policies, and we'll keep this feedback in mind as we consider the various approaches. Thanks for submitting this!

djivko:
Hey @hsimon-hashicorp Thanks for the follow up. It's much appreciated

PR #15535, opened 2022-05-20
Clarify KU/EKU parameters on sign-verbatim

cipherboy:

---

This resolves #[13364](https://github.com/hashicorp/vault/issues/13364) with a documentation update. In particular, modifying the behavior of `sign-verbatim` at this point in time is not possible, so we attempt to clarify the documentation.

Resolves: #13364

### DIFF
--- a/website/content/api-docs/secret/pki.mdx
+++ b/website/content/api-docs/secret/pki.mdx
@@ -696,18 +696,29 @@ have access.**
 - `csr` `(string: <required>)` - Specifies the PEM-encoded CSR.
 
 - `key_usage` `(list: ["DigitalSignature", "KeyAgreement", "KeyEncipherment"])` -
-  Specifies the allowed key usage constraint on issued certificates. Valid
+  Specifies the default key usage constraint on the issued certificate. Valid
   values can be found at https://golang.org/pkg/crypto/x509/#KeyUsage - simply
   drop the `KeyUsage` part of the value. Values are not case-sensitive. To
   specify no key usage constraints, set this to an empty list.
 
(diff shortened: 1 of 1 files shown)


kitography (approved):
That makes sense to me.

website/content/api-docs/secret/pki.mdx
```diff
@@ -696,19 +696,30 @@ have access.**
 - `csr` `(string: <required>)` - Specifies the PEM-encoded CSR.
 
 - `key_usage` `(list: ["DigitalSignature", "KeyAgreement", "KeyEncipherment"])` -
-  Specifies the allowed key usage constraint on issued certificates. Valid
+  Specifies the default key usage constraint on the issued certificate. Valid
   values can be found at https://golang.org/pkg/crypto/x509/#KeyUsage - simply
   drop the `KeyUsage` part of the value. Values are not case-sensitive. To
   specify no key usage constraints, set this to an empty list.
```
stevendpclark, line 702: Should we also remove this sentence in both field definitions? `To specify no key usage constraints, set this to an empty list.`

cipherboy: Being pedantic, but the [KeyUsage](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3) field is a constraint ("restriction" in the RFC) on the certificate (and we call it the KU constraint above). 

How about "To specify no _default_ key usage constraints, ..."?

stevendpclark: Yes that sounds good

stevendpclark (approved):

PR #15536, opened 2022-05-20
api/monitor: Adding log format to monitor command and debug

davidadeleon:
Adding log_format flag to monitor endpoint as well as debug command. This will allow the option for vault logs being emitted via monitor to come through as JSON format. The default behavior is to emit in the standard log format.

Usage:
`vault monitor -log-level=DEBUG -log-format=JSON`

`curl -H "X-Vault-Request: true" -H "X-Vault-Token: $(vault print token)" http://127.0.0.1:8200/v1/sys/monitor?log_format=json&log_level=trace`

`vault debug -target=log -log-format=json`

### DIFF
--- a/api/sys_monitor.go
+++ b/api/sys_monitor.go
@@ -5,17 +5,25 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+
+	"github.com/hashicorp/vault/sdk/helper/logging"
 )
 
 // Monitor returns a channel that outputs strings containing the log messages
 // coming from the server.
-func (c *Sys) Monitor(ctx context.Context, logLevel string) (chan string, error) {
+func (c *Sys) Monitor(ctx context.Context, logLevel string, logFormat string) (chan string, error) {
 	r := c.c.NewRequest(http.MethodGet, "/v1/sys/monitor")
 
 	if logLevel == "" {
 		r.Params.Add("log_level", "info")
 	} else {
 		r.Params.Add("log_level", logLevel)
+	}
+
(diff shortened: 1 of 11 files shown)


jasonodonnell (approved):
LGTM!

vault/logical_system_paths.go
```diff
@@ -1388,6 +1388,11 @@ func (b *SystemBackend) monitorPath() *framework.Path {
 				Description: "Log level to view system logs at. Currently supported values are \"trace\", \"debug\", \"info\", \"warn\", \"error\".",
 				Query:       true,
 			},
+			"log_format": {
+				Type:        framework.TypeString,
```
ncabatoff, line 1392: I think you should add a default of `standard` in case an older vault cli is talking to a version that supports this.

api/sys_monitor.go
```diff
@@ -18,6 +20,12 @@ func (c *Sys) Monitor(ctx context.Context, logLevel string) (chan string, error)
 		r.Params.Add("log_level", logLevel)
 	}
 
+	if logFormat == "" || logFormat == logging.UnspecifiedFormat.String() {
```
ncabatoff, line 23: Why check UnspecifiedFormat?

davidadeleon: Some of our tests rely upon parsing of the logformat environment variable which if unset is set to "unspecified".

ccapurso (approved):
Just one small comment on one of the doc changes but otherwise looks good!

PR #15541, opened 2022-05-20
PKI: Do not error out on unknown issuers/keys on delete api calls.

stevendpclark:
- No longer error out when we fail to lookup the passed in issuer_ref or key_ref values on delete apis.
 - Add more key related unit tests


cipherboy (approved):
Makes sense, thank you!

PR #15543, opened 2022-05-20
Allow issuer/:issuer_ref/sign-verbatim/:role, add error on missing role

cipherboy:
In #15277, we missed that the new `/issuer/issuer_ref/sign-verbatim` path lacked a role-allowed variant (like `/sign-verbatim/:role` does). Add the missing parameter.

Additionally, I noticed that `sign-verbatim` (since early versions), wouldn't warn or err on a missing role when one was requested. Since we don't currently err, we can't now err, so add a warning instead.

### DIFF
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -267,7 +267,7 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
 			if err != nil {
 				return nil, err
 			}
-			if role == nil && roleMode == roleRequired {
+			if role == nil && (roleMode == roleRequired || role == roleOptional) {
 				return logical.ErrorResponse(fmt.Sprintf("unknown role: %s", roleName)), nil
 			}
 			labels = []metrics.Label{{"role", roleName}}
--- a/builtin/logical/pki/backend.go
+++ b/builtin/logical/pki/backend.go
@@ -267,7 +267,7 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
 			if err != nil {
 				return nil, err
 			}
-			if role == nil && (roleMode == roleRequired || role == roleOptional) {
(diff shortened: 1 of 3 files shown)


stevendpclark (approved):

changelog/15543.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
stevendpclark, line 1: I think this should be a `change` type now?

```suggestion
```release-note:change
```

builtin/logical/pki/path_issue_sign.go
```diff
@@ -219,6 +219,14 @@ func (b *backend) pathSignVerbatim(ctx context.Context, req *logical.Request, da
 		entry.Issuer = role.Issuer
 	}
 
+	// Either we have a non-sign-verbatim path, or we have a role named
+	// sign-verbatim being used with the sign-verbatim endpoint.
+	pathShouldHaveRole := !strings.HasSuffix(req.Path, "sign-verbatim")
```
stevendpclark, line 224: Instead of all this should we not fix this within backend.metricsWrap ? If it was provided a roleName and it failed to lookup the role to fail?

cipherboy: Ahhh, that's probably better.

builtin/logical/pki/backend.go
```diff
@@ -267,7 +267,7 @@ func (b *backend) metricsWrap(callType string, roleMode int, ofunc roleOperation
 			if err != nil {
 				return nil, err
 			}
-			if role == nil && roleMode == roleRequired {
+			if role == nil && (roleMode == roleRequired || role == roleOptional) {
```
stevendpclark, line 270: shouldn't this read more like this? As now I *believe* if we don't specify a role it will fail?
```suggestion
			if role == nil && (roleMode == roleRequired || len(roleName) > 0 {
```

cipherboy:
Thanks all! Merging...

PR #15545, opened 2022-05-20
Add role patching test case

cipherboy:
This updates almost all fields individually in a role, with the exceptions of (key_type, key_bits, and signature_bits) due to inter-dependencies between them. 

Also, fixes a bug spotted by @stevendpclark during review of #15510, where both Update and Patch issuer didn't validate the new name.


stevendpclark (approved):
Thanks! I assume we will eventually update the api-docs about the supported patch ops we now have?

cipherboy:
Yeah, I could probably do that here.

stevendpclark (approved):

cipherboy:
Thanks!

PR #15548, opened 2022-05-20
remove token type

Monkeychip:
According to the backend: >Vault login with a token doesn’t actually talk to vault, it’s just shorthand for “copy this token into my token helper, by default into ~/.vault-token”.  I would omit it (token) from the enforcement options, auth/token with mfa is more of a step-up mfa (ent mfa) thing.

This PR filters it out as an option in the template.


hashishaw (approved):
Nice! Test coverage for this would be excellent

PR #15200, opened 2022-04-27
MFA Config

zofskeez:
Side branch for merging phase 2 mfa config feature work.

PR #15549, opened 2022-05-20
UI/mfa small cleanup

hashishaw:
Address some of the small items from my review of https://github.com/hashicorp/vault/pull/15200

### DIFF
--- a/ui/app/templates/components/mfa-login-enforcement-header.hbs
+++ b/ui/app/templates/components/mfa-login-enforcement-header.hbs
@@ -1,5 +1,5 @@
 {{#if @isInline}}
-  <h3 class="title is-5" data-test-mfa-title>Enforcement</h3>
+  <h3 class="title is-5" data-test-mleh-title>Enforcement</h3>
 {{else}}
   <PageHeader as |p|>
     <p.top>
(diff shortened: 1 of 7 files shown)


ui/app/templates/components/mfa-login-enforcement-header.hbs
```diff
@@ -1,5 +1,5 @@
 {{#if @isInline}}
-  <h3 class="title is-5" data-test-mleh-title>Enforcement</h3>
+  <h3 class="title is-5" data-test-mfa-title>Enforcement</h3>
```
zofskeez, line 2: I went with the abbreviated component name for acceptance tests when both the header and form components are rendered in the same view to avoid possible collisions. I can tweak if necessary when I get to writing those.

zofskeez (approved):
LGTM. Thanks for reviewing and making the changes!

ui/app/templates/components/mfa-login-enforcement-header.hbs
hashishaw: ohhh i thought it was a version of "bleh" 😂 I can change these back, I agree that something specific like this is good for avoiding collisions

zofskeez: Haha I totally see that now! We created a new slang. _mleh_ is a mix of _meh_ and _blah_ and represents mundane indifference :laughing:

PR #15550, opened 2022-05-20
Add deprecated tags to token_type and policy

robmonte:
Adding deprecated tags to the fields `token_type` and `policy` as they were long-ago deprecated in Consul and no longer supported in new versions. Documentation and explanation of this change will be added in the neat future separately.


swenson (approved):
LGTM

PR #15559, opened 2022-05-20
Convert not_before_duration to seconds before returning it

Gabrielopesantos:
# Description
Found a bug in an improvement that was merged to main the other day (#15250).
The SSH Role field `not_before_duration` field is returned in nanoseconds instead of seconds (as expected).
The following requests fixes this issue by converting the value of the fields to seconds before returning it.

## Screenshots of the bug (`not_before_duration` meant to be 30 seconds)
ui
cli

PS: The screenshots are a bit bigger than expected.

## Others
Not sure if it's worth creating an issue for this.

@cipherboy 
Sorry for pinging, just thought it would be faster since you have context of what was done.


cipherboy:
@Gabrielopesantos Thanks for catching this! I'll take a look on Monday, but I've tagged us so we'll hopefully remember.

cipherboy (approved):
Typically we don't need a changelog entry for this that haven't hit a RC or anything yet, so I'd usually ask you to remove it. But its such a small change that it seems a waste to re-run CI...

PR #15560, opened 2022-05-20
UI/Fix validation issues, regression from Ember upgrade

hellobontempo:


ui/app/utils/validators.js
```diff
@@ -3,17 +3,20 @@ import { isPresent } from '@ember/utils';
...
+    const overMax = max && valueLength > max;
+    return underMin || overMax ? false : true;
   }
-  return isValid;
+  return nullable;
 };
 
 export const number = (value, { nullable = false, asString } = {}) => {
-  if (!value) return nullable;
+  // since 0 is falsy, !value is true even though 0 is valid here
```
hellobontempo, line 18: This was source part I of the bug. All validations are treated as a string, even tough default integer values only become a string once the user has interacted with the form. So in KV for example, the validation decorator attempts to validate the default input of `0` and since`!0 = true` the check is invalid.

Alternative approach - since they are form values, do we want all of our default inputs to be strings? (This would also address the issue above)

ui/app/models/secret-v2.js
```diff
@@ -8,7 +8,7 @@ import { withModelValidations } from 'vault/decorators/model-validations';
 
 const validations = {
   maxVersions: [
-    { type: 'number', options: { asString: true }, message: 'Maximum versions must be a number.' },
```
hellobontempo, line 11: This seems a little confusing because originally the `defaultValue` is an integer, and then after interacting with the form it becomes a string.

ui/lib/core/addon/components/form-field.hbs
```diff
@@ -261,7 +261,7 @@
           readonly={{this.isReadOnly}}
           autocomplete="off"
           spellcheck="false"
-          value={{or (get @model this.valuePath) @attr.options.defaultValue}}
+          value={{or @attr.options.defaultValue (get @model this.valuePath)}}
```
hashishaw, line 264: This will always set the value to the passed default if exists even if the model has a value at `valuePath`, so I think we should put this back

ui/tests/unit/utils/validators-test.js
```diff
@@ -6,10 +6,18 @@ module('Unit | Util | validators', function (hooks) {
...
-    assert.true(isValid);
+    let isValid;
+    const check = (value) => (isValid = validators.presence(value));
+    check(null);
+    assert.false(isValid, 'Invalid when value is null');
+    check(true);
+    assert.true(isValid, 'Valid when value is true');
+    check('');
+    assert.false(isValid, 'Invalid when value is empty string');
+    check(0);
```
hashishaw, line 17: 👏

ui/lib/core/addon/components/form-field.hbs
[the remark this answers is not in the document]
hellobontempo: But - I'm not sure if this switch could cause other issues

hashishaw (changes requested):
Just one thing! Good job tracking this down

ui/app/components/mount-backend-form.js
```diff
@@ -88,10 +87,22 @@ export default Component.extend({
     }
   },
 
+  isModelValid(model) {
```
hashishaw, line 90: nit that doesn't need to be changed, but the naming `isSomething` doesn't make me think it's a function, it seems like an attribute. Great that it returns a boolean though! Personally i would name it something like `checkModelValidity`

ui/app/templates/components/mount-backend-form.hbs
```diff
@@ -68,7 +68,7 @@
           type="submit"
           data-test-mount-submit="true"
           class="button is-primary {{if this.mountBackend.isRunning 'loading'}}"
-          disabled={{or this.mountBackend.isRunning this.isFormInvalid}}
+          disabled={{or this.mountBackend.isRunning}}
```
hashishaw, line 71: we don't need the `or` anymore

ui/app/components/mount-backend-form.js
hellobontempo: this is helpful! i always get caught up on naming, and want it to be clear that it was a boolean situation (so the conditional below made sense) will change as I take naming quite seriously (...too seriously)

ui/app/components/mount-backend-form.js
```diff
@@ -120,7 +134,6 @@ export default Component.extend({
       } catch (err) {
         if (err.httpStatus === 403) {
           this.mountIssue = true;
-          this.set('isFormInvalid', this.mountIssue);
```
hellobontempo, line 123: Originally this was to disable the submit button - which we no longer want to do

ui/app/decorators/model-validations.js
```diff
@@ -110,9 +112,19 @@ export function withModelValidations(validations) {
               }
             }
           }
+          errors.push(state[key].errors);
```
hellobontempo, line 115: This is new since the PR was approved

ui/lib/core/addon/templates/components/alert-inline.hbs
```diff
@@ -1,8 +1,18 @@
-<Icon @name={{this.alertType.glyph}} class={{this.alertType.glyphClass}} />
-<p class={{this.textClass}} data-test-inline-error-message>
-  {{#if (has-block)}}
-    {{yield}}
+<div
+  {{did-update this.refresh @message}}
```
zofskeez, line 2: I like how you got around this requirement. This is basically to force some kind of more dramatic view change if the message itself changes?

zofskeez (commented):
Nice work! It might be nice to add a model-validations test for the new `invalidFormMessage` property.

ui/app/decorators/model-validations.js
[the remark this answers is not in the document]
hellobontempo: I like that! I wanted to leverage what was pre-existing...and ran out of brain energy in how to do so. Good suggestion!

ui/app/decorators/model-validations.js
```diff
@@ -110,9 +112,19 @@ export function withModelValidations(validations) {
...
+        invalidFormMessage = this.generateErrorCountMessage(errors);
+
+        return { isValid, state, invalidFormMessage };
+      }
+
+      generateErrorCountMessage(errors) {
+        // returns form-specific message: 'There is an error/are N errors with this form.'
+        let errorCount = errors.map((e) => e.length).reduce((a, b) => a + b, 0);
+        let isPlural = errorCount > 1 ? `are ${errorCount} errors` : false;
+        return `There ${isPlural ? isPlural : 'is an error'} with this form.`;
```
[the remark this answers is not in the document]
hellobontempo: Good call!

ui/app/models/secret-v2.js
hellobontempo: Yeah - this was part of the original bug actually. `maxVersions` is returned as an integer from the API and defined on the model as a number (in the model's attrs) but when interacting with the form (when validations happen) it's a string. 

max-versions

ui/lib/core/addon/components/alert-inline.js
```diff
@@ -1,7 +1,10 @@
-import Component from '@ember/component';
-import { computed } from '@ember/object';
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { later } from '@ember/runloop';
+import { tracked } from '@glimmer/tracking';
 import { messageTypes } from 'core/helpers/message-types';
 import layout from '../templates/components/alert-inline';
+import { setComponentTemplate } from '@ember/component';
```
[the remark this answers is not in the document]
hellobontempo: Aha! Thank you

ui/lib/core/addon/templates/components/alert-inline.hbs
hellobontempo: Yeah! Before updating the count, the loading spinner displays briefly
fake-load

ui/app/models/secret-v2.js
zofskeez: Thanks. So I think if we are not concerned with validating type in the number validator than it is fine to take out the `asString` option.

Issue #10879, opened 2021-02-10
Feature request for SSH secrets engine to create keypairs (CA mode)


nvx:
**Is your feature request related to a problem? Please describe.**
When using Vault with Nomad, consul-template, or Vault Agent a common pattern to create TLS certificates with the PKI engine has Vault generate both the private key and the signed certificate. By virtue of having a separate endpoint to sign a CSR for an existing key vs having Vault generate a new private key, this can ensure via policy that secret keys are rotated frequently.

Unfortunately this pattern is not possible with the SSH secrets engine in CA mode for user keys, an already-generated public/private keypair must be generated and the private key is sent to Vault for creation of the certificate. This makes it difficult to use in eg Nomad where creating the private key ahead of time is difficult (unless a long lived static keypair is stored in eg Vault KV store), and makes it impossible to enforce short key lifetimes.

**Describe the solution you'd like**
Duplicating the existing [POST /ssh/sign/:name](https://www.vaultproject.io/api/secret/ssh#sign-ssh-key) endpoint as /ssh/issue/:name (this follows naming of the equivalent endpoints in the PKI engine) that does not take a public_key, and instead generates a new public/private keypair internally in Vault which is returned in the response.

**Describe alternatives you've considered**
Scripts external to Vault could automate the generation of the certificate (although this becomes much harder within Nomad), but this can not guarantee by policy that secret keys are cycled.

**Additional context**
https://www.vaultproject.io/api-docs/secret/pki#generate-certificate
https://www.vaultproject.io/api-docs/secret/pki#sign-certificate

cipherboy:
@Gabrielopesantos What about this one? You'll need to wire up a new response handler [in the backend](https://github.com/hashicorp/vault/blob/main/builtin/logical/ssh/backend.go#L63), validate the [requested key type against the role](https://github.com/hashicorp/vault/blob/7c72d9d3d6f127ff368d6f47634684c6fb60cb22/builtin/logical/ssh/path_sign.go#L118), and then [generate a new key pair](https://github.com/hashicorp/vault/blob/7c72d9d3d6f127ff368d6f47634684c6fb60cb22/builtin/logical/ssh/path_config_ca.go#L288) and then largely refactor/duplicate the logic of [signing](https://github.com/hashicorp/vault/blob/7c72d9d3d6f127ff368d6f47634684c6fb60cb22/builtin/logical/ssh/path_sign.go#L107-L206). 

[...]

PR #15561, opened 2022-05-21
SSH secrets engine - Enabled creation of key pairs (CA Mode)

Gabrielopesantos:
# Description
Issue: #10879
Created a new endpoint (/ssh/issue/:name) that does not take a public_key (as /ssh/sign/:name does), and instead generates a new public/private key pair internally which is returned in the response.


builtin/logical/ssh/path_issue.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if role.KeyType != "ca" {
+		return logical.ErrorResponse("role key type \"%s\" not allowed to issue key pairs", role.KeyType), nil
+	}
+
+	keyType := data.Get("key_type").(string)
+	keyBits := data.Get("key_bits").(int)
+
+	keyTypeLengths, keyPresent := role.AllowedUserKeyTypesLengths[keyType]
+	if keyPresent {
```
cipherboy, line 90: I think you'd err here too; if the keyType isn't in the allowed list, you shouldn't allow the request.

builtin/logical/ssh/path_issue.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	if role.KeyType != "ca" {
+		return logical.ErrorResponse("role key type \"%s\" not allowed to issue key pairs", role.KeyType), nil
+	}
+
+	keyType := data.Get("key_type").(string)
+	keyBits := data.Get("key_bits").(int)
+
+	keyTypeLengths, keyPresent := role.AllowedUserKeyTypesLengths[keyType]
```
cipherboy, line 89: I think you'll end up wanting to hoist this to a package level internal constant:

https://github.com/hashicorp/vault/blob/7c72d9d3d6f127ff368d6f47634684c6fb60cb22/builtin/logical/ssh/path_sign.go#L476-L481 

There's a series of aliases we allow (for better usability) and you've gotta check them all (for compatibility with the existing validatiog logic). 

And just to clarify this, they are parameters/field in the allowed map, not in the requested params. There you can keep it simple and align with the PKI endpoint (rsa/ec/ed25519).

cipherboy: Additionally, this check (lines 89 through 102) needs to be conditional on whether or not the map is even present:

https://github.com/hashicorp/vault/blob/7c72d9d3d6f127ff368d6f47634684c6fb60cb22/builtin/logical/ssh/path_sign.go#L450

If the map isn't present, you should allow all key types/bits. :-)

Well, I'd deny RSA with bits < 2048, but yeah.... Oh and Go doesn't really have DSA support any more, so please don't implement it either :D It is supported on signing since users can generate DSA keys, but its generally considered weak now.

cipherboy (commented):
Looking like a great start @Gabrielopesantos! 

A few nits here and there around how validation of type/bits parameters occurs, that's tricky but expected :-) In retrospect, I really made that more complicated than I intended to. 

You might be able to refactor the existing logic into taking a keyType/keyBits parameter and hoisting the type/bits extraction to a separate method.

And regarding your comment:

> Also, for now everything is in a single function but I could also refactor the existing `pathSignCertificate` to support this new case.

I think I'd maybe take the common logic (124 and down in the new implementation?) and refactor it to a shared call. So three functions (`pathSignCertificate`, `pathIssueCertificate`, and perhaps `pathSignIssueCertificateHelper` or something of the sort). 

Just a thought. :-)

builtin/logical/ssh/path_issue.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, fmt.Errorf("failed to generate or parse the keys")
+	}
+
+	// Sign key
+	userPublicKey, err := parsePublicSSHKey(publicKey)
+	if err != nil {
+		return logical.ErrorResponse(fmt.Sprintf("failed to parse public_key as SSH key: %s", err)), nil
+	}
+
+	err = b.validateSignedKeyRequirements(userPublicKey, role)
```
cipherboy, line 119: This call is actually redundant after the 89...102 and could be removed.

builtin/logical/ssh/path_issue.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	extensions, err := b.calculateExtensions(data, req, role)
+	if err != nil {
+		return logical.ErrorResponse(err.Error()), nil
+	}
+
+	privateKeyEntry, err := caKey(ctx, req.Storage, caPrivateKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CA private key: %w", err)
```
Gabrielopesantos, line 164: I am working on the creation of the common logic function but found something a bit odd, why do some return statements return as `nil, err` and others `response, nil`? Example above.
I was trying to treat everything as `response, nil` but it seems to be breaking the tests.

cipherboy: @Gabrielopesantos it depends on the type of error, I'd return both tbh.

I think these eventually get switched between `errutil.UserError` and `errutil.InternalError` based on something (keywords?). In the former case (line 159 in your diff above), this probably is already a user error. But lacking the private key is more of an internal/operator error than a user error. My 2c.

I _think_ errors from the handler are logged explicitly, whereas `logical.ErrorResponses` aren't logged. I think its maybe [here](https://github.com/hashicorp/vault/blob/7c3c89eec2db780e19d098a72343c426ae1b3536/sdk/logical/response_util.go#L17) ? Not entirely sure :-D 

And yes, there's really no rhyme or reason to one or the other sometimes.

builtin/logical/ssh/path_issue.go
```diff
@@ -0,0 +1,207 @@
...
+
+	signedSSHCertificate := ssh.MarshalAuthorizedKey(certificate)
+	if len(signedSSHCertificate) == 0 {
+		return nil, fmt.Errorf("error marshaling signed certificate")
+	}
+
+	response := &logical.Response{
+		Data: map[string]interface{}{
+			"serial_number":    strconv.FormatUint(certificate.Serial, 16),
+			"certificate":      string(signedSSHCertificate),
```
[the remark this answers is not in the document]
Gabrielopesantos: I changed it to `certificate` because `pki/issue` also has `certificate`, and is more precise (as you said) but then with would not be consistent with ssh/sign.

Gabrielopesantos:
Just pushed a version that should cover most requested changes/comments.

There are two files `path_issue.go` and `path_sign.go` using stuff from each other, maybe joining everything a `path_issue_sign.go` as in the pki package is the way to go.

builtin/logical/ssh/path_issue.go
sgmiller: I don't think that errors are automatically logged.  And yeah, this pattern isn't used consistently at all.

sgmiller:
> Hello @cipherboy, just pushed a version that should cover most requested changes/comments.
> 
> There are two files `path_issue.go` and `path_sign.go` using stuff from each other, maybe joining everything a `path_issue_sign.go` as in the pki package is the way to go.
> 
> What is the behavior expected if no value are defined? Currently, the `key_type` and `key_bits` do not have any default values set. So, if not filled, `key_type` will be an empty string and `key_bits` 0. I think its ok to leave `key_bits` as it is handled in `generateSSHKeyPair`, its set to different values depending on the `key_type`. However, in the case of `key_type`, if not defined it returns an error. Took a look at other endpoints that have a `key_type` field and they seem to have "rsa" as default. Is it ok to also define "rsa" in this endpoint?

I'm fine with either an error or an implicit default.  If the goal is to make the call as simple as possible to make then the default works.

Gabrielopesantos:
Hello, added a few tests which should cover which should the implementation of this new endpoint, as most of it is reusing the existing `ssh/sign`. Let me know if I am missing any case. With this, I am opening the pull request for review.

Also, while testing found something that I am not sure if it expected. I defined a `allowed_user_key_lengths` as:
`"ssh-rsa": int{2048, 3072, 4096}, "ecdsa-sha2-nistp521": 0`, however it seems that it is created as the following.
AllowedUserKeyTypesLengths: map[ecdsa-sha2-nistp521:[0 0] ssh-rsa:[0 0 0 2048 3072 4096]]
Is this expected?

builtin/logical/ssh/path_issue.go
Gabrielopesantos: https://github.com/hashicorp/vault/blob/7c72d9d3d6f127ff368d6f47634684c6fb60cb22/builtin/logical/ssh/path_sign.go#L476-L481


Is it okay to make this global variable as it can be mutated here?
https://github.com/hashicorp/vault/blob/d897316a9e337a2a065ac738c1240490f33eb2ed/builtin/logical/ssh/path_sign.go#L412-L414

cipherboy: Ah sharp eyes, probably best not to mutate it then. I was just thinking that I was regretting writing such complex algorithmic verification...

Gabrielopesantos:
Hello, made a commit that addresses both comments and proposes a solution for this https://github.com/hashicorp/vault/pull/15561#discussion_r887309928 . 

There are a few things that still need to be addressed such as help messages, error messages and creating the changelog. Let me know if am forgetting anything.

builtin/logical/ssh/path_issue.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func (b *backend) pathIssueCertificate(ctx context.Context, req *logical.Request, data *framework.FieldData, role *sshRole, keySpecs *keySpecs) (*logical.Response, error) {
+	publicKey, privateKey, err := generateSSHKeyPair(rand.Reader, keySpecs.Type, keySpecs.Bits)
+	if err != nil {
+		return nil, err
+	}
+
+	if publicKey == "" || privateKey == "" {
+		return nil, fmt.Errorf("failed to generate or parse the keys")
```
cipherboy, line 111: Some of these could probably be `errors.New` as there's no formatting occurring here.

cipherboy: This particular one could be removed, as these should never be empty strings based on `generateSSHKeyPair`'s definition. Since we control that, I'm inclined to elide this check.

cipherboy:
Hey @Gabrielopesantos Sorry for the delay in getting back to this. I think you need to run `make fmt` but otherwise (minus what you identified in your last comment) this looks good :)

builtin/logical/ssh/path_issue.go
```diff
@@ -0,0 +1,265 @@
...
+				}
+			}
+		}
+
+		if !keyAllowed {
+			return nil, fmt.Errorf("key_type provided not in allowed_user_key_types")
+		}
+
+		if !bitsAllowed {
+			return nil, fmt.Errorf("key_bits not in list of allowed values for key_type provided")
```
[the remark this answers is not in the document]
Gabrielopesantos: I am not too good with these descriptions so I will accept every suggestion.

Gabrielopesantos:
No worries, I know this improvement is set for a 1.12.*** milestone so isn't a priority.

cipherboy (approved):
Love the refactor! I'll have a colleague check this out before merging though.

stevendpclark (approved):
Overall this is really great! I have some small nits but nothing preventing this from merging.

 An api documentation update is required for this change though, if you do not feel comfortable writing the updated documentation, we can handle that in a separate PR.

builtin/logical/ssh/path_issue_sign.go
```diff
@@ -0,0 +1,539 @@
...
+	CriticalOptions map[string]string
+	Extensions      map[string]string
+}
+
+func (b *backend) pathSignIssueCertificateHelper(ctx context.Context, req *logical.Request, data *framework.FieldData, role *sshRole, publicKey ssh.PublicKey) (*logical.Response, error) {
+	// Note that these various functions always return "user errors" so we pass
+	// them as 4xx values
+	keyID, err := b.calculateKeyID(data, req, role, publicKey)
+	if err != nil {
+		return logical.ErrorResponse(err.Error()), err
```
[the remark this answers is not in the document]
cipherboy: Not to nitpick your nitpick, but this was just a copy/paste refactor of the existing code:

https://github.com/hashicorp/vault/blob/7c72d9d3d6f127ff368d6f47634684c6fb60cb22/builtin/logical/ssh/path_sign.go#L125-L128

And worst, it seems consistently used that way. :/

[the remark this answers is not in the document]
stevendpclark: Actually there is a behavior change here, as the callers are wrapping any error into a logical.ErrorResponse so some existing error scenarios we did return a 500 will be switched to a 400.

builtin/logical/ssh/backend_test.go
```diff
@@ -1955,6 +2001,39 @@ func signCertificateStep(
...
+		Operation: logical.UpdateOperation,
+		Path:      "issue/" + role,
+		Data: map[string]interface{}{
+			"key_type": keyType,
+			"key_bits": keyBits,
+		},
+		ErrorOk: true,
+		// Returns like nil, err break this
+		Check: func(resp *logical.Response) error {
+			if expectError {
```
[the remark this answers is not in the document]
Gabrielopesantos: Thank you for the feedback, makes sense. Would something like the following after the `if expectError` block be enough to handle it?

```go
if resp.Error() != nil {
    return fmt.Errorf("unexpected error response returned: %v", resp.Error())
}
```

builtin/logical/ssh/path_issue_sign.go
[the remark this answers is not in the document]
Gabrielopesantos: Also makes sense, I remember making this change during development because was having some kind of error. The latest commit reverts these returns to how they were initially (also not consistent). If consistency in the return statements is what we are looking for, I can make the change to `return nil, err` as suggested.

builtin/logical/ssh/backend_test.go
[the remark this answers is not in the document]
stevendpclark: Yup that's perfect, just an FYI the following is a little more common (but don't bother changing it)

```golang
if resp.IsError() {
    return fmt.Errorf("unexpected error response returned: %v", resp.Error())
}
```

builtin/logical/ssh/path_issue_sign.go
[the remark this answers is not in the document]
stevendpclark: Thanks! What you have changed it too, reverting to what it was, is the correct approach, much appreciated.

Gabrielopesantos:
Pushed a commit with the documentation proposal, it is basically a copy of the existing endpoint (with some minor changes), not sure if it enough or might require extra information.

I think I also found an issue with a previous PR (made by me), there is a field in the documentation of the `ssh/sign` endpoint that is not supported as its supposed defined when creating a role (`ssh/roles`).
```
- `not_before_duration` `(duration: "30s")` – Specifies the duration by which to backdate the `ValidAfter` property.
```

cipherboy:
This looks excellent @Gabrielopesantos! Going to go ahead and merge this, and I'll get a fix opened for that docs change you noted (needs to be a separate PR as it goes back to 1.11) :)

PR #15562, opened 2022-05-23
Add more PKI usage best practices to documentation

cipherboy:
1. Add note about cross-cluster CRL URIs
2. Add note that short TTLs are relative to quantity
3. Add note to make sure default is configured
4. Add note about automating certificate renewal

1, 2, and 4 were suggested by @gitrgoliveira, thank you!

### DIFF
--- a/website/content/docs/secrets/pki/considerations.mdx
+++ b/website/content/docs/secrets/pki/considerations.mdx
@@ -163,6 +163,13 @@ OCSP servers manually using the `config/urls` endpoint. It is supported to have
 more than one of each of these by passing in the multiple URLs as a
 comma-separated string parameter.
 
+~> Note: when using Vault Enterprise's Performance Replication features with a
+   PKI Secrets Engine mount, each cluster will have its own CRL; this means
+   each cluster's unique CRL address should be included in the [AIA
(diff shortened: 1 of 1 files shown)


website/content/docs/secrets/pki/considerations.mdx
```diff
@@ -72,6 +72,15 @@ by the HSM and one backed by Vault -- can satisfy both use cases. Operators
 can make roles setting maximum TTLs for each issuer and consumers of the
 mount can decide which to use.
 
+### Always Configure a Default Issuer
+
+For backwards compatibility, [the default issuer](/api-docs/secret/pki#read-issuers-configuration)
+is used to service the existing PKI endpoints. When certificates are revoked
```
sgmiller, line 78: This language assumes a past and present.  "existing" won't mean much to newcomers.  Instead I'd phrase this as the endpoints with no issuer or similar.

sgmiller (changes requested):
Otherwise looks good.

website/content/docs/secrets/pki/considerations.mdx
cipherboy: Good point, thanks!

PR #15565, opened 2022-05-23
Glimmerize and add component test for MountAccessorSelect

Monkeychip:
This PR glimmerize the component `MountAccessorSelect` component, which was modified in the MFA 1.11 work. It also adds component tests.

### DIFF
--- a/ui/app/components/mount-accessor-select.js
+++ b/ui/app/components/mount-accessor-select.js
@@ -13,11 +13,11 @@ import { action } from '@ember/object';
  * ```
  * @param {string} value - the selected value.
  * @param {function} onChange - the parent function that handles when a new value is selected.
- * @param {boolean} [showAccessor] - whether or not you should show the value or the more detailed accessor off the class
+ * @param {boolean} [showAccessor] - whether or not you should show the value or the more detailed accessor off the class.
  * @param {boolean} [noDefault] - whether or not there is a default value.
- * @param {boolean} [filterToken] -whether or not you should filter out type "token".
- * @param {string} [name] - name on the label
- * @param {string} [label] - label above the select input
+ * @param {boolean} [filterToken] - whether or not you should filter out type "token".
+ * @param {string} [name] - name on the label.
+ * @param {string} [label] - label above the select input.
  * @param {string} [helpText] - text shown in tooltip.
  */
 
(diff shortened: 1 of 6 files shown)


ui/app/components/mount-accessor-select.js
```diff
@@ -1,34 +1,49 @@
...
+  }
 
-  authMethods: task(function* () {
+  @dropTask *authMethods() {
     let methods = yield this.store.findAll('auth-method');
-    if (!this.value && !this.noDefault) {
-      this.set('value', methods.get('firstObject.accessor'));
-      this.onChange(this.value);
+    if (!this.args.value && !this.args.noDefault) {
+      // ARG can't set value as args.
```
hellobontempo, line 39: This comment is a tad unclear, not sure if it's a leftover note-to-self or to help explain the following lines.

ui/app/components/mount-accessor-select.js
```diff
@@ -1,34 +1,49 @@
...
-  value: null,
-  filterToken: false,
-  noDefault: false,
-  onChange: () => {},
+export default class MountAccessorSelect extends Component {
+  @service store;
 
-  init() {
-    this._super(...arguments);
+  filterToken = false;
```
hellobontempo, line 27: Since these attrs have default values, would it be better to use getters here? ([docs](https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_arguments))

ui/tests/integration/components/mfa-login-enforcement-form-test.js
```diff
@@ -241,7 +241,12 @@ module('Integration | Component | mfa-login-enforcement-form', function (hooks)
       } else {
         const key = target.label === 'Authentication method' ? 'auth-method' : 'accessor';
         const value = target.label === 'Authentication method' ? 'userpass' : 'auth_userpass_1234';
-        await fillIn(`[data-test-mlef-select="${key}"] select`, value);
+        if (key === 'accessor') {
+          // glimmerized component that does not bind the data-test to the hbs file, and instead has it on the select inside the component
```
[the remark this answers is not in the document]
zofskeez: So this is targeting the `select` element which is nested underneath the element with the test selector applied which should work with `fillIn`. It should work the same as the original implementation since classic components automatically add a wrapping div and apply attributes. I'm fine with leaving as is but happy to help get it working for future reference as well.

PR #15566, opened 2022-05-23
auth/okta: documents API token minimal permissions

austingebauer:
This PR adds documentation to recommended a minimally permissive role to use when generating the API token used by Vault's Okta auth method.


swenson (approved):
LGTM

PR #15569, opened 2022-05-23
Vault documentation: vault overview page proposal

taoism4504:
The writers and I have been trying to standardize on the product overview page based on Laura's [RFC](https://docs.google.com/document/d/1kp5AUJaQg7Wrq472Ebad9Bvi4t5UYn5TvWtjsM2HsA8/edit).

Terraform has already implemented the new[ overview ](https://www.terraform.io/intro)page and I am following suit with Vault's overview page.

:mag: [Deploy Preview](https://vault-git-docs-vault-overview-page-hashicorp.vercel.app/docs/what-is-vault)


yhyakuna (approved):
LGTM

website/content/docs/what-is-vault.mdx
```diff
@@ -18,8 +18,22 @@ difficult and platform-specific. Adding on key rolling, secure storage, and
...
 
-Examples work best to showcase Vault. Please see the
-[use cases](/docs/use-cases).
+### How does Vault work?
+
+![How Vault Works](/img/how-vault-works.png)
+
+Vault validates, authorizes, and provides clients (users, machines, apps) access to secrets or stored sensitive data. The core Vault worklflow consists of four stages:
+
+![Vault Workflow](/img/vault-workflow-diagram1.png)
```
trujillo-adam, line 27: I actually wonder if you need to cut either the graphic or the list below because they say the same things. In fact, I would prefer to keep the text because it would be easier to edit.

website/content/docs/what-is-vault.mdx
```diff
@@ -18,8 +18,22 @@ difficult and platform-specific. Adding on key rolling, secure storage, and
...
+* **Access**: Vault provides access to secrets, keys, and encryption capabilities based on policies associated with the client’s identity.
+
+### Why Vault?
+
+Most enterprises today have credentials sprawled across their organizations, such as passwords, API keys, credentials that are in plain text, app source code, config files, and more. Since these credentials live everywhere, the sprawl can lead enterprises to a state where it's difficult and daunting to really know who has access and authorization to what. Further, having credentials in plain text increases the potential for malicious attacks, both by internal and external attackers. Vault was designed with these challenges in mind by helping you take all these credentials and centralize them so they are defined in one location, thereby, reducing unwanted exposure to credentials. But, Vault takes it a few steps further by making sure clients (users, apps, systems) are authenticated and explicitly authorized to access resources while providing an audit trail that captures and preserves a history of client actions.
```
trujillo-adam, line 36: This is a pretty daunting block of text that I can see a visitor just glossing over. Do you think its possible to edit it down and/or breaking it into a few shorter paragraphs? Maybe turn it into a list of challenges?

website/content/docs/what-is-vault.mdx
```diff
@@ -51,16 +65,75 @@ The key features of Vault are:
   Revocation assists in key rolling as well as locking down systems in the
   case of an intrusion.
 
-## What is HCP Vault?
+-> **Tip:** Learn more about [Vault use cases](/docs/use-cases).
```
trujillo-adam, line 68: FWIW, Laura and I have been using the block quote component instead of callout boxes on What is . . .? pages. This might be a personal thing, but I would also highlight "Use cases" as the keyword instead of "Tip" (or "Note", etc.) because I think it's more meaningful to readers when they are scanning.

website/content/docs/what-is-vault.mdx
```diff
@@ -51,16 +65,75 @@ The key features of Vault are:
...
+
+### What is HCP Vault?
+
+HashiCorp Cloud Platform (HCP) Vault is a hosted version of Vault, which is operated by HashiCorp to allow organizations to get up and running quickly. HCP Vault uses the same binary as self-hosted Vault, which means you will have a consistent user experience. You can use the same Vault clients to communicate with HCP Vault as you use to communicate with a self-hosted Vault.
+
+-> **Tip:** Learn more about [HCP Vault](https://cloud.hashicorp.com/docs/vault) or [get started](https://learn.hashicorp.com/collections/vault/cloud) by setting up a managed Vault cluster.
+
+### Why HCP Vault?
```
trujillo-adam, line 76: Unless this is a requirement from someone above our pay grade, I would consider this trimming this down to the first paragraph and linking out to the HCP site. It might be more information than the visitor might want at this stage if they're just trying to learn about Vault.

trujillo-adam (commented):
Nice work! I added a few questions/concerns.

website/content/docs/what-is-vault.mdx
taoism4504: Thanks Adam. I'm going to have Laura weigh on your suggestion about highlighting 'use cases' instead of 'Tip'. I thought that was the convention used throughout our docs (and do see that convention used as well in Terraform docs). I tend to lean toward highlighting the 'Tip' or 'Note' since linking to the 'use cases' automatically highlights the words.

website/content/docs/what-is-vault.mdx
taoism4504: Ok, I can see where it may appear to be duplicate info. I was thinking of providing a bit more detail in the bulleted list. Basically, add more context to it. I think the visual gives user a nice quick visual of the workflow and they can refer to the bulleted list for more info. @laurapacilio , thoughts on this?

website/content/docs/what-is-vault.mdx
taoism4504: In lieu of a 20-minute video, I opted for text, but I can see how it may be a bit overwhelming. Looks like Jeff made some suggestions regarding the content and even broke it up a bit so it's more palatable.

website/content/docs/what-is-vault.mdx
taoism4504: I might be able to convince the powers  to relocate some of the content to the actual HCP Vault docs :)

website/content/docs/what-is-vault.mdx
laurapacilio: hmmm So I think to make things more accessible, we should probably repeat the workflow in the text and use the bullets to add information that the picture can't capture. The picture should be reinforcing the text. My opinion is that users shouldn't HAVE to even look at the picture to understand the point, but that the picture should kind of....illustrate the points further. Like bonus content :) 

The way we handled this in Terraform docs is we kind of listed out the workflow item int he written bullet and tried to tie it to a real-world example. Whereas the picture just reinforces the workflow and tells the user something about how the Terraform project is organized/communicates with Terraform/the providers. You don't have to do it exactly the same as us! But I'm wondering if you could:
- Start with the list. 
- Rather than repeating the text in the graphic for each stage, could you expand? Give us an understanding of what that actually MEANS in practice. "Vault provides access to secrets, keys, and encryption capabilities based on policies associated with the client’s identity." Could you add like "for example...blah blah". 

This way it's not repetitive. Each has a purpose!

laurapacilio (approved):
Approving so I don't block :) some thoughts for you to consider.  Really great work on this. I know how hard it is to create pages like this. It's coming along really well!!!

PR #15577, opened 2022-05-24
Remove certs with SHA1 from tests

swenson:
In preparation for moving to Go 1.18


swenson:
Closing for now, as this will definitely break that AWS credential engine.

PR #15578, opened 2022-05-24
Test for MFA end user setup

Monkeychip:
This PR tests the following:

1. The status menu either shows Multi-factor Authentication or not depending on if the user has an entityId or not. This was done by modifying the cluster test which was the only place I could find that was doing any status menu testing.

2. On the MFA end user setup that if a valid UUID is entered a QR code is shown.

3. If entering a UUID returns a warning from the API that is shown instead of a QR code and the option to restart the setup is given.

4. Restarting the setup works.


ui/app/templates/components/mfa-setup-step-one.hbs
```diff
@@ -1,26 +1,37 @@
...
-  TOTP Multi-factor authentication (MFA) can be enabled here if it is required by your administrator. This will ensure that
-  you are not prevented from logging into Vault in the future, once MFA is fully enforced.
-</p>
-<form id="mfa-setup-step-one" {{on "submit" this.verifyUUID}}>
-  <MessageError @errorMessage={{this.error}} class="has-top-margin-s" />
-  <div class="field has-top-margin-l">
-    <label class="is-label">
-      Method ID
-    </label>
+<div ...attributes>
```
Monkeychip, line 1: only thing I did to this file was add some data-test selectors and the div here to wrap the hbs in an div with attributes

ui/app/templates/components/mfa-setup-step-two.hbs
```diff
@@ -1,34 +1,49 @@
...
-</p>
-<div class="field has-top-margin-l">
-  <MessageError @errorMessage={{this.error}} class="has-top-margin-s" />
-  {{#if @warning}}
-    <AlertBanner @type="info" @title="MFA enabled" @message={{@warning}} class="has-top-margin-l" />
-  {{else}}
-    <div class="list-item-row">
-      <div class="center-display">
-        <QrCode @text={{@qrCode}} @colorLight="#F7F7F7" @width={{155}} @height={{155}} @correctLevel="L" />
+<div ...attributes>
```
Monkeychip, line 1: again, only thing I did to this file was add some data-test selectors and the div here to wrap the hbs in an div with attributes

zofskeez (approved):
LGTM once the test runs passes. Thanks!

PR #15582, opened 2022-05-24
Vault-5619: Transit BYOK Documentation

rculpepper:
Docs changes for Transit BYOK API endpoints and description for constructing the ciphertext for the key


website/content/docs/secrets/transit.mdx
```diff
@@ -232,6 +232,57 @@ the proper permission, it can use this secrets engine.
     data, since the process would not be able to get access to the plaintext
     data.
 
+## BYOK
```
sgmiller, line 235: Lets be explicit here: "Bring Your Own Key (BYOK)"

sgmiller: Also, I'd like some language discouraging BYOK when not needed.  Something like "While creating a key within the Transit Secrets Engine and letting Vault create and manage the key material is more secure, there may be a need to bring a key from an existing HSM or system.  The key import functionality enables this case."

sgmiller (changes requested):
Just a couple notes.

website/content/docs/secrets/transit.mdx
```diff
@@ -232,6 +232,57 @@ the proper permission, it can use this secrets engine.
...
+### HSM
+
+If the key is being imported from an HSM that supports PKCS#11, there are
+two possible scenarios:
+
+- If the HSM supports the CKM_RSA_AES_KEY_WRAP mechanism, that can be used to wrap the
+target key using the wrapping key.
+
+- Otherwise, two mechanisms can be combined to wrap the target key. First, an AES key should
+be generated and then used to wrap the target key using the CKM_AES_KEY_WRAP_PAD mechanism.
```
cipherboy, line 260: Just a note, this mechanism is wrong. Per the [updated PKCS#11 v3.0 spec](http://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc370634387) and the Transit implementation, this should be `CKM_AES_KEY_WRAP_KWP`.

cipherboy (changes requested):

PR #15587, opened 2022-05-24
Add Patch KV helper

digivava:
This PR is an addition onto the already-merged KV helpers PR (https://github.com/hashicorp/vault/pull/15305), to add a helper method to the api package that allows the Patch command. The logic is mostly the same as the CLI. It also provides a new `WithOption` KVOption to allow for providing an arbitrary key-value option (for future-proofing if we don't get around to adding new With* functions when new options are developed).


api/kv_v2.go
```diff
@@ -38,17 +38,40 @@ type KVVersionMetadata struct {
...
+
+// WithMethod can optionally be passed to dictate which type of
+// patch to perform in a Patch request. If set to "patch", then an HTTP PATCH
+// request will be issued. If set to "rw", then a read will be performed,
+// then a local update, followed by a remote update. Defaults to "patch".
+func WithMethod(method string) KVOption {
+	return func() (string, interface{}) {
+		return KVOptionMethod, method
 	}
 }
```
VinnyHC, line 76: Might be more clear to wrap your new WithOption function:

```suggestion
func WithMethod(method string) KVOption {
	return WithOption(KVOptionMethod, method)
}
```

api/kv_v2.go
```diff
@@ -217,6 +240,72 @@ func (kv *kvv2) Put(ctx context.Context, secretPath string, data map[string]inte
 	return kvSecret, nil
 }
 
+func (kv *kvv2) Patch(ctx context.Context, secretPath string, data map[string]interface{}, opts ...KVOption) (*KVSecret, error) {
```
VinnyHC, line 243: This probably deserves a comment

digivava:
I'm adding the no-changelog label since I plan to tag and release this along with https://github.com/hashicorp/vault/pull/15305 which already has a changelog entry that should cover this in the release notes (as well as future PRs I'm about to do for other KV helper functions like PutMetadata, etc.) 

@mladlow Let me know if you think otherwise!

mladlow:
@digivava changelog plan sounds good!

api/kv_v2.go
```diff
@@ -399,3 +493,87 @@ func extractFullMetadata(secret *Secret) (*KVMetadata, error) {
...
+	if err != nil {
+		// If it's a 405, that probably means the server is running a pre-1.9
+		// Vault version that doesn't support the HTTP PATCH method.
+		// Fall back to the old way of doing it.
+		if re, ok := err.(*ResponseError); ok && re.StatusCode == 405 {
+			return readThenWrite(ctx, client, path, wrappedData)
+		}
+
+		if re, ok := err.(*ResponseError); ok && re.StatusCode == 403 {
+			return nil, fmt.Errorf("received 403 from Vault server; please ensure that token's policy has \"patch\" capability: %w", err)
```
digivava, line 508: Whenever you get a chance to review this @ccapurso , I wanted to ask you if this seems like okay behavior for the `api` package version of this error-handling scenario. 

In the `command` package's version of this method [here](https://github.com/hashicorp/vault/blob/main/command/kv_patch.go#L345-L348), you can see that it doesn't error in this scenario, it throws a warning and resorts back to "rw" style updates. Since it's hard to throw a "warning" in the client library, I thought it might be good to simplify this logic and just error out like I have here. Note that I removed the `rwFallback` variable altogether. However, I can see an argument to be made that these helpers should work the same way as it does in the CLI, so... let me know what you think.

(In case you're wondering why I didn't just reuse those methods altogether, I opted to not share code between api/ and command/ because there is some terminal-printing-specific code in command/ that might be hard to untangle, and we plan on moving the api repo out of this monorepo in the near-ish future anyway.)

digivava: Another idea @VinnyHC had was to perhaps simplify this altogether and make this Patch helper only capable of one style of Patch request. But that would mean either being exclusionary to pre-1.9.0 users or not utilizing the JSON merge patch style... so we would appreciate your context!

VinnyHC: Apologies, I wasn't very clear; my question was around the behavior/signature of the function `Patch` and if we should explicitly require a method to be passed in or have a default with a mechanism to override it.
The current code relies on a default value, where we allow the users to alter the default method by passing in the `WithMethod` KVOption. Should we be more forceful/explicit and have it as an argument in `Patch`?
Like something similar to this:
```go
func (kv *kvv2) Patch(ctx context.Context, secretPath string, data map[string]interface{}, mergeMethod string, opts ...KVOption) {...}
```
Further, if we do the above option would we want to implement something like:
```go
type MergeMethod string

const(
	Patch MergeMethod = "patch"
	RW MergeMethod = "rw"
)

func (kv *kvv2) Patch(ctx context.Context, secretPath string, data map[string]interface{}, m MergeMethod, opts ...KVOption) {...}
```

api/kv_v2.go
```diff
@@ -217,6 +236,81 @@ func (kv *kvv2) Put(ctx context.Context, secretPath string, data map[string]inte
...
+// Patch additively updates the most recent version of a key-value secret,
+// differentiating it from Put which will fully overwrite the previous data.
+// Only the key-value pairs that are new or changing need to be provided.
+//
+// The WithMethod KVOption function can optionally be passed to dictate which
+// kind of patch to perform, as older Vault server versions (pre-1.9.0) may
+// only be able to use the old "rw" (read-then-write) style of partial update,
+// whereas newer Vault servers can use the default value of "patch" if the
+// client token's policy has the "patch" capability.
+func (kv *kvv2) Patch(ctx context.Context, secretPath string, data map[string]interface{}, opts ...KVOption) (*KVSecret, error) {
```
digivava, line 248: @ccapurso What would the user need to provide in the `data` map if they wanted to delete the value at a given key? Is that even possible, with `client.Logical().JSONMergePatch` at least? (I may need to add an explanation to the godoc comment here.)

api/kv_v2.go
```diff
@@ -399,3 +465,92 @@ func extractFullMetadata(secret *Secret) (*KVMetadata, error) {
...
+		"data": newData,
+	}
+	options := make(map[string]interface{})
+	for _, opt := range opts {
+		k, v := opt()
+		options[k] = v
+	}
+	if len(opts) > 0 {
+		wrappedData["options"] = options
+	}
```
ccapurso, line 484: The KVv2 API ignores unexpected options but just wanted to call out that this pattern would be potentially providing things that might not be deemed valid in the future.

api/kv_v2.go
ccapurso: Consistency between the CLI is an interesting topic. The reason the warning exists and a fallback occurs is that the `vault kv patch` command preexisted support for the HTTP `PATCH` method.  Vault needed to provide proper HTTP `PATCH` support via the CLI while ensuring backward compatibility to ensure we didn't break existing workflows. I am not sure we need to do the same here, however. There is less likelihood for someone to take proper action given the programmatic use. With that said, I think that erroring here makes sense.

api/kv_v2.go
```diff
@@ -399,3 +465,92 @@ func extractFullMetadata(secret *Secret) (*KVMetadata, error) {
...
+
+	// Make sure the secret already exists
+	if existingVersion == nil || existingVersion.Data == nil {
+		return nil, fmt.Errorf("no existing secret was found at %s when doing read-then-write patch operation: %w", secretPath, err)
+	}
+
+	// Verify existing secret has metadata
+	if existingVersion.VersionMetadata == nil {
+		return nil, fmt.Errorf("no metadata found at %s; patch can only be used on existing data", secretPath)
```
ccapurso, line 536: Is this case actually possible? A metadata entry must exist for a KVv2 data entry. If there was no metadata, the client would have received a 404 response.

api/kv_v2.go
```diff
@@ -399,3 +465,92 @@ func extractFullMetadata(secret *Secret) (*KVMetadata, error) {
...
+	}
+
+	// Verify existing secret has data
+	if existingVersion.Data == nil {
```
ccapurso, line 540: This check is redundant as it is included in the if statement on line 530.

ccapurso (approved):
Looks good, just a couple small comments. It might also be interesting to add a test where we perform a JSON merge patch for an entry that does not exist to prevent regressions with our error handling as that will result in a 404.

api/kv_v2.go
digivava: Whoops, you're right, thank you!

api/kv_v2.go
digivava: I think you're right. I was basing it off of [this](https://github.com/hashicorp/vault/blob/main/command/kv_patch.go#L261-L275) equivalent check in the CLI code, but now I see those checks are just there as part of the map key check and type assertion.

I'm never sure whether it's better practice to keep checks like this around, where they shouldn't ever happen but would cause a nil pointer deref panic in the WithCheckAndSet line below if someone ever somehow changed the Get code to return nil metadata... shouldn't I keep the check in there for that reason, as unlikely as it is?

api/kv_v2.go
digivava: I see, thanks for the context. I'll keep the RW functionality available this time since I've already coded it here and there's no sense in deleting it if it's helpful for pre-1.9.0 users, but next time maybe I don't need to worry about old Vault servers from before two server versions ago when writing new functions like this.

api/kv_v2.go
digivava: True, but in that case I would hope that the server would just respond with some kind of invalid option 400 error in that scenario so the user would know that their request had failed and why.

ccapurso: Yup, it would in the future. I just didn't know if we wanted to potentially future-proof by filtering it out. I don't think it's strictly necessary _today_ so mostly was pointing it out.

api/kv_v2.go
ccapurso: Ah, that's fair. I think that was mostly guarding against potential panics due to performing type assertions on nil. This check doesn't hurt to keep. Thanks for the context!

VinnyHC (approved):
This looks good to me!

PR #15665, opened 2022-05-27
MFA method acceptance tests

zofskeez:
Adds acceptance tests for mfa method config workflows.


ui/app/templates/vault/cluster/access/mfa/enforcements/enforcement/index.hbs
```diff
@@ -21,30 +21,51 @@
...
   </ToolbarActions>
 </Toolbar>
 
 {{#if (eq this.tab "targets")}}
   {{#each @model.targets as |target|}}
-    <LinkedBlock class="list-item-row" @disabled={{not target.link}} @params={{union (array target.link) target.linkModels}}>
+    <LinkedBlock
+      class="list-item-row"
+      @disabled={{not target.link}}
+      @params={{union (array target.link) target.linkModels}}
```
Monkeychip, line 66: TIL "union" Never seen that before.

ui/lib/core/addon/templates/components/masked-input.hbs
```diff
@@ -1,7 +1,7 @@
 <div
   class="masked-input {{if this.displayOnly 'display-only'}} {{if this.allowCopy 'allow-copy'}}"
-  data-test-masked-input
-  data-test-field
+  data-test-masked-input={{this.name}}
```
Monkeychip, line 3: good call. does it affect other tests however?

ui/tests/acceptance/mfa-login-enforcement-test.js
```diff
@@ -0,0 +1,215 @@
...
+    await click('[data-test-popup-menu-trigger]');
+    await click('[data-test-list-item-link="edit"]');
+    assert.equal(
+      currentRouteName(),
+      'vault.cluster.access.mfa.enforcements.enforcement.edit',
+      'Edit more menu action transitions to enforcement edit route'
+    );
+  });
+
+  test('it should display login enforcement', async function (assert) {
```
Monkeychip, line 103: Do you need to add assert.expect(x) on these tests?

Monkeychip (approved):
A couple of non-blocking comments. Once it passes the UI tests, review is approved.

ui/app/templates/vault/cluster/access/mfa/enforcements/enforcement/index.hbs
zofskeez: First time I've used it myself. It's a part of [ember-composable-helpers](https://github.com/DockYard/ember-composable-helpers#union)

ui/lib/core/addon/templates/components/masked-input.hbs
zofskeez: :disappointed: yes. It's the reason behind the test failures so I reverted these changes. I'm surprised since doing `document.querySelector('some-attribute')` when the attribute has a value will still return the node.

ui/tests/acceptance/mfa-method-test.js
```diff
@@ -0,0 +1,285 @@
...
+        methods.addObjects(this.server.db[`mfa${type}Methods`].where({}));
+        return methods;
+      }, []);
+    return authPage.login();
+  });
+  hooks.after(function () {
+    ENV['ember-cli-mirage'].handler = null;
+  });
+
+  test('it should display landing page when no methods exist', async function (assert) {
```
Monkeychip, line 30: same question as before, you'll know better than me but should we start out every test with assert.expect(x)?

Monkeychip (approved):
LGTM!

ui/tests/acceptance/mfa-login-enforcement-test.js
zofskeez: Oops I forgot to merge main so the file change history was off. This was part of an earlier PR. I will answer the question in the other comment thought!

ui/tests/acceptance/mfa-method-test.js
zofskeez: So it's only required by the lint rule if there are assertions outside of the main scope.
```js
test('foo', async function (assert) {
  assert.expect(1);
  
  this.server.get('/some-endpoint', () => {
    assert.true('endpoint is hit');
  });
});
```
I always used to add it for every test but haven't really in this codebase since no one else really used it until it was required by the lint updates.

Monkeychip: ah, that makes sense. Thanks for the clarification.

PR #15672, opened 2022-05-27
Add note about X.509 SHA-1 deprecation to relevant plugins

swenson:


website/content/docs/deprecation/faq.mdx
```diff
@@ -46,6 +47,31 @@ This workaround will fail in a future version of Go, however, the Go team has no
```
tomhjp, line 47: This will be quite a tricky thing for users to verify they are ready to upgrade. Might it be helpful if we give them some tips for checking existing certificates before the upgrade, and/or e.g. some log lines they should expect to see if they hit this deprecation?

website/content/partials/x509-sha1-deprecation.mdx
```diff
@@ -0,0 +1,5 @@
+~> **Note**: This engine can use external X.509 certificates or load certificates into Vault.
```
cipherboy, line 1: Given the wide set of plugins here, which aren't strictly limited to x.509 certificate usage, you're probably best adding a notice here saying not just use/load but also TLS (which is the bigger place this will likely fail).

website/content/api-docs/secret/kmip.mdx
```diff
@@ -6,6 +6,8 @@ description: This is the API documentation for the Vault KMIP secrets engine.
 
 # KMIP Secrets Engine (API)
 
+@include 'x509-sha1-deprecation.mdx'
```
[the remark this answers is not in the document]
swenson: It could still hit errors if the mTLS certificate has a SHA-1 signature. I think it can also load X.509 certificates to sign as well?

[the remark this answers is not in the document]
cipherboy: I don't think KMIP lets you bring your own mTLS CA. The KMIP creates its own CA and doesn't allow configuration. In the BYO CSR thing to mint the mTLS cred, we'll use our own hash algo there (its from the common certutil code). We don't allow SHA-1 there, so I don't think this is valid. And no, KMIP doesn't allow loading X.509 certs AFAICT. 

Note that on the Go side, _loading_ the cert is fine, it is just _verifying_ the cert (with the SHA-1 sig) that will fail. I'm not sure about signing with SHA-1, presumably that's been disabled as well. 

For general mTLS, I think that should be doc'd elsewhere, not in KMIP.

website/content/partials/x509-sha1-deprecation.mdx
cipherboy: See comment above; in general _loading_ will be fine, its the actual TLS usage that'll fail IIUC.

cipherboy: I don't actually know if _outbound_ connections will verify certs also. That might be worth testing and seeing if Go still allows SHA-1 client certs to a non-SHA-1 server.

website/content/api-docs/secret/kmip.mdx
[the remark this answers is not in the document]
swenson: Ah, that sounds right. I'll remove the notice from KMIP.

website/content/partials/x509-sha1-deprecation.mdx
swenson: Ah, right. But most of the time we're using the certs for TLS connections, or validating signatures (`aws`, `cert` auth methods), so I think this applies. But I'll clarify that it isn't loading, but signature verification.

cipherboy: Yeah, hence my suggestion to include TLS in the mix ;-)

cipherboy (approved):

website/content/docs/deprecation/faq.mdx
```diff
@@ -39,13 +40,48 @@ These features were deprecated in prior releases of Vault. We are targeting the
...
 
 Starting with Vault 1.12.0, Vault will be built with Go 1.18.
-The Go 1.18 standard library [rejects X.509 certificates](https://tip.golang.org/doc/go1.18#sha1) whose signatures use a SHA-1 hash.
+The Go 1.18 standard library X.509 signature validation [rejects signatures](https://go.dev/doc/go1.18#sha1) that use a SHA-1 hash.
 
 If this issue impacts your usage of Vault, you can temporarily work around it by deploying Vault with the environment variable `GODEBUG=x509sha1=1` set.
 This workaround will fail in a future version of Go, however, the Go team has not said when they will remove this workaround.
 
-Here are the use cases that may still use certificates with SHA-1:
+If you want to check if your certificate or CA contains a problematic signature, you can use the OpenSSL CLI:
```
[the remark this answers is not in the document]
swenson:
```suggestion
If you want to check whether a certificate or CA contains a problematic signature, you can use the OpenSSL CLI:
```

Issue #15390, opened 2022-05-12
ServiceAccount incompatability with EKS 1.21+


tim-sendible:
**Describe the bug**
Apologies if this is the wrong place. I received the following alert from AWS, and I think this indicates that Vault is not using a newer version of the Kubernetes client SDK, which will result in ServiceAccounts not working fairly shortly:

```
We have identified applications running in one or more of your Amazon EKS clusters that are not refreshing service account tokens. Applications making requests to Kubernetes API server with expired tokens will fail. You can resolve the issue by updating your application and its dependencies to use newer versions of Kubernetes client SDK that automatically refreshes the tokens.

What is the problem?

Kubernetes version 1.21 graduated BoundServiceAccountTokenVolume feature [1] to beta and enabled it by default. This feature improves security of service account tokens by requiring a one hour expiry time, over the previous default of no expiration. This means that applications that do not refetch service account tokens periodically will receive an HTTP 401 unauthorized error response on requests to Kubernetes API server with expired tokens. You can learn more about the BoundServiceAccountToken feature in EKS Kubernetes 1.21 release notes [2].

To enable a smooth migration of applications to the newer time-bound service account tokens, EKS v1.21+ extends the lifetime of service account tokens to 90 days. Applications on EKS v1.21+ clusters that make API server requests with tokens that are older than 90 days will receive an HTTP 401 unauthorized  error response.

How can you resolve the issue?

To make the transition to time bound service account tokens easier, Kubernetes has updated the below official versions of client SDKs to automatically refetch tokens before the one hour expiration:

* Go v0.15.7 and later
* Python v12.0.0 and later
* Java v9.0.0 and later
* Javascript v0.10.3 and later
* Ruby master branch
* Haskell v0.3.0.0

We recommend that you update your application and its dependencies to use one of the above client SDK versions if you are on an older version. 
```

**To Reproduce**
1. Deploy Vault into EKS 1.22 (we used the official vault helm chart)
2. Ensure that serviceaccount.create is 'true' and that an eks role-arn annotation is added to it.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.10.2
* Vault CLI Version (retrieve with `vault version`): n/a
* Server Operating System/Architecture: EKS 1.22

rodosskiand:
https://www.vaultproject.io/docs/auth/kubernetes

`disable_iss_validation: true` is a workaround?

tim-sendible:
[...]

PR #15675, opened 2022-05-28
docs/k8s: updated helm doc for short-lived SA tokens

tsaarni:
Helm docs still have example for Kubernetes Auth Method that store the SA token in configuration https://www.vaultproject.io/docs/platform/k8s/helm/examples/kubernetes-auth. Update the example to allow rotated tokens to be reloaded automatically.

xref #15390

### DIFF
--- a/website/content/docs/platform/k8s/helm/examples/kubernetes-auth.mdx
+++ b/website/content/docs/platform/k8s/helm/examples/kubernetes-auth.mdx
@@ -14,33 +14,19 @@ In this example, we will walk through how to set up the [Kubernetes Auth Method]
 
 This assumes the following commands will be run inside a Vault pod running in Kubernetes.
 
-You will optionally need the following variables:
-
-```bash
-# JWT is a service account token that has access to the Kubernetes TokenReview API
-# You can retrieve this from inside a pod at: /var/run/secrets/kubernetes.io/serviceaccount/token
-JWT=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
-
-# Address of Kubernetes itself as viewed from inside a running pod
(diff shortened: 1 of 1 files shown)


tsaarni:
This is only doc update so skipping the Changelog entry.

I think there is nothing specific to Helm in this doc, so not sure if the page is absolutely necessary, but since the page now exists it would be nice to have up to date info!

website/content/docs/platform/k8s/helm/examples/kubernetes-auth.mdx
```diff
@@ -14,33 +14,18 @@ In this example, we will walk through how to set up the [Kubernetes Auth Method]
...
-
 Exec into the Vault pod:
 
 ```bash
-kubectl exec -it vault-0 /bin/sh
+kubectl exec -it vault-0 -- /bin/sh
 ```
 
-Then run the following command to configure the Kubernetes Auth Method:
+First, login using the `root` token and then run the following commands to configure the Kubernetes Auth Method:
```
tomhjp, line 23:
```suggestion
If you didn't set `server.dev.enabled=true`, you'll need to log in to Vault first using `vault login`. Then run the following commands to configure the Kubernetes Auth Method:
```

tomhjp (commented):
Sorry for the delay, this looks like a great update, just one suggestion as I don't want us to encourage non-dev installers to use the root token.

website/content/docs/platform/k8s/helm/examples/kubernetes-auth.mdx
tsaarni: @tomhjp Thank you for the review. I've fixed according to comments.

Issue #15661, opened 2022-05-27
LIST sys/quotas/lease-count is not documented


remilapeyre:
**EDIT:** `LIST sys/quotas/lease-count` is not documented at https://www.vaultproject.io/api-docs/system/lease-count-quotas.

Since it was absent from the documentation and that it returns 404 when there is no lease-count quota I thought it did not exist. It actually does and this is just a documentation issue. I will open a PR shortly to document it.

The rest of the message can be ignored as I was initially confused by the absence of documentation. 

---

**Describe the bug**
It is not possible to list the lease-count quotas. 

**To Reproduce**
1. Run `vault write sys/quotas/lease-count/hello max_leases=1000`
2. Run `vault list sys/quotas/lease-count`, the response is a 404 error because this is not implemented. 

**Expected behavior**

It should be possible to list existing lease-counts to check their configuration. The rate-limit quota support this feature. 

Not being able to list them mean that there is no way to know whether some have been set and I think this should count as a bug. This is especially important as we want to remove them from our clusters as there is other bugs affecting lease-counts that can result in dead-lock rendering the leader unresponsive. 

I would gladly contribute a PR but this is part of Vault Enterprise.

This impacts all versions since Vault 1.6.

hsimon-hashicorp:
Hi @remilapeyre - would you be willing to file a support request for this Enterprise enhancement request? The `vault` repo is OSS-specific, so I want to make sure it doesn't get dropped. Thanks!

remilapeyre:
Hi @hsimon-hashicorp, after spending more time looking into this it is just an issue with the documentation. I opened https://github.com/hashicorp/vault/pull/15679 to fix it.

We do also have some issues with lease-counts sometime deadlocking our Vault clusters but it was not the case when I opened this issue. I will contact the support when we have a reproducible example.

PR #15679, opened 2022-05-29
Document `LIST sys/quotas/lease-count`

remilapeyre:
Closes https://github.com/hashicorp/vault/issues/15661


raskchanky (approved):
Thanks @remilapeyre !

PR #15682, opened 2022-05-30
Spawn fewer goroutines to make TestRateLimitQuota_Allow less flaky…

ncabatoff:
…by depending less on Go's scheduler.  Example failure this aims to fix: https://app.circleci.com/pipelines/github/hashicorp/vault/35308/workflows/b2ee4055-2e94-4aa1-a68b-e2214a89dd8c/jobs/456346

### DIFF
--- a/vault/quotas/quotas_rate_limit_test.go
+++ b/vault/quotas/quotas_rate_limit_test.go
@@ -118,11 +118,11 @@ func TestRateLimitQuota_Allow(t *testing.T) {
 		numFail := cr.atomicNumFail.Load()
 
 		// ensure there were some failed requests for the namespace
-		require.NotZerof(t, numFail, "expected some requests to fail; addr: %s, numSuccess: %d, numFail: %d, elapsed: %d", addr, numAllow, numFail, elapsed)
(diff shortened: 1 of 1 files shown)


vault/quotas/quotas_rate_limit_test.go
```diff
@@ -147,46 +147,45 @@ func TestRateLimitQuota_Allow_WithBlock(t *testing.T) {
...
+	defer cancel()
+
 	reqFunc := func(addr string, atomicNumAllow, atomicNumFail *atomic.Int32) {
 		defer wg.Done()
 
-		resp, err := rlq.allow(context.Background(), &Request{ClientAddress: addr})
-		if err != nil {
-			return
-		}
+		for ctx.Err() == nil {
```
stevendpclark, line 156: Shouldn't this have another condition based on time/number of loops? If I understand this code correctly, if `rlq.allow` never returns a failure, the test will never timeout.

ncabatoff: The timeout is embedded in the ctx.

stevendpclark: Oops completely missed that.

stevendpclark (approved):

PR #15687, opened 2022-05-30
Add caching for CRT Go builds

tomhjp:
Adds caching to the CRT Go builds based on the recommended config from here: https://github.com/actions/cache/blob/14c4fd4871149762bbba2312aaf4b031861333d5/examples.md#go---modules

Purely added for speed/DX, as I noticed the checks take quite a long time on PRs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,6 +75,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      # cache/restore go mod
+      - uses: actions/cache@v2
+        with:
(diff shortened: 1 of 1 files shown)


tomhjp:
I only just noticed after looking at the runs on this PR in more detail that the majority of the time is spent building the UI. I'll take another look at that bit.

tomhjp:
This doesn't even seem to reduce Go build times by more than the cache download time unfortunately. And the UI build is already caching based on the lock file. I'll close for now.

PR #15703, opened 2022-05-31
Add integration tests for aliased PKI paths (root/rotate, root/replace)

stevendpclark:
- Add tests for the two api endpoints
 - Also return the issuer_name and key_name fields within the generate root api response
 - Update api-docs for Generate Root to match what we return as api response and the params we accept.

### DIFF
--- a/builtin/logical/pki/path_root.go
+++ b/builtin/logical/pki/path_root.go
@@ -231,6 +231,7 @@ func (b *backend) pathCAGenerateRoot(ctx context.Context, req *logical.Request, 
 	resp.Data["issuer_id"] = myIssuer.ID
 	resp.Data["issuer_name"] = myIssuer.Name
 	resp.Data["key_id"] = myKey.ID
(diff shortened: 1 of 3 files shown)


sgmiller (approved):

builtin/logical/pki/path_root.go
```diff
@@ -231,6 +231,7 @@ func (b *backend) pathCAGenerateRoot(ctx context.Context, req *logical.Request,
 	resp.Data["issuer_id"] = myIssuer.ID
 	resp.Data["issuer_name"] = myIssuer.Name
 	resp.Data["key_id"] = myKey.ID
+	resp.Data["key_name"] = myKey.Name
```
cipherboy, line 234: Do we return `key_name` on the fetch APIs?

stevendpclark: We do not on the fetch issuer api.

cipherboy (approved):

PR #15711, opened 2022-05-31
add codeowners to protect release dirs

modrake:
Locks down changes to release directories so that they require approval from release engineering, and so they have service account rights for pipeline runs and backport-assistant functionality.


claire-labry (approved):
🚢

PR #15714, opened 2022-05-31
UI/client history binary bug

hellobontempo:
Fixes empty state on the binary showing error message: `attempt to map 'null'`


hashishaw (approved):
Good catch! Is there test coverage we could add?

hellobontempo:
> Good catch! Is there test coverage we could add?

I wrote a test when I originally hit this bug after merging some backend changes. But for some reason this only happened on the binary? 😭  For peace of mind I'll see if I can add another test

PR #15715, opened 2022-05-31
Option to show radio button in form field

arnav28:
- For possible values, now we can render either dropdown or radio buttons
- Drop down will be used for larger data set, whereas radio buttons should
  be used when we have only couple of options (example totp mfa)
- Added test for radio button functionality
- Added missing test for ttl without toggle

Screen Shot 2022-05-27 at 11 40 14 AM

### DIFF
--- a/ui/lib/core/addon/components/form-field.hbs
+++ b/ui/lib/core/addon/components/form-field.hbs
@@ -27,12 +27,12 @@
             <RadioButton
               class="radio"
               name={{@attr.name}}
-              id={{or val.value val}}
+              id={{dasherize (or val.value val)}}
               @value={{or val.value val}}
               @onChange={{this.setAndBroadcast}}
               @groupValue={{(get @model this.valuePath)}}
             />
-            <label for={{or val.value val}} class="has-left-margin-xs">
+            <label for={{dasherize (or val.value val)}} class="has-left-margin-xs">
               {{or val.value val}}
             </label>
           </div>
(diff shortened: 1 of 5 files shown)


ui/lib/core/addon/components/form-field.hbs
```diff
@@ -20,22 +20,42 @@
...
-          {{/each}}
-        </select>
+    {{#if (eq @attr.options.editType "radio")}}
+      <div class="control columns is-gapless" data-test-radio-input>
+        {{#each (path-or-array @attr.options.possibleValues @model) as |val|}}
+          <div class="column is-1 is-flex-center has-text-grey">
+            <RadioButton
+              class="radio"
+              name={{@attr.name}}
+              id={{or val.value val}}
```
hashishaw, line 30: should we `dasherize` this value in case it's a value with spaces? (We would have to do the same for the label's `for` attr too)

ui/tests/pages/components/form-field.js
```diff
@@ -41,6 +46,9 @@ export default {
     inputChecked: attribute('checked', 'input[type=checkbox]'),
     selectValue: value('select'),
   }),
+  selectRadioInput: async function (value) {
```
hashishaw, line 49: This seems like it's only used for tests. Is there a different way we could trigger the selection in the test, for example a `click` action on the radio with the desired ID?

hashishaw (commented):
A couple small things, great addition!

ui/tests/pages/components/form-field.js
arnav28: I was thinking since this is part of form field, we can leverage this for future tests also.

hashishaw (approved):
Thank you! 🚀

PR #15719, opened 2022-05-31
Update AWS auth method certificates

swenson:
Add tests that the `rsa2048` document can also be verified using the
`pkcs7` field for AWS auth.

Due to the use of SHA-1-based signatures for the `identity` and `pkcs7`
methods, we want to encourage moving toward using the RSA 2048 workflow,
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-rsa2048.html

This doesn't require code changes for Vault necessarily, but adding in
the (many) certificates will help end users.

Also adds `rsa2048` option to API to fetch the RSA 2048 signature.

I will make a PR to update to the AWS auth docs to document the RSA 2048
flow soon after this.

I ran the AWS acceptance tests manually to verify that the RSA 2048 flow works as expected.


api/auth/aws/aws.go
```diff
@@ -108,7 +109,7 @@ func (a *AWSAuth) Login(ctx context.Context, client *api.Client) (*api.Secret, e
 			}
 			pkcs7 := strings.TrimSpace(resp)
 			loginData["pkcs7"] = pkcs7
-		} else {
+		} else if a.signatureType == identityType {
```
austingebauer, line 112: Curious where you saw that the `identity` type uses SHA-1? I was looking at [our docs](https://www.vaultproject.io/api-docs/auth/aws#signature) and [instance-identity-documents.html#verify-iid](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html#verify-iid) and only see usage of SHA-256. 

Somehow I've managed to not look at the code for this auth method (so far :)) so trying to learn a bit.

builtin/credential/aws/certificates.go
```diff
@@ -0,0 +1,881 @@
...
+		}
+		defaultCertificates = append(defaultCertificates, decodedCert)
+	}
+}
+
+// These certificates are for verifying PKCS#7 DSA signatures.
+// Copied from:
+//   curl https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-pkcs7.html | pcregrep -M -o -e '(?s)-----BEGIN CERTIFICATE-----[^>]*-----END CERTIFICATE-----'
+// Last updated: 2022-05-31
+var pkcs7RawCerts = `-----BEGIN CERTIFICATE-----
```
austingebauer, line 31: nit: perhaps can use `const` for these?

builtin/credential/aws/certificates.go
```diff
@@ -0,0 +1,881 @@
...
+A1UEChMXQW1hem9uIFdlYiBTZXJ2aWNlcyBMTEOCCQCHvR56vNju1DASBgNVHRMB
+Af8ECDAGAQH/AgEAMA0GCSqGSIb3DQEBCwUAA4GBACrKjWj460GUPZCGm3/z0dIz
+M2BPuH769wcOsqfFZcMKEysSFK91tVtUb1soFwH4/Lb/T0PqNrvtEwD1Nva5k0h2
+xZhNNRmDuhOhW1K9wCcnHGRBwY5t4lYL6hNV6hcrqYwGMjTjcAjBG2yMgznSNFle
+Rwi/S3BFXISixNx9cILu
+-----END CERTIFICATE-----`
+
+// These certificates are for verifying RSA 2048 signatures.
+// Copied from:
+// curl https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-rsa2048.html | pcregrep -M -o -e '(?s)-----BEGIN CERTIFICATE-----[^>]*-----END CERTIFICATE-----'
```
austingebauer, line 316: Nice! I was going to ask if you automated pulling these into the code somehow 😄

api/auth/aws/aws.go
```diff
@@ -42,6 +42,7 @@ const (
 	ec2Type              = "ec2"
 	pkcs7Type            = "pkcs7"
 	identityType         = "identity"
+	rsa2048Type          = "rsa2048"
```
austingebauer, line 45: nit: can add this type to the comment on `AWSAuth.signatureType`

austingebauer (commented):
Left a few comments and questions. Generally looks good to me.

api/auth/aws/aws.go
swenson: I believe it is the AWS certificate -- it's signed with sha1.

builtin/credential/aws/certificates.go
swenson: Crafting the right grep was a challenge on macOS :)

austingebauer (approved):
LGTM

swenson:
Thanks!

PR #15720, opened 2022-05-31
[VLT-204] Developer Quickstart

VinnyHC:
This PR introduces the developer quickstart flag for configuring a `-dev` server with some predefined options:

- **Identity System** - The server will be automatically initialized
  with an identity entity and an associated Userpass authentication method for
  exploring the identity system. The entity has a policy associated with it to enable
  reading, writing and deleting secrets.
- **App Role Authentication** - The server will be automatically initialized
  with an App Role authentication method for exploring a common method to
  authenticate applications.
- **Sample Secret** - The server will be automatically initialized
  with a secret stored in the KV secret engine.
- **Sample Policies** - The server will be automatically initialized
  with a collection of sample policies to give access to the
  aforementioned authentication mechanisms.

This can be tested by executing the vault binary with the appropriate flags:
```shell
make dev
vault server -dev -dev-quickstart
```


command/server_dev.go
```diff
@@ -0,0 +1,828 @@
...
+	// identity
+	r = &logical.Request{
+		Operation:   logical.CreateOperation,
+		ClientToken: init.RootToken,
+		Path:        "identity/entity",
+		Data: map[string]interface{}{
+			"name": options.idName,
+			"metadata": map[string]string{
+				"organization": "ACME Inc.",
+				"team":         "Squad-2",
```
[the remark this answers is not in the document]
VinnyHC: At the moment I think it makes sense to keep the `quickstartOptions` struct as lean as possible, only sharing info that needs to be shown in the banner, since we won't be providing a method to manipulate these values. In the future, we can improve this UX but I don't want to paint ourselves into a corner.

website/content/docs/concepts/dev-server.mdx
```diff
@@ -54,3 +54,25 @@ If you're new to Vault, you may want to pick up with [Your First Secret](/intro/
...
+a configured Vault server.
+- **Identity System** - The server will be automatically initialized
+  with an identity entity and an associated `userpass` authentication method for
+  exploring the identity system. The entity has a policy associated with it to enable
+  reading, writing and deleting secrets.
+- **AppRole Authentication** - The server will be automatically initialized
+  with an `approle` authentication method for exploring a common method to
+  authenticate applications.
+- **Sample Secret** - The server will be automatically initialized
+  with a secret stored in the KVv2 secrets engine.
```
digivava, line 72: Would it maybe be worth adding the secret name/relevant CLI command here too so people have another quick reference besides the startup banner?

command/server_dev.go
```diff
@@ -0,0 +1,900 @@
...
+	for name, policy := range options.policyMap {
+		if err := devAddPolicy(ctx, init, core, name, policy); err != nil {
+			return nil, err
+		}
+	}
+
+	// userpass
+	err := qsEnableUserpass(ctx, init, core, options)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initalize quickstart userpass: %w", err)
```
digivava, line 582: Typo! These next few error messages say "initalize" instead of "initialize".

command/server_dev.go
```diff
@@ -0,0 +1,900 @@
...
+func devAddSecret(ctx context.Context, init *vault.InitResult, core *vault.Core, secret quickstartSecret) error {
+	r := &logical.Request{
+		Operation:   logical.UpdateOperation,
+		ClientToken: init.RootToken,
+		Path:        fmt.Sprintf("secret/data/%s", secret.key),
+		Data: map[string]interface{}{
+			"data": secret.data,
+		},
+	}
+	resp, err := core.HandleRequest(ctx, r)
```
digivava, line 617: Just for my own curiosity since I haven't stumbled across this before, what is the difference between using this method and Logical().Write in the API package? (since I think I've seen other code in `command/` using that, or at least Logical().RawRequest)

command/server_dev.go
```diff
@@ -0,0 +1,900 @@
...
+
+	r = &logical.Request{
+		Operation:   logical.CreateOperation,
+		ClientToken: init.RootToken,
+		Path:        "identity/entity-alias",
+		Data: map[string]interface{}{
+			"name":           options.authUserpass.username,
+			"canonical_id":   id,
+			"mount_accessor": accessor,
+			"custom-metadata": map[string]string{
```
digivava, line 759: Should this be `custom_metadata`? Or is this a weird idiosyncrasy of the identity endpoint...

command/server_dev.go
```diff
@@ -0,0 +1,900 @@
...
+	roleID, ok := resp.Data["role_id"].(string)
+	if !ok {
+		return fmt.Errorf("failed to parse role id")
+	}
+	options.authAppRole.roleID = roleID
+
+	r = &logical.Request{
+		Operation:   logical.UpdateOperation,
+		ClientToken: init.RootToken,
+		Path:        fmt.Sprintf("auth/approle/role/%s/secret-id", options.authAppRole.roleName),
```
digivava, line 833: How long does this secret-id last? And is it one-time use?

command/server_dev.go
```diff
@@ -0,0 +1,900 @@
...
+		c.UI.Warn(fmt.Sprintf("Policy Name: %s", name))
+		c.UI.Warn(fmt.Sprintf("Policy Path: %s/v1/sys/policy/%s", e, name))
+		c.UI.Warn(fmt.Sprintf("     Policy: %s", policy))
+		c.UI.Warn("")
+	}
+
+	c.UI.Warn(wrapAtLength("A sample secret is available at:"))
+	c.UI.Warn("")
+	c.UI.Warn(fmt.Sprintf("  Secret Path: %s/v1/secret/data/sample-secret", e))
+	c.UI.Warn(fmt.Sprintf("Metadata Path: %s/v1/secret/metadata/sample-secret", e))
```
digivava, line 899: I wonder if people may say "okay I see the full URL paths here but what do I do with that?"

Perhaps it's worth adding a one-liner like "This secret can be accessed with `vault kv get -mount=secret sample-secret`." ? 

(using that 1.11 mount flag syntax to jump past any /data/ confusion, assuming the dev-quickstart flag isn't going to be backported before 1.11.)

digivava (commented):
Super excited about this flag, great work! Just a few quick items.

command/server_dev.go
VinnyHC: With no options being provided I believe the secret id lasts forever and has infinite uses.
`vault read auth/approle/role/app-read-only`
```text                                                                                                                                                                                                                    
Key                        Value
---                        -----
bind_secret_id             true
local_secret_ids           false
secret_id_bound_cidrs      <nil>
secret_id_num_uses         0
secret_id_ttl              0s
token_bound_cidrs          []
token_explicit_max_ttl     0s
token_max_ttl              0s
token_no_default_policy    false
token_num_uses             0
token_period               0s
token_policies             [read-only]
token_ttl                  0s
token_type                 default
```

command/server_dev.go
VinnyHC: There was concern about making the banner too verbose and difficult to read, but I'm happy to add these in. My worry is that the same reasoning could be applied to the sections, for example, including the CLI command for policy operations, which would indeed make the banner a few lines longer. Maybe this is better suited for the docs?

command/server_dev.go
VinnyHC: This method appears to be the pattern folks use when interacting with server-side resources in situations like this, as opposed to the client-side (which I suppose a good portion of `command` utilizes) where `(*Client).Logical().Write()` is more appropriate. [Example](https://github.com/hashicorp/vault/blob/06c56355b59c14a950de627ac3724769bb3c0834/command/server.go#L1754-L1773)

website/content/docs/concepts/dev-server.mdx
VinnyHC: Yeah, I think that's a great addition! I'll update this with commands and some more details about the resources.

PR #15731, opened 2022-06-01
fix off by one error in activity log nil padding for month data

HridoyRoy:
This PR fixes an off-by-one error in the nil month padding code in the activity log, which I added in the following PR: https://github.com/hashicorp/vault/pull/15420/files

### DIFF
--- a/vault/activity_log.go
+++ b/vault/activity_log.go
@@ -1736,6 +1736,9 @@ func modifyResponseMonths(months []*ResponseMonth, start time.Time, end time.Tim
 		return months
 	}
 	start = timeutil.StartOfMonth(start)
+	if timeutil.IsCurrentMonth(end, time.Now().UTC()) {
+		end = timeutil.StartOfMonth(end).AddDate(0, -1, 0)
+	}
 	end = timeutil.EndOfMonth(end)
 	modifiedResponseMonths := make([]*ResponseMonth, 0)
 	firstMonth, err := time.Parse(time.RFC3339, months[0].Timestamp)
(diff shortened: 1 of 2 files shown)


vault/activity_log.go
```diff
@@ -1736,6 +1736,9 @@ func modifyResponseMonths(months []*ResponseMonth, start time.Time, end time.Tim
 		return months
 	}
 	start = timeutil.StartOfMonth(start)
+	if timeutil.IsCurrentMonth(end, time.Now().UTC()) {
+		end = timeutil.StartOfMonth(end).AddDate(0, -1, 0)
```
swayne275, line 1740: it looks like this is setting `end` to the start of last month. is that intended?

my confusion comes from "should end be the end of a month or the start of a month? i'd think end"

raskchanky (approved):

vault/activity_log.go
HridoyRoy: `end` is a function input so do you mean we should assign `<new_variable> := end`, do the intermediate arithmetic on `new_variable`, and then assign `end = new_variable`?

swayne275: yeah something like that. it seems to me like we're changing the meaning of `end` throughout this code, so it might be a good candidate to use a more targeted new variable?

HridoyRoy: Sounds good, I'll open a separate PR for it!

swayne275: your call - it's not a huge deal but i think it would improve clarity?

PR #15735, opened 2022-06-01
VAULT-6371 Fix issue with lease quotas on read requests that generate leases

VioletHynes:
The crux of the issue here was that lease quota checks get routed differently to rate limit checks. In the case of lease quota checks, the MountPoint was not always present on the logical.Request. In essence, when searching for an applicable quota to apply to the request, it would not correctly find the applicable quota as it uses the MountPoint as part of searching for applicable quotas. It seems to apply only to endpoints that generate a lease on Read.

Lease quotas now apply properly on quotas based on mounts that when generating a lease on a Read:
```
violethynes@violethynes-C02CW1WWMD6R vault-enterprise % vault read mysql-database/creds/my-role
Error reading mysql-database/creds/my-role: Error making API request.

URL: GET http://127.0.0.1:8200/v1/mysql-database/creds/my-role
Code: 429. Errors:

* 1 error occurred:
	* request path "mysql-database/creds/my-role": lease count quota exceeded

```

Tests added as part of enterprise PR.

### DIFF
--- a/changelog/15735.txt
+++ b/changelog/15735.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-quotas/lease-count: Fix lease-count quotas on non-auth mounts not properly being enforced
+quotas/lease-count: Fix lease-count quotas on mounts not properly being enforced when the lease generating request is a read
 ```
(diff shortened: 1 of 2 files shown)


ncabatoff:
> I wasn't sure if there was an appropriate place to put a test for this, as the layers I could find for lease quotas created logical.Requests with the MountPoints included

Any of the tests in quotas_ent_test look like a good basis for a test for this, as they're all blackbox tests using the API, e.g. TestQuotas_LeaseCount_Counting.

ncabatoff (approved):

PR #15741, opened 2022-06-01
Update AWS auth docs for SHA-1 deprecation

swenson:
We now recommend `/rsa2048` as the preferred AWS signature moving
foward, as `/pkcs7` and `/signature` will stop working by default in
Vault 1.12 without setting `GODEBUG=x509sha1=1` in the Vault deployment
due to the move to Go 1.18.

I also took this oppoturnity to try to make the docs less confusing
and more consistent with all of the usages of signature, PKCS#7, DSA,
and RSA terminology.

Feedback definitely welcome.


website/content/api-docs/auth/aws.mdx
```diff
@@ -282,10 +284,21 @@ $ curl \
 ## Create Certificate Configuration
 
 Registers an AWS public key to be used to verify the instance identity
-documents. While the PKCS#7 signature of the identity documents have DSA
-digest, the identity signature will have RSA digest, and hence the public
-keys for each type varies respectively. Indicate the type of the public key
-using the "type" parameter.
+documents. Indicate the type of the public key using the "type" parameter.
```
benashz, line 287: nit: I think we can probably format the parameter as code to be consistent in the other docs.
```suggestion
documents. Indicate the type of the public key using the `type` parameter.
```

website/content/api-docs/auth/aws.mdx
```diff
@@ -1023,15 +1044,23 @@ along with its RSA digest can be supplied to this endpoint.
...
-  `identity` parameter when using the ec2 auth method.
-- `pkcs7` `(string: <required-ec2>)` - PKCS7 signature of the identity document with
-  all `\n` characters removed. Either this needs to be set _OR_ both `identity`
-  and `signature` need to be set when using the ec2 auth method.
+  Either both of this and `signature` must be set _OR_ `pkcs7` must be set
+  when using the ec2 auth method.
+- `signature` `(string: <required-ec2>)` - Base64-encoded SHA256 RSA signature of
+  the instance identity document, which can usually be obtained from the
+  `http://169.254.169.254/latest/dynamic/instance-identity/document` endpoint.
+  Either both of this _AND_ `identity` must be set _OR_ `pkcs7` must be set
```
benashz, line 1056:
```suggestion
  Either both this _AND_ `identity` must be set _OR_ `pkcs7` must be set
```

swenson:
Thanks!

PR #15742, opened 2022-06-01
Add parsing for NSS-wrapped Ed25519 keys

cipherboy:
NSS wraps Ed25519 using the PKCS#8 standard structure. The Go standard
library as of Go 1.18.x doesn't support parsing this key type with the
OID used by NSS; it requires the 1.3.101.112/RFC 8410 format, rather
than the RFC 5915-esque structure supported here.

```
Co-authored-by: Rachel Culpepper <84159930+rculpepper@users.noreply.github.com>
Signed-off-by: Alexander Scheel <alex.scheel@hashicorp.com>
```

### DIFF
--- a/sdk/helper/keysutil/policy.go
+++ b/sdk/helper/keysutil/policy.go
@@ -1385,7 +1385,17 @@ func (p *Policy) Import(ctx context.Context, storage logical.Storage, key []byte
 	} else {
 		parsedPrivateKey, err := x509.ParsePKCS8PrivateKey(key)
 		if err != nil {
-			return fmt.Errorf("error parsing asymmetric key: %s", err)
+			if strings.Contains(err.Error(), "unknown elliptic curve") {
+				var edErr error
+				parsedPrivateKey, edErr = ParsePKCS8Ed25519PrivateKey(key)
+				if edErr != nil {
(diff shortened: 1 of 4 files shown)


sdk/helper/keysutil/policy.go
```diff
@@ -1385,7 +1385,17 @@ func (p *Policy) Import(ctx context.Context, storage logical.Storage, key []byte
 	} else {
 		parsedPrivateKey, err := x509.ParsePKCS8PrivateKey(key)
 		if err != nil {
-			return fmt.Errorf("error parsing asymmetric key: %s", err)
+			if strings.Contains(err.Error(), "unknown elliptic curve") {
```
stevendpclark, line 1388: This seems a little brittle against possible future changes from an error message? 

Are there any reasons we don't just try to call `ParsePKCS8Ed25519PrivateKey` on any err?

cipherboy: We might have a bad ASN.1 blob or something else, that'd trigger a non-unknown curve error. If that's the case, the Ed25519 parsing won't do anything, but give a weird, unrelated error message.

stevendpclark (approved):
Overall this seems good to me, but another pair of eyes is highly recommended before merge.

sgmiller (approved):
lgtm!

cipherboy:
Thanks all! Merging...

PR #15744, opened 2022-06-01
Glimmerize components using outer-html

Monkeychip:
This covers all the components, except `auth-jwt` that use the outer-html component. The idea was to remove outer-html which is an old way to make a tagless template only components. However, auth-jwt is a little more complicated and I'll work on that in a separate PR.

For background on Tagless components and Glimmer [see here](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-tagless-components.md#:~:text=Ember%20allows%20you%20to%20disable,style%20allowed%20with%20Glimmer%20components). 

> Ember allows you to disable the wrapper element on a component (turning it into a “tagless” component). This is now the preferred style and the only style allowed with Glimmer components.

I checked the instance of each of these components to make sure they were still rendering as expected. They are.


ui/app/components/alert-popup.js
```diff
@@ -8,14 +6,22 @@ import OuterHTML from './outer-html';
...
 
-export default OuterHTML.extend({
-  type: null,
-  message: null,
-  close: null,
-});
+import Component from '@glimmer/component';
+
+export default class AlertPopup extends Component {
+  get type() {
```
Monkeychip, line 18: these defaults are required unlike in the wizard-section component.

ui/app/components/alert-popup.js
```diff
@@ -8,14 +6,22 @@ import OuterHTML from './outer-html';
...
+ * @param {function} close=null - The close action which will close the alert.
  *
  */
 
-export default OuterHTML.extend({
-  type: null,
-  message: null,
-  close: null,
-});
+import Component from '@glimmer/component';
```
hellobontempo, line 15: same comment about moving import to the top of the page

ui/app/components/auth-jwt.js
```diff
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import { inject as service } from '@ember/service';
+// ARG NOTE: Once you remove outer-html after glimmerizing you can remove the outer-html component
```
hellobontempo, line 3: Would it be helpful to add this todo onto the excel spreadsheet?

ui/app/components/alert-popup.js
```diff
@@ -8,14 +6,22 @@ import OuterHTML from './outer-html';
  * // All properties are passed in from the flashMessage service.
  *   <AlertPopup @type={{message-types flash.type}} @message={{flash.message}} @close={{close}}/>```
  *
- * @param type=null {String} - The alert type. This comes from the message-types helper.
- * @param [message=null] {String} - The alert message.
- * @param close=null {Func} - The close action which will close the alert.
+ * @param {string} type=null - The alert type. This comes from the message-types helper.
+ * @param {string} [message=null] - The alert message.
+ * @param {function} close=null - The close action which will close the alert.
```
hellobontempo, line 11: I noticed `@isPreformatted` in the template file, should we add it here as a param?

hellobontempo (approved):
some small nits, but nothing blocking! yaaaay glimmerization ✨  thanks for tackling!

ui/app/components/alert-popup.js
Monkeychip: Good call. I just copied over the original documentation, and didn't check if it covered everything. Nice catch.

ui/app/components/auth-jwt.js
Monkeychip: I've been adding comments on the components. You'll see them as little yellow dog ears on the components name. That's where I've been keeping comments, including this one.

hellobontempo: awesome - thank you!

PR #15750, opened 2022-06-02
website: fix internal redirect links

BRKalow:
Fix links which point at redirects defined in `website/redirects.js`

The formatting changes are a result of us running the content through a parser and re-stringifying it from an AST. Functionally nothing should change.


topfunky (approved):
👍🏻

PR #15751, opened 2022-06-02
Support for CPS URLs in Custom Policy Identifiers.

kitography:
Allows a new method of adding Policy Information that enables the use of CPS URLs and (explicit text) user notices.  This is done in one field (in the same field as previously).

Eg: 
```
vault write pki/roles/puppy policy_identifiers='[{"oid":"1.3.6.1.4.1.7.8","notice":"I am a user Notice"}, {"oid":"1.3.6.1.4.1.44947.1.2.4","cps":"https://example.com"}]'

Success! Data written to: pki/roles/puppy
```

On the backend each policy_identifier is stored separately - those might be an oid or string_json form of policy identifier with qualifiers.

### DIFF
--- a/builtin/logical/pki/path_roles.go
+++ b/builtin/logical/pki/path_roles.go
@@ -3,6 +3,7 @@ package pki
 import (
 	"context"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
@@ -384,8 +385,10 @@ for "generate_lease".`,
 			},
 
 			"policy_identifiers": {
-				Type:        framework.TypeCommaStringSlice,
-				Description: `A comma-separated string or list of policy OIDs.`,
+				Type: framework.TypeCommaStringSlice,
+				Description: `A comma-separated string or list of policy OIDs, or a JSON list of qualified policy
+information, which must include an oid, and may include a notice and/or cps url, using the form 
+[{"oid"="1.3.6.1.4.1.7.8","notice"="I am a user Notice"}, {"oid"="1.3.6.1.4.1.44947.1.2.4 ","cps"="https://example.com"}].`,
 			},
 
 			"basic_constraints_valid_for_non_ca": {
@@ -658,7 +661,7 @@ func (b *backend) pathRoleCreate(ctx context.Context, req *logical.Request, data
 		NoStore:                       data.Get("no_store").(bool),
 		RequireCN:                     data.Get("require_cn").(bool),
 		AllowedSerialNumbers:          data.Get("allowed_serial_numbers").([]string),
-		PolicyIdentifiers:             data.Get("policy_identifiers").([]string),
+		PolicyIdentifiers:             getPolicyIdentifier(data, nil),
 		BasicConstraintsValidForNonCA: data.Get("basic_constraints_valid_for_non_ca").(bool),
 		NotBeforeDuration:             time.Duration(data.Get("not_before_duration").(int)) * time.Second,
 		NotAfter:                      data.Get("not_after").(string),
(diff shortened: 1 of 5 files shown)


kitography:
For something that doesn't have breaking tests (needs to use a second parameter):

https://github.com/hashicorp/vault/pull/new/VAULT-5436-pki-support-for-cps-url-in-custom-policy-identifiers-json

builtin/logical/pki/path_roles.go
```diff
@@ -1116,3 +1117,42 @@ const pathListRolesHelpDesc = `Roles will be listed by the role name.`
...
+}
+
+func parsePolicyIdentifiersFromJson(policyIdentifiers string) ([]string, error) {
+	var entries []certutil.PolicyIdentifierWithQualifierEntry
+	var policyIdentifierList []string
+	err := json.Unmarshal([]byte(policyIdentifiers), &entries)
+	if err != nil {
+		return policyIdentifierList, err
+	}
+	policyIdentifierList = make([]string, len(entries))
```
stevendpclark, line 1149: I *believe* you want the following form of make? As I think the append below will start at len(entries) and not 0?

```suggestion
	policyIdentifierList = make([]string, 0, len(entries))
```

stevendpclark, line 1149: I *believe* you want the following form of make? As I think the append below will start at len(entries) and not 0?

```suggestion
	policyIdentifierList = make([]string, 0, len(entries))
```

sdk/helper/certutil/helpers.go
```diff
@@ -446,18 +446,30 @@ func ParsePublicKeyPEM(data []byte) (interface{}, error) {
...
-	for _, oidstr := range data.Params.PolicyIdentifiers {
-		oid, err := StringToOid(oidstr)
+	oidOnly := true
+	for _, oidStr := range data.Params.PolicyIdentifiers {
+		oid, err := StringToOid(oidStr)
 		if err == nil {
 			certTemplate.PolicyIdentifiers = append(certTemplate.PolicyIdentifiers, oid)
 		}
+		if err != nil {
+			oidOnly = false
```
stevendpclark, line 458: I may be way off the mark here but I would have expected us to store non-oid parseable values into a separate list that would then be fed into the `CreatePolicyInformationExtensionFromStorageStrings` call below?

Is there not a chance that we could have some of the same elements now within certTemplate.PolicyIdentifiers and within certTemplate.ExtraExtensions?

stevendpclark, line 458: I may be way off the mark here but I would have expected us to store non-oid parseable values into a separate list that would then be fed into the `CreatePolicyInformationExtensionFromStorageStrings` call below?

Is there not a chance that we could have some of the same elements now within certTemplate.PolicyIdentifiers and within certTemplate.ExtraExtensions?

builtin/logical/pki/path_roles.go
kitography: Thank you!  That solves where my nil policyIdentifiers were coming from :)

sdk/helper/certutil/helpers.go
kitography: Theoretically (though not practically) possible.  If someone had in their policy identifier extensions, before validation was added, the fields {"oid":"2.5.1.4.8"}, and managed to build this with an old version of go, that might be trimmed to numbers, and possibly successfully parsed as a oid, and could be successfully parsed as a policy json object now.

It's not a problem if some of the same elements are in the `certTemplate.PolicyIdentifiers` and `certTemplate.ExtraExtensions[policyInformationExtension]`.  Everything in `certTemplate.PolicyIdentifiers` is overwritten when there is a policy extension in extra extensions.  Since the extra extension is only created if it can parse _all_ the policy identifiers, it will include everything in `certTemplate.PolicyIdentifiers` (no one will look there and be misled).

builtin/logical/pki/path_roles_test.go
```diff
@@ -1019,3 +1025,141 @@ func TestPki_RolePatch(t *testing.T) {
...
+		}
+		// Validate the ASN
+		certificateAsn, err := getPolicyInformationExtensionOffCertificate(*issueResp)
+		if err != nil {
+			t.Fatalf("bad [%d], getting extension from %v err: %v resp: %#v", index, testCase.Input, err, issueResp)
+		}
+		certificateB64 := make([]byte, len(certificateAsn)*2)
+		base64.StdEncoding.Encode(certificateB64, certificateAsn)
+		certificateString := string(certificateB64[:])
+		assert.Contains(t, certificateString, testCase.ASN)
```
sgmiller, line 1134: Appreciate the full round trip here!

sgmiller (approved):
Looks good to me!

stevendpclark (approved):
👍

PR #15753, opened 2022-06-02
Moves json to it's own field (erroring out if both included).

kitography:


kitography:
This isn't really where a second field is useful - and changes API super late in the game.

PR #15754, opened 2022-06-02
PKI: Only set issuers with an associated key as default on import

stevendpclark:
- Do not set the first issuer we attempt to import as the default issuer unless
   it has a corresponding key.
 - Add the ability to set a default issuer if none exist and we import it's corresponding key after the fact.
 - Add a warning to an end-user if we imported multiple issuers with keys and we
   choose one of them as the default value.


cipherboy (approved):
Great catch, thank you @stevendpclark!

PR #15755, opened 2022-06-02
Add PutMetadata and PatchMetadata KV helpers

digivava:
This is the final PR in the series of PRs aimed at creating helper methods in the `api` package for the KV secrets engine. (See https://github.com/hashicorp/vault/pull/15305, https://github.com/hashicorp/vault/pull/15587, and https://github.com/hashicorp/vault/pull/15637 for the others.)

Note: There appears to be a pre-existing bug (or at least I'm unsure of how to get around it) with the `delete_version_after` field which prevents it from being reset back to 0s. If someone can show me how to get around this I would be most appreciative!


api/kv_v2.go
```diff
@@ -30,6 +30,26 @@ type KVMetadata struct {
...
+//
+// The struct's fields are all pointers. When used with the PatchMetadata
+// method, a pointer to a field's zero value (e.g. false for *bool) implies
+// that field should be reset to its zero value after update, whereas a field
+// set to a null pointer (e.g. null for *bool) implies the field should remain
+// unchanged.
+//
+// Since maps are already pointers, use an empty map to remove all
+// custom metadata. To explicitly set the zero value for time.Duration,
+// you can use a pointer to time.ParseDuration("0s").
```
averche, line 45: I believe the default value of time.Duration is a zero value, so no need to use Parse.

api/kv_v2.go
```diff
@@ -238,6 +258,41 @@ func (kv *kvv2) Put(ctx context.Context, secretPath string, data map[string]inte
...
+	md := make(map[string]interface{})
+	casRequiredKey := "cas_required"
+	deleteVersionAfterKey := "delete_version_after"
+	maxVersionsKey := "max_versions"
+	customMetadataKey := "custom_metadata"
+
+	md[customMetadataKey] = metadata.CustomMetadata
+	if metadata.MaxVersions != nil {
+		md[maxVersionsKey] = *(metadata.MaxVersions)
+	}
```
averche, line 280: If we are keeping the input properties values as pointers, should we cast the nil to a zero value so that it actually replaces the existing one?

api/kv_v2.go
```diff
@@ -551,3 +625,69 @@ func readThenWrite(ctx context.Context, client *Client, mountPath string, secret
...
+	customMetadataKey := "custom_metadata"
+
+	// --[Explicit Zero Value For The User, Explicit Null On The Backend]--
+	// The KVMetadataInput struct is designed to have pointer fields so that
+	// the user can easily express the desire to explicitly set a field back to its zero
+	// value (e.g. false), as opposed to just having the field remain
+	// unchanged (e.g. nil).
+	// However, in the actual JSONMergePatch request performed by Logical,
+	// the meanings are reversed, where a null value means to delete the
+	// value at that field, and a zero value means to do nothing.
```
averche, line 643: Would it be confusing for users to have the semantics reversed for two different functions within the same library? Perhaps we should look at other PATCH methods as examples.

api/kv_v2.go
```diff
@@ -551,3 +625,69 @@ func readThenWrite(ctx context.Context, client *Client, mountPath string, secret
 
 	return updatedSecret, nil
 }
+
+func setMetadataMap(metadata KVMetadataInput) (map[string]interface{}, error) {
```
averche, line 629: A better name might be `toMetadataMap` or `constructMetadataMap`

api/kv_v2.go
```diff
@@ -551,3 +625,69 @@ func readThenWrite(ctx context.Context, client *Client, mountPath string, secret
...
+	if metadata.DeleteVersionAfter != nil {
+		derefVal := *metadata.DeleteVersionAfter
+		strDuration := derefVal.String()
+		parsedVal, err := time.ParseDuration(strDuration)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse DeleteVersionAfter duration: %w", err)
+		}
+		if parsedVal == 0 {
+			// TODO: This doesn't work until a bug is fixed.. currently
+			// there seems to be no way to reset the delete_version_after field.
```
averche, line 685: Is the bug on the server side?

vault/external_tests/api/kv_helpers_test.go
```diff
@@ -295,5 +296,137 @@ func TestKVHelpers(t *testing.T) {
...
+			t.Fatalf("did not expect \"cat\" key to be removed")
+		}
+
+		// now let's do another patch to test the "explicit zero value" use case
+		explicitFalse := false
+		explicitZero := 0
+		explicitTimeZero, err := time.ParseDuration("0s")
+		if err != nil {
+			t.Fatal(err)
+		}
```
averche, line 401:
```suggestion
		var (
			explicitFalse    bool
			explicitZero     int
			explicitTimeZero time.Duration
		)
```
I think this would give the same effect of default-initializing the values.

averche:
This looks great overall! I also love the fact that the KV PR's are. split into multiple smaller ones -- it makes it much easier to review!

api/kv_v2.go
digivava: The reason it behaves differently is because the Logical().JSONMergePatch method accepts a map, and this method accepts a struct. With the map can just choose to not include those values if you don't want to modify them, but with a struct, an unset value is `nil` or the zero value, which both mean something specific in JSON merge patch semantics. (Which is why I create my own map out of what I believe they intended with their KVMetadataInput struct, thanks to pointers)

The reason I chose for it to be a struct is just the standard "why statically typed things are nice" argument, in that users can easily see which metadata fields they even have the ability to modify (because as you can see, it's not _all_ of the metadata fields), and they don't have to worry about messing up by accidentally calling the map key `delete_versions_after` instead of `delete_version_after` like I did 500 times. 😆 

We could allow them to do a map instead like Logical().JSONMergePatch, but alas, we'd lose that built-in field knowledge I suppose. I guess the question is, which aspect of the UX is more important (knowing what fields you can edit, or not having to deal with pointers).

Your comment here is definitely worth thinking about though! Curious what others think here too!

(And yeah I'll look at some other PATCH methods, but my experiences with PATCH at previous companies always seemed to be situations where the teams had designed their code in a way where they passed around structs but didn't use pointers, and had to do crazy hacky things to get the zero value working, haha... so I think for PATCH it's a question of either maps or structs-with-pointers but not plain structs.)

api/kv_v2.go
```diff
@@ -30,6 +30,26 @@ type KVMetadata struct {
...
+// a secret without any data yet.
+//
+// The struct's fields are all pointers. When used with the PatchMetadata
+// method, a pointer to a field's zero value (e.g. false for *bool) implies
+// that field should be reset to its zero value after update, whereas a field
+// set to a null pointer (e.g. null for *bool) implies the field should remain
+// unchanged.
+//
+// Since maps are already pointers, use an empty map to remove all
+// custom metadata. To explicitly set the zero value for time.Duration,
```
ccapurso, line 44: Passing an empty map to patch custom metadata will actually leave the original value unchanged.

api/kv_v2.go
```diff
@@ -238,6 +258,41 @@ func (kv *kvv2) Put(ctx context.Context, secretPath string, data map[string]inte
 	return kvSecret, nil
 }
 
+// PutMetadata can be used to fully replace a subset of metadata fields for a
+// given KV v2 secret.
```
ccapurso, line 262: The confusing part about the KVv2 metadata API is that it _mostly_ acts as a patch operation today. In other words, it acts as a patch for all the top-level scalar values. It does _not_ provide patch support for the fields nested under the `custom_metadata` field.

api/kv_v2.go
ccapurso: In general, a null value is interpreted as "remove this field" in JSON merge patches. That's not actually how Vault's patch handling works, however. Vault will not blindly remove all fields whose values are null as that would leave the resource in a bad state. For example, that would allow a user to remove the `created_at` field in a KVv2 metadata entry. Vault actually translates null into the zero value of that type. This happens in `FieldData.getPrimitive`, I believe. A field will only be removed for arbitrary maps, like underlying fields in `custom_metadata`.

api/kv_v2.go
ccapurso: This is in fact a bug that exists in the KV secret engine. It is ultimately due to the underlying [durationpb](https://pkg.go.dev/google.golang.org/protobuf/types/known/durationpb#Duration) type that the `delete_version_after` field uses. The struct members have `omitempty` json tags. The JSON patch library that Vault uses accepts byte arrays of marshaled JSON. A duration of "0s" effectively gets marshaled to an empty map, see [this example](https://go.dev/play/p/dEUyjmHdm2h). I have identified a potential fix but would like to chat with some folks before merging it.

api/kv_v2.go
averche: Thanks for the detailed explanations! This is indeed a tricky issue to get around...

api/kv_v2.go
digivava: Thanks for the thorough review, including of my explanations! This comment is actually in reference to how I'm asking the user to supply it in a KVMetadataInput struct, which (as I explain [here](https://github.com/hashicorp/vault/pull/15755/files#r888240307)), is not actually what gets sent to the /metadata endpoint. 

If you look at the toMetadataMap function, you'll see that I actually take this empty map and change it to a `nil` so that the /metadata endpoint behaves correctly, whereas I don't include nils from the KVMetadataInput struct at all:

```
	if metadata.CustomMetadata != nil {
		if len(metadata.CustomMetadata) == 0 { // empty non-nil map means delete all the keys
			md[customMetadataKey] = nil
		} else {
			md[customMetadataKey] = metadata.CustomMetadata
		}
	}
```

Since the call to the /metadata endpoint is behind the scenes, I figure making the KVMetadataInput struct as easy to use as possible regardless of how our backend endpoint is designed is most important (as in, letting users only set the fields they want to change, and leaving the others untouched i.e. nil.)

This is easy for users imo but hard for us devs and reviewers to understand at a glance, especially when familiar with how the /metadata endpoints work behind the scenes, so let me know if this still doesn't make sense!

ccapurso: Oh gotcha, my mistake! Thanks for the clarification.

api/kv_v2.go
```diff
@@ -30,6 +30,26 @@ type KVMetadata struct {
 	Raw      *Secret
 }
 
+// KVMetadataInput is the subset of metadata that can be manually modified for
+// a KV v2 secret. This can also be passed to the PutMetadata method to create
+// a secret without any data yet.
+//
+// The struct's fields are all pointers. When used with the PatchMetadata
```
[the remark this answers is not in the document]
digivava: Yeah I think you may be right here. I was trying not to double up pointlessly on structs but honestly the pointer thing can be annoying for Put when it's the more basic use case.

api/kv_v2.go
ccapurso: This has been fixed! https://github.com/hashicorp/vault/pull/15792

api/kv_v2.go
digivava: @ccapurso Gotcha! Okay thanks for the clarification there. So I think I can remove the "switch these fields to nil if they're the default value" bit here, since the null only comes into play with custom_metadata. I'll just change this to "if the field in the struct wasn't nil, then use the value (of the deref'd pointer), whatever it was". Thanks!

vault/external_tests/api/kv_helpers_test.go
```diff
@@ -295,5 +296,137 @@ func TestKVHelpers(t *testing.T) {
...
+		if md4.MaxVersions != 0 || md4.CASRequired != false {
+			t.Fatalf("expected fields to be reset to their zero values but they were %d and %t instead", md4.MaxVersions, md4.CASRequired)
+		}
+
+		// TODO: Bring this test back to life once the bug is fixed where we can't actually reset delete-version-after to zero...
+		// if md4.DeleteVersionAfter.String() != "0s" {
+		// 	t.Fatalf("expected delete-version-after to be reset to its zero value but instead it was %d", md4.DeleteVersionAfter.String())
+		// }
+
+		////
```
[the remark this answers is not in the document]
digivava: Totally agree, it's getting unwieldy. I will refactor it.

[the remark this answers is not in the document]
digivava: Okay, so I went off to change this into multiple tests with a `TestMain` function to initialize the TestVaultCluster (since spinning that up and taking that down is what takes the longest, so I want to make sure each test shares the same test cluster). But then I realized `vault.NewTestCluster` and `vault.TestWaitActive` depends on `t`, which TestMain doesn't have access to. 

So what I decided to do is just split this single massive "kv v2 helpers" test up into more `t.Run`s. That way it's hopefully a little easier to read and they can still rely on the same shared test cluster setup.

api/kv_v2.go
digivava: Oh I just realized--Put needs to be its own method separate from Patch because "patch" is actually a capability verb on a policy. Users without "patch" capability but who have "update" need their own method for Put. 

So maybe casting `nil` on Put to the zero value is the right approach, as you first said.

api/kv_v2.go
digivava: Interesting... my tests seem to show PatchMetadata working as expected on the `custom_metadata` field, at least with regard to the checks I've got [here](https://github.com/hashicorp/vault/blob/1773d0b153802b0478358e02f1a89f883ce100ae/vault/external_tests/api/kv_helpers_test.go#L521-L536). 

Or did you just mean that regular PutMetadata will perform a full overwrite of the contents of the custom_metadata field? If so, I think that's fair to expect from a PUT.

api/kv_v2.go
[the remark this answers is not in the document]
averche: > // we haven't specified CASRequired at all, and thus the value is still false!

I think this only applies to PATCH operation, right? A PUT operation will always overwrite the existing values.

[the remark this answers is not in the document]
digivava: Oh that's true huh. Yes now that we're casting nil to the zero value on PUT then yes, I think splitting these up could make it more clear indeed.

api/kv_v2.go
```diff
@@ -238,6 +262,48 @@ func (kv *kvv2) Put(ctx context.Context, secretPath string, data map[string]inte
...
+//
+// To only partially replace the values of these metadata fields, use PatchMetadata.
+//
+// This method can also be used to create a new secret with just metadata and no secret data yet.
+func (kv *kvv2) PutMetadata(ctx context.Context, secretPath string, metadata KVMetadataInput) error {
+	pathToWriteTo := fmt.Sprintf("%s/metadata/%s", kv.mountPath, secretPath)
+
+	// convert values to a map we can pass to Logical,
+	// casting nils to the appropriate zero value
+	md := make(map[string]interface{})
```
averche, line 277: Nit: since the variable is used throughout the scope of the function, I think a longer name is warranted. `metadataMap` maybe? One or two letter names are usually used for short for loop blocks.

api/kv_v2.go
```diff
@@ -417,7 +502,7 @@ func (kv *kvv2) Rollback(ctx context.Context, secretPath string, toVersion int)
 
 func extractCustomMetadata(secret *Secret) (map[string]interface{}, error) {
 	// Logical Writes return the metadata directly, Reads return it nested inside the "metadata" key
-	cmI, ok := secret.Data["custom_metadata"]
+	cmI, ok := secret.Data[customMetadataKey]
```
averche, line 505: Same comment as above regarding variable names. It's hard to tell what `cmI` stands for

api/kv_v2.go
```diff
@@ -427,9 +512,9 @@ func extractCustomMetadata(secret *Secret) (map[string]interface{}, error) {
 		if !ok {
 			return nil, fmt.Errorf("unexpected type for 'metadata' element: %T (%#v)", mI, mI)
 		}
-		cmI, ok = mM["custom_metadata"]
+		cmI, ok = mM[customMetadataKey]
 		if !ok {
-			return nil, fmt.Errorf("metadata missing expected field \"custom_metadata\":%v", mM)
+			return nil, fmt.Errorf("metadata missing expected field \"%s\":%v", customMetadataKey, mM)
```
averche, line 517:
```suggestion
			return nil, fmt.Errorf("metadata missing expected field %q: %v", customMetadataKey, mM)
```

api/kv_v2.go
```diff
@@ -671,3 +756,33 @@ func readThenWrite(ctx context.Context, client *Client, mountPath string, secret
...
+		if len(metadata.CustomMetadata) == 0 { // empty non-nil map means delete all the keys
+			md[customMetadataKey] = nil
+		} else {
+			md[customMetadataKey] = metadata.CustomMetadata
+		}
+	}
+	if metadata.DeleteVersionAfter != nil {
+		derefVal := *metadata.DeleteVersionAfter
+		strDuration := derefVal.String()
+		md[deleteVersionAfterKey] = strDuration
```
averche, line 784:
```suggestion
		md[deleteVersionAfterKey] = (*metadata.DeleteVersionAfter).String()
```
This might be easier to read

vault/external_tests/api/kv_helpers_test.go
```diff
@@ -303,38 +381,196 @@ func TestKVHelpers(t *testing.T) {
...
+		}
+
+		if _, ok := md3.CustomMetadata["cat"]; !ok {
+			t.Fatalf("did not expect \"cat\" key to be removed")
+		}
+	})
+
+	t.Run("patch with explicit zero values", func(t *testing.T) {
+		teardownTest, _ := setupKVv2Test(t)
+		defer teardownTest(t)
```
averche, line 541: Great pattern 👍 Thanks for cleaning up the tests!

api/kv_v2.go
digivava: Ooh, I didn't know about this one, thanks!

api/kv_v2.go
```diff
@@ -46,6 +65,11 @@ const (
 	KVOptionMethod         = "method"
 	KVMergeMethodPatch     = "patch"
 	KVMergeMethodReadWrite = "rw"
+
+	casRequiredKey        = "cas_required"
+	deleteVersionAfterKey = "delete_version_after"
+	maxVersionsKey        = "max_versions"
+	customMetadataKey     = "custom_metadata"
```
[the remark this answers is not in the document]
averche: Sorry, I was not clear. I meant they should still be consts but within the functions:

```go
func blah() {
     const (
	     casRequiredKey        = "cas_required"
	     deleteVersionAfterKey = "delete_version_after"
	     maxVersionsKey        = "max_versions"
	     customMetadataKey     = "custom_metadata"
     )
     ...
}
```

averche (approved):
Great work 👍

PR #15767, opened 2022-06-02
refactor some code in modifyResponseMonths and ensure that the last mo…

HridoyRoy:
…nth comparison with end is comparing end of month with end of month

### DIFF
--- a/vault/activity_log.go
+++ b/vault/activity_log.go
@@ -1736,6 +1736,7 @@ func modifyResponseMonths(months []*ResponseMonth, start time.Time, end time.Tim
 		return months
 	}
 	start = timeutil.StartOfMonth(start)
+	end = timeutil.EndOfMonth(end)
 	if timeutil.IsCurrentMonth(end, time.Now().UTC()) {
 		end = timeutil.EndOfMonth(timeutil.StartOfMonth(end).AddDate(0, -1, 0))
 	}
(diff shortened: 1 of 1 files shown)


vault/activity_log.go
```diff
@@ -1737,13 +1737,11 @@ func modifyResponseMonths(months []*ResponseMonth, start time.Time, end time.Tim
 	}
 	start = timeutil.StartOfMonth(start)
 	if timeutil.IsCurrentMonth(end, time.Now().UTC()) {
-		end = timeutil.StartOfMonth(end).AddDate(0, -1, 0)
+		end = timeutil.EndOfMonth(timeutil.StartOfMonth(end).AddDate(0, -1, 0))
 	}
-	end = timeutil.EndOfMonth(end)
```
swayne275, line 1742: do we still need to pin `end` to be the end of the provided `end` month in the case where `end` is not the current month?

e.g. what if `end` is in the middle of last month?

hellobontempo: is "provided" end month here referencing the `end_date` query param? If so, in the UI we send the 20th of the month, as we expect the backend to re-adjust it to the last day of that month.

hellobontempo: following up, appears to be something different - as the response looks correct on my end. disregard!

HridoyRoy: Fixed!

swayne275 (approved):

PR #15769, opened 2022-06-02
Revert UI: replace localStorage with sessionStorage

hashishaw:
In #14054 we moved from localStorage to using sessionStorage in an effort to make the Vault UI auth sessions tab-specific rather than browser-wide. However, after hearing from the community we are going to revisit our approach to browser security and revert this change for now.


arnav28 (approved):
🚀

PR #15770, opened 2022-06-02
Add details to CHANGELOG and 1.10 upgrade note regarding new 412 error response resulting from SSCTs.

ncabatoff:


VioletHynes (approved):

bkrodgers:
Hi there.  I might be the customer that prompted adding this (I complained to support about this not being listed clearly enough just yesterday after it caused an outage).  While we're still troubleshooting the underlying cause, I don't think this new phrasing is accurate.  We did not see alternate error codes in this situation on the prior version.  The errors themselves are new, were happening 6000-10000 times a day, and immediately started raising complaints we hadn't seen before the upgrade.  Had this been happening but just with a different, more generic error code instead of 412, those complaints would have been the same.  The breaking change isn't that the error code changed, these are new errors.

PR #15777, opened 2022-06-02
Integration tests for mfa method form

arnav28:


ui/app/templates/components/mfa/method-form.hbs
```diff
@@ -28,6 +29,7 @@
   @isActive={{this.isEditModalActive}}
   @confirmText={{@model.type}}
   @onConfirm={{perform this.save}}
+  @testSelector="mfa"
```
hashishaw, line 32: Is this necessary? We should instead follow our convention of using `data-test-*` selectors for test selectors

ui/tests/integration/components/mfa/method-form-test.js
```diff
@@ -0,0 +1,77 @@
...
+  setupMirage(hooks);
+
+  hooks.beforeEach(function () {
+    this.store = this.owner.lookup('service:store');
+    this.model = this.store.createRecord('mfa-method');
+    this.model.type = 'totp';
+  });
+
+  test('it should render correct fields', async function (assert) {
+    await render(hbs`
```
hashishaw, line 18: Can we put `assert.expect` on this test too please?

ui/app/templates/components/mfa/method-form.hbs
arnav28: This Confirmation model component uses testSelector value and prefixes it with data-test-confirmation-modal.

ui/tests/integration/components/mfa/method-form-test.js
arnav28: Sure. 👍

ui/tests/integration/components/mfa/method-form-test.js
```diff
@@ -0,0 +1,77 @@
...
+      <Mfa::MethodForm
+        @hasActions="true"
+        @model={{this.model}}
+      />
+      <div id="modal-wormhole"></div>
+    `);
+    assert.dom('[data-test-input="issuer"]').hasValue('Vault', 'Issuer input is populated');
+    assert.dom('[data-test-ttl-value="Period"]').hasValue('30', 'Period input ttl is populated');
+    let SHA512radioBtn = this.element.querySelectorAll('input[name=algorithm]')[2];
```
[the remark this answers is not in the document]
arnav28: Oh true, thanks!!

PR #15786, opened 2022-06-03
Glimmer continued

Monkeychip:
This PR glimmerizes the following components:

1. generated-item-list
2. hover-copy-button
3. home-link

Small components. It makes it easier to have small contained components glimmerized before hitting the more nuanced components, at least that's what I'm telling myself.

### DIFF
--- a/ui/app/templates/vault/cluster/access/method/item/list.hbs
+++ b/ui/app/templates/vault/cluster/access/method/item/list.hbs
@@ -1,6 +1,5 @@
 <GeneratedItemList
   @model={{this.model}}
-  @method={{this.method}}
   @itemType={{this.itemType}}
   @paths={{this.paths}}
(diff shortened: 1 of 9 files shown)


arnav28 (approved):
🚀

ui/app/templates/vault/cluster/access/method/item/list.hbs
```diff
@@ -1,6 +1,5 @@
 <GeneratedItemList
   @model={{this.model}}
-  @method={{this.method}}
```
arnav28, line 3: Method is not being used?

Monkeychip: Nope, did a double/triple check. It was never in either the js or hbs file. Good catch though.

PR #15789, opened 2022-06-03
UI: calendar widget fix

hashishaw:
Fixes calendar widget allowing user to select an end month prior to the start month (ex. `May 2022 - January 2021`)
calendar-bug


hellobontempo (approved):
thank you for tackling!! 🚀

PR #15810, opened 2022-06-06
docs: Update CSI Provider command line arguments

tomhjp:
Updates the website with the latest flags, and also make some other small improvements to help users link things together.

### DIFF
--- a/website/content/docs/platform/k8s/csi/configurations.mdx
+++ b/website/content/docs/platform/k8s/csi/configurations.mdx
@@ -7,6 +7,12 @@ description: This section documents the configurables for the Vault CSI Provider
 # Command line arguments
 
 The following command line arguments are supported by the Vault CSI provider.
+Most settings support being set by, in ascending order of precedence:
+
+- Environment variables
+- Command line arguments
+- Secret Provider Class parameters
+
 If installing via the helm chart, they can be set using e.g.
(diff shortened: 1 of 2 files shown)


swenson (approved):
LGTM

tvoran (approved):

brettjacobson:
@tomhjp I'd like to point out that maybe it is worth clarifying in these docs; these fields do NOT support quoted values. At least for sure the `vault-addr` and the `vault-tls-ca-cert` options do not.  I am very used to quoting string value options on CLIs, but here they are NOT allowed. And the error messages gave no hints that this was my problem for 4 hours :(  At least they are not allowed to be quoted in the `extraArgs` part of the Helm chart (hopefully the env way of setting these via helm chart will ship soon)

brettjacobson:
@tomhjp I'd like to point out that maybe it is worth clarifying in these docs; these fields do NOT support quoted values. At least for sure the `vault-addr` and the `vault-tls-ca-cert` options do not.  I am very used to quoting string value options on CLIs, but here they are NOT allowed. And the error messages gave no hints that this was my problem for 4 hours :(  At least they are not allowed to be quoted in the `extraArgs` part of the Helm chart (hopefully the env way of setting these via helm chart will ship soon)

PR #15813, opened 2022-06-06
Move Mfa components to MFA folder

Monkeychip:
Cleanup PR: There was an MFA folder where some of the MFA components were living. However, there was an handful of other Mfa components that were in the main component folder. This PR moves those remaining files into the MFA folder.

Before:
image


arnav28 (approved):
Thanks for cleaning it up. 🚀

Issue #15639, opened 2022-05-26
Bump up containerd version to fix CVE-2021-43816


bhmj:
**Is your feature request related to a problem? Please describe.**
The current Vault release (v1.10.3) depends on a github.com/containerd/containerd@v1.5.7 which contains a vulnerability [CVE-2021-43816](https://nvd.nist.gov/vuln/detail/CVE-2021-43816) rated as High/Critical.

**Describe the solution you'd like**
Bump up the imported containerd version to 1.5.9+ since the vulnerability was fixed as described [here](https://github.com/containerd/containerd/security/advisories/GHSA-mvff-h3cj-wj9c).

PR #15816, opened 2022-06-06
Update containerd/containerd indirect test dep

cipherboy:
This dependency is pulled in from our testing infra and not in our final
Vault version. However, updating this dep pulls in newer versions of
other deps (such as `protobuf`) which are used at runtime. Updated via:

```
$ go get github.com/containerd/containerd@v1.5.13 && go mod tidy
```

Since docker/docker has an indirect dep on containerd, I've updated it
as well:

```
$ go get github.com/docker/docker@v20.10.17 && go mod tidy
```

---

Inclined not to backport to 1.11 since this is a testing dep. Anyone know if we need [to update `vault-testing-stepwise`](https://github.com/hashicorp/vault-testing-stepwise/blob/main/go.mod#L6) as well? 

Inclined not to add a changelog as the motivation was for the containerd test dep, but if others feel the other updates warrant a changelog, happy to add one. 

Resolves: #15639 

/cc @mickael-hc


cipherboy:
Marking draft until https://github.com/shirou/gopsutil/pull/1310 is resolved to fix the build.

stevendpclark:
Note that [CVE-2022-31030](https://github.com/advisories/GHSA-5ffw-gxpp-mxpf) has come out since so we should also bump `github.com/containerd/containerd` to 1.5.13

cipherboy:
Per conversation with @ncabatoff, by using `go get` rather than `go get -u`, I've avoided the `x/sys` update. I've also pulled in the newer `containerd/container` and `docker/docker` versions @stevendpclark.

stevendpclark (approved):

cipherboy:
Thanks all! Merging...

Issue #15347, opened 2022-05-10
Ui not showing KV secret metadata tab


JanMa:
**Describe the bug**
When using the UI to access KV secrets, it does not show the metadata of a given secret although the tokens policies grant access to it. Only if the policy grants metadata read access to all secrets, the tab is shown.

**To Reproduce**
1. Create a token with the following policy:
```
# List top-level key/value secrets metadata.
path "secret/metadata/" {
  capabilities = ["list"]
}

# List, create, update, and delete key/value secrets
path "secret/data/team_acme/*" {
  capabilities = ["create", "read", "update", "delete"]
}

path "secret/metadata/team_acme/*" {
  capabilities = ["read", "delete", "list"]
}

path "secret/delete/team_acme/*" {
  capabilities = ["update"]
}

path "secret/undelete/team_acme/*" {
  capabilities = ["update"]
}

path "secret/destroy/team_acme/*" {
  capabilities = ["update"]
}
```
2. Create some secret in the `team_acme` folder
3. Log in with the token and access secret in the UI
4. The `metadata` tab in the secret details is not shown.

![ui-meta-hidden](https://user-images.githubusercontent.com/44572196/167614357-e07264d3-54f2-4883-8444-d490ef8ee5dd.png)

**Expected behavior**
The UI should show the `metadata` tab because the policy allows accessing it.

**Workaround**
If the policy grants `list` and `read` access to all secrets, the `metadata` tab is shown in the secret details.
```
# List top-level key/value secrets metadata.
path "secret/metadata/*" {
  capabilities = ["list", "read"]
}

# List, create, update, and delete key/value secrets
path "secret/data/team_acme/*" {
  capabilities = ["create", "read", "update", "delete"]
}

path "secret/metadata/team_acme/*" {
  capabilities = ["read", "delete", "list"]
}

path "secret/delete/team_acme/*" {
  capabilities = ["update"]
}

path "secret/undelete/team_acme/*" {
  capabilities = ["update"]
}

path "secret/destroy/team_acme/*" {
  capabilities = ["update"]
}
```

![ui-meta-shown](https://user-images.githubusercontent.com/44572196/167614466-251c4c93-e693-4845-a6d3-f23be6513c3e.png)

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.10.2
* Vault CLI Version (retrieve with `vault version`): 1.10.2
* Server Operating System/Architecture: Vault HELM chart

hellobontempo:
[...]

PR #15824, opened 2022-06-06
UI: Fix metadata tab not showing given policy

hashishaw:
This PR fixes the path what capabilities are checked against when evaluating whether to show the metadata tab on KV v2 secrets engine. Since metadata is only ever connected to a secret with an ID, we need to provide the ID on that capability path so that it is checked for each specific secret view. 

Repro steps for the bug are excellently outlined in issue #15347

### DIFF
--- a/ui/tests/acceptance/secrets/backend/kv/secret-test.js
+++ b/ui/tests/acceptance/secrets/backend/kv/secret-test.js
@@ -559,7 +559,7 @@ module('Acceptance | secrets/secret/create', function (hooks) {
     await logout.visit();
     await authPage.login(userToken);
     await settled();
-    await showPage.visit({ backend: enginePath, id: secretPath });
+    await visit(`/vault/secrets/${enginePath}/show/${secretPath}`);
     assert.dom('[data-test-empty-state-title]').hasText('You do not have permission to read this secret.');
     assert.dom('[data-test-secret-metadata-tab]').exists('Metadata tab exists');
     await editPage.metadataTab();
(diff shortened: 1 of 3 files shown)


ui/tests/acceptance/secrets/backend/kv/secret-test.js
```diff
@@ -548,15 +559,15 @@ module('Acceptance | secrets/secret/create', function (hooks) {
     await logout.visit();
     await authPage.login(userToken);
     await settled();
-    await click(`[data-test-auth-backend-link=${enginePath}]`);
-
-    await click(`[data-test-secret-link="${secretPath}"]`);
-
+    await showPage.visit({ backend: enginePath, id: secretPath });
```
hellobontempo, line 562: For some reason the URL here is formatting incorrectly
Screen Shot 2022-06-06 at 3 18 54 PM

navigating directly to the URL seems to work, though
```suggestion
    await visit(`/vault/secrets/${enginePath}/show/${secretPath}`);
```

ui/tests/acceptance/secrets/backend/kv/secret-test.js
```diff
@@ -29,6 +29,18 @@ let writeSecret = async function (backend, path, key, val) {
   return editPage.createSecret(path, key, val);
 };
 
+let deleteEngine = async function (enginePath, assert) {
```
hellobontempo, line 32: 👏

hellobontempo (changes requested):
Just a small test fix! Thanks for tackling

hellobontempo (approved):

PR #15837, opened 2022-06-07
Provide more warnings to users for deprecated or unrecognized paremeters

robmonte:
## Summary

This PR is to showcase Vault providing more warnings when users provide unrecognized or deprecated parameters in requests. Currently, warnings are only given when the request generates a response (e.g. a `vault read` but not a `vault write`). This change adds a response to requests that previously didn't have one, like `vault write`, but only when there are warnings to provide.

A secondary positive impact of this change is that it aids with troubleshooting. For example, if you have a typo in a command, e.g. `vault write consul/roles/management consul_polices="global-management"`, it will warn that "consul_polices" is unrecognized, pointing out to the user that they misspelled the word "policies".

&nbsp;

## Examples using Consul secrets engine

Just an unrecognized parameter:
```
❯ vault write consul/roles/management \
   consul_policies="global-management" \
   not-a-real-parameter="does not exist"

  * Endpoint ignored these unrecognized parameters: [not-a-real-parameter]
```

Just a deprecated parameter:
```
❯ vault write consul/roles/oldtoken \
   policy="$(base64 <<< 'key "" { policy = "read" }')"

  * Endpoint received these deprecated parameters which may be removed in future Vault releases: [policy]
```

Both deprecated parameters and unrecognized parameters:
```
❯ vault write consul/roles/oldtoken \
   not-a-real-parameter="doesn't exist" \
   token_type="client" \
   policy="$(base64 <<< 'key "" { policy = "read" }')"

  * Endpoint received these deprecated parameters which may be removed in future Vault releases: [policy token_type]

  * Endpoint ignored these unrecognized parameters: [not-a-real-parameter]
```

### DIFF
--- a/builtin/logical/database/rotation_test.go
+++ b/builtin/logical/database/rotation_test.go
@@ -81,6 +81,7 @@ func TestBackend_StaticRole_Rotate_basic(t *testing.T) {
 	}
 
 	data = map[string]interface{}{
+		"name":                "plugin-role-test",
 		"db_name":             "plugin-test",
 		"rotation_statements": testRoleStaticUpdate,
 		"username":            dbUser,
@@ -140,12 +141,10 @@ func TestBackend_StaticRole_Rotate_basic(t *testing.T) {
 	}
 
 	// Trigger rotation
-	data = map[string]interface{}{"name": "plugin-role-test"}
 	req = &logical.Request{
 		Operation: logical.UpdateOperation,
 		Path:      "rotate-role/plugin-role-test",
 		Storage:   config.StorageView,
-		Data:      data,
 	}
 	resp, err = b.HandleRequest(namespace.RootContext(nil), req)
 	if err != nil || (resp != nil && resp.IsError()) {
--- a/builtin/logical/database/rotation_test.go
+++ b/builtin/logical/database/rotation_test.go
@@ -81,7 +81,6 @@ func TestBackend_StaticRole_Rotate_basic(t *testing.T) {
 	}
 
 	data = map[string]interface{}{
-		"name":                "plugin-role-test",
 		"db_name":             "plugin-test",
 		"rotation_statements": testRoleStaticUpdate,
 		"username":            dbUser,
--- a/changelog/15837.txt
+++ b/changelog/15837.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+core: Provide more warnings to users when ignored, replaced, or deprecated parameters are provided
+```
--- a/command/kv_put.go
+++ b/command/kv_put.go
@@ -203,7 +203,7 @@ func (c *KVPutCommand) Run(args []string) int {
 	}
 
 	// If the response has only warnings, we still want to print output
-	if secret == nil || checkWarningResponse(secret) {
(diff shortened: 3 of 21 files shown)


robmonte:
This feature is currently incomplete.

This change now creates a response where there previously wasn't for commands like `vault write`. This means that typical output such as \
`Success! Data written to: consul/roles/oldtoken` \
will no longer be printed in the CLI.

Further digging is necessary to resolve this.

changelog/15837.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+core: Provide more warnings to users when ignored, replaced, or deprecated parameters are provided
```
robmonte, line 2: Unsure if "core" is the correct tag here, but using it for now.

builtin/logical/database/rotation_test.go
```diff
@@ -81,7 +81,6 @@ func TestBackend_StaticRole_Rotate_basic(t *testing.T) {
 	}
 
 	data = map[string]interface{}{
-		"name":                "plugin-role-test",
```
robmonte, line 84: This was causing a test failure because the test expected an empty response. The presence of "name" would trigger a response with warnings for replaced parameters because the path already provides the name of the role.

command/kv_test.go
```diff
@@ -262,7 +262,7 @@ func TestKVPutCommand(t *testing.T) {
 		}
 		combined := ui.OutputWriter.String() + ui.ErrorWriter.String()
 		if !strings.Contains(combined, "Success!") {
-			t.Errorf("expected %q to contain %q", combined, "created_time")
+			t.Errorf("expected %q to contain %q", combined, "Success!")
```
robmonte, line 265: This looks like a typo I caught where we check the output for the words "Success!" but then if not found, the outputted error said "created_time" instead.

command/kv_put.go
```diff
@@ -201,11 +201,14 @@ func (c *KVPutCommand) Run(args []string) int {
...
 	}
-	if secret == nil {
+
+	// If the response has only warnings, we still want to print output
+	if secret == nil || checkWarningResponse(secret) {
 		// Don't output anything unless using the "table" format
 		if Format(c.UI) == "table" {
 			c.UI.Info(fmt.Sprintf("Success! Data written to: %s", fullPath))
 		}
+
```
robmonte, line 211: This change is separate from `write.go`'s because a kvv1 put do not return the same output as kvv2.

command/kv_put.go
```diff
@@ -201,11 +201,14 @@ func (c *KVPutCommand) Run(args []string) int {
...
 	}
-	if secret == nil {
+
+	// If the response has only warnings, we still want to print output
+	if secret == nil || secret.IsOnlyWarnings() {
 		// Don't output anything unless using the "table" format
 		if Format(c.UI) == "table" {
 			c.UI.Info(fmt.Sprintf("Success! Data written to: %s", fullPath))
 		}
+
```
robmonte, line 211: This change is separate from `write.go`'s change because a kvv1 put do not return the same output as kvv2.

PR #15852, opened 2022-06-07
Change tooltip for token_bound_certs and glimmerize string-list component

Monkeychip:
This PR modifies the tokenBoundCerts string-list helpText.  Request from HCP.

It also glimmerizes the string-list component.

If the attr.name is tokenBoundCidrs then the help text is modified from the same helpText used on the CLI and API.

**Now:**
image

**Before:**
image


ui/lib/core/package.json
```diff
@@ -17,6 +17,7 @@
     "ember-composable-helpers": "*",
     "ember-concurrency": "*",
     "ember-maybe-in-element": "*",
+    "ember/render-modifiers": "*",
```
Monkeychip, line 20: allows me to use did-render and did-update inside the hbs file

PR #15866, opened 2022-06-08
pass context to postgres queries

bhowe34:
I have observed postgres backend query times far exceed request time in cases where the client cancels the request (usually because postgres is not responding fast enough). This solves the issue by allowing the postgres query to be canceled when the associated request is cancelled.

### DIFF
--- a/changelog/15866.txt
+++ b/changelog/15866.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+physical/postgresql: pass context to queries.
+```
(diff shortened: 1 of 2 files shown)


cipherboy (commented):
This makes sense with me, but asking for review from my Ecosystem colleagues... :)

swenson (approved):
LGTM on behalf of ecosystem

changelog/15866.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+physical/postgresql: pass context to queries.
```
cipherboy, line 2: Perhaps?

```suggestion
physical/postgresql: pass context to queries to propagate timeouts and cancellations on requests.
```

PR #15869, opened 2022-06-08
don't create leases for AWS STS secrets

bhowe34:
AWS STS credentials can neither be renewed nor revoked on a per credential basis (credentials can be revoked by revoking all credentials for a role issued before a given time). Given these limitations creating a lease for the credentials isn't very useful and creates lots of requests to the storage backend, especially when vault processes all existing leases on startup. For these reasons it would make sense to not create leases for AWS STS credentials.

To support existing leases during upgrade the revoke and renew logic will still check for STS leases.


calvn:
Hi @bhowe34, we've discussed this internally and think that this is a reasonable enhancement request. We're currently wrapping up tasks for our next major release, but will continue to track this internally and prioritize reviewing this in the near future.

bhowe34:
Looks like you are correct so I made the same change for federation tokens.

robmonte (approved):
Thanks for your submission! Looks good to me. There's a test case to update to accommodate this change but I'll take care of that in a separate PR.

PR #15876, opened 2022-06-08
Use manual_chain for cross-signed intermediates

cipherboy:
This adds a note that manual_chain is required for cross-signed
intermediates, as Vault will not automatically associate the
cross-signed pair during chain construction. During issuance, the chain
is used verbatim from the issuer, so no chain detection will be used
then.


stevendpclark (approved):
TOC update perhaps on the considerations.mdx?

stevendpclark (approved):

PR #15878, opened 2022-06-08
Only run CI on relevant changesets

cipherboy:

Testing... :-)


cipherboy:
I don't think this approach is viable without a bunch more integration work with CircleCI. I think it'll be easier if/when we switch to GH Actions as a lot of the necessary information is exposed that way.

PR #15880, opened 2022-06-08
Make KV structs public so the KV helper methods show in godoc

digivava:
The new KV helper methods (Get, PutMetadata, etc.) are not that useful if users cannot find them, and various usability studies we've done have shown that many users rely on the [godoc](https://pkg.go.dev/github.com/hashicorp/vault/api) page for discovery of the api package's functionality.

Since the kvv1 and kvv2 structs are private, their methods do not show in the godoc.

By changing them to public but keeping their fields private, we still force users to go through our constructor method (`client.KVv1(mountPath)` and `client.KVv2(mountPath)`) to initiate the structs, but their methods will now be displayed in the godoc.

This also matches how the similarly-designed `Logical` and `Auth` struct work today.

### DIFF
--- a/api/kv.go
+++ b/api/kv.go
@@ -30,8 +30,8 @@ type KVSecret struct {
 //
 // Learn more about the KV secrets engine here:
 // https://www.vaultproject.io/docs/secrets/kv
-func (c *Client) KVv1(mountPath string) *kvv1 {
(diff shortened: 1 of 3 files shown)


VinnyHC (approved):
Great catch!

PR #15882, opened 2022-06-08
docs for activity log noncontiguous billing period changes

HridoyRoy:


website/content/api-docs/system/internal-counters.mdx
```diff
@@ -94,11 +94,14 @@ period, which is represented by the `start_time` and `end_time` parameters.
 
 There are a few things to keep in mind while using this API.
 
-- The activity information only accounts for the activity of the latest
-contiguous months in the billing period. For example, if the billing period is
-from Jan 2022 to June 2022, and the activity subsystem in Vault was not
-capturing data for the months Jan to March, the response will only include
-information for May and June.
+- For vault versions before 1.11, the activity information only accounts for
```
hellobontempo, line 97: small nit: should `vault` be capitalized here?

website/content/api-docs/system/internal-counters.mdx
```diff
@@ -94,11 +94,14 @@ period, which is represented by the `start_time` and `end_time` parameters.
...
-capturing data for the months Jan to March, the response will only include
-information for May and June.
+- For vault versions before 1.11, the activity information only accounts for 
+the activity of the latest contiguous months in the billing period. 
+For example, if the billing period is from Jan 2022 to June 2022, and the
+activity subsystem in Vault was not capturing data for the months Jan to March,
+the response will only include information for May and June.
+
+- As of 1.11, this behavior has been modified to return all available data within 
+the specified billing period.
```
hellobontempo, line 104: Thanks for adding! 👏

hellobontempo (approved):
LGTM! 🚀 
I feel like it should be even clearer that "billing period" means start and end time params (I know it says in the paragraph above, but sometimes folks skim). Perhaps one way to further clarify is calling out that without any date params the response defaults to the last 12 months (so if it's currently June 2022, `/activity` returns Jun 2021-May 2022)

website/content/api-docs/system/internal-counters.mdx
HridoyRoy: Yes, thanks for catching!

website/content/api-docs/system/internal-counters.mdx
HridoyRoy: No problem!

HridoyRoy:
> LGTM! 🚀
> I feel like it should be even clearer that "billing period" means start and end time params (I know it says in the paragraph above, but sometimes folks skim). Perhaps one way to further clarify is calling out that without any date params the response defaults to the last 12 months (so if it's currently June 2022, `/activity` returns Jun 2021-May 2022)

Yeah, I like that! I've added a line to the end of that paragraph detailing what happens without start and end times specified as params.

PR #15886, opened 2022-06-08
pki: Do not use a static issuer/key name within the migration

stevendpclark:
- Selecting a constant default value exposed a possible edge case
   that the migration would fail if a previous migration contained the
   same issuer or key name.


cipherboy (approved):
Makes sense. This was mostly named for convenience and no internal code checks for "current" as a value.

schultz-is (approved):
lgtm

PR #15894, opened 2022-06-08
add OSS noop check for valid ent storage

ccapurso:
Vault Enterprise will need to ensure that a supported storage backend is configured upon startup. This PR includes the OSS pieces to support that check within the `vault server` command.


akshya96 (approved):
LGTM!!

PR #15898, opened 2022-06-08
Remove deprecated core-js version from production builds

zofskeez:
Several dependencies used by Vault UI were using a now deprecated version of core-js (2.6.11) which was causing vulnerability scans to fail. To resolve this, the following steps were taken to remove the deprecated version of core-js from production builds:

- updated `ember-auto-import` to 2.4.2 - [apps are now responsible for requiring webpack](https://github.com/ef4/ember-auto-import/blob/main/docs/upgrade-guide-2.0.md#apps-responsible-for-installing-webpack)
- removed `ember-cli-element-closest-polyfill` - [Element.closest() is supported by all target browsers](https://caniuse.com/element-closest)
- removed `ember-maybe-import-regenerator` - [ES6 generators are supported by all target browsers](https://caniuse.com/es6-generators)
- removed `@babel/polyfill` use (now deprecated and uses deprecated core-js version) via `ember-cli-babel` config
- added `eslint-plugin-compat` for linting code compatibility against target browsers

Doing a search in the vendor file of a production ember build after the changes yields zero results for core-js and/or 2.6.11. Additionally, the build time dropped significantly from 6 minutes to just over 1 minute: 

**Before**
ember-build-main
**After**
ember-build-no-polyfills

It would be ideal to use [@babel/plugin-transform-runtime](https://babeljs.io/docs/en/babel-plugin-transform-runtime) with core-js 3 but I couldn't get it working with `ember-cli-babel` and the `useBabelConfig` option. Hopefully `ember-cli-babel` will be updated to remove the `@babel/polyfill` entirely and support the recommendation to use core-js directly.

Related `ember-cli-babel` issues:
https://github.com/babel/ember-cli-babel/issues/443
https://github.com/babel/ember-cli-babel/issues/444
https://github.com/babel/ember-cli-babel/issues/417
https://github.com/babel/ember-cli-babel/issues/446


hashishaw (approved):
Excellent description of changes and how you approached this. Thank you so much! 🚀

Issue #14043, opened 2022-02-14
Boolean parse errors in Zookeeper backend configuration


tjs-intel:
**Describe the bug**
When providing boolean options to the Zookeeper backend (ex. `tls_enabled`, `tls_skip_verify`) in the Zookeeper configuration HCL, I get the error `unknown type for string *ast.LiteralType` and Vault fails to start.

A workaround is to specify the options as boolean strings instead, ex. `tls_enabled = "true"`.

**To Reproduce**
1. Add an `ha_storage "zookeeper"` stanza with `tls_enabled = true`
2. Start the Vault server

**Expected behavior**
Vault should respect the types in the documentation, ex:
> [tls_enabled](https://www.vaultproject.io/docs/v1.4.x/configuration/storage/zookeeper#tls_enabled) (bool: false) – Specifies if TLS communication with the Zookeeper backend has to be enabled.

**Environment:**
* Vault Server Version (retrieve with `vault status`): 1.4.3 (also tried on 1.9.3)
* Vault CLI Version (retrieve with `vault version`): N/A
* Server Operating System/Architecture: x86_64 GNU/Linux

Vault server configuration file(s):

```hcl
ha_storage "zookeeper" {
  address = ***
  path = "vault/"
  tls_enabled = true
}
```

PR #15900, opened 2022-06-08
Parse ha_storage in config 

akshya96:
https://github.com/hashicorp/vault/issues/14043
The gh issue is specific to zookeeper but this is applicable for all ha_storage backends 
When providing boolean options to an ha_storage backend (ex. tls_enabled, tls_skip_verify) vault cannot parse them. It only accepts string values for boolean fields i.e; "true" and "false"

The parsing works well for storage backends. Therefore, changed the parsing similar to parsing values for storage backend

### DIFF
--- a/changelog/15900.txt
+++ b/changelog/15900.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: Fixes parsing boolean values for ha_storage backends in config
+```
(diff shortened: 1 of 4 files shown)


ccapurso (approved):
Looks good! A small test to verify the changes address https://github.com/hashicorp/vault/issues/14043 would be prudent. There are likely already tests for config parsing at this level so we might just need to modify one slightly.

ccapurso (approved):

PR #15901, opened 2022-06-08
docs/postgres: replaces lib/pq with pgx

austingebauer:
This PR updates documentation that was referencing the [lib/pq](https://github.com/lib/pq) package which is being replaced by [jackc/pgx](https://github.com/jackc/pgx).

Related PR: https://github.com/hashicorp/vault/pull/15343

### DIFF
--- a/website/content/docs/configuration/storage/cockroachdb.mdx
+++ b/website/content/docs/configuration/storage/cockroachdb.mdx
@@ -33,8 +33,8 @@ uses that driver to interact with the database.
 
 - `connection_url` `(string: <required>)` – Specifies the connection string to
   use to authenticate and connect to CockroachDB. A full list of supported
-  parameters can be found in [the pgx library documentation][pgxlib]. For example
-  connection string URLs, see the examples section below.
+  parameters can be found in the [pgx library][pgxlib] and [PostgreSQL connection string][pg_conn_docs]
+  documentation. For example connection string URLs, see the examples section below.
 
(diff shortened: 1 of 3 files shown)


website/content/docs/configuration/storage/cockroachdb.mdx
```diff
@@ -63,4 +63,4 @@ storage "cockroachdb" {
 ```
 
 [cockroachdb]: https://www.cockroachlabs.com/
-[pglib]: https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters
+[pgxlib]: https://pkg.go.dev/github.com/jackc/pgx/v4#hdr-Establishing_a_Connection
```
austingebauer, line 66: There might be a better reference for what params the connection string can take. I wasn't able to find something in the pgx docs outside of the pgconn library it uses. We might be able to reference https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING.

swenson: :+1: on that

swenson (approved):
LGTM; the connection string URL doc fix would also be cool

website/content/docs/configuration/storage/cockroachdb.mdx
austingebauer: Added in https://github.com/hashicorp/vault/pull/15901/commits/ea62e91912cf3d9121928530fce66545e260a11e 👍

calvn (approved):

PR #15904, opened 2022-06-09
Removed IRC reference in architecture internals doc

peteski22:


changelog/15904.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
cipherboy, line 1: I'd probably remove the changelog entry based on the discussion on the earlier thread.

peteski22: No problem, I was thinking the same thing but then succumbed after searching for other doc related changelog entries and saw there were >1 (2 maybe). I'll remove it. 👍🏼

PR #15906, opened 2022-06-09
VAULT-5095 Update docs to reflect that child namespaces do not inherit parent quotas

VioletHynes:
We made a decision on Slack: https://hashicorp.slack.com/archives/CMRSLF924/p1654718659650829

To update the docs instead of fix this issue as it's ultimately a breaking change if we were to enable this behaviour now.

I wasn't sure about the additions in the API docs but it felt like the best place to put the information and I felt it was important that this was documented as it could seem unintuitive (as root namespace quotas do get inherited)

### DIFF
--- a/website/content/api-docs/system/rate-limit-quotas.mdx
+++ b/website/content/api-docs/system/rate-limit-quotas.mdx
@@ -27,7 +27,8 @@ either be a namespace or mount.
   `userpass` in `namespace1`. Updating this field on an existing quota can have
   "moving" effects. For example, updating `auth/userpass` to
   `namespace1/auth/userpass` moves this quota from being a global mount quota to a
-  namespace specific mount quota. **Note, namespaces are supported in Enterprise only**.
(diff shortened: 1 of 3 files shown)


website/content/api-docs/system/rate-limit-quotas.mdx
```diff
@@ -27,7 +27,8 @@ either be a namespace or mount.
   `userpass` in `namespace1`. Updating this field on an existing quota can have
   "moving" effects. For example, updating `auth/userpass` to
   `namespace1/auth/userpass` moves this quota from being a global mount quota to a
-  namespace specific mount quota. **Note, namespaces are supported in Enterprise only**.
+  namespace specific mount quota. Quotas on a namespace are not inherited by child
```
ncabatoff, line 30:
```suggestion
  namespace specific mount quota. Quotas on a non-root namespace are not inherited by child
```

ncabatoff (approved):

PR #15908, opened 2022-06-09
Import SDK for 1.8.13 version

stevendpclark:
Import new SDK version from #15907 and a `go mod tidy`


mladlow (commented):
I'm not as comfortable reviewing this type of PR when there are other go.sum changes - I wouldn't necessarily expect to see the gocql and snappy changes here, and would like to know that they're expected & understood before I approve.

stevendpclark:
Fair enough, seems like the same thing happened within #15141. I'll revert the `go.sum` changes for now and not perform the `go mod tidy` on 1.9.x either.

PR #15911, opened 2022-06-09
Glimmerize mount-backend-form

Monkeychip:
This PR glimmerizes the mount-backend-form component. I went ahead and tackled this one because we will be working in this component for the upcoming PKI work. It will make it a little easier for the future work having this component updated. Specifically, having tracked properties instead of computed.

### DIFF
--- a/ui/app/components/mount-backend-form.js
+++ b/ui/app/components/mount-backend-form.js
@@ -78,7 +78,7 @@ export default class MountBackendForm extends Component {
     // change it here to match the new type
     let isUnchanged = list.findBy('type', currentPath);
     if (!currentPath || isUnchanged) {
-      mount.set('path', type);
+      mount.path = type;
     }
   }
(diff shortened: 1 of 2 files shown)


ui/app/components/mount-backend-form.js
```diff
@@ -86,112 +80,113 @@ export default Component.extend({
     if (!currentPath || isUnchanged) {
       mount.set('path', type);
```
hellobontempo, line 81: since it's glimmerized, can we do `mount.path = type` here?

hellobontempo, line 81: since it's glimmerized, can we do `mount.path = type` here?

Monkeychip: Yes, let me check that out.

arnav28 (approved):

PR #15913, opened 2022-06-09
Update minimum required go version for sdk

digivava:
The sdk package is now relying on `//go:build` directives, so the minimum required Go version needs to be `go 1.17`.


cipherboy (approved):
Looks good to me! Updating to 1.17 and `go mod tidy` afterwards gives me the same result.

PR #15915, opened 2022-06-09
Update minimum required go version for api

digivava:
Now that `sdk` requires a minimum version of go 1.17 to work, api should reflect the same since sdk is a dependency.


cipherboy (approved):
Looks good to me, I was able to reproduce it locally. Thank you!

PR #15920, opened 2022-06-09
VAULT-6091 Document Duration Format String

VioletHynes:
This duration format string is used in many, many places in the docs. I've tried to add a central page to describe it, and then reference this central page from the places that reference it. In many places, they had their own, unique ways of describing this string format.

I likely missed somewhere but I did my best to search out every page that was referencing this string in any way. Let me know if I missed anything obvious.


website/content/docs/configuration/time-format.mdx
```diff
@@ -0,0 +1,39 @@
+---
+layout: docs
+page_title: Time String Format
+description: A description of Vault's time string format used throughout its configuration.
+---
+
+# Time String Format
+
+Vault uses a unique but ultimately simple time string format as part of its configuration and APIs.
```
taoism4504, line 9:
```suggestion
Vault uses a unique time string format as part of its configuration and APIs.
```

website/content/docs/configuration/time-format.mdx
```diff
@@ -0,0 +1,39 @@
...
+This format is used throughout Vault wherever time units, such as durations, need to be given. For
```
taoism4504, line 10: suggestion: replace 'need to be given' with 'are provided'. Or, you can also say 'are given'.

taoism4504 (approved):
minor nit/suggestion, but LG. As far as placement of the information, I'm inclined to say it look like it belongs under Concepts though I can see why you put it under Configuration. The reason why I say Concepts is because the whole page, especially, the 'how it works' part, lays it out conceptually. I think there would be a stronger case for it to live under Configuration if you  provided some configuration examples. My 2cents.

website/content/docs/configuration/telemetry.mdx
```diff
@@ -29,7 +29,7 @@ The following options are available on all telemetry configurations.
 
 - `usage_gauge_period` `(string: "10m")` - Specifies the interval at which high-cardinality
   usage data is collected, such as token counts, entity counts, and secret counts.
-   A value of "none" disables the collection.
+   A value of "none" disables the collection.  Uses [time format strings](/configuration/time-format).
```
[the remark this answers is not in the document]
VioletHynes: This is a good point - these are definitely durations, not times. I'll update

website/content/docs/configuration/time-format.mdx
ncabatoff, line 10: In some places we reference https://golang.org/pkg/time/#ParseDuration, which is almost what Vault does (the differences IIRC are that we also support the `d` unit, and we assume no unit means seconds.) 

I suspect some of those references to ParseDuration are incorrect and we actually support "vault style durations" there.  There do seem to be a handful where we really are only support Go-style durations though, e.g. pg & mssql & mysql's ttls, pki crls, which may be bugs.  And then command-line flags that take durations are a hybrid, where we treat no-unit as seconds but we don't support the 'd' unit... which is probably also a bug.

Originally I was going to say that you should qualify that "wherever", but I think this exceptions are few enough and probably fixable that I guess we can let it stand.  Instead, do you want to search for references we make to the Go ParseDuration in mdx files and point them to your doc?

VioletHynes: Agreed - I think it makes sense to consider inconsistencies as bugs until proven otherwise.

After this initial review, I was planning to go through and update all of the many, many pages that reference this - so updating anywhere that references ParseDuration would definitely be included, but thanks for calling it out!

ncabatoff (approved):

PR #15923, opened 2022-06-09
When database plugin Clean is called, close connections in goroutines

swenson:
We are seeing long pre-seal times when nodes are being shut down, which
seems to be delays in releasing the global stateLock, which further
seems to be delayed by cleaning up the builtin database plugin.

The database plugin `Clean()` eventually calls, synchronously, `Close()`
on every database instance that has been loaded. The database instance
`Close()` calls can take a long time to complete (e.g.,
[mongo](https://github.com/hashicorp/vault/blob/v1.10.3/plugins/database/mongodb/connection_producer.go#L132)
can take up to a minute to close its client before it will return).

So, instead, we change the `Clean()` method to close all of the database
instances in goroutines so they will be closed more quickly and won't
block the release of the global stateLock. This should be safe because
`Clean()` will only be called when the plugin is being unloaded, and so
the connections will no longer be used.

In addition, this adds metrics tracking how many of each type of
databases are being used by the builtin database plugin.

### DIFF
--- a/builtin/logical/database/backend.go
+++ b/builtin/logical/database/backend.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/armon/go-metrics"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-secure-stdlib/strutil"
 	"github.com/hashicorp/go-uuid"
@@ -280,8 +281,28 @@ func (b *databaseBackend) GetConnectionWithConfig(ctx context.Context, name stri
 		id:       id,
 		name:     name,
 	}
+	b.addConnectionsCounter(dbw, 1)
 	b.connections[name] = dbi
 	return dbi, nil
+}
+
+// addConnectionsCounter is used to keep metrics on how many and what types of databases we have open
+func (b *databaseBackend) addConnectionsCounter(dbw databaseVersionWrapper, amount int) {
+	// keep track of what databases we open
+	labels := make([]metrics.Label, 0, 2)
+	dbType, err := dbw.Type()
+	if err == nil {
+		labels = append(labels, metrics.Label{Name: "type", Value: dbType})
+	} else {
+		b.Logger().Debug("Error getting database type", "err", err)
+		labels = append(labels, metrics.Label{Name: "type", Value: "unknown"})
+	}
+	if dbw.isV4() {
+		labels = append(labels, metrics.Label{Name: "version", Value: "4"})
+	} else {
+		labels = append(labels, metrics.Label{Name: "version", Value: "5"})
+	}
+	metrics.IncrCounterWithLabels([]string{"secrets", "database", "backend", "connections", "count"}, float32(amount), labels)
 }
 
 // invalidateQueue cancels any background queue loading and destroys the queue.
@@ -311,6 +332,7 @@ func (b *databaseBackend) clearConnection(name string) error {
 	if ok {
 		// Ignore error here since the database client is always killed
 		db.Close()
+		b.addConnectionsCounter(db.database, -1)
 		delete(b.connections, name)
 	}
 	return nil
@@ -331,6 +353,7 @@ func (b *databaseBackend) CloseIfShutdown(db *dbPluginInstance, err error) {
 			// Ensure we are deleting the correct connection
 			mapDB, ok := b.connections[db.name]
 			if ok && db.id == mapDB.id {
+				b.addConnectionsCounter(db.database, -1)
 				delete(b.connections, db.name)
 			}
 		}()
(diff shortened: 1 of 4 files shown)


builtin/logical/database/mocks_test.go
```diff
@@ -36,8 +36,7 @@ func (m *mockNewDatabase) DeleteUser(ctx context.Context, req v5.DeleteUserReque
 }
 
 func (m *mockNewDatabase) Type() (string, error) {
-	args := m.Called()
-	return args.String(0), args.Error(1)
+	return "mock", nil
```
calvn, line 39: Is there a reason why we're hardcoding the return here? 

I don't think we're currently doing this, but this wouldn't allow us to mock up errors from this method on tests if we needed to via something like `mockDB.On("Type").Return("", errors.New("error from type"))`.

swenson: If we don't, we have to add `mockDB.On("Type")` to nearly every method or the tests will panic, since we use it in metrics collection.

Since this isn't used for much else that I saw, it seemed safer to just hardcode it here.

builtin/logical/database/backend.go
```diff
@@ -336,18 +359,26 @@ func (b *databaseBackend) CloseIfShutdown(db *dbPluginInstance, err error) {
...
-		db.Close()
-	}
+	// copy all connections so we can close asynchronously
+	connectionsCopy := b.connections
 	b.connections = make(map[string]*dbPluginInstance)
+
+	for _, db := range connectionsCopy {
+		go func(db *dbPluginInstance) {
+			b.addConnectionsCounter(db.database, -1)
+			db.Close()
```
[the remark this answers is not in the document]
swenson: Hmm. I mean, theoretically the process closing should close all of the associated sockets if they hadn't already been cleanly closed, which should terminate all of the connections.

builtin/logical/database/backend.go
```diff
@@ -280,10 +281,30 @@ func (b *databaseBackend) GetConnectionWithConfig(ctx context.Context, name stri
...
+	} else {
+		b.Logger().Debug("Error getting database type", "err", err)
+		labels = append(labels, metrics.Label{Name: "type", Value: "unknown"})
+	}
+	if dbw.isV4() {
+		labels = append(labels, metrics.Label{Name: "version", Value: "4"})
+	} else {
+		labels = append(labels, metrics.Label{Name: "version", Value: "5"})
+	}
+	metrics.IncrCounterWithLabels([]string{"secrets", "database", "backend", "connections", "count"}, float32(amount), labels)
```
calvn, line 305: Totally a nit here, but we could eliminate the `else`'s here with something like this:
```suggestion
	dbType, err := dbw.Type()
	if err != nil {
		b.Logger().Debug("Error getting database type", "err", err)
		dbType = "unknown"
	} 
	version := 5
	if dbw.isV4() {
		version = 4
	}
	
	labels := []metrics.Label{{"type", dbType}, {"version", version}}
	metrics.IncrCounterWithLabels([]string{"secrets", "database", "backend", "connections", "count"}, float32(amount), labels)
```

swenson: I like it. Thanks!

builtin/logical/database/backend.go
```diff
@@ -308,6 +327,7 @@ func (b *databaseBackend) clearConnection(name string) error {
 	if ok {
 		// Ignore error here since the database client is always killed
 		db.Close()
+		b.addConnectionsCounter(db.database, -1)
```
ncabatoff, line 330: Technically counters should only ever increase, I think you want a gauge instead.

builtin/logical/database/backend.go
```diff
@@ -27,6 +30,32 @@ const (
 	minRootCredRollbackAge = 1 * time.Minute
 )
 
+// mutable for tests only
+var cleanupMaxWaitTime = 500 * time.Millisecond
+
+// metrics collection
+var (
+	gaugeSync = sync.Mutex{}
+	gauges    = map[string]*int32{}
```
ncabatoff, line 39: I'm always discomforted when I see non-readonly globals, can we pull this into databaseBackend instead?  I'm also not sure we need a generic gauge manager, why not just have an atomic `numConnections` field in databaseBackend?  That way we won't have to worry about the lock to protect the map and any possible unpleasantness it might cause us.

swenson: A single `numConnections` wouldn't allow us to see the different types of databases though.

I wanted to avoid putting these into `databaseBackend` because we might have multiple `databaseBackend`s running, and we probably would like to aggregate these gauges across all of them. With gauges (instead of counters), I don't *think* that `armon/go-metrics` or `statsd` will automatically aggregate them for us (whereas with counters, I believe `statsd` at least would).

builtin/logical/database/backend.go
```diff
@@ -339,18 +386,45 @@ func (b *databaseBackend) CloseIfShutdown(db *dbPluginInstance, err error) {
...
-	for _, db := range b.connections {
-		db.Close()
-	}
+	// copy all connections so we can close asynchronously
+	connectionsCopy := b.connections
 	b.connections = make(map[string]*dbPluginInstance)
+
+	// we will try to wait for all the connections to close
+	// ... but not too long
+	sem := semaphore.NewWeighted(int64(len(connectionsCopy)))
```
swenson, line 408: This seems like a useful pattern that I have seen messed up a lot, and with [a lot](https://stackoverflow.com/questions/32840687/timeout-for-waitgroup-wait) of bad answers for on the internet; I'm tempted to move this out to a library since [I don't see one already made](https://pkg.go.dev/search?q=waitgroup) (what I want is something like a "TimedWaitGroup" or "WaitGroupWithTimeout".

swenson:
#15944 might be a better, simpler version of this. If it is approved and merged, I can close this and perhaps add a different PR for just adding metrics, the semaphore/waitgroup, or both.

builtin/logical/database/backend.go
swenson: Ideally, I would like to add a tag with the instance of the database backend -- maybe the mount path? But I don't see any way to get that. We could tag by the BackendUUID though.

swenson:
I'm going to close this and rework it for just metrics, I think. Thanks everyone!

PR #15929, opened 2022-06-10
ssh: Do not convert errors into logical.ErrorResponse in issue path

stevendpclark:
Do not wrap all errors into logical.ErrorResponse as I believe that switches the return code from a 5XX to a 4XX. Also if the `pathSignIssueCertificateHelper` returns an error response value, do not add data to it.


cipherboy (approved):
Makes sense to me.

PR #15933, opened 2022-06-10
database plugin: Invalidate queue should cancel context first

swenson:
To signal to any credentials rotating goroutines that they should cancel
pending operations, which reduces lock contention.

### DIFF
--- a/builtin/logical/database/backend.go
+++ b/builtin/logical/database/backend.go
@@ -286,12 +286,15 @@ func (b *databaseBackend) GetConnectionWithConfig(ctx context.Context, name stri
 
 // invalidateQueue cancels any background queue loading and destroys the queue.
 func (b *databaseBackend) invalidateQueue() {
-	b.Lock()
-	defer b.Unlock()
-
+	// cancel context before grabbing lock to start closing any open connections
+	// this is safe to do without the lock since it is only written to once in initialization
+	// and can be canceled multiple times safely
 	if b.cancelQueue != nil {
 		b.cancelQueue()
 	}
+	b.Lock()
+	defer b.Unlock()
+
 	b.credRotationQueue = nil
 }
 


builtin/logical/database/backend.go
```diff
@@ -286,12 +286,15 @@ func (b *databaseBackend) GetConnectionWithConfig(ctx context.Context, name stri
...
-	b.Lock()
-	defer b.Unlock()
-
+	// cancel context before grabbing lock to start closing any open connections
+	// this is safe to do without the lock since it is only written to once in initialization
+	// and can be canceled multiple times safely
 	if b.cancelQueue != nil {
 		b.cancelQueue()
 	}
+	b.Lock()
```
ncabatoff, line 295: Technically I would argue we don't need the lock, and we don't need to nil out credRotationQueue, since that field is only initialized by Factory, so it's not like databaseBackend can get usefully recycled.  But it's probably safest to make the minimal change, so ignore me.

ncabatoff (approved):

builtin/logical/database/backend.go
austingebauer: +1, doesn't seem like the lock is needed. I'm also okay with leaving it.

austingebauer (approved):

builtin/logical/database/backend.go
calvn: That's true, though there are several places were we check whether `credRotationQueue` nil or not and only perform things like rotation (async) if non-nil so not sure if it'd impact those calls.

calvn (approved):

builtin/logical/database/backend.go
swenson: Like @calvn says, I think we have to leave the lock since there are several times when we check if `credRotationQueue` is `nil` and if not, then read from it. Not having the lock would introduce a potential (unlikely) race condition where we set `credRotationQueue` nil here, but after the `!= nil` check elsewhere but before they use `credRotationQueue`.

swenson:
Thanks everyone!

builtin/logical/database/backend.go
swenson: The more I think about it, the more I think we can safely use a different pattern instead of doing the `credRotationQueue != nil` pattern. Something like:

```go
select {
case <-b.ctx.Done():
default:
  // credRotationQueue is valid, use it
}
```

And instead of setting `credRotationQueue = nil`, we just rely on the context that we're canceling anyway.

I'll see if I can take a stab at this, and limiting the use of the other lock, in another PR I think.

PR #15934, opened 2022-06-10
Convert /internal/ui/mounts to publicly supported endpoint

AnPucel:


website/content/api-docs/system/mounts-all.mdx
```diff
@@ -1,35 +1,31 @@
 ---
 layout: api
-page_title: /sys/internal/ui/mounts - HTTP API
+page_title: /sys/mounts-all - HTTP API
 description: >-
-  The `/sys/internal/ui/mounts` endpoint is used to manage mount listing
+  The `/sys/mounts-all` endpoint is used to manage mount listing
```
hashishaw, line 5: Something to be aware of in the documentation is that this endpoint only returns mount paths which the current user has access to. This is very different from most of our endpoints so I think it should be called out. 

Maybe also a callout that this endpoint must be present in the policy for the UI to work as expected.

hashishaw (approved):
Approved from the UI perspective (all files under `/ui`) ✅

maxb:
This change seems rather disruptive to me. It willl break at least `vault kv ...` CLI commands if the client and server versions are not both with or both without this change, as well as some functionality in consul-template, and all of the third party code which has already observed this is the only way to cleanly support both KVv1 and KVv2 access transparently.

A more practical way forward might be to simply document that the endpoint is now a stable API for general usage, despite its path.

AnPucel:
This is going to be a disruptive change because consul-template relies on this path and it breaks the integration tests. I'm going to close this PR and update the ticket.

PR #15944, opened 2022-06-10
Cleanup and simplify lock usage in database plugin

swenson:
Following up from discussions in #15923 and #15933, I wanted to split
out a separate PR that drastically reduced the complexity of the use of
the databaseBackend lock. We no longer need it at all for the
`credRotationQueue`, and we can move it to be solely used in a few,
small connections map management functions.

### DIFF
--- a/builtin/logical/database/backend.go
+++ b/builtin/logical/database/backend.go
@@ -58,10 +58,9 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
 	// Create a context with a cancel method for processing any WAL entries and
 	// populating the queue
 	initCtx := context.Background()
-	ictx, cancel := context.WithCancel(initCtx)
-	b.cancelQueue = cancel
+	b.ctx, b.cancelQueue = context.WithCancel(initCtx)
 	// Load queue and kickoff new periodic ticker
-	go b.initQueue(ictx, conf, conf.System.ReplicationState())
+	go b.initQueue(b.ctx, conf, conf.System.ReplicationState())
 	return b, nil
 }
 
@@ -110,13 +109,14 @@ func Backend(conf *logical.BackendConfig) *databaseBackend {
 }
 
 type databaseBackend struct {
+	// used to synchronize access to the connections map
+	connLock sync.RWMutex
 	// connections holds configured database connections by config name
 	connections map[string]*dbPluginInstance
 	logger      log.Logger
 
 	*framework.Backend
-	sync.RWMutex
-	// CredRotationQueue is an in-memory priority queue used to track Static Roles
+	// credRotationQueue is an in-memory priority queue used to track Static Roles
 	// that require periodic rotation. Backends will have a PriorityQueue
 	// initialized on setup, but only backends that are mounted by a primary
 	// server or mounted as a local mount will perform the rotations.
@@ -124,12 +124,44 @@ type databaseBackend struct {
 	// cancelQueue is used to remove the priority queue and terminate the
 	// background ticker.
 	credRotationQueue *queue.PriorityQueue
-	cancelQueue       context.CancelFunc
+	// context used for canceling operations
+	ctx         context.Context
+	cancelQueue context.CancelFunc
 
 	// roleLocks is used to lock modifications to roles in the queue, to ensure
 	// concurrent requests are not modifying the same role and possibly causing
 	// issues with the priority queue.
 	roleLocks []*locksutil.LockEntry
+}
+
+func (b *databaseBackend) connGet(name string) *dbPluginInstance {
+	b.connLock.RLock()
+	defer b.connLock.RUnlock()
+	return b.connections[name]
+}
+
+func (b *databaseBackend) connPop(name string) *dbPluginInstance {
+	b.connLock.Lock()
+	defer b.connLock.Unlock()
+	dbi := b.connections[name]
+	delete(b.connections, name)
+	return dbi
+}
+
+func (b *databaseBackend) connPut(name string, newDbi *dbPluginInstance) *dbPluginInstance {
+	b.connLock.Lock()
+	defer b.connLock.Unlock()
+	dbi := b.connections[name]
+	b.connections[name] = newDbi
+	return dbi
+}
+
+func (b *databaseBackend) connClear() map[string]*dbPluginInstance {
+	b.connLock.Lock()
+	defer b.connLock.Unlock()
+	old := b.connections
+	b.connections = make(map[string]*dbPluginInstance)
+	return old
 }
 
 func (b *databaseBackend) DatabaseConfig(ctx context.Context, s logical.Storage, name string) (*DatabaseConfig, error) {
@@ -236,22 +268,8 @@ func (b *databaseBackend) GetConnection(ctx context.Context, s logical.Storage, 
 }
 
 func (b *databaseBackend) GetConnectionWithConfig(ctx context.Context, name string, config *DatabaseConfig) (*dbPluginInstance, error) {
-	b.RLock()
-	unlockFunc := b.RUnlock
-	defer func() { unlockFunc() }()
-
-	dbi, ok := b.connections[name]
-	if ok {
-		return dbi, nil
-	}
-
-	// Upgrade lock
-	b.RUnlock()
-	b.Lock()
-	unlockFunc = b.Unlock
-
-	dbi, ok = b.connections[name]
-	if ok {
+	dbi := b.connGet(name)
+	if dbi != nil {
 		return dbi, nil
 	}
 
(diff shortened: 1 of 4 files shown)


builtin/logical/database/backend.go
```diff
@@ -324,33 +327,33 @@ func (b *databaseBackend) CloseIfShutdown(db *dbPluginInstance, err error) {
...
+	b.cancelQueue()
 
-	b.Lock()
-	defer b.Unlock()
-
-	for _, db := range b.connections {
-		db.Close()
+	connections := b.connClear()
+	for _, db := range connections {
+		go db.Close()
```
swenson, line 355: couldn't help but slip this in :)

builtin/logical/database/path_config_connection.go
```diff
@@ -344,16 +344,14 @@ func (b *databaseBackend) connectionWriteHandler() framework.OperationFunc {
...
+		oldConn := b.connPut(name, &dbPluginInstance{
 			database: dbw,
 			name:     name,
 			id:       id,
+		})
+		if oldConn != nil {
+			oldConn.Close()
 		}
 
 		err = storeConfig(ctx, req.Storage, name, config)
```
ncabatoff, line 357: Are we breaking any assumptions anywhere by calling storeConfig without a lock now?

ncabatoff (commented):
I have a couple of misgivings but overall I think this is a lot clearer than what it replaces.

builtin/logical/database/backend.go
```diff
@@ -324,33 +327,33 @@ func (b *databaseBackend) CloseIfShutdown(db *dbPluginInstance, err error) {
...
-			defer b.Unlock()
 			db.Close()
 
 			// Ensure we are deleting the correct connection
-			mapDB, ok := b.connections[db.name]
-			if ok && db.id == mapDB.id {
-				delete(b.connections, db.name)
+			mapDB := b.connPop(db.name)
+			if mapDB != nil && db.id != mapDB.id {
+				// oops, put it back
```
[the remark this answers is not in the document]
swenson: I can do that. I didn't love this either, but I was trying to preserve the original intent.

builtin/logical/database/path_config_connection.go
swenson: As far as I can tell, no. I don't think there is anything unsafe about multiple calls to `storeConfig`, other than only one of them would win (which is also true even with a lock).

builtin/logical/database/backend.go
```diff
@@ -110,28 +109,72 @@ func Backend(conf *logical.BackendConfig) *databaseBackend {
...
 	*framework.Backend
-	sync.RWMutex
-	// CredRotationQueue is an in-memory priority queue used to track Static Roles
+	// credRotationQueue is an in-memory priority queue used to track Static Roles
 	// that require periodic rotation. Backends will have a PriorityQueue
 	// initialized on setup, but only backends that are mounted by a primary
 	// server or mounted as a local mount will perform the rotations.
 	//
 	// cancelQueue is used to remove the priority queue and terminate the
 	// background ticker.
```
calvn, line 127: Nit: Can we move this docstring closer to cancelQueue?

builtin/logical/database/backend.go
```diff
@@ -58,15 +58,16 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
 	// Create a context with a cancel method for processing any WAL entries and
 	// populating the queue
 	initCtx := context.Background()
-	ictx, cancel := context.WithCancel(initCtx)
-	b.cancelQueue = cancel
+	b.ctx, b.cancelQueue = context.WithCancel(initCtx)
```
calvn, line 61: Since we're always setting this within `Backend`, do we have to set it in `Factory` at all?

builtin/logical/database/backend.go
```diff
@@ -110,28 +111,72 @@ func Backend(conf *logical.BackendConfig) *databaseBackend {
...
 	// that require periodic rotation. Backends will have a PriorityQueue
 	// initialized on setup, but only backends that are mounted by a primary
 	// server or mounted as a local mount will perform the rotations.
 	//
 	// cancelQueue is used to remove the priority queue and terminate the
 	// background ticker.
 	credRotationQueue *queue.PriorityQueue
-	cancelQueue       context.CancelFunc
+	// context used for canceling operations
+	ctx         context.Context
```
calvn, line 130: Nit: Can we call this `ctxQueue` since this is used for that primarily?

builtin/logical/database/path_rotate_credentials.go
```diff
@@ -78,22 +78,30 @@ func (b *databaseBackend) pathRotateRootCredentialsUpdate() framework.OperationF
...
 			}
 			// Even on error, still remove the connection
-			delete(b.connections, name)
+			// be careful with a deadlock here
+			if !unlocked {
+				// set to true before unlocking to avoid race condition
+				unlocked = true
+				dbi.Unlock()
+			}
+			b.ClearConnection(name)
```
calvn, line 104: I think that this can be racy if another request gets to grab the write lock between this unlock and clear connection, potentially clearing out a newly set connection.

builtin/logical/database/backend.go
swenson: It is still necessary if we want to tie the backend context to the passed in context, but I think I can refactor it a little to simplify it.

builtin/logical/database/backend.go
swenson: 👍

builtin/logical/database/path_rotate_credentials.go
```diff
@@ -78,22 +78,30 @@ func (b *databaseBackend) pathRotateRootCredentialsUpdate() framework.OperationF
...
 		dbi.Lock()
-		defer dbi.Unlock()
-
+		unlocked := false
+		defer func() {
+			if !unlocked {
+				// set to true before unlocking to avoid race condition
+				unlocked = true
+				dbi.Unlock()
+			}
```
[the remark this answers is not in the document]
swenson: It is, because the other `defer` could panic, and we want to ensure we *always* unlock.

builtin/logical/database/path_rotate_credentials.go
swenson: Good call. I'll add a method to only clear the connection if the ID matches.

builtin/logical/database/backend.go
```diff
@@ -103,35 +98,76 @@ func Backend(conf *logical.BackendConfig) *databaseBackend {
...
 	// that require periodic rotation. Backends will have a PriorityQueue
 	// initialized on setup, but only backends that are mounted by a primary
 	// server or mounted as a local mount will perform the rotations.
-	//
-	// cancelQueue is used to remove the priority queue and terminate the
-	// background ticker.
 	credRotationQueue *queue.PriorityQueue
-	cancelQueue       context.CancelFunc
+	// queueCtx is the context for the priority queue
+	queueCtx context.Context
```
tomhjp, line 120: nit: If you generate the queue context in the same function where it's consumed, then you only really need to store the `Done()` channel (and the cancel func) on the struct. That will help stop this context getting abused for other uses by accident in future too.

builtin/logical/database/backend.go
```diff
@@ -280,38 +302,34 @@ func (b *databaseBackend) GetConnectionWithConfig(ctx context.Context, name stri
...
-// invalidateQueue cancels any background queue loading and destroys the queue.
-func (b *databaseBackend) invalidateQueue() {
-	// cancel context before grabbing lock to start closing any open connections
-	// this is safe to do without the lock since it is only written to once in initialization
-	// and can be canceled multiple times safely
-	if b.cancelQueue != nil {
-		b.cancelQueue()
+	oldConn := b.connPut(name, dbi)
+	if oldConn != nil {
+		err := oldConn.Close()
```
tomhjp, line 307: Does it matter if two threads are racing and both call close on the `oldConn`?

builtin/logical/database/path_rotate_credentials.go
```diff
@@ -78,22 +78,27 @@ func (b *databaseBackend) pathRotateRootCredentialsUpdate() framework.OperationF
...
+		defer unlock()
 		defer func() {
 			// Close the plugin
 			dbi.closed = true
 			if err := dbi.database.Close(); err != nil {
 				b.Logger().Error("error closing the database plugin connection", "err", err)
 			}
 			// Even on error, still remove the connection
-			delete(b.connections, name)
+			// be careful with a deadlock here
```
tomhjp, line 99: As a first time reader of this code, the deadlock wasn't immediately obvious to me. Could we spell it out a bit more?

swenson: :+1:

builtin/logical/database/backend.go
swenson: That should be safe -- they are both protected by the lock in the `oldConn`.

builtin/logical/database/path_rotate_credentials.go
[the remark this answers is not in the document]
calvn: The initial defer call (i.e. `dbi.Unlock()`) will get invoked since it's already pushed to the stack, so even if it panics during the second defer (e.g. by the `Close` call) it should safely `dbi.Unlock()` without us having to call it explicitly within that second defer func on L99.

Example: https://go.dev/play/p/-V8L4XnrFHD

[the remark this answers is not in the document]
swenson: We need to unlock before we call `b.ClearConnectionId()`, but we can't guarantee that `dbi.database.Close()` won't panic, and we also can't guarantee that the other `defer` runs first.

I suppose we could combine them and move the `unlock()` call to the top of this `defer`, and just remove the other `defer` entirely?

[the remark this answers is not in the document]
swenson: We chatted about this and sorted it out into simpler `defer`s.

tomhjp (approved):
LGTM

builtin/logical/database/backend.go
```diff
@@ -103,35 +98,76 @@ func Backend(conf *logical.BackendConfig) *databaseBackend {
...
+	defer b.connLock.Unlock()
+	dbi := b.connections[name]
+	delete(b.connections, name)
+	return dbi
+}
+
+func (b *databaseBackend) connPopIfEqual(name, id string) *dbPluginInstance {
+	b.connLock.Lock()
+	defer b.connLock.Unlock()
+	dbi := b.connections[name]
```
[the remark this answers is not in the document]
swenson: Good call.

swenson:
Thanks!

PR #15958, opened 2022-06-13
docs: update double use of "note" in client faq

kylemac:
stumbled across an info box that had "note:" written twice


cipherboy (approved):

cipherboy:
Thank you!

PR #15964, opened 2022-06-13
UI/Update CircleCi Config for UI Tests

hellobontempo:


Monkeychip (approved):
Thanks for tackling!

hellobontempo:
Hiya @samsalisbury! I requested your review just as it's my first time editing the `.circleci` directory and I was unsure about this step from the README. How do I just merge the source files without running `make ci-config` and also merging the generated `config.yml` file? Also I realize you updated the README quite a while ago...no worries if this is no longer relevant to you :) 

```
When merging this `.circleci` directory:

- Do not merge the generated `./config.yml` file, instead:
- Merge the source files under `./config/`, and then
- Run `make ci-config` to re-generate the merged `./config.yml`
```

.circleci/config/jobs/test-ui.yml
```diff
@@ -9,19 +9,8 @@ steps:
...
-          | apt-key add -
-        echo "deb http://dl.google.com/linux/chrome/deb/ stable main" \
-          | tee /etc/apt/sources.list.d/google-chrome.list
-        apt-get update
-        apt-get -y install google-chrome-stable
-        rm /etc/apt/sources.list.d/google-chrome.list
-        rm -rf /var/lib/apt/lists/* /var/cache/apt/*
-
         # Add ./bin to the PATH so vault binary can be run by Ember tests
+        set -x
```
samsalisbury, line 13: In general I avoid using `set -x` in CI, because it has a tendency to leak secrets into the logs (if not immediately, then later on when someone makes further changes). It's fine for a quick debugging session as long as you're careful, but I tend to avoid it completely and just add manual logging, e.g. `echo "SOMEVAR=$SOMEVAR"` or `echo "this conditional was hit!" or whatever, when I'm debugging.

samsalisbury (approved):

.circleci/config/jobs/test-ui.yml
hellobontempo: I carried this over from our previous testing set up, because it failed and didn't seem to set the path properly on the following line `export PATH="${PWD}/bin:${PATH}"` and I wasn't sure why. 

I read about avoiding use, so I'd prefer to remove it if there's a better way!

PR #15965, opened 2022-06-13
docs/vault-k8s: update the service annotation

tvoran:
The injector's `service` annotation is really the vault address to
use, and not just the name of the service.

Also change a couple mentions of "controller" to "injector".

Related to https://github.com/hashicorp/vault-k8s/issues/299


swenson (approved):
LGTM

PR #15968, opened 2022-06-14
Docs/ekm sql provider corrections

aphorise:
See below screenshots for 1x correction (wrong role reference) and 2x Addition (one specification & another warning).

Screenshot 2022-06-14 at 10 28 54
Screenshot 2022-06-14 at 10 27 20
Screenshot 2022-06-14 at 10 52 11


aphorise:
'successfully' in the screen shot changed to 'successful'

website/content/api-docs/auth/aws.mdx
```diff
@@ -6,10 +6,6 @@ description: This is the API documentation for the Vault AWS auth method.
 
 # AWS Auth Method (API)
 
-@include 'x509-sha1-deprecation.mdx'
-
-@include 'aws-sha1-deprecation.mdx'
-
```
tomhjp, line 12: Looks like all the changes in this file are unintentional?

website/content/docs/platform/mssql/installation.mdx
```diff
@@ -12,6 +12,7 @@ For upgrade instructions, see [upgrading](/docs/platform/mssql/upgrading).
 ## Prerequisites
 
 * Vault Enterprise server 1.9+ with a license for the Advanced Data Protection Key Management module
+* Microsoft Windows Server editions for installing Vault EKM provider.
```
tomhjp, line 15:
```suggestion
* Microsoft Windows Server operating system
```

website/content/docs/platform/mssql/installation.mdx
```diff
@@ -158,6 +159,15 @@ installation.
...
     ```
 
+    !> **Warning:** Encountering the SQL error response `Provider error code: 2050`
+    at this stage is typically caused by an incorrect license that's lacking
+    "Key Management Transparent Data Encryption". Using the Vault Audit Logs confirm
+    that EKM is in network reach to Vault API path `/sys/license/status`. Failing 
+    attempts by EKM to connect to Vault in order to check its license or the use 
+    of the wrong license are likely causes for the `2050` error response. This 
+    would need to be recitified for the above SQL to be successful before 
+    commencing any farther.
```
tomhjp, line 169: I don't think a danger alert is called for here. I'd also prefer to refer to the troubleshooting page so we can go into more detail there on the possible causes without interrupting the flow too much.

```suggestion
    -> **Note:** This is the first step at which the EKM provider will communicate with Vault. If
    Vault is misconfigured, this step is likely to fail. See
    [troubleshooting](/docs/platform/mssql/troubleshooting) for tips on specific error codes.
```

For the troubleshooting page, I think we should add a new section titled "Error Codes", and just put 2050 as the only entry for now, but I'll add more error codes at a later date.

```
## Error Codes

During installation, the EKM provider registers a manifest of coded event logs to aid debugging. You may see the following error codes during operation.

### 2050 License Error

The EKM provider was unable to verify that Vault has the correct license features. This
could be due to:

- An incompatible Vault Enterprise license - see the installation [prerequisites](/docs/platform/mssql/installation#prerequisites) for the required license feature.
- Lack of network connectivity - Check Vault's audit logs to see if any requests are made to
   authenticate via AppRole or query the `/sys/license/status` API.
- Misconfigured AppRole auth - Ensure you provided the correct Role ID and Secret ID when
   configuring the SQL Server `CREDENTIAL`. See the
   [installation instructions](/docs/platform/mssql/installation) for an end-to-end working example.
- Incorrect policy permissions - The EKM provider requires the `read` capability
   on the path `sys/license/status`. See the `tde-policy` created in the
   [installation instructions](/docs/platform/mssql/installation#configuring-vault)
   for an example of a working policy.
```

tomhjp: Please could we add the Error Codes section to the trouble shooting page as well? You're welcome to copy my suggestion wholesale or make edits as you like, but I just want to make sure we're linking to a page with something helpful on it for this specific callout.

aphorise: Hey @tomhjp  I've added your text verbatim - sorry for the delay.
Screenshot 2022-07-01 at 09 35 42

aphorise: Hey @tomhjp I've added your text verbatim - sorry for the delay.
Screenshot 2022-07-01 at 09 40 51

tomhjp (approved):
Thanks! LGTM 👍 I'll try to help get this merged ASAP

PR #15971, opened 2022-06-14
KMSE distribute key bug

zofskeez:
When distributing a key to a provider via the _Distribute key_ action in the provider toolbar the key was not appearing in the _Keys_ tab of the provider view until after a refresh. This fixes that and adds a test for the workflow.

### DIFF
--- a/ui/tests/acceptance/enterprise-kmse-test.js
+++ b/ui/tests/acceptance/enterprise-kmse-test.js
@@ -6,7 +6,7 @@ import logout from 'vault/tests/pages/logout';
 import mountSecrets from 'vault/tests/pages/settings/mount-secret-backend';
 import { setupMirage } from 'ember-cli-mirage/test-support';
 
-module('Acceptance | Enterprise | kmse', function (hooks) {
+module('Acceptance | Enterprise | keymgmt', function (hooks) {
   setupApplicationTest(hooks);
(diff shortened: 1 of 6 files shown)


ui/tests/acceptance/enterprise-kmse-test.js
```diff
@@ -0,0 +1,45 @@
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import { click, fillIn } from '@ember/test-helpers';
+import authPage from 'vault/tests/pages/auth';
+import logout from 'vault/tests/pages/logout';
+import mountSecrets from 'vault/tests/pages/settings/mount-secret-backend';
+import { setupMirage } from 'ember-cli-mirage/test-support';
+
+module('Acceptance | Enterprise | kmse', function (hooks) {
```
hashishaw, line 9: Thanks for adding this acceptance test! One nit: our other tests related to this secrets engine call it `keymgmt`, can we keep the naming consistent with this one?

hashishaw (approved):
Thanks for tackling this!

ui/tests/acceptance/enterprise-kmse-test.js
zofskeez: Sure thing I updated it.

PR #15975, opened 2022-06-14
KMSE key delete fails silently

zofskeez:
This was noticed when attempting to delete a KMSE key that was distributed to a provider. The delete action for secret backend list items was not doing anything with the error thrown from the Ember Data adapter. This PR is to update that action to display the error in a flash message.


Monkeychip (approved):
Is there an easy way to make this an alertBanner message instead of a flashMessage? This way the error stays on screen and can be copied or screenshot'ed by the user who may want to return to it.  If that's not the current pattern, ignore me.

PR #15979, opened 2022-06-14
KMSE provider list menu fix

arnav28:
- Backend value had to be mapped to the payload so capabilities call
can be triggered. Based on the response from capabilities, options are
rendered in the more menu dropdown.

Screen Shot 2022-06-14 at 10 33 14 AM


hashishaw (approved):
Nice work, thank you!

PR #15980, opened 2022-06-14
PKI cleanup continued: queryParams and moving routes/models/serializers to pki folder

Monkeychip:
Moved as many of the models as I could into the pki folder and cleaned up routing to be based on tabs. I suspect there's an Ember data/memory loss issue in the other models and I'm going to next try and glimmerize the relevant ones before moving them. (Hitting maxium call stack issues when moving the remaining models).


ui/app/helpers/options-for-backend.js
```diff
@@ -34,13 +34,13 @@ const SECRET_BACKENDS = {
         editComponent: 'pki/role-pki-edit',
       },
       {
-        name: 'certs',
+        name: 'cert',
```
Monkeychip, line 37: to match tab and route name together.

ui/app/routes/vault/cluster/secrets/backend/list.js
```diff
@@ -80,7 +80,7 @@ export default Route.extend({
       ssh: 'role-ssh',
       transform: this.modelTypeForTransform(tab),
       aws: 'role-aws',
-      pki: tab === 'certs' ? 'pki/pki-certificate' : 'pki/pki-role',
+      pki: `pki/${tab || 'pki-role'}`,
```
Monkeychip, line 83: this ultimately is why I have this PR. It gets me setup for a much easier time to add new queryParams that are associated by tab name and all the files can be kept in the pki folder. I have three additional tabs I need to add in the PKI project. I am following the keymgmt pattern, which I really like.

ui/app/routes/vault/cluster/settings/configure-secret-backend/section.js
```diff
@@ -2,21 +2,28 @@ import AdapterError from '@ember-data/adapter/error';
...
   pki: ['cert', 'urls', 'crl', 'tidy'],
 };
 export default Route.extend({
   fetchModel() {
     const { section_name: sectionName } = this.paramsFor(this.routeName);
     const backendModel = this.modelFor('vault.cluster.settings.configure-secret-backend');
-    const modelType = `${backendModel.get('type')}-config`;
+    const type = backendModel.get('type');
+    let modelType;
+    if (type === 'pki') {
```
Monkeychip, line 15: not ideal, but this route was not setup to handle folders. It assumed all models would be in the root models directory, which IMO is a little unmanageable or at least hard to view if you don't implement a sub-directory structure e.g. models/pki/model-name.

zofskeez: I think we should agree upon a directory structure going forward or follow the recommended ember conventions to avoid having to handle exceptions like this when dynamically loading things.

zofskeez (approved):
By changing the model and tab name to match the route is that necessary for the view to function or just for consistency? I'm thinking of someone coming along in the future and changing one of those for whatever reason and not having that implied knowledge may cause things to break.

In the secret-edit route specifically there are conditions for model type and capabilities and they seem to keep growing so not necessarily in the scope of this work but I wonder if we should think of a way to streamline resolving those dynamic values?

Monkeychip:
> By changing the model and tab name to match the route is that necessary for the view to function or just for consistency? I'm thinking of someone coming along in the future and changing one of those for whatever reason and not having that implied knowledge may cause things to break.
> 
> In the secret-edit route specifically there are conditions for model type and capabilities and they seem to keep growing so not necessarily in the scope of this work but I wonder if we should think of a way to streamline resolving those dynamic values?

Yeah, for this engine specifically the model, route and tab name all needed to match. Let's chat a bit about some ideas for approaching this problem. It's out of scope, but I might be able to do something here to set a pattern in the future. I have some freedom to go against the grain in this project and set us up for something better in the future.

PR #15986, opened 2022-06-14
agent: add disable_idle_connections configurable

jasonodonnell:
A [commit to go-retryablehttp](https://github.com/hashicorp/go-retryablehttp/commit/171d0c90ea3fe19417437c478bf0aedc589ea919) changed the default behavior of closing every successful request to instead keeping the connection open. This was done to allow users of the package to utilize TCP keep alives. This resulted in a large number of connections being left open for 90 seconds by Vault Agent 1.6.0+.

This PR adds a new configurable, `disable_idle_connections`, which takes a string list to disable idle connections for various features in Vault Agent (`auto-auth`, `caching` and `templating`).

A small example can be found here:
```hcl
 "auto_auth" = {
  "method" = {
    "config" = {
      "role" = "app"
    }
    "type" = "kubernetes"
  }

  "sink" = {
    "config" = {
      "path" = "/home/vault/.token"
    }
    "type" = "file"
  }
}

"disable_idle_connections" = ["auto-auth", "templating"]
"exit_after_auth" = false
"pid_file" = "/home/vault/.pid"

"template" = {
  "contents" = "{{ with secret \"secret/hashiconf\" }}{{ .Data | toJSONPretty }}{{ end }}"
  "destination" = "/vault/secrets/kv-secret"
  "error_on_missing_key" = true
}

"vault" = {
  "address" = "https://vault.vault.svc.cluster.local:8200"
}
```

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -720,24 +720,6 @@ func (c *Client) SetMaxRetries(retries int) {
 	c.config.MaxRetries = retries
 }
 
-func (c *Client) SetDisableKeepAlives(disable bool) {
-	c.modifyLock.RLock()
-	defer c.modifyLock.RUnlock()
-	c.config.modifyLock.Lock()
-	defer c.config.modifyLock.Unlock()
-
-	c.config.HttpClient.Transport.(*http.Transport).DisableKeepAlives = disable
-}
-
-func (c *Client) DisableKeepAlives() bool {
-	c.modifyLock.RLock()
-	defer c.modifyLock.RUnlock()
-	c.config.modifyLock.Lock()
-	defer c.config.modifyLock.Unlock()
-
-	return c.config.HttpClient.Transport.(*http.Transport).DisableKeepAlives
-}
-
 func (c *Client) SetMaxIdleConnections(idle int) {
 	c.modifyLock.RLock()
 	defer c.modifyLock.RUnlock()
--- a/api/client.go
+++ b/api/client.go
@@ -720,6 +720,24 @@ func (c *Client) SetMaxRetries(retries int) {
 	c.config.MaxRetries = retries
 }
 
+func (c *Client) SetDisableKeepAlives(disable bool) {
+	c.modifyLock.RLock()
+	defer c.modifyLock.RUnlock()
+	c.config.modifyLock.Lock()
+	defer c.config.modifyLock.Unlock()
+
+	c.config.HttpClient.Transport.(*http.Transport).DisableKeepAlives = disable
+}
+
+func (c *Client) DisableKeepAlives() bool {
+	c.modifyLock.RLock()
+	defer c.modifyLock.RUnlock()
+	c.config.modifyLock.Lock()
+	defer c.config.modifyLock.Unlock()
+
+	return c.config.HttpClient.Transport.(*http.Transport).DisableKeepAlives
+}
+
 func (c *Client) MaxRetries() int {
 	c.modifyLock.RLock()
 	defer c.modifyLock.RUnlock()
--- a/api/client.go
+++ b/api/client.go
@@ -738,6 +738,24 @@ func (c *Client) DisableKeepAlives() bool {
 	return c.config.HttpClient.Transport.(*http.Transport).DisableKeepAlives
 }
 
+func (c *Client) SetMaxIdleConnections(idle int) {
+	c.modifyLock.RLock()
+	defer c.modifyLock.RUnlock()
+	c.config.modifyLock.Lock()
+	defer c.config.modifyLock.Unlock()
+
+	c.config.HttpClient.Transport.(*http.Transport).MaxIdleConns = idle
+}
+
+func (c *Client) MaxIdleConnections() int {
+	c.modifyLock.RLock()
+	defer c.modifyLock.RUnlock()
+	c.config.modifyLock.Lock()
+	defer c.config.modifyLock.Unlock()
+
+	return c.config.HttpClient.Transport.(*http.Transport).MaxIdleConns
+}
+
 func (c *Client) MaxRetries() int {
 	c.modifyLock.RLock()
 	defer c.modifyLock.RUnlock()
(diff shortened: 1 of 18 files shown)


command/agent.go
```diff
@@ -368,13 +368,22 @@ func (c *AgentCommand) Run(args []string) int {
...
 		}
 		templateNamespace = client.Headers().Get(consts.NamespaceHeaderName)
+
+		sinkClient, err := client.CloneWithHeaders()
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error cloning client for file sink: %v", err))
+			return 1
+		}
+
+		sinkClient.SetDisableKeepAlives(config.DisableKeepAlivesAutoAuth)
```
swenson, line 378: Do we want the default to be enabling keep-alives, or disabling them? (Since Go favors `false`, here we are enabling them by default I believe.)

jasonodonnell: `config.DisableKeepAlivesAutoAuth` is only true if `auto-auth` is found in the configurable string as seen here: https://github.com/hashicorp/vault/pull/15986/files#diff-f309f8618f6d7bea196a7f4eb10f16a73ecc98da0b4dddfdf39278c35b1067afR276.

swenson: Right. I was more asking, do we want the default behavior to be keep-alives, or the old behavior of disabling them by default (and closing connections)?

command/agent/config/config.go
```diff
@@ -24,12 +24,16 @@ import (
...
-	TemplateConfig *TemplateConfig            `hcl:"template_config"`
-	Templates      []*ctconfig.TemplateConfig `hcl:"templates"`
+	AutoAuth                    *AutoAuth                  `hcl:"auto_auth"`
+	ExitAfterAuth               bool                       `hcl:"exit_after_auth"`
+	Cache                       *Cache                     `hcl:"cache"`
+	Vault                       *Vault                     `hcl:"vault"`
+	TemplateConfig              *TemplateConfig            `hcl:"template_config"`
+	Templates                   []*ctconfig.TemplateConfig `hcl:"templates"`
+	DisableKeepAlives           string                     `hcl:"disable_keep_alives"`
+	DisableKeepAlivesCaching    bool
```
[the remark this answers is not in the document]
jasonodonnell: Done!

command/agent/config/config.go
```diff
@@ -24,12 +24,16 @@ import (
...
-	Vault          *Vault                     `hcl:"vault"`
-	TemplateConfig *TemplateConfig            `hcl:"template_config"`
-	Templates      []*ctconfig.TemplateConfig `hcl:"templates"`
+	AutoAuth                    *AutoAuth                  `hcl:"auto_auth"`
+	ExitAfterAuth               bool                       `hcl:"exit_after_auth"`
+	Cache                       *Cache                     `hcl:"cache"`
+	Vault                       *Vault                     `hcl:"vault"`
+	TemplateConfig              *TemplateConfig            `hcl:"template_config"`
+	Templates                   []*ctconfig.TemplateConfig `hcl:"templates"`
+	DisableKeepAlives           string                     `hcl:"disable_keep_alives"`
```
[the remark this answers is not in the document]
jasonodonnell: I decided to go with a string, actually. The UX for lists wasn't as good as a string and it would introduce the first list parameter in agent. Went for the simpler format.

api/client.go
```diff
@@ -720,6 +720,42 @@ func (c *Client) SetMaxRetries(retries int) {
 	c.config.MaxRetries = retries
 }
 
+func (c *Client) SetDisableKeepAlives(disable bool) {
```
swenson, line 723: I don't think these are used anymore.

command/agent.go
```diff
@@ -368,13 +368,24 @@ func (c *AgentCommand) Run(args []string) int {
...
 		templateNamespace = client.Headers().Get(consts.NamespaceHeaderName)
+
+		sinkClient, err := client.CloneWithHeaders()
+		if err != nil {
+			c.UI.Error(fmt.Sprintf("Error cloning client for file sink: %v", err))
+			return 1
+		}
+
+		if config.DisableIdleConnsAutoAuth {
+			sinkClient.SetMaxIdleConnections(-1)
```
swenson, line 379: Just to clarify: this is sufficient, and we don't need to worry about the keep alives parameter anymore?

command/agent/config/config.go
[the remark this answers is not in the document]
jasonodonnell: Actually, went with the slice.

command/agent/config/config.go
```diff
@@ -260,6 +266,25 @@ func LoadConfig(path string) (*Config, error) {
 		result.Vault.Retry.NumRetries = 0
 	}
 
+	if disableIdleConnsEnv := os.Getenv(DisableIdleConnsEnv); disableIdleConnsEnv != "" {
+		result.DisableIdleConns = disableIdleConnsEnv
+	}
+
+	if result.DisableIdleConns != "" {
+		diableIdleConns := strings.ToLower(result.DisableIdleConns)
+		if strings.Contains(diableIdleConns, "auto-auth") {
```
swenson: this might result in users abusing our lack of checking the syntax of this parameter, i.e., this allows them to give us `"auto-auth caching"` or `auto-auth;caching` or `auto-authcaching`. I worry that this sets a bad precedent for other list parameters.

api/client.go
jasonodonnell: They aren't, I've opted to keep them because they might be useful some day, but I could remove if needed.

command/agent/config/config.go
jasonodonnell: This was switched to a slice, so it shouldn't matter anymore. 😄

api/client.go
jasonodonnell: Removed them anyway.

swenson (approved):
LGTM

calvn (approved):

PR #15995, opened 2022-06-15
UI/Remove browserstack

hellobontempo:
Since [removing IE11 support](https://www.vaultproject.io/docs/browser-support) we no longer need browserstack tests


ui/README.md
```diff
@@ -1,15 +1,15 @@
+**Table of Contents**
```
hellobontempo, line 1: Gets deleted by running `yarn doctoc README.md` unless moved to outside the comment block

hellobontempo:
Wrong Jira ticket used in branch name!

Issue #15596, opened 2022-05-25
PKI: Some CN values are always rejected, even with allow_any_name and pki/sign-verbatim


addreas:
Ran into an issue where I want to generate certificates with multiple `@` signs, which currently seems to be impossible even with `allow_any_name` and `pki/sign-verbatim`. 

I'm hitting [this error](https://github.com/hashicorp/vault/blob/70551bf26c356a25a41945b3195e246237bb1ff0/builtin/logical/pki/cert_util.go#L1055) and the culprit seems to be that the validation [here](https://github.com/hashicorp/vault/blob/70551bf26c356a25a41945b3195e246237bb1ff0/builtin/logical/pki/cert_util.go#L272) is being a bit too harsh?

cipherboy:
Hi @addreas -- we discussed this with the cryptosec team. 

I think the validation could be loosened there if we set all of `allow_any_name=true`, `enforce_hostnames=false`, and `allow_wildcard_certificates=true` in the role. Given that the rest of the checks would otherwise be ignored if all three of those flags are set as such, I think there's an argument to be made about not erring. 

If you want to open a PR, I'd be happy to review it :-) There some [existing test scenarios](https://github.com/hashicorp/vault/blob/main/builtin/logical/pki/backend_test.go#L4299-L4505) that could be adapted for validating this case works.

addreas:
Based on the docs for `allow_any_name` that is what I would expect as well. The only thing that isn't mentioned there is the email related stuff that I'm hitting and isn't currently configurable. The perhaps more surprising behavior is that of `pki/sign-verbatim`, which the docs say should allow signing anything, when it in fact uses the same validation logic internally which blocks some names.

My interpretation of this is that the intention is that there should be at least one combination of role settings that bypass every single check in [validateNames](https://github.com/hashicorp/vault/blob/a2543b80d0c951869579ab6f21b73f4732cb1d9d/builtin/logical/pki/cert_util.go#L249). Looking at the comments there are some assumptions made that prevent that, though, despite being completely reasonable assumptions.

In the case where a weird CN is required, for whatever (most likely legacy) reason, the problematic cases are these:
- Multiple occurrences of `@` always being blocked as that is not allowed in an email.
- Multiple occurrences of `*` always being blocked because of adherence to RFC 6125.
- A `*` coming after a `.` always being blocked because wildcards are only allowed in the leftmost label.
[...]

PR #15996, opened 2022-06-15
Add explicit cn_validation field to PKI Roles

cipherboy:
This new parameter allows disabling all validations on a common name, enabled by default on `sign-verbatim` and issuer generation options.
    
Presently, the default behavior is to allow either an email address (denoted with an `@` in the name) or a hostname to pass validation. Operators can restrict roles to just a single option (e.g., for email certs, limit CNs to have strictly email addresses and not hostnames).
    
By setting the value to `disabled`, CNs of other formats can be accepted without validating their contents against our minimal correctness checks for email/hostname/wildcard that we typically apply even when broad permissions (`allow_any_name=true`, `enforce_hostnames=false`, and `allow_wildcard_certificates=true`) are granted on the role.

---

Resolves: #15596


stevendpclark (commented):
This looks good to me.

I would like to see a call out in the help texts for the new param that the specified validations are or'd and not and'd together which is obviousish now but if we introduce others in the future might not be.

PR #15999, opened 2022-06-15
Refactor PKI tests for speed

cipherboy:
I've personally introduced a number of slow tests to the PKI backend. :-) Here's an attempt at trying to fix that a bit.

See individual commits for individual test speed up counts on my machine. There's definitely more to do but this is a start. :-)

Before: github.com/hashicorp/vault/builtin/logical/pki	355.430s
After: github.com/hashicorp/vault/builtin/logical/pki	151.182s

Edit: add some more refactors. This gets us down by over half!


stevendpclark (approved):
💯

PR #16001, opened 2022-06-15
Added nonce field to payload for okta sign in.

linda9379:
Added a nonce field for a unique uuid in the payload for okta sign in using the imported uuid library


ui/app/components/auth-form.js
```diff
@@ -276,6 +277,10 @@ export default Component.extend(DEFAULTS, {
       if (this.customPath || backend.id) {
         data.path = this.customPath || backend.id;
       }
+      // add nonce field for okta backend
+      if (backend.id == 'okta') {
+        data.nonce = uuidv4();
```
zofskeez, line 282: Since we have discontinued support for IE 11 I wonder if we need to add the `uuid` library as a dependency or if [crypto.randomUUID()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) would do the trick?

zofskeez (commented):
Nice work! Just one comment regarding the `uuid` dependency.

ui/tests/unit/adapters/cluster-test.js
```diff
@@ -114,11 +114,12 @@ module('Unit | Adapter | cluster', function (hooks) {
       'ldap:userpass options OK'
     );
 
+    data = { password: 'password', username: 'username', nonce: 'uuid' };
```
hashishaw, line 117: 👏 hooray for updating the tests!

hashishaw (approved):
Great work! 🎉

ui/app/components/auth-form.js
linda9379: I removed the `uuid` library and used `crypto.randomUUID()` instead to generate the uuid values. Thank you for the suggestion!

PR #16002, opened 2022-06-15
Use new -mount syntax for all KV subcommands in 1.11 docs

digivava:
I have updated all docs pages dealing with the KV CLI to use the new `-mount` flag syntax, which was created to mitigate confusion caused by the previous syntax for interacting with KV secrets in the CLI.

These docs pages will now match the text produced by calling a vault kv subcommand with the `-help` flag.

## Background
This syntax is optional, allowing users to do `vault kv get -mount=secret foo` instead of the old `vault kv get secret/foo`. While the old way may look prettier, it turned out this syntax caused a lot of confusion for new users, who quickly run into issues caused by the fact that `secret/foo` looks like a path but is not actually the real path to the secret for KV v2 (`secret/data/foo`). The real path is used by policies and the API, so users fresh off the Getting Started guide often don't understand why their policies or CLI commands don't work when they try to use the same "path" for both.

## Caveats
I continued to use the old syntax on the index page for CLI commands in general, since that page is aimed at showcasing concepts that are true across all Vault CLI commands. This mount flag syntax only works for the kv subcommands, so I didn't want to muddy the message of that page. So, on that page I call out the existence of this flag in a KV-specific paragraph down below, but otherwise leave those examples alone.

PR #16003, opened 2022-06-15
Backport of Use new -mount syntax for all KV subcommands in 1.11 docs into release/1.11.x

hc-github-team-secure-vault-core:
## Backport

This PR is auto-generated from #16002 to be assessed for backporting due to the inclusion of the label backport/1.11.x.

> merge conflict error: POST https://api.github.com/repos/hashicorp/vault/merges: 409 Merge conflict []

---

I have updated all docs pages dealing with the KV CLI to use the new `-mount` flag syntax, which was created to mitigate confusion caused by the previous syntax for interacting with KV secrets in the CLI.

These docs pages will now match the text produced by calling a vault kv subcommand with the `-help` flag.

## Background
This syntax is optional, allowing users to do `vault kv get -mount=secret foo` instead of the old `vault kv get secret/foo`. While the old way may look prettier, it turned out this syntax caused a lot of confusion for new users, who quickly run into issues caused by the fact that `secret/foo` looks like a path but is not actually the real path to the secret for KV v2 (`secret/data/foo`). The real path is used by policies and the API, so users fresh off the Getting Started guide often don't understand why their policies or CLI commands don't work when they try to use the same "path" for both.

## Caveats
I continued to use the old syntax on the index page for CLI commands in general, since that page is aimed at showcasing concepts that are true across all Vault CLI commands. This mount flag syntax only works for the kv subcommands, so I didn't want to muddy the message of that page. So, on that page I call out the existence of this flag in a KV-specific paragraph down below, but otherwise leave those examples alone.

---

<details>
<summary> Overview of commits </summary>

  - 531abac2c10b08ae4b4dc65c6e96a983f98ae19a 

</details>


digivava:
Closing as I ended up having to do a manual backport

PR #16008, opened 2022-06-16
Add region reference

rowansmithhc:
The contents here appears to get called when `vault path-help auth/aws/login` is called, and we don't have an example for `region` so it should be added - ref https://www.vaultproject.io/docs/auth/aws#perform-the-login-operation

### DIFF
--- a/builtin/credential/aws/path_login.go
+++ b/builtin/credential/aws/path_login.go
@@ -122,7 +122,7 @@ needs to be supplied along with 'identity' parameter.`,
 			"region": {
 				Type: framework.TypeString,
 				Description: `The AWS region to which the request should be directed to. Defaults to 
-'us-east-1', can be adjusted if necessary for regions such as 'us-gov-west-1','us-gov-east-1'.`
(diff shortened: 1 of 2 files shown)


builtin/credential/aws/path_login.go
```diff
@@ -119,6 +119,11 @@ binary.`,
 				Description: `Base64 encoded SHA256 RSA signature of the instance identity document. This
 needs to be supplied along with 'identity' parameter.`,
 			},
+			"region": {
```
swenson, line 122: This parameter doesn't appear to be used anywhere?

swenson:
After discussions with @rowansmithhc, it looks like this was attempting to document `region` parameter for `vault login -method=aws`; however, `region` is already documented there:

```sh
$ vault auth help aws
...
  region=<string>
      Explicit AWS region to reach out to for authentication request signing. A value
      of "auto" enables auto-detection of region based on the precedence described above.
      Defaults to "us-east-1" if not specified.
...
```

I can see how the relationship between the `vault login` parameters in `builtin/credential/aws/cli.go` are easily confused for the slightly different `vault write auth/aws/login` parameters (where "region" is not needed since it is baked into the credentials).

PR #16018, opened 2022-06-16
ssh: Fix template regex test for defaultExtensions to allow additional text

stevendpclark:
- The regex to identify if our defaultExtensions contains a template was
   a little too greedy, requiring the entire field to be just the regex. Allow
   additional text within the value field to be added


cipherboy (approved):
LGTM, thanks Steve!

PR #16019, opened 2022-06-16
Refactor PKI storage calls to take a shared struct

cipherboy:
This will allow us to refactor the storage functions to take additional
parameters (or backend-inferred values) in the future. In particular, as
we look towards adding a storage cache layer, we'll need to add this to
the backend, which is now accessible from all storage functions.

---

There has admittedly been a little bit of scope creep in order to make things consistent. But I believe at the top-level handler, we'll now fetch the new storage context struct once and pass that down the chain for use.

### DIFF
--- a/builtin/logical/pki/ca_util.go
+++ b/builtin/logical/pki/ca_util.go
@@ -17,7 +17,7 @@ import (
 	"github.com/hashicorp/vault/sdk/logical"
 )
 
-func (b *backend) getGenerationParams(ctx context.Context, storage logical.Storage, data *framework.FieldData) (exported bool, format string, role *roleEntry, errorResp *logical.Response) {
+func getGenerationParams(sc *storageContext, data *framework.FieldData) (exported bool, format string, role *roleEntry, errorResp *logical.Response) {
 	exportedStr := data.Get("exported").(string)
 	switch exportedStr {
 	case "exported":
@@ -37,7 +37,8 @@ func (b *backend) getGenerationParams(ctx context.Context, storage logical.Stora
 			`the "format" path parameter must be "pem", "der", or "pem_bundle"`)
 		return
 	}
-	keyType, keyBits, err := getKeyTypeAndBitsForRole(ctx, b, storage, data)
+
+	keyType, keyBits, err := sc.getKeyTypeAndBitsForRole(data)
 	if err != nil {
 		errorResp = logical.ErrorResponse(err.Error())
 		return
@@ -75,7 +76,10 @@ func (b *backend) getGenerationParams(ctx context.Context, storage logical.Stora
 	return
 }
 
-func generateCABundle(ctx context.Context, b *backend, input *inputBundle, data *certutil.CreationBundle, randomSource io.Reader) (*certutil.ParsedCertBundle, error) {
+func generateCABundle(sc *storageContext, input *inputBundle, data *certutil.CreationBundle, randomSource io.Reader) (*certutil.ParsedCertBundle, error) {
+	ctx := sc.Context
+	b := sc.Backend
+
 	if kmsRequested(input) {
 		keyId, err := getManagedKeyId(input.apiData)
 		if err != nil {
@@ -90,7 +94,7 @@ func generateCABundle(ctx context.Context, b *backend, input *inputBundle, data 
 			return nil, err
 		}
 
-		keyEntry, err := getExistingKeyFromRef(ctx, input.req.Storage, keyRef)
+		keyEntry, err := sc.getExistingKeyFromRef(keyRef)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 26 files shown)


builtin/logical/pki/storage.go
```diff
@@ -265,11 +279,11 @@ func importKey(ctx context.Context, b *backend, s logical.Storage, keyValue stri
 
 	foundExistingKeyWithName := false
 	for _, identifier := range knownKeys {
-		existingKey, err := fetchKeyById(ctx, s, identifier)
+		existingKey, err := sc.fetchKeyById(identifier)
 		if err != nil {
 			return nil, false, err
 		}
-		areEqual, err := comparePublicKey(ctx, b, existingKey, pkForImportingKey)
+		areEqual, err := comparePublicKey(sc.Context, sc.Backend, existingKey, pkForImportingKey)
```
victorr, line 286: Maybe `comparePublicKey()`  could be a `storageContext` method as well?

builtin/logical/pki/cert_util.go
```diff
@@ -670,7 +673,7 @@ func generateCert(ctx context.Context,
 
 		if data.SigningBundle == nil {
 			// Generating a self-signed root certificate
-			entries, err := getURLs(ctx, input.req)
+			entries, err := getURLs(ctx, input.req.Storage)
```
victorr, line 676: It is somewhat unfortunate that both `storageContext` and `inputBundle` have the storage object. Perhaps use `sc.Storage` instead? 

If you are feeling cautious or pedantic, then maybe you should add a guard to functions that take these two objects to verify that their storage object is then same?

builtin/logical/pki/ca_util.go
```diff
@@ -233,12 +240,12 @@ func getKeyTypeAndBitsFromPublicKeyForRole(pubKey crypto.PublicKey) (certutil.Pr
 	return keyType, keyBits, nil
 }
 
-func getExistingKeyFromRef(ctx context.Context, s logical.Storage, keyRef string) (*keyEntry, error) {
-	keyId, err := resolveKeyReference(ctx, s, keyRef)
+func getExistingKeyFromRef(sc *storageContext, keyRef string) (*keyEntry, error) {
```
victorr, line 243: This one looks like it could be a `storageContext` method.

victorr (approved):
Nice refactoring!

stevendpclark (commented):
Overall I really like this refactoring! 

It does feel at times that we have certain functions as a member variable of `storageContext` that I question such as `getGenerationParams`, `getIssuerName`, `getKeyName` as some examples. 

I would argue that the only functions that should be members of `storageContext` should reside within `storage.go`, I could be convinced for the `config_util.go` perhaps as well.

builtin/logical/pki/storage.go
cipherboy: I updated it to take a storageContext parameter, but I followed @stevendpclark's guidance and didn't make this a `sc` member.

cipherboy:
Ok, this has been revised and rebased following discussion above. :-)

builtin/logical/pki/ca_util.go
```diff
@@ -180,15 +187,15 @@ func getKeyTypeAndBitsForRole(ctx context.Context, b *backend, storage logical.S
 			return "", 0, errors.New("unable to determine managed key id" + err.Error())
 		}
 
-		pubKeyManagedKey, err := getManagedKeyPublicKey(ctx, b, keyId)
+		pubKeyManagedKey, err := getManagedKeyPublicKey(sc.Context, sc.Backend, keyId)
```
kitography, line 190: It would make a lot of sense to me if getManagedKeyPublicKey was a function off storageContext (commenting here since managed_key_util isn't changed) though this is a oss change so, maybe not?

builtin/logical/pki/cert_util.go
```diff
@@ -139,7 +139,7 @@ func fetchCAInfoByIssuerId(ctx context.Context, b *backend, req *logical.Request
 		LeafNotAfterBehavior: entry.LeafNotAfterBehavior,
 	}
 
-	entries, err := getURLs(ctx, req)
+	entries, err := getURLs(sc.Context, sc.Storage)
```
kitography, line 142: as a get, this would also make sense to change to a sc call.

builtin/logical/pki/key_util.go
```diff
@@ -9,11 +9,10 @@ import (
 
 	"github.com/hashicorp/vault/sdk/helper/certutil"
 	"github.com/hashicorp/vault/sdk/helper/errutil"
-	"github.com/hashicorp/vault/sdk/logical"
 )
 
-func comparePublicKey(ctx context.Context, b *backend, key *keyEntry, publicKey crypto.PublicKey) (bool, error) {
-	publicKeyForKeyEntry, err := getPublicKey(ctx, b, key)
+func comparePublicKey(sc *storageContext, key *keyEntry, publicKey crypto.PublicKey) (bool, error) {
+	publicKeyForKeyEntry, err := getPublicKey(sc.Context, sc.Backend, key)
```
kitography, line 15: why not change this below as well?

kitography (approved):
This looks really good.  There's a couple of places it might have made more sense to me if you went deeper, but this is cleaner, and makes more sense, with the refactoring done, and if it's sufficient, then no need to do more.

builtin/logical/pki/cert_util.go
cipherboy: I had it as such, then discussion with Steve kinda turned me against it. Maybe with the caching layer it makes sense, I dunno.

stevendpclark: Happy to be considered wrong about this, seems like I might be in the minority opinion on this topic.

cipherboy: For getURLs in particular, if we do AIA info, this is going to change so I'm not inclined to do this now.

cipherboy:
Thanks all! Given we have consensus that this moves in the right direction, I think the remaining nits can be addressed in subsequent PRs. :-)

Issue #4046, opened 2018-02-28
Vault Neo4j database plugin for secret engine


peterclemenko:
**Feature Request:**
A Vault plugin for allowing connection to a Neo4J database would help with using Vault to authenticate with Neo4j.

pocorschi:
I'm guessing there's still no plugin available for neo4j correct @aoighost ?

leandrohmvieira:
Any updates on this?

jamesdhope:
any update please?

hsimon-hashicorp:
Hi folks! I don't have any update for you, I apologize about that. This might be an opportunity to have a community-supported secrets engine / storage backend in the meantime. I'll keep tracking this request as well, and provide feedback to the team. Thanks!

eh-steve:
Hey all, I've opened this PR as an initial implementation: https://github.com/hashicorp/vault/pull/16020

aphorise:
@peterclemenko @leandrohmvieira @pocorschi @jamesdhope - Hey I'm curious if you folks have tried the plugin that's been done by @eh-steve (👍) & if you're okay to close this issue.

hsimon-hashicorp:
Closing due to available workaround and issue age

laocoon2525:
I have implemented a V5 plugin compatible version for Neo4j. The code is still beta:
https://github.com/HomaiLabs/neo4j-vault-database-plugin
I would appropriate any feedback/review.

PR #16020, opened 2022-06-16
Neo4j database plugin

eh-steve:
Addresses https://github.com/hashicorp/vault/issues/4046

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,7 +120,7 @@ jobs:
     environment:
     - CIRCLECI_CLI_VERSION: 0.1.5546
     - GO_TAGS: ''
-    - GO_VERSION: 1.17.11
+    - GO_VERSION: 1.18.3
     - GOFUMPT_VERSION: 0.2.1
     - GOTESTSUM_VERSION: 0.5.2
   test-go-remote-docker:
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -120,7 +120,7 @@ jobs:
     environment:
     - CIRCLECI_CLI_VERSION: 0.1.5546
     - GO_TAGS: ''
-    - GO_VERSION: 1.17.11
+    - GO_VERSION: 1.18.3
```
cipherboy, line 123: Out of curiosity, is there a change in this PR that requires the updated Go version?

Updating Go versions is a little more involved than this (see some past PRs for example) and is usually something we do closer to release time.

swenson (commented):
Why not just put this in a separate repository and build as an external plug-in?

eh-steve:
> Why not just put this in a separate repository and build as an external plug-in?

I can do, I just got the impression from the open issue that it's something that might be wanted built in.

Happy to move it elsewhere and close this PR if you prefer?

swenson:
@eh-steve I think at this point we'd probably just want to have it as a separate plugin, just to avoid plugin sprawl in the main `vault` repo.

Thanks!

.circleci/config.yml
swenson: This would cause the tests to fail due to the SHA-1 deprecation, so even more reason to move to a separate repo. :)

eh-steve:
Moved to https://github.com/vivacitylabs/neo4j-database-plugin

PR #16024, opened 2022-06-16
Small refactor of mount entry manipulation

VioletHynes:
Will be used in the PR for path suffix bassed lease-count quotas also. All Ent tests pass.

This seemed to be the nicest refactor to this section possible (the rest is very small, and error handling is done for the MountEntry), and also seems to be much more in line with the normal way to get API paths.

### DIFF
--- a/vault/logical_system_quotas.go
+++ b/vault/logical_system_quotas.go
@@ -193,15 +193,9 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 				return logical.ErrorResponse("invalid mount path %q", mountPath), nil
 			}
 
-			var newMountPath string
-			if me.Table == mountTableType {
-				newMountPath = me.Path
-			} else {
-				newMountPath = me.Table + "/" + me.Path
-			}
-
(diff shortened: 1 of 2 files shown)


vault/logical_system_quotas.go
```diff
@@ -193,15 +193,9 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
...
-			var newMountPath string
-			if me.Table == mountTableType {
-				newMountPath = me.Path
-			} else {
-				newMountPath = me.Table + "/" + me.Path
-			}
-
-			pathSuffix = strings.TrimSuffix(strings.TrimPrefix(mountPath, newMountPath), "/")
-			mountPath = newMountPath
+			mountAPIPath := me.APIPathNoNamespace()
```
hghaf099, line 196: Would it make sense to add the pathSuffix line to this function as well? I am suggesting it because I am not sure if this function is used for other reasons than finally extracting pathSuffix.

hghaf099 (approved):
LGTM, I just have a suggestion, but I don't have a strong feeling about that.

PR #16025, opened 2022-06-16
Remove the limit on the number of secret and auth engines

remilapeyre:
Vault currently has a limit on the number of secret and auth engines
that can be created. The exact limit depends on the storage backend
used but it is documented at https://www.vaultproject.io/docs/internals/limits
that 14000 is the maximum that can be supported by the Integrated Storage
backend.

This limit also creates a lower limit on the number of namespaces because
each namespace will have multiple secret and auth engines that will count
toward the 14000 global limit.

This is an operational burden because a Vault cluster can reach this limit
while being very low on RAM, CPU and disk usage, and stop accepting engine
creation without previous notice.

When the limit is reached, it should be possible most of the time to
reduce the number of engines by reorganizing things inside the Vault, for
example by merging two KV engines and using subfolders instead of separate
engines.

This can however be a challenge because the organization model was often
chosen years before, and is built in a lot of already deployed applications
that would need to update the path they are using to read their secrets,
all policies would have to change.

It is also not possible at all with some secret engines, for example
when using the PKI secret engine a mounut entry is required for all CA
rootsm and some secret engines like transit so can't easily be merged
without a lot of manual work.

Making this king of changes globally across the Vault also requires the
coordination of multiple business units making it nearly impossible to
coordinate.

This limit is therefore hard to work around and while the telemetry
gives clear metrics regarding the mount tables size, there is not much
that can be done by operators to avoid the impeding doom.

This limit on the total number of engines is linked to the maximum size
of a storage entry in Vault. When writing the list of mount points to
persistent storage, all mount points of a given type are serialized and
must fit in a single storage entry. This explains why a cluster can have
very low resource usage and suddenly stop accepting creating engines as
the next one makes it go over the threshold.

This means that there is as of today two ways to work around this limit:
  - increase the maximum size of the storage entries by updating
    `max_entry_size`. This has two major drawbacks: it only delays the
    problem a bit longer, and it will degrade the performance of Vault as
    bigger storage entries mean bigger network exchanges that will
    negatively impact the Raft heartbeats.
  - changing the storage backend to another one that does not have this
    limit will also work. For example, the `postgres` storage backend
    uses a `BYTEA` column whose limit is much larger than what is
    supported by Raft, changing the storage backend of an existing Vault
    cluster can however be very difficult. This is also not
    recommendable as only the Consul and Vault storage backends are
    officially supported.

On the other hand the current design has two major advantages:
  - atomicity: because the whole table is stored in a single mount entry,
  it is certain that the whole table will be updated at once and no
  corruption can happen,
  - performance: reading/writing a mount table only requires a read/write
  through the cryptographic barrier and is very efficient.

I've been looking at ways to improve the situation, making it possible to
grow the mount table past the limit of a single storage entry while
keeping performance statisfactory and guarding against data corruption.

After trying multiple schemes I settled on a design when the mount table
is split on the physical storage in multiple chunks with a header
containing metadata.

The current format stores the whole table in compressed JSON at
"core/auth" (or "core/mounts", "core/local-auth", "core/local-mounts"):

		┌─────────────────────────────┐
		│core/auth                    │
		│                             │
		│   type:     string          │
		│   entries:  list of objects │
		│                             │
		└─────────────────────────────┘

The new format stores at "core/auth" a header containing only the type,
version of the physical format, and list of chunks containing the mount
entries. The actual mount entries are separately in the chunks:

		┌─────────────────────────────┐
		│ core/auth                   │
		│                             │
		│   type:     string          │
		│   version:  int             │
		│   chunks:   list of uuids   │
		│                             │
		└───────┬─────────────────────┘
				│
				│      ┌────────────────────────┐
				│      │ core/auth/uuid1        │
				│      │                        │
				├─────▶│    chunk of compressed │
				│      │    entries             │
				│      │                        │
				│      └────────────────────────┘
				│
				│      ┌────────────────────────┐
				│      │ core/auth/uuid2        │
				│      │                        │
				└─────▶│    chunk of compressed │
					   │    entries             │
					   │                        │
					   └────────────────────────┘

Each chunk stores up to 512kb of the mount entries so none of the
storage entries goes over the maximum size and the whole table can be
read by first reading the header at "core/auth", then reading each of
the chunks and finally uncompressing the mount entries.

Correctness
-----------

Because the mount table is now stored in multiple storage entries updating
it requires multiple writes to the storage backend and it must never get
corrupted even if an IO occurs when writing one of the chunks.

To achive this the entries are first compressed and split in chunks in
memory, and each chunk is written to storage separately. Once each chunk
has been written the header is updated to points to the new list of chunks.
Once this is done the chunks pointed to previously are deleted.

Taking the previous example with two chunks the sequence of operation
would be:

           ┌────────────────────┐        ┌───────────────────┐         ┌─────────────────────────────┐
           │                    │        │                   │         │                             │
           │   Write chunk to   │        │  Write chunk to   │         │  Update header at core/auth │
           │   core/auth/uuid3  │───────▶│  core/auth/uuid4  │────────▶│  with addresses of the      │──┐
           │                    │        │                   │         │  new chunks                 │  │
           │                    │        │                   │         │                             │  │
           └────────────────────┘        └───────────────────┘         └─────────────────────────────┘  │
                                                                                                        │
        ┌───────────────────────────────────────────────────────────────────────────────────────────────┘
        │
        │  ┌───────────────────────────┐   ┌──────────────────────────┐   ┌──────────────────────────┐
        │  │                           │   │                          │   │                          │
        │  │  List all chunks under    │   │  Remove core/uuid/uuid1  │   │  Remove core/uuid/uuid2  │
        └─▶│  at core/auth/            │──▶│                          │──▶│                          │
           │                           │   │                          │   │                          │
           └───────────────────────────┘   └──────────────────────────┘   └──────────────────────────┘

Any error happening when writing the chunks, before the header is
updated would not cause any issue as the header would not have been
updated.

Any error happening after the header has been written would not cause
corruption either. It would just leave some useless chunks but they will
get removed the next time the table is updated.

Performance
-----------

Writing the mount table needs to be performant. Currently when reading
or writing the mount table it takes only one operation.

The number of operations of the new scheme depends on the number of chunks
required to fit the full list of mount entries.

With n being the number of chunks the total number of operations is:

  - (n + 1) `Get` when reading the table
  - (n + 1) `Put` + 1 List + n `Delete` when writing the table

The time taken to read/write the table grows linearly with the number of
chunks and this should have acceptable performance, especially since the
behavioras of today is to just fail without alternative.

I also implemented a shortcut when the table only contains one chunk
(i.e. all existing tables) so that for such table reading uses 2 Get
and writing only uses 1 Put. This makes the performance in this case on
par with the current process.

Compatibility with Vault < 1.12 and transparent upgrades
--------------------------------------------------------

The migration from the current encoding of the mount tables, and in the
future the migration from one encoding version to the next one must
obviously happen transparently for the user.

Even during a major upgrade where some servers are running Vault 1.11 and
some others are running Vault 1.12, the leadership can switch back and
forth from the old servers and the new ones and the mount table version
must not be an issue.

To achieve this Vault 1.12 will default to using the current representation
unless an operator explicitly chose to upgrade the mount table format
by setting the `VAULT_DEFAULT_MOUNT_TABLE_VERSION` environment variable
to 1.

When `VAULT_DEFAULT_MOUNT_TABLE_VERSION` is set to 1 the mount table
physical representation is upgraded on the next save, for example when
a new auth or secret engine is enabled.

Once the mount table has been upgraded to the v1 representation it will
not be downgraded by setting `VAULT_DEFAULT_MOUNT_TABLE_VERSION` to 0,
this is on purpose to avoid trying to fit a table too large into a single
entry of the Raft backend. If an operator wishes to do this they can do
so manually by using `sys/raw`.

In the future the default version used in the absence of the `
VAULT_DEFAULT_MOUNT_TABLE_VERSION` environment variable could be changed
from 0 to 1.

Possible optimisations
----------------------

I've looked at multiple optimisations on top of this proposal but
finally decided against them for now. Here's a list of things I've tried:

  - Storing the pages using Protobuf instead of JSON, but that gives
  little benefit and adds a lot of code.

  - Splitting the pages logically, where all pages can be decoded
  separately rather than having to load them all and decode them as a
  block. This works and I think is an interesting improvement as it
  makes it possible to not update the whole table when there is a change
  but only the page containing the entry that has been updated. I've not
  implemented it here because it adds a lot of code:
    - to split the pages cleanly,
    - to update only one page when a single entry has been added/removed/updated,
    - to track fragmentation and compact the mount table when needed.
  Because it adds a lot of code for something that is not really needed I
  don't think it is currently worth it. Only users with a large amount
  of churn in the mount tables would benefit from that and, given the
  current limits on the mount tables, I don't think anyone is doing this.

  - Using a tree to physically store the entries of the mount table. This
  would make it possible to improve some operations like deleting all the
  auth and secret engines at once when a namespace is removed, however it
  doesn't seem to me like this would be beneficial as there is a lot to
  when deleting a namespace like revoking all leases, updating the mount
  tables is probably not a bottleneck and worth the extra complexity.

### DIFF
--- a/vault/auth.go
+++ b/vault/auth.go
@@ -552,50 +552,142 @@ func (c *Core) taintCredEntry(ctx context.Context, nsID, path string, updateStor
 	return nil
 }
 
-// loadCredentials is invoked as part of postUnseal to load the auth table
-func (c *Core) loadCredentials(ctx context.Context) error {
-	// Load the existing mount table
-	raw, err := c.barrier.Get(ctx, coreAuthConfigPath)
+// loadTable reads a mount table header and decodes the mount table according to
+// its physical format version.
+func (c *Core) loadTable(ctx context.Context, path string) (*MountTable, bool, error) {
+	header, err := c.barrier.Get(ctx, path)
 	if err != nil {
-		c.logger.Error("failed to read auth table", "error", err)
-		return errLoadAuthFailed
+		c.logger.Error("failed to read mount table header", "error", err)
+		return nil, false, errLoadAuthFailed
 	}
-	rawLocal, err := c.barrier.Get(ctx, coreLocalAuthConfigPath)
-	if err != nil {
-		c.logger.Error("failed to read local auth table", "error", err)
-		return errLoadAuthFailed
+	if header == nil {
+		return nil, false, nil
+	}
+	size := len(header.Value)
+
+	// Decode the header into mount table
+	mountTable := new(MountTable)
+	if err := jsonutil.DecodeJSON(header.Value, mountTable); err != nil {
+		c.logger.Error("failed to decompress or decode the mount table header", "error", err)
+		return nil, false, err
 	}
 
+	c.logger.Debug("decoding mount table", "version", mountTable.Version)
+	switch mountTable.Version {
+	case 0:
+		// There is nothing special to do, version 0 has all the mount entries
+		// in the table header
+	case 1:
+		extraSize, err := c.decodeMountTableV1(ctx, mountTable, path)
+		if err != nil {
+			c.logger.Error(err.Error())
+			return nil, false, err
+		}
+		size += extraSize
+	default:
+		err := fmt.Errorf("unknown mount table version %d for table %q", mountTable.Version, path)
+		c.logger.Error(err.Error())
+		return nil, false, err
+	}
+
+	// Populate the namespace in memory
+	var mountEntries []*MountEntry
+	for _, entry := range mountTable.Entries {
+		if entry.NamespaceID == "" {
+			entry.NamespaceID = namespace.RootNamespaceID
+		}
+		ns, err := NamespaceByID(ctx, entry.NamespaceID, c)
+		if err != nil {
+			return nil, false, err
+		}
+		if ns == nil {
+			c.logger.Error("namespace on mount entry not found", "namespace_id", entry.NamespaceID, "mount_path", entry.Path, "mount_description", entry.Description)
+			continue
+		}
+
+		entry.namespace = ns
+		mountEntries = append(mountEntries, entry)
+	}
+	mountTable.Entries = mountEntries
+
+	if len(mountTable.Entries) > 0 {
+		isLocal := strings.Contains(path, "local")
+		isAuth := strings.Contains(path, "auth")
+		c.tableMetrics(len(mountTable.Entries), isLocal, isAuth, size)
+	}
+
+	expectedTableVersion, err := defaultMountTableVersion()
+	if err != nil {
+		return nil, false, err
+	}
+
+	return mountTable, mountTable.Version < expectedTableVersion, nil
+}
+
+// decodeMountTableV1 decode a table stored in the v1 physical format where the
+// entries have been split in chunks of 512kb and stored separately.
+// It returns both the decoded mount table and the total physical size used by
+// the header and all the chunks.
+func (c *Core) decodeMountTableV1(ctx context.Context, mountTable *MountTable, path string) (int, error) {
+
+	c.logger.Debug("loading table chunks", "chunks", mountTable.Chunks)
+
+	// Read the chunks
+	var compressedEntries []byte
+	for _, chunk := range mountTable.Chunks {
+		c.logger.Debug("loading mount table chunk", "chunk", chunk)
+
+		p, err := c.barrier.Get(ctx, fmt.Sprintf("%s/%s", path, chunk))
+		if err != nil {
+			c.logger.Error("failed to read mount table chunk", "error", err)
+			return 0, err
+		}
+		if p == nil {
+			err := fmt.Errorf("failed to find chunk")
+			c.logger.Error(err.Error(), "chunk", chunk)
+			return 0, err
+		}
+
+		compressedEntries = append(compressedEntries, p.Value...)
+	}
+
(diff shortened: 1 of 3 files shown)


ncabatoff:
Hi Remi,

Thank you for creating this PR, it addresses a serious pain point affecting many of our users.  I love the extremely detailed description you provided, including performance and correctness analyses.  However, at the moment I’m not inclined to merge it.  We’d prefer to maintain and support a general solution that we could re-use for other larger-than-one-storage-entry use cases.  As you point out in your Possible Optimizations section, there’s not that much motivation to do anything other than read-all-entries and write-all-entries for mount tables.  In other use cases the situation is different, and adopting this solution now would mean we would have to deal with possible migration issues when upgrading, and then again later when we want to migrate mount tables to a more general solution for datasets larger than one storage entry.

I know people have been waiting a very long time for us to address this limitation, the workarounds aren’t great, and unfortunately I can’t commit to a specific timeline for when we will release a fix for mount limits.  But we do have a solution in mind and we are working towards it.

digivava:
While there's still more work to do to support users with high numbers of mount and namespaces, I wanted to mention that we did release some helpful configuration in Vault Enterprise 1.17 today: the `max_mount_and_namespace_table_entry_size` field. This allows users to increase the size limit just for KV entries that store mount table and namespace metadata, without letting other storage entries unintentionally balloon in size.

There is future work being planned to continue addressing other struggles in this area.

PR #16028, opened 2022-06-16
OIDC Config Routing

zofskeez:
Adds routing for OIDC configuration and nested crud routes for applications, keys, assignments, providers and scopes. Since controllers are optional they may be added later on an as needed basis.

### DIFF
--- a/ui/app/router.js
+++ b/ui/app/router.js
@@ -108,6 +108,44 @@ Router.map(function () {
           this.route('index', { path: '/' });
           this.route('create');
         });
+        this.route('oidc', function () {
+          this.route('index', { path: '/' });
+          this.route('applications', function () {
+            this.route('index', { path: '/' });
+            this.route('create');
+            this.route('application', { path: '/:id' }, function () {
+              this.route('edit');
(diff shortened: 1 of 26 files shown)


ui/app/router.js
```diff
@@ -108,6 +108,44 @@ Router.map(function () {
           this.route('index', { path: '/' });
           this.route('create');
         });
+        this.route('oidc', function () {
```
Monkeychip, line 111: I see the need for the individual routes, and this is my preferred approach, but are there any concerns with having a very large router file? If we follow this pattern in the future (mind you I don't think we'll add many more secret engines, but we could...).

Monkeychip (approved):
Once the test past and you make the changes we discussed. Thanks for tackling.

ui/app/router.js
zofskeez: I don't think there are any concerns with how big the router gets or at least I haven't heard of any issues with performance. I think the benefits of having clearly defined routes outweigh having a single route and then dynamically loading different resource types.

PR #16032, opened 2022-06-17
UI/Glimmerize modal & confirmation modal component

hellobontempo:
Glimmerizes modal and confirmation modal components and adds tests


ui/lib/core/addon/components/confirmation-modal.hbs
```diff
@@ -1,14 +1,6 @@
-<Modal
-  @title={{this.title}}
-  @onClose={{this.onClose}}
-  @isActive={{this.isActive}}
-  @type={{this.type}}
-  @showCloseButton={{true}}
->
+<Modal @title={{@title}} @onClose={{@onClose}} @isActive={{@isActive}} @type={{this.type}} @showCloseButton={{true}}>
```
Monkeychip, line 1: Is isActive being used anymore. looks like maybe you moved the mut action to the onClose action?

Monkeychip: I see it in the js file, but is it used on the actual hbs component at all?

Monkeychip (approved):
Just one question. Also, if you haven't already don't forget to mark these components as glimmerized on the google sheet.

Thanks for tackling.

ui/lib/core/addon/components/confirmation-modal.hbs
```diff
@@ -21,21 +13,21 @@
...
+      class={{concat "button " this.buttonClass}}
       disabled={{not-eq this.confirmationInput this.confirmText}}
-      onclick={{this.onConfirm}}
-      data-test-confirm-button={{this.testSelector}}
+      {{on "click" @onConfirm}}
+      data-test-confirm-button={{@testSelector}}
     >
       {{this.buttonText}}
     </button>
-    <button type="button" class="button is-secondary" onclick={{action (mut this.isActive) false}} data-test-cancel-button>
```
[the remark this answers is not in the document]
hellobontempo: I carried this over to mimic how the `Modal` works. Since the passed in `onClose` action changes `isActive` to `false`, I thought it'd be better to keep this consistent and handle changing `isActive` in only one place - the function passed in. What do you think?

in `database-connection.hbs`:

```
<ConfirmationModal
  @title="Delete connection?"
  @onClose={{action (mut this.isDeleteModalActive) false}}
  @isActive={{this.isDeleteModalActive}}
  @confirmText={{@model.name}}
  @toConfirmMsg="deleting the connection"
  @onConfirm={{action "delete"}}
  @testSelector="delete"
>
```

ui/lib/core/addon/components/info-table-row.js
```diff
@@ -2,7 +2,6 @@ import { typeOf } from '@ember/utils';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
-// import layout from '../templates/components/info-table-row';
```
hellobontempo, line 5: stumbled across this and figured i could go ahead and remove!

ui/lib/core/addon/components/confirmation-modal.hbs
```diff
@@ -21,21 +13,21 @@
         class="input has-margin-top"
         autocomplete="off"
         spellcheck="false"
-        data-test-confirmation-modal-input={{this.testSelector}}
+        data-test-confirmation-modal-input={{@testSelector}}
```
zofskeez, line 16: Since test selectors are stripped out of production builds I'm not sure that dedicating a component argument to a test selector value is best. What about spreading the attributes on either the `Modal` or `section` element? You could then remove the value for `data-test-confirmation-modal-input` selector and target it by first selecting the top level element:

```js
<ConfirmationModal data-test-confirm-modal="some-value" />
// confirmation-modal.hbs
<Modal ...attributes></Modal>
// in a test
assert.dom('[data-test-confirm-modal="some-value"] [data-test-confirmation-modal-input]').exists();
```

ui/lib/core/addon/components/confirmation-modal.hbs
```diff
@@ -21,21 +13,21 @@
...
       />
     </div>
   </section>
   <footer class="modal-card-foot modal-card-foot-outlined">
     <button
       type="button"
-      class="button {{this.buttonClass}}"
+      class={{concat "button " this.buttonClass}}
```
zofskeez, line 23: This works but could also be written like:
```hbs
class="button {{this.buttonClass}}"
```

zofskeez (commented):
✨ Thanks for updating! ✨ I left a suggestion and a comment regarding test selectors and/or test selector values as component arguments. Might be a good topic for a group discussion.

ui/lib/core/addon/components/confirmation-modal.hbs
hellobontempo: This is a good call. I think in other components I've just passed in an existing attr to help differentiate, I'll update to do something similar here

PR #16040, opened 2022-06-17
Fix flaky test

Monkeychip:
While testing, I don't believe the setTimeout written into the date-base helper is working as intended for testing. So higher up I check for testing and return, essentially ignoring the interval.


hashishaw (approved):
Nice find! Hope it works, looks good so far 🚀

PR #16046, opened 2022-06-17
Add PKI Caching Layer

cipherboy:
This PR adds a PKI caching layer.

We cache the following things: 

 - Key Entries, IDs, and Names
 - Issuer Entries, IDs, and Names
 - Issuer Cert Bundles, Parsed Cert Bundles, and CAInfoBundles

The latter allows us to fast-path the signing process in the optimal case, avoiding all storage reads to load the issuer. Previously we'd have to:

 - Resolve the reference to an ID
 - Read the entry
 - Parse the entry into a bundle
 - Parse the bundle into a parsed bundle
 - Fetch the URLs
 - Build the CAInfoBundle

Now that's all handled as part of two calls (the one for resolution remains and one for fetching the info bundle) that are both cached in a LRU cache for the optimal case.

---

Using my `raftvault` script and a `benchmark-vault` invocation, I get numbers like:

Vault 1.10.4

```
op         count    mean        95th%       99th%       successRatio
pki issue  1131860  2.760386ms  5.614354ms  7.650496ms  100.00%
```

Vault 1.11.0-rc1

```
op         count    mean        95th%       99th%       successRatio
pki issue  962404  3.258479ms  6.803232ms  9.422979ms  100.00%
```

Vault 1.12 Main

```
op         count    mean        95th%       99th%       successRatio
pki issue  997797  3.147056ms  6.429387ms  8.719978ms  100.00%
```

This Branch

```
op         count    mean        95th%       99th%       successRatio
pki issue  1218442  2.547024ms  5.120058ms  6.941694ms  100.00%
```

```
$ source raftvault && go build . && time ./benchmark-vault -workers=32 -duration=100s -vault_addr=$VAULT_ADDR -vault_token=$VAULT_TOKEN -pki_config_json=configs/pki/all-nostore-ec-256.json -pki_setup_delay=20ms -pct_pki_issue=100
```

This gives us a net ~7% improvement over 1.10 (which lacked multiple issuers but did incur parsing overhead) and a ~22-27% over 1.11.0-rc1 (with multiple storage calls and parsing overhead). Numbers may vary.

### DIFF
--- a/builtin/logical/pki/storage_cache_issuers.go
+++ b/builtin/logical/pki/storage_cache_issuers.go
@@ -37,11 +37,17 @@ func InitIssuerStorageCache() *issuerStorageCache {
 	return &ret
 }
 
-func (c *issuerStorageCache) Invalidate() {
+func (c *issuerStorageCache) Invalidate(op func() error) error {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
 	c.invalid = true
+
+	if op != nil {
+		return op()
+	}
+
+	return nil
 }
 
 func (c *issuerStorageCache) reloadOnInvalidation(ctx context.Context, s logical.Storage) error {
--- a/builtin/logical/pki/storage_cache_issuers.go
+++ b/builtin/logical/pki/storage_cache_issuers.go
@@ -18,6 +18,7 @@ type issuerStorageCache struct {
 	entries       *lru.Cache
 	bundles       *lru.Cache
 	parsedBundles *lru.Cache
+	infoBundles   *lru.Cache
 }
 
 func InitIssuerStorageCache() *issuerStorageCache {
@@ -37,6 +38,11 @@ func InitIssuerStorageCache() *issuerStorageCache {
 	}
 
 	ret.parsedBundles, err = lru.New(32)
+	if err != nil {
+		panic(err)
+	}
+
+	ret.infoBundles, err = lru.New(32)
 	if err != nil {
 		panic(err)
 	}
@@ -72,6 +78,7 @@ func (c *issuerStorageCache) reloadOnInvalidation(ctx context.Context, s logical
 	c.entries.Purge()
 	c.bundles.Purge()
 	c.parsedBundles.Purge()
+	c.infoBundles.Purge()
 
 	// List all issuers which exist.
 	strList, err := s.List(ctx, issuerPrefix)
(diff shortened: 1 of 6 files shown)


builtin/logical/pki/storage_cache_keys.go
```diff
@@ -0,0 +1,251 @@
...
+
+type keyStorageCache struct {
+	lock      sync.RWMutex
+	invalid   bool
+	idSet     map[keyID]bool
+	nameIDMap map[string]keyID
+	entries   *lru.Cache
+}
+
+func InitKeyStorageCache() *keyStorageCache {
```
sgmiller, line 20: Could these be private functions?

builtin/logical/pki/storage_cache_issuers.go
```diff
@@ -0,0 +1,385 @@
...
+		needUnlock = false
+
+		if err := c.reloadOnInvalidation(ctx, s); err != nil {
+			return IssuerRefNotFound, err
+		}
+
+		// Now re-read-lock.
+		c.lock.RLock()
+		needUnlock = true
+	}
```
sgmiller, line 201: This pattern repeats in all the getters.  Opportunity for a helper?

sgmiller: Granted it'd need to pass the non-repeated part to the helper for proper defer semantics and without generics would break return values, so meh, maybe not.

cipherboy: Yeah Steve has a fix-up commit he was playing with here: https://github.com/hashicorp/vault/commit/7c327497d2838e28e949020dafaed30c4d2d98e8 

I think there's bigger questions around locks though (there's the global `issuersLock` on the PKI `backend` directly too -- does it make sense to upgrade that to a read/write lock? What about write starvation w.r.t. lots of reads? Is the present double locking an issue? Are there races within this existing code? &c)

Overall I think this approach isn't viable in _this_ commit. (It was also my second time writing this code, the first attempt didn't pass any tests either. :D)

I think we'd probably be better served knowing that this is possible -- if there's issues resulting with slow performance, we _can_ fix them and we'd (even with fast-and-maybe-not-good code) get only a little better perf than <=1.10 (10%-ish). I think an actual design for this caching layer should perhaps be done a little more carefully. 

I'm on the fence about merging the struct changes, I think if we get consensus, we might want to do that anyways, just since there's a few nice refactors in it.

Perhaps I should just mark this one draft?

PR #16048, opened 2022-06-17
Add database plugin metrics around connections

swenson:
This is a replacement for #15923 that takes into account recent lock
cleanup.

I went ahead and added back in the hanging plugin test, which I meant to
add in #15944 but forgot.

I tested this by spinning up a statsd sink in the tests and verifying I
got a stream of metrics:

```
$ nc -u -l 8125 | grep backend
test.swenson-Q9Q0L72D39.secrets.database.backend.connections.count.pgx.5.:1.000000|g
test.swenson-Q9Q0L72D39.secrets.database.backend.connections.count.pgx.5.:0.000000|g
test.swenson-Q9Q0L72D39.secrets.database.backend.connections.count.pgx.5.:1.000000|g
test.swenson-Q9Q0L72D39.secrets.database.backend.connections.count.pgx.5.:0.000000|g
```

### DIFF
--- a/builtin/logical/database/backend.go
+++ b/builtin/logical/database/backend.go
@@ -6,13 +6,14 @@ import (
 	"net/rpc"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/armon/go-metrics"
 	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-secure-stdlib/strutil"
 	"github.com/hashicorp/go-uuid"
+	"github.com/hashicorp/vault/helper/metricsutil"
+	"github.com/hashicorp/vault/internalshared/configutil"
 	v4 "github.com/hashicorp/vault/sdk/database/dbplugin"
 	v5 "github.com/hashicorp/vault/sdk/database/dbplugin/v5"
 	"github.com/hashicorp/vault/sdk/database/helper/dbutil"
@@ -28,29 +29,6 @@ const (
 	databaseStaticRolePath = "static-role/"
 	minRootCredRollbackAge = 1 * time.Minute
 )
-
-// metrics collection
-var (
-	gaugeSync = sync.Mutex{}
-	gauges    = map[string]*int32{}
-	gaugeKey  = []string{"secrets", "database", "backend", "connections", "count"}
-)
-
-func createGauge(name string) {
-	gaugeSync.Lock()
-	defer gaugeSync.Unlock()
-	gauges[name] = new(int32)
-}
-
-func addToGauge(dbType, version string, uuid string, amount int32) {
-	labels := []metrics.Label{{"type", dbType}, {"version", version}, {"backend", uuid}}
-	gaugeName := fmt.Sprintf("%s-%s", dbType, version)
-	if _, ok := gauges[gaugeName]; !ok {
-		createGauge(gaugeName)
-	}
-	val := atomic.AddInt32(gauges[gaugeName], amount)
-	metrics.SetGaugeWithLabels(gaugeKey, float32(val), labels)
-}
 
 type dbPluginInstance struct {
 	sync.RWMutex
@@ -82,6 +60,20 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
 	b.credRotationQueue = queue.New()
 	// Load queue and kickoff new periodic ticker
 	go b.initQueue(b.queueCtx, conf, conf.System.ReplicationState())
+
+	// collect metrics on number of plugin instances
+	pluginInstanceGaugeProcess, err := metricsutil.NewGaugeCollectionProcess(
+		[]string{"secrets", "database", "backend", "pluginInstances", "count"},
+		[]metricsutil.Label{},
+		b.collectPluginInstanceGaugeValues,
+		metrics.Default(),
+		configutil.UsageGaugeDefaultPeriod, // TODO: we seem to have no way to access the actual configured values for these?
+		configutil.MaximumGaugeCardinalityDefault,
+		b.logger)
+	if err != nil {
+		return nil, err
+	}
+	go pluginInstanceGaugeProcess.Run()
 	return b, nil
 }
 
@@ -121,12 +113,32 @@ func Backend(conf *logical.BackendConfig) *databaseBackend {
 		BackendType:       logical.TypeLogical,
 	}
 
-	b.backendUUID = conf.BackendUUID
 	b.logger = conf.Logger
 	b.connections = make(map[string]*dbPluginInstance)
 	b.queueCtx, b.cancelQueueCtx = context.WithCancel(context.Background())
 	b.roleLocks = locksutil.CreateLocks()
 	return &b
+}
+
+func (b *databaseBackend) collectPluginInstanceGaugeValues(context.Context) ([]metricsutil.GaugeLabelValues, error) {
+	b.connLock.RLock()
+	defer b.connLock.RUnlock()
+	counts := map[string]int{}
+	for _, v := range b.connections {
+		dbType, err := v.database.Type()
+		if err != nil {
+			continue
+		}
+		if _, ok := counts[dbType]; !ok {
+			counts[dbType] = 0
+		}
+		counts[dbType] += 1
+	}
+	var gauges []metricsutil.GaugeLabelValues
+	for k, v := range counts {
+		gauges = append(gauges, metricsutil.GaugeLabelValues{Labels: []metricsutil.Label{{Name: "dbType", Value: k}}, Value: float32(v)})
+	}
+	return gauges, nil
 }
 
 type databaseBackend struct {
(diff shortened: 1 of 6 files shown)


builtin/logical/database/backend.go
```diff
@@ -27,6 +29,29 @@ const (
 	minRootCredRollbackAge = 1 * time.Minute
 )
 
+// metrics collection
+var (
+	gaugeSync = sync.Mutex{}
+	gauges    = map[string]*int32{}
+	gaugeKey  = []string{"secrets", "database", "backend", "connections", "count"}
```
calvn, line 36: Would it make sense to have these be part of `databaseBackend` instead of globals? If the keys and/or labels are common across database backend mounts, would we still be able to aggregate them, if that's the concern around having them local to the struct?

builtin/logical/database/backend.go
```diff
@@ -125,6 +151,23 @@ type databaseBackend struct {
 	// concurrent requests are not modifying the same role and possibly causing
 	// issues with the priority queue.
 	roleLocks []*locksutil.LockEntry
+	// backendUUID is a unique identifier for this backend
+	backendUUID string
+}
+
+// addConnectionsCounter is used to keep metrics on how many and what types of databases we have open
+func (b *databaseBackend) addConnectionsCounter(dbw databaseVersionWrapper, amount int) {
```
calvn, line 159: Would it be more accurate to call this something like `updateConnectionsGauge`?

builtin/logical/database/backend.go
swenson: I thought it would make more sense to have them as globals. Let me see how `armon/go-metrics` handles duplication and aggregation though if we have multiple gauges with the same name.

builtin/logical/database/backend.go
swenson: :+1:

builtin/logical/database/backend.go
swenson: Okay, I dug in a little, and `armon/go-metrics` does not do any aggregation for us, so different backends will report metrics with the same label and could trample the others backends.

The BackendUUID can be used to differentiate, but it tends to generate metrics with long, inscrutable names in `statsd`, so I'm debating truncating it or replacing it with an atomic counter.

builtin/logical/database/backend.go
```diff
@@ -27,6 +29,29 @@ const (
...
+	defer gaugeSync.Unlock()
+	gauges[name] = new(int32)
+}
+
+func addToGauge(dbType, version string, uuid string, amount int32) {
+	labels := []metrics.Label{{"type", dbType}, {"version", version}, {"backend", uuid}}
+	gaugeName := fmt.Sprintf("%s-%s", dbType, version)
+	if _, ok := gauges[gaugeName]; !ok {
+		createGauge(gaugeName)
+	}
```
tomhjp, line 50: None of the reads of `gauges` are inside a `gaugeSync` lock, but the writes are. It seems like all of the calls to `addToGauge` and `createGauge` are ultimately covered by the `connLock`, although that's per database mount and this is global. If you end up moving the gauge variables inside the database backend, then I think you could remove all the gauge locking? And if you keep it global, I think you can cover everything by turning `createGauge` into `createOrGetGauge` with an exclusive lock for the whole function body, and return a plain `*int32` value from the map.

builtin/logical/database/backend.go
```diff
@@ -27,6 +29,29 @@ const (
...
+	gaugeKey  = []string{"secrets", "database", "backend", "connections", "count"}
+)
+
+func createGauge(name string) {
+	gaugeSync.Lock()
+	defer gaugeSync.Unlock()
+	gauges[name] = new(int32)
+}
+
+func addToGauge(dbType, version string, uuid string, amount int32) {
```
tomhjp, line 45: For consistency, either:

```suggestion
func addToGauge(dbType, version, uuid string, amount int32) {
```

Or:

```suggestion
func addToGauge(dbType string, version string, uuid string, amount int32) {
```

builtin/logical/database/backend.go
```diff
@@ -27,6 +29,29 @@ const (
...
+	labels := []metrics.Label{{"type", dbType}, {"version", version}, {"backend", uuid}}
```
tomhjp, line 46: `"version"` may cause some confusion when plugin versioning is introduced in the near future. In order of preference, I would suggest either omitting it or giving it a more detailed name. Is using v4 vs v5 of the plugin interface especially valuable to have in the metrics? The other two labels seem much more worthwhile to me.

swenson:
@tomhjp @calvn I reworked it to use the common `metricsutil` gauge tools (though I had to adapt them a little bit). I think this will address most of the concerns.

swenson:
Hmmm. I wonder if this is really necessary since we do have *some* metrics around database initialization -- https://www.vaultproject.io/docs/internals/telemetry#secrets-engines-metrics ?

builtin/logical/database/backend.go
```diff
@@ -57,6 +60,20 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
...
+		[]metricsutil.Label{},
+		b.collectPluginInstanceGaugeValues,
+		metrics.Default(),
+		configutil.UsageGaugeDefaultPeriod, // TODO: we seem to have no way to access the actual configured values for these?
+		configutil.MaximumGaugeCardinalityDefault,
+		b.logger)
+	if err != nil {
+		return nil, err
+	}
+	go pluginInstanceGaugeProcess.Run()
```
calvn, line 76: This goroutine is leaky since we never call `Stop()` to return from it. A backend's lifecycle may end without the vault process ending on state changes like active to standby, going from unsealed to sealed, or on unmount. We may want to call `Stop` within `databaseBackend.clean()`.

helper/metricsutil/wrapped_metrics.go
```diff
@@ -49,6 +49,23 @@ type Metrics interface {
 
 var _ Metrics = &ClusterMetricSink{}
 
+type SinkWrapper struct {
+	metrics.MetricSink
+}
+
+func (s SinkWrapper) AddDurationWithLabels(key []string, d time.Duration, labels []Label) {
```
calvn, line 56: These seem to be unused.

builtin/logical/database/backend.go
```diff
@@ -57,6 +60,20 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
...
 	// Load queue and kickoff new periodic ticker
 	go b.initQueue(b.queueCtx, conf, conf.System.ReplicationState())
+
+	// collect metrics on number of plugin instances
+	pluginInstanceGaugeProcess, err := metricsutil.NewGaugeCollectionProcess(
+		[]string{"secrets", "database", "backend", "pluginInstances", "count"},
+		[]metricsutil.Label{},
+		b.collectPluginInstanceGaugeValues,
+		metrics.Default(),
+		configutil.UsageGaugeDefaultPeriod, // TODO: we seem to have no way to access the actual configured values for these?
```
calvn, line 70: There isn't a way to directly access Vault Core level config values from mounts. We'd probably have to thread it through `conf *logical.BackendConfig`, but I don't think common that we propagate global config values to mounts.

swenson: It's probably fine then. We could always add a backend-specific config for it in the future.

helper/metricsutil/wrapped_metrics.go
swenson: I believe they are necessary to meet the interface requirements for `Metrics`

builtin/logical/database/backend.go
swenson: Good catch. I'll add that to `clean()`

builtin/logical/database/backend_test.go
```diff
@@ -1461,6 +1461,87 @@ func TestBackend_ConnectionURL_redacted(t *testing.T) {
...
+var _ v5.Database = (*hangingPlugin)(nil)
+
+func TestBackend_PluginMain_Hanging(t *testing.T) {
+	if os.Getenv(pluginutil.PluginVaultVersionEnv) == "" {
+		return
+	}
+	v5.Serve(&hangingPlugin{})
+}
+
+func TestBackend_Closes_Cleanly_Even_If_Plugin_Hangs(t *testing.T) {
```
calvn, line 1502: I like what this test is doing, though the name looks a bit long. What do you think of shortening it to something like `TestBackend_AsyncClose` and adding a comment for further clarification?

tomhjp: For my 2c, I don't mind long test names - it self documents the test, means test failure output gives more information, and doesn't really have any overheads because it's never called from code.

helper/metricsutil/wrapped_metrics.go
```diff
@@ -49,6 +49,23 @@ type Metrics interface {
 
 var _ Metrics = &ClusterMetricSink{}
 
+type SinkWrapper struct {
```
tomhjp, line 52:
```suggestion
// SinkWrapper implements `metricsutil.Metrics` using an instance of
// armon/go-metrics `MetricSink` as the underlying implementation.
type SinkWrapper struct {
```

builtin/logical/database/backend_test.go
swenson: :+1:

tomhjp (commented):
LGTM once Calvin's comments are resolved 👍

builtin/logical/database/backend.go
```diff
@@ -57,6 +60,20 @@ func Factory(ctx context.Context, conf *logical.BackendConfig) (logical.Backend,
 	b.credRotationQueue = queue.New()
 	// Load queue and kickoff new periodic ticker
 	go b.initQueue(b.queueCtx, conf, conf.System.ReplicationState())
+
+	// collect metrics on number of plugin instances
+	pluginInstanceGaugeProcess, err := metricsutil.NewGaugeCollectionProcess(
```
tomhjp, line 65: Looks like you're missing an assignment to `b.gaugeCollectionProcess` here?

swenson: Oops! Good catch :)

builtin/logical/database/backend.go
tomhjp: Should we delete or update the TODO note in the meantime?

tomhjp (approved):
LGTM 👍

calvn (approved):

swenson:
Thanks!

PR #16049, opened 2022-06-17
Start of implementation of the plumbing for role resolution logic on auth mounts

VioletHynes:
This PR is meant as a start to the role resolution logic for role based quotas. This PR doesn't add any new functionality (hence no changelog), but acts as a basis for the role resolution required for role based quotas.

At this point, I'm hoping most for feedback relating to approach (and sanity checks) before I continue with the implementation for the other auth methods in implementing their Operation to get the role information. The new logic for rate limit quotas can be considered an example of how it might work in the creation of both types of quota.

Left to do:
- Adding the `ResolveRoleOperation` to all auth methods that support roles
- Adding the `role` field to the Quota structs, and using it as part of retrieving/creating quotas
- Performing the `ResolveRoleOperation` when checking if a quota applies (likely somewhere in the `applyQuota` flow for login requests)

It's also unclear how much of this should be done as part of the 'plumbing' Jira (VAULT-6612) and which should be done as part of the implementation of the quota features - so I'm also looking for feedback on that. Regardless, after review/iteration/this approach is agreed upon, this will be merged before the next bits of work.

After this change, the following requests fail (assuming approle is enabled):
`vault write sys/quotas/rate-limit/path-suffix-rate rate=2 interval="1m" path="/auth/approle/login" role="abc"`
`vault write sys/quotas/rate-limit/path-suffix-rate rate=2 interval="1m" path="/auth/token" role="abc"`
`vault write sys/quotas/rate-limit/path-suffix-rate rate=2 interval="1m" path="" role="abc"`

And this one succeeds (notably, the quota does not actually apply only to the role, as of right now):
`vault write sys/quotas/rate-limit/path-suffix-rate rate=2 interval="1m" path="/auth/approle/" role="abc"`

### DIFF
--- a/builtin/credential/approle/path_login.go
+++ b/builtin/credential/approle/path_login.go
@@ -75,21 +75,9 @@ func (b *backend) pathLoginResolveRole(ctx context.Context, req *logical.Request
 
 	roleName := roleIDIndex.Name
 
-	roleLock := b.roleLock(roleName)
-	roleLock.RLock()
-
-	role, err := b.roleEntry(ctx, req.Storage, roleName)
-	roleLock.RUnlock()
-	if err != nil {
-		return nil, err
-	}
-	if role == nil {
-		return logical.ErrorResponse("invalid role ID"), nil
-	}
-
 	return &logical.Response{
 		Data: map[string]interface{}{
-			"role": role.name,
+			"role": roleName,
 		},
 	}, nil
 }
--- a/builtin/credential/approle/path_login.go
+++ b/builtin/credential/approle/path_login.go
@@ -33,6 +33,9 @@ func pathLogin(b *backend) *framework.Path {
 			logical.AliasLookaheadOperation: &framework.PathOperation{
 				Callback: b.pathLoginUpdateAliasLookahead,
 			},
+			logical.ResolveRoleOperation: &framework.PathOperation{
+				Callback: b.pathLoginResolveRole,
+			},
 		},
 		HelpSynopsis:    pathLoginHelpSys,
 		HelpDescription: pathLoginHelpDesc,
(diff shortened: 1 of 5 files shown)


builtin/credential/approle/path_login.go
```diff
@@ -54,6 +57,31 @@ func (b *backend) pathLoginUpdateAliasLookahead(ctx context.Context, req *logica
...
+	}
+
+	roleName := roleIDIndex.Name
+
+	return &logical.Response{
+		Data: map[string]interface{}{
+			"role": roleName,
+		},
+	}, nil
+}
```
ccapurso, line 83: It might make sense to create a small helper func like `logical.ResolveRoleResponse` that takes the role name and returns

```
&logical.Response{
    Data: map[string]interface{}{
        "role": roleName,
    },
}, nil
```

It's so basic but if we're introducing this to numerous auth methods it will provide consistency. What do you think?

VioletHynes: I like that idea a lot - especially as we'll be expecting all of the responses to be uniform. I'll make this change

vault/logical_system_quotas.go
```diff
@@ -207,6 +212,26 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
 			return logical.ErrorResponse("quota rule with similar properties exists under the name %q", quotaByFactors.QuotaName()), nil
 		}
 
+		role := d.Get("role").(string)
+		// If this is a quota with a role, ensure the backend supports role resolution
+		if role != "" {
+			if pathSuffix != "" {
+				return logical.ErrorResponse("Quotas cannot contain both a path suffix and a role. If a role is provided, path must be a valid auth mount with a concept of roles"), nil
```
ccapurso, line 219: Love the informative error message!

builtin/credential/approle/path_login_test.go
```diff
@@ -301,3 +301,56 @@ func generateRenewRequest(s logical.Storage, auth *logical.Auth) *logical.Reques
...
+	}
+	resp, err = b.HandleRequest(context.Background(), roleSecretIDReq)
+	if err != nil || (resp != nil && resp.IsError()) {
+		t.Fatalf("err:%v resp:%#v", err, resp)
+	}
+	secretID := resp.Data["secret_id"]
+
+	loginData := map[string]interface{}{
+		"role_id":   roleID,
+		"secret_id": secretID,
```
ccapurso, line 336: It doesn't look like the approle `ResolveRoleOperation` handler requires a `secret_id`. Is this necessary?

VioletHynes: It isn't necessary, but it's what would be passed to it in reality, as this is what a valid login request would look like (and we will use the arguments to the login request to resolve the role). I'm inclined to leave it in, as it's meant to work when given the same data as a valid login request.

ccapurso: Makes sense, thanks for the context.

vault/logical_system_quotas.go
```diff
@@ -207,6 +212,26 @@ func (b *SystemBackend) handleRateLimitQuotasUpdate() framework.OperationFunc {
...
+			}
+			authBackend := b.Core.router.MatchingBackend(namespace.ContextWithNamespace(ctx, ns), mountPath)
+			if authBackend == nil || authBackend.Type() != logical.TypeCredential {
+				return logical.ErrorResponse("Mount path '%s' is not a valid auth method and therefore unsuitable for use with role-based quotas", mountPath), nil
+			}
+			// We will always error as we aren't supplying real data, but we're looking for "unsupported operation" in particular
```
ccapurso, line 225: Thanks for the informative comment!

ccapurso (approved):
Looks good!

hghaf099 (approved):
LGTM

PR #16051, opened 2022-06-17
change rotation_period to algorithm

hellobontempo:
Fix default config listed [here](https://www.vaultproject.io/docs/concepts/oidc-provider#keys). The first attr says `rotation_period` but this should read: `algorithm` ([key docs](https://www.vaultproject.io/api-docs/secret/identity/tokens#algorithm))
Screen Shot 2022-06-17 at 1 18 46 PM


zofskeez (approved):
LGTM good catch!

austingebauer:
@hellobontempo - I added the backport label here so this can get published on the live site. Thanks for fixing this!

PR #16055, opened 2022-06-17
fix: crash if datadog telemetry is down

fiftin:
Vault is dependent on the dogstatsd service in our cluster, and crashes in a loop at startup if that service is unavailable.

Relates to https://github.com/hashicorp/vault/issues/8463

### DIFF
--- a/internalshared/configutil/datadog_sink.go
+++ b/internalshared/configutil/datadog_sink.go
@@ -1,6 +1,8 @@
 package configutil
 
 import (
+	"time"
+
 	"github.com/armon/go-metrics"
 	"github.com/armon/go-metrics/datadog"
 	"github.com/mitchellh/cli"
@@ -13,6 +15,7 @@ type DatadogSink struct {
 	propagateHostname bool
 	sink              *datadog.DogStatsdSink
 	logger            cli.Ui
+	tryedAt           *time.Time
 }
 
 func NewDatadogSink(addr string, hostName string, logger cli.Ui) *DatadogSink {
@@ -53,6 +56,11 @@ func (s *DatadogSink) IncrCounter(key []string, val float32) {
 }
 
 func (s *DatadogSink) EmitKey(key []string, val float32) {
+	sink := s.getSink()
+	if sink == nil {
+		return
+	}
+	sink.EmitKey(key, val)
 }
 
 func (s *DatadogSink) AddSample(key []string, val float32) {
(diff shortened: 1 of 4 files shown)


internalshared/configutil/datadog_sink.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func NewDatadogSink(addr string, hostName string, logger cli.Ui) (*DatadogSink, error) {
+	return &DatadogSink{
+		addr:              addr,
+		hostName:          hostName,
+		propagateHostname: false,
+		logger:            logger,
+	}, nil
+}
```
raskchanky, line 25: Since there's no way for this to error, what are your thoughts on removing the returned nil error value here and also removing the error check in `internalshared/configutil/telemetry.go` where this is called?

raskchanky (commented):
This solution seems pretty reasonable to me, but I'm going to request a second opinion, as the issue in question had a lot of feedback on it

internalshared/configutil/datadog_sink.go
fiftin: Fixed

internalshared/configutil/datadog_sink.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (s *DatadogSink) getSink() *datadog.DogStatsdSink {
+	if s.sink != nil {
+		return s.sink
+	}
+
+	sink, err := datadog.NewDogStatsdSink(s.addr, s.hostName)
+	if err != nil {
+		s.logger.Warn("failed to connect to datadog: " + err.Error())
```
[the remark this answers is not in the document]
fiftin: @ncabatoff Agree, it can be noisy. We can log the error one time and log info-message when connection established.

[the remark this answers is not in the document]
fiftin: Or we can log the error with interval.

internalshared/configutil/datadog_sink.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"time"
+
+	"github.com/armon/go-metrics"
+	"github.com/armon/go-metrics/datadog"
+	"github.com/mitchellh/cli"
+)
+
+type DatadogSink struct {
+	tags              []string
```
ncabatoff, line 12: You probably need to add a mutex field for getSink to lock to make concurrent use of that method safe.

internalshared/configutil/datadog_sink.go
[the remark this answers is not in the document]
fiftin: I added interval for attempts to connect.

internalshared/configutil/datadog_sink.go
fiftin: @ncabatoff Done

internalshared/configutil/datadog_sink.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	sink.EmitKey(key, val)
+}
+
+func (s *DatadogSink) AddSample(key []string, val float32) {
+	s.AddSampleWithLabels(key, val, nil)
+}
+
+func (s *DatadogSink) getSink() *datadog.DogStatsdSink {
+	s.lock.Lock()
```
ncabatoff, line 79: I think at this point the change is "correct", in that we could merge it now without causing breakage.  I'm a little concerned about the performance implications here though.  Since a basic mutex can only be locked by a single goroutine, we're now serializing all metrics writers when a datadog sink exists.  Maybe we're better off using an RWMutex.  Then we'd only Rlock in the happy path, escalating to a write lock if we get to the point where we need to actually modify s.attemptedToConnectAt.

fiftin: @ncabatoff Done. 
I refactored the code to make it testable, because the code gets complicated.

fiftin:
Hi @ncabatoff 

Do you have a time to check my PR?

PR #16063, opened 2022-06-20
website: Update replication docs to mention Integrated Storage

lucymhdavies:
Currently only mentions Consul in the initial paragraph


raskchanky (approved):
Thanks @lucymhdavies!

PR #16065, opened 2022-06-20
Create polling function for correct answer for okta number challenge

linda9379:
Implemented polling function to poll for correct answer in Okta number challenge every 1s. We keep polling for an answer until a non-404 response is given. Temporarily printing out the correct answer as the display screens are not implemented yet (will be done after this task).


linda9379:
Pull request created to the wrong branch.

PR #16079, opened 2022-06-21
Add role resolution operations to cert and aws auth types (VAULT-6612)

VioletHynes:
This commit adds role resolution operations to the cert and aws auth types. It's mostly just refactoring the logic that gets the role into its own method, then calling that method from the role resolution function and the login function proper.

I thought it made sense to add this to a couple more auth methods when awaiting review, and now I'll be looking at more common functionality and then moving back to the auth methods again.

After this, I'll need to do a few things. Roughly ordered:
- Adding the role field to the Quota structs, and using it as part of retrieving/creating quotas
Performing the ResolveRoleOperation when checking if a quota applies
- Upversion SDK so I can add this to the plugin auth methods (alicloud etc)
- Add the role resolution functionality to the plugin auth methods (and upversion them)
- Pull in new versions of plugins to Vault

Adding no changelog for the same reason as https://github.com/hashicorp/vault/pull/16049 - no new functionality is added here, will come as part of the commits that actually 'make it work'


builtin/credential/aws/path_login.go
```diff
@@ -128,13 +128,211 @@ needs to be supplied along with 'identity' parameter.`,
...
+
+	roleName := data.Get("role").(string)
+
+	// If roleName is not supplied, a role in the name of the instance's AMI ID will be looked for
+	if roleName == "" {
+		roleName = identityDocParsed.AmiID
+	}
+
+	// Get the entry for the role used by the instance
+	roleEntry, err := b.role(ctx, req.Storage, roleName)
```
hghaf099, line 219: It seems that these lines exist in two places, is this intentional?

hghaf099 (commented):
One minor comment, otherwise looks good!

builtin/credential/aws/path_login.go
hghaf099: To be clear as what I meant by two places, I am referring to `pathLoginUpdateEc2` function. In `pathLoginUpdateEc2`, `pathLoginEc2GetRoleNameAndIdentityDocumentParsed ` is called (which is refactored from the former) and then `roleEntry` is used to check for a CIDR match and other things. 
So, the question is what is getting `roleEntry` used for here? Is it to make sure it exists? If that is the case, would  adding a comment describing the intended use of fetching it be beneficial?

VioletHynes: Ahh yes I see what you mean. Essentially:

roleEntry does not contain the roleName, but we need to check it to ensure the role exists. It would be possible to return the roleEntry _and_ the roleName to avoid the doubling up but that feels a little strange to me. Ideally, the roleEntry would contain the name and we'd only return and handle that in both functions, but that's not possible.

hghaf099: I think adding a comment describing that would be useful!

hghaf099 (approved):

builtin/credential/aws/path_login.go
```diff
@@ -128,13 +128,211 @@ needs to be supplied along with 'identity' parameter.`,
...
+	case anyIam && !allIam:
+		return logical.ErrorResponse("supplied some of the auth values for the iam auth type but not all"), nil
+	case anyIam:
+		return b.pathLoginResolveRoleIam(ctx, req, data)
+	default:
+		return logical.ErrorResponse("didn't supply required authentication values"), nil
+	}
+}
+
+func (b *backend) pathLoginEc2GetRoleNameAndIdentityDocumentParsed(ctx context.Context, req *logical.Request, data *framework.FieldData) (string, *identityDocument, *logical.Response, error) {
```
ccapurso, line 160: This is a nit so feel free to ignore :). The method name here is a bit long and might be hard to grok at quick glance. Maybe something like `pathLoginEc2GetRoleNameAndIdentityDoc` which is slightly shorter is still descriptive enough to get the intent across? What do you think?

ccapurso: Cool, thanks!

builtin/credential/cert/path_login_test.go
```diff
@@ -0,0 +1,43 @@
+package cert
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+func TestAppRole_RoleResolve(t *testing.T) {
```
ccapurso, line 10:
```suggestion
func TestCert_RoleResolve(t *testing.T) {
```

PR #16080, opened 2022-06-21
OIDC/Kubernetes docs: Improve instructions for setting bound_audiences

tomhjp:
Adds instructions for finding the correct `bound_audiences` value to use when using a default token and the Kubernetes cluster has been configured with non-default values for its API audiences. Previously, there was only a small throwaway comment with no help for the non-default scenario, but this is a very common scenario as several managed Kubernetes offerings configure the audience differently from the issuer.

Relates to https://github.com/hashicorp/vault-plugin-auth-jwt/issues/206.


austingebauer (approved):
LGTM. Thanks, @tomhjp!

tomhjp:
Thanks for the offer @BasmehSabeer - feel free to make suggestions or raise another PR if you think it can be improved further, but I'll merge this one for now.

PR #16091, opened 2022-06-21
UI/OIDC models

hellobontempo:
Creates models (attrs + capabilities) for each OIDC config element

Also fixes routing so that the `/oidc` page is the landing page not `oidc/clients` (will reroute/transition to `oidc/clients` if there are any applications have been configured)

Note: Model attrs are intended to be an outline, but will likely need to be adjusted when the forms are created


ui/app/models/oidc/assignment.js
```diff
@@ -0,0 +1,42 @@
+import Model, { attr, hasMany } from '@ember-data/model';
+import lazyCapabilities, { apiPath } from 'vault/macros/lazy-capabilities';
+
+export default class OidcAssignmentModel extends Model {
+  @attr('string') name;
+  @hasMany('identity/entity') identity_entities;
```
hellobontempo, line 6: are these relationships helpful here? or should we remove

ui/app/templates/components/oidc/provider-config-header.hbs
```diff
@@ -0,0 +1,21 @@
+<PageHeader as |p|>
```
hellobontempo, line 1: Goal here is to have this header component display the tabs if it _is_ a route index page:
Screen Shot 2022-06-21 at 11 42 18 AM
<hr>
otherwise show the title with breadcrumb nav, with a title:
Screen Shot 2022-06-21 at 11 43 18 AM

 **or**

ui/app/models/oidc/provider.js
```diff
@@ -0,0 +1,51 @@
...
+    editType: 'searchSelect',
+  })
+  scopesSupported;
+
+  // form has a radio option to allow_all, or limit access to selected 'application'
+  // if limited, expose assignment dropdown and add via search-select
+  @attr('array', {
+    editType: 'searchSelect',
+  })
+  allowedClientIds;
```
zofskeez, line 22: Another candidate for `hasMany` if we need the full model

zofskeez (changes requested):
Nice work and thanks for tackling the models! I left some general comments. The properties on the `assignment` model should be straightened out before we merge.

ui/app/models/oidc/assignment.js
```diff
@@ -0,0 +1,42 @@
...
+  })
+  entityIds;
+
+  @attr('array', {
+    label: 'Groups',
+    editType: 'searchSelect',
+    fallbackComponent: 'string-list',
+    models: ['identity/group'],
+  })
+  groupIds;
```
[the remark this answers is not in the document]
hellobontempo: Ahh - so we don't need to further define it as an `@attr` if you define a relationship? Good to know - Thank you!

ui/app/models/oidc/client.js
```diff
@@ -0,0 +1,136 @@
...
+    return expandAttributeMeta(this, [
+      'name',
+      'clientType',
+      'redirectUris',
+      'key',
+      'idTokenTtl',
+      'accessTokenTtl',
+      'assignments',
+    ]);
+  }
```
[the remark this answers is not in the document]
hellobontempo: good call!

ui/app/models/oidc/client.js
```diff
@@ -0,0 +1,139 @@
...
+    label: 'Redirect URIs',
+    subText:
+      'One of these values must exactly match the redirect_uri parameter value used in each authentication request.',
+    editType: 'stringArray',
+  })
+  redirectUris;
+
+  // >> MORE OPTIONS TOGGLE <<
+
+  // might be a good candidate for using @belongsTo relationship?
```
zofskeez, line 30: 👍

zofskeez (approved):
LGTM!

PR #16095, opened 2022-06-21
dockerfile: use ubi-minimal:8.6

tvoran:
Just updating the base for the UBI images.


cipherboy (approved):
I like newer RHEL versions :-)

PR #16111, opened 2022-06-22
Add endpoints to provide ability to modify logging verbosity

ccapurso:
This PR introduces some endpoints to allow users to modify the logging verbosity of all loggers or an individual logger. One prerequisite of this is to ensure that the base logger that Vault uses is initialized using `IndependentLevels` so that individual loggers can be modified.

The two endpoints that have been introduced are `sys/loggers` (all loggers) and `sys/loggers/:name` (an individual logger, e.g. `core`). Each support a `POST` and `DELETE` to modify and revert back to Vault configuration, respectively.

There were a handful of loggers that were not accounted for in `Core.allLoggers` so I have added calls to `Core.AddLogger` for them. I performed an audit of calls to `logger.Named` to ensure the ones that I did add made sense to track. There are many short-lived loggers which have been deliberately left out.

This PR also includes a new doc page to outline the new endpoints.


vault/core.go
```diff
@@ -2816,6 +2816,19 @@ func (c *Core) SetLogLevel(level log.Level) {
 	}
 }
 
+func (c *Core) SetLogLevelByName(name string, level log.Level) error {
+	c.allLoggersLock.RLock()
+	defer c.allLoggersLock.RUnlock()
+	for _, logger := range c.allLoggers {
+		if logger.Name() == name {
```
ncabatoff, line 2823: Equality, or prefix match (only from dot delimiter boundaries)?

ncabatoff: My thinking is that if I want to turn up logging for "core.identity", I shouldn't have to know that there may be subsystems within identity that create subloggers (e.g. "core.identity.storagepacker") that I probably also want turned up.  I may not even be aware they exist based on the log level I usually use.  Maybe we could take a regexp?

vault/logical_system.go
```diff
@@ -4436,6 +4436,111 @@ func (b *SystemBackend) handleVersionHistoryList(ctx context.Context, req *logic
 	return logical.ListResponseWithInfo(respKeys, respKeyInfo), nil
 }
 
+func getLogLevel(logLevel string) (log.Level, error) {
```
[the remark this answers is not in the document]
ccapurso: I was actually planning on using that but it actually doesn't support everything that we accept in configuration, for example "warning" or "notice". I don't know what the history of supporting those is but it seems prudent to remain consistent between both workflows.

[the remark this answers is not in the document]
ncabatoff: Good catch!

vault/core.go
ccapurso: That's a good question. I think that supporting equality would provide a better UX. One sub logger might be far noisier than the one the user is intending to modify. If we get feedback that folks want the ability to override an entire class of loggers then we could add an optional API param to allow for it.

vault/logical_system.go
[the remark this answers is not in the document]
ncabatoff: Maybe add a comment so someone doesn't come later and make this change unawares?

vault/core.go
ncabatoff: I had that hesitation too, which was why I suggested regexps, that way both options would be available.

ncabatoff: I'm probably trying to make it do too much, let's go with your simpler approach for now.

ccapurso: This actually made me notice something, @ncabatoff. It looks like there are numerous sub-loggers that are not accounted for in `allLoggers` by calling `AddLogger`. I'm going to go through and audit calls to `logger.Named` and call `AddLogger` to ensure that these new API endpoints can actually modify all the logger names that users will see in their logs.

ccapurso: There are some loggers that are short-lived (e.g. for initialization of a backend) that I'll leave out of `allLoggers`.

ncabatoff (approved):

ncabatoff (approved):

PR #16115, opened 2022-06-22
VAULT-6613 Add role support for rate limit quotas (OSS Changes)

VioletHynes:
The subset of OSS changes required for role support for rate limit quotas.

PR best reviewed by the ent counterpart here: https://github.com/hashicorp/vault-enterprise/pull/2996


ccapurso (approved):
As recommended, provided my feedback on ent. Looks good!

PR #16124, opened 2022-06-23
Add signature_bits to sign-intermediate, sign-verbatim

cipherboy:
Both of these two signing requests lacked the `signature_bits` parameter, prohibiting callers from modifying the signature size on RSA key sizes. 

Notably, this parameter is present on root generation and role based signing. Since this parameter might change from request to request, I opted not to pull it out of the role and instead use the API-provided value.


stevendpclark (approved):
lgtm

PR #16125, opened 2022-06-23
UI/RadioCard Adjust Sizes

hellobontempo:
Make the RadioCard component adjust size based on how many cards, and the width of the screen
radio-flex


Monkeychip (approved):
Thank you!

PR #16130, opened 2022-06-23
UI/util for updating a hasMany relationship via SearchSelect

hellobontempo:
In the OIDC config, many of the forms present an opportunity to leverage a `hasMany` relationship between models giving us a tidy way to handle the interrelated attrs.

examples of future use:

- Form to create a client (application) and search select for assignments 
Screen Shot 2022-06-23 at 3 37 03 PM

- Form to create a `key` and search select for clients (applications)

Screen Shot 2022-06-23 at 3 37 50 PM

- Form to create a `provider` and search select for clients and scopes 

Screen Shot 2022-06-23 at 3 39 52 PM

- Form to create an 'assignment' and search select entities and groups

Screen Shot 2022-06-23 at 3 40 43 PM

### DIFF
--- a/ui/lib/core/addon/components/search-select.js
+++ b/ui/lib/core/addon/components/search-select.js
@@ -15,7 +15,7 @@ import layout from '../templates/components/search-select';
  *
  * @param {string} id - The name of the form field
  * @param {Array} models - An array of model types to fetch from the API.
- * @param {function} onChange - The onchange action for this form field.
+ * @param {function} onChange - The onchange action for this form field. ** SEE UTIL ** search-select-has-many.js if selecting models from a hasMany relationship
  * @param {string | Array} inputValue -  A comma-separated string or an array of strings -- array of ids for models.
  * @param {string} label - Label for this form field
  * @param {string} fallbackComponent - name of component to be rendered if the API call 403s
--- a/ui/lib/core/addon/utils/search-select-has-many.js
+++ b/ui/lib/core/addon/utils/search-select-has-many.js
@@ -1,0 +1,33 @@
+/**
+ * Util to add/remove models in a hasMany relationship via the search-select component
+ *
+ * @example of using the util within an action in a component
+ *```js
+ * @action
+ * async onSearchSelectChange(selectedIds) {
+ *    const methods = await this.args.model.mfa_methods;
+ *   handleHasManySelection(selectedIds, methods, this.store, 'mfa-method');
+ *  }
+ *
+ * @param selectedIds array of selected options from search-select component
+ * @param modelCollection array-like, list of models from the hasMany relationship
+ * @param store the store so we can call peekRecord()
+ * @param modelRecord string passed to peekRecord
+ */
+
+export default function handleHasManySelection(selectedIds, modelCollection, store, modelRecord) {
+  // first check for existing models that have been removed from selection
+  modelCollection.forEach((model) => {
+    if (!selectedIds.includes(model.id)) {
+      modelCollection.removeObject(model);
+    }
+  });
+  // now check for selected items that don't exist and add them to the model
+  const modelIds = modelCollection.mapBy('id');
+  selectedIds.forEach((id) => {
+    if (!modelIds.includes(id)) {
+      const model = store.peekRecord(modelRecord, id);
+      modelCollection.addObject(model);
+    }
+  });
+}
--- a/ui/app/components/mfa/mfa-login-enforcement-form.js
+++ b/ui/app/components/mfa/mfa-login-enforcement-form.js
@@ -3,6 +3,7 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { task } from 'ember-concurrency';
+import handleHasManySelection from 'core/utils/search-select-has-many';
 
 /**
  * @module MfaLoginEnforcementForm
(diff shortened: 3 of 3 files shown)


ui/lib/core/addon/utils/search-select-has-many.js
```diff
@@ -0,0 +1,33 @@
+/**
```
hellobontempo, line 1: I prepended the name of this file with `search-select` so the util didn't get lost and would appear while searching for the `search-select` component.

ui/lib/core/addon/utils/search-select-has-many.js
```diff
@@ -0,0 +1,33 @@
...
+ *   handleHasManySelection(selectedIds, methods, this.store, 'mfa-method');
+ *  }
+ *
+ * @param selectedIds array of selected options from search-select component
+ * @param modelCollection array-like, list of models from the hasMany relationship
+ * @param store the store so we can call peekRecord()
+ * @param modelRecord string passed to peekRecord
+ */
+
+export default function handleHasManySelection(selectedIds, modelCollection, store, modelRecord) {
```
hellobontempo, line 18: but named the function differently to more accurately declare what it's doing.

ui/lib/core/addon/utils/search-select-has-many.js
Monkeychip: smart!

ui/lib/core/addon/utils/search-select-has-many.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @example of using the util within an action in a component
+ *```js
+ * @action
+ * async onSearchSelectChange(selectedIds) {
+ *    const methods = await this.args.model.mfa_methods;
+ *   handleHasManySelection(selectedIds, methods, this.store, 'mfa-method');
+ *  }
+ *
+ * @param selectedIds array of selected options from search-select component
```
Monkeychip, line 12: When you say selected options, does that mean an option that was selected, or is this an array of options they can select?

ui/lib/core/addon/utils/search-select-has-many.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param modelCollection array-like, list of models from the hasMany relationship
```
Monkeychip, line 13: Is there any reason you're not using the convention for documentation of the `@param {Array} modelCollection - Array-like, list of ...`

ui/lib/core/addon/utils/search-select-has-many.js
hellobontempo: an array of options they selected - this handled by the search-select component:
 `selectedOptions` is returned via `this.onChange` which is sent to the callback function 

```  handleChange() {
    if (this.selectedOptions.length && typeof this.selectedOptions.firstObject === 'object') {
      if (this.passObject) {
        this.onChange(Array.from(this.selectedOptions, (option) => ({ id: option.id, isNew: !!option.new })));
      } else {
        this.onChange(Array.from(this.selectedOptions, (option) => option.id));
      }
    } else {
      this.onChange(this.selectedOptions);
    }
  },

**Review by Monkeychip — APPROVED** [review]
Just a couple of small questions. Thanks for adding this.

File: ui/lib/core/addon/utils/search-select-has-many.js (hunk 4 above)
Comment (hellobontempo) [reply ui/lib/core/addon/utils/search-select-has-many.js:13]: Since this isn't a component - I just pulled the jsdoc syntax so provide some explanation of what's going on since it's a newer pattern. 

I wasn't sure how folks would feel about it. I can commit to 100% jsdoc syntax if this documentation is helpful! But wanted feedback first before committing.

File: ui/lib/core/addon/utils/search-select-has-many.js (hunk 3 above)
Comment (Monkeychip) [reply ui/lib/core/addon/utils/search-select-has-many.js:12]: Got it. In most cases is selectedOptions one thing? I guess what confuses me, and it's not your language but the original of SearchSelect is why options is plural if this is coming from a search component. I would think you'd be searching, adding just one thing. That's how i've seen the searchSelect component work. Again, not something for you to address, more just the plural is messing with my thinking of what this param represents.

File: ui/lib/core/addon/utils/search-select-has-many.js (hunk 4 above)
Comment (Monkeychip) [reply ui/lib/core/addon/utils/search-select-has-many.js:13]: I reads a little confusing without the of separation from the name to the description, which is why I thought of this. For example: @param store the store so we can call peekRecord(). 

I think my brain is used to documentation style we use for components, so it took a moment to adjust. My preference, which isn't a strong one, is to follow the styling we use for documentation in components.

File: ui/lib/core/addon/utils/search-select-has-many.js (hunk 1 above)
Comment (zofskeez) [reply ui/lib/core/addon/utils/search-select-has-many.js:1]: 💯

File: ui/lib/core/addon/components/search-select.js [hunk 5]
```diff
@@ -15,7 +15,7 @@ import layout from '../templates/components/search-select';
  *
  * @param {string} id - The name of the form field
  * @param {Array} models - An array of model types to fetch from the API.
- * @param {function} onChange - The onchange action for this form field.
+ * @param {function} onChange - The onchange action for this form field. ** SEE UTIL ** search-select-has-many.js if selecting models from a hasMany relationship
```
zofskeez, line 18: 👍

ui/lib/core/addon/utils/search-select-has-many.js
zofskeez: The `SearchSelect` component always returns an array in the `onChange` callback even if you can only select a single item. I think for oidc assignments as an example, since entities and groups are arrays you should be able to select multiple via `SearchSelect`.

zofskeez (approved):
LGTM thanks for creating this!

PR #16131, opened 2022-06-23
UI/OIDC client form

hellobontempo:
Application (client) create view for OIDC provider config
create-app


ui/app/templates/components/oidc/assignment-form.hbs
```diff
@@ -1,4 +1,6 @@
-<div>
+{{! TODO remove div, add unless @isInline conditional for header }}
```
hashishaw, line 1: This comment is complete, yes?

ui/app/templates/components/oidc/client-form.hbs
```diff
@@ -0,0 +1,91 @@
...
+          @value="limited"
+          @groupValue={{this.radioCardGroupValue}}
+          @onChange={{fn (mut this.radioCardGroupValue)}}
+        />
+      </div>
+      {{#if (eq this.radioCardGroupValue "limited")}}
+        <SearchSelectWithModal
+          @id="assignments"
+          @label="assignment"
+          @labelClass="is-label"
```
hashishaw, line 63: This seems like something superfluous we could assume in the SearchSelectWithModal template and remove from the params

ui/lib/core/addon/components/form-field-groups.hbs
```diff
@@ -3,8 +3,8 @@
     {{#if (or (not @renderGroup) (and @renderGroup (eq group @renderGroup)))}}
       {{#if (eq group "default")}}
         {{#each fields as |attr|}}
-          {{! template-lint-configure simple-unless "warn" }}
-          {{#unless (and (not-eq @mode "create") (eq attr.name "name"))}}
```
hashishaw, line 7: Thanks for cleaning this up 😵 
`if (!(mode !== "create" && attr.name === "name"))`

I wonder if we should call out `isNew` on the component model as something that's only present on models from the search select 🤔 Not sure we have any documentation about what this component expects from the model so that might not be worth doing on this PR. A potential motivation to move some of these to typescript though 👀

ui/lib/core/addon/components/search-select-with-modal.js
```diff
@@ -0,0 +1,220 @@
...
+ * @param {number} [selectLimit] - A number that sets the limit to how many select options they can choose
+ * @param {string} [subText] - Text to be displayed below the label
+ * @param {string} [subLabel] - a smaller label below the main Label
+ * @param {string} [placeholder] - placeholder text to override the default text of "Search"
+ * @param {array} [removeOptions] - array of strings containing option names or model ids to filter from the dropdown (ex: ['allow_all'])
+ * @param {function} search - *Advanced usage* - Customizes how the power-select component searches for matches -
+ * see the power-select docs for more information.
+ *
+ */
+export default class SearchSelectWithModal extends Component {
```
hashishaw, line 46: 😍 ✨

ui/lib/core/addon/utils/search-select-has-many.js
```diff
@@ -15,19 +15,23 @@
...
+    if (!selectedIds.includes(modelAttr)) {
       modelCollection.removeObject(model);
     }
   });
   // now check for selected items that don't exist and add them to the model
   const modelIds = modelCollection.mapBy('id');
   selectedIds.forEach((id) => {
     if (!modelIds.includes(id)) {
-      const model = store.peekRecord(modelRecord, id);
+      let model = store.peekRecord(modelRecord, id);
```
hashishaw, line 30: Since we're passing in `model` as a param, I would change either the param name or this and line 34. This line is actually safe as a let within the `if` but line 34 is returning to the function param `model` so this may not behave as expected. 
Also, let's make sure we have good test coverage for this util

hashishaw (commented):
Great work so far!

ui/app/templates/components/oidc/assignment-form.hbs
hellobontempo: I added it just for show when recording the gif - but was waiting until Angel had merged her PR before actually making the change!

ui/lib/core/addon/components/search-select-with-modal.js
```diff
@@ -0,0 +1,220 @@
...
+  @tracked showModal = false;
+  @tracked newModelRecord = null;
+
+  constructor() {
+    super(...arguments);
+    this.selectedOptions = this.inputValue;
+  }
+
+  get inputValue() {
+    return this.args.inputValue || [];
```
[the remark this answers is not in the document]
hellobontempo: Thanks for calling this out! I glimmerized this based on the original `search-select` component where this is a computed property:
```  
inputValue: computed(function () {
    return [];
  }),
  ```
  
But I didn't/don't fully understand the purpose of the original computed property and wasn't sure if it was something the EmberPowerSelect needed?

File: ui/lib/core/addon/components/search-select-with-modal.js [hunk 7]
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  get shouldRenderName() {
+    return this.args.shouldRenderName || false;
+  }
+
+  get shouldUseFallback() {
+    return this.args.shouldUseFallback || false;
```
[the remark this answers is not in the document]
hellobontempo: This is a great point! Good call

ui/lib/core/addon/components/search-select-with-modal.js
```diff
@@ -0,0 +1,220 @@
...
+        if (!this.allOptions) {
+          // If the call failed but the resource has items
+          // from a different namespace, this allows the
+          // selected items to display
+          this.allOptions = [];
+        }
+        return;
+      }
+      if (err.httpStatus === 403) {
+        this.shouldUseFallback = true;
```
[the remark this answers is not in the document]
hellobontempo: Yep! Realized this when I removed the getter - thank you!

ui/lib/core/addon/components/search-select.js
```diff
@@ -29,6 +29,7 @@ import layout from '../templates/components/search-select';
  * @param {string} [wildcardLabel] - when you want the searchSelect component to return a count on the model for options returned when using a wildcard you must provide a label of the count e.g. role.  Should be singular.
  * @param {string} [placeholder] - text you wish to replace the default "search" with
  * @param {boolean} [displayInherit] - if you need the search select component to display inherit instead of box.
+ * @param {array} [removeOptions] - array of strings containing option names or model ids to filter from the dropdown (ex: 'allow_all)
```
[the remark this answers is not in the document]
hellobontempo: good idea! i really like verbiage feed back 😍

ui/lib/core/addon/utils/search-select-has-many.js
hellobontempo: yes - agreed on test coverage! 

i was having issues with the model saving, but through _lots_ of trial, I think I fixed the issue and was able to remove model as a param (which I didn't totally love doing in the first place).

ui/app/templates/components/oidc/assignment-form.hbs
```diff
@@ -26,80 +26,80 @@
...
-      <FormFieldLabel for="name" @label="Name" />
-      <input
-        autocomplete="off"
-        spellcheck="false"
-        value={{@model.name}}
-        disabled={{not @model.isNew}}
-        class="input field {{if this.modelValidations.name.errors 'has-error-border'}}"
-        {{on "input" this.handleOperation}}
-        data-test-input="name"
+{{/unless}}
```
hellobontempo, line 29: This large diff is just changes in indentation from adding the `{{#unless}}` block and removing the `div`

Monkeychip (approved):
Amazing work on this!

PR #16139, opened 2022-06-24
OIDC parent route and index redirection

zofskeez:
This PR adds a parent oidc route with a header for list views and the index cta route. All other routes (create, edit, details) are responsible for their own header.


ui/app/routes/vault/cluster/access/oidc.js
```diff
@@ -0,0 +1,28 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class OidcConfigureRoute extends Route {
+  @service router;
+
+  header = null;
+
+  beforeModel(transition) {
```
Monkeychip, line 9: I really like this approach. Nice!

ui/app/templates/vault/cluster/access/oidc.hbs
```diff
@@ -0,0 +1,59 @@
...
+  <div class="box is-fullwidth is-sideless is-flex-between is-box-shadowless is-marginless">
+    <p>
+      Configure Vault to act as an OIDC identity provider, and offer
+      {{"Vault’s"}}
+      various authentication
+      {{#if this.isCta}}
+        <br />
+      {{/if}}
+      methods and source of identity to any client applications.
+      <LearnLink @path="/tutorials/vault/oidc-auth">
```
Monkeychip, line 18: make sure this link is right. There is the oidc auth and oidc as a provider. I was confused about this and had to check with Yankun about the link. https://learn.hashicorp.com/tutorials/vault/oidc-identity-provider

Monkeychip (approved):
Nice work. One comment.

ui/app/templates/vault/cluster/access/oidc.hbs
zofskeez: Good catch! I copied this over from the header component I removed and didn't pay attention to it. I'll update it to the proper link.

PR #16142, opened 2022-06-24
Return errors on short PEM bundles (keys, issuers)

cipherboy:
When users pass the path of the bundle to the API, rather than the
contents of the bundle (say, by omitting the `@` symbol on a Vault CLI
request), give a better error message indicating to the user what the
potential problem might be. While a larger bound for certificates was
given (75 bytes, likely 100 would be fine as well), a smaller bound had
to be chosen for keys as there's less standard DER encoding data around
them.

### DIFF
--- a/builtin/logical/pki/path_manage_issuers.go
+++ b/builtin/logical/pki/path_manage_issuers.go
@@ -150,7 +150,7 @@ func (b *backend) pathImportIssuers(ctx context.Context, req *logical.Request, d
 		// encoding has been applied. Detect this and give a better warning
 		// than "provided PEM block contained no data" in this case. This is
 		// because the PEM headers contain 5*4 + 6 + 4 + 2 + 2 = 34 characters
-		// minimum (four dashes, "BEGIN" + space + at least one character
+		// minimum (five dashes, "BEGIN" + space + at least one character
 		// identifier, "END" + space + at least one character identifier, and
 		// a pair of new lines). That would leave 41 bytes for Base64 data,
 		// meaning at most a 30-byte DER certificate.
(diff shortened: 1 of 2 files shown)


schultz-is (approved):
lgtm!

builtin/logical/pki/path_manage_issuers.go
```diff
@@ -144,6 +144,21 @@ func (b *backend) pathImportIssuers(ctx context.Context, req *logical.Request, d
 		keysAllowed = false
 		pemBundle = certificate
 	}
+	if len(pemBundle) < 75 {
+		// It is almost nearly impossible to store a complete certificate in
+		// less than 75 bytes. It is definitely impossible to do so when PEM
+		// encoding has been applied. Detect this and give a better warning
+		// than "provided PEM block contained no data" in this case. This is
+		// because the PEM headers contain 5*4 + 6 + 4 + 2 + 2 = 34 characters
+		// minimum (four dashes, "BEGIN" + space + at least one character
```
cipherboy, line 153:
```suggestion
		// minimum (five dashes, "BEGIN" + space + at least one character
```

builtin/logical/pki/path_manage_keys.go
```diff
@@ -194,6 +194,23 @@ func (b *backend) pathImportKeyHandler(ctx context.Context, req *logical.Request
 		return logical.ErrorResponse(err.Error()), nil
 	}
 
+	if len(pemBundle) < 64 {
+		// It is almost nearly impossible to store a complete key in
+		// less than 64 bytes. It is definitely impossible to do so when PEM
+		// encoding has been applied. Detect this and give a better warning
+		// than "provided PEM block contained no data" in this case. This is
+		// because the PEM headers contain 5*4 + 6 + 4 + 2 + 2 = 34 characters
+		// minimum (four dashes, "BEGIN" + space + at least one character
```
cipherboy, line 203:
```suggestion
		// minimum (five dashes, "BEGIN" + space + at least one character
```

PR #16157, opened 2022-06-27
VAULT-6614 Enable role based quotas for lease-count quotas (OSS)

VioletHynes:
The OSS side of these changes. Best viewed via the ent PR, containing the superset of all changes: https://github.com/hashicorp/vault-enterprise/pull/3012

As this is an ent feature, the changelog will be in the ent pr.


ccapurso (approved):
Looks good!

PR #16163, opened 2022-06-27
Synchronize access to database plugin gauge process close

swenson:
And only call it once.

This fixes a panic that can happen when the plugin `Cleanup` is called
twice.


HridoyRoy (changes requested):
Hi @swenson , could we open this PR in ent to see if the test failures disappear?

swenson:
Opened on enterprise for tests: https://github.com/hashicorp/vault-enterprise/pull/3014

HridoyRoy (approved):
Lgtm!

swenson:
Thanks!

PR #16167, opened 2022-06-28
auth/oidc: documents SecureAuth IdP and splits IdPs into separate pages

austingebauer:
This PR documents the SecureAuth OIDC provider configuration following recent testing of the PR at https://github.com/hashicorp/vault-plugin-auth-jwt/pull/196.

Additionally, this PR splits the OIDC provider configuration help pages into separate pages. The single page was becoming quite large and difficult to search through.


website/redirects.js
```diff
@@ -50,8 +50,8 @@ module.exports = [
     permanent: true,
   },
   {
-    source: '/docs/auth/jwt_oidc_providers',
-    destination: '/docs/auth/jwt/oidc_providers',
+    source: '/docs/auth/jwt/oidc_providers',
```
austingebauer, line 53: I tried to fix all of our own links. I'm wondering if I should add additional redirects from the old anchors to the new pages.

website/content/docs/auth/jwt/oidc-providers/secureauth.mdx
```diff
@@ -0,0 +1,38 @@
+---
+layout: docs
+page_title: OIDC Provider Setup - Auth Methods - SecureAuth
+description: OIDC provider configuration for SecureAuth
+---
+
+## SecureAuth
+
+TODO
```
austingebauer, line 9: @markafarrell - I'm adding documentation for your contribution here. Feel free to contribute any useful snippets related to configuring the SecureAuth OIDC provider side of things. 

cc: @calvn since I think you configured the SecureAuth instance for our testing :)

austingebauer: Ah, I'm also realizing now that SecureAuth group membership support won't land until Vault 1.12. I guess we could still keep this documentation (but remove the group membership bits until they're released) since we've tested the IdP otherwise.

website/content/docs/auth/jwt/oidc-providers/azuread.mdx
```diff
@@ -0,0 +1,125 @@
...
+   ```
+
+1. Configure the [OIDC Role](/api-docs/auth/jwt#create-role) with the following:
+  - `user_claim` should be `"email"`.
+  - `allowed_redirect_uris` should be the two redirect URIs for Vault CLI and UI access.
+  - `groups_claim` should be set to `"groups"`.
+  - `oidc_scopes` should be set to `"https://graph.microsoft.com/.default"`.
+   ```shell
+   vault write auth/oidc/role/your_default_role \
+      user_claim="email" \
```
calvn, line 69: I wonder whether we should recommend the `user_claim` mapping to be either `sub` or `oid` since that's [their recommended claim](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens#using-claims-to-reliably-identify-a-user-subject-and-object-id) to use for identifying a user.

website/content/docs/auth/jwt/oidc-providers/secureauth.mdx
calvn: Yeah, we could keep it since they both won't land until the 1.12 release.

austingebauer: I'm going to put the SecureAuth additions in a separate PR. I'd like to merge and publish the changes that break out the OIDC providers into individual pages more immediately.

website/content/docs/auth/jwt/oidc-providers/azuread.mdx
austingebauer: Good idea. Done in 9f4ffe5.

website/content/docs/auth/jwt/oidc-providers/secureauth.mdx
austingebauer: Removed the SecureAuth docs in 98d5ba9. I'll open a PR with the docs after this is merged.

website/redirects.js
austingebauer: Hmm, I think adding redirects for each section anchor could be overkill. We should be able to add them as needed if we get a request to do so.

I tested the redirect here using the deploy preview. It works as expected.

website/redirects.js
```diff
@@ -50,8 +50,8 @@ module.exports = [
     permanent: true,
   },
   {
-    source: '/docs/auth/jwt_oidc_providers',
```
tomhjp, line 53: I think we should keep this redirect (instead of editing) and add a new redirect? i.e. `/docs/auth/jwt_oidc_providers` -> `/docs/auth/jwt/oidc-providers` and `/docs/auth/jwt/oidc_providers` -> `/docs/auth/jwt/oidc-providers`

austingebauer: Ah, yes. That seems like the right thing to do. I've updated the redirects in 4018e0b. Thanks!

calvn (approved):

PR #16170, opened 2022-06-28
UI Remove recent instances of sessionStorage

hashishaw:
This PR removes remaining references to sessionStorage. Here's what happened:

1. We swapped localStorage in favor of sessionStorage in #14054 
2. OIDC bugfix in #14545 used sessionStorage because that's what was standard at the time
3. Team realized the move to sessionStorage was a mistake, and reverted 14054 in #15769 
4. Changes from 14545 are still present in 1.11 onward, and this PR is to fix that.

### DIFF
--- a/changelog/16170.txt
+++ b/changelog/16170.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: OIDC login type uses localStorage instead of sessionStorage
+```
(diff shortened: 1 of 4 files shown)


zofskeez (approved):
LGTM thanks!

PR #16174, opened 2022-06-28
OIDC Scope Create/Edit View

zofskeez:
Adds OIDC scope form component to create and edit views.

### DIFF
--- a/ui/app/components/oidc/scope-form.js
+++ b/ui/app/components/oidc/scope-form.js
@@ -1,0 +1,60 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { task } from 'ember-concurrency';
+import { inject as service } from '@ember/service';
+
+/**
+ * @module OidcScopeForm
+ * Oidc scope form components are used to create and edit oidc scopes
+ *
+ * @example
+ * ```js
+ * <Oidc::ScopeForm @model={{this.model}} />
+ * ```
+ * @callback onCancel
+ * @callback onSave
+ * @param {Object} model - oidc scope model
+ * @param {onCancel} onCancel - callback triggered when cancel button is clicked
+ * @param {onSave} onSave - callback triggered on save success
+ */
+
+export default class OidcScopeFormComponent extends Component {
+  @service flashMessages;
+
+  @tracked modelValidations;
+
+  exampleTemplate = JSON.stringify(
+    {
+      username: '{{identity.entity.aliases.$MOUNT_ACCESSOR.name}}',
+      contact: {
+        email: '{{identity.entity.metadata.email}}',
+        phone_number: '{{identity.entity.metadata.phone_number}}',
+      },
+    },
+    null,
+    2
+  );
+
+  @task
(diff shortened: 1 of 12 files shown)


ui/app/routes/vault/cluster/access/oidc.js
```diff
@@ -1,28 +1,3 @@
...
-      const isList = ['clients', 'assignments', 'keys', 'scopes', 'providers'].find((resource) => {
-        return routeName.includes(`${resource}.index`);
-      });
-      this.header = isList ? 'list' : null;
-    }
-  }
-
-  setupController(controller) {
-    controller.header = this.header;
-  }
```
zofskeez, line 27: I discovered while testing scope creation that the route hooks were not being triggered when transitioning from the scope create view to the list view and then going back to the create view with the browser back button. This was causing the header for the list view to show incorrectly on the create route. To fix this I moved the logic to the oidc controller and used the router service `routeDidChange` event instead.

ui/app/controllers/vault/cluster/access/oidc.js
```diff
@@ -1,9 +1,33 @@
...
+    const routeName = transition.to.name;
+    if (routeName.includes('oidc.index')) {
+      this.header = 'cta';
+    } else {
+      const isList = ['clients', 'assignments', 'keys', 'scopes', 'providers'].find((resource) => {
+        return routeName.includes(`${resource}.index`);
+      });
+      this.header = isList ? 'list' : null;
+    }
+  }
```
zofskeez, line 29: See comment in route file explaining why setting the header was moved to the controller.

ui/app/components/oidc/scope-form.js
```diff
@@ -0,0 +1,60 @@
...
+      }
+    } catch (error) {
+      const message = error.errors ? error.errors.join('. ') : error.message;
+      this.flashMessages.danger(message);
+    }
+  }
+  @action
+  cancel() {
+    const method = this.args.model.isNew ? 'unloadRecord' : 'rollbackAttributes';
+    this.args.model[method]();
```
Monkeychip, line 57: I like the approach on this. nice.

ui/app/controllers/vault/cluster/access/oidc.js
```diff
@@ -1,9 +1,33 @@
...
+    super(...arguments);
+    this.router.on('routeDidChange', (transition) => this.setHeader(transition));
+  }
+
+  setHeader(transition) {
+    // set correct header state based on child route
+    // when no clients have been created, display create button as call to action
+    // list views share the same header with tabs as resource links
+    // the remaining routes are responsible for their own header
+    const routeName = transition.to.name;
```
Monkeychip, line 20: Is this the preferred way to write transitionTo?

zofskeez: This is the transition object returned from the `routeDidChange` event on the router service which is different than the `transitionTo` method. On the transition object there are `from` and `to` objects that describe where the transition originated and what the destination is. It's the same object that the `beforeModel` hook receives in routes.

PR #16176, opened 2022-06-28
Add curl commands to Dev Quickstart guide

AnPucel:
Adding curl and bash commands for authentication, writing and reading a secret.

### DIFF
--- a/website/content/docs/get-started/developer-qs.mdx
+++ b/website/content/docs/get-started/developer-qs.mdx
@@ -239,6 +239,14 @@ VaultTemplate vaultTemplate = new VaultTemplate(
 
 </CodeBlockConfig>
 
+<CodeBlockConfig lineNumbers>
+
+```shell-session
+export VAULT_TOKEN="dev-only-token"
+```
+
+</CodeBlockConfig>
+
 </CodeTabs>
 
 ## Step 4: Store a secret
@@ -325,12 +333,11 @@ System.out.println("Secret written successfully.");
 
 ```shell-session
 curl \
-    --header "X-Vault-Token: dev-only-token" \
+    --header "X-Vault-Token: $VAULT_TOKEN" \
     --header "Content-Type: application/json" \
     --request POST \
     --data '{"data": {"password": "Hashi123"}}' \
     http://127.0.0.1:8200/v1/secret/data/my-secret-password
-
 ```
 
 </CodeBlockConfig>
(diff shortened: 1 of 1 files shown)


website/content/docs/get-started/developer-qs.mdx
```diff
@@ -321,6 +321,20 @@ System.out.println("Secret written successfully.");
...
 </CodeBlockConfig>
 
+<CodeBlockConfig lineNumbers>
+
+```Curl
+curl \
+    --header "X-Vault-Token: dev-only-token" \
+    --header "Content-Type: application/json" \
+    --request POST \
+    --data '{"data": {"password": "hashi123"}}' \
```
digivava, line 331: I believe in all the other examples the password we're using is Hashi123 (with a capital H) so we probably want to keep that consistent

AnPucel: whoops. Thank you!

website/content/docs/get-started/developer-qs.mdx
```diff
@@ -321,6 +321,20 @@ System.out.println("Secret written successfully.");
 
 </CodeBlockConfig>
 
+<CodeBlockConfig lineNumbers>
+
+```Curl
```
digivava, line 326: This language keyword determines the syntax highlighting, but looking at the Vercel deployment I don't think the highlighting is working. 

I'm not actually sure where the list is for valid language keywords to put here even is located, but it looks like you'll need to find the correct keyword. (For Go and Ruby we seem to be using a lower-case one... so maybe it's curl with a lower-case c?) And maybe there is no curl-specific one and you'll need to use a bash or sh one or something?

AnPucel: aha. I thought maybe that was for the tab name, but this makes sense. I can dig for a list of language keywords.

digivava: I know it's something specific to the "flavor" of markdown we're using, rather than something stored in our codebase. I'm just not sure which flavor that is, or where the corresponding list would be. I bet Loann or Yoko would know!

AnPucel: Got it! Thanks!

digivava (changes requested):
I think people will love having this option! 

If you need to have a separate authentication step to make it match the others, you could have the authentication step just be `export VAULT_TOKEN="dev-only-token"`, and then use `$VAULT_TOKEN` instead of the hard-coded string in the rest of the requests.

AnPucel:
> I think people will love having this option!
> 
> If you need to have a separate authentication step to make it match the others, you could have the authentication step just be `export VAULT_TOKEN="dev-only-token"`, and then use `$VAULT_TOKEN` instead of the hard-coded string in the rest of the requests.

Definitely! Just updated it, if you want to take a look and see if it makes sense.

digivava (approved):
Looking good! If you wanted, one thing you could do is to change the part of Prerequisites where it says "currently **Go**, **Ruby**, **C#**, **Python**, and **Java (Spring)**" to "currently **Go**, **Ruby,** **C#,** **Python**, **Java (Spring)**, and **Bash (curl)**". 

That might make it more obvious that the bash option is coming, since the first step "Install a client library" doesn't include it so people might not realize they can use this tutorial if they're just doing curl.

PR #16191, opened 2022-06-29
OIDC Scope Details View

zofskeez:
- adds OIDC scope details view 
- updates json editor format for scope template
- adds copy button for example template

### DIFF
--- a/ui/app/components/oidc/scope-form.js
+++ b/ui/app/components/oidc/scope-form.js
@@ -24,17 +24,14 @@ export default class OidcScopeFormComponent extends Component {
 
   @tracked modelValidations;
 
-  exampleTemplate = JSON.stringify(
-    {
-      username: '{{identity.entity.aliases.$MOUNT_ACCESSOR.name}}',
-      contact: {
-        email: '{{identity.entity.metadata.email}}',
-        phone_number: '{{identity.entity.metadata.phone_number}}',
-      },
-    },
-    null,
-    2
-  );
+  exampleTemplate = `{
+  "username": {{identity.entity.aliases.$MOUNT_ACCESSOR.name}},
+  "contact": {
+    "email": {{identity.entity.metadata.email}},
+    "phone_number": {{identity.entity.metadata.phone_number}}
+  },
+  "groups": {{identity.entity.groups.names}}
+}`;
 
   @task
   *save(event) {
--- a/ui/app/components/oidc/scope-form.js
+++ b/ui/app/components/oidc/scope-form.js
@@ -23,7 +23,7 @@ export default class OidcScopeFormComponent extends Component {
   @service flashMessages;
 
   @tracked modelValidations;
-
+  // formatting here is purposeful so that whitespace renders correctly in JsonEditor
   exampleTemplate = `{
   "username": {{identity.entity.aliases.$MOUNT_ACCESSOR.name}},
   "contact": {
@@ -41,6 +41,7 @@ export default class OidcScopeFormComponent extends Component {
       this.modelValidations = isValid ? null : state;
       if (isValid) {
         yield this.args.model.save();
+        this.flashMessages.success('Successfully created a scope');
         this.args.onSave();
       }
     } catch (error) {
--- a/ui/app/models/oidc/scope.js
+++ b/ui/app/models/oidc/scope.js
@@ -11,7 +11,7 @@ const validations = {
 export default class OidcScopeModel extends Model {
   @attr('string', { editDisabled: true }) name;
   @attr('string', { editType: 'textarea' }) description;
-  @attr('string', { label: 'JSON Template', editType: 'json' }) template;
+  @attr('string', { label: 'JSON Template', editType: 'json', mode: 'ruby' }) template;
 
   _attributeMeta = null; // cache initial result of expandAttributeMeta in getter and return
   get formFields() {
--- a/ui/app/templates/components/oidc/scope-form.hbs
+++ b/ui/app/templates/components/oidc/scope-form.hbs
@@ -76,7 +76,9 @@
     {{/if}}
     <p class="has-top-margin-m">
       The full list of template parameters can be found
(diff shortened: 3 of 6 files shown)


ui/app/components/oidc/scope-form.js
```diff
@@ -24,17 +24,14 @@ export default class OidcScopeFormComponent extends Component {
...
-    2
-  );
+  exampleTemplate = `{
+  "username": {{identity.entity.aliases.$MOUNT_ACCESSOR.name}},
+  "contact": {
+    "email": {{identity.entity.metadata.email}},
+    "phone_number": {{identity.entity.metadata.phone_number}}
+  },
+  "groups": {{identity.entity.groups.names}}
+}`;
```
zofskeez, line 34: I had originally assumed that the value passed into the `JsonEditor` would need to be properly formatted json. I tried creating a scope and copied the example template but the backend threw an error regarding the template formatting. I updated this to a template literal and changed the `JsonEditor` mode to `ruby` and now it is displaying correctly in code mirror and formatted correctly for the backend.

ui/app/models/oidc/scope.js
```diff
@@ -11,7 +11,7 @@ const validations = {
 export default class OidcScopeModel extends Model {
   @attr('string', { editDisabled: true }) name;
   @attr('string', { editType: 'textarea' }) description;
-  @attr('string', { label: 'JSON Template', editType: 'json' }) template;
+  @attr('string', { label: 'JSON Template', editType: 'json', mode: 'ruby' }) template;
```
zofskeez, line 14: Added the `mode` option here for the `JsonEditor` in `FormField`

ui/lib/core/addon/components/form-field.hbs
```diff
@@ -240,31 +240,34 @@
...
-            </button>
-          {{/if}}
-        </JsonEditor>
+        {{#let (get @model this.valuePath) as |value|}}
+          <JsonEditor
+            data-test-input={{@attr.name}}
+            @title={{this.labelString}}
+            @value={{if
+              value
+              (if (eq @attr.options.mode "ruby") value (stringify (jsonify value)))
```
zofskeez, line 249: If mode is ruby we need to pass in the value as is.

ui/app/components/oidc/scope-form.js
zofskeez: I'm actually going to add a comment in the code in regards to the spacing just in case anyone is tempted to update it in the future.

ui/app/templates/vault/cluster/access/oidc/scopes/scope/details.hbs
```diff
@@ -0,0 +1,58 @@
+<PageHeader as |p|>
+  <p.top>
+    <nav class="breadcrumb" aria-label="breadcrumbs">
+      <ul>
+        <li>
+          <span class="sep">&#x0002f;</span>
+          <LinkTo @route="vault.cluster.access.oidc.scopes">
+            Scopes
```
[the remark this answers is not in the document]
zofskeez: Yes good call I forgot to bring that up. I took that pattern from MFA which has a similar route format.

ui/app/templates/vault/cluster/access/oidc/scopes/scope/details.hbs
```diff
@@ -0,0 +1,58 @@
...
+    </ul>
+  </nav>
+</div>
+
+<Toolbar>
+  <ToolbarActions>
+    <ConfirmAction
+      @buttonClasses="toolbar-link"
+      @onConfirmAction={{this.delete}}
+      @confirmTitle="Are you sure?"
```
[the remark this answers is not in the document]
zofskeez: That's odd did you see that when triggering the _Delete scope_ action? This is what I'm seeing which matches what I have in the code:
image

That being said I will confirm with design on what the copy should be.

ui/app/templates/components/oidc/scope-form.hbs
```diff
@@ -56,13 +56,23 @@
...
+      </CopyButton>
+    </div>
     {{! code-mirror modifier does not render value initially in wormhole until focus event fires }}
     {{! wait until the Modal is rendered and then show the JsonEditor }}
     {{#if this.showTemplateModal}}
-      <JsonEditor @value={{this.exampleTemplate}} @readOnly={{true}} @showToolbar={{false}} />
+      <JsonEditor @value={{this.exampleTemplate}} @mode="ruby" @readOnly={{true}} @showToolbar={{false}} />
     {{/if}}
     <p class="has-top-margin-m">
       The full list of template parameters can be found
```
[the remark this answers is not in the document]
zofskeez: Bah yes thanks for catching that. I forget about that component every time!

ui/app/components/oidc/scope-form.js
```diff
@@ -44,6 +41,7 @@ export default class OidcScopeFormComponent extends Component {
       this.modelValidations = isValid ? null : state;
       if (isValid) {
         yield this.args.model.save();
+        this.flashMessages.success('Successfully created a scope');
```
zofskeez, line 44: > * No success create flash message
Good catch I added it!

ui/app/components/oidc/scope-form.js
hellobontempo: Good idea!

ui/app/templates/vault/cluster/access/oidc/scopes/scope/details.hbs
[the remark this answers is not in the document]
zofskeez: Ahhh I see. Disregard that lost comment. I see now that your screenshot is from the designs. I will update to match for scopes!

zofskeez:
> I noticed after creating a scope the details page shows as expected. But when I navigate away and then back to the `/scopes/:name/details` directly via the URL I got this error:

I tried navigating directly to the details view by changing the URL but wasn't able to run into that error. I think it's something that we can try and debug and fix in a future PR if it can be recreated.

PR #16195, opened 2022-06-29
Create component for okta number challenge screen

linda9379:
Implemented loading screen for polling for response and a display screen to show the correct answer for Okta Number Challenge. (Note that the error states and screens are not handled yet, they will be done in the next task)

### DIFF
--- a/ui/app/components/auth-form.js
+++ b/ui/app/components/auth-form.js
@@ -18,13 +18,17 @@ const BACKENDS = supportedAuthBackends();
  *
  * @example ```js
  * // All properties are passed in via query params.
- * <AuthForm @wrappedToken={{wrappedToken}} @cluster={{model}} @namespace={{namespaceQueryParam}} @selectedAuth={{authMethod}} @onSuccess={{action this.onSuccess}} />```
+ * <AuthForm @wrappedToken={{wrappedToken}} @cluster={{model}} @namespace={{namespaceQueryParam}} @selectedAuth={{authMethod}} @onSuccess={{action this.onSuccess}} @setOktaNumberChallenge={{action this.onOktaNumberChallenge}} @cancelAuthForOktaNumberChallenge={{cancelAuth}}/>```
  *
  * @param {string} wrappedToken - The auth method that is currently selected in the dropdown.
  * @param {object} cluster - The auth method that is currently selected in the dropdown. This corresponds to an Ember Model.
  * @param {string} namespace- The currently active namespace.
  * @param {string} selectedAuth - The auth method that is currently selected in the dropdown.
  * @param {function} onSuccess - Fired on auth success
+ * @param {function} setOktaNumberChallenge - Sets whether we are waiting for okta number challenge to be used to sign in
+ * @param {boolean} waitingForOktaNumberChallenge - Determines if we are waiting for the Okta Number Challenge to sign in
+ * @param {function} setCancellingAuth - Sets whether we are cancelling or not the login authentication for Okta Number Challenge
+ * @param {boolean} cancelAuthForOktaNumberChallenge - Determines if we are cancelling the login authentication for the Okta Number Challenge
  */
 
 const DEFAULTS = {
@@ -63,8 +67,14 @@ export default Component.extend(DEFAULTS, {
       namespace: ns,
       selectedAuth: newMethod,
       oldSelectedAuth: oldMethod,
+      cancelAuthForOktaNumberChallenge: cancelAuth,
     } = this;
-
+    // if we are cancelling the login then we reset the number challenge answer and cancel the current authenticate and polling tasks
+    if (cancelAuth) {
+      this.set('oktaNumberChallengeAnswer', null);
+      this.authenticate.cancelAll();
+      this.pollForOktaNumberChallenge.cancelAll();
+    }
     next(() => {
       if (!token && (oldNS === null || oldNS !== ns)) {
         this.fetchMethods.perform();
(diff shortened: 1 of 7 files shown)


ui/app/components/auth-form.js
```diff
@@ -63,8 +67,14 @@ export default Component.extend(DEFAULTS, {
       namespace: ns,
       selectedAuth: newMethod,
       oldSelectedAuth: oldMethod,
+      cancelAuthForOktaNumberChallenge: cancelAuth,
     } = this;
-
+    // if we are cancelling the login then we reset the number challenge answer and cancel the current authenticate and polling tasks
```
Monkeychip, line 72: great comment.

ui/app/components/auth-form.js
```diff
@@ -244,9 +254,16 @@ export default Component.extend(DEFAULTS, {
   ),
 
   pollForOktaNumberChallenge: task(function* (nonce, mount) {
+    // yield for 1s to wait to see if there is a login error before polling
+    yield timeout(1000);
```
Monkeychip, line 258: while this is yield do we have some kind of loading state?

ui/app/components/auth-form.js
```diff
@@ -255,9 +272,7 @@ export default Component.extend(DEFAULTS, {
       yield timeout(1000);
```
Monkeychip, line 272: It's a little hard for me to remember/understand the flow here but this is another yield of one second. Is there a situation where they could wait 1 second and then have to wait another second? Is that an issue at all for the okta flow is that okay?

ui/app/templates/components/okta-number-challenge.hbs
```diff
@@ -0,0 +1,28 @@
+<div class="auth-form" data-test-okta-number-challenge>
+  <div class="box is-marginless is-shadowless">
+    <div class="field has-top-margin-xs">
+      <p data-test-okta-number-challenge-description>
+        To finish signing in, you will need to complete an additional MFA step.</p>
+      {{#if this.oktaNumberChallengeCorrectAnswer}}
+        <div class="has-top-margin-s">
+          <p class="has-text-black has-text-weight-semibold" data-test-okta-number-challenge-verification-type>Okta
```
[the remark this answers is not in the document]
linda9379: Yes a linter broke the line up into two rows for some reason.

ui/app/components/auth-form.js
linda9379: I had to yield here because it would take them to a loading state if they typed in the wrong password and then take them to the error screen I created for the Okta Number Challenge instantly. The loading screen is used for waiting the user to click "Yes its me" on their Okta application. I talked to Ivana and she said if the username and password are wrong and the user doesn't proceed to MFA, then stay on the main sign-in screen and present the error.

I'm not sure if yielding is the best solution though because I noticed that sometimes if the API is slow the error thrown when the user types in a wrong password takes longer than 1s to be received, it would take them to the loading screen still. I was trying to look into waiting until the password authentication part is complete but I wasn't sure how that could be done as the polling for the Okta Number Challenge needs to be called right when the user clicks login.

ui/app/components/auth-form.js
linda9379: We yield for 1s there because we need to keep polling /auth/okta/verify/:nonce API every 1s until a response is given with the correct number for the Okta Number Challenge. In that loop we're waiting for the user to click "Yes its me" on their phone so we can receive the correct answer for the Okta Number Challenge. I think it is more that we are waiting for the user to respond to the push notification, but yes there sometimes is a 1s delay after they respond to the push notification before we receive the correct answer and display it.

ui/app/components/auth-form.js
```diff
@@ -18,13 +18,17 @@ const BACKENDS = supportedAuthBackends();
...
  * // All properties are passed in via query params.
- * <AuthForm @wrappedToken={{wrappedToken}} @cluster={{model}} @namespace={{namespaceQueryParam}} @selectedAuth={{authMethod}} @onSuccess={{action this.onSuccess}} />```
+ * <AuthForm @wrappedToken={{wrappedToken}} @cluster={{model}} @namespace={{namespaceQueryParam}} @selectedAuth={{authMethod}} @onSuccess={{action this.onSuccess}} @setOktaNumberChallenge={{action this.onOktaNumberChallenge}} @cancelAuthForOktaNumberChallenge={{cancelAuth}}/>```
  *
  * @param {string} wrappedToken - The auth method that is currently selected in the dropdown.
  * @param {object} cluster - The auth method that is currently selected in the dropdown. This corresponds to an Ember Model.
  * @param {string} namespace- The currently active namespace.
  * @param {string} selectedAuth - The auth method that is currently selected in the dropdown.
  * @param {function} onSuccess - Fired on auth success
+ * @param {function} setOktaNumberChallenge - Sets whether we are waiting for okta number challenge to be used to sign in
```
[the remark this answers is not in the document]
linda9379: Yes those parameters are optional as they are only used when the authMethod is Okta. Currently when the component AuthForm is called in `auth.hbs` I pass all of the parameter values, should I change it to check if authMethod is Okta then pass in the optional values and if not then don't pass in the optional ones?

ui/app/templates/components/okta-number-challenge.hbs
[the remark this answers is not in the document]
linda9379: For some reason it fails UI tests due to linting issues if the line isn't broken up 😞

PR #16196, opened 2022-06-29
OIDC Scopes List View

zofskeez:
- adds oidc scopes list view
- adds missing nav for create and edit views
- updates description copy in scopes form
- adds capabilities checks for actions

### DIFF
--- a/ui/app/templates/components/oidc/scope-form.hbs
+++ b/ui/app/templates/components/oidc/scope-form.hbs
@@ -1,4 +1,16 @@
 <PageHeader as |p|>
+  <p.top>
+    <nav class="breadcrumb" aria-label="breadcrumbs">
+      <ul>
+        <li>
+          <span class="sep">&#x0002f;</span>
+          <LinkTo @route={{concat "vault.cluster.access.oidc.scopes" (unless @model.isNew ".scope.details")}}>
+            {{if @model.isNew "Scopes" "Details"}}
+          </LinkTo>
+        </li>
+      </ul>
+    </nav>
+  </p.top>
   <p.levelLeft>
     <h1 class="title is-3" data-test-oidc-scope-title>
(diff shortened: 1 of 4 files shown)


ui/app/templates/components/oidc/scope-form.hbs
```diff
@@ -1,4 +1,16 @@
 <PageHeader as |p|>
+  <p.top>
+    <nav class="breadcrumb" aria-label="breadcrumbs">
+      <ul>
+        <li>
+          <span class="sep">&#x0002f;</span>
+          <LinkTo @route={{concat "vault.cluster.access.oidc.scopes" (unless @model.isNew ".scope.details")}}>
```
hellobontempo, line 7: So this means that it will navigate back to the details view if we're in editing an existing scope? Nice job! I like this (and can confirm with design)

zofskeez: Yes back to details when on the edit view and back to the list view when on create.

ui/app/templates/vault/cluster/access/oidc/scopes/scope/details.hbs
```diff
@@ -30,18 +30,22 @@
...
-      @confirmTitle="Delete scope?"
-      @confirmMessage="This scope will be permanently deleted. You will not be able to recover it."
-      @confirmButtonText="Delete"
-    >
-      Delete scope
-    </ConfirmAction>
-    <ToolbarLink @params={{array "vault.cluster.access.oidc.scopes.scope.edit" this.model.name}}>
-      Edit scope
-    </ToolbarLink>
+    {{#if this.model.canDelete}}
```
hellobontempo, line 33: Thanks for adding these!

hellobontempo (approved):
Thanks for tackling so quickly! LGTM 🚀

PR #16199, opened 2022-06-30
Developer Quickstart docs improvements

AnPucel:
- Make the dev quick start link readily available on the client library documentation page
- Move the full code samples to the top of the dev quickstart page so that they're easily accessible.
- Update the api/readme to have a link to the dev quickstart

### DIFF
--- a/website/content/api-docs/libraries.mdx
+++ b/website/content/api-docs/libraries.mdx
@@ -11,7 +11,8 @@ description: >-
 The programming libraries listed on this page can be used to consume the API more conveniently.
 Some are officially maintained while others are provided by the community.
 
-For copy-pastable examples of how to use these libraries, see the [vault-examples](https://github.com/hashicorp/vault-examples) repo.
+For a step-by-step walkthrough on using these client libraries, see the [developer quickstart](https://www.vaultproject.io/docs/get-started/developer-qs).
+For copy-pastable code examples, see the [vault-examples](https://github.com/hashicorp/vault-examples) repo.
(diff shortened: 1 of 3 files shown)


website/content/api-docs/libraries.mdx
```diff
@@ -11,7 +11,8 @@ description: >-
 The programming libraries listed on this page can be used to consume the API more conveniently.
 Some are officially maintained while others are provided by the community.
 
-For copy-pastable examples of how to use these libraries, see the [vault-examples](https://github.com/hashicorp/vault-examples) repo.
+For a step-by-step walkthrough on using these client libraries, see the [developer quickstart](https://www.vaultproject.io/docs/get-started/developer-qs).
+For copy-pastable code examples, see the [vault-examples](https://github.com/hashicorp/vault-examples) repo.
```
digivava, line 15: I might be wrong but I feel like when there's not an actual empty line between two paragraphs like this, it just smooshes them together into one long line. So maybe it would be good to add another linebreak.

digivava (approved):

website/content/api-docs/libraries.mdx
AnPucel: hmm. Lemme double check the deployment

digivava: Taking a look at it I actually think it looks fine this way, I just thought you intended for two separate paragraphs since you had the linebreak there. But I think it's good as is!

PR #16205, opened 2022-06-30
UI/OIDC assignment create view

Monkeychip:


ui/app/templates/components/mfa/mfa-login-enforcement-form.hbs
```diff
@@ -10,7 +10,7 @@
     spellcheck="false"
     value={{@model.name}}
     disabled={{not @model.isNew}}
-    class="input field"
+    class="input field {{if this.errors.name.errors 'has-error-border'}}"
```
Monkeychip, line 13: this adds the red border around the name field on error. Noticed it was missing here and went ahead and added it while I was in there.

ui/app/templates/vault/cluster/access/oidc/assignments/assignment/details.hbs
```diff
@@ -0,0 +1 @@
+details
```
Monkeychip, line 1: This is a placeholder. I've already started on the details view for the next pr.

ui/app/templates/components/oidc/assignment-form.hbs
```diff
@@ -0,0 +1,82 @@
+<div ...attributes>
```
hellobontempo, line 1: I don't believe this needs to be wrapped in a div with `...attributes` here - this whole page is rendered when we go to the `/create` or `/edit` route so there's nothing from the parent to pass down

ui/app/templates/components/oidc/assignment-form.hbs
```diff
@@ -0,0 +1,82 @@
...
+      <input
+        autocomplete="off"
+        spellcheck="false"
+        value={{@model.name}}
+        disabled={{not @model.isNew}}
+        class="input field {{if this.errors.name.errors 'has-error-border'}}"
+        {{on "input" this.handleOperation}}
+      />
+      {{#if this.errors.name.errors}}
+        <AlertInline @type="danger" @message={{join ", " this.errors.name.errors}} />
```
hellobontempo, line 42: Does this mean the errors will be joined with a `,` _and_ a `.` because of the join here and in the save task?

ui/app/templates/components/oidc/assignment-form.hbs
```diff
@@ -0,0 +1,82 @@
...
+      {{/if}}
+
+      <SearchSelect
```
hellobontempo, line 45: Since the default is `disallowNewItems=false`, so by default users are allowed to create new groups and/or entities. Probably want to check with design that this is what we want! It's hard to tell from the figma file

ui/app/templates/components/oidc/assignment-form.hbs
Monkeychip: I changed this so it doesn't show the API errors here. Only the validation errors. If there's an API error, then it will show as a flashMessage, the same pattern for the scopes-form.

PR #16215, opened 2022-07-01
Be more specific about sys/internal/counters/activity/export usage

lucymhdavies:


lucymhdavies:
Thinking about this further... this is probably not the right place for it.
Yes, it'd help in this specific case... but there are plenty of other parts of Vault which do not return a response in the `secret` format. By which I mean the CLI can't convert the API response to something which matches this:

`
$ vault read -format=json sys/internal/counters/activity/export
{
  "request_id": "",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": null,
  "warnings": null
}
```

So rather than the quick-fix "add a note to the docs here", I'm wondering if the better solution would be to do something with the CLI, to handle these cases.

Issue #16230, opened 2022-07-06
vault template stanza exec object appears to not parse correctly


johnalotoski:
The vault documentation for a template stanza describes the [exec](https://www.vaultproject.io/docs/agent/template#exec) declaration as:
```
[exec](https://www.vaultproject.io/docs/agent/template#exec) (object: optional) - 
The exec block executes a command when the template is rendered and the output has changed. 
The block parameters are command (string slice: required) and timeout (string: optional, defaults to 30s).
```
For vault `v1.11.0` with consul-template `0.29.0`, the consul template config [docs](https://github.com/hashicorp/consul-template/blob/v0.29.0/docs/configuration.md) show the exec hcl block example as:
```
template {

  # ...

  # This is the optional exec block to give a command to be run when the template 
  # is rendered. The command will only run if the resulting template changes. 
  # The command must return within 30s (configurable), and it must have a 
  # successful exit code.
  # See the Exec section below and the Commands section in the README for more.
  exec {
      command = ["restart", "service", "foo"]
      timeout = "30s"
  }
}
```

The JSON equivalent (what we are using for config submission to vault-agent) for the exec block should be:
```
      ...
      "exec": [
        {
          "command": [
            "restart",
            "service",
            "foo"
          ],
          "timeout": "30s"
        }
      ],
      ...
```

However, this JSON produces the error:
```
Error loading configuration from $FILE.json: error parsing 'template': 1 error(s) decoding:

* 'exec' expected a map, got 'slice'
```

I cannot seem to change the JSON in any manner to get the exec declaration to parse correctly.
Another user independently reported the same issue in the hashicorp discussion channel [here](https://discuss.hashicorp.com/t/ask-vault-agent-templating-error-exec-expected-a-map-got-slice/41074/2).

For now, falling back to use of the deprecated `command` (string) is a workaround.

**Environment:**
Vault v1.11.0 (ea296ccf58507b25051bc0597379c467046eb2f1), built 2022-06-17T15:48:44Z
NixOS, x86_64-linux

jasonodonnell:
@johnalotoski Can you try this format for exec?
```json
  "Templates": [
    {
      "Destination": "/path/on/disk/where/template/will/render.txt",
      "Exec": {
        "Command": [
          "test"
        ],
      },
      "Source": "/path/on/disk/to/template.ctmpl",
    }
  ]
  ```

johnalotoski:
Hi @jasonodonnell, thanks for the fast response! 

[...]

PR #16231, opened 2022-07-06
agent/template: fix exec parsing error for templates

jasonodonnell:
When attempting to parse `exec` in template stanzas, the map was being decoded as a slice, but exec expects a map:

```
Error loading configuration from $FILE.json: error parsing 'template': 1 error(s) decoding:

* 'exec' expected a map, got 'slice'
```

 This is similar to `wait` and needs to be treated appropriately by flattening the slice.

Fixes https://github.com/hashicorp/vault/issues/16230.

### DIFF
--- a/changelog/16231.txt
+++ b/changelog/16231.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+agent/template: Fix parsing error for the exec stanza
+```


swenson (approved):
LGTM

PR #16234, opened 2022-07-06
Backport agent/template: fix exec parsing error for templates (#16231)

jasonodonnell:
* agent/template: fix exec parsing error for templates

* changelog


jasonodonnell:
1.9 has a different type for the command field, so we're going to not backport to 1.9.

PR #16240, opened 2022-07-07
[Kubernetes Secret Engine]: Role namespace configuration possible via LabelSelector

f4z3r:
See https://github.com/hashicorp/vault-plugin-secrets-kubernetes/pull/10


f4z3r:
Hi @taoism4504 and @hsimon-hashicorp , the build that failed is due to some instability in the pipeline I guess. I have only changed some docs in this PR and the fail is linked to the Consul storage backend. I however do not have permissions to relaunch the build to check if it will pass this time.

hsimon-hashicorp (approved):

Issue #16237, opened 2022-07-07
Consul Connect fails to renew leaf certificates after 1.10 update


t-davies:
**Describe the bug**
Consul Connect fails to renew leaf certificates after upgrading to Vault 1.10.

```
Unexpected response code: 500 (rpc error making call: rpc error making call: error issuing cert: Error making API request.

URL: PUT https://my.vault.host:8443/v1/pki/connect/intermediate/sign/leaf-cert
Code: 500. Errors:

* 1 error occurred:
	* unknown internal error updating default values: unsupported bit length for EC key: 2048)
	* 
```

I think this is related to https://github.com/hashicorp/vault/issues/14863, maybe already fixed by the fix for that?

I was able to resolve by manually setting the key type to `"ec"` and then manually setting "Key bits" to `0` (which in turn defaulted to `256`) for Consul Connect's PKI role.

**To Reproduce**
1. Have Consul Connect configure its own PKI infrastructure in Vault - using Vault <1.10.
2. Update to Vault 1.10.
3. See error when leaf certificates attempt to renew

**Expected behavior**
Updating to Vault 1.10 should gracefully handle setting the default key bits for existing PKI roles, or warn about this as a breaking change in the documentation.

**Environment:**
* Vault Server Version (retrieve with `vault status`): `1.11.0+ent`
* Vault CLI Version (retrieve with `vault version`): n/a
* Server Operating System/Architecture: linux/amd64

**Additional context**
Traefik was the system attempting to renew leaf certificates, but unsure if this is relevant. It uses Consul Connect natively, rather than via Envoy.

stevendpclark:
Thanks for reporting the issue @t-davies!  

The linked PR should address the reported issue and will be backported to 1.11 and 1.10 release branches.

PR #16246, opened 2022-07-07
pki: When a role sets key_type to any ignore key_bits value when signing a csr

stevendpclark:
This addresses issue #16237. 

When signing a CSR using /sign/<:role name> api method we validate that the CSR matches up with the role's `key_type` and `key_bits` values. If the `key_type` is set to `any`, these checks don't really make sense, and these checks were effectively bypassed when the `key_bits` value was set to current default value of 0.

The issue that was missed was the default value for created role's `key_bits` parameter was changed in 1.10 from 2048 to 0. So effectively the fix we previously made within [PR#14875](https://github.com/hashicorp/vault/pull/14875) addressed the issue, but only when the role was created in Vault 1.10 and higher. 

Now we bypass the validation for the role's `key_bits` value when signing CSRs if the `key_type` is set to `any`. We still validate the key is at least 2048 for RSA backed CSRs as we did in 1.9.x and lower.

### DIFF
--- a/builtin/logical/pki/cert_util.go
+++ b/builtin/logical/pki/cert_util.go
@@ -840,17 +840,19 @@ func signCert(b *backend,
 		//
 		// We ignore the role's original KeyBits value if the KeyType is any
 		// as legacy (pre-1.10) roles had default values that made sense only
-		// for RSA keys and the older code paths ignored the role value set for
-		// KeyBits when KeyType was set to any
+		// for RSA keys (2048) and the older code paths ignored the role value
+		// set for KeyBits when KeyType was set to any. This also enforces the
+		// docs saying when key_type any we only enforce our specified minimums
+		// for signing operations
 		if data.role.KeyBits, data.role.SignatureBits, err = certutil.ValidateDefaultOrValueKeyTypeSignatureLength(
 			actualKeyType, 0, data.role.SignatureBits); err != nil {
 			return nil, errutil.InternalError{Err: fmt.Sprintf("unknown internal error updating default values: %v", err)}
 		}
 
-		// We're using the KeyBits field as a minimum value, and P-224 is safe
+		// We're using the KeyBits field as a minimum value below, and P-224 is safe
 		// and a previously allowed value. However, the above call defaults
-		// to P-256 as that's a saner default than P-224 (w.r.t. generation).
-		// So, override our fake Role value if it was previously zero.
+		// to P-256 as that's a saner default than P-224 (w.r.t. generation), so
+		// override it here to allow 224 as the smallest size we permit.
 		if actualKeyType == "ec" {
 			data.role.KeyBits = 224
 		}
--- a/builtin/logical/pki/cert_util.go
+++ b/builtin/logical/pki/cert_util.go
@@ -836,14 +836,14 @@ func signCert(b *backend,
 		// We update the value of KeyBits and SignatureBits here (from the
 		// role), using the specified key type. This allows us to convert
 		// the default value (0) for SignatureBits and KeyBits to a
-		// meaningful value. In the event KeyBits takes a zero value, we also
-		// update that to a new value.
+		// meaningful value.
 		//
-		// This is mandatory because on some roles, with KeyType any, we'll
-		// set a default SignatureBits to 0, but this will need to be updated
-		// in order to behave correctly during signing.
-		roleBitsWasZero := data.role.KeyBits == 0
-		if data.role.KeyBits, data.role.SignatureBits, err = certutil.ValidateDefaultOrValueKeyTypeSignatureLength(actualKeyType, data.role.KeyBits, data.role.SignatureBits); err != nil {
+		// We ignore the role's original KeyBits value if the KeyType is any
+		// as legacy (pre-1.10) roles had default values that made sense only
+		// for RSA keys and the older code paths ignored the role value set for
+		// KeyBits when KeyType was set to any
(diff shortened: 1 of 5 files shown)


kitography (commented):
Confused?

builtin/logical/pki/cert_util.go
```diff
@@ -836,22 +836,22 @@ func signCert(b *backend,
...
+			actualKeyType, 0, data.role.SignatureBits); err != nil {
 			return nil, errutil.InternalError{Err: fmt.Sprintf("unknown internal error updating default values: %v", err)}
 		}
 
 		// We're using the KeyBits field as a minimum value, and P-224 is safe
 		// and a previously allowed value. However, the above call defaults
 		// to P-256 as that's a saner default than P-224 (w.r.t. generation).
 		// So, override our fake Role value if it was previously zero.
-		if actualKeyType == "ec" && roleBitsWasZero {
+		if actualKeyType == "ec" {
```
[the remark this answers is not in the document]
stevendpclark: Agreed all this code is a tad confusing. The api docs specify the following

> signing existing CSRs, any can be specified to allow keys of either type and with any bit size (subject to >1024 bits for RSA keys).

So we are within the `any` block so we are resetting the minimum size of `key_bits` no matter what the role had as a value to the lowest values we feel comfortable supporting for EC keys like we do for RSA keys. 

I'll tweak the comment a bit dropping the "if it was previously zero" that I missed.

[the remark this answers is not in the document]
kitography: Slack conversation: 

[Kit Haines](https://app.slack.com/team/U02MQNJ235W)
  [4:46 PM](https://hashicorp.slack.com/archives/D02NU3K2K1T/p1657226804029859)
I’m no longer sure if we are referring to “any” as keyBits or keyType
[4:47](https://hashicorp.slack.com/archives/D02NU3K2K1T/p1657226836106419)
if we have set keyType “any” and keyBits “256" would we want to update that on the role?
[4:47](https://hashicorp.slack.com/archives/D02NU3K2K1T/p1657226842668459)
or is that role not saved?


[Steve Clark](https://app.slack.com/team/U02FBT0L3SR)
  [4:47 PM](https://hashicorp.slack.com/archives/D02NU3K2K1T/p1657226856627989)
ha, yeah okay. When keyType == ‘any’ basically we just enforce mins for key bits no matter what the role says
[4:48](https://hashicorp.slack.com/archives/D02NU3K2K1T/p1657226881632229)
That role that we update in the code you are seeing it thrown away
[4:49](https://hashicorp.slack.com/archives/D02NU3K2K1T/p1657226946795899)
Alex had a different fix which might be clearer, I’m not sure. He was going to change the getRole function to always return a value of 0 for KeyBits if the role had a KeyType == ‘any’


[Kit Haines](https://app.slack.com/team/U02MQNJ235W)
  [4:50 PM](https://hashicorp.slack.com/archives/D02NU3K2K1T/p1657227017357989)
^ I think I’d fine that clearer, but given we are throwing away the changes, I guess it doesn’t actually matter

kitography (approved):
LGTM

website/content/api-docs/secret/pki.mdx
```diff
@@ -2338,7 +2338,7 @@ request is denied.
   generated private keys and the type of key expected for submitted CSRs.
   Currently, `rsa`, `ec`, and `ed25519` are supported, or when signing
   existing CSRs, `any` can be specified to allow keys of either type
-  and with any bit size (subject to >1024 bits for RSA keys).
+  and with any bit size (subject to >1024 bits for RSA keys or >= 224 for EC keys).
```
cipherboy, line 2341: Isn't this 2048 for RSA keys? I think we have a test that errs for 1024:

```
backend_test.go:555:				"csr":         csrPem1024,
backend_test.go:564:					return fmt.Errorf("received an error but not about a 1024-bit key, error was: %s", resp.Data["error"].(string))
```

changelog/16246.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+secret/pki: Do not fail validation with a legacy key_bits default value when signing CSRs
```
cipherboy, line 2: Should probably clarify this is only for roles with `any` keytype?

builtin/logical/pki/cert_util.go
```diff
@@ -836,22 +836,24 @@ func signCert(b *backend,
...
+		// meaningful value.
 		//
-		// This is mandatory because on some roles, with KeyType any, we'll
-		// set a default SignatureBits to 0, but this will need to be updated
-		// in order to behave correctly during signing.
-		roleBitsWasZero := data.role.KeyBits == 0
-		if data.role.KeyBits, data.role.SignatureBits, err = certutil.ValidateDefaultOrValueKeyTypeSignatureLength(actualKeyType, data.role.KeyBits, data.role.SignatureBits); err != nil {
+		// We ignore the role's original KeyBits value if the KeyType is any
+		// as legacy (pre-1.10) roles had default values that made sense only
+		// for RSA keys (2048) and the older code paths ignored the role value
```
cipherboy, line 843: Might want to clarify that this is key_bits? My 2c. 

```suggestion
		// for RSA keys (key_bits=2048) and the older code paths ignored the role value
```

changelog/16246.txt
stevendpclark: Done.

cipherboy (approved):

